PR #35004, opened 2025-03-11
Add hostname resolution for OCB ddprofiling extension

mackjmr:
### What does this PR do?

This PR adds hostname resolution for the ddprofiling extension in the OCB path. 

### Motivation
OTAGENT-302


comp/otelcol/ddprofilingextension/impl/extension.go
```diff
@@ -87,13 +107,39 @@ func (e *ddExtension) startForOCB() error {
...
+	if e.cfg.ProfilerOptions.Env != "" {
+		tags.WriteString(fmt.Sprintf(",default_env:%s", e.cfg.ProfilerOptions.Env))
+	}
+
+	if source.Kind == "host" {
+		profilerOptions = append(profilerOptions, profiler.WithHostname(source.Identifier))
+		tags.WriteString(fmt.Sprintf(",host:%s", source.Identifier))
+	}
+
+	if source.Kind == "task_arn" {
```
peterg17, line 130: I'm not sure if this works exactly, `task_arn` can be a tag on profiles emitted from classic ECS clusters (not Fargate ECS). We rely on `r.conf.FargateOrchestrator` in our agent code (see https://github.com/DataDog/datadog-agent/blob/067029a5c7b6ca88688f52be8ec68618a6313874/pkg/trace/api/profiles.go#L82) and i'm not sure if this is available here. I believe under the hood, this relies on env vars (https://github.com/DataDog/datadog-agent/blob/067029a5c7b6ca88688f52be8ec68618a6313874/pkg/util/fargate/detection.go#L21).

comp/otelcol/ddprofilingextension/impl/extension.go
```diff
@@ -87,13 +107,39 @@ func (e *ddExtension) startForOCB() error {
...
+		tags.WriteString(fmt.Sprintf(",orchestrator:fargate_ECS,task_arn:%s", source.Identifier))
```
peterg17, line 131: I would try to follow our implementation in profiles.go, which lowercases the orchestrator name (which could also be `fargate_eks`:
https://github.com/DataDog/datadog-agent/blob/067029a5c7b6ca88688f52be8ec68618a6313874/pkg/trace/api/profiles.go#L83

peterg17 (changes requested):

comp/otelcol/ddprofilingextension/impl/extension.go
mackjmr: I've lowercased it. Unfortunately, our hostname resolution does not support EKS fargate yet, only ECS fargate. I'm working on adding support for this (https://datadoghq.atlassian.net/browse/OTAGENT-316), and will make a PR to update this code once it's supported.

comp/otelcol/ddprofilingextension/impl/extension.go
peterg17: Ok - it's probably worth renaming that field since it is common to have non-Fargate with `task_arn` (classic ECS). But I understand that it's what you have to use in this case, so that should be a separate issue/PR probably.

comp/otelcol/ddprofilingextension/impl/extension.go
peterg17: Cool, thanks for the update!

peterg17 (approved):
I think it's ok after our conversations - but only testing / time will tell for sure 😅 . Might be worth testing a few configurations, which I know can be time consuming so up to you how necessary that is (not sure if the customers using this will be using Fargate). For example, we used this way back when with testing Fargate billing/usage: https://github.com/DataDog/profile-ruby-app-on-amazon-ecs-experiment

comp/otelcol/ddprofilingextension/impl/extension.go
mackjmr: Agreed, will rename once I add EKS fargate support

comp/otelcol/ddprofilingextension/impl/extension.go
```diff
@@ -87,13 +106,41 @@ func (e *ddExtension) startForOCB() error {
...
 
+	source, err := e.sourceProvider.Source(context.Background())
+	if err != nil {
+		return err
+	}
+
+	var tags strings.Builder
+	// hardcode agent version to 7.64.0. The backend requires this tag, but this path does not include agent,
+	// so hardcode value.
+	tags.WriteString("agent_version:7.64.0")
```
songy23, line 121: Is this expected to be the version moving forward?

comp/otelcol/ddprofilingextension/impl/extension_test.go
```diff
@@ -181,8 +181,67 @@ func TestOSSExtension(t *testing.T) {
...
+type fargateSourceProvider struct{}
+
+func (*fargateSourceProvider) Source(ctx context.Context) (source.Source, error) {
+	return source.Source{
+		Kind:       "task_arn",
+		Identifier: "arn:aws:ecs:us-east-1:123456789012:cluster/default",
+	}, nil
+}
+
+func TestOSSExtensionFargate(t *testing.T) {
```
songy23, line 193: Could you add a test on non fargate host?

mackjmr: Added a test for regular host.

comp/otelcol/ddprofilingextension/impl/extension.go
mackjmr: I've tested this and the version ends up appearing in the profiles, so I don't think this is a good approach, it could lead to confusion if you are using agent ex v 7.65 and the profiles show 7.64. I'll try to find a better way.

mackjmr: I've updated this in https://github.com/DataDog/datadog-agent/pull/35004/commits/bccba7da215087dcb7831e28745c02aa6dc6d400 to use the version of `github.com/DataDog/datadog-agent/comp/trace/agent/def ` instead, checking with profiling team if this is OK

songy23 (approved):

PR #35005, opened 2025-03-11
ddExtension.NotifyConfig check conf is not nil

ogaca-dd:
### What does this PR do?
This PR checks that `conf` in `ddExtension.NotifyConfig` is not nil.


truthbk (approved):
🙇

PR #35023, opened 2025-03-11
[AGENTRUN-173] Create diagnose component

GustavoCaso:
### What does this PR do?

This componetize the diagnose package

### Motivation

Continue working on encapsulating existing functionality in components. 
Also, now the diagnose component uses the API and Flare provider to inject its functionality using value groups

I've updated the flare component to accept a local diagnose `[]byte` to allow reducing the flare complexity for when it runs locally. It no longer have to create the diagnose result inside the component, but rather as part of the flare command. That way we can get rid a multiple components in the flare component 🎉 

Also, the API component no longer has to manually register the diagnose endpoint

I wanted to avoid creating a global registry of diagnoses, but for now it is simpler to do this way, and it introduces less changes on this PR. 

The cluster agent behaves a bit differently,  but I ensure the functionality remains the same 🤞  

I've added unit tests for the new component, still manual confirmation that:
- The `diagnose` command works as expected 
- The `flare` result is the same when running locally and when requesting to the server
- the `flare` for the cluster agent works as expected


vickenty (approved):

kkhor-datadog (approved):

iglendd (approved):
Looks good

L3n41c (approved):
LGTM for the subset of files owned by @DataDog/container-integrations

TCheruy (approved):

ajgajg1134 (approved):
APM changes look good to me and are very limited

cmd/cluster-agent/subcommands/start/command.go
```diff
@@ -281,8 +286,15 @@ func start(log log.Component,
 		)
 	}
 
+	// Register Diagnose functions
+	diagnosecatalog := diagnose.GetCatalog()
```
gabedos, line 290: Small nit (legibility): `diagnosecatalog` `=>` `diagnoseCatalog`

cmd/cluster-agent/subcommands/start/command.go
```diff
@@ -281,8 +286,15 @@ func start(log log.Component,
...
+
+	diagnosecatalog.Register("connectivity-datadog-autodiscovery", func(_ diagnose.Config) []diagnose.Diagnosis {
```
gabedos, line 292: Just a thought: For the catalogs being registered, are their names particularly important (ie what's the consequence of a typo)? I see you have these names somewhat defined in the comp/core/diagnose/def/component.go. Would it be worthwhile to define these as enum/constants and export them wherever needed?

```
type CatalogType string

const (
    AutodiscoveryConnectivity   CatalogType = "connectivity-datadog-autodiscovery"
    ...
)
```

gabedos (approved):
Looks good for cluster agent. Small comments

gh123man (approved):
ALP owned files LGTM.

cmd/cluster-agent/subcommands/start/command.go
GustavoCaso: That is a great point! 
I will do that 😄

cmd/cluster-agent/subcommands/start/command.go
GustavoCaso: Nice catch

comp/core/diagnose/impl/diagnose.go
```diff
@@ -0,0 +1,339 @@
...
+		err = format.Text(writer, diagnose.Config{Verbose: true}, diagnoseResult)
+	}
+
+	if err != nil {
+		return nil, err
+	}
+	writer.Flush()
+
+	return buffer.Bytes(), nil
+}
```
louis-cqrl, line 118: could we factorise this part to avoid code redundancy ?

louis-cqrl (approved):
🎉 Very nice work, left a nit comment feel free to apply or not

comp/core/diagnose/impl/diagnose.go
GustavoCaso: Done https://github.com/DataDog/datadog-agent/pull/35023/commits/a8c5d07f0bef99b3b9cf1e01c0880fda0a264a77

cmd/cluster-agent/subcommands/start/command.go
GustavoCaso: Done https://github.com/DataDog/datadog-agent/pull/35023/commits/a72e077bdd8e058a37f25063d7a15222e1e2d890

kangyili (approved):

sarah-witt (approved):

dustmop (approved):
LGTM, one small typo I noticed

PR #35029, opened 2025-03-11
Use dd-pkg to create signatures for linux packages

spencergilbert:
### What does this PR do?

Create detached signatures for linux packages using `dd-pkg`, this should not impact existing release tooling.

### Motivation

Support the use of centralized signing, and eventually remove access to the deb/rpm signing keys.

[Created custom build off this branch](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/58562170/), this runs the package deploy jobs which are successful and currently ignore the newly added signatures.

### DIFF
--- a/.gitlab/common/shared.yml
+++ b/.gitlab/common/shared.yml
@@ -47,7 +47,10 @@
   - GITHUB_INSTALLATION_ID=$($CI_PROJECT_DIR/tools/ci/fetch_secret.sh $AGENT_GITHUB_APP installation_id) || exit $?; export GITHUB_INSTALLATION_ID
   - echo "Using agent GitHub App"
 
-# Install `dd-pkg` and lint packages produced by Omnibus, supports only deb and rpm packages
+# Shared `dd-pkg` steps:
+# - Install `dd-pkg`, logging the version in use for the job
+# - Lint packages produced by Omnibus (supports only deb and rpm packages)
+# - Create detached signatures for packages produced by Omnibus, to be used by agent-release-management in downstream pipelines
 .create_signature_and_lint_linux_packages:
   - curl -sSL "https://dd-package-tools.s3.amazonaws.com/dd-pkg/${DD_PKG_VERSION}/dd-pkg_Linux_${DD_PKG_ARCH}.tar.gz" | tar -xz -C /usr/local/bin dd-pkg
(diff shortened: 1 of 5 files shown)


FlorentClarret (approved):

.gitlab/common/shared.yml
```diff
@@ -48,10 +48,12 @@
   - echo "Using agent GitHub App"
 
 # Install `dd-pkg` and lint packages produced by Omnibus, supports only deb and rpm packages
```
KevinFairise2, line 50: Maybe we want to update the comment

.gitlab/packaging/rpm.yml
```diff
@@ -13,7 +13,7 @@
     - RPM_SIGNING_PASSPHRASE=$($CI_PROJECT_DIR/tools/ci/fetch_secret.sh $RPM_SIGNING_PASSPHRASE) || exit $?; export RPM_SIGNING_PASSPHRASE
     - dda inv -e omnibus.build --release-version "$RELEASE_VERSION" --base-dir $OMNIBUS_BASE_DIR --skip-deps --target-project=${DD_PROJECT} ${OMNIBUS_EXTRA_ARGS}
```
KevinFairise2, line 14: Is it an issue if the omnibus.build task is already signing the package?

KevinFairise2 (approved):

.gitlab/packaging/rpm.yml
spencergilbert: No, the `dd-pkg sign` command is creating a detached signature that we'll eventually use to verify the contents of the package are as expected - _not_ the signature included in the package for end-users to check.

.gitlab/common/shared.yml
spencergilbert: Updated in [aee33f1](https://github.com/DataDog/datadog-agent/pull/35029/commits/aee33f1e0da0794c07e006810a996361c5660c3f)

.gitlab/packaging/rpm.yml
spencergilbert: So to clarify further - this process is separate to the "actual" package signing. Those changes will land later and will hopefully be more tightly scoped to each package (or at least package type).

PR #35031, opened 2025-03-11
[NPM-4291] Separate network_filter + ConnectionTuple into its own package

pimlu:
### What does this PR do?
This PR changes `network_filter` to use `ConnectionTuple` and moves them into their own separate packages.

### Motivation
We are going to use `network_filter` in the network path collector for dynamic paths.

Build should pass


pimlu:
Actually I might not need this, hold on...

PR #35032, opened 2025-03-11
[AGENTCFG-154] Adding an Agent metadata field to inform if the Agent is FIPS-enabled

rahulkaukuntla:
### What does this PR do?
We will stop reporting the `feature_fips_enabled` metadata field, and will use a new `fips_mode` field instead, which uses the `pkg/fips` module to determine if an agent has FIPS enabled. The actual REDAPL table will continue to have a `feature_fips_enabled` to accommodate legacy customers. 

Update: the [REDAPL table was updated](https://github.com/DataDog/infrastructure-resources/pull/1923).

### DIFF
--- a/comp/metadata/inventoryagent/inventoryagentimpl/inventoryagent.go
+++ b/comp/metadata/inventoryagent/inventoryagentimpl/inventoryagent.go
@@ -182,7 +182,7 @@ func (ia *inventoryagent) initData() {
 	ia.data["agent_startup_time_ms"] = pkgconfigsetup.StartTime.UnixMilli()
 	ia.data["flavor"] = flavor.GetFlavor()
 	if val, err := fips.Enabled(); err == nil {
-		ia.data["is_fips"] = val
+		ia.data["fips_mode"] = val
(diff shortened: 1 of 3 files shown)


comp/metadata/inventoryagent/inventoryagentimpl/inventoryagent.go
```diff
@@ -180,6 +181,11 @@ func (ia *inventoryagent) initData() {
 	ia.data["agent_version"] = version.AgentVersion
 	ia.data["agent_startup_time_ms"] = pkgconfigsetup.StartTime.UnixMilli()
 	ia.data["flavor"] = flavor.GetFlavor()
+	if val, err := fips.Enabled(); err == nil {
+		ia.data["fips_mode"] = val
+	} else {
```
hush-hush, line 186: We our metadata payload should be constant, so even it we get an error the field should be present in the payload.

hush-hush (changes requested):

hush-hush (approved):

PR #35044, opened 2025-03-12
Reduce some remote-tagger logs level

misteriaud:
### What does this PR do?
This PR reduces the log level from `info` to `debug` for some recurrent log messages and rephrases counter-intuitive logs.

### Motivation
The current implementation of the remote-tagger generates a large number of logs in a loop, which is counter-intuitive and creates unnecessary noise.
Example:
```
TRACE | INFO | (backoff@v2.2.1+incompatible/retry.go:37 in RetryNotify) | tagger stream established successfully
TRACE | WARN | (/usr/local/go/src/runtime/asm_amd64.s:1695 in goexit) | error received from remote tagger: rpc error: code = Unauthenticated desc = invalid auth token: Invalid session token
```

### DIFF
--- a/comp/core/tagger/impl-remote/remote.go
+++ b/comp/core/tagger/impl-remote/remote.go
@@ -589,11 +589,11 @@ func (t *remoteTagger) startTaggerStream(maxElapsed time.Duration) error {
 			Prefixes:    prefixes,
 		})
 		if err != nil {
-			t.log.Infof("unable to establish stream, will possibly retry: %s", err)
+			t.log.Debug("unable to establish stream, will possibly retry: %s", err)
 			return err
 		}
 
-		t.log.Info("tagger stream established successfully")
+		t.log.Debug("tagger stream successfully initialized")
 
 		return nil
 	}, expBackoff)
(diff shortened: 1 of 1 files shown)


wdhif (approved):

comp/core/tagger/impl-remote/remote.go
```diff
@@ -589,11 +589,11 @@ func (t *remoteTagger) startTaggerStream(maxElapsed time.Duration) error {
 			Prefixes:    prefixes,
 		})
 		if err != nil {
-			t.log.Infof("unable to establish stream, will possibly retry: %s", err)
+			t.log.Debug("unable to establish stream, will possibly retry: %s", err)
 			return err
 		}
 
-		t.log.Info("tagger stream established successfully")
+		t.log.Debug("tagger stream successfully initialized")
```
wdhif, line 596: Should this be kept as `Info`? Unless it does log too much?

misteriaud: Good point ! I moved it after the first successful request sent. I assert that the communication is established AND that it have been authenticated by the server.

wdhif (approved):

GustavoCaso (approved):

PR #35048, opened 2025-03-12
[EBPF] gpu: introduce the device manager

gjulianm:
### What does this PR do?

This PR introduces a single `DeviceCache` singleton that collects all devices from NVML. It also introduces a `Device` type that wraps `nvml.Device`, with some attributes pre-computed for speed.

### Motivation

Uniformize usage of NVML, cache arguments for performance improvements.

Unit tests added, existing ones should pass.

Synthetic benchmarks do not show a significant improvement as we mock NVML, but in real-life profiles the cost of calling NVML functions in the consumer loop was not negligible so we expect to see performance improvements there.

### DIFF
--- a/pkg/gpu/nvml/devices.go
+++ b/pkg/gpu/nvml/devices.go
@@ -22,6 +22,7 @@ type Device struct {
 	UUID      string
 	CoreCount int
 	Index     int
+	Memory    uint64
 }
 
 // NewDevice creates a new Device from an nvml.Device and caches some properties
@@ -47,12 +48,18 @@ func NewDevice(dev nvml.Device) (*Device, error) {
 		return nil, fmt.Errorf("error getting index: %s", nvml.ErrorString(ret))
 	}
 
+	memInfo, ret := dev.GetMemoryInfo()
+	if ret != nvml.SUCCESS {
+		return nil, fmt.Errorf("error getting memory info: %s", nvml.ErrorString(ret))
+	}
+
 	return &Device{
 		Device:    dev,
 		SMVersion: smVersion,
 		UUID:      uuid,
 		CoreCount: cores,
 		Index:     index,
+		Memory:    memInfo.Total,
 	}, nil
 }
 
--- a/pkg/gpu/nvml/devices.go
+++ b/pkg/gpu/nvml/devices.go
@@ -100,6 +100,7 @@ func NewDeviceCache() (DeviceCache, error) {
 func NewDeviceCacheWithOptions(nvmlLib nvml.Interface) (DeviceCache, error) {
 	cache := &deviceCache{
 		uuidToDevice: make(map[string]*Device),
+		smVersionSet: make(map[uint32]struct{}),
 	}
 
 	count, ret := nvmlLib.DeviceGetCount()
--- a/pkg/gpu/nvml/devices.go
+++ b/pkg/gpu/nvml/devices.go
@@ -11,7 +11,7 @@ package nvml
 import (
 	"fmt"
 
-	"github.com/NVIDIA/go-nvml/nvml"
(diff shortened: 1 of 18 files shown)


pkg/gpu/nvml/devices.go
```diff
@@ -0,0 +1,174 @@
...
+	// GetDeviceByIndex returns a device by its index
+	GetDeviceByIndex(index int) (*Device, error)
+	// DeviceCount returns the number of devices in the cache
+	DeviceCount() int
+	// GetSMVersionSet returns a set of all SM versions in the cache
+	GetSMVersionSet() map[uint32]struct{}
+	// GetAllDevices returns all devices in the cache
+	GetAllDevices() []*Device
+	// GetCores returns the number of cores for a device with a given UUID. Returns an error if the device is not found.
+	GetCores(uuid string) (uint64, error)
```
val06, line 81: let's remove the Device from the func names:

GetByUUID
GetByIndex
Count
SMVersionSet
All
Cores

pkg/gpu/nvml/devices.go
```diff
@@ -0,0 +1,174 @@
...
+	lib, err := GetNvmlLib()
+	if err != nil {
+		return nil, err
+	}
+
+	return NewDeviceCacheWithOptions(lib)
+}
+
+// NewDeviceCacheWithOptions creates a new DeviceCache with an already initialized NVML library
+func NewDeviceCacheWithOptions(nvmlLib nvml.Interface) (DeviceCache, error) {
```
val06, line 102: why do we need this ctor? if we have the nvmlLib singleton, why not enforcing usage of it?

pkg/gpu/nvml/devices.go
```diff
@@ -0,0 +1,174 @@
...
+		if ret != nvml.SUCCESS {
+			log.Warnf("error getting device by index %d: %s", i, nvml.ErrorString(ret))
+			continue
+		}
+
+		device, err := NewDevice(dev)
+		if err != nil {
+			log.Warnf("error creating device index %d: %s", i, err)
+			continue
+		}
```
val06, line 124: are we ok with having a partial list of devices? should we continue or report an error in this case? (how will this gap affect our metrics calculation logic?)

pkg/gpu/nvml/devices.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+// GetDeviceByUUID returns a device by its UUID
+func (c *deviceCache) GetDeviceByUUID(uuid string) (*Device, error) {
+	device, ok := c.uuidToDevice[uuid]
+	if !ok {
+		return nil, fmt.Errorf("device with uuid %s not found", uuid)
+	}
+	return device, nil
+}
```
val06, line 141: maybe keep the same interface as map get? return device, bool and not device, error ?

val06 (commented):
left some comments

pkg/gpu/nvml/devices.go
gjulianm: Changed all

pkg/gpu/nvml/devices.go
gjulianm: For tests, so that we can pass the nvml mock instead of setting the global singleton which is a bit more fragile.

pkg/gpu/nvml/devices.go
gjulianm: I'm following the logic we had in the core-check collector. I think it's better to have a partial list, the assumption is that if we're having failures here it's likely that CUDA itself will show failures and won't run anything on that device. But I'm not sure if it's even possible to get errors on these basic identification functions.

val06 (approved):

PR #33799, opened 2025-02-06
Serverless FIPS

apiarian-datadog:
### What does this PR do?
In conjunction with the datadog-lambda-extension build process intoroduced in https://github.com/DataDog/datadog-lambda-extension/pull/556 we are adding support for FIPS in the agent with goboring.

### Describe how you validated your changes
Reviewed tests here. Also deployed a wide array of self-monitoring lambda functions to a govcloud region talking to a datadog org in ddog-gov.

### Possible Drawbacks / Trade-offs
We chose to go with the goboring approach rather than the msgo approach because the latter would have required more substantial tooling changes for our build system.

PR #35051, opened 2025-03-12
[serverless] Use AWS FIPs endpoints when getting secrets

nhulston:
### What does this PR do?

If we detect a user is in a Govcloud region, use FIPs endpoints.

### Motivation

Make our Lambda extension FIPs compliant

I tested manually on Govcloud and on commercial.

Added a unit test

### Additional Notes

### DIFF
--- a/pkg/serverless/apikey/env.go
+++ b/pkg/serverless/apikey/env.go
@@ -18,11 +18,9 @@ import (
 // the appropriate AWS service. KMS, SM, etc.
 type decryptFunc func(string, aws.FIPSEndpointState) (string, error)
 
-func getSecretEnvVars(envVars []string, kmsFunc decryptFunc, smFunc decryptFunc) map[string]string {
-	awsRegion := os.Getenv(lambdaRegionEnvVar)
-	isGovRegion := strings.HasPrefix(awsRegion, "us-gov-")
+func getSecretEnvVars(envVars []string, kmsFunc decryptFunc, smFunc decryptFunc, shouldUseFips bool) map[string]string {
 	fipsEndpointState := aws.FIPSEndpointStateUnset
-	if isGovRegion {
+	if shouldUseFips {
 		fipsEndpointState = aws.FIPSEndpointStateEnabled
 		log.Debug("Govcloud region detected. Using FIPs endpoints for secrets management.")
 	}
@@ -75,7 +73,9 @@ func getSecretEnvVars(envVars []string, kmsFunc decryptFunc, smFunc decryptFunc)
 // DD_LOGS_CONFIGURATION, and will have dual shipping enabled without exposing
 // their API key in plaintext through environment variables.
 func setSecretsFromEnv(envVars []string) {
-	for envKey, envVal := range getSecretEnvVars(envVars, readAPIKeyFromKMS, readAPIKeyFromSecretsManager) {
+	awsRegion := os.Getenv(lambdaRegionEnvVar)
+	shouldUseFips := strings.HasPrefix(awsRegion, "us-gov-")
+	for envKey, envVal := range getSecretEnvVars(envVars, readAPIKeyFromKMS, readAPIKeyFromSecretsManager, shouldUseFips) {
 		os.Setenv(envKey, strings.TrimSpace(envVal))
 	}
 }
--- a/pkg/serverless/apikey/env.go
+++ b/pkg/serverless/apikey/env.go
@@ -6,6 +6,7 @@
 package apikey
 
 import (
+	"github.com/aws/aws-sdk-go-v2/aws"
 	"os"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


pkg/serverless/apikey/env.go
```diff
@@ -15,9 +16,19 @@ import (
...
+	awsRegion := os.Getenv(lambdaRegionEnvVar)
+	isGovRegion := strings.HasPrefix(awsRegion, "us-gov-")
+	var fipsEndpointState aws.FIPSEndpointState
+	if isGovRegion {
+		fipsEndpointState = aws.FIPSEndpointStateEnabled
+		log.Info("Govcloud region detected. Using FIPs endpoints for secrets management.")
+	} else {
+		fipsEndpointState = aws.FIPSEndpointStateUnset
+	}
+
```
duncanista, line 31:
```suggestion
    fipsEndpointState := aws.FIPSEndpointStateUnset
	if isGovRegion {
		fipsEndpointState = aws.FIPSEndpointStateEnabled
		log.Debug("Govcloud region detected. Using FIPs endpoints for secrets management.")
	}
```

hghotra (approved):
✅

duncanista (approved):

estherk15 (approved):

pkg/serverless/apikey/env.go
```diff
@@ -15,9 +16,17 @@ import (
...
-type decryptFunc func(string) (string, error)
+type decryptFunc func(string, aws.FIPSEndpointState) (string, error)
 
 func getSecretEnvVars(envVars []string, kmsFunc decryptFunc, smFunc decryptFunc) map[string]string {
+	awsRegion := os.Getenv(lambdaRegionEnvVar)
+	isGovRegion := strings.HasPrefix(awsRegion, "us-gov-")
+	fipsEndpointState := aws.FIPSEndpointStateUnset
+	if isGovRegion {
+		fipsEndpointState = aws.FIPSEndpointStateEnabled
+		log.Debug("Govcloud region detected. Using FIPs endpoints for secrets management.")
```
apiarian-datadog, line 27: can we do `Info` in this situation?

pkg/serverless/apikey/env.go
```diff
@@ -15,9 +16,17 @@ import (
...
 // decryptFunc is a function that takes in a value and retrieves it from
 // the appropriate AWS service. KMS, SM, etc.
-type decryptFunc func(string) (string, error)
+type decryptFunc func(string, aws.FIPSEndpointState) (string, error)
 
 func getSecretEnvVars(envVars []string, kmsFunc decryptFunc, smFunc decryptFunc) map[string]string {
+	awsRegion := os.Getenv(lambdaRegionEnvVar)
+	isGovRegion := strings.HasPrefix(awsRegion, "us-gov-")
+	fipsEndpointState := aws.FIPSEndpointStateUnset
+	if isGovRegion {
```
apiarian-datadog, line 25: we need to make a "fips-only" build of the agent. should this logic only be done there?

apiarian-datadog: this is the plan for the fips-only build: https://github.com/DataDog/datadog-agent/pull/33799

pkg/serverless/apikey/env.go
nhulston: cc @duncanista 

apparently `Info` increases billing for customers

pkg/serverless/apikey/env.go
nhulston: Yeah that makes sense to me. If there's a simple way to detect if we're in the `fips-only` build, we should just use that instead of parsing the region.

should i go ahead with this PR and make the change later, or wait for the `fips-only` build first?

pkg/serverless/apikey/env.go
duncanista: And it spams

pkg/serverless/apikey/env.go
apiarian-datadog: feel free to merge this but if you can extract the "should we fips" logic into a function, we can implement that function as a "false" normally and move the "true" into the serverlessfips build later.

pkg/serverless/apikey/env.go
apiarian-datadog: fair enough. for our testing we can always turn on debug mode. i'd be more concerned if there was some way for a gov region *not* to be fips, in which case i'd push more to have a message to make it clear which mode we're using.

how often do we hit this code? are we pulling the api key every time the lambda runs?

nhulston: There could be cases where customers are running in a Govcloud region, but they don't care about FIPs right? (I think you told me that?)

No, this only runs once per execution context. After we get the API key, it's just stored in memory

PR #35072, opened 2025-03-12
Fix flaky OTel e2e tests using telemetrygen

liustanley:
### What does this PR do?
Fix flaky OTel tests that are using telemetrygen. Traces were not being consistently received from telemetrygen, so this PR modifies tests using telemetrygen to use calendar-go instead.


test/new-e2e/tests/otel/otel-agent/config/sampling.yml
```diff
@@ -17,7 +17,7 @@ exporters:
 processors:
   batch:
   probabilistic_sampler:
-    sampling_percentage: 50
+    sampling_percentage: 100
```
truthbk, line 20: Are really testing sampling if we sample at `100%`?

liustanley: Thanks for catching this, reverted this change.

truthbk (approved):

PR #35075, opened 2025-03-13
Index parameter captures in output buffer of dynamic instrumentation snapshots

grantseltzer:
### What does this PR do?

Adds logic where each parameter that's captured in BPF records at what index in the buffer it's content starts at. This allows for individual parsing of the parameters which benefits reliability of partial snapshot captures. In other words, if there's an issue parsing one parameter, the others can still be captured.

It also adds logic to properly enforce a parameter limit of 20.

### Motivation

Better reliability of partial snapshot captures

Updated tests.

### Additional Notes

This also enables us to add logic where we can fallback from verifier/compilation errors by disabling one/more parameters but not others.

### DIFF
--- a/pkg/dynamicinstrumentation/codegen/c/base_event.h
+++ b/pkg/dynamicinstrumentation/codegen/c/base_event.h
@@ -2,14 +2,15 @@
 #define DI_BASE_EVENT_H
 
 #include "ktypes.h"
+#include "macros.h"
 
 // standard fields which all events created in bpf will contain, regardless of the function that the
 // probe is instrumenting
 struct base_event {
     __u32 pid; // process ID
     __u32 uid; // user ID
-    __u64 program_counters[10]; // program counters representing the stack trace of the instrumented function invocation
-    __u64 param_indicies[20]; // indicies of where each parameter starts in argument buffer
+    __u64 program_counters[STACK_DEPTH_LIMIT]; // program counters representing the stack trace of the instrumented function invocation
+    __u64 param_indicies[MAX_FIELD_AND_PARAM_COUNT]; // indicies of where each parameter starts in argument buffer
     char probe_id[36]; // identifier for each user-configured instrumentation point, it's a standard 36 character UUID
 };
 
--- a/pkg/dynamicinstrumentation/codegen/c/base_event.h
+++ b/pkg/dynamicinstrumentation/codegen/c/base_event.h
@@ -10,7 +10,7 @@ struct base_event {
     __u32 uid; // user ID
     __u64 program_counters[10]; // program counters representing the stack trace of the instrumented function invocation
     char probe_id[36]; // identifier for each user-configured instrumentation point, it's a standard 36 character UUID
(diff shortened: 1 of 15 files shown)


pkg/dynamicinstrumentation/diconfig/binary_inspection.go
```diff
@@ -60,6 +60,16 @@ func AnalyzeBinary(procInfo *ditypes.ProcessInfo) error {
 
 	procInfo.TypeMap = typeMap
 
+	// Enforce limit on number of parameters
+	for funcName := range procInfo.TypeMap.Functions {
+		for i, param := range procInfo.TypeMap.Functions[funcName] {
+			if i >= 20 {
```
grantseltzer, line 66: TODO: use a constant for this

pkg/dynamicinstrumentation/codegen/c/base_event.h
```diff
@@ -10,6 +10,7 @@ struct base_event {
     __u32 uid; // user ID
     __u64 program_counters[10]; // program counters representing the stack trace of the instrumented function invocation
     char probe_id[36]; // identifier for each user-configured instrumentation point, it's a standard 36 character UUID
+    __u64 param_indicies[20];
```
grantseltzer, line 13: TODO: use a constant for length of this array.

pkg/dynamicinstrumentation/ditypes/analysis.go
```diff
@@ -56,6 +56,21 @@ const (
 	CaptureDepthReached                             // CaptureDepthReached means the parameter wasn't captures because the data type has too many levels
 )
 
+func (r NotCaptureReason) String() string {
```
tylfin, line 59: What's the point of having `NotCaptureReason uint8` and not `NotCaptureReason string`?

pkg/dynamicinstrumentation/ditypes/ebpf_linux.go
```diff
@@ -7,8 +7,9 @@ type BaseEvent struct {
 	Pid              uint32
 	Uid              uint32
 	Program_counters [10]uint64
+	Param_indicies   [20]uint64
 	Probe_id         [36]byte
 	Pad_cgo_0        [4]byte
 }
 
-const SizeofBaseEvent = 0x80
+const SizeofBaseEvent = 0x120
```
tylfin, line 15: Is it worth dynamically calculating this to reduce the probability that we forget to update it?

```suggestion
var SizeofBaseEvent = unsafe.Sizeof(BaseEvent{})
```

pkg/dynamicinstrumentation/ditypes/analysis.go
grantseltzer: It's a good point. Memory efficiency and type safety are generally benefits to uint8 but not very relevant in this case. I'm going to leave it for consistency for now.

tylfin (approved):

PR #33232, opened 2025-01-22
[ASCII-2580] Implement function to safely fetch or create artifacts concurrently

misteriaud:
### What does this PR do?
This PR improves the mechanism for creating and fetching authentication artifacts used for inter-process communication between Agents. It introduces a more robust and fault-tolerant approach that eliminates race conditions, maintains process independence, and ensures reliable artifact retrieval.

### Motivation
The existing authentication mechanism relies on a strict startup order where the coreAgent generates artifacts before subAgents start. Initially, this worked well because subAgents acted only as clients. However, with subAgents now also acting as servers, this assumption has become problematic.

In reality, agent processes do not follow a guaranteed startup sequence, leading to race conditions when subAgents attempt to load authentication artifacts before they are created by the coreAgent. This results in failures, inconsistent behaviors, and unnecessary complexity.

Several constraints make this problem particularly challenging:
- **Configurable artifact paths**: Authentication artifact locations are configurable, preventing pre-generation at installation.
- **Independent startup model**: The system was designed so that Agents can start independently without strict dependencies. In host-based deployments, the coreAgent is historically the last to start, making it even more critical for subAgents to handle authentication artifact retrieval reliably.
- **Robustness and fault tolerance**: Simple retries or waiting for the coreAgent would introduce tighter coupling between processes, reducing flexibility and making the system more fragile.

Previous workarounds, such as lazy-loading or background fetching, have led to unreliable behavior and inconsistent patterns. A more structured approach is necessary to ensure subAgents can reliably obtain authentication artifacts without unnecessary dependencies.

### Implementation Description

1. **Optimistic Fetching:**
   - Attempts to retrieve the artifact from disk.
   - If valid, it is returned immediately.

2. **Lock-Based Creation:**
   - Uses a temporary file to prevent partial artifacts if creation fails.
   - If the artifact is missing, a file-based write lock is acquired.
   - Processes can independently verify if the artifact is valid, avoiding read locks.

3. **Retry-Based Fetching:**
   - If another process holds the lock, retries until the artifact is available.
   - Ensures reliable retrieval without race conditions.

### Describe how you validated your changes

I added unit tests covering the core mechanism. e2e tests involving CoreAgent and TraceAgent ensure real-world reliability, as they are currently the only processes authorized to create `auth_token`.

### Possible Drawbacks / Trade-offs

This mechanism relies on OS-level file locking. In worst-case scenarios, Agent initialization may be delayed if another process locks the file and takes time to release it.

#### Kubernetes Compatibility
[...]

PR #35081, opened 2025-03-13
[AGENTRUN-139] Allow processAgent to create auth artifacts

misteriaud:
### What does this PR do?
This PR enable the processAgent run command to create the auth_artifacts if they are not present yet.
This was made possible by:
- #33232
- https://github.com/DataDog/helm-charts/pull/1743

Additionally, it increases the default timeout duration from 10 seconds to 30 seconds. This change addresses issues in Kubernetes deployments where subAgents have the auth-token volume mounted as read-only, requiring them to wait for the coreAgent to generate the artifacts. The previous 10-second timeout was occasionally insufficient given the observed time between container startups.

### Motivation
This PR is part of the effort to reduce uncertainty and lack of determinism in the artifact creation step.

### DIFF
--- a/cmd/process-agent/command/main_common.go
+++ b/cmd/process-agent/command/main_common.go
@@ -17,7 +17,7 @@ import (
 	"github.com/DataDog/datadog-agent/cmd/agent/common/misconfig"
 	"github.com/DataDog/datadog-agent/comp/agent/autoexit"
 	"github.com/DataDog/datadog-agent/comp/agent/autoexit/autoexitimpl"
-	"github.com/DataDog/datadog-agent/comp/api/authtoken/fetchonlyimpl"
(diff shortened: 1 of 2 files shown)


misteriaud:

daniel-taf (approved):

cmd/process-agent/command/main_common.go
```diff
@@ -156,7 +156,7 @@ func runApp(ctx context.Context, globalParams *GlobalParams) error {
 		}),
 
 		// Provide authtoken module
-		fetchonlyimpl.Module(), // TODO IPC: replace by createandfetchimpl when the auth-token volume is mounted on processAgent container on Windows
```
dustmop, line 159: 🥳

dustmop (approved):
LGTM for agent-configuration

PR #35082, opened 2025-03-13
fix(dependabot): Enable automatic review request on dependency PR

chouetz:
### What does this PR do?
- Split the initial `add_reviewers_and_ask_review` in 2 different workflows
- Consider also renovate bot to ask review.
- Update the add_review tests to rely on stable test data

### Motivation
The original workflow was not working, because when a PR is triggered from a bot, a subsequent workflow will not be triggered. In our condition:
- if we add the `ask-review` with the GITHUB_TOKEN on dependabot PR, the `ask-review` job of the `label-analysis` workflow is [skipped](https://github.com/DataDog/datadog-agent/actions/runs/13831439753/job/38696397129?pr=35080). 
- if we try to use a token from a github app this is not possible because PR from bots are considered like forks (so they don't have access to secrets)
As a consequence the solution is:
- to have a dedicated workflow using GITHUB_TOKEN to set reviewers
- then trigger separate workflow run with the Github app token
- thus the `ask-review` will add reviewers

The goal is to have automatic reviewer addition on dependency update PR to have a fast review on these PR

Validated with a similar configuration on a personal repository (see [here](https://github.com/chouetz/psychic-guacamole/pull/79#issuecomment-2718296391), this comment was added because the label addition is done by Github App)

### DIFF
--- a/tasks/issue.py
+++ b/tasks/issue.py
@@ -124,4 +124,3 @@ def add_reviewers(ctx, pr_id, dry_run=False):
         return
     # Teams are added by slug, so we need to remove the @DataDog/ prefix
     pr.create_review_request(team_reviewers=[owner.casefold().removeprefix("@datadog/") for owner in owners])
-    pr.add_to_labels("ask-review")
--- a/tasks/issue.py
+++ b/tasks/issue.py
@@ -74,7 +74,7 @@ def ask_reviews(_, pr_id):
 
 
 @task
-def add_reviewers(ctx, pr_id, dry_run=False):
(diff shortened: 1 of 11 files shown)


tasks/issue.py
```diff
@@ -74,7 +74,7 @@ def ask_reviews(_, pr_id):
 
 
 @task
-def add_reviewers(ctx, pr_id, dry_run=False):
+def add_reviewers(ctx, pr_id, dry_run=False, owner_file=".github/CODEOWNERS"):
```
chouetz, line 77: All updates from here are to make tests not relying on the "official" go code structure, to have them more robust

CelianR (approved):

PR #35083, opened 2025-03-13
[AGENTRUN-177] create grpc component

GustavoCaso:
### What does this PR do?

This PR add the gRPC server component. 

It allows to decouple the gRPC server logic from the API component.

In combination with this PR https://github.com/DataDog/datadog-agent/pull/35023 it removes all the dependencies of the API. So to register routes you have to use the API Providers via components or if you need to add a gRPC route, you would need to modify the gRPC server implementation.

The gRPC server component is abstract enough to allow for new implementations for other API. For example if the cluster agent binary migrates to use the API and gRPC component it could easily create a new grpc server implementation defining the endpoints it want to support. 

### Motivation

Continue the work of decoupling the different components 😄 

Added Unit Test to validate the changes are working as expected 😄


guyarb (approved):

davidor (approved):
Nothing really to review from container-platform. Only one file with a minor change :+1:

louis-cqrl (approved):

vickenty (approved):
LGTM for AMP changes, although I'd like to suggest bit more descriptive name for the grpc component: e.g. `comp/api/grpcserver`. The new component contains code specific to agent's api authenticated and not a general purpose grpc server (like the one one gets by using grpc directly), its name should probably reflect that.

dustmop (approved):
LGTM for agent-configuration

sarah-witt (approved):

PR #35098, opened 2025-03-13
fix(protobuf): Improve the check for the pre-commit hook

chouetz:
### What does this PR do?
Update the test so that it allows doing a commit.

### Motivation
With the current version the check will always fail in the pre-commit hook if the files are not committed leading to a dead end. Now it updates the files and compares the state before/after to allow update on protocol buffer files.

Tested an update on protoc 30.0


tasks/libs/common/git.py
```diff
@@ -47,6 +47,18 @@ def get_staged_files(ctx, commit="HEAD", include_deleted_files=False) -> Iterabl
                 yield file
 
 
+def get_unstaged_files(ctx, re_filter=None) -> Iterable[str]:
+    """
+    Get the list of unstaged files in the repository.
+    """
+
+    files = ctx.run("git diff --name-only", hide=True).stdout.splitlines()
```
CelianR, line 55: Question: Should we include deleted files? (or add an option to include them or not)
Maybe you can rely on `get_file_modifications` / modify it to not use `git merge-base`

tasks/protobuf.py
```diff
@@ -105,8 +109,8 @@ def generate(ctx, do_mockgen=True):
                     f"protoc -I{proto_root} -I{protodep_root} --grpc-gateway_out=logtostderr=true:{repo_root} {targets}"
                 )
 
-        # Mockgen
-        if do_mockgen:
+        # Mockgen (no done in pre-commit as it is slow)
```
CelianR, line 112:
```suggestion
        # Mockgen (not done in pre-commit as it is slow)
```

CelianR (approved):

tasks/libs/common/git.py
chouetz: I suppose I can ignore them with an option like `get_staged_files`

PR #35100, opened 2025-03-13
Add separators for large numbers

ajgajg1134:
### What does this PR do?
Add separators to a variety of large int literals that we own.

While making the change I happened to run into some dead code in testutils and so removed `TestSpan()`

### Motivation
Separators improve human readability. We have previously missed changes to literals like these that resulted in breaking changes, with improved readability human reviewers are more likely to _see_ when a value changes, especially for large numbers like `1000000` (e.g. can you at a glance tell what that number is? I know I certainly can't. Is it the same as `100000`?)

I added a custom Datadog Code Security rule that looks for these and suggests fixes. On making these changes I ensured that the values stayed the same. Unit tests and visual verification should be sufficient here.

Maybe someone out there doesn't like these separators?

### DIFF
--- a/pkg/trace/testutil/span.go
+++ b/pkg/trace/testutil/span.go
@@ -371,8 +371,8 @@ func GetTestSpan() *pb.Span {
 		Service:  "fennel_IS amazing!",
 		Name:     "something &&<@# that should be a metric!",
 		Resource: "NOT touched because it is going to be hashed",
-		Start:    9223372036854775807,
-		Duration: 9223372036854775807,
+		Start:    9_223_372_036_854_775_807,
+		Duration: 9_223_372_036_854_775_807,
 		Meta:     map[string]string{"http.host": "192.168.0.1"},
 		Metrics:  map[string]float64{"http.monitor": 41.99},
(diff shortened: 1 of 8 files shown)


ekump (approved):

pkg/trace/testutil/span.go
```diff
@@ -393,26 +393,3 @@ func GetTestSpan() *pb.Span {
 	traceutil.ComputeTopLevel(trace)
 	return trace[0]
 }
-
-// TestSpan returns a fix span with hardcoded info, useful for reproducible tests
-func TestSpan() *pb.Span {
```
truthbk, line 398: Any reason to remove this? Unused?

truthbk (approved):
Looks good, just calling out the `TestSpan()` removal to make sure it was intentional, and understand why.

pkg/trace/testutil/span.go
ajgajg1134: Ah I'll update the description - yeah in modifying this code I noticed that this function was unused so removed it.

PR #35101, opened 2025-03-13
Auto Multiline V2 - Correctly handle custom samples as JSON from the config or an env var

gh123man:
### What does this PR do?

It turns out our config parser can't automatically handle JSON strings when unmarshalling, and this needs to be explicitly handled. This PR makes the parsing behavior consistent for `DD_LOGS_CONFIG_AUTO_MULTI_LINE_DETECTION_CUSTOM_SAMPLES` with other complex config keys that support JSON in env vars. 

Unit test coverage, Local sanity test.

### DIFF
--- a/pkg/logs/internal/decoder/auto_multiline_detection/user_samples.go
+++ b/pkg/logs/internal/decoder/auto_multiline_detection/user_samples.go
@@ -7,6 +7,7 @@
 package automultilinedetection
 
 import (
+	"encoding/json"
 	"regexp"
 
 	"github.com/DataDog/datadog-agent/pkg/config/model"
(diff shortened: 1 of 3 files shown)


pkg/logs/internal/decoder/auto_multiline_detection/user_samples.go
```diff
@@ -47,7 +48,18 @@ type UserSamples struct {
 func NewUserSamples(config model.Reader) *UserSamples {
 	tokenizer := NewTokenizer(0)
 	s := make([]*UserSample, 0)
-	err := structure.UnmarshalKey(config, "logs_config.auto_multi_line_detection_custom_samples", &s)
+	var err error
+	raw := config.Get("logs_config.auto_multi_line_detection_custom_samples")
+	if raw == nil {
+		return &UserSamples{
```
ddrthall, line 54: Nit: this looks like it's deserving a log.Error to mirror current functionality

ddrthall (approved):

pkg/logs/internal/decoder/auto_multiline_detection/user_samples.go
gh123man: Actually - `Get` will return `nil` if the setting is omitted, which is expected (ex the user simply didn't set it).
So I'm not sure an additional error log is necessary here. 

Please let me know if I'm missing anything though!

ddrthall (approved):

PR #34496, opened 2025-02-26
Remove .pyi files

FlorentClarret:
### What does this PR do?

Remove .pyi files

### Motivation

These files contains type hints for IDE. They are not used at runtime so we don't need to ship them. 

Credits to @alopezz for the idea.

PR #35107, opened 2025-03-13
Don't remove Python type annotation files

ofek:
### What does this PR do?

Partial revert of https://github.com/DataDog/datadog-agent/pull/34496

### Motivation

It is quite possible to use these at runtime, which we are about to do: https://github.com/DataDog/integrations-core/pull/19838


FlorentClarret (approved):

Kyle-Neale (approved):

ofek:
Minor disk increase but this is required to prevent runtime issues.

PR #35111, opened 2025-03-13
Add support to trigger ci-platform-machine-images workflow

jorgetomtz:
<details open><summary>

# Problem
</summary>

We are porting over the definition of our Windows runner AMIs to a different repo. In order to continue doing
dogfooding of release candidates, we need to update the version of the agent we install on our fleet of runners.
This is already done for the old repo so porting over to new one.

</details>

<details open><summary>

# Solution
</summary>

Refactor a bit the current functions that trigger the GH workflow on buildenv repository
so that we can support ci-platform-machine-images.

</details>

<details><summary>

# Testing Done
</summary>

I ran `dda inv -e github.update-windows-runner-version` locally and I was able to trigger both workflows
https://github.com/DataDog/buildenv/actions/runs/14093183746
https://github.com/DataDog/ci-platform-machine-images/actions/runs/14093141532
Things failed since I don't have the same permissions as the bot user which will trigger this on CI but the logic
to start the workflows works.

</details>

### DIFF
--- a/tasks/github_tasks.py
+++ b/tasks/github_tasks.py
@@ -17,8 +17,8 @@ from tasks.libs.ciproviders.github_actions_tools import (
     follow_workflow_run,
     print_failed_jobs_logs,
     print_workflow_conclusion,
-    trigger_buildenv_workflow,
     trigger_macos_workflow,
+    trigger_windows_bump_workflow,
 )
 from tasks.libs.common.color import Color, color_message
 from tasks.libs.common.constants import DEFAULT_INTEGRATIONS_CORE_BRANCH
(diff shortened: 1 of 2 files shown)


tasks/github_tasks.py
```diff
@@ -113,17 +113,30 @@ def trigger_macos(
         raise Exit(message=f"Macos {workflow_type} workflow {conclusion}", code=1)
 
 
-def _update_windows_runner_version(new_version=None, buildenv_ref="master"):
+def _update_windows_runner_version(new_version=None, repo="buildenv"):
```
spencergilbert, line 116: **❓ question:** ‏we'd never want to update the version off a test branch, right?

spencergilbert (approved):

tasks/github_tasks.py
jorgetomtz: yeah we don't. The jobs have rules set up to only trigger on tags https://github.com/DataDog/datadog-agent/blob/main/.gitlab/trigger_release/trigger_release.yml#L93-L130

CelianR (approved):

PR #35115, opened 2025-03-14
ci(installer): Force a retry for the \`new-e2e-installer\` job

chouetz:
### What does this PR do?
Set `retry: 2` for the `new-e2e-installer` job

### Motivation
This job has a [high level of flakiness](https://app.datadoghq.com/ci/pipeline-executions?query=ci_level%3Ajob%20%40ci.pipeline.name%3A%22DataDog%2Fdatadog-agent%22%20%40ci.job.name%3A%22new-e2e-installer%22%20%40git.branch%3Amain%20-%40ci.status%3Askipped&agg_m=count&agg_m_source=base&agg_t=count&fromUser=false&index=cipipeline&start=1741339243941&end=1741944043941&paused=false) which is often caused by external dependencies. Adding an automatic retry for any reason is not really satisfying but will decrease the number of interruptions for the on-call rotation 

n/a

### DIFF
--- a/.gitlab/e2e/e2e.yml
+++ b/.gitlab/e2e/e2e.yml
@@ -530,6 +530,7 @@ new-e2e-installer:
     TEAM: fleet
     FLEET_INSTALL_METHOD: "install_script"
     E2E_LOGS_PROCESSING_TEST_DEPTH: 2
+  retry: 1  # Force retry on this specific job to limit flakiness
 
 new-e2e-installer-windows:
   extends: .new_e2e_template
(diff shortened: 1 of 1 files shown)


KevinFairise2:
I think this is an expensive retry because these tests create a lot of VM. In the other hand the "install-script" test we have are all retried once and it proved to reduce their flakiness. Since these tests are installation tests that are more exposed to flakes than others I think this is fine to retry them

L3n41c (approved):

PR #35116, opened 2025-03-14
[AGENTRUN-40] add dependency audit for all binaries and flavors

GustavoCaso:
### What does this PR do?

This PR introduces an audit check at CI for dependencies for all the binaries, expect serverless.

Our codebase is complex and many packages are used by many binaries that have being the cause of numerous issues with regards of binary size and RSS utilization. 

Currently, we have a few checks SMP and  the package compression. Also, we have the import diff GH comment. None of those checks forces the creator of the PR nor the owner of the binary to reflect on a change that adds a new package to the binary, no it blocks CI. 

The `inv -e go-deps.test-dependency-list` blocks CI if a new package is added or removed, it adds enough friction that requires a codewoner reviewing the change. With the hopes that of the addition is not desire it would start a conversation to figure out the best path forward. In the case of a deletion running the `inv -e go-deps.generate-dependency-list` task would update all the relevant files. 

Another useful usage is that we have a single place to look for what is included in the final binary (`dependencies_*.text`) files. That way if we are investigating a particular binary to reduce its overhead, we can start by looking at the dependency file and look for outliers.

This check is far from perfect, and there is room for improvement, but I think is a nice step in the right direction. I'm open to suggestion and feedback.


tasks/go_deps.py
```diff
@@ -292,3 +294,126 @@ def graph(
...
+                if not os.path.isfile(filename):
+                    print(
+                        f"File {filename} does not exist. To execute the dependencies list check for the {binary} binary, please run the task `inv -e go-deps.generate --binaries {binary}"
+                    )
+                    continue
+
+                deps_file = open(filename)
+                deps = deps_file.read()
+                deps_list = deps.splitlines()
+                deps_file.close()
```
CelianR, line 352:
```suggestion
                with open(filename) as deps_file:
                    deps_list = list(deps_file)
```
This should be the same and be proper 👍

tasks/go_deps.py
```diff
@@ -292,3 +294,126 @@ def graph(
...
+
+    if len(mismatch_binaries) > 0:
+        message = io.StringIO()
+
+        for mismatch_binary in mismatch_binaries:
+            if mismatch_binary.differences > 0:
+                message.write(
+                    color_message(
+                        f"You added some dependencies to {mismatch_binary.binary} ({mismatch_binary.os}/{mismatch_binary.arch}). Adding new dependencies to the binary increases its size. Do we really need to add this dependency?\n",
+                        "red",
```
CelianR, line 372:
```suggestion
                        "orange",
```

.gitlab/source_test/golang_deps_diff.yml
```diff
@@ -63,3 +63,16 @@ golang_deps_send_count_metrics:
     # Get API key to send metrics
     - DD_API_KEY=$($CI_PROJECT_DIR/tools/ci/fetch_secret.sh $AGENT_API_KEY_ORG2 token) || exit $?; export DD_API_KEY
     - dda inv -e go-deps.send-count-metrics --git-sha "${CI_COMMIT_SHA}" --git-ref "${CI_COMMIT_REF_NAME}"
+    
+golang_deps_test:
+  stage: source_test
+  image: registry.ddbuild.io/ci/datadog-agent-buildimages/deb_x64$DATADOG_AGENT_BUILDIMAGES_SUFFIX:$DATADOG_AGENT_BUILDIMAGES
+  tags: ["arch:amd64"]
+  rules:
+    - when: on_success
```
CelianR, line 72:
```suggestion
    - !reference  [.except_mergequeue]
    - when: on_success
```

In addition, should this be triggered only if go files change?

tasks/go_deps.py
```diff
@@ -292,3 +294,126 @@ def graph(
...
+
+                goos, goarch = GOOS_MAPPING[platform], GOARCH_MAPPING[arch]
+
+                filename = os.path.join(ctx.cwd, f"dependencies_{flavor.name}_{goos}_{goarch}.txt")
+
+                deps = compute_binary_dependencies_list(ctx, build, flavor, platform, arch)
+
+                f = open(filename, "w")
+                f.write("\n".join(deps))
+                f.close()
```
CelianR, line 419:
```suggestion
                with open(filename, "w") as f:
                    f.write("\n".join(deps))
```

CelianR (approved):

hush-hush (commented):
Could we bundle this new comment within `inv tidy` to avoid having to run yet another command when working on a PR ?

tobz (approved):

ichinaski (approved):

cmd/system-probe/dependencies_base_linux_amd64.txt
brycekahle, line 1: What are the `vendor/*` dependencies?

paulcacheux: I believe this is https://github.com/golang/go/tree/master/src/vendor

brycekahle: Thanks

tasks/go_deps.py
```diff
@@ -292,3 +294,126 @@ def graph(
...
+    'cluster-agent-cloudfoundry',
+    'cluster-agent',
+    'dogstatsd',
+    'heroku-process-agent',
+    'heroku-trace-agent',
+    'sbomgen',
+    'installer',
+    'system-probe',
+]
+MisMacthBinary = namedtuple('MisMacthBinary', ['binary', 'os', 'arch', 'differences'])
```
brycekahle, line 316:
```suggestion
MisMatchBinary = namedtuple('MisMatchBinary', ['binary', 'os', 'arch', 'differences'])
```

brycekahle:
Do we need this for internal packages too? I worry this will encourage copy-paste rather than proper code reuse, since you can copy the code you need to an existing imported package and not trip up this audit.

PR #35118, opened 2025-03-14
[CWS] delay kills on first rule disarmer period

spikat:
### What does this PR do?

Tld;r: It modifies the kill action behavior, by adding a new guard: first delaying kills to ensure the rule is not too aggressive.

Here is the full kill logic:
image

In detail:
* once a rule with a kill action triggers for the first time, we will start enqueueing kills for this rules until now + disarmer period, and:
- * if the disarmer guards don't triggers during this period (1min by default), the enqueued kills are performed.
- * if during this period the guards triggered, the rule kill action enters a new state `dismantled`. Once dismantled a kill rule action is permanently disarmed (loading a new ruleset with the same rule ID will reset the state)
* it's not possible anymore to change per rule disarmers parameters, now we can only disable them
* new added metrics:
- * `enforcement.kill_queued`, incremented when we enqueue a kill (per rule_id)
- * `enforcement.kill_queued_discarded`, incremented when a enqueued kill is discarded due to a rule disarm (per rule_id)
- * `enforcement.rule_dismantled`, incremented when a rule is dismantled (per rule_id and disarmer type)
- * Also, the already present `enforcement.process_killed` can differentiate if the kill was queued or not.
* The kill disarmer periods are now caped to 1 min max
* An event waiting for its action reports to be resolved can now be delayed up to 1 min 2 sec (it don't change anything for hash action which is still flushed after max 5sec)
* New kill action report status:
- * `rule_dismantled` when the rule has been dismantled
- * `kill_queued` when the kill has been  queued
* Now if a rule with a kill action with container scope triggers on a process not in a container, nothing will be killed
* To kill a process (directly of after an enqueue), there now those guards:
- * pid should exists in proc fs
- * process created at should match (withing 2sec)
- * process path should match
* Lots of unit tests has been added to validate the new state machine

### DIFF
--- a/pkg/security/probe/probe_ebpf.go
+++ b/pkg/security/probe/probe_ebpf.go
@@ -151,9 +151,7 @@ type EBPFProbe struct {
 	discarderRateLimiter         *rate.Limiter
 
 	// kill action
-	killListMap           *lib.Map
-	supportsBPFSendSignal bool
-	processKiller         *ProcessKiller
+	processKiller *ProcessKiller
 
 	isRuntimeDiscarded bool
 	constantOffsets    map[string]uint64
@@ -473,11 +471,6 @@ func (p *EBPFProbe) Init() error {
 	}
 
 	p.eventStream.SetMonitor(p.monitors.eventStreamMonitor)
-
-	p.killListMap, err = managerhelper.Map(p.Manager, "kill_list")
-	if err != nil {
-		return err
-	}
 
 	p.processKiller.Start(p.ctx, &p.wg)
 
@@ -2191,10 +2184,6 @@ func (p *EBPFProbe) initManagerOptionsConstants() {
 			Value: utils.BoolTouint64(p.config.Probe.NetworkFlowMonitorEnabled),
 		},
 		manager.ConstantEditor{
-			Name:  "send_signal",
-			Value: utils.BoolTouint64(p.kernelVersion.SupportBPFSendSignal()),
-		},
-		manager.ConstantEditor{
 			Name:  "anomaly_syscalls",
 			Value: utils.BoolTouint64(slices.Contains(p.config.RuntimeSecurity.AnomalyDetectionEventTypes, model.SyscallsEventType)),
 		},
@@ -2418,8 +2407,6 @@ func NewEBPFProbe(probe *Probe, config *config.Config, opts Opts) (*EBPFProbe, e
 	p.useMmapableMaps = p.kernelVersion.HaveMmapableMaps()
 
 	p.Manager = ebpf.NewRuntimeSecurityManager(p.useRingBuffers)
-
-	p.supportsBPFSendSignal = p.kernelVersion.SupportBPFSendSignal()
 
 	p.monitors = NewEBPFMonitors(p)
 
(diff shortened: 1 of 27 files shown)


pkg/security/probe/process_killer.go
```diff
@@ -459,13 +613,27 @@ const (
 	executableDisarmerType disarmerType = "executable"
 )
 
+type processContext struct {
```
safchain, line 616: I would rather call it killContext

pkg/security/probe/process_killer.go
```diff
@@ -221,67 +283,84 @@ func (p *ProcessKiller) KillAndReport(kill *rules.KillDefinition, rule *rules.Ru
...
-		KilledAt:   killedAt,
 		Pid:        ev.ProcessContext.Pid,
 		rule:       rule,
 	}
+
+	// TODO: merge periods to keep only one?
+	if disarmer != nil && p.FirstPeriodEnqueued(disarmer, sig, pcs, max(disarmer.container.period, disarmer.executable.period)) {
+		log.Warnf("rule %s triggered on first period, putting pids to kill on wait list", rule.ID)
+		report.Status = KillActionStatusQueued
+		report.resolved = true
```
safchain, line 321: Not sure to understand why we mark it as resolved, is it not going to send directly the event without waiting for the action report ?

pkg/security/probe/process_killer.go
```diff
@@ -162,8 +203,28 @@ func (p *ProcessKiller) isRuleAllowed(rule *rules.Rule) bool {
 	return slices.Contains(p.sourceAllowed, rule.Policy.Source)
 }
 
+// called once a rule got disarmed
+func (p *ProcessKiller) updateKillQueueAlarm(disarmer *ruleDisarmer) {
+	// check if we have another rule with queued kills and reset the alarm to
+	// the correct value. Disable the alarm otherwise.
+	// NB: we should not kill anything, but if we do it's ok
+	if alarm := p.KillQueuedPidsAndGetNextAlarm(); alarm != nil {
```
safchain, line 211: should we split this function into 2 functions : one for killing another one for getting the nextAlarm. It is a bit confusing to see a function that aims to kill here (even there a check within the function itself).

pkg/security/probe/process_killer.go
```diff
@@ -387,6 +486,34 @@ func (p *ProcessKiller) Start(ctx context.Context, wg *sync.WaitGroup) {
...
+				case <-p.disableKillQueue:
+					if killQueueAlarm != nil {
+						killQueueAlarm.Stop()
+						killQueueAlarm = nil
+					}
+					killQueueChan = nil
+					p.currentKillQueueAlarm = nil
+					break
+
+				case <-killQueueChan:
```
safchain, line 506: would it possible to use only one ticker let say 250ms call the killqueue function ? it will avoid having 3 chan and most of the time the queue list will be empty.

pkg/security/probe/process_killer.go
spikat: yeah I already though about renaming it :) will do, thanks

pkg/security/probe/process_killer.go
spikat: The idea it's to iterate only once and prevent races when setting a new alarm. I'd rather keep it as it

pkg/security/probe/process_killer.go
spikat: It will yes. I've still to find a better way to handle reports for queued kills on rules with long periods. We can discuss offline

pkg/security/probe/process_killer.go
spikat: Yeah .. I looked to have a start/stop alarm but at the end the implementation is a bit too complex for nothing I guess, I'll update it with a single chan

pkg/security/probe/process_killer.go
spikat: done

pkg/security/probe/process_killer.go
spikat: updated

pkg/security/probe/probe_ebpf.go
```diff
@@ -2794,26 +2781,7 @@ func (p *EBPFProbe) HandleActions(ctx *eval.Context, rule *rules.Rule) {
...
-				if stat.Ino != inode {
-					return fmt.Errorf("failed to kill process %d, incorrect inode %d vs %d", pid, stat.Ino, inode)
-				}
-
-				if p.supportsBPFSendSignal {
-					if err := p.killListMap.Put(uint32(pid), uint32(sig)); err != nil {
-						seclog.Warnf("failed to kill process with eBPF %d: %s", pid, err)
-					}
-				}
-				return p.processKiller.KillFromUserspace(pid, sig, ev)
```
YoannGh, line 2815: This is outside of the scope of this PR but I'm wondering how can the `KillFromUserspace` method work when the process to kill and the system-probe process don't belong to the pid namespace?

spikat: I  don't think we could kill anything in this case yep :/

pkg/security/probe/process_killer.go
```diff
@@ -502,29 +748,42 @@ func (c *disarmerCache[K, V]) flush() int {
...
 }
 
-func (rd *ruleDisarmer) allow(cache *disarmerCache[string, bool], key string, onDisarm func()) bool {
+func (rd *ruleDisarmer) allow(cache *disarmerCache[string, bool], key string, onDisarm func(dismantled bool)) bool {
 	rd.Lock()
 	defer rd.Unlock()
 
+	if rd.disarmed {
+		return false
+	}
```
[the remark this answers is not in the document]
spikat: good point! I just pushed a new version with a quick exit but only on dismantled rules, thanks!

lebauce (approved):

YoannGh (approved):

PR #35119, opened 2025-03-14
(fleet) repair package when it is reinstalled while a previous installation cannot safely be deleted

bmermet:
### What does this PR do?

This PR makes it possible to reinstall packages even if the files from a previous installation of the same version of the package is stil in use. In that case it checks that the directory contains all the expected files and copies any missing file.

The change is scoped to windows for now, to minimise the likelihood of introducing a regression and because that need hasn't surfaced on linux yet.

### Motivation

Make package installation more resilient on windows.

The new code is unit tested and the behavior of the datadog-installer is checked by numerous E2E tests.

### DIFF
--- a/pkg/fleet/installer/repository/repository.go
+++ b/pkg/fleet/installer/repository/repository.go
@@ -7,12 +7,16 @@
 package repository
 
 import (
+	"bytes"
 	"context"
+	"crypto/sha256"
 	"errors"
 	"fmt"
+	"io"
 	"io/fs"
 	"os"
 	"path/filepath"
+	"runtime"
 
 	"github.com/DataDog/datadog-agent/pkg/fleet/installer/paths"
 	"github.com/DataDog/datadog-agent/pkg/util/log"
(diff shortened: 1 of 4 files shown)


raphaelgavache (approved):

pkg/fleet/installer/repository/repository.go
```diff
@@ -481,3 +505,139 @@ func (l *link) Delete() error {
 	l.packagePath = nil
 	return nil
 }
+
+func buildFileMap(rootPath string) (map[string]struct{}, error) {
+	files := make(map[string]struct{})
```
andrewlock, line 510: Am I right in assuming this map would be case sensitive? On Windows it probably _shouldn't_ be case sensitive, right?

pkg/fleet/installer/repository/repository_test.go
```diff
@@ -286,3 +286,120 @@ func TestDeleteHookDenyDoesNotPreventReinstall(t *testing.T) {
...
+		"subdir/file3.txt":  "content3",
+		"subdir/file4.txt":  "content4",
+		"subdir2/file5.txt": "content5",
+		"subdir2/file6.txt": "content6",
+	}
+	targetFiles := map[string]string{
+		"file1.txt":         "wrong_content",
+		"file2.txt":         "content2",
+		"subdir/file3.txt":  "wrong_content",
+		"subdir/file4.txt":  "content4",
```
andrewlock, line 369: nit: we should probably have some tests cases for case sensitivity

andrewlock (approved):
Nice 👍 Just one question about case sensitivity - Windows file paths are not case sensitive, so I think we need to account for that in the comparison code

pkg/fleet/installer/repository/repository.go
andrewlock: As discussed, for simplicity we will ignore case insensitivity as an issue, as we don't expect the case to change between the OCI and the file system, and if it does, the repair will fail anyway, which is acceptable in this unexpected scenario. We can update this to explicitly do case insensitive comparisons if need be, but for now, it's fine

PR #35128, opened 2025-03-14
tagger: Add gpu_driver_version tag to GPU entities

gjulianm:
### What does this PR do?

Adds the `gpu_driver_version` tag to GPU metrics.

### Motivation

Enable comparison of behavior between different GPU drivers.

E2E test passing.

### DIFF
--- a/comp/core/tagger/tags/tags.go
+++ b/comp/core/tagger/tags/tags.go
@@ -105,6 +105,9 @@ const (
 	// KubeGPUUUID is the tag for the Kubernetes Resource GPU UUID
 	KubeGPUUUID = "gpu_uuid"
 
+	// GPUDriver is the tag for the GPU driver
+	GPUDriver = "gpu_driver"
+
 	// OpenshiftDeploymentConfig is the tag for the OpenShift deployment config name
 	OpenshiftDeploymentConfig = "oshift_deployment_config"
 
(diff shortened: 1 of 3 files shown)


val06 (approved):

comp/core/tagger/tags/tags.go
```diff
@@ -105,6 +105,9 @@ const (
 	// KubeGPUUUID is the tag for the Kubernetes Resource GPU UUID
 	KubeGPUUUID = "gpu_uuid"
 
+	// GPUDriver is the tag for the GPU driver
+	GPUDriver = "gpu_driver"
```
wdhif, line 109: Should this be more like `gpu_driver_version` to be closer to the real value meaning?

gjulianm: It makes more sense indeed, changed

wdhif: Sorry for nitpicking again, but the variable should probably be renamed to `GPUDriverVersion`.

comp/core/tagger/tags/tags.go
```diff
@@ -105,6 +105,9 @@ const (
 	// KubeGPUUUID is the tag for the Kubernetes Resource GPU UUID
 	KubeGPUUUID = "gpu_uuid"
 
+	// GPUDriverVersion is the tag for the GPU driver
```
wdhif, line 108:
```suggestion
	// GPUDriverVersion is the tag for the GPU driver version
```

comp/core/tagger/tags/tags.go
wdhif: Perfect, just added this: https://github.com/DataDog/datadog-agent/pull/35128#discussion_r2003467433
Approved 👍

wdhif (approved):

PR #35132, opened 2025-03-14
upgrade(installer): Generate experiment units from stable ones

BaptisteFoy:
### What does this PR do?
Adds an invoke task to generate experiment systemd units from the stable ones.
Adds the corresponding CI check job.

### Motivation
Avoid drift between experiment & stable units

E2E, manual testing, unit tests


arbll (approved):

arbll (approved):

pkg/fleet/installer/packages/embedded/datadog-agent-trace-exp.service
chouetz, line 1: Should you align the file name according to the update below?

tasks/installer.py
```diff
@@ -91,3 +92,70 @@ def build_linux_script(ctx, flavor, version, bin_amd64, bin_arm64, output):
...
+
+@task
+def generate_experiment_units(ctx, check=False):
+    '''
+    Generates systemd units for the experiment service.
+    '''
+
+    files = glob.glob('./pkg/fleet/installer/packages/embedded/*.service')
+    for stable_path in files:
+        if "datadog-installer" in stable_path or "-exp.service" in stable_path:
```
chouetz, line 105: The second condition seems strange to me, as it means you explicitly ignore the `-exp.service`, while this task is supposed to generate the experiment services. And below you rename the "simple" `service` as `exp-service`, so your intent is not really clear.
Could you clarify please?
Thanks in advance

.gitlab/lint/technical_linters.yml
```diff
@@ -60,3 +60,8 @@ validate_modules:
   script:
     - dda inv -e modules.validate
     - dda inv -e modules.validate-used-by-otel
+
+validate_experiment_systemd_units:
+  extends: .lint
+  script:
```
CelianR, line 66: You should add a rule such that it is triggered only on specific file changes (`./pkg/fleet/installer/packages/embedded/*.service` I guess) or on main

tasks/installer.py
```diff
@@ -91,3 +92,70 @@ def build_linux_script(ctx, flavor, version, bin_amd64, bin_arm64, output):
...
+
+
+@task
+def generate_experiment_units(ctx, check=False):
+    '''
+    Generates systemd units for the experiment service.
+    '''
+
+    files = glob.glob('./pkg/fleet/installer/packages/embedded/*.service')
+    for stable_path in files:
```
chouetz, line 104: 🥜 the paths you are iterating into are not `stable` yet, IIUC. Regarding my comment below, you have only the stable ones after your initial filter, correct?

CelianR (approved):
Minor change to do

tasks/installer.py
```diff
@@ -91,3 +92,70 @@ def build_linux_script(ctx, flavor, version, bin_amd64, bin_arm64, output):
...
+        if "datadog-installer" in stable_path or "-exp.service" in stable_path:
+            continue
+        experiment_path = stable_path.replace(".service", "-exp.service")
+
+        experiment_file = ""
+        with open(stable_path) as f:
+            for line in f:
```
chouetz, line 111: Your file format looks like an `ini` or `toml`. Have you considered using a parser like [toml](https://pypi.org/project/toml/)? The only drawback of such parser would be keeping comments in the correct order, if any. If this is not a strong requirement on your side it could make this loop quite simpler, wdyt?

chouetz: In any case I think you could extract this part in a dedicated method with some unit test to ensure it's working as expected, to keep it robust.

tasks/installer.py
```diff
@@ -91,3 +92,70 @@ def build_linux_script(ctx, flavor, version, bin_amd64, bin_arm64, output):
...
+                if "Description=" in line:
+                    line = line.replace("\n", "") + " Experiment\n"
+                line = line.replace("stable", "experiment")
+                experiment_file += line
+
+        if not check:
+            with open(experiment_path, 'w') as f:
+                f.write(experiment_file)
+        else:
+            with open(experiment_path) as f:
```
chouetz, line 157: Are we sure this file exists? Otherwise you might crash here

chouetz (approved):
Such a check could be added as a [pre-commit hook](https://github.com/DataDog/datadog-agent/blob/main/.pre-commit-config.yaml) I suppose.
As I only did suggestions, I consider this is good to go, thanks

tasks/installer.py
BaptisteFoy: I'll add a comment in the code to reflect my message.

The generated files are the `-exp.service` files, while the generator files are the `.service` files (no `-exp`). We don't want to re-execute the generation on generated files, else we'll have `-exp-exp.service` files and so on.

tasks/installer.py
BaptisteFoy: Good call. We want to raise an exception in any case but it's better if we control it.

tasks/installer.py
BaptisteFoy: That's correct yes

PR #35136, opened 2025-03-14
[CNM] clean up network ID / vpc ID logs

akarpz:
### What does this PR do?

if the system-probe HTTP client is nil (this should never happen), we will never do any retries as it will immediately return an empty string and no error. This PR also cleans up the logging a bit to make it clearer what's happening.

### DIFF
--- a/pkg/process/checks/net.go
+++ b/pkg/process/checks/net.go
@@ -504,7 +504,7 @@ func retryGetNetworkID(sysProbeClient *http.Client) (string, error) {
 			return networkID, nil
 		}
 		log.Debugf(
-			"failed to get network ID from system-probe (attempt %d/%d): %s",
+			"failed to get network ID from host or system-probe (attempt %d/%d): %s",
 			attempt,
 			maxRetries,
 			err,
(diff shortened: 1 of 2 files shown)


pkg/process/checks/net.go
```diff
@@ -516,13 +516,16 @@ func retryGetNetworkID(sysProbeClient *http.Client) (string, error) {
 
 // getNetworkID fetches network_id from the current netNS or from the system probe if necessary, where the root netNS is used
 func getNetworkID(sysProbeClient *http.Client) (string, error) {
-	networkID, err := network.GetNetworkID(context.TODO())
-	if err != nil && sysProbeClient != nil {
+	networkID, err := network.GetNetworkID(context.Background())
```
hmahmood, line 519: Does `network.GetNetworkID` handle the case `sysProbeClient == nil` case?

akarpz: the previous issue is that we were not retrying the initial fetch of networkID (not the sysprobe RPC version) if the sysProbeClient was nil. we would just bail out of the whole thing and silently leave networkID as an empty string. this will now properly retry, and the sysprobe RPC will never happen if the client is nil

hmahmood (approved):

pkg/process/checks/net.go
akarpz: but I don't think it's practically possible for the client to be nil anyways, so I think this PR is functionally just cleaning some logs

PR #35138, opened 2025-03-14
Propagate agent host tags for oracle

natashadada:
### What does this PR do?
This PR updates the agent host tags that we use in the oracle integration. This change only affects DBM customers with the `propagate_agent_tags` flag set to true. By getting host tags, we not only use the tags in the conf.yaml file but also other tags on the agent host. We already have this for our other integrations, so adding it for oracle too.

### Motivation
Metrics emitted by DBM integrations are tagged with the tags from the database instance host. We would like to update the `propagate_agent_tags` option (default false) that allows us to also consider the tags of the agent host. If this setting is enabled, we will propagate all agent host tags. This allows user to opt-in to having their postgres metrics tagged with the agent host tags.

### DIFF
--- a/releasenotes/notes/add-propagate-agent-host-tags-flag-bc1fcd6ba35cbf12.yaml
+++ b/releasenotes/notes/add-propagate-agent-host-tags-flag-bc1fcd6ba35cbf12.yaml
@@ -8,4 +8,4 @@
 ---
 enhancements:
   - |
-    Add the propagate_agent_tags setting. When set to `true`, the tags from the agent host are added to the check's tag for all instances.
+    Add the `propagate_agent_tags setting`. When set to `true`, the tags from the Agent host are added to the check's tag for all instances.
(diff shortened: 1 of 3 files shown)


sethsamuel (approved):

releasenotes/notes/add-propagate-agent-host-tags-flag-bc1fcd6ba35cbf12.yaml
```diff
@@ -0,0 +1,11 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+enhancements:
+  - |
+    Add the propagate_agent_tags setting. When set to `true`, the tags from the agent host are added to the check's tag for all instances.
```
brett0000FF, line 11:
```suggestion
    Add the `propagate_agent_tags setting`. When set to `true`, the tags from the Agent host are added to the check's tag for all instances.
```

brett0000FF (approved):

brett0000FF (approved):
🚀

PR #35142, opened 2025-03-15
usm: tests: Add test buildflag for test utils and test helper files

guyarb:
### What does this PR do?

Adds `test` buildflag for test utilities of USM

### Motivation

Prevent test only code from being accessed in production code


amitslavin (approved):
WDYT about adding test buid for the following as well:
gRPC pkg:
  * pkg/network/usm/testutil/grpc/builder.go
  * pkg/network/usm/testutil/grpc/client.go
  * pkg/network/usm/testutil/grpc/server.go

alexn-dd (approved):

PR #35145, opened 2025-03-15
fix(network): Prefer resolved host IP if available

BaptisteFoy:
### What does this PR do?
Some customers explicitly tie their host IP to their hostname in `/etc/hosts`. They expect this IP to be reported to the backend. Today this is not the case, we take the IP of the first interface that becomes up. This is not configurable in most OSes.

Having specifically set the host IP in `/etc/hosts` means that `hostname -i` will return the hardcoded IP. `hostname -i` just resolves the hostname. This PR implements it in the agent by trying to resolve the hostname. If it succeeds we use this IP, else we fallback to the existing mechanism

Also, for Fleet Automation IP display, this PR sends all network interfaces as part of the `inventoryhost` payload.

### Motivation
Major customer rely on the host IP in some monitors to trigger & for remediation, we need to display the one they expect.

Using a VM:
1. Add an entry in /etc/hosts (IP / hostname pair)
2. Set the hostname in datadog.yaml
3. Restart the agent
4. Check in the infra list that I see my hostname with the set IP

May break existing IP address resolution, but we haven't seen many customers use that feature. In any case if they set resolution in `/etc/hosts` the new IP will be aligned with what they put in their host configuration.

### Additional Notes
Gated behind a `inventories_network_use_hostname_resolver` boolean parameter.


pkg/config/setup/config.go
```diff
@@ -925,6 +925,7 @@ func InitConfig(config pkgconfigmodel.Setup) {
 	config.BindEnvAndSetDefault("inventories_min_interval", 0) // 0 == default interval from inventories
 	// Seconds to wait to sent metadata payload to the backend after startup
 	config.BindEnvAndSetDefault("inventories_first_run_delay", 60)
+	config.BindEnvAndSetDefault("inventories_network_use_hostname_resolver", false) // resolve the hostname to get the IP address
```
hush-hush, line 928: We're trying to deprecate the `inventory` prefix:

```suggestion
	config.BindEnvAndSetDefault("metadata_ip_resolution_from_hostname", false) // resolve the hostname to get the IP address
```

pkg/gohai/gohai.go
```diff
@@ -114,10 +115,24 @@ func getGohaiInfo(isContainerized, withProcesses bool) *gohai {
...
+				ipv4s, ipv6s, err := network.ResolveFromHostname(hostname)
+				if err != nil {
+					log.Errorf("failed to resolve hostname to IP addresses: %s", err) //nolint:errcheck
+				} else {
+					if len(ipv4s) > 0 {
+						networkInfo.IPAddress = ipv4s[0]
+					}
+					if len(ipv6s) > 0 {
+						networkInfo.IPAddressV6 = utils.NewValue(ipv6s[0])
+					}
```
hush-hush, line 133: Why do we always use the first elem ?

releasenotes/notes/inventories_network_use_hostname_resolver-1c19f97cd2208277.yaml
```diff
@@ -0,0 +1,12 @@
...
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+enhancements:
+  - |
+    Adds a config option to use hostname DNS resolution as the
+    preferred method to compute the host IP address
```
hush-hush, line 12: We should mention the setting name here and we should update the config template if this is a publicly documented setting.

comp/metadata/inventoryhost/inventoryhostimpl/inventoryhost.go
```diff
@@ -83,6 +83,7 @@ type hostMetadata struct {
 	IPAddress   string `json:"ip_address"`
 	IPv6Address string `json:"ipv6_address"`
 	MacAddress  string `json:"mac_address"`
+	Interfaces  string `json:"interfaces"`
```
hush-hush, line 86: If this is a new element, will it be taken into account in the backend ?

pkg/gohai/gohai_test.go
```diff
@@ -46,7 +46,7 @@ func TestGetPayloadContainerizedWithDocker0(t *testing.T) {
 	docker0Detected = true
 	defer func() { docker0Detected = oldDocker0Detected }()
 
-	gohai := GetPayload(false)
+	gohai := GetPayload("hostname", false, false)
```
hush-hush, line 49: Can we add a unit test when the second elem it true ?

hush-hush (changes requested):

releasenotes/notes/inventories_network_use_hostname_resolver-1c19f97cd2208277.yaml
BaptisteFoy: I'm not 100% sure we want to make this publicly available yet, we definitely don't have enough customer data regarding the need. 

But agreed on adding the setting name in the release note, I'll update

comp/metadata/inventoryhost/inventoryhostimpl/inventoryhost.go
BaptisteFoy: The plan is to use it in FA, so yes

pkg/gohai/gohai_test.go
BaptisteFoy: Done in [address review](https://github.com/DataDog/datadog-agent/pull/35145/commits/9445dd6ad2f9317c46b0b5f134d517ca9ccb19a4)

releasenotes/notes/inventories_network_use_hostname_resolver-1c19f97cd2208277.yaml
BaptisteFoy: Done in [address review](https://github.com/DataDog/datadog-agent/pull/35145/commits/9445dd6ad2f9317c46b0b5f134d517ca9ccb19a4)

pkg/config/setup/config.go
BaptisteFoy: Done in [address review](https://github.com/DataDog/datadog-agent/pull/35145/commits/9445dd6ad2f9317c46b0b5f134d517ca9ccb19a4)

releasenotes/notes/inventories_network_use_hostname_resolver-1c19f97cd2208277.yaml
hush-hush: We can always add the feature without a release note :shrug:

hush-hush (approved):

urseberry (approved):

PR #34886, opened 2025-03-07
ci(gitlab rules): Remove useless mentions of a6/a7

chouetz:
### What does this PR do?
- Remove not used mention of agent6/7
- Rename entries on `release.json` as we only have a single version to manage
- Rename also the env variable `RELEASE_VERSION`
- Fix the `linter.gitlab-ci` test: some variables are compared to `null` to test if they are defined, and we set them to "". It was masking the configuration issue raised on the first commit of this PR.

### Motivation
Simplification, improvements

### Describe how you validated your changes
I rely on unit tests and CI pipeline 🤞🏽

PR #35159, opened 2025-03-17
[Agent6] Remove occurrence of agent6/agent7

chouetz:
### What does this PR do?
Same as #34886

### Motivation
Fix the [failing workflow](https://github.com/DataDog/datadog-agent/actions/runs/13895463868/job/38875058080)

I need to merge to re-run the workflow


CelianR (approved):
Question: Should we continue updating release.py since we use worktrees now? Maybe we could remove it to avoid changes in the future

chouetz:
> Question: Should we continue updating release.py since we use worktrees now? Maybe we could remove it to avoid changes in the future

You are perfectly right, this is not necessary. I did it for the invoke tests only, we could remove everything

PR #35169, opened 2025-03-17
Add SBOM on fargate using procfs

safchain:
### What does this PR do?

Use procfs to scan container file system. For each container, except the core agent one (self pid), we use the /proc/pid/root path to trigger a scan.

This feature is limited to Fargate instance as this is the main target for now.

### DIFF
--- a/pkg/collector/corechecks/sbom/processor.go
+++ b/pkg/collector/corechecks/sbom/processor.go
@@ -284,9 +284,10 @@ func (p *processor) triggerProcfsScan(ctr *workloadmeta.Container) {
 func (p *processor) processProcfsScanResult(result sbom.ScanResult) {
 	log.Debugf("processing procfs scanresult: %v", result)
 	sbom := &model.SBOMEntity{
-		Status:             model.SBOMStatus_SUCCESS,
-		Id:                 result.RequestID,
-		Type:               model.SBOMSourceType_CONTAINER_FILE_SYSTEM,
+		Status: model.SBOMStatus_SUCCESS,
+		Id:     result.RequestID,
+		//Type:               model.SBOMSourceType_CONTAINER_FILE_SYSTEM,
+		Type:               model.SBOMSourceType_HOST_IMAGE,
 		InUse:              true,
 		GeneratedAt:        timestamppb.New(result.CreatedAt),
 		GenerationDuration: convertDuration(result.Duration),
--- a/pkg/collector/corechecks/sbom/processor.go
+++ b/pkg/collector/corechecks/sbom/processor.go
@@ -280,10 +280,9 @@ func (p *processor) triggerProcfsScan(ctr *workloadmeta.Container) {
 func (p *processor) processProcfsScanResult(result sbom.ScanResult) {
 	log.Debugf("processing procfs scanresult: %v", result)
 	sbom := &model.SBOMEntity{
-		Status: model.SBOMStatus_SUCCESS,
-		Id:     result.RequestID,
-		//Type:               model.SBOMSourceType_CONTAINER_FILE_SYSTEM,
-		Type:               model.SBOMSourceType_HOST_IMAGE,
+		Status:             model.SBOMStatus_SUCCESS,
+		Id:                 result.RequestID,
+		Type:               model.SBOMSourceType_HOST_IMAGE, //Change this to SBOMSourceType_CONTAINER_FILE_SYSTEM once BE is ready
 		InUse:              true,
 		GeneratedAt:        timestamppb.New(result.CreatedAt),
 		GenerationDuration: convertDuration(result.Duration),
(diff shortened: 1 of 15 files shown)


pkg/collector/corechecks/sbom/processor.go
```diff
@@ -259,6 +262,70 @@ func (p *processor) triggerHostScan() {
...
 }
 
+func (p *processor) triggerProcfsScan(ctr *workloadmeta.Container) {
+	// Allowed only on Fargate instance for now
+	if !fargate.IsFargateInstance() {
+		log.Warnf("procfs scan supported only on Fargate instances, ignoring : %s", ctr.ID)
+		return
+	}
+
+	if !pkgconfigsetup.Datadog().GetBool("sbom.container.enabled") {
```
paulcacheux, line 272: the `sbom.host.enabled` config is passed when creating the processor. It would be great to do the same thing here https://github.com/DataDog/datadog-agent/blob/d5a84b2612bdfaa746f5f8b0ed99d27806d5c1fe/pkg/collector/corechecks/sbom/check.go#L163

paulcacheux: and IMO the fargate check could be done in the same spot, when creating the processor. Otherwise we are going to do the same check for each container event

pkg/sbom/collectors/procfs/procfs_linux.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/shirou/gopsutil/v4/process"
+
+	"github.com/DataDog/datadog-agent/pkg/sbom"
+	"github.com/DataDog/datadog-agent/pkg/security/secl/containerutils"
+	"github.com/DataDog/datadog-agent/pkg/security/utils"
```
paulcacheux, line 18: It's really not amazing that we start using some pkg/security stuff here, since it's going to be in the core agent

paulcacheux (approved):

pkg/sbom/collectors/procfs/procfs_linux.go
safchain: yeah I agree, I'll see if I extract only what is needed or move it to a common package

pkg/collector/corechecks/sbom/processor.go
safchain: I can pass the boolean in the processor creation but I'm not sure there is a huge value, we'll still have to check on each container events as this is done for host.

pkg/collector/corechecks/sbom/processor.go
```diff
@@ -135,6 +137,7 @@ func (p *processor) processContainerImagesEvents(evBundle workloadmeta.EventBund
 				continue
 			}
 			p.registerContainer(container)
+			p.triggerProcfsScan(event.Entity.(*workloadmeta.Container))
```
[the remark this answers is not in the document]
safchain: yes it is going to scan all the containers except the DD one which is filtered out later when scanning procfs. I known this is not ideal but I don't want to put the procfs scan here just for one container. I can put a comment.

pkg/collector/corechecks/sbom/processor.go
safchain: I changed the logic.

pkg/sbom/collectors/procfs/procfs_linux.go
safchain: I removed this dependency. The algo is more naive but still work

paulcacheux (approved):

pkg/collector/corechecks/sbom/processor.go
```diff
@@ -259,6 +268,60 @@ func (p *processor) triggerHostScan() {
...
+		log.Errorf("Failed to trigger SBOM generation for procfs: %s", err)
+	}
+}
+
+func (p *processor) processProcfsScanResult(result sbom.ScanResult) {
+	log.Debugf("processing procfs scanresult: %v", result)
+	sbom := &model.SBOMEntity{
+		Status: model.SBOMStatus_SUCCESS,
+		Id:     result.RequestID,
+		//Type:               model.SBOMSourceType_CONTAINER_FILE_SYSTEM,
```
dustmop, line 285: Could this line be removed?

dustmop (approved):
LGTM for agent-configuration

pkg/sbom/collectors/procfs/utils_linux.go
```diff
@@ -0,0 +1,31 @@
...
+
+func cgroupContains(pid uint32, str string) (bool, error) {
+	cgroupPath := kernel.HostProc(strconv.FormatUint(uint64(pid), 10), "cgroup")
+
+	data, err := os.ReadFile(cgroupPath)
+	if err != nil {
+		return false, err
+	}
+
+	return bytes.Contains(data, []byte(str)), nil
```
spikat, line 24: Isn't this match a bit loose?

spikat (approved):

pkg/collector/corechecks/sbom/processor.go
safchain: I removed it and added a comment

lebauce (approved):

PR #35176, opened 2025-03-17
[tasks] Run test tasks for all modules in parallel

KSerrania:
### What does this PR do?

Updates the test and lint tasks to run their commands on all modules at once, instead of per module. In detail:
- the test and lint commands still accept the same inputs (a list of modules and/or packages, or `--get-impacted-packages` / `--get-modified-packages` that compute the list of relevant modules and packages)
- these inputs are transformed into a list of packages to test / lint across modules
- the test / lint commands are run on this list, instead of once per module.

Refactors all core code to not mention modules, and removes the indirection induced by the `test_core` method (that allowed running a given command on multiple modules).

Other updates:
- The `TestWasher` now takes an explicit result json file as input, instead of relying on the default. This was needed to make it work with the non-default result json file I use in E2E tests.
  - As a result, the JUnit upload methods also take an explicit result json file as input, since they use the `TestWasher`.
- `get_impacted_packages`'s early return when specific files are modified now happens before the computation of the dependency graph, making this case ~1 minute faster.
- The E2E test jobs now expose their result json in artifacts.
- The `--save-result-json` option has been renamed to `--result-json`, more appropriate since the argument is a string.
- A few cosmetic updates to log lines printed in the test / lint jobs.

### Motivation

This allows `go test` to better parallelize tests, reducing the duration of full test runs by 10-15 minutes.

Average duration on `main` before PR:
image

Average duration on the PR:
image

Ran CI, verified that all jobs pass.
Verified the test outputs against `main`, verified that all modules tested and linted on `main` are also tested and linted in this PR.

Ran [CI on another dev branch](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/60317243) to verify the behavior of the `--only-impacted-packages` option, and verified that all test jobs pass.

The result json is now always produced, since it is needed for the failure output, test washer, and junit tar production. Before, we were relying on temporary per-module test result files, but they do not exist anymore.

### Additional Notes

n/a


paulcacheux (approved):
LGTM for security-agent related files

tasks/gotest.py
```diff
@@ -116,83 +115,94 @@ def test_flavor(
...
-            if lines is not None and 'DONE 0 tests' in lines[-1]:
-                cov_path = os.path.join(module_path, PROFILE_COV)
-                print(color_message(f"No tests were run, skipping coverage report. Removing {cov_path}.", "orange"))
-                try:
-                    os.remove(cov_path)
-                except FileNotFoundError as e:
-                    print(f"Couldn't remove coverage file {cov_path}\n{e}")
-                return
+    # Produce the result json file, which is used to show the failures at the end of the test run
+    if result_json:
```
chouetz, line 139: As you set a default value this condition is always true (unless we call the task with `--result-json ""` which is unlikely), correct? If yes we can probably get rid of this line.

tasks/gotest.py
```diff
@@ -116,83 +115,94 @@ def test_flavor(
...
     test_profiler: TestProfiler,
     coverage: bool = False,
+    result_json: str = DEFAULT_TEST_OUTPUT_JSON,
 ):
     """
     Runs unit tests for given flavor, build tags, and modules.
     """
+
+    # Early return if no modules are given
+    # Can happen when --only-impacted-packages or --only-modified-packages is used
```
chouetz, line 127: Does it mean the logic to send the full list of modules is not applied in such case?

tasks/gotest.py
```diff
@@ -655,6 +658,11 @@ def get_impacted_packages(ctx, build_tags=None):
         print(f"Triggering all tests because a file matching one of the {TRIGGER_ALL_TESTS_PATHS} was modified")
         return get_default_modules().values()
 
+    if build_tags is None:
```
chouetz, line 661: What is the purpose of moving this block here?

tasks/linter.py
```diff
@@ -273,34 +273,38 @@ def lint_flavor(
...
-                module_result.lint_outputs.append(lint_result)
-                if lint_result.exited != 0:
-                    module_result.failed = True
-        module_results.append(module_result)
-
-    return test_core(
-        modules, flavor, ModuleLintResult, "golangci_lint", command, headless_mode=headless_mode
-    ), execution_times
+    # Compute full list of targets to run linters against
+    targets = []
```
chouetz, line 277: As it is used twice (maybe pending the resolution of the FIXME below), what do you think of having the `targets` generation in a dedicated method?

chouetz (approved):
Only minor comments. The code looks simpler and is more efficient, good game!

tasks/gotest.py
KSerrania: Yes, and it is on purpose: `--only-impacted-packages` and `--only-modified-packages` may compute that no go packages need to be tested, in which case we get an empty module list.

tasks/gotest.py
KSerrania: I prefer keeping that safeguard for the `--result-json ""` case (which people may want to apply to not overwrite their current file). It does not change much complexity-wise, and prevents the command from crashing otherwise.

tasks/gotest.py
KSerrania: That's a git shenanigan - what I did was move the previous block (starting with `for file in TRIGGER_ALL_TESTS_PATHS`) up, so that the early return is evaluated before the very costly `dependencies = create_dependencies(ctx, build_tags)` line.

tasks/linter.py
KSerrania: I'd have liked to, but it is more complex than it seems because the code block is not exactly the same. Here, the inner loop reads `for target in module.lint_targets:`, while in `gotest.py`, it reads `for target in module.test_targets:`.

I could make it work with more refactoring of the `GoModule` class, but I don't feel that it is worth it in this (already big) PR.

tasks/gotest.py
```diff
@@ -205,17 +215,13 @@ def sanitize_env_vars():
             del os.environ[env]
 
 
-def process_test_result(test_results, junit_tar: str, flavor: AgentFlavor, test_washer: bool) -> bool:
+def process_test_result(test_result, result_json: str, junit_tar: str, flavor: AgentFlavor, test_washer: bool) -> bool:
```
KevinFairise2, line 218: Can we use `result_json_path` field of `test_result` instead here?

KevinFairise2 (approved):

tasks/linter.py
chouetz: I am not good at the game of 7 differences 😄

clarkb7 (approved):

PR #35187, opened 2025-03-17
rework authtoken mock component

misteriaud:
### What does this PR do?
This PR improves the mock implementation of the `authtoken.Component` by generating and saving the authentication artifacts in-memory rather than on disk. It also provides a helper to start a mock server using the generated IPC certificate.

The remote workloadmeta has been updated to avoid directly reading the `auth_token` from the disk, instead using the utility function `SetAuthToken`.

### Motivation
This abstracts the complexity of setting the correct configuration for `auth_token_file_path` and prevents leaving files behind after tests.

This only affects code running on tests


mellon85 (approved):

davidor (approved):

ichinaski (approved):

vickenty (approved):

daniel-taf (approved):

hush-hush (approved):

comp/api/authtoken/mock/mock.go
```diff
@@ -0,0 +1,65 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+//go:build test
```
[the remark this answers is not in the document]
pgimalac: The package should only be imported from tests, but it doesn't hurt to have the build tag to avoid bad imports

comp/otelcol/ddflareextension/impl/go.mod
```diff
@@ -3,13 +3,13 @@ module github.com/DataDog/datadog-agent/comp/otelcol/ddflareextension/impl
 go 1.23.0
 
 require (
+	github.com/DataDog/datadog-agent/comp/api/authtoken v0.0.0-00010101000000-000000000000
```
pgimalac, line 6: This is not a new module so you should be able to set a real version directly

pkg/api/util/util_notest_helper.go
```diff
@@ -0,0 +1,18 @@
...
+
+//go:build !test
+
+package util
+
+// SetAuthTokenInMemory is only expected to be used for unit-tests
+// In a non-test environment, this function is a no-op
+func SetAuthTokenInMemory() {
+	// No-op implementation
+}
```
pgimalac, line 14: **🥜 nitpick:** ‏‏If this only called in unit tests then you could just not define it when the test build tag is not used ?

pkg/api/util/util_test_helper.go
```diff
@@ -0,0 +1,68 @@
...
+
+// SetAuthTokenInMemory is only expected to be used for unit-tests
+// It sets the auth token, client TLS config and server TLS config in memory
+// and initializes the initSource to setAuthTokenInMemory
+func SetAuthTokenInMemory() {
+	if initSource != uninitialized {
+		if initSource != setAuthTokenInMemory {
+			panic("the auth stack have been initialized by un underlying part of the code")
+		}
+		fmt.Printf("the auth stack have been initialized before calling SetAuthTokenInMemory")
```
pgimalac, line 27: **❓ question:** ‏Any reason not to use the logger in this file ?

pkg/api/util/util_test_helper.go
```diff
@@ -0,0 +1,68 @@
...
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+)
+
+// SetAuthTokenInMemory is only expected to be used for unit-tests
+// It sets the auth token, client TLS config and server TLS config in memory
+// and initializes the initSource to setAuthTokenInMemory
+func SetAuthTokenInMemory() {
+	if initSource != uninitialized {
```
pgimalac, line 23: **🔨 warning:** ‏You're accessing `initSource` without any synchronization, is that safe ? Could you have multiple concurrent calls to this function ?

pkg/api/util/util_test_helper.go
```diff
@@ -0,0 +1,68 @@
...
+	clientTLSConfig = &tls.Config{
+		InsecureSkipVerify: true,
+	}
+	initSource = setAuthTokenInMemory
+}
+
+// CleanupAuthTokenInMemory is only expected to be used for unit-tests
+// It cleans up the auth token, client TLS config and server TLS config in memory
+// and initializes the initSource to uninitialized
+func CleanupAuthTokenInMemory() {
```
pgimalac, line 58: **💬 suggestion:** ‏You could have `SetAuthTokenInMemory` take a `*testing.T` as parameter, and add the cleanup directly in that function ?

pkg/api/util/util_test_helper.go
misteriaud: No reason in particular indeed, I updated it

pkg/api/util/util_test_helper.go
misteriaud: We can imagine multiple concurrent calls in the case of `t.Parallel()` usage. I update the lock to be acquired at the beginning of the function.

pkg/api/util/util_test_helper.go
misteriaud: The cleanup stage wasn't finally relevant to have since:
- it is store in memory, so it will be cleaned when the test finished anyway
- there is no need for per-test values, so test A can `SetAuthTokenInMemory` and test B can reuse them without problem.

It also prevents test B to use cleaned values in the situation where testA end earlier than testB, which would be problematic.

pgimalac: If it just needs to be called once whenever we run tests then you might as well have an `init` function in a file behind a `//go:build test` tag

songy23 (approved):
Approval for OTel

pkg/api/util/util_test_helper.go
misteriaud: A and B names are interchangeables, so if testB start first, it will initialized the auth globals by calling `SetAuthTokenInMemory`.
Ideally this code will be moved to the mock internal implementation when we will get rid of the `pkg/api/util` globals (which is WIP).

comp/otelcol/ddflareextension/impl/go.mod
misteriaud: If I use a real version will it be able to access the new authtoken mock implementation that I'm introducing in this PR ?

pgimalac: Yes thanks to the `replace` directives we're always using the "current" version (the pinned version is only for code importing this from outside) 👍

pgimalac (approved):

PR #35191, opened 2025-03-17
APM - Enable container based second primary tags

raphaelgavache:
### What does this PR do?
Enable [extraction of container tags for APM stats by default](https://docs.datadoghq.com/tracing/guide/setting_primary_tags_to_scope/?tab=helm#container-based-second-primary-tags), this allows choosing container tags as secondary primary tags.
To disable the extraction set in the agent configuration the following env
```
datadog:
  #...
  env:
    - name: DD_APM_FEATURES
      value: 'disable_cid_stats'
```

### DIFF
--- a/pkg/trace/stats/concentrator.go
+++ b/pkg/trace/stats/concentrator.go
@@ -40,6 +40,7 @@ type Concentrator struct {
 	bsize         int64
 	exit          chan struct{}
 	exitWG        sync.WaitGroup
+	cidStats      bool
 	agentEnv      string
 	agentHostname string
 	agentVersion  string
@@ -54,10 +55,12 @@ func NewConcentrator(conf *config.AgentConfig, writer Writer, now time.Time, sta
 		ComputeStatsBySpanKind: conf.ComputeStatsBySpanKind,
 		BucketInterval:         bsize,
 	}, now)
+	_, disabledCIDStats := conf.Features["disable_cid_stats"]
 	c := Concentrator{
 		spanConcentrator: sc,
 		Writer:           writer,
 		exit:             make(chan struct{}),
+		cidStats:         !disabledCIDStats,
 		agentEnv:         conf.DefaultEnv,
 		agentHostname:    conf.Hostname,
 		agentVersion:     conf.AgentVersion,
@@ -113,20 +116,11 @@ type Input struct {
 }
 
 // NewStatsInput allocates a stats input for an incoming trace payload
-func NewStatsInput(numChunks int, containerID string, clientComputedStats bool, conf *config.AgentConfig) Input {
+func NewStatsInput(numChunks int, containerID string, clientComputedStats bool) Input {
 	if clientComputedStats {
 		return Input{}
 	}
-	in := Input{Traces: make([]traceutil.ProcessedTrace, 0, numChunks)}
-	_, enabledCIDStats := conf.Features["enable_cid_stats"]
-	_, disabledCIDStats := conf.Features["disable_cid_stats"]
-	enableContainers := enabledCIDStats || (conf.FargateOrchestrator != config.OrchestratorUnknown)
-	if enableContainers && !disabledCIDStats {
-		// only allow the ContainerID stats dimension if we're in a Fargate instance or it's
-		// been explicitly enabled and it's not prohibited by the disable_cid_stats feature flag.
-		in.ContainerID = containerID
-	}
-	return in
+	return Input{Traces: make([]traceutil.ProcessedTrace, 0, numChunks), ContainerID: containerID}
 }
 
 // Add applies the given input to the concentrator.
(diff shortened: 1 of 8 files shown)


songy23 (approved):
approval for OTel

pkg/trace/stats/concentrator.go
```diff
@@ -113,20 +116,11 @@ type Input struct {
...
-	_, enabledCIDStats := conf.Features["enable_cid_stats"]
-	_, disabledCIDStats := conf.Features["disable_cid_stats"]
-	enableContainers := enabledCIDStats || (conf.FargateOrchestrator != config.OrchestratorUnknown)
-	if enableContainers && !disabledCIDStats {
-		// only allow the ContainerID stats dimension if we're in a Fargate instance or it's
-		// been explicitly enabled and it's not prohibited by the disable_cid_stats feature flag.
-		in.ContainerID = containerID
-	}
-	return in
+	return Input{Traces: make([]traceutil.ProcessedTrace, 0, numChunks)}
```
raphaelgavache, line 123: mistake, adding back the containerID here

releasenotes/notes/enable-apm-stats-cid-collection-9a3371f51218c238.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    APM: Container-based second primary tags are now available by default.
```
jdgumz, line 4: > APM: Primary tags can now be selected from container-based tags by default.

We want to move away from "second primary" as a term.

jdgumz (changes requested):
Just a small suggestion on the update language and an implementation question.

One other thing: I believe we check for resolved container tags _before_ the stats concentrator and then again right before writing out the stats payload. I can't remember why but it'd be worth checking both places.

pkg/trace/stats/concentrator.go
```diff
@@ -139,6 +133,9 @@ func (c *Concentrator) Add(t Input) {
 // addNow adds the given input into the concentrator.
 // Callers must guard!
 func (c *Concentrator) addNow(pt *traceutil.ProcessedTrace, containerID string, containerTags []string) {
+	if !c.cidStats {
+		containerID = ""
+	}
```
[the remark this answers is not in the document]
raphaelgavache: As the test shows below, if cid = "" container tags don't pass. That was the previous behaviour too, filter was only on containerID.

releasenotes/notes/enable-apm-stats-cid-collection-9a3371f51218c238.yaml
raphaelgavache: what would be your naming choice instead? It's the current one in public doc https://docs.datadoghq.com/tracing/guide/setting_primary_tags_to_scope/?tab=helm#container-based-second-primary-tags

brett0000FF: Yeah, I haven't heard anything about changing the term. I definitely agree the term is confusing and could be improved, but I don't think we have a new term standardized?

brett0000FF (approved):

pkg/trace/stats/concentrator.go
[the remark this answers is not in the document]
jdgumz: Perhaps we drop them later if the container ID is an empty string.

releasenotes/notes/enable-apm-stats-cid-collection-9a3371f51218c238.yaml
brett0000FF: Great! Sounds good. Thanks for the info.

jdgumz (approved):

releasenotes/notes/enable-apm-stats-cid-collection-9a3371f51218c238.yaml
raphaelgavache: Removed the `second`, searching for "primary tag" in the public doc [quickly gives the result](https://docs.datadoghq.com/tracing/guide/setting_primary_tags_to_scope/?tab=helm)

pkg/trace/stats/concentrator.go
```diff
@@ -113,20 +116,11 @@ type Input struct {
...
-	_, enabledCIDStats := conf.Features["enable_cid_stats"]
-	_, disabledCIDStats := conf.Features["disable_cid_stats"]
-	enableContainers := enabledCIDStats || (conf.FargateOrchestrator != config.OrchestratorUnknown)
-	if enableContainers && !disabledCIDStats {
-		// only allow the ContainerID stats dimension if we're in a Fargate instance or it's
-		// been explicitly enabled and it's not prohibited by the disable_cid_stats feature flag.
-		in.ContainerID = containerID
-	}
-	return in
+	return Input{Traces: make([]traceutil.ProcessedTrace, 0, numChunks), ContainerID: containerID}
```
ichinaski, line 123: why don't we check here the `conf.Features["disable_cid_stats"]` anymore (and/or Fargate presence), similar to what we do on the concentrator constructor?

raphaelgavache: it's moved to the addStats in the concentrator, it was only checked there

ichinaski (approved):

PR #35194, opened 2025-03-17
Fixes to types captured on stack, structs in heap

grantseltzer:
### What does this PR do?

Fixes a series of unreported bugs that I observed in how location expressions were generated for types that are passed on the stack. For example, structs that are large enough to be passed on the stack instead of individual fields passed via registers were not be captured correctly. This covers a really common occurrence.

It also fixes issues with collisions in cases where structs have multiple fields of the same type resulting in incorrect data being captured.

### Motivation

Bug fixes

Ran e2e tests.

### Additional Notes

I have a fix for issues with arrays in a followup PR coming this week.


tylfin (approved):
LGTM

PR #35214, opened 2025-03-17
Add orch_cluster_id host tag

clamoriniere:
### What does this PR do?

Add a new host tag called ``orch_cluster_id`` that will be set to the
kubernetes cluster ID generated by the cluster-agent. This is added only
for the hosts that are part of a kubernetes cluster.
This tag is also used on every cluster level resources like
`kubernetes_state.*` metrics, kubernetes events and orchestrator level
resources.

### Motivation

Introduce a new tag to ensure accuracy of query result if several cluster use the same name.
since `orch_cluster_id` correspond to a UUID generated by kubernetes API-Server, it should be unique. 

E2e tests have been updated to check the presence of the new tag.
However to test it manually, you can deploy the agent and cluster-agent in a kubernetes cluster, then check
that the `orch_cluster_id` tag is present on `container.cpu.usage`, `kubernetes_state.daemonset.*` metrics
and `orchestrator resources`.

here is a notebook with the test result https://dddev.datadoghq.com/notebook/11879318/cedric-test-orch-cluster-id-tag-pr?view=view-mode


davidor (approved):

L3n41c (commented):
I left some suggestions to get rid of `fmt.Sprintf` and to replace it with strings concatenation because it’s significantly more efficient:

```go
package main

import (
        "fmt"
        "testing"

        "github.com/stretchr/testify/assert"
)

const (
        tagName = "foo"
        tagValue = "bar"
)

func BenchmarkConcatenation(b *testing.B) {
        var tag string
        for i := 0; i < b.N; i++ {
                tag = tagName + ":" + tagValue
        }
        assert.Equal(b, "foo:bar", tag)
}

func BenchmarkSprintf(b *testing.B) {
        var tag string
        for i := 0; i < b.N; i++ {
                tag = fmt.Sprintf("%s:%s", tagName, tagValue)
        }
        assert.Equal(b, "foo:bar", tag)
}
```
```console
$ go test -bench .
goos: linux
goarch: amd64
pkg: testconcat
cpu: 12th Gen Intel(R) Core(TM) i9-12900H
BenchmarkConcatenation-20       1000000000               0.2350 ns/op
BenchmarkSprintf-20             25886013                49.00 ns/op
PASS
ok      testconcat      1.588s
```

comp/metadata/host/hostimpl/hosttags/tags.go
```diff
@@ -125,14 +126,19 @@ func Get(ctx context.Context, cached bool, conf model.Reader) *Tags {
 	hname, _ := hostname.Get(ctx)
 	clusterName := clustername.GetClusterNameTagValue(ctx, hname)
 	if clusterName != "" {
-		clusterNameTags := []string{"kube_cluster_name:" + clusterName}
+		clusterNameTags := []string{fmt.Sprintf("%s:%s", tags.KubeClusterName, clusterName)}
```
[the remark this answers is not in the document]
clamoriniere: hi @L3n41c 

thanks for the tips, I was thinking that it was the opposite, but didn't tried to benchmark it. I should have.
I will update my PR to use it.

releasenotes/notes/Add-orch_cluster_id-host-tag-123dadf739977cd4.yaml
```diff
@@ -0,0 +1,16 @@
...
+# Each section note must be formatted as reStructuredText.
+---
+enhancements:
+  - |
+    Add a new host tag called ``orch_cluster_id`` that will be set to the
+    kubernetes cluster ID generated by the cluster-agent. This is added only
+    for the hosts that are part of a kubernetes cluster.
+    This tag is also used on every cluster level resources like
+    `kubernetes_state.*` metrics, kubernetes events and orchestrator level
+    resources.
```
maycmlee, line 16:
```suggestion
    Add a new host tag called ``orch_cluster_id`` that is set to the
    Kubernetes cluster ID generated by the cluster-agent. This is added only
    for the hosts that are part of a Kubernetes cluster.
    This tag is also used on every cluster-level resources like
    `kubernetes_state.*` metrics, kubernetes events, and orchestrator level
    resources.
```

maycmlee (approved):
Just some small suggestion, but approving to not block merge.

dustmop (approved):
LGTM for agent-configuration

pkg/collector/corechecks/cluster/ksm/kubernetes_state.go
```diff
@@ -904,12 +908,26 @@ func (k *KSMCheck) getClusterName() {
...
+	if err != nil {
+		log.Warnf("Error retrieving the cluster ID: %s", err)
+		return
+	}
+	k.clusterIDTagValue = clusterID
+}
+
 // initTags avoids keeping a nil Tags field in the check instance
 // Sets the kube_cluster_name tag for all metrics.
+// Sets the orch_cluster_is tag for all metrics.
```
jennchenn, line 922:
```suggestion
// Sets the orch_cluster_id tag for all metrics.
```

jennchenn (approved):

PR #35216, opened 2025-03-18
Add logs to show elem, tree in Autodiscovery

taromn:
### What does this PR do?
- show elem value
- show tree that will be processed 

### Motivation
To improve visibility
```
| CORE | ERROR | (comp/core/autodiscovery/configresolver/configresolver.go:xxx in resolveDataWithTemplateVars) | Unknown type: xxx
```
↑ Now we are not able to see the value of elem that caused the error. 
We can see just the type of it.
It doesn't provide enough information for troubleshooting.

### DIFF
--- a/comp/core/autodiscovery/configresolver/configresolver.go
+++ b/comp/core/autodiscovery/configresolver/configresolver.go
@@ -216,6 +216,8 @@ func resolveDataWithTemplateVars(ctx context.Context, data integration.Data, svc
 		return data, err
 	}
 
+	log.Debugf("tree to parse is: %v", tree)
+
 	type treePointer struct {
 		get func() interface{}
 		set func(interface{})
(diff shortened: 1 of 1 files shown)


wdhif:
Hey 👋 thanks for the PR! I'm just worried that this change could leak config data in logs. I'll link back with my team and let you know 👍

wdhif:
> Hey 👋 thanks for the PR! I'm just worried that this change could leak config data in logs. I'll link back with my team and let you know 👍

In its current state, this PR could leak configuration data into the logs. Are you able to try the `agent status` command as it should be sufficient for debugging purposes.

taromn:
Thanks for the review!

wdhif (changes requested):
Submitting review regarding outlined concerns: https://github.com/DataDog/datadog-agent/pull/35216#issuecomment-2743863963

PR #35220, opened 2025-03-18
Add container instance arn to ecs tasks

aureleoules:
### What does this PR do?

We retrieve the container instance arn from the workloadmeta ecs collector and pass it down to the ecs tasks.
All tasks collected by the same agent share the same container instance because the agent is deployed as a "Daemon" on ECS which deploys the agent on all of the cluster hosts

### Motivation

https://datadoghq.atlassian.net/browse/CAP-2350

### DIFF
--- a/pkg/util/ecs/metadata/v1/types.go
+++ b/pkg/util/ecs/metadata/v1/types.go
@@ -13,8 +13,9 @@ type Commands struct {
 // Instance represents the instance metadata exposed by the ECS introspection endpoint.
 // See http://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-agent-introspection.html
 type Instance struct {
-	Cluster string `json:"Cluster"`
-	Version string `json:"Version"`
+	Cluster              string `json:"Cluster"`
+	Version              string `json:"Version"`
+	ContainerInstanceARN string `json:"ContainerInstanceARN"`
 }
 
 // Tasks represents the list of task exposed by the ECS introspection endpoint.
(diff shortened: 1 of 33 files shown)


fisherevans (approved):

jhgilbert (approved):

pkg/util/ecs/metadata/v1/types.go
```diff
@@ -13,8 +13,9 @@ type Commands struct {
 // Instance represents the instance metadata exposed by the ECS introspection endpoint.
 // See http://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-agent-introspection.html
 type Instance struct {
-	Cluster string `json:"Cluster"`
-	Version string `json:"Version"`
+	Cluster              string `json:"Cluster"`
+	Version              string `json:"Version"`
+	ContainerInstanceARN string `json:"ContainerInstanceARN"`
```
L3n41c, line 18: Given that we’re already in the `Instance` `struct`, having `Instance` as well in the field name sounds redundant.
Would it be confusing to simplify this by:
```suggestion
	Arn string `json:"Arn"`
```

L3n41c (approved):
👍🏻 LGTM for files owned by @DataDog/container-integrations.

pkg/util/ecs/metadata/v1/types.go
aureleoules: This is the response format AWS uses so we can't change the json struct tag

adel121:
Good for @DataDog/container-platform owned files

adel121 (approved):
Good for @DataDog/container-platform  owned files

PR #35223, opened 2025-03-18
[EBPF] gpu: disable fatbin parsing by default

gjulianm:
### What does this PR do?

This PR disables fatbin parsing by default, adding a config flag.

### Motivation

As we want to split shared memory usage from the memory metrics and fatbin parsing is computationally expensive, we're disabling it by default until we add more features that require that feature.

CI passing.

### DIFF
--- a/pkg/gpu/probe_test.go
+++ b/pkg/gpu/probe_test.go
@@ -46,6 +46,9 @@ func (s *probeTestSuite) getProbe() *Probe {
 	// Avoid waiting for the initial sync to finish in tests, we don't need it
 	cfg.InitialProcessSync = false
 
+	// Enable fatbin parsing in tests so we can validate it runs
+	cfg.EnableFatbinParsing = true
+
 	deps := ProbeDependencies{
 		NvmlLib:        testutil.GetBasicNvmlMock(),
 		ProcessMonitor: consumerstestutil.NewTestProcessConsumer(t),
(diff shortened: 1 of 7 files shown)


pkg/gpu/probe_test.go
```diff
@@ -46,6 +46,9 @@ func (s *probeTestSuite) getProbe() *Probe {
 	// Avoid waiting for the initial sync to finish in tests, we don't need it
 	cfg.InitialProcessSync = false
 
+	// Enable fatbin parsing in tests so we can validate it runs
+	cfg.EnableFatbinParsing = true
```
val06, line 50: let's add another case to test this flag as false, so we can ensure the memory.usage metric is equal to the global mamory (aka const and shared memory are reported as 0)

val06 (changes requested):
reviewed

hush-hush (approved):

pkg/gpu/probe_test.go
gjulianm: In this test we're not actually parsing fatbin files, as the file that runs is a C sample without CUDA calls. This toggle just ensures that the code that tries to find the kernel doesn't crash and just continues when it can't find the kernel. The stream tests do try to get the kernel data, so I added a [test there](https://github.com/DataDog/datadog-agent/pull/35223/commits/afd271381a874680625763c5f063fa97f57c7584) to ensure the config flag actually changes behavior.

val06 (approved):

PR #35236, opened 2025-03-18
[CWS] Add kernel compilation flags in sysctl snapshot event

Gui774ume:
### What does this PR do?

This PR adds the local kernel compilation flags to the `sysctl_snapshot` event.

### Motivation

Kernel compilation flags are part of the configuration parameters that should be checked in order to assess the hardening posture of a host.

### DIFF
--- a/pkg/config/setup/system_probe_cws.go
+++ b/pkg/config/setup/system_probe_cws.go
@@ -124,6 +124,7 @@ func initCWSSystemProbeConfig(cfg pkgconfigmodel.Config) {
 	cfg.BindEnvAndSetDefault("runtime_security_config.sysctl.snapshot.enabled", true)
 	cfg.BindEnvAndSetDefault("runtime_security_config.sysctl.snapshot.period", "1h")
 	cfg.BindEnvAndSetDefault("runtime_security_config.sysctl.snapshot.ignored_base_names", []string{"netdev_rss_key", "stable_secret"})
+	cfg.BindEnvAndSetDefault("runtime_security_config.sysctl.snapshot.kernel_compilation_flags", kernelCompilationFlags)
 
 	// CWS - UserSessions
 	cfg.BindEnvAndSetDefault("runtime_security_config.user_sessions.cache_size", 1024)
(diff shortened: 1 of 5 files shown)


pkg/security/probe/sysctl/snapshot.go
```diff
@@ -247,6 +257,75 @@ func (s *Snapshot) snapshotKernelCmdline(ignoredBaseNames []string) error {
 	return nil
 }
 
+func (s *Snapshot) getKernelConfigPath() (string, error) {
+	kernelVersion, err := os.ReadFile(kernel.HostProc("/sys/kernel/osrelease"))
```
brycekahle, line 261: can you use `pkg/util/kernel.Release()`? It is memoized.

pkg/security/probe/sysctl/snapshot.go
```diff
@@ -47,11 +50,11 @@ type SnapshotEvent struct {
 }
 
 // NewSnapshotEvent returns a new sysctl snapshot event
-func NewSnapshotEvent(ignoredBaseNames []string) (*SnapshotEvent, error) {
+func NewSnapshotEvent(ignoredBaseNames []string, kernelCompilationFLags []string) (*SnapshotEvent, error) {
```
brycekahle, line 53:
```suggestion
func NewSnapshotEvent(ignoredBaseNames []string, kernelCompilationFlags []string) (*SnapshotEvent, error) {
```

pkg/security/probe/sysctl/snapshot.go
```diff
@@ -247,6 +257,75 @@ func (s *Snapshot) snapshotKernelCmdline(ignoredBaseNames []string) error {
...
+}
+
+func (s *Snapshot) parseKernelConfig(r io.Reader, kernelCompilationFLags []string) error {
+	scanner := bufio.NewScanner(r)
+	for scanner.Scan() {
+		line := strings.TrimSpace(scanner.Text())
+		if strings.HasPrefix(line, "#") {
+			if strings.HasSuffix(line, "is not set") {
+				key := strings.Fields(line)[1]
+				if slices.Contains(kernelCompilationFLags, key) {
```
brycekahle, line 283: `bytes.Fields` tends to be better from an allocation point of view, if you only need one/few of the fields.
```suggestion
				key := bytes.Fields(line)[1]
				if slices.Contains(kernelCompilationFlags, string(key)) {
```

brycekahle, line 283: Would it be worth building a map to check here, rather than doing `slices.Contains` in a loop?

brycekahle (approved):
minor typos and suggestions, but nothing blocking

pkg/config/setup/system_probe_cws.go
```diff
@@ -149,3 +150,106 @@ func initCWSSystemProbeConfig(cfg pkgconfigmodel.Config) {
 
 	cfg.BindEnvAndSetDefault("runtime_security_config.network_monitoring.enabled", false)
 }
+
+var (
```
hush-hush, line 154: This list would end up in all the agent metadata payload of the Agent and in REDAPL. This would be a significant increase in size.

Do we need this in the configuration ? Could we define a empty slice here and move the list to the system-probe code ? how likely are users to edit this list ?

pkg/security/probe/sysctl/snapshot.go
Gui774ume: Updated, thanks !

pkg/security/probe/sysctl/snapshot.go
Gui774ume: Updated, thanks !

Gui774ume: You're right ... I'll generate a map from the slice config. Thanks !

pkg/security/probe/sysctl/snapshot.go
```diff
@@ -247,6 +257,75 @@ func (s *Snapshot) snapshotKernelCmdline(ignoredBaseNames []string) error {
...
+					s.KernelCompilationConfiguration[key] = "not_set"
+				}
+			}
+			continue
+		}
+		parts := strings.SplitN(line, "=", 2)
+		if len(parts) == 2 && slices.Contains(kernelCompilationFLags, parts[0]) {
+			s.KernelCompilationConfiguration[parts[0]] = strings.ReplaceAll(parts[1], "\"", "")
```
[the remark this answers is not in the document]
Gui774ume: Trim should be enough, thanks !

pkg/config/setup/system_probe_cws.go
Gui774ume: No worries, I moved it to the `system-probe` package. It's more of a safeguard in case we missed something, or if anything goes wrong with this feature.

pkg/security/probe/sysctl/snapshot.go
```diff
@@ -247,6 +258,75 @@ func (s *Snapshot) snapshotKernelCmdline(ignoredBaseNames []string) error {
...
+		if strings.HasPrefix(line, "#") {
+			if strings.HasSuffix(line, "is not set") {
+				key := bytes.Fields([]byte(line))[1]
+				if _, ok := kernelCompilationFlags[string(key)]; ok {
+					s.KernelCompilationConfiguration[string(key)] = "not_set"
+				}
+			}
+			continue
+		}
+		parts := strings.SplitN(line, "=", 2)
```
paulcacheux, line 290: small nit, maybe for a follow up, but `strings.Cut` is much cleaner and doesn't allocate

pkg/util/kernel/fs.go
```diff
@@ -59,6 +59,19 @@ var SysFSRoot = funcs.MemoizeNoError(func() string {
 	return "/sys"
 })
 
+// BootRoot retrieves the current boot dir we should use
+var BootRoot = funcs.MemoizeNoError(func() string {
+	if v := os.Getenv("HOST_BOOT"); v != "" {
+		return v
+	}
+	if os.Getenv("DOCKER_DD_AGENT") != "" {
```
paulcacheux, line 67: `env.IsContainerized` is doing the same thing, but it more cleanly expresses the goal

safchain (approved):

paulcacheux (approved):

spikat (approved):

hush-hush (approved):

pkg/security/probe/sysctl/snapshot.go
Gui774ume: Done 👍🏻

pkg/security/probe/sysctl/snapshot.go
```diff
@@ -247,6 +258,75 @@ func (s *Snapshot) snapshotKernelCmdline(ignoredBaseNames []string) error {
 	return nil
 }
 
+func (s *Snapshot) getKernelConfigPath() (string, error) {
+	kernelVersion, err := kernel.Release()
+	if err != nil {
+		return "", err
+	}
+	configPath := fmt.Sprintf(kernel.HostBoot("/config-%s"), strings.TrimSpace(string(kernelVersion)))
```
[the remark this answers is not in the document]
Gui774ume: It's a symlink no ? I'll keep it this way for now, we can always add the sylink later if need be 👍🏻

PR #35237, opened 2025-03-18
[Dynamic Instrumentation] Make `param_stack` a per-CPU map to prevent data corruption

GreenMatan:
### What does this PR do?
This PR modifies the `param_stack` map (of type `BPF_MAP_TYPE_STACK`) to be per-CPU rather than shared across all CPUs.

Previously, `param_stack` was a single instance shared across all CPUs for a single BPF program, potentially leading to corruption when multiple CPUs executed the same BPF program concurrently.
Now, each CPU gets its own instance of `param_stack`, ensuring execution remains isolated and correct per CPU.

### Motivation
Avoid multiple CPUs race. Now there is `param_stack_cpu_<i>` map per-cpu, generated dynamically with templating.

Dynamic Instrumentation's e2e tests are sufficient for now. They all pass.

### Additional Notes
Fixes DEBUG-3259.

### DIFF
--- a/pkg/dynamicinstrumentation/codegen/c/dynamicinstrumentation.c
+++ b/pkg/dynamicinstrumentation/codegen/c/dynamicinstrumentation.c
@@ -114,6 +114,22 @@ int {{.GetBPFFuncName}}(struct pt_regs *ctx)
         return 0;
     }
 
+    u32 cpu = bpf_get_smp_processor_id();
+
+    void *param_stack_map;
+    switch (cpu) {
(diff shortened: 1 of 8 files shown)


grantseltzer (approved):
LGTM! The only thing that I think about with this is that at some point we may explore getting rid of templating, instead moving the location expression generation into a simple VM in BPF. At that point we'd certainly have to create the bpf maps in Go which is regardless equally good to this approach. But for now this'll work just fine.

PR #35250, opened 2025-03-19
[CONTP-681] Add 'fips' to cluster-agent version tag if it is built with fips flag

zhuminyi:
### What does this PR do?
If agent version has prefix containing fips, the version string will become numberVersion + prefix (with "fips")

### Motivation
When dca fips image is used, the agent version doesn't include prefix "fips" because it only includes number in current setup. 
Screenshot 2025-03-18 at 12 31 04 PM

With this change, the new version will be like  "7.64.0-rc.13-fips"

Validate by checking fed.staging with the version tag of 7.65.9-rc.1-fips

### DIFF
--- a/cmd/cluster-agent/api/agent/agent.go
+++ b/cmd/cluster-agent/api/agent/agent.go
@@ -12,6 +12,7 @@ import (
 	"encoding/json"
 	"io"
 	"net/http"
+	"strings"
 
 	"github.com/gorilla/mux"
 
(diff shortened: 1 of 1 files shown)


chouquette:
AFAICS this change doesn't cause `FIPS` to be included in the version string:
```
./datadog-cluster-agent version
Cluster Agent 7.65.0-devel - Meta: git.544.1c500d6 - Commit: 1c500d6 - Serialization version: v5.0.144 - Go version: go1.23.6
```

PR #35257, opened 2025-03-19
upgrade(installer): Make agent experiment a simple unit

BaptisteFoy:
### What does this PR do?
Transform the datadog-agent-exp unit into a "simple" unit instead of a "oneshot" unit.

This PR makes sure that any given units except the core agent can only reference the following units:
- Itself (exp or stable)
- Core agent (exp or stable)

In turn, the core agent can only reference:
- Itself (stable or experiments)
- Other units in the same tree (stable references stables, experiment references experiments)

This will let us add or remove units more smoothly

### Motivation
Remove dependencies between units

E2E, manual tests on CentOS 7

We can't have a timeout baked in the unit, so we use /usr/bin/timeout. It's included in coreutils and should be available everywhere. If there's an OS where it's not present, we can add it to the agent package (<50kb).

### DIFF
--- a/test/new-e2e/tests/installer/host/systemd.go
+++ b/test/new-e2e/tests/installer/host/systemd.go
@@ -39,7 +39,7 @@ func (h *Host) LastJournaldTimestamp() JournaldTimestamp {
 func (h *Host) AssertUnitProperty(unit, property, value string) {
 	res, err := h.remote.Execute(fmt.Sprintf("sudo systemctl show -p %s %s", property, unit))
 	require.NoError(h.t, err)
-	require.Equal(h.t, fmt.Sprintf("%s=%s\n", property, value), res)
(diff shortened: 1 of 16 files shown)


test/new-e2e/tests/installer/host/systemd.go
```diff
@@ -58,7 +58,7 @@ func eventMatches(log journaldLog, event SystemdEvent) bool {
 	if log.Unit != event.Unit {
 		return false
 	}
-	match, _ := regexp.MatchString(event.Pattern, log.Message)
```
BaptisteFoy, line 61: Case insensitiveness as older OSes (CentOS / AL2) use `failed` instead of `Failed`.

raphaelgavache (approved):
nice, lgtm

coignetp (approved):

PR #35259, opened 2025-03-19
Use agent package from Gitlab artifacts for e2e tests

KevinFairise2:
### What does this PR do?

Pass the `--local-package` flag to all e2e tests, so that they will use the package from Gitlab artifacts instead of downloading the agent from s3 bucket.
The PR is not doing a lot, all the logic is done on test-infra-definitions side.

Note: This PR will only change the way the agent is installed for the tests that rely on the e2e framework to handle the agent installation. More PRs will come to update the way the agent is installed in other tests.

### Motivation

Stop relying on the install-script for the e2e tests running in the CI. That is a first step to allow us using non-signed agent package in the pipeline.
This PR might also reduce the cost we have from download the packages behind Cloudfront.


CelianR (approved):

test/new-e2e/tests/agent-platform/package-signing/package_signing_test.go
```diff
@@ -92,6 +92,26 @@ func TestPackageSigningComponent(t *testing.T) {
...
+		is.Env().RemoteHost.MustExecute("sudo apt-get update && sudo DEBIAN_FRONTEND=noninteractive apt-get install -y apt-transport-https curl gnupg")
+		aptUsrShareKeyring := "/usr/share/keyrings/datadog-archive-keyring.gpg"
+		is.Env().RemoteHost.MustExecute(fmt.Sprintf("sudo touch %s && sudo chmod a+r %s", aptUsrShareKeyring, aptUsrShareKeyring))
+		keys := []string{"DATADOG_APT_KEY_CURRENT.public", "DATADOG_APT_KEY_C0962C7D.public", "DATADOG_APT_KEY_F14F620E.public", "DATADOG_APT_KEY_382E94DE.public"}
+		for _, key := range keys {
+			is.Env().RemoteHost.MustExecute(fmt.Sprintf("sudo curl --retry 5 -o \"/tmp/%s\" \"https://keys.datadoghq.com/%s\"", key, key))
+			is.Env().RemoteHost.MustExecute(fmt.Sprintf("sudo cat \"/tmp/%s\" | sudo gpg --import --batch --no-default-keyring --keyring \"%s\"", key, aptUsrShareKeyring))
+		}
+		aptTrustedDKeyring := "/etc/apt/trusted.gpg.d/datadog-archive-keyring.gpg"
+		is.Env().RemoteHost.MustExecute(fmt.Sprintf("sudo cp %s %s", aptUsrShareKeyring, aptTrustedDKeyring))
```
chouetz, line 106:
```suggestion
		aptTrustedKeyring := "/etc/apt/trusted.gpg.d/datadog-archive-keyring.gpg"
		is.Env().RemoteHost.MustExecute(fmt.Sprintf("sudo cp %s %s", aptUsrShareKeyring, aptTrustedKeyring))
```
Fixes a typo. As discussed, I think this step is not necessary, but we can keep it to be safe

chouetz (approved):
Probably only minor comments

PR #35261, opened 2025-03-19
GPU: added host_gpu tag to host tags

val06:
### What does this PR do?

[Jira Ticket](https://datadoghq.atlassian.net/browse/EBPF-685)

Adds a new host_gpu:true tag to hosts that have at least one gpu device.
This extra tag is collected only when `collect_gpu_tags` config value is `true`, currently the default is `false`

### Motivation

allow effective filtering of gpu hosts in GPU Monitoring product

- added UTs
- manually tested one host with gpu devices and another host without

It is critical to return the host payload very fast, hence we need to make sure the GPU tags retrieval isn't a costly operation.
Added manual benchmark test to measure the gpu.GetTags function (main bottleneck is loading the native library)

### Additional Notes

The implementation is using go-nvml library to match the behavior of the GPU core-check.
**Currently only linux hosts are supported and only Nvidia as GPU vendor.**

### DIFF
--- a/pkg/gpu/tags/tags_nix.go
+++ b/pkg/gpu/tags/tags_nix.go
@@ -13,6 +13,7 @@ import (
 	"github.com/NVIDIA/go-nvml/pkg/nvml"
 )
 
+// used for tests to mock the NVML library
 var nvmlLibrary nvml.Interface
 
 // GetTags returns gpu_host:true if any NVIDIA GPUs are present, nil otherwise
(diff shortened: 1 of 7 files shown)


gjulianm (approved):

pkg/gpu/tags/tags_nix.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"github.com/DataDog/datadog-agent/pkg/util/log"
+	"github.com/NVIDIA/go-nvml/pkg/nvml"
+)
+
+var nvmlLibrary nvml.Interface
+
+// GetTags returns gpu_host:true if any NVIDIA GPUs are present, nil otherwise
+func GetTags() []string {
+	if nvmlLibrary == nil {
```
hush-hush, line 20: This should only be called once, do we need to save the nvm lib in a global var ? If it's needed we need some king of mutex/sync to protect from race cond.

hush-hush (changes requested):

hush-hush (approved):

jhgilbert (approved):

PR #35271, opened 2025-03-19
ci(source_test): Add a serverless-init test

chouetz:
### What does this PR do?
Add a `serverless-init` test in the gitlab pipeline of `datadog-agent`

### Motivation
We need to confirm that serverless-init can build properly before new PRs are merged into the agent. We have dealt with multiple incidents in the last year due to an engineer outside our team making an update to a dependency or unrelated file that broke the deployment for serverless-init. We want to make it clearer that serverless-init depends on the Datadog agent by having this go test command run
https://datadoghq.atlassian.net/browse/ACIX-585

The tests passed on the executed [job](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/856458116)
```
$ go test -tags "serverless otlp test" ./cmd/serverless-init/.
ok  	github.com/DataDog/datadog-agent/cmd/serverless-init	0.168s
```

### DIFF
--- a/.gitlab/source_test/linux.yml
+++ b/.gitlab/source_test/linux.yml
@@ -131,6 +131,19 @@ tests_rpm-arm64-py3:
   variables:
     CONDA_ENV: ddpy3
 
+tests_serverless-init:
+  extends: .linux_x64
+  stage: source_test
+  needs: ["go_deps"]
+  rules:
+    - !reference [.except_disable_unit_tests]
+    - !reference [.except_mergequeue]
+    - when: on_success
+  script:
+    - !reference [.retrieve_linux_go_deps]
+    - go test -tags "serverless otlp test" ./cmd/serverless-init/.
+
+
 # Check consistency of go.mod file with project imports
 go_mod_tidy_check:
   stage: source_test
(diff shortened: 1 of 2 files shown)


CelianR (approved):
IIUC, we have to build this everytime even though serveless files are not changed? If this is not the case, we should add a rule for change paths

chouetz:
> IIUC, we have to build this everytime even though serveless files are not changed? If this is not the case, we should add a rule for change paths

Yes I understood it's not "serverless" related, we must ensure at any time this test is ok. As it's not a costly one I chose to execute it all the time.

.gitlab/source_test/linux.yml
```diff
@@ -131,6 +131,19 @@ tests_rpm-arm64-py3:
   variables:
     CONDA_ENV: ddpy3
 
+tests_serverless:
```
nina9753, line 134: small nit: can we update the test name to be `build_serverless-init` or `tests_serverless-init`

nina9753 (approved):
lgtm! just one small nit

nina9753:
> IIUC, we have to build this everytime even though serveless files are not changed? If this is not the case, we should add a rule for change paths

Serverless-init relies on the dependencies built from the agent so we would want to build/test this on all pr merged into the datadog agent

PR #35275, opened 2025-03-19
[AGNTLOG-56] Update registry writer to not write atomically when Agent runs on ECS Fargate

aqian01:
### What does this PR do?
Adds a new configuration option `logs_config.atomic_registry_write` to control whether registry writes are atomic. The registry writing functionality is now a component with two implementations (atomic and non-atomic) that are swapped based on the config value. Defaults to true, but automatically set to false when running on ECS Fargate.

### Motivation

There is an increased memory leak when the Agent runs on ECS Fargate. This is due to the way AWS has implemented Fargate containers and disk management. When you create a file in their containers and delete it, it doesn’t delete it but instead [creates it in memory](https://github.com/aws/amazon-ecs-agent/issues/3594). Atomically writing creates a temporary file to write to which would cause our issue.

[Ticket](https://datadoghq.atlassian.net/jira/software/c/projects/AGNTLOG/boards/8654?assignee=712020%3A58eaddc6-2677-4b63-82b0-a98b3e9f4a84&selectedIssue=AGNTLOG-56)

[Issue](https://datadoghq.atlassian.net/browse/AGENT-12980)

Unit tests and deploying my changes with ECS Fargate.

image


estherk15 (approved):

releasenotes/notes/update-registry-non-atomic-c204f35fc48f9b82.yaml
```diff
@@ -0,0 +1,11 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+enhancements:
+  - |
+    Update registry writer to not write atomically when Agent runs on ECS Fargate.
```
hush-hush, line 11: Should we explain why this is an `enhancements` ? I'm not sure the current version of this releasenote is useful to users.

comp/logs/auditor/impl/auditor.go
```diff
@@ -77,10 +78,17 @@ type Provides struct {
 // newAuditor is the public constructor for the auditor
 func newAuditor(deps Dependencies) *registryAuditor {
 	runPath := deps.Config.GetString("logs_config.run_path")
-	// filename := deps.Config.GetString("logs_config.registry_filename")
 	filename := DefaultRegistryFilename
```
hush-hush, line 81: Why not get the value from the config ?

pkg/config/setup/config_darwin.go
```diff
@@ -31,6 +31,8 @@ const (
 	defaultStatsdSocket = ""
 	//DefaultStreamlogsLogFile points to the stream logs log file that will be used if not configured
 	DefaultStreamlogsLogFile = "/opt/datadog-agent/logs/streamlogs_info/streamlogs.log"
+	// DefaultRegistryFilename is the default registry filename
+	DefaultRegistryFilename = "registry.json"
```
hush-hush, line 35: Since it's the same value for all system, can we move this to config.go ? Also, do we need this default to be public ?

hush-hush (changes requested):

pkg/config/setup/config.go
```diff
@@ -120,6 +120,9 @@ const (
 
 	// DefaultNetworkPathMaxTTL defines the default maximum TTL for traceroute tests
 	DefaultNetworkPathMaxTTL = 30
+
+	// DefaultRegistryFilename is the default registry filename
+	DefaultRegistryFilename = "registry.json"
```
hush-hush, line 125: Why more this default here if it's never used in the config ? This creates unnecessary dependency to the `setup` package.

releasenotes/notes/update-registry-non-atomic-c204f35fc48f9b82.yaml
hush-hush, line 11: Sorry to be nit-picky but we should explain why using atomic write are not a good thing for ECS Fargate (most dev will go with the assumption that atomic write are faster than mutexes).

hush-hush (approved):
Added one last comment on the releasenote but looks good otherwise.

ddrthall (approved):

PR #35280, opened 2025-03-19
Unset'ing a setting removes it from AllSettings

dustmop:
### What does this PR do?

Unset'ing a setting removes it from AllSettings
    
For settings that have no default, but then get assigned a value later, calling Unset will remove that setting entirely. The same is true for SetWithoutSource when passed a nil value.

### Motivation

Match behavior with Viper.

Unit test added to viper and nodetreemodel to validate the behavior works in both.


pkg/config/setup/config.go
```diff
@@ -272,7 +272,15 @@ func init() {
...
-	systemProbe = viperconfig.NewConfig("system-probe", "DD", strings.NewReplacer(".", "_")) // nolint: forbidigo // legit use case
+	if envvar == "enable" {
+		systemProbe = nodetreemodel.NewConfig("system-probe", "DD", strings.NewReplacer(".", "_")) // nolint: forbidigo // legit use case
+	} else if envvar == "tee" {
+		viperConfig := viperconfig.NewConfig("system-probe", "DD", strings.NewReplacer(".", "_"))      // nolint: forbidigo // legit use case
+		nodetreeConfig := nodetreemodel.NewConfig("system-probe", "DD", strings.NewReplacer(".", "_")) // nolint: forbidigo // legit use case
+		systemProbe = teeconfig.NewTeeConfig(viperConfig, nodetreeConfig)
+	} else {
+		systemProbe = viperconfig.NewConfig("system-probe", "DD", strings.NewReplacer(".", "_")) // nolint: forbidigo // legit use case
+	}
```
hush-hush, line 283: Why not reuse the if/else above ?

pkg/config/nodetreemodel/config.go
```diff
@@ -215,6 +215,10 @@ func (c *ntmConfig) Set(key string, newValue interface{}, source model.Source) {
 
 // SetWithoutSource assigns the value to the given key using source Unknown
 func (c *ntmConfig) SetWithoutSource(key string, value interface{}) {
+	if value == nil {
+		c.UnsetForSource(key, model.SourceUnknown)
+		return
+	}
```
hush-hush, line 221: Do we know where this is used ? Could is be possible to migrate use cases to UnsetForSource instead ?

dustmop: I was initially concerned with having feature drift, but then again the method `SetWithoutSource` should only be used in tests, and there only a small number of tests that rely on this behavior of how `AllSettings` acts. I updated the tests.

hush-hush (approved):

PR #32310, opened 2024-12-17
remove unused choco related files

sabrina-datadog:
### What does this PR do?
Remove the unused chocolatey jobs:
- offline chocolate build job (windows_choco_offline_7_x64)
- choco deploy job (publish_choco_7_x64) 

Rename and reorganize the file structure without "online" since there is no more offline.

### Motivation
After chocolatey has been automated, there is no need for the manual components. ([BARX-684](https://datadoghq.atlassian.net/browse/BARX-684))

### Describe how you validated your changes

1. checked that the offline job is no longer in the choco build stage and the choco publish job isn't in the install_script_deploy stage
image

2. manually ran the choco build jobs to ensure they still work

### Additional Notes

[BARX-684]: https://datadoghq.atlassian.net/browse/BARX-684?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #35299, opened 2025-03-20
Allow custom choco releases

AliDatadog:
### What does this PR do?
This PR fixes an issue with choco jobs failing for custom builds http://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/856594274.

I think the issue was introduced in https://github.com/DataDog/datadog-agent/pull/32310/files#diff-2bb2ce257e8d6f43475d0663271e9aa730428491a54fadde5075ae5ab50126e0, adding ^$ to the regex would enforce strict matching.

### DIFF
--- a/tasks/winbuildscripts/Generate-Chocolatey-Package.ps1
+++ b/tasks/winbuildscripts/Generate-Chocolatey-Package.ps1
@@ -63,7 +63,7 @@ if (![string]::IsNullOrEmpty($VersionOverride)) {
 $copyright = "Datadog {0}" -f (Get-Date).Year
 
 $releasePattern = "(\d+\.\d+\.\d+)"
-$releaseCandidatePattern = "^(\d+\.\d+\.\d+)-rc\.(\d+)$"
+$releaseCandidatePattern = "(\d+\.\d+\.\d+)-rc\.(\d+)"
 $develPattern = "^(\d+\.\d+\.\d+)-devel\.git\.\d+\.(.+)"
 
 # Build the package in a temporary directory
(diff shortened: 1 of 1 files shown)


tasks/winbuildscripts/Generate-Chocolatey-Package.ps1
```diff
@@ -62,7 +62,7 @@ if (![string]::IsNullOrEmpty($VersionOverride)) {
 }
 $copyright = "Datadog {0}" -f (Get-Date).Year
 
-$releasePattern = "^(\d+\.\d+\.\d+)$"
+$releasePattern = "(\d+\.\d+\.\d+)"
 $releaseCandidatePattern = "^(\d+\.\d+\.\d+)-rc\.(\d+)$"
```
chouquette, line 66: Are you keeping this pattern as is because the assumption is that it's never used for custom builds, and therefor will never need to handle a suffix?

chouquette (approved):
LGTM with a side question

tasks/winbuildscripts/Generate-Chocolatey-Package.ps1
AliDatadog: I'm just reverting to the previous state honestly

clarkb7: i used it for the fips preview builds that encountered this issue.
```
7.63.0-rc.7-fips-preview
```

chouquette: Agreed for `releasePattern` but my question was about `releaseCandidatePattern` which was also changed by https://github.com/DataDog/datadog-agent/pull/32310/files#diff-2bb2ce257e8d6f43475d0663271e9aa730428491a54fadde5075ae5ab50126e0R64-R66

clarkb7 (approved):

PR #35303, opened 2025-03-20
[CWS] Add 'log' rule action

lebauce:
### What does this PR do?

Add a 'log' action to display a log when a rule is triggered.


pkg/security/secl/rules/ruleset.go
```diff
@@ -609,6 +610,36 @@ func (rs *RuleSet) runSetActions(_ eval.Event, ctx *eval.Context, rule *Rule) er
...
+
+		switch {
+		// other actions are handled by ruleset listeners
+		case action.Def.Log != nil:
+			message := action.Def.Log.Message
+			if message == "" {
+				message = fmt.Sprintf("Rule %s triggered", rule.ID)
+			}
+
+			switch action.Def.Log.Level {
```
paulcacheux, line 627: might be a good idea to do a `strings.ToUpper` or similar, to be a bit more resilient to cases

paulcacheux (approved):

PR #35308, opened 2025-03-20
Avoid calling diagnose on nil environment

KevinFairise2:
### What does this PR do?

The Host diagnose method panics and break normal workflow if the environment is `nil`. Which can happen if the environment failed to initialize properly

### Motivation

Avoid unexpected panics that prevent resource cleaning

### DIFF
--- a/test/new-e2e/pkg/e2e/suite.go
+++ b/test/new-e2e/pkg/e2e/suite.go
@@ -510,11 +510,13 @@ func (bs *BaseSuite[Env]) SetupSuite() {
 		// run environment diagnose
 		if diagnosableEnv, ok := any(bs.env).(common.Diagnosable); ok {
 			// at least one test failed, diagnose the environment
-			diagnose, diagnoseErr := diagnosableEnv.Diagnose(bs.SessionOutputDir())
-			if diagnoseErr != nil {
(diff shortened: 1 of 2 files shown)


test/new-e2e/pkg/e2e/suite.go
```diff
@@ -510,11 +510,13 @@ func (bs *BaseSuite[Env]) SetupSuite() {
 		// run environment diagnose
 		if diagnosableEnv, ok := any(bs.env).(common.Diagnosable); ok {
```
pgimalac, line 511: You can just do this to avoid adding one indentation level below
```suggestion
		if diagnosableEnv, ok := any(bs.env).(common.Diagnosable); ok && diagnosableEnv != nil {
```

test/new-e2e/pkg/e2e/suite.go
```diff
@@ -593,11 +595,14 @@ func (bs *BaseSuite[Env]) AfterTest(suiteName, testName string) {
 			// run environment diagnose if the test failed
 			if diagnosableEnv, ok := any(bs.env).(common.Diagnosable); ok {
```
pgimalac, line 596: Same

pgimalac (approved):

PR #35311, opened 2025-03-20
[AGENTCFG-155] Adding a secrets section to datadog-agent status command output

rahulkaukuntla:
### What does this PR do?
Refer to the description provided [here](https://datadoghq.atlassian.net/browse/AGENTCFG-155?focusedCommentId=2234118)

The output after enabling secrets, and running the status command (just included the secrets sections):
Screenshot 2025-03-20 at 10 50 19 AM

If secrets are not enabled, this is what the secrets section of the status command looks like:
Screenshot 2025-03-20 at 10 50 35 AM


comp/core/secrets/secretsimpl/status.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	stats["executable"] = r.backendCommand
+
+	permissionMsg := "OK, the executable has the correct permissions"
+	err := checkRights(r.backendCommand, r.commandAllowGroupExec)
+	if err != nil {
+		permissionMsg = fmt.Sprintf("error: %s", err)
+	}
+	stats["executable_permissions"] = permissionMsg
```
hush-hush, line 51: This seems only used for JSON version, if so we should produce a more friendly JSON.

Something like this would be more user friendly:
```
{
    "executable_correct_permissions": true/false,
    "executable_permissions_message": "OK, the executable has the correct permissions",
}
```

hush-hush (changes requested):

hush-hush (approved):

comp/core/secrets/secretsimpl/status.go
rahulkaukuntla: I made the suggested change.

PR #33943, opened 2025-02-11
[usm] add regular and raw tracepoints /sched_process_exit

yuri-lipnesh:
### What does this PR do?
Adds eBPF programs for `tracepoint/sched/sched_process_exit` (kernel < 4.17) and
raw_tracepoint/sched_process_exit (kernel>=4.17)
These programs clean terminated processes from `ssl_read_args` and `ssl_read_ex_args` maps.

### Motivation
[...]

PR #35314, opened 2025-03-20
[EBPF] attacher: add OnSyncCallback

gjulianm:
### What does this PR do?

Adds the `OnSyncCallback` option to the uprobe attacher, to allow callers to periodically run a custom function that requires to know the currently alive PIDs.

### Motivation

Replicating in the attacher the behavior introduced in https://github.com/DataDog/datadog-agent/pull/33943/, to allow for the SSL monitor refactor

Unit tests passing


pkg/ebpf/uprobes/attacher.go
```diff
@@ -219,6 +219,10 @@ type AttacherConfig struct {
 
 	// If shared libraries tracing is enabled, this is the name of the library set used to filter the events
 	SharedLibsLibset sharedlibraries.Libset
+
+	// OnSyncCallback is an optional function that gets called when the attacher performs a sync. Receives as an argument
+	// the set of alive PIDs in the system.
+	OnSyncCallback func(map[uint32]struct{})
```
val06, line 225: are there any limitations on the implementation? must it be fast? can it lead to potential deadlocks?

gjulianm: If it takes too long it can delay other syncs, but the usual sync interval is in the order of seconds (it's user configurable)

val06: Consider adding that to the comment in code.

gjulianm: Added!

val06 (approved):

PR #35315, opened 2025-03-20
fix oci rollback in action

clarkb7:
### What does this PR do?
Fix the MSI action sequence of the OCI install rollback action to ensure it runs when the OCI install action fails.
Previously it only ran if a later action failed.


test/new-e2e/tests/installer/windows/suites/apm-library-dotnet-package/install_no_iis_test.go
```diff
@@ -40,3 +45,21 @@ func (s *testDotnetLibraryInstallSuiteWithoutIIS) TestInstallDotnetLibraryPackag
...
+		installerwindows.WithMSIArg("DD_APM_INSTRUMENTATION_ENABLED=iis"),
+		// TODO: remove override once image is published in prod
+		// TODO: support DD_INSTALLER_REGISTRY_URL
+		installerwindows.WithMSIArg("SITE=datad0g.com"),
+		installerwindows.WithMSIArg(fmt.Sprintf("DD_APM_INSTRUMENTATION_LIBRARIES=dotnet:%s", version)),
+		installerwindows.WithMSILogFile("install-rollback.log"),
+		installerwindows.WithMSIDevEnvOverrides("CURRENT_AGENT"),
+	))
+	defer s.Installer().Purge()
+
```
bmermet, line 61: Is there a way to check that the rollback runs in the MSI logs for instance?  Because if we fix the [behavior causing the todo above](https://github.com/DataDog/datadog-agent/pull/35315/files#diff-c30d36cd1d8f63231112482274f1ede2e3ed30429899f2cd64e669063872d432R43-R46), this will pass even if the rollback does not run.

jack0x2 (approved):

L3n41c (approved):
👍🏻 for file owned by @DataDog/container-integrations.

test/new-e2e/tests/installer/windows/suites/apm-library-dotnet-package/install_no_iis_test.go
clarkb7: hmm interesting point. We should be able to do that. I created a card for it so that we can get this merged as is for 7.65.0
https://datadoghq.atlassian.net/browse/WINA-1347

PR #30543, opened 2024-10-28
enhancement: add 'remote agent' component for remotely-provided status and flare support

tobz:
### What does this PR do?

This PR introduces a new "remote agent" component, which is used to provide `agent status` and `agent flare` capabilities for remote processes (agents) by communicating over gRPC.

The component is added to the normal Agent stack, and waits for remote agents to register themselves, providing some basic information like a unique ID, a display name (i.e., `Process Agent`) and an API endpoint address. When the new component is queried for status or flare information, it concurrently reaches out to all registered remote agents, collects the relevant data, and then hands it back.

This effectively allows any remote agent to coordinate with the Core Agent by registering itself and integrating neatly with the existing `agent status`/`agent flare` infrastructure, supporting use cases like Agent Data Plane, Agent Kernel, JMXFetch, embedded OpenTelemetry Collector, and more.

### Motivation

This is needed for Agent Data Plane but is generically applicable to many other use cases as well.

### Describe how to test/QA your changes

Bundled in this PR is a one-off "remote agent" implementation that can be used to simulate a remote agent.

```shell
# Build the Agent and the `remote-agent` tool:
inv agent.build
inv agent.build-remote-agent

# In another shell, run the Agent:
bin/agent/agent -c bin/agent/dist run

# Check the Agent's status output, and observe that the `Remote Agents` section
# towards the end indicates "No remote agents registered":
bin/agent/agent -c bin/agent/dist status

# Run the `remote-agent` binary, pretending to be a remote agent:
#
# This will emit some basic information that the remote agent has been registered,
# and there will be a corresponding log line on the Agent side saying the same:
bin/remote-agent -agent-id remote-agent-1 -display-name="Remote Agent #1" \
                 -listen-addr 127.0.0.1:13580 \
                 -agent-ipc-address localhost:5001 \
                 -agent-auth-token-file bin/agent/dist/auth_token

# Check the Agent's status again, and observe that there is a new entry under the
# `Remote Agents` section for our remote agent:
bin/agent/agent -c bin/agent/dist status

# Now build a flare (non-local, since this depends on runtime state) and observe that
# the resulting flare archive contains a file at `<hostname>/remote-agent-1/example.txt`
# with the contents of `Hello, world!`.
#
# You'll need to enter an e-mail address (doesn't need to be real) and when asked if the
# archive should be uploaded, you can say no:
bin/agent/agent -c bin/agent/dist flare
```

### Additional Notes

- [x] Add remote agent example for QA purposes
- [x] Add unit tests
[...]

PR #35319, opened 2025-03-20
[AGTMETRICS-294] Add RemoteAgent telemetry request

rayz:
### What does this PR do?
Adds the grpc definitions `GetTelemetryRequest` and `GetTelemetryResponse` to `RemoteAgent` in order to get telemetry data from remote agents.

The only supported format for the telemetry response so far is the prometheus text exposition format.

### Motivation
This will be used to get ADP's internal telemetry.

```
# Build the Agent and the `remote-agent` tool:
inv agent.build
inv agent.build-remote-agent

# In another shell, run the Agent:
bin/agent/agent -c bin/agent/dist run

# Run the `remote-agent` binary, pretending to be a remote agent:
#
# This will emit some basic information that the remote agent has been registered,
# and there will be a corresponding log line on the Agent side saying the same:
bin/remote-agent -agent-id remote-agent-1 -display-name="Remote Agent #1" \
                 -listen-addr 127.0.0.1:13580 \
                 -agent-ipc-address localhost:5001 \
                 -agent-auth-token-file bin/agent/dist/auth_token

# Go to `http://localhost:5000/telemetry` 
# See that the telemetry is working as the same before and has the addition of 
# remote_agent_test_telemetry emitted from the dummy remote-agent.
```


comp/core/remoteagentregistry/impl/remoteagentregistry.go
```diff
@@ -178,6 +187,77 @@ func (ra *remoteAgentRegistry) start() {
...
+
+	for agentID, details := range c.agentMap {
+		go func() {
+			// Retrieve the telemetry data in exposition format from the remote agent
+			resp, err := details.client.GetTelemetry(ctx, &pb.GetTelemetryRequest{}, grpc.WaitForReady(true))
+			if err != nil {
+				log.Warnf("Failed to query remote agent '%s' for telemetry data: %v", agentID, err)
+				return
+			}
+			// TODO: parse the response and add the metrics to the channel
```
rayz, line 235: The idea is to retrieve the text response from the grpc call and parsing it with prometheus `expfmt`.

pkg/proto/datadog/remoteagent/remoteagent.proto
```diff
@@ -72,3 +72,12 @@ message GetFlareFilesResponse {
...
   map<string, bytes> files = 1;
 }
+
+message GetTelemetryRequest {}
+
+message GetTelemetryResponse {
+  // Text based metrics in Prometheus exposition format.
+  //
+  // This needs to be parsed into the appropriate metrics.
+  string telemetry = 1;
```
tobz, line 82: I think we should make this a `oneof` and then call this variant `prom_text`, just to give ourselves a little more freedom to incrementally update things in the future. For example, Prometheus "native" histograms can't be represented in the text exposition format, so if we ever wanted to switch to supporting them, we'd have to either update this field to have dual meaning (and update the related parsing code) or add a new field just for the additional format, and still update the parsing code to decide which field has priority or not, etc.

By making it a `oneof`, we come out of the gate enforcing that there can be multiple potential formats, but that only one can ever be present in the payload.

```
message GetTelemetryResponse {
  // Telemetry payload.
  oneof payload {
    // Prometheus "text" exposition format.
    string prom_text = 1;

    // In the future, we then might add:
    // bytes prom_protobuf = 2;
    //
    // or potentially utilize the native Protocol Buffers definition:
    // io.prometheus.write.v2.Request prom_protobuf = 2;
    //
    // and so on...
  }
}
```

pkg/proto/datadog/remoteagent/remoteagent.proto
```diff
@@ -72,3 +72,21 @@ message GetFlareFilesResponse {
...
+    // Prometheus "text" exposition format.
+    string prom_text = 1;
+
+    // In the future, we then might add:
+    // bytes prom_protobuf = 2;
+    //
+    // or potentially utilize the native Protocol Buffers definition:
+    // io.prometheus.write.v2.Request prom_protobuf = 2;
+    //
+    // and so on...
```
tobz, line 90: This example code I gave was purely for demonstrating my thought process in the PR feedback, but I didn't intend it to be added as-is... so I would remove this whole doc comment block.

tobz (changes requested):

comp/core/remoteagentregistry/impl/remoteagentregistry.go
```diff
@@ -83,10 +92,12 @@ func newRemoteAgent(reqs Requires) *remoteAgentRegistry {
...
-	conf         config.Component
-	agentMap     map[string]*remoteAgentDetails
-	agentMapMu   sync.Mutex
-	shutdownChan chan struct{}
+	conf              config.Component
+	agentMap          map[string]*remoteAgentDetails
+	agentMapMu        sync.Mutex
+	shutdownChan      chan struct{}
+	telemetry         telemetry.Component
+	RegistryCollector prometheus.Collector
```
tobz, line 100: Is this actually getting used? 🤔

comp/core/remoteagentregistry/impl/remoteagentregistry.go
```diff
@@ -142,9 +153,14 @@ func (ra *remoteAgentRegistry) RegisterRemoteAgent(registration *remoteagentregi
 	return recommendedRefreshInterval, nil
 }
 
+func (ra *remoteAgentRegistry) RegisterCollector() {
```
tobz, line 156: I forgot the visibility rules for Go methods, but are we able to make this private and still be able to call it from the test?

It strikes me that we shouldn't allow this be be called directly from anything outside the implementation.

comp/core/remoteagentregistry/impl/remoteagentregistry.go
rayz: Removed 👍

comp/core/remoteagentregistry/impl/remoteagentregistry.go
rayz: Updated the test to call `start` instead 👍

tobz (approved):

pkg/proto/datadog/remoteagent/remoteagent.proto
```diff
@@ -72,3 +72,15 @@ message GetFlareFilesResponse {
...
+
+message GetTelemetryRequest {}
+
+message GetTelemetryResponse {
+  // Telemetry payload.
+  oneof payload {
+    // Prometheus "text" exposition format.
+    //
+    // https://prometheus.io/docs/instrumenting/exposition_formats/#text-based-format
+    string prom_text = 1;
```
jeremy-hanna, line 84: I'm not very familiar with the prometheus text exposition format but from my observation with this polling approach we're expecting several metrics bundled together in this response - do we have expectations/concerns around message length? I see the [this grpc client has configured it](https://github.com/DataDog/datadog-agent/pull/30543/files#diff-62826caa101a7872ffae0270f7e77f55abf48945d1ca5abde6429b94a0c183e6R358) higher than the 4MB default for flares and that is probably enough with the polling interval, it might be worth making that limit configurable as we [do the grpc server](https://github.com/DataDog/datadog-agent/blob/main/comp/api/grpcserver/impl-agent/grpc.go#L87-L88) to have a knob to turn there as we build this functionality out

jeremy-hanna (approved):
Nice work! LGTM 🔥

iglendd (approved):

comp/core/remoteagentregistry/impl/remoteagentregistry.go
```diff
@@ -178,6 +193,118 @@ func (ra *remoteAgentRegistry) start() {
...
+		c.agentMapMu.Unlock()
+		return
+	}
+
+	data := make(chan *pb.GetTelemetryResponse, agentsLen)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	for agentID, details := range c.agentMap {
```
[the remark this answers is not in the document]
rayz: I think it will be more in the dozens or lower. Some potential candidates for remote agents: agent kernal, jmxfetch ... but as of now the only remote agent in use is ADP.

PR #35330, opened 2025-03-20
retry network ID fetch in netpath

akarpz:
### What does this PR do?

we already have similar retry logic in the network tracer, but netpath can benefit too because it often fails to fetch a network ID. also it was only checking on EC2, but it should use the helper logic to get it from user provided config or GCE


pkg/networkpath/traceroute/runner/runner.go
```diff
@@ -282,3 +277,25 @@ func createGatewayLookup(telemetryComp telemetryComponent.Component) (network.Ga
...
+		if err == nil {
+			return networkID, nil
+		}
+		log.Debugf(
+			"failed to fetch network ID (attempt %d/%d): %s",
+			attempt,
+			maxRetries,
+			err,
+		)
+		time.Sleep(time.Duration(250*attempt) * time.Millisecond)
```
pimlu, line 298: Nit: Doesn't need to sleep after the final try

pimlu (approved):

PR #35341, opened 2025-03-21
[CWS] Add test on empty scope

lebauce:
### What does this PR do?

Adds a test for scope that failed to be resolved such as process context.

### Motivation

There was a regression regarding this corner case so we want to make
sure we do not forget about this case.


paulcacheux (approved):
Thank you !

PR #35343, opened 2025-03-21
[CWS] Embed multiple default policies on windows

mftoure:
### What does this PR do?
This PR enables embedding multiple default policies on windows, not just the default.policy file.

### DIFF
--- a/omnibus/config/software/datadog-agent-finalize.rb
+++ b/omnibus/config/software/datadog-agent-finalize.rb
@@ -39,7 +39,6 @@ build do
             if ENV['WINDOWS_DDPROCMON_DRIVER'] and not ENV['WINDOWS_DDPROCMON_DRIVER'].empty? and not windows_arch_i386?
               move "#{install_dir}/etc/datadog-agent/security-agent.yaml.example", conf_dir_root, :force=>true
               move "#{install_dir}/etc/datadog-agent/runtime-security.d", conf_dir_root, :force=>true
-              move "#{conf_dir_root}/runtime-security.d/default.policy", "#{conf_dir_root}/runtime-security.d/default.policy.example", :force=>true
             end
             if ENV['WINDOWS_APMINJECT_MODULE'] and not ENV['WINDOWS_APMINJECT_MODULE'].empty?
               move "#{install_dir}/etc/datadog-agent/apm-inject.yaml.example", conf_dir_root, :force=>true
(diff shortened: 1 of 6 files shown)


lebauce (approved):

omnibus/config/software/datadog-agent-finalize.rb
```diff
@@ -39,7 +39,6 @@
             if ENV['WINDOWS_DDPROCMON_DRIVER'] and not ENV['WINDOWS_DDPROCMON_DRIVER'].empty? and not windows_arch_i386?
               move "#{install_dir}/etc/datadog-agent/security-agent.yaml.example", conf_dir_root, :force=>true
               move "#{install_dir}/etc/datadog-agent/runtime-security.d", conf_dir_root, :force=>true
-              move "#{conf_dir_root}/runtime-security.d/default.policy", "#{conf_dir_root}/runtime-security.d/default.policy.example", :force=>true
```
chouquette, line 42: Don't we risk overriding customers' `default.policy` file if they contain modifications?

I'm not familiar at all with these files, so please forgive the question if it's silly

mftoure: Customers are expected to create additional policy files if they want to create their rules.
This behaviour (risk to override the default.policy ) was the same before this PR.

clarkb7: Previously the MSI custom action would only copy `default.policy.example` to `default.policy` if it does not already exist.

tools/windows/DatadogAgentInstaller/WixSetup/Datadog Agent/AgentInstaller.cs
```diff
@@ -610,9 +610,10 @@ private Dir CreateAppDataFolder()
                 ));
 
             appData.AddDir(new Dir(new Id("security.d"),
-                                    "runtime-security.d",
-                                    new WixSharp.File($@"{EtcSource}\runtime-security.d\default.policy.example")
+                       "runtime-security.d",
+                       new Files($@"{EtcSource}\runtime-security.d\*.*")
```
clarkb7, line 614: These files will be deleted at uninstall time. and during upgrades they will be deleted and then written by the new version.
Is this the expected behavior?

clarkb7: isn't it expected to be updated remotely once the Agent starts?

mftoure: There are multiple providers for policies. What you are talking about is the remote config provider. The changes here are just for the file provider. (the user puts a .policy file in the expected policy, the agent loads it)

chouquette (approved):
LGTM for agent-delivery file

clarkb7 (approved):

PR #35351, opened 2025-03-21
AGNTLOG-2 - reduce payload memory usage in the logs pipeline

ddrthall:
### What does this PR do?
The logs payload entity stores metadata for all of the respective messages batched within, but also stores the message contents themselves. This isn't required, all consumers of the logs payload will interact with the encoded bytes buffer rather than the raw contents. Additionally, dropping these message contents on payload construction can save considerable amounts of RSS memory for log agents experiencing enough load to buffer multiple payloads. 

This PR shallow copies the message metadata into each payload, rather than injecting the whole message entity.

Validation was completed by a combination of 1. unit test additions, 2. successful code compilation, 3. SMP performance runs

Adding memory allocation in the critical path (for the metadata shallow copy) isn't ideal, but given we're already allocating a new array for the entirety of the contents in the same code path its impact shouldn't be noticeable.

### Additional Notes

The memory savings for a system under heavy load is extreme. Attached is the memory comparison used in the file_to_blackhole_1000ms_latency_linear_load regression test:
Screenshot 2025-03-21 at 11 25 17 AM

### DIFF
--- a/pkg/logs/message/message.go
+++ b/pkg/logs/message/message.go
@@ -33,8 +33,8 @@ var EscapedLineFeed = []byte(`\n`)
 
 // Payload represents an encoded collection of messages ready to be sent to the intake
 type Payload struct {
-	// The slice of sources messages encoded in the payload
-	Messages []*Message
+	// The slice of sources message metadata encoded in the payload
+	MessageMetas []*MessageMetadata
 	// The encoded bytes to be sent to the intake (sometimes compressed)
 	Encoded []byte
 	// The content encoding. A header for HTTP, empty for TCP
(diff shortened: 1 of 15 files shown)


pkg/logs/message/message.go
```diff
@@ -43,15 +43,30 @@ type Payload struct {
 	UnencodedSize int
 }
 
+func NewPayload(messages []*Message, encoded []byte, encoding string, unencodedSize int) *Payload {
+	messageMetas := make([]*MessageMetadata, len(messages))
+	for i, m := range messages {
+		// Split the metadata from the message content to avoid holding the entire message in memory
```
gh123man, line 49: 👏 🥳

pkg/logs/message/message_test.go
```diff
@@ -21,5 +23,98 @@ func TestMessage(t *testing.T) {
...
+	var payload *Payload
+
+	// Create scope to allow message to be GC'd
+	func() {
+		// Create message with content we want to track
+		content := make([]byte, 1000000) // Large enough to be noticeable for GC
+		message := NewMessage(content, nil, "", 2)
+
+		// Set up finalizer to track when content is GC'd
+		runtime.SetFinalizer(&message.content, trackGC)
```
gh123man, line 90: This is a cool test. 
This seems like a nice way to assert memory behaviors in unit tests and possibly prevent memory regressions in the future 👍

hestonhoffman (approved):

gh123man (approved):

PR #35365, opened 2025-03-21
Downgrade to net462

andrewlock:
### What does this PR do?

Downgrades the .NET Framework version required by the installer from 4.7.2 to 4.6.2.

### Motivation

Windows Server 2016 ships with .NET Framework 4.6.2. While the 4.7.2 installer will run on this older framework, it's strongly discouraged, as if you reference anything not available in the earlier version, you will crash and fail installation.

Recent changes implicitly added a new dependency by using the `ValueTuple` type which is used under the hood when you use the C# syntax:

```csharp
         // 👇 Converted to value tuple
private (string Name, string Version) ParseVersion(string library)
{
}
```

This change was only introduced recently, hence why the problem has not previously appeared. 

The existing tests should validate that the installer still works as expected. The fact that this was caught in manual testing shows that there is a test missing here, but automating tests for Windows Server 2016 could be challenging (I will defer to the agent team if they deem them to be required)

Targeting a lower TFM means you have fewer features available, but that is an unfortunate consequence of supporting Windows Server 2016.

### Additional Notes

It's possible that _only_ removing the `ValueTuple` code would also resolve the issue, but without downgrading the project TFMs there's a risk that a similar change will be introduced again, especially as validating on old TFMs is generally tricky (this was only caught during manual testing).

As a general rule, compiling for earlier TFMs and running on later versions is safe, whereas the reverse is not safe.


tools/windows/DatadogAgentInstaller/CustomActions/InstallOciPackages.cs
```diff
@@ -173,5 +173,17 @@ public static ActionResult RollbackActions(Session session)
         {
             return new InstallOciPackages(new SessionWrapper(session)).RollbackState();
         }
+
+        private class NameVersionPair
```
julien-lebot, line 177: Why don't we use regular Tuples? https://learn.microsoft.com/en-us/dotnet/api/system.tuple?view=netframework-4.6.2

andrewlock: We could, I just think they're ugly with `Item1` and `Item2` 😅 Knowing what they mean elsewhere in the code base is harder than just using a simple class

julien-lebot: Fair enough!

julien-lebot (approved):

PR #35378, opened 2025-03-21
Fix deadlock and stability issues in Go DI

grantseltzer:
### What does this PR do?

This fixes a deadlock, a couple of other bugs, and updates tests. The bugs are:

- A deadlock occurs when probe configuration is updated. A mutex is shared to guard 3 different maps. Updating configuration doesn't work as a result.
    - This was fixed by using sync.Map instead.
- Slices were not properly being converted to captured values before being uploaded to datadog backend.
- A more difficult to understand bug occurs in bpf space with the param stack maps. Instrumentation would occasionally not work at all until I did a full rebuild. Checking kernel logs I would see a lot of these errors:

```
sample_service-2611190 [002] ....1 7050153.972650: bpf_trace_printk: github_com_DataDog_datadog_agent_pkg_dynamicinstrumentation_testutil_sample_test_struct probe in go-di-sample-service has triggered
  sample_service-2611190 [002] ....1 7050153.972711: bpf_trace_printk: there is no `param_stack` map associated with currently executing cpu 2
```

I fixed this by changing to the idiomatic way of populating the bpf array of maps in user space.

- Updates test fixtures in accordance with my most recent PR.

### Motivation

Ensure stability.

Ran e2e tests.

### DIFF
--- a/pkg/dynamicinstrumentation/codegen/c/maps.h
+++ b/pkg/dynamicinstrumentation/codegen/c/maps.h
@@ -25,10 +25,7 @@ BPF_PERCPU_ARRAY_MAP(temp_storage_array, __u64[4000], 1);
 // when reading the values in the collection.
 BPF_HASH_MAP(collection_limits, char[6], __u16, 1024);
 
-// The param_stack_cpu_<i> map is used as a stack for the location expressions
-// to operate on values and addresses. We use templating to create per-cpu map
-{{- range $i, $val := until .InstrumentationInfo.InstrumentationOptions.NumCPUs }}
-BPF_STACK_MAP(param_stack_cpu_{{$i}}, __u64, 2048);
-{{- end }}
-
+// The param_stack map is used as a stack for the location expressions
+// to operate on values and addresses.
+BPF_STACK_MAP(param_stack, __u64, 2048);
 #endif
--- a/pkg/dynamicinstrumentation/codegen/c/maps.h
+++ b/pkg/dynamicinstrumentation/codegen/c/maps.h
@@ -22,7 +22,7 @@ struct inner_param_stack {
 // The param_stacks map is to set up a unique stack for each CPU.
 struct {
     __uint(type, BPF_MAP_TYPE_ARRAY_OF_MAPS);
-    __uint(max_entries, 32); // we arbitrarily set this to support 32 CPUs
+    __uint(max_entries, 0);
     __uint(key_size, sizeof(__u32));
     __array(values, struct inner_param_stack);
 } param_stacks SEC(".maps");
--- a/pkg/dynamicinstrumentation/codegen/c/maps.h
+++ b/pkg/dynamicinstrumentation/codegen/c/maps.h
@@ -10,6 +10,23 @@ struct {
     __uint(max_entries, 1 << 24);
 } events SEC(".maps");
 
+
+// The param_stack map is used as a stack for the location expressions
+// to operate on values and addresses.
+struct inner_param_stack {
+    __uint(type, BPF_MAP_TYPE_STACK);
+    __uint(max_entries, 2048);
+    __uint(value_size, sizeof(__u64));
+};
+
+// The param_stacks map is to set up a unique stack for each CPU.
+struct {
+    __uint(type, BPF_MAP_TYPE_ARRAY_OF_MAPS);
+    __uint(max_entries, 32); // we arbitrarily set this to support 32 CPUs
+    __uint(key_size, sizeof(__u32));
+    __array(values, struct inner_param_stack);
+} param_stacks SEC(".maps");
+
 // The zeroval map is used to have pre-zero'd data which bpf code can
 // use to zero out event buffers (similar to memset, but verifier friendly).
 BPF_ARRAY_MAP(zeroval, char[PARAM_BUFFER_SIZE], 1);
@@ -24,8 +41,4 @@ BPF_PERCPU_ARRAY_MAP(temp_storage_array, __u64[4000], 1);
 // of collections such as slices so that they can later be referenced
 // when reading the values in the collection.
 BPF_HASH_MAP(collection_limits, char[6], __u16, 1024);
-
-// The param_stack map is used as a stack for the location expressions
-// to operate on values and addresses.
-BPF_STACK_MAP(param_stack, __u64, 2048);
 #endif
--- a/pkg/dynamicinstrumentation/ditypes/config.go
+++ b/pkg/dynamicinstrumentation/ditypes/config.go
@@ -14,7 +14,6 @@ import (
 	"io"
 	"strconv"
 	"strings"
-	"sync"
 
 	"github.com/DataDog/datadog-agent/pkg/util/log"
 
@@ -139,13 +138,10 @@ type ProcessInfo struct {
 	ProbesByID             ProbesByID
 	InstrumentationUprobes map[ProbeID]*link.Link
 	InstrumentationObjects map[ProbeID]*ebpf.Collection
-	mu                     sync.RWMutex
 }
 
 // SetupConfigUprobe sets the configuration probe for the process
 func (pi *ProcessInfo) SetupConfigUprobe() (*ebpf.Map, error) {
-	pi.mu.Lock()
-	defer pi.mu.Unlock()
 
 	configProbe, ok := pi.ProbesByID[ConfigBPFProbeID]
 	if !ok {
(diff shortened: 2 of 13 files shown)


pkg/dynamicinstrumentation/ditypes/config.go
```diff
@@ -136,30 +239,39 @@ type ProcessInfo struct {
...
+	if pi.InstrumentationUprobes == nil {
+		return nil, fmt.Errorf("uprobes map not initialized for process %s", pi.ServiceName)
+	}
+	configLink := pi.InstrumentationUprobes.Get(ConfigBPFProbeID)
+	if configLink == nil {
 		return nil, fmt.Errorf("config uprobe was not set for process %s", pi.ServiceName)
 	}
 	pi.ConfigurationUprobe = configLink
-	delete(pi.InstrumentationUprobes, ConfigBPFProbeID)
+	pi.InstrumentationUprobes.Delete(ConfigBPFProbeID)
```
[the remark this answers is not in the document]
grantseltzer: Hm ok, stepping through this right now to better understand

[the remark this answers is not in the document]
grantseltzer: Ok this makes sense now. If two concurrent calls occur the link from `InstrumentationUprobesMap` could be deleted by one thread before being retrieved by the other threads. Even though this wouldn't occur because it's only called a single time per process which have unique config probe links, but still. 

I'll work on using a RWMutex. Is it still needed to use a sync map?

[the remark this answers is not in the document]
grantseltzer: Also SetupConfigUprobe should be renamed, its purpose is to remove the uprobe link from the map so it doesn't get instrumented elsewhere in the code (the maps are used elsewhere)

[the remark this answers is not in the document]
grantseltzer: Actually after further thought, I'm not convinced there is actually a problematic race condition. Even if two threads call this function one after the other, the same thing would happen where one thread wouldn't be able to retrieve the config uprobe link. It's deleted completely, not reassigned. 

What could possibly make sense is to use `sync.Once` here.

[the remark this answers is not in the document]
tylfin: Yeah, I wouldn't overthink it, perhaps just revert to the mutex approach, switch to an RWLock and add it to the missing usages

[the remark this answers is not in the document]
grantseltzer: Ok yea that's fine by me

pkg/dynamicinstrumentation/ebpf/ebpf.go
```diff
@@ -54,7 +53,43 @@ func AttachBPFUprobe(procInfo *ditypes.ProcessInfo, probe *ditypes.Probe) error
...
 	}
 
-	mapReplacements := map[string]*ebpf.Map{}
+	inner := &ebpf.MapSpec{
+		Type:       ebpf.Stack,
+		MaxEntries: 2048,
+		ValueSize:  8,
+	}
+
+	outer := &ebpf.MapSpec{
```
brycekahle, line 62: This will create a new map, rather than using the one from the object file? Is that intended? I see you are using `MapReplacements`, but I don't think that is required.

You do need to setup the inner maps, but you can get the outer map with `outer := spec.Maps["param_stacks"]`

pkg/dynamicinstrumentation/ebpf/ebpf.go
```diff
@@ -54,7 +53,43 @@ func AttachBPFUprobe(procInfo *ditypes.ProcessInfo, probe *ditypes.Probe) error
...
+		Name:       "param_stacks",
+		InnerMap:   inner,
+		Type:       ebpf.ArrayOfMaps,
+		MaxEntries: uint32(32), // we arbitrarily set this to support 32 CPUs
```
brycekahle, line 66: You can get the exact value of available CPUs easily:
```
numCPUs, err := kernel.PossibleCPUs()
if err != nil {
	return nil, fmt.Errorf("error getting possible cpus: %s", err)
}
outer.MaxEntries = numCPUs
```

pkg/dynamicinstrumentation/codegen/c/maps.h
```diff
@@ -10,6 +10,23 @@ struct {
...
+struct inner_param_stack {
+    __uint(type, BPF_MAP_TYPE_STACK);
+    __uint(max_entries, 2048);
+    __uint(value_size, sizeof(__u64));
+};
+
+// The param_stacks map is to set up a unique stack for each CPU.
+struct {
+    __uint(type, BPF_MAP_TYPE_ARRAY_OF_MAPS);
+    __uint(max_entries, 32); // we arbitrarily set this to support 32 CPUs
```
brycekahle, line 25: I'd suggest setting this to `0` to ensure that userspace is properly setting it.

pkg/dynamicinstrumentation/ebpf/ebpf.go
brycekahle: Another reason to do this, 32 cpus is not sufficient for even our internal environments. I think we have some nodes with 96 cpus.

pkg/dynamicinstrumentation/codegen/c/maps.h
grantseltzer: Ah, so if I set to 0 we can override it in userspace?

brycekahle: You can always override in userspace, but setting to 0 means it will fail to load if you fail set a value in userspace

pkg/dynamicinstrumentation/ebpf/ebpf.go
grantseltzer: It was my intention, but am I able to resize the outer map based on the number of CPUs at runtime?

brycekahle: Maps are always created by the loader. By modifying the spec here, you are changing what the loader does before creation. The reason to use `MapReplacements` is if you need to share an existing map with a program. This is why it is used for the ring buffer.

pkg/dynamicinstrumentation/codegen/c/maps.h
grantseltzer: Yea making this change produces this error (why I originally set it to an arbitrary number):

```
2025-03-24 15:43:54 CDT | SYS-PROBE | INFO | (pkg/dynamicinstrumentation/diconfig/config_manager.go:287 in applyConfigUpdate) | Couldn't load and attach bpf programs could not load bpf collection for probe g504163d-f367-4522-8905-fe8bc34eb975: using replacement map param_stacks: MaxEntries: 4 changed to 0: map spec is incompatible with existing map
```

grantseltzer: I suppose if we can fully make the outer map in userspace and retrieve it in bpf space then we can set length in Go safely. Is there a way to do that without having a map (with 1 entry) holding this outer map, which holds the stack maps?

brycekahle: maps are always created in userspace. The loader does this (simplified):
1. read object file and collect map specs
1. create maps for each spec with the `bpf` syscall, kernel gives back a `fd`
1. modify ebpf bytecode of programs to reference the `fd`
1. load ebpf programs into kernel

By changing the spec in `spec.Maps["param_stacks"]`, it is the same as if it was changed in the object file

brycekahle: > `map spec is incompatible with existing map`

this is because `MapReplacements` is trying to ensure `spec.Maps["param_stacks"]` matches what you pass to `MapReplacements["param_stacks"]`. This is so the ebpf code which is expecting a specific map spec continues to work.

You don't need `MapReplacements` here at all for `param_stacks`.

grantseltzer: If that's the case, why would I be getting the above pasted error?

grantseltzer: Ah I see then, So I can edit the spec of the outer map before loading. This was because I was making a new map.

brycekahle: Yes, exactly.

pkg/dynamicinstrumentation/ebpf/ebpf.go
```diff
@@ -53,42 +54,29 @@ func AttachBPFUprobe(procInfo *ditypes.ProcessInfo, probe *ditypes.Probe) error
 		return fmt.Errorf("could not create bpf collection for probe %s: %w", probe.ID, err)
 	}
 
+	outerMapSpec := spec.Maps["param_stacks"]
+	outerMapSpec.MaxEntries = uint32(runtime.NumCPU())
```
brycekahle, line 58: Please use `kernel.PossibleCPUs()`, `runtime.NumCPU()` is not guaranteed to match the kernel.

tylfin (approved):

PR #35380, opened 2025-03-21
[Bug fix] Add shared compression kind for different pipeline

DDuongNguyen:
### What does this PR do?
This PR fixes an issue where event platform pipelines don't properly inherit the compression settings from `logs_config.compression_kind`. Previously, pipelines with specialized prefixes (eg `service_discovery.forwarder`, `agent_telemetry`, etc.) would default to using `DefaultLogCompressionKind `(gzip) even when `logs_config.compression_kind` was set to `zstd`. This created inconsistent compression across different pipelines.
The fix modifies the `compressionKind()` method to first check the pipeline-specific setting, then fall back to the global `logs_config.compression_kind` setting, and only if both are missing use the hardcoded default.

### Motivation
This change ensures consistent compression behavior across all agent pipelines. When a user configures logs_config.compression_kind: zstd, all pipelines should respect that setting instead of silently using different compression algorithms for different data types. This was causing confusion and poor performance due to inconsistent compression settings between data pipelines.

1. Turn on agent debug log
2. Set logs_config.compression_kind to zstd or something that is different from the the defaultLogCompressionKind (currently gzip)
3. Turn on the agent and check for this debug log:
```
log.Debugf("Using shared pipeline compression kind: %s for %s", sharedPipelineCompressionKind, l.prefix)
```

and see if its correctly using the `logs_config.compression_kind` instead of the `DefaultLogCompressionKind`

### DIFF
--- a/comp/logs/agent/config/config_keys.go
+++ b/comp/logs/agent/config/config_keys.go
@@ -109,16 +109,28 @@ func (l *LogsConfigKeys) devModeUseProto() bool {
 	return l.getConfig().GetBool(l.getConfigKey("dev_mode_use_proto"))
 }
 
+// compressionKind returns the compression kind for the logs agent
+// However, other pipelines (agent_telemetry, databse_monitoring,network_device, container,sbom, service_discovery, compliance, runtime_security) also use this function.
+// Todo: Refactor this to eventually account for different config/global keys for different pipelines other than the logs pipeline.
 func (l *LogsConfigKeys) compressionKind() string {
-	compressionKind := l.getConfig().GetString(l.getConfigKey("compression_kind"))
-	switch compressionKind {
-	case "zstd", "gzip":
+	configKey := l.getConfigKey("compression_kind")
+	compressionKind := l.getConfig().GetString(configKey)
+
+	if l.prefix != "logs_config." {
+		sharedPipelineCompressionKind := l.getConfig().GetString("logs_config.compression_kind")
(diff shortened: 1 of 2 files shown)


git-thuerk-done (approved):

comp/logs/agent/config/config_keys.go
```diff
@@ -109,16 +109,28 @@ func (l *LogsConfigKeys) devModeUseProto() bool {
...
+// However, other pipelines (agent_telemetry, databse_monitoring,network_device, container,sbom, service_discovery, compliance, runtime_security) also use this function.
+// Todo: Refactor this to eventually account for different config/global keys for different pipelines other than the logs pipeline.
 func (l *LogsConfigKeys) compressionKind() string {
-	compressionKind := l.getConfig().GetString(l.getConfigKey("compression_kind"))
-	switch compressionKind {
-	case "zstd", "gzip":
+	configKey := l.getConfigKey("compression_kind")
+	compressionKind := l.getConfig().GetString(configKey)
+
+	if l.prefix != "logs_config." {
```
paulcacheux, line 119: (I'm working on runtime security, so I'm impacted by this PR). I don't understand it at all. If I specifically ask for `zstd` in my prefix specific config, it's not going to be actually used ?
Why not using some `IsConfigured` or similar to inherit when it's not defined in the config ?
Also why should this be done just for the compression and not for other configs: like NoSSL, or the compression level ?

paulcacheux (changes requested):

PR #35388, opened 2025-03-24
[EBPF] gpu: fix initializing of NVML library for host tags

gjulianm:
### What does this PR do?

This PR fixes the initialization of the NVML library for getting host tags.

### Motivation

Fix a crash in non-NVIDIA hosts. See also #incident-36466 

Unit tests added to cover this case. Validated in `gizmo` staging cluster too.


pkg/gpu/tags/tags_nix.go
```diff
@@ -9,23 +9,43 @@
...
-// GetTags returns gpu_host:true if any NVIDIA GPUs are present, nil otherwise
-func GetTags() []string {
-	if nvmlLibrary == nil {
-		nvmlLibrary = nvml.New()
-		if ret := nvmlLibrary.Init(); ret != nvml.SUCCESS && ret != nvml.ERROR_ALREADY_INITIALIZED {
-			log.Warnf("Failed to get gpu host tags, failed to initialize NVML: %v", nvml.ErrorString(ret))
-			return nil
-		}
+func ensureNvmlLibrary(getLibrary func(opts ...nvml.LibraryOption) nvml.Interface) error {
+	mtx.Lock()
```
val06, line 25: nit: add **another** check for nil before acquiring the lock

val06 (approved):
small comment

val06 (approved):

pkg/gpu/tags/tags_nix.go
brycekahle: Isn't reading before acquiring the mutex creating a data race?

brycekahle: > A data race is defined as a write to a memory location happening concurrently with another read or write to that same location, unless all the accesses involved are atomic data accesses as provided by the sync/atomic package.

gjulianm: In this case not, the variable is only set to a value once all the initialization is complete and nothing more needs to be done. Setting the pointer is an atomic operation in arm/x86 according to the Go spec so we won't have partial values either.

brycekahle: > the variable is only set to a value once all the initialization is complete and nothing more needs to be done

This feels brittle and implicit. Someone could accidentally change the code to make this not true, and then this would break. Can we use a `RWMutex` instead if you are trying to speed up reads?

PR #35400, opened 2025-03-24
[EBPF] gpu: add NVML library singleton

gjulianm:
### What does this PR do?

This PR adds a `GetNvmlLib` function to have a single point where we initialize and retrieve the NVML library. It's in a separate package so that we can import it from other packages without bringing in the entire GPU package, which is system-probe specific.

### Motivation

Reduce changes of errors due to erroneous handling of NVML init.

Added unit tests

### DIFF
--- a/cmd/system-probe/config/ns.go
+++ b/cmd/system-probe/config/ns.go
@@ -73,6 +73,6 @@ func discoveryNS(k ...string) string {
 }
 
 // gpuNS adds `gpu_monitoring` namespace to config key
-func gpuNS(k ...string) string {
+func GpuNS(k ...string) string {
 	return NSkey("gpu_monitoring", k...)
 }
--- a/cmd/system-probe/config/ns.go
+++ b/cmd/system-probe/config/ns.go
@@ -73,6 +73,6 @@ func discoveryNS(k ...string) string {
 }
 
 // gpuNS adds `gpu_monitoring` namespace to config key
-func GpuNS(k ...string) string {
+func gpuNS(k ...string) string {
 	return NSkey("gpu_monitoring", k...)
 }
--- a/pkg/gpu/config/config.go
+++ b/pkg/gpu/config/config.go
@@ -30,8 +30,6 @@ type Config struct {
 	ScanProcessesInterval time.Duration
 	// InitialProcessSync indicates whether the probe should sync the process list on startup.
 	InitialProcessSync bool
-	// NVMLLibraryPath is the path of the native libnvidia-ml.so library
-	NVMLLibraryPath string
 	// ConfigureCgroupPerms indicates whether the probe should configure cgroup permissions for GPU monitoring
 	ConfigureCgroupPerms bool
 	// EnableFatbinParsing indicates whether the probe should enable fatbin parsing.
@@ -45,7 +43,6 @@ func New() *Config {
 		Config:                *ebpf.NewConfig(),
 		ScanProcessesInterval: time.Duration(spCfg.GetInt(sysconfig.FullKeyPath(GPUNS, "process_scan_interval_seconds"))) * time.Second,
 		InitialProcessSync:    spCfg.GetBool(sysconfig.FullKeyPath(GPUNS, "initial_process_sync")),
-		NVMLLibraryPath:       spCfg.GetString(sysconfig.FullKeyPath(GPUNS, "nvml_lib_path")),
 		Enabled:               spCfg.GetBool(sysconfig.FullKeyPath(GPUNS, "enabled")),
 		ConfigureCgroupPerms:  spCfg.GetBool(sysconfig.FullKeyPath(GPUNS, "configure_cgroup_perms")),
 		EnableFatbinParsing:   spCfg.GetBool(sysconfig.FullKeyPath(GPUNS, "enable_fatbin_parsing")),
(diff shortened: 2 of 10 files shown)


pkg/gpu/nvml/lib.go
```diff
@@ -0,0 +1,88 @@
...
+		var libpath string
+		if flavor.GetFlavor() == flavor.SystemProbe {
+			cfg := pkgconfigsetup.SystemProbe()
+			// Use the config directly here to avoid importing the entire gpu
+			// config package, which includes system-probe specific imports
+			libpath = cfg.GetString(sysconfig.GpuNS("nvml_lib_path"))
+		} else {
+			cfg := pkgconfigsetup.Datadog()
+			libpath = cfg.GetString("nvml_lib_path")
+		}
```
val06, line 60: should we consider having just one config value for nvml_lib_path?

val06: for simplicity...

pkg/gpu/nvml/lib.go
```diff
@@ -0,0 +1,88 @@
...
+
+		c.lib = nvmlNewFunc(nvml.WithLibraryPath(libpath))
```
val06, line 62: can the nvmlNewFunc fail and then the next function will cause panic?

pkg/gpu/nvml/lib.go
```diff
@@ -0,0 +1,88 @@
...
+	"sync"
+
+	"github.com/NVIDIA/go-nvml/pkg/nvml"
+
+	sysconfig "github.com/DataDog/datadog-agent/cmd/system-probe/config"
+	pkgconfigsetup "github.com/DataDog/datadog-agent/pkg/config/setup"
+	"github.com/DataDog/datadog-agent/pkg/util/flavor"
+)
+
+type nvmlCache struct {
```
val06, line 22: what do you think about adding a `Shutdown` (or `Close`) method to this singleton?

val06 (commented):
left comments

cmd/system-probe/config/ns.go
```diff
@@ -73,6 +73,6 @@ func discoveryNS(k ...string) string {
 }
 
 // gpuNS adds `gpu_monitoring` namespace to config key
-func gpuNS(k ...string) string {
```
hush-hush, line 76: Let move this somewhere in the `pkg` folder. We should not have code in comp or pkg importing `cmd`

hush-hush (changes requested):

pkg/gpu/nvml/lib.go
gjulianm: AFAIK we can't see the system-probe config from the agent and vice-versa, no?

cmd/system-probe/config/ns.go
gjulianm: Fixed by having the "gpu_monitoring" name in a separate package (`pkg/gpu/config/consts`) to avoid issues. However, I see we're importing things from `cmd` in other places (e.g. `pkg/ebpf/config.go`). What problems could that cause?

pkg/gpu/nvml/lib.go
gjulianm: We could add it but I'm not sure where to use it right now

pkg/gpu/nvml/lib.go
gjulianm: In theory it shouldn't, but added a check just in case

pkg/gpu/config/config.go
```diff
@@ -45,7 +43,6 @@ func New() *Config {
 		Config:                *ebpf.NewConfig(),
 		ScanProcessesInterval: time.Duration(spCfg.GetInt(sysconfig.FullKeyPath(GPUNS, "process_scan_interval_seconds"))) * time.Second,
 		InitialProcessSync:    spCfg.GetBool(sysconfig.FullKeyPath(GPUNS, "initial_process_sync")),
-		NVMLLibraryPath:       spCfg.GetString(sysconfig.FullKeyPath(GPUNS, "nvml_lib_path")),
```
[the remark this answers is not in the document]
gjulianm: Moved some things around to have a separate `pkg/gpu/config/consts` that we can import everywhere.

cmd/system-probe/config/ns.go
dustmop: We try to avoid having "spaghetti dependencies" within the agent codebase. Genereally speaking, `cmd` should be the entry points of the binary, so it can import `pkg` and `comp` but not vice versa. It works okay in this case because the subfolder of `cmd/system-probe/config` (which `pkg/ebpf/config.go` depends on) is its own package, but that just obscures the issue: it would be better if such config lived in a lower dependency. Fixing that issue doesn't have to happen in this PR, but it would be nice to not make it worse than it currently is.

brycekahle: We can certainly move the config into a different package, but it wouldn't change any of the semantics. This package is the config for the system-probe. Having it here felt a little more discoverable than having to create a `pkg/system-probe/config`.

pkg/gpu/nvml/lib.go
val06: cc @hush-hush , can we consolidate the config param to be shared between the core and the system-probe agent's components?

hush-hush (approved):

cmd/system-probe/config/ns.go
hush-hush: The general direction the Agent is taking is to have building blocks that can be shipped to customers in different ways. The `cmd` folder is one of them but we start to have the same product available for serverless, agentless, IOT or simply combined in different binary (like the process monitoring product being a standalone process on Windows but not on Linux).

Having a product import code from `cmd` breaks or complexify those other ways to ship features. Like @dustmop said, this is the general direction and can be improve over time in other PRs.

val06 (approved):

dustmop (approved):
LGTM for agent-configuration, thanks for the fixes!

dustmop (approved):
LGTM for agent-configuration, thanks for the fixes!

PR #35401, opened 2025-03-24
[EBPF] gpu: use singleton for NVML access

gjulianm:
### What does this PR do?

This PR is based on #35400 and replaces all instantiations of the NVML library with a call to the singleton.

### Motivation

Single entry point for NVML creation, centralize code and parameters.

Existing tests should pass.

### DIFF
--- a/comp/core/workloadmeta/collectors/internal/nvml/nvml.go
+++ b/comp/core/workloadmeta/collectors/internal/nvml/nvml.go
@@ -19,6 +19,7 @@ import (
 	workloadmeta "github.com/DataDog/datadog-agent/comp/core/workloadmeta/def"
 	"github.com/DataDog/datadog-agent/pkg/config/env"
 	"github.com/DataDog/datadog-agent/pkg/errors"
+	ddnvml "github.com/DataDog/datadog-agent/pkg/gpu/nvml"
 	"github.com/DataDog/datadog-agent/pkg/util/log"
 )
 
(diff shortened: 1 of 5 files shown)


val06 (approved):

comp/core/workloadmeta/collectors/internal/nvml/nvml.go
```diff
@@ -274,12 +259,13 @@ func (c *collector) Start(_ context.Context, store workloadmeta.Component) error
 
 // Pull collects the GPUs available on the node and notifies the store
 func (c *collector) Pull(_ context.Context) error {
-	nvmlLib, err := c.getNVML()
+	nvmlLib, err := ddnvml.GetNvmlLib()
 	if err != nil {
 		return fmt.Errorf("failed to get NVML library: %w", err)
 	}
+	c.nvmlLib = nvmlLib
```
val06, line 266: can we get rid of this field if we have the singleton?

pkg/gpu/nvml/testutil.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"testing"
+
+	"github.com/NVIDIA/go-nvml/pkg/nvml"
+)
+
+// SetSingletonForTest sets the singleton NVML library for testing purposes.
+// This is useful to test the NVML library without having to initialize it
+// manually. It automatically restores the original NVML library on test
+func SetSingletonForTest(t *testing.T, lib nvml.Interface) {
```
val06, line 19:
```suggestion
func WithMockNvml(t *testing.T, lib nvml.Interface) {
```

pkg/gpu/nvml/testutil.go
```diff
@@ -0,0 +1,33 @@
...
+	singleton.isInitialized = true
+
+	t.Cleanup(func() {
+		singleton.mu.Lock()
+		defer singleton.mu.Unlock()
+
+		singleton.lib = nil
+		singleton.isInitialized = false
+	})
+}
```
val06, line 33: I think it is better to restore a previous state, rather than setting the lib to nil and initialized to false. 
([example](https://github.com/DataDog/datadog-agent/pull/35741/files#diff-8faeff52914d9ff889b6c6b72ce25029b9942a0e7377ad94cf426a411104624fR20))

val06 (commented):
added few more comments

comp/core/workloadmeta/collectors/internal/nvml/nvml.go
gjulianm: It makes it easier to have the field, that way we avoid having to error check every time we get the singleton

pkg/gpu/nvml/testutil.go
gjulianm: Changed

pkg/gpu/nvml/testutil.go
gjulianm: I prefer to restore it completely. In tests the singleton should always be uninitialized by default. Restoring the previous value makes it seem like leaving it initialized is supported, which is not and can cause test failures that would depend on execution order. We also shouldn't support nested setting of this variable either.

PR #35402, opened 2025-03-24
gpum: changed gpu_host tag implementation 

val06:
### What does this PR do?

Changes the way we retrieve gpu_host tag, instead of using nvml API, scan /proc/driver/nvidia directory.
if present and has more than 1 entry, we generate the tag

### Motivation

Remove dependency on nvml lib properly mounted into core agent container.
This will allow us to use the `gpu_host` tag as a discovery feature to enable gpu monitoring product

added UTs

### Additional Notes
`/host/proc` is already mounted by the core agent

[Jira ticket](https://datadoghq.atlassian.net/browse/EBPF-685)


pkg/gpu/tags/tags_test.go
```diff
@@ -3,92 +3,111 @@
...
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			nvmlLibrary = &tt.nvml
-			gotTags := GetTags()
+			defer func() {
+				// Clean up any existing files before each test
+				if err := os.RemoveAll(filepath.Join(tmpDir, "driver")); err != nil {
+					t.Fatalf("Failed to clean up test directory: %v", err)
+				}
+			}()
```
gjulianm, line 103: nit: Instead of this, I'd create the tmpDir for each test and pass it as an argument to `setup()`, that way you're completely sure it's a new diretory tree

gjulianm (approved):

pkg/gpu/tags/tags_test.go
val06: i would still need to have a cleanup function that removes those tmp dirs, so i don't really see a benefit in that approach

gjulianm: `t.TempDir` automatically cleans up the directory after the specific test is run

PR #35409, opened 2025-03-24
Fix issue where discovery client erroring would permenantly break KSM

sblumenthal:
### What does this PR do?

This PR fixes an issue with the `GlobalResourceTypeCache` where it could wind up in a permanent error state if the discovery client call failed during initialization. This would result in the KSM check permanently being unable to start until manual action was taken.

### Motivation

See [this notebook](https://app.datadoghq.com/notebook/11964057/ksm-alert-investigation-3-21?view=view-mode) detailing the investigation following an alert

### DIFF
--- a/pkg/collector/corechecks/cluster/ksm/kubernetes_state.go
+++ b/pkg/collector/corechecks/cluster/ksm/kubernetes_state.go
@@ -347,11 +347,6 @@ func (k *KSMCheck) Configure(senderManager sender.SenderManager, integrationConf
 					return err
 				}
 
-				apiServerClient, err := apiserver.GetAPIClient()
-				if err != nil {
-					return err
-				}
-
 				err = apiserver.InitializeGlobalResourceTypeCache(apiServerClient.Cl.Discovery())
 				if err != nil {
 					return err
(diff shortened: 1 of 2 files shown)


pkg/collector/corechecks/cluster/ksm/kubernetes_state.go
```diff
@@ -347,11 +347,6 @@ func (k *KSMCheck) Configure(senderManager sender.SenderManager, integrationConf
 					return err
 				}
 
-				apiServerClient, err := apiserver.GetAPIClient()
```
sblumenthal, line 350: Removed because this is duplicated in the 4 lines directly above

pkg/util/kubernetes/apiserver/resourcetypes.go
```diff
@@ -43,12 +46,24 @@ func InitializeGlobalResourceTypeCache(discoveryClient discovery.DiscoveryInterf
...
+			AttemptMethod: func() error {
+				err := resourceCache.prepopulateCache()
+				if err != nil {
+					return fmt.Errorf("failed to prepopulate resource type cache: %w", err)
+				}
+				return nil
+			},
+			Strategy:          retry.Backoff,
+			InitialRetryDelay: 30 * time.Second,
+			MaxRetryDelay:     5 * time.Minute,
```
sblumenthal, line 60: tried to pick sensible defaults, happy to change them if there is an existing pattern

justin-lesko (approved):

adel121 (approved):

PR #28130, opened 2024-08-01
add retry logic in metadata v4 client

kangyili:
### What does this PR do?

When the agent is deployed on a large EC2 instance running on ECS, such as with 500 tasks running on the same instance, the Datadog agent can encounter timeout errors:
```
2024-08-01 13:24:21 UTC \| CORE \| WARN \| (comp/core/workloadmeta/collectors/internal/ecs/v4parser.go:85 in getTaskWithTagsFromV4Endpoint) \| failed to get task with tags from metadata v4 API: Get "http://169.254.170.2/v4/12345678-44bd-4cc6-9fd9-4f4340c123456/taskWithTags": context deadline exceeded (Client.Timeout exceeded while awaiting headers) |  
```
This is because the default timeout value is not sufficient to wait for responses from the ECS agent.
https://github.com/DataDog/datadog-agent/blob/main/pkg/config/setup/config.go#L524

Other types of error can occur as well when updating or restarting the ECS agent.

This PR improves metadata client by adding a retry logic to reduce these errors.
1. `ecs_metadata_retry_initial_interval ` is the initial interval between retries
2. `ecs_metadata_retry_max_elapsed_time` is the maximum time to retry before giving up
3. `ecs_metadata_retry_timeout_factor` is a factor used to increases the request timeout on each retry

### Describe how to test/QA your changes

It just adds an extra layer to minimise any temporary errors. We can deploy agent on ECS to see if we can see any regression.

PR #35414, opened 2025-03-24
[workloadmeta/collectors/internal/ecs] Use retry option for ECS collector client when getting resource tags

jennchenn:
### What does this PR do?
Add retry option when `DD_ECS_COLLECT_RESOURCE_TAGS_EC2` enabled to prevent throttling issues.

### Motivation

This change makes use of the same retry configuration introduced for `DD_ECS_TASK_COLLECTION_ENABLED` here: https://github.com/DataDog/datadog-agent/pull/28130 to avoid throttling when customers set `DD_ECS_TASK_COLLECTION_ENABLED` to `true`.

1. Deployed the changes on an ECS cluster and verified that metrics/etc. still seemed to be collected
2. Basic usage should also be covered by E2E tests

### Additional Notes

I tried to throttle a workload but was not able to hit a timeout while fetching resource tags to validate that this fix is working.


gabedos (approved):
Using V4 retry logic into V1 looks good to me.

PR #35447, opened 2025-03-25
fix(installer): Use the installed installer binary to run postinst

BaptisteFoy:
### What does this PR do?
- Adds the installer as a separate layer of the agent OCI
- When starting an agent experiment from Remote Config on Linux, if the installer layer exists, then extract that installer & use it. If not, use the installer linked to the daemon.

### Motivation
Use the post install scripts from the agent / installer pair we just downloaded

E2E, lots of manual QA


arbll (approved):
two small comments, lgtm beyond that

chouquette (approved):
LGTM for agent-delivery file

PR #35454, opened 2025-03-25
[EBPF] refactor util/kernel pkg

val06:
### What does this PR do?

This PR is moving code between packages and updated imports, no changes to the implementation of the funcs from the original package

Introducing a new `util/ebpf` pkg to include functions that require cilium pkgs as dependency.
Those functions are used by ebpf based products (CNM, USM, CWS), hence the name.

### Motivation

Avoid importing heavy cilium dependencies when using a simple util function from the util/kernel pkg (adds ~0.9MB and often fails binary_size_check)

linters and existing tests should still pass. no logical code changes in this PR except for moving code between packages

### DIFF
--- a/pkg/collector/corechecks/ebpf/probe/ebpfcheck/probe.go
+++ b/pkg/collector/corechecks/ebpf/probe/ebpfcheck/probe.go
@@ -19,6 +19,7 @@ import (
 	"time"
 	"unsafe"
 
+	ebpf2 "github.com/DataDog/datadog-agent/pkg/util/ebpf"
 	manager "github.com/DataDog/ebpf-manager"
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/asm"
@@ -42,7 +43,7 @@ import (
 // 5.5 for security_perf_event_open
 // 5.0 required for /proc/kallsyms BTF program full names
 // 4.15 required for map names from kernel
-var minimumKernelVersion = kernel.VersionCode(5, 5, 0)
+var minimumKernelVersion = ebpf2.VersionCode(5, 5, 0)
 
 const maxMapsTracked = 20
 
@@ -64,7 +65,7 @@ type Probe struct {
 
 // NewProbe creates a [Probe]
 func NewProbe(cfg *ddebpf.Config) (*Probe, error) {
-	kv, err := kernel.HostVersion()
+	kv, err := ebpf2.HostVersion()
 	if err != nil {
 		return nil, fmt.Errorf("kernel version: %s", err)
 	}
@@ -132,11 +133,11 @@ func startEBPFCheck(buf bytecode.AssetReader, opts manager.Options) (*Probe, err
 
 	// `security_bpf_map_alloc` was renamed to `security_bpf_map_create`
 	// in this commit: https://github.com/torvalds/linux/commit/a2431c7eabcf9bd5a1e7a1f7ecded40fdda4a8c5
-	kv, err := kernel.HostVersion()
+	kv, err := ebpf2.HostVersion()
 	if err != nil {
 		return nil, fmt.Errorf("failed to get kernel version: %s", err)
 	}
-	if kv >= kernel.VersionCode(6, 9, 0) {
+	if kv >= ebpf2.VersionCode(6, 9, 0) {
 		delete(collSpec.Programs, "k_map_alloc")
 	} else {
 		delete(collSpec.Programs, "k_map_create")
--- a/pkg/collector/corechecks/ebpf/probe/ebpfcheck/probe.go
+++ b/pkg/collector/corechecks/ebpf/probe/ebpfcheck/probe.go
@@ -43,7 +43,7 @@ import (
 // 5.5 for security_perf_event_open
 // 5.0 required for /proc/kallsyms BTF program full names
 // 4.15 required for map names from kernel
-var minimumKernelVersion = kernelversion.VersionCode(5, 5, 0)
+var minimumKernelVersion = kernelversion.FromCode(5, 5, 0)
 
 const maxMapsTracked = 20
 
@@ -65,7 +65,7 @@ type Probe struct {
 
 // NewProbe creates a [Probe]
 func NewProbe(cfg *ddebpf.Config) (*Probe, error) {
-	kv, err := kernelversion.HostVersion()
+	kv, err := kernelversion.Host()
 	if err != nil {
 		return nil, fmt.Errorf("kernel version: %s", err)
 	}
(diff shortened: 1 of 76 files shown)


pkg/network/usm/sharedlibraries/ebpf.go
```diff
@@ -15,6 +15,7 @@ import (
 	"strings"
 	"sync"
 
+	ebpf2 "github.com/DataDog/datadog-agent/pkg/util/ebpf"
```
guyarb, line 18: please move to the correct section

pkg/util/ebpf/download_headers.go
pgimalac, line 1: Could you add codeowners for pkg/util/ebpf please ?
I think Agent Runtimes owns it right now because we own `pkg` or `pkg/util` 😄

pkg/network/usm/usm_grpc_monitor_test.go
```diff
@@ -14,6 +14,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/DataDog/datadog-agent/pkg/util/ebpf"
```
guyarb, line 17: please move to the correct section

pkg/network/usm/usm_http2_monitor_test.go
```diff
@@ -26,6 +26,7 @@ import (
 	"time"
 	"unsafe"
 
+	ebpf2 "github.com/DataDog/datadog-agent/pkg/util/ebpf"
```
guyarb, line 29: please move to the correct section

pkg/network/protocols/http/protocol.go
```diff
@@ -12,6 +12,7 @@ import (
 	"io"
 	"unsafe"
 
+	ebpf2 "github.com/DataDog/datadog-agent/pkg/util/ebpf"
```
guyarb, line 15: please move to the correct section

guyarb (changes requested):

pkg/collector/corechecks/ebpf/probe/ebpfcheck/probe.go
```diff
@@ -19,6 +19,7 @@ import (
 	"time"
 	"unsafe"
 
+	ebpf2 "github.com/DataDog/datadog-agent/pkg/util/ebpf"
```
pgimalac, line 22: What do you think of `ebpfutils` or something similar ?
`ebpf` vs `ebpf2` is not super clear 😄

paulcacheux:
I'm not seeing the point, since this new package is still imported by all non-system-probe agents, what am I missing ?

guyarb (approved):

hmahmood:
`ebpfutil.Version[Code]` does not really read very well, as in what version are we talking about here? `kernel.Version` was pretty clear about which version this was referring. Could we do a sub-package `kernel` under `utils/ebpf`? Or maybe rename `Version[Code]` to `KernelVersion[Code]`?

val06:
> I'm not seeing the point, since this new package is still imported by all non-system-probe agents, what am I missing ?

Hmm I think all usages of the new util/ebpf pkg should have linux_bpf build flag and omitted from some of the agent packages

paulcacheux (approved):

brycekahle (changes requested):
It looks like we only import `features.LinuxVersionCode()` from `github.com/cilium/ebpf` in this package. Can we just pull that code in, rather than this massive refactor?

brycekahle:
> use the original implementation

If you mean the string parsing, that is incorrect. The vDSO parsing is the only foolproof way to get the kernel version code.

brycekahle:
> internal cilium structs

They aren't really cilium-specific. They are parsing a standard Linux data structure.

I don't like this refactor because knowing the Linux kernel version is NOT ebpf-specific functionality. We just happen to use a library that also deals with ebpf functionality to do it. That is why I think pulling in the parsing logic makes more sense.

val06:
> > internal cilium structs
> 
> They aren't really cilium-specific. They are parsing a standard Linux data structure.
> 
> I don't like this refactor because knowing the Linux kernel version is NOT ebpf-specific functionality. We just happen to use a library that also deals with ebpf functionality to do it. That is why I think pulling in the parsing logic makes more sense.

we can discuss a different name for the package, rather than ebpf, see also @hmahmood [comment](https://github.com/DataDog/datadog-agent/pull/35454#issuecomment-2751564527), however i am still not convinced for a need to bring all those files. 

Regardless of the implementation, I think it is useful to break the large `util/kernel` package into smaller sub-packages. If you feel like it, feel free to make another PR after this is merged, to replace cilium's API with the "in-house" vdso implementation

hmahmood:
> > > internal cilium structs
> > 
> > 
> > They aren't really cilium-specific. They are parsing a standard Linux data structure.
> > I don't like this refactor because knowing the Linux kernel version is NOT ebpf-specific functionality. We just happen to use a library that also deals with ebpf functionality to do it. That is why I think pulling in the parsing logic makes more sense.
> 
> we can discuss a different name for the package, rather than ebpf, see also @hmahmood [comment](https://github.com/DataDog/datadog-agent/pull/35454#issuecomment-2751564527), however i am still not convinced for a need to bring all those files.
> 
> Regardless of the implementation, I think it is useful to break the large `util/kernel` package into smaller sub-packages. If you feel like it, feel free to make another PR after this is merged, to replace cilium's API with the "in-house" vdso implementation
> 
> what do you think about moving it as a sub-package under `util/kernel` and calling it `version`. then also rename exported structs to omit the "Version" from the names: (cc @hmahmood )
> 
> ```
> version.Host
> version.MustHost
> version.Parse
> version.FromCode
> version.FromReleaseString
> ```
> 
> and also keep the Version struct so it will be `version.Version`
> 
> I can also add an alias to the package in all places and call it `kernelversion` in the import line

For me `kernel.Version` still reads much better. With a `version` package, we would have code like `version.Code`, which still doesn't tell me what kind of version this is, so the user has to alias the package in the code to be clear what version it is.

val06:
> > > > internal cilium structs
> > > 
> > > 
> > > They aren't really cilium-specific. They are parsing a standard Linux data structure.
> > > I don't like this refactor because knowing the Linux kernel version is NOT ebpf-specific functionality. We just happen to use a library that also deals with ebpf functionality to do it. That is why I think pulling in the parsing logic makes more sense.
> > 
> > 
> > we can discuss a different name for the package, rather than ebpf, see also @hmahmood [comment](https://github.com/DataDog/datadog-agent/pull/35454#issuecomment-2751564527), however i am still not convinced for a need to bring all those files.
> > Regardless of the implementation, I think it is useful to break the large `util/kernel` package into smaller sub-packages. If you feel like it, feel free to make another PR after this is merged, to replace cilium's API with the "in-house" vdso implementation
> > what do you think about moving it as a sub-package under `util/kernel` and calling it `version`. then also rename exported structs to omit the "Version" from the names: (cc @hmahmood )
> > ```
> > version.Host
> > version.MustHost
> > version.Parse
> > version.FromCode
> > version.FromReleaseString
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and also keep the Version struct so it will be `version.Version`
> > I can also add an alias to the package in all places and call it `kernelversion` in the import line
> 
> For me `kernel.Version` still reads much better. With a `version` package, we would have code like `version.Code`, which still doesn't tell me what kind of version this is, so the user has to alias the package in the code to be clear what version it is.

i will add the `kernelversion` alias to all existing imports + the comments above the exported functions should make the purpose very clear

hmahmood (approved):

vitkyrka (approved):

PR #35460, opened 2025-03-25
Update agent-platform install-script test to not rely on the install-script directly

KevinFairise2:
### What does this PR do?

We want to stop relying on the install-script to install the agent in the e2e test framework.
The main reason is that we now want to support installing unsigned packages built from dev laptops. The install-script expect your package to be available on a apt/yum repository and signed properly.

https://docs.google.com/document/d/1_L07m7Xxg_xuG6Y2d-PE18Ezu_bEoLlGqFBMCxOSQTQ/edit?tab=t.0

These tests should also be renamed at some point, because their goal is more to test that the agent can be installed properly than testing that the install-script works. Testing the install-script is already covered by e2e tests running on agent-linux-install-script repository.
It might also reduce the cost of the e2e tests infrastructure because pulling the agent from public Cloudfront should bring more traffic through the NAT Gateway that represent most of the cost of the infra.

### DIFF
--- a/.gitlab/e2e/e2e.yml
+++ b/.gitlab/e2e/e2e.yml
@@ -99,7 +99,7 @@
     - !reference [.needs_new_e2e_template]
     - datadog-agent-7-x64
   variables:
-    LOCAL_PACKAGE_FLAG: "--local-package OMNIBUS_PACKAGE_DIR"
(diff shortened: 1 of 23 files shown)


KevinFairise2:

KevinFairise2:

KevinFairise2:

KevinFairise2:

PR #35462, opened 2025-03-25
Prepare fixes for tests when nodetreemodel is used by default

dustmop:
### What does this PR do?

This PR fixes a number of issues we found when testing out nodetreemodel as the default config implementation. In particular, the unit tests make very different assumptions than the actual agent binaries, often treating the config more dynamically.

Fixes include:
* Introduce `SetTestOnlyDynamicSchema`, which allows more flexible usage patterns for tests
  - it's okay to read the config before its schema is built
  - it's okay to modify the config schema after it gets built
  - unknown keys can be assigned and retrieved
* Call this method for all mock config constructors and test-only constructors
* Replace calls to SetDefault with `SetWithoutSource`
* Use `UnsetForSource` where applicable

### Motivation

Found issues with tests when testing out nodetreemodel as the default config implementation.

Set `DD_CONF_NODETREEMODEL=enable` and ran the tests. Some of the tests are still failing with this env var, those will be fixed in follow-up PRs.

### DIFF
--- a/comp/core/configsync/configsyncimpl/sync_integration_test.go
+++ b/comp/core/configsync/configsyncimpl/sync_integration_test.go
@@ -18,7 +18,7 @@ import (
 	"go.uber.org/atomic"
 
 	pkgconfigmodel "github.com/DataDog/datadog-agent/pkg/config/model"
-	viperconfig "github.com/DataDog/datadog-agent/pkg/config/viperconfig"
+	"github.com/DataDog/datadog-agent/pkg/config/viperconfig"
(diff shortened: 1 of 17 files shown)


rahulkaukuntla (approved):

davidor (approved):
:+1: for the file owned by container-platform

L3n41c (approved):
👍🏻 LGTM for the file owned by @DataDog/container-integrations.

Jakobovski (approved):

PR #35464, opened 2025-03-25
[APM] Attach ContainerID on aggregated CSS payloads

ichinaski:
### What does this PR do?

Flushing ClientStat buckets on the ClientStatsAggregator was missing this field, which is kept on the aggregation key.

This PR ensures that the ContainerID used during the aggregation is also propagated when flushing the payloads.

The change is tested via a unit test and an additional E2E test.


songy23 (approved):
(no impact on OTel)

ajgajg1134 (approved):

PR #35467, opened 2025-03-25
[CONTP-724] add csi configs for dca ac

adel121:
### What does this PR do?

Adds feature flags to enable/disable csi driver on three levels:
* Global: `csi.enabled`
* Admission controller: `admission_controller.csi.enabled`
* Inject Config webhook: `admission_controller.inject_config.csi.enabled`

The inject webhook will leverage CSI only if all flags are set to `true`.

### Motivation

We allow enabling/disabling CSI on different level to allow the user enable/disable the feature on different scopes. This will be useful when other features decide to leverage CSI, user should be able to enable/disable CSI for different features independently, and should also be able to disable/enable it globally.

We already have E2E tests, and I updated the unit tests also.

None.


hush-hush (approved):

pkg/config/setup/config.go
```diff
@@ -767,6 +768,7 @@ func InitConfig(config pkgconfigmodel.Setup) {
 	config.BindEnvAndSetDefault("admission_controller.certificate.secret_name", "webhook-certificate") // name of the Secret object containing the webhook certificate
 	config.BindEnvAndSetDefault("admission_controller.webhook_name", "datadog-webhook")
 	config.BindEnvAndSetDefault("admission_controller.inject_config.enabled", true)
+	config.BindEnvAndSetDefault("admission_controller.inject_config.csi.enabled", false)
```
gabedos, line 771: Just thinking out loud. If a customer wants to enabled the csi driver injection via the config webhook. They would need to manually set all 3 `csi_enabled`, `admission_controller.csi.enabled`, and `admission_controller.inject_config.csi.enabled` to `true` in various locations in their helm chart/operator.

Would it make sense to default the higher level attributes to `true` like `csi_enabled` and `admission_controller.csi.enabled` but then keep the flag directly related to the feature `false`. This would then give users the flexibility to enable and disable more easily.

gabedos (approved):
Small comment idea for defaults

pkg/config/setup/config.go
adel121: Totally agree. 

I defaulted everything to `false` for now because we don't want to enable it by default anywhere at all. 

Once everything is stable, we can set better defaults as you suggested

PR #35468, opened 2025-03-25
[ebpf] simplify system-probe config imports

val06:
### What does this PR do?

Introduces a new `config_linux_bpf.go` to use the ebpf imports only for system-probe builds
otherwise fallback to config_linux which doesn't support ProcessEventDataStream and ebpf prebuild fallback 

### Motivation

prevent importing ebpf related imports into the core-agent (that is using system-probe/config pkg)

linter and tests should pass

### Additional Notes

helper PR for the larger [refactor](https://github.com/DataDog/datadog-agent/pull/35454)

### DIFF
--- a/cmd/system-probe/config/config_unsupported.go
+++ b/cmd/system-probe/config/config_unsupported.go
@@ -3,7 +3,7 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2016-present Datadog, Inc.
 
-//go:build !unix && !linux_bpf && !windows
+//go:build !unix && !windows
 
 package config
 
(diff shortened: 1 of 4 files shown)


brycekahle:
This PR highlights a problem where the non-system-processes may have a different picture of what the running system-probe config is compared to what system-probe believes internally.

cmd/system-probe/config/config_unsupported.go
```diff
@@ -3,7 +3,7 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2016-present Datadog, Inc.
 
-//go:build !unix && !windows
+//go:build !unix && !linux_bpf && !windows
```
brycekahle, line 6: `unix` is inclusive of `linux`, so I don't think `!linux_bpf` is required.

brycekahle: If we had `linux_bpf` but no `linux`, I think a lot of other things would break.

val06: removed

brycekahle (approved):

PR #35483, opened 2025-03-25
Add framework to support the distributed sends architecture

ddrthall:
### What does this PR do?
This PR introduces a nonfunctional architectural shift that brings the underlying codebase closer to the enablement of RTT Fairness/Distributed Sends. PR that enables these features and documents the full change scope is https://github.com/DataDog/datadog-agent/pull/35484

 The crux of this PR is the multiplexing of the sender object in the logs pipeline. Specifically a single "sender" instance can now possess a variety of payload queues and workers for those queues. Each "worker" has the full functionality of the older sender objects, and this architectural shift allows for the abstraction and separation of sender count from pipeline count.

Unit test coverage was expanded, and existing test cases were largely relied upon to verify that no regressions were introduced in this change.

### DIFF
--- a/pkg/logs/sender/sender.go
+++ b/pkg/logs/sender/sender.go
@@ -6,171 +6,157 @@
 package sender
 
 import (
-	"strconv"
 	"sync"
-	"time"
 
 	pkgconfigmodel "github.com/DataDog/datadog-agent/pkg/config/model"
+	"github.com/DataDog/datadog-agent/pkg/logs/auditor"
 	"github.com/DataDog/datadog-agent/pkg/logs/client"
 	"github.com/DataDog/datadog-agent/pkg/logs/message"
 	"github.com/DataDog/datadog-agent/pkg/logs/metrics"
-	"github.com/DataDog/datadog-agent/pkg/telemetry"
+	"github.com/DataDog/datadog-agent/pkg/util/log"
+
+	"go.uber.org/atomic"
 )
 
-var (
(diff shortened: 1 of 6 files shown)


pkg/logs/sender/sender.go
```diff
@@ -6,171 +6,157 @@
...
+	}
 
-		if s.senderDoneChan != nil && s.flushWg != nil {
-			// Wait for all destinations to finish sending the payload
-			senderDoneWg.Wait()
-			// Decrement the wait group when this payload has been sent
-			s.flushWg.Done()
+	queues := make([]chan *message.Payload, queueCount)
+	for idx := range queueCount {
+		queues[idx] = make(chan *message.Payload, workersPerQueue+1)
```
gh123man, line 110: Perhapse I am missing something so please correct me, but why do we want the queue length defined in termms of the # of workers?

ddrthall: The original implementation had each sender having an input channel of size 1. Since the input channel is now shared between multiple senders, equivalent behavior is to scale it by the number of workers (with the +1 being somewhat arbitrary, but "free" given the payload halving MR)

gh123man: cool makes sense. 

maybe worth dropping a code comment here for future reference.

gh123man (approved):

pkg/logs/sender/sender.go
ddrthall: I've dropped the comment in the next MR in the chain (https://github.com/DataDog/datadog-agent/pull/35685) so as to not spend another five years waiting on pipeline completion

PR #35374, opened 2025-03-21
AGNTLOG-148 - retire legacy log sender components and harden serverless behavior

ddrthall:
### What does this PR do?
1. Perform a final, nonfunctional retirement of the legacy logs pipeline sender flow. All systems should already be explicitly using the new logic except for epforwarder, which this MR shifts to an equivalent function in the new sender logic.
2. Add logic to harden serverless behavior, with the goal of preventing race conditions on the wait group used for serverless flushing. 

### Describe how you validated your changes

Validation was performed via a soak in the serverless testing environment - log throughput and accuracy was judged correct, and we no longer observed panics in the serverless runtime.

PR #35484, opened 2025-03-25
Enable distributed sends and RTT fairness for core logs pipeline

ddrthall:
### What does this PR do?
This PR enables distributed sends/RTT fairness for the majority of the codebase. Log flow in this new system is slightly different for various flavors of the pipeline.

#### Original Architecture
original

In the original architecture, every log pipeline would be associated with a single sender. That sender in turn pushed each of its received logs synchronously to the intake. This model worked well when the system was cpu bound rather than network bound, as that meant that the logs pipeline was the critical time sink and the sender entities would quickly spin through whatever log backlog existed. However, in systems where the transport time to the intake was a significant limiting factor, the synchronous nature of the sender was a major hurdle that could prevent what would otherwise be a timely transfer of logs.

In order to make the system more responsive in more use cases, three points of the sender have been expanded with the option of concurrent operation: the input queues, the queue processing logic, and the destination output connections. Different use of the system are set to utilize different points of concurrency. In order to mask this behavior to the caller, the first, universal change in the system architecture is to have every pipeline interact with a singular, shared sender object. This sender can be configured to perform the right level of concurrency at the right stage of the process.

#### HTTP Mode
http

Log pipelines pointed at http-based intakes have opted to make use of concurrency in the destination entity, as follows:

All pipelines are funneled into a single queue and worker to allow for full sender utilization, regardless of how busy individual pipelines are or are not. Each individual destination entity in the sender will be able to send multiple http payloads simultaneously via the RTT fairness algorithm. Specifically, a "minWorker" value is calculated from the number of pipelines in the provider and a "maxWorker" value is calculated as 10 times the min. A target latency (150ms at this time) is established in the code, which is the latency the minimum worker count is expected to be able to handle. Observed latency higher than the target latency will prompt the worker count to increase at the same relative amount, e.g. an observed latency 2x higher than the target latency will prompt 2x workers, up to the max worker count. 

By default the minWorker component of the destination is set equal to the pipeline count, which mirrors the original behavior of having one sender per pipeline.

#### TCP Mode
tcp

Log pipelines pointed at tcp-based intakes have opted to make use of concurrency at the queue processing stage, as follows:

All pipelines are funneled into a single queue within the sender, and from there dispatched to one of several worker entities. Each workers funnels a log synchronously through the destinations (destination minWorkers == maxWorkers == 1). This system requires less of an overhaul in the tcp destination entity (which isn't widely utilized and doesn't actually support > 1 worker in its pool), but still allows a level of concurrency for each individual pipeline with the queue worker count.

By default, the queue worker count is set equal to the pipeline count, which mirrors the original behavior of having one sender per pipeline.

#### Legacy Mode
legacy

Legacy mode (enabled via the config flag logs_config.disable_distributed_senders) will attempt to mirror the original architecture's behavior while keeping the new control flow of the sender. It does this by making use of concurrency at the queue level, as follows:

All pipelines pull a queue from the sender to pipe their output into. With legacy mode enabled, this queue will be distinct per pipeline, and every entity listening from a given queue will have concurrency set to 1 (e.g. queue worker count = 1 and destination max workers = 1). This setup is important for some systems that might not want to introduce concurrency to the logs pipeline because said concurrency will disrupt log ordering guarantees. 

#### Configuration
In addition to the new configuration option `logs_config.disable_distributed_senders`, the endpoint's `batch_max_concurrent_sends` configuration option has been integrated in to the new architecture. Agents with this option set (as a reasonable number of them do in the wild) will adjust worker pool behavior for HTTP Mode and Legacy Mode. In HTTP, a given value of x for `batch_max_concurrent_sends` will set minWorkers == maxWorkers == numberOfPipelines * `batch_max_concurrent_sends`. This is the closest reasonable mirror to the setting's impact on current production code, which produces an equivalent amount of HTTP workers spread out over a number of senders. Similarly, Legacy mode will set minWorkers == maxWorkers == `batch_max_concurrent_sends` for each of its numberOfPipelines senders. 

Utilizing this configuration as the primary method for worker adjustment comes with a bit of a tradeoff, as a user who is trying to optimize for 6 workers with the default 4 pipelines can instead only choose between 4 (`batch_max_concurrent_sends` = 1) or 8 (`batch_max_concurrent_sends` = 2). This has been judged a reasonable tradeoff for the foreseeable future, as the current architecture's behavior is no different and having excess workers allocated shouldn't translate to a noticeable resource waste.

### Motivation
Distributing the senders amongst the pipelines with an RTT fairness algorithm backing a large default concurrency increase will allow for a much larger throughput of logs data in systems operating under heavy load and high latency. Example throughput change in SMP's file_to_blackhole_1000ms_latency_linear_load regression test:
image

We also see noticeable decreased RSS usage for systems under a reasonable amount of load but experiencing moderately high latency. Example RSS change in SMP's file_to_blackhole_500ms_latency regression test:
image

Extensive performance testing had a byproduct of serving as useful correctness verification. Unit tests have been expanded where necessary to lock down specific behaviors. A number of cases have additionally been verified by hand, including reliable endpoint behavior, connection reuse for http, and destination worker reset on 429 http responses.

Increased connections to datadog's fabric as a result of the increased destination concurrency will be mitigated by the enablement of http2 in the fleet, but nonzero. The latency threshold after which connection counts will increase was chosen to reduce the chance of seeing a large increase in connections on release.

### Additional Notes

For record keeping purposes, the PRs that built the underlying framework for this change are
https://github.com/DataDog/datadog-agent/pull/35351
https://github.com/DataDog/datadog-agent/pull/35329
https://github.com/DataDog/datadog-agent/pull/35483
https://github.com/DataDog/datadog-agent/pull/35685
https://github.com/DataDog/datadog-agent/pull/35374

### DIFF
--- a/comp/logs/agent/agentimpl/agent_serverless_init.go
+++ b/comp/logs/agent/agentimpl/agent_serverless_init.go
@@ -49,7 +49,19 @@ func (a *logAgent) SetupPipeline(
 	destinationsCtx := client.NewDestinationsContext()
 
 	// setup the pipeline provider that provides pairs of processor and sender
-	pipelineProvider := pipeline.NewServerlessProvider(a.config.GetInt("logs_config.pipelines"), a.auditor, diagnosticMessageReceiver, processingRules, a.endpoints, destinationsCtx, NewStatusProvider(), a.hostname, a.config, a.compression)
+	pipelineProvider := pipeline.NewProvider(
+		a.config.GetInt("logs_config.pipelines"),
+		a.auditor,
+		diagnosticMessageReceiver,
+		processingRules, a.endpoints,
+		destinationsCtx,
+		NewStatusProvider(),
+		a.hostname,
+		a.config,
+		a.compression,
+		true, // disable distributed sending for serverless
+		true, // serverless
+	)
 
 	lnchrs := launchers.NewLaunchers(a.sources, pipelineProvider, a.auditor, a.tracker)
 	lnchrs.AddLauncher(channel.NewLauncher())
--- a/comp/otelcol/logsagentpipeline/logsagentpipelineimpl/agent.go
+++ b/comp/otelcol/logsagentpipeline/logsagentpipelineimpl/agent.go
@@ -214,7 +214,20 @@ func (a *Agent) SetupPipeline(
 	destinationsCtx := client.NewDestinationsContext()
 
 	// setup the pipeline provider that provides pairs of processor and sender
-	pipelineProvider := pipeline.NewProvider(a.config.GetInt("logs_config.pipelines"), auditor, &diagnostic.NoopMessageReceiver{}, processingRules, a.endpoints, destinationsCtx, NewStatusProvider(), a.hostname, a.config, a.compression)
+	pipelineProvider := pipeline.NewProvider(
+		a.config.GetInt("logs_config.pipelines"),
+		auditor,
+		&diagnostic.NoopMessageReceiver{},
+		processingRules,
+		a.endpoints,
+		destinationsCtx,
+		NewStatusProvider(),
+		a.hostname,
+		a.config,
+		a.compression,
+		a.config.GetBool("logs_config.disable_distributed_senders"),
+		false, // serverless
+	)
 
 	a.auditor = auditor
 	a.destinationsCtx = destinationsCtx
--- a/pkg/compliance/reporter.go
+++ b/pkg/compliance/reporter.go
@@ -44,7 +44,21 @@ func NewLogReporter(hostname string, sourceName, sourceType string, endpoints *c
 	auditor.Start()
 
 	// setup the pipeline provider that provides pairs of processor and sender
-	pipelineProvider := pipeline.NewProvider(4, auditor, &diagnostic.NoopMessageReceiver{}, nil, endpoints, dstcontext, &common.NoopStatusProvider{}, hostnameimpl.NewHostnameService(), pkgconfigsetup.Datadog(), compression)
+	cfg := pkgconfigsetup.Datadog()
+	pipelineProvider := pipeline.NewProvider(
+		4,
+		auditor,
+		&diagnostic.NoopMessageReceiver{},
+		nil, // processingRules
+		endpoints,
+		dstcontext,
+		&common.NoopStatusProvider{},
+		hostnameimpl.NewHostnameService(),
+		cfg,
+		compression,
+		cfg.GetBool("logs_config.disable_distributed_senders"),
+		false, // serverless
+	)
 	pipelineProvider.Start()
 
 	logSource := sources.NewLogSource(
--- a/pkg/security/reporter/reporter.go
+++ b/pkg/security/reporter/reporter.go
@@ -52,7 +52,21 @@ func newReporter(hostname string, stopper startstop.Stopper, sourceName, sourceT
 	stopper.Add(auditor)
 
 	// setup the pipeline provider that provides pairs of processor and sender
-	pipelineProvider := pipeline.NewProvider(4, auditor, &diagnostic.NoopMessageReceiver{}, nil, endpoints, context, &seccommon.NoopStatusProvider{}, hostnameimpl.NewHostnameService(), pkgconfigsetup.Datadog(), compression)
+	cfg := pkgconfigsetup.Datadog()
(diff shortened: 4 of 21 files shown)


comp/logs/agent/agentimpl/agent_serverless_init.go
```diff
@@ -49,7 +49,19 @@ func (a *logAgent) SetupPipeline(
...
+		a.config.GetInt("logs_config.pipelines"),
+		a.auditor,
+		diagnosticMessageReceiver,
+		processingRules, a.endpoints,
+		destinationsCtx,
+		NewStatusProvider(),
+		a.hostname,
+		a.config,
+		a.compression,
+		false, // disable distributed sending for serverless
```
ddrthall, line 62: Note: serverless support for distributed sending (via the TCP Mode) is present in the codebase. However, I'm forcing serverless to Legacy Mode unless the serverless team has a strong opinion otherwise, since it is marginally safer.

pkg/compliance/reporter.go
```diff
@@ -44,7 +44,21 @@ func NewLogReporter(hostname string, sourceName, sourceType string, endpoints *c
...
+		auditor,
+		&diagnostic.NoopMessageReceiver{},
+		nil, // processingRules
+		endpoints,
+		dstcontext,
+		&common.NoopStatusProvider{},
+		hostnameimpl.NewHostnameService(),
+		cfg,
+		compression,
+		cfg.GetBool("logs_config.disable_distributed_senders"),
```
ddrthall, line 59: Notice for the compliance team: seeking confirmation that this is the behavior you want from the logs pipeline

pkg/security/reporter/reporter.go
```diff
@@ -52,7 +52,21 @@ func newReporter(hostname string, stopper startstop.Stopper, sourceName, sourceT
...
+		auditor,
+		&diagnostic.NoopMessageReceiver{},
+		nil,
+		endpoints,
+		context,
+		&seccommon.NoopStatusProvider{},
+		hostnameimpl.NewHostnameService(),
+		cfg,
+		compression,
+		cfg.GetBool("logs_config.disable_distributed_senders"),
```
ddrthall, line 67: Notice for the security team: seeking confirmation that this is the behavior you want from the logs pipeline

comp/otelcol/logsagentpipeline/logsagentpipelineimpl/agent.go
```diff
@@ -214,7 +214,20 @@ func (a *Agent) SetupPipeline(
...
+		auditor,
+		&diagnostic.NoopMessageReceiver{},
+		processingRules,
+		a.endpoints,
+		destinationsCtx,
+		NewStatusProvider(),
+		a.hostname,
+		a.config,
+		a.compression,
+		a.config.GetBool("logs_config.disable_distributed_senders"),
```
ddrthall, line 228: Notice for the otel team: seeking confirmation that this is the behavior you want from the logs pipeline

comp/logs/agent/agentimpl/agent_serverless_init.go
```diff
@@ -49,7 +49,19 @@ func (a *logAgent) SetupPipeline(
...
+		a.config.GetInt("logs_config.pipelines"),
+		a.auditor,
+		diagnosticMessageReceiver,
+		processingRules, a.endpoints,
+		destinationsCtx,
+		NewStatusProvider(),
+		a.hostname,
+		a.config,
+		a.compression,
+		true, // disable distributed sending for serverless
```
ddrthall, line 62: Note: serverless support for distributed sending (via the TCP Mode) is present in the codebase. However, I'm forcing serverless to Legacy Mode unless the serverless team has a strong opinion otherwise, since it is marginally safer.

michaelcretzman (approved):

DDuongNguyen (approved):
LGTM! 👍🏼

liustanley (approved):

dustmop (approved):

pkg/compliance/reporter.go
paulcacheux: I think the behavior we want is as close to the default/what logs is using as possible to not hit any hidden bugs

paulcacheux (approved):

gh123man (approved):
:shipit: 
excellent PR description and diagrams. I foresee these being referenced long into the future :)

PR #35490, opened 2025-03-25
move kernel headers logic to own package

brycekahle:
### What does this PR do?

Moves kernel header related logic to a subpackage

### Motivation

Intentional usage of the headers logic, since it carries some large imports with it.

### DIFF
--- a/pkg/util/kernel/headers/find_headers.go
+++ b/pkg/util/kernel/headers/find_headers.go
@@ -5,7 +5,8 @@
 
 //go:build linux && linux_bpf
 
-package kernel
+// Package headers is utilities for downloading Linux kernel headers
+package headers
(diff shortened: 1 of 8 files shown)


pkg/util/kernel/headers/find_headers.go
```diff
@@ -5,7 +5,8 @@
 
 //go:build linux && linux_bpf
 
-package kernel
+// Package headers is utilities for downloading Linux kernel headers
+package headers
```
hmahmood, line 9: Do we also want to remove the `header` prefix from the struct names, etc. in this new package?

hmahmood (approved):

val06 (approved):

PR #32530, opened 2024-12-27
Add WLAN Integration

NouemanKHAL:
### What does this PR do?

Adds a WLAN integration as a corecheck in the agent.

### Motivation

Monitor the current Wi-Fi interface in use and collect various metrics to check the health of the Wi-Fi connection.
For now, the check collects the following data:
- RSSI (Received signal strength indicator)
- Noise (on MacOs)
- Transmit Rate
- Receive Rate (on Windows)
- Channel Swap Events
- Roaming Events
- PHYMode (to determine whether the Wi-Fi interface is active or not)

### Describe how you validated your changes

- Unit Tests
- Manual Testing

### Possible Drawbacks / Trade-offs
- Requires Location Services Access on Mac OS

PR #35495, opened 2025-03-26
Add Windows platform for WiFi check

iglendd:
### What does this PR do?
Add Windows implementation to WiFi check provided in the base branch (https://github.com/DataDog/datadog-agent/pull/32530) which provided both cross-platform and Mac OS  code.


pkg/collector/corechecks/net/wlan/wlan_darwin.go
```diff
@@ -20,6 +20,38 @@ import (
...
+const (
+	phyModeNone phyMode = iota // No PHY mode.
+	phyMode11a                 // IEEE 802.11a PHY.
+	phyMode11b                 // IEEE 802.11b PHY.
+	phyMode11g                 // IEEE 802.11g PHY.
+	phyMode11n                 // IEEE 802.11n PHY.
+	phyMode11ac                // IEEE 802.11ac PHY.
+)
+
+func formatPhy(phy phyMode) string {
```
NouemanKHAL, line 36:
```suggestion
func (phyMode p) String() string {
```
Let's define the `String()` method instead to be a [Stringer](https://go.dev/tour/methods/17) Interface.

pkg/collector/corechecks/net/wlan/wlan_darwin.go
```diff
@@ -50,7 +82,7 @@ func GetWiFiInfo() (WiFiInfo, error) {
 		Noise:           int(info.noise),
 		TransmitRate:    float64(info.transmitRate),
 		HardwareAddress: hardwareAddress,
-		ActivePHYMode:   int(info.activePHYMode),
+		PhyMode:         formatPhy(phyMode(info.activePHYMode)),
```
NouemanKHAL, line 85:
```suggestion
		PhyMode:         phyMode(info.activePHYMode),
```
This should probably work if we define the String() method.
if not, we can just call string().

pkg/collector/corechecks/net/wlan/wlan_nix.go
```diff
@@ -3,13 +3,13 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2016-present Datadog, Inc.
 
-//go:build !darwin
+//go:build freebsd || linux || netbsd || openbsd || solaris || dragonfly || aix
```
NouemanKHAL, line 6:
```suggestion
//go:build !darwin && !windows
```
How about this?

iglendd: yeah, this is better, I just copied above from some other source, thank you

pkg/collector/corechecks/net/wlan/wlan_darwin.go
iglendd: Sure, if it is more idiomatic (in fact CoPilot suggested just that). I just did similar things for windows code. I did not feel that it worth to objectivezing half a dozen of enumerations. But I will do your suggestion, thank you.

iglendd: Also I am replacing "unknown" on "" because in wlan.go (and tests) I saw substitution of "" on "unknown"

NouemanKHAL: Yes, that's the idiomatic way of doing string enums in go, example: https://gobyexample.com/enums

pkg/collector/corechecks/net/wlan/wlan_darwin.go
iglendd: Thank you, I cannot believe I missed that (plus I did not check CLI results). Addressed.

pkg/collector/corechecks/net/wlan/wlan_windows.go
pgimalac, line 1: Do we need to keep all the commented out types and constants ?

pkg/collector/corechecks/net/wlan/wlan_windows.go
```diff
@@ -0,0 +1,696 @@
...
+	}
+
+	return string(dot11Ssid.UcSSID[:ssidLen])
+}
+
+func formatMacAddress(mac []byte) string {
+	if len(mac) == 6 {
+		return fmt.Sprintf("%02X:%02X:%02X:%02X:%02X:%02X", mac[0], mac[1], mac[2], mac[3], mac[4], mac[5])
+	} else if len(mac) >= 8 {
+		return fmt.Sprintf("%02X:%02X:%02X:%02X:%02X:%02X:%02X:%02X", mac[0], mac[1], mac[2], mac[3], mac[4], mac[5], mac[6], mac[7])
```
pgimalac, line 343: Are we fine with not having the remaining bytes if we have more than 8 ?

.github/CODEOWNERS
```diff
@@ -431,6 +431,7 @@
 /pkg/collector/corechecks/network-devices/         @DataDog/ndm-integrations
 /pkg/collector/corechecks/orchestrator/   @DataDog/container-app
 /pkg/collector/corechecks/net/          @DataDog/agent-runtimes
+/pkg/collector/corechecks/net/wlan/     @DataDog/windows-agent
```
pgimalac, line 434: I believe wlan is owned by agent-runtimes, windows-agent should probably just own the windows implementation ?

pgimalac (approved):
LGTM

pkg/collector/corechecks/net/wlan/wlan_windows.go
iglendd: Good question. The reason that I am not 100% sure if we would need telemetry from the commented out code. I am a bit torn (and would rather avoid coding/discovering it again since it was a bit of pain).

Also, thank you for the review, it is a partial PR since it will not be merged directly into main branch but instead its target is a branch which has Mac code plus overall check structure which is not merged yet.

pkg/collector/corechecks/net/wlan/wlan_windows.go
iglendd: It should not be more than 6 bytes. In fact it should be only 6 bytes. I was a bit overzealous. I will simplify the code to report only on "proper" mac address. Thank you.

pgimalac: Yeah I actually started reading the wikipedia article as I wasn't sure if it could be something else than 6
I figured this might also handle some different kind of address

NouemanKHAL (approved):

.github/CODEOWNERS
iglendd: No, the change was deliberate and not incidental. Will discuss OOTB.

PR #35497, opened 2025-03-26
[DJM] Handle custom host tags

aboitreaud:
### What does this PR do?

For all DJM installer scripts, extract the custom host tags from `DD_TAGS` and `DD_EXTRA_TAGS`. The split is done on either space or comma (space for legacy reasons from our old init scripts). We allow for key-only tags, thus not checking for `:` in the tags, as this is generally [allowed by Datadog](https://docs.datadoghq.com/getting_started/tagging/#define-tags) (`A tag can be in the format value or <KEY>:<VALUE>`)

For the default installer script, add tag extraction on `DD_EXTRA_TAGS`, on top of  what's already done (`DD_TAGS` and `DD_HOST_TAGS` now deprecated). The split is on comma, not modified.

DD_EXTRA_TAGS is a param specified [here in agent config](https://github.com/DataDog/datadog-agent/blob/main/pkg/config/config_template.yaml#L144).

Lastly in DJM scripts, we are setting so tags in clear on the install spans reported to Datadog so that it makes debugging easier (`is_master`/`spark_node`, `emr_version`, `databricks_is_job_cluster `, `workspace`, and `cluster_id`).

### DIFF
--- a/pkg/fleet/installer/setup/defaultscript/default_script.go
+++ b/pkg/fleet/installer/setup/defaultscript/default_script.go
@@ -149,7 +149,7 @@ func setConfigTags(s *common.Setup) {
 		}
 	}
 	if tags, ok := os.LookupEnv("DD_EXTRA_TAGS"); ok {
-		s.Config.DatadogYAML.Tags = strings.Split(tags, ",")
+		s.Config.DatadogYAML.ExtraTags = strings.Split(tags, ",")
 	}
 }
 
--- a/pkg/fleet/installer/setup/defaultscript/default_script.go
+++ b/pkg/fleet/installer/setup/defaultscript/default_script.go
@@ -148,6 +148,9 @@ func setConfigTags(s *common.Setup) {
 			s.Config.DatadogYAML.Tags = strings.Split(tags, ",")
 		}
 	}
+	if tags, ok := os.LookupEnv("DD_EXTRA_TAGS"); ok {
+		s.Config.DatadogYAML.Tags = strings.Split(tags, ",")
+	}
 }
 
 // installAgentPackage installs the agent package
--- a/pkg/fleet/installer/setup/djm/databricks.go
+++ b/pkg/fleet/installer/setup/djm/databricks.go
@@ -156,6 +156,7 @@ func setupCommonHostTags(s *common.Setup) {
 		setHostTag(s, "runid", runID, false)
 	}
 	setHostTag(s, "data_workload_monitoring_trial", "true", false)
+	addCustomHostTags(s)
 }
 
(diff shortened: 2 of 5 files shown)


pkg/fleet/installer/setup/defaultscript/default_script.go
```diff
@@ -148,6 +148,9 @@ func setConfigTags(s *common.Setup) {
 			s.Config.DatadogYAML.Tags = strings.Split(tags, ",")
 		}
 	}
+	if tags, ok := os.LookupEnv("DD_EXTRA_TAGS"); ok {
+		s.Config.DatadogYAML.Tags = strings.Split(tags, ",")
+	}
```
BaptisteFoy, line 153: Shouldn't we append them to existing tags?

BaptisteFoy: TBH I wonder if we shouldn't even add them in a separate ExtraTags entry...

aboitreaud: Oops good catch, and agree, let's put them in a new`extra_tags` section.

aboitreaud: I checked, `extra_tags` are treated equally as `tags` in the UI and in the status output.

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -174,27 +175,30 @@ func getJobAndRunIDs() (jobID, runID string, ok bool) {
...
 	}
 	if normalize != nil {
 		value = normalize(value)
 	}
-	setHostTag(s, tagKey, value)
+	setHostTag(s, tagKey, value, valueOnTag)
 }
 
-func setHostTag(s *common.Setup, tagKey, value string) {
+func setHostTag(s *common.Setup, tagKey, value string, valueOnTag bool) {
```
paul-laffon-dd, line 189: Wondering if there might be a more readable way to choose tags to include, as having an extra boolean means that every function calls need to include this new parameter

What do you think of:
- not changing this function (keeping `s.Span.SetTag("host_tag_set."+tagKey, "true")` for all tags)
- manually adding `s.Span.SetTag("host_tag_value."+tagKey, value)` (or a part as a function) for tag values that we explicitly want to have

aboitreaud: True, I agree. I initially chose that boolean approach because the value extraction logic is in `setIfExists` so what you suggest would make us duplicate that logic for all tags that we want to append in clear.

BaptisteFoy (approved):

paul-laffon-dd (approved):

PR #35505, opened 2025-03-26
Restart quality gates on flake

Pythyu:
### What does this PR do?

Restart the `static_quality_gates` job when flaking when getting rpm artifacts


KevinFairise2 (commented):
I think the exit code is not forwarded properly on Gitlab, you might need a hack similar to that: https://github.com/DataDog/datadog-agent/blob/44a1a1670c3ffa62275d062357022d3769e84829/.gitlab/deploy_packages/oci.yml#L19
Might be worth testing it, because I do not think it will ever be restarted in the current state

KevinFairise2 (approved):

spencergilbert (approved):

PR #35522, opened 2025-03-26
[CWS] Add agent tags on custom events

spikat:
### What does this PR do?

It adds agent image tags on custom events (ruleset_loaded, heartbit, selftests, self_tests) for both ebpf and ebpfless.

Example:
image

### DIFF
--- a/pkg/security/module/server.go
+++ b/pkg/security/module/server.go
@@ -231,16 +231,28 @@ func (a *APIServer) dequeue(now time.Time, cb func(msg *pendingMsg) bool) {
 	})
 }
 
-func (a *APIServer) updateMsgTags(msg *api.SecurityEventMessage, includeGlobalTags bool) {
-	// on fargate, append global tags
-	if includeGlobalTags && fargate.IsFargateInstance() {
-		for _, tag := range a.getGlobalTags() {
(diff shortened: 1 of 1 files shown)


pkg/security/module/server.go
```diff
@@ -231,17 +231,29 @@ func (a *APIServer) dequeue(now time.Time, cb func(msg *pendingMsg) bool) {
...
-	// on fargate, append global tags
-	if includeGlobalTags && fargate.IsFargateInstance() {
-		for _, tag := range a.getGlobalTags() {
-			key, _, _ := strings.Cut(tag, ":")
-			if !slices.ContainsFunc(msg.Tags, func(t string) bool {
-				return strings.HasPrefix(t, key+":")
-			}) {
-				msg.Tags = append(msg.Tags, tag)
+func (a *APIServer) updateMsgTags(msg *api.SecurityEventMessage, isCustomEvent bool) {
+	if isCustomEvent {
```
safchain, line 235: I would suggest to remove the isCustomEvent bool and put the if condition when call this function

spikat: suggestion applied

safchain (approved):

PR #35524, opened 2025-03-26
[NPM-4178] Add second goroutine to ParallelTraceroute for ICMP

pimlu:
### What does this PR do?
This PR adds another receive goroutine to the parallel traceroute logic.
### Motivation
We are using raw sockets instead of gopacket, so we'll need to read from two sockets at once (TCP and ICMP).

New tests should pass with race enabled:
```
go test -race -timeout 10s -tags linux,linux_bpf,npm,process,test github.com/DataDog/datadog-agent/pkg/networkpath/traceroute/common -count=1
```

### DIFF
--- a/pkg/networkpath/traceroute/common/traceroute_parallel.go
+++ b/pkg/networkpath/traceroute/common/traceroute_parallel.go
@@ -97,9 +97,17 @@ func TracerouteParallel(ctx context.Context, t TracerouteDriver, p TraceroutePar
 		log.Tracef("found probe %+v", probe)
 		resultsMu.Lock()
 		defer resultsMu.Unlock()
+		previous := results[probe.TTL]
+
 		// packets can get delivered twice - only use the first received probe to avoid overestimating RTT.
 		// this is also important for SACK because SACK traceroute returns the lowest TTL found from ACKs
-		if results[probe.TTL] == nil {
+		shouldUpdate := previous == nil
+		// but also just in case, never let ICMP responses "cover up" actual destination responses
+		if previous != nil && !previous.IsDest && probe.IsDest {
+			shouldUpdate = true
+		}
+
+		if shouldUpdate {
 			results[probe.TTL] = probe
 		}
 	}
@@ -206,7 +214,7 @@ func ToHops(p TracerouteParallelParams, probes []*ProbeResponse) ([]*Hop, error)
 	return hops, nil
 }
 
-var logOnce sync.Once
+var badPktLimit = log.NewLogLimit(10, 5*time.Minute)
 
 // CheckParallelRetryable returns whether ReceiveProbe failed due to a real error or just an irrelevant packet
 func CheckParallelRetryable(funcName string, err error) bool {
@@ -214,9 +222,9 @@ func CheckParallelRetryable(funcName string, err error) bool {
 	if errors.Is(err, ErrReceiveProbeNoPkt) {
 		return true
 	} else if errors.As(err, &badPktErr) {
-		logOnce.Do(func() {
-			log.Warnf("%s() saw a malformed packet (this will only log once): %s", funcName, err)
-		})
+		if badPktLimit.ShouldLog() {
+			log.Warnf("%s() saw a malformed packet: %s", funcName, err)
+		}
 		return true
 	}
 	return false
--- a/pkg/networkpath/traceroute/common/traceroute_parallel.go
+++ b/pkg/networkpath/traceroute/common/traceroute_parallel.go
@@ -44,8 +44,16 @@ type ProbeResponse struct {
 	IsDest bool
 }
 
+// TracerouteDriverInfo is metadata about a TracerouteDriver
+type TracerouteDriverInfo struct {
+	// whether this driver uses a separate socket to read ICMP, and implements ReceiveICMPProbe
+	UsesReceiveICMPProbe bool
+}
+
 // TracerouteDriver is an implementation of traceroute send+receive of packets
 type TracerouteDriver interface {
+	// GetDriverInfo returns metadata about this driver
+	GetDriverInfo() TracerouteDriverInfo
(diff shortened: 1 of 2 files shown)


pkg/networkpath/traceroute/common/traceroute_parallel.go
```diff
@@ -90,38 +131,40 @@ func TracerouteParallel(ctx context.Context, t TracerouteDriver, p TraceroutePar
...
-			if probe.TTL < p.MinTTL || probe.TTL > p.MaxTTL {
-				return fmt.Errorf("received an invalid TTL (expected TTL in [%d, %d]): %d", p.MinTTL, p.MaxTTL, probe.TTL)
-			}
-
-			results[probe.TTL-p.MinTTL] = probe
-			// no need to send more probes if we found the destination
-			if probe.IsDest {
-				writerCancel()
+	handleProbeFunc := func(funcName string, probeFunc func(timeout time.Duration) (*ProbeResponse, error)) {
+		g.Go(func() error {
```
pimlu, line 135: This is mostly an indentation change - now this receive goroutine is created twice

ken-schneider (approved):

pkg/networkpath/traceroute/common/traceroute_parallel.go
```diff
@@ -142,5 +185,39 @@ func TracerouteParallel(ctx context.Context, t TracerouteDriver, p TraceroutePar
...
+
+// CheckParallelRetryable returns whether ReceiveProbe failed due to a real error or just an irrelevant packet
+func CheckParallelRetryable(funcName string, err error) bool {
+	badPktErr := &BadPacketError{}
+	if errors.Is(err, ErrReceiveProbeNoPkt) {
+		return true
+	} else if errors.As(err, &badPktErr) {
+		logOnce.Do(func() {
+			log.Warnf("%s() saw a malformed packet (this will only log once): %s", funcName, err)
+		})
```
AlexandreYang, line 219: Should we use a log limiter/throttler instead? Example here https://github.com/DataDog/datadog-agent/pull/35584/files#diff-f49ca5e1322ed2a691794f41e6695d6fbb5e0a7f1f859476e862a2ded212004aR113

AlexandreYang (approved):

pkg/networkpath/traceroute/common/traceroute_parallel.go
pimlu: Ah that is a good idea, thanks!

pkg/networkpath/traceroute/common/traceroute_parallel.go
```diff
@@ -54,16 +68,51 @@ type TracerouteParallelParams struct {
 	SendDelay time.Duration
 }
 
+// ProbeCount returns the number of probes that will be sent
+func (p TracerouteParallelParams) ProbeCount() int {
+	if p.MinTTL > p.MaxTTL {
+		return 0
```
hmahmood, line 74: When can this happen?

pkg/networkpath/traceroute/common/traceroute_parallel.go
```diff
@@ -90,38 +139,40 @@ func TracerouteParallel(ctx context.Context, t TracerouteDriver, p TraceroutePar
...
-			// no need to send more probes if we found the destination
-			if probe.IsDest {
-				writerCancel()
-			}
-		}
-	})
+		})
+	}
+	handleProbeFunc("ReceiveProbe", t.ReceiveProbe)
+	handleProbeFunc("ReceiveICMPProbe", t.ReceiveICMPProbe)
```
hmahmood, line 175: Do we always do icmp and non-icmp probes (i.e. can one or the other be behind a config)? Isn't this going to double the amount of traffic we send out?

pimlu: > Isn't this going to double the amount of traffic we send out?
That only happens in `SendProbe()` not the receive callbacks. For the read side of things, it means we have to recvmsg from two raw sockets concurrently, but not send more traffic

pkg/networkpath/traceroute/common/traceroute_parallel.go
pimlu: To be honest I don't think it can happen, I just didn't like this potentially returning a negative number. I can remove it

pkg/networkpath/traceroute/common/traceroute_parallel.go
hmahmood: OK, I guess that is for another PR, but should we consider adding configs to disable/enable icmp maybe?

pkg/networkpath/traceroute/common/traceroute_parallel.go
hmahmood: Would that be a misconfiguration? Maybe we should check these values when they are set/passed in.

pimlu: Yeah, it does get checked [at the start of TracerouteParallel](https://github.com/DataDog/datadog-agent/pull/35524/files/c8891a7dd17bc28f92a7c200bb2d1fcf571f4008#diff-c09d0b99bf5241878eb9a6c6d3cabbd08b2b59d496e2a335348986327227266dL59)

pkg/networkpath/traceroute/common/traceroute_parallel.go
pimlu: Yeah this will definitely be needed for windows in the future because the windows packet capture uses a single goroutine.  And it wouldn't be strange to use this with gopacket in the future which doesn't need an ICMP goroutine. I will add a config flag for this

pimlu: I added a bit extra called `TracerouteDriverInfo` which allows the driver to configure this behavior

hmahmood (approved):

PR #35527, opened 2025-03-26
[NPM-4293] Implement parallel SACK traceroute

pimlu:
Stacked on: [\[NPM-4178\] Add second goroutine to ParallelTraceroute for ICMP](https://github.com/DataDog/datadog-agent/pull/35524)

### What does this PR do?
This PR adds a parallel SACK-based traceroute implementation and a portable binary `portable_sack` which runs it standalone.

It's not yet integrated into the system-probe traceroute runner.
### Motivation
SACK traceroute gives higher quality results and is accepted by more firewalls

1. Build:
`go build github.com/DataDog/datadog-agent/pkg/networkpath/traceroute/sack/portable_sack`
2. Run on an EC2 instance (this one's the IP for example.com):
`sudo ./portable_sack 142.250.64.110:443`

I created new parsing logic in `frame_parser.go` which aims to combine TCP/ICMP and hopefully IPv6 in the future into one gopacket entrypoint. However, until I switch over the SYN traceroute implementations to `ParallelTraceroute`, we will have some redundant parsing. I can take this out if you don't think it's a good idea.

### Additional Notes

[NPM-4178]: https://datadoghq.atlassian.net/browse/NPM-4178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/networkpath/traceroute/sack/portable_sack/portable_sack.go
+++ b/pkg/networkpath/traceroute/sack/portable_sack/portable_sack.go
@@ -24,7 +24,7 @@ import (
 func main() {
 	loglevel := os.Getenv("LOG_LEVEL")
 	if loglevel == "" {
-		loglevel = "debug"
+		loglevel = "warn"
 	}
 
 	err := pkglogsetup.SetupLogger(
--- a/pkg/networkpath/traceroute/sack/portable_sack/portable_sack.go
+++ b/pkg/networkpath/traceroute/sack/portable_sack/portable_sack.go
@@ -1,0 +1,76 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-present Datadog, Inc.
+
+// Package main contains a portable binary to easily run SACK traceroutes
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/netip"
+	"os"
+	"time"
+
+	"github.com/DataDog/datadog-agent/pkg/networkpath/traceroute/common"
+	"github.com/DataDog/datadog-agent/pkg/networkpath/traceroute/sack"
+
+	pkgconfigsetup "github.com/DataDog/datadog-agent/pkg/config/setup"
+	pkglogsetup "github.com/DataDog/datadog-agent/pkg/util/log/setup"
+)
+
+func main() {
+	loglevel := os.Getenv("LOG_LEVEL")
+	if loglevel == "" {
+		loglevel = "debug"
+	}
+
+	err := pkglogsetup.SetupLogger(
+		pkglogsetup.LoggerName("sack"),
+		loglevel,
+		"",
+		"",
+		false,
+		true,
+		false,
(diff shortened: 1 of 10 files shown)


hmahmood:
Why the separate binary?

pimlu:
> Why the separate binary?
No real reason besides making it quicker/easier to test. (The dev loop inside system-probe means waiting for startup, curling the unix socket, waiting for shutdown etc.) It could be deleted before merging.

hmahmood:
> > Why the separate binary?
> 
> No real reason besides making it quicker/easier to test. (The dev loop inside system-probe means slower build, waiting for startup, curling the unix socket, waiting for shutdown etc.) It could be deleted before merging.

You could also just write a unit test that only runs if a certain build tag is set, or a certain argument is passed to the test. For example we use this to run "integration" tests in the backend: https://github.com/DataDog/dd-go/blob/prod/testutil/integration/integration.go

AlexandreYang (approved):
Nice !!

pkg/networkpath/traceroute/sack/portable_sack/portable_sack.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/DataDog/datadog-agent/pkg/networkpath/traceroute/sack"
+
+	pkgconfigsetup "github.com/DataDog/datadog-agent/pkg/config/setup"
+	pkglogsetup "github.com/DataDog/datadog-agent/pkg/util/log/setup"
+)
+
+func main() {
+	loglevel := os.Getenv("LOG_LEVEL")
+	if loglevel == "" {
+		loglevel = "debug"
```
AmitaiBl, line 27: so if we do not set that, we are going with debug log level

pkg/networkpath/traceroute/sack/sack_driver.go
```diff
@@ -0,0 +1,403 @@
...
+
+func (s *sackDriver) readAndParse(conn *ipv4.RawConn) error {
+	n, err := conn.Read(s.buffer)
+	if errors.Is(err, os.ErrDeadlineExceeded) {
+		return common.ErrReceiveProbeNoPkt
+	} else if err != nil {
+		return fmt.Errorf("sackDriver failed to ReadFromIP: %w", err)
+	} else if n == 0 {
+		return common.ErrReceiveProbeNoPkt
+	}
```
hmahmood, line 387: nit: the `else`'s are superfluous since all clauses are returning.

hmahmood (approved):

pkg/networkpath/traceroute/sack/sack_driver.go
pimlu: That's true, I actually did it so gofmt saves me a newline 😆 But it's probably a readability improvement so I will update this

pkg/networkpath/traceroute/sack/portable_sack/portable_sack.go
pimlu: Yeah, I'll change this to warn by default actually

PR #35532, opened 2025-03-26
feat: add a serverless no-op trace stats concentrator

apiarian-datadog:
### What does this PR do?
The trace stats concentrator is responsible for computing trace stats in the agent. In the `serverless` and `serverless-init` builds of the agent we actually want to rely on the datadog backend to do the computation. This change overrides the concentrator that we create to be a NoOpConcentrator for serverless environments.

### Motivation
[FRSLES-609](https://datadoghq.atlassian.net/browse/FRSLES-609), [APMS-13109](https://datadoghq.atlassian.net/browse/APMS-13109), and [SLES-6162](https://datadoghq.atlassian.net/browse/SLES-2162) . 

I have deployed a version of this code and confirmed that before the change we were getting exactly doubled trace metrics and wih the change the metrics are no longer doubled. This was helped by grouping the metrics by a second primary tag which is only correctly applied by the backend calculation, not the agent calculation.

[FRSLES-609]: https://datadoghq.atlassian.net/browse/FRSLES-609?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[APMS-13109]: https://datadoghq.atlassian.net/browse/APMS-13109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/trace/stats/concentrator_noop.go
```diff
@@ -0,0 +1,15 @@
...
+package stats
+
+
+type NoOpConcentrator struct {}
+
+func (n *NoOpConcentrator) Start() { }
+
+func (n *NoOpConcentrator) Stop() { }
+
+func (n *NoOpConcentrator) Add(t Input) { }
```
apiarian-datadog, line 15: ultimately https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/agent/agent.go#L428 in the agent Process method is what we're overriding here. maybe there's a better way to do this? perhaps by checking some sort of config in addition to the `p.ClientComputedStats` further up that Process method?

PR #35538, opened 2025-03-27
Exclude files by path always

jose-manuel-almaza:
### What does this PR do?
Always exclude configured paths from copyright checking

### Motivation
When the list of files to be processed is received, the exclusion by path is not considered, which is considered if the list is not received and has to be created.

Touch the file `internal/third_party/golang/expansion/expand.go` and then run:

`inv linter.copyrights --only-staged-files --debug`
`inv linter.copyrights --debug`

And checking that file is excluded in all cases

### Additional Notes

This is a clear example where our tests are coupled to the implementation, and simply fixing a bug without changing the requirements means we have to change the tests.


chouetz (approved):
Thanks for this contribution!

PR #35551, opened 2025-03-27
[NDMII-3391] deduplicate devices during autodiscovery

jedupau:
### What does this PR do?
This PR duplicates devices during SNMP autodiscovery based on three things:
- sysName
- sysOID
- sysDescr
- sysUptime

To do this, when a device is discovered we add the infos of the device in a list of pending services.
When an IP is fully scanned, we flush this list of pending services by activating all possible services.
A service can be activated only if all lower IPs have been scanned.
This new logic lives in a new `snmp_deduper` file

### Motivation
Some devices can answer to several IPs and it creates duplicated devices

I made a simulation on CML with a router responding to two IPs and another one responding to only one IP.
This new version of the code scheduled me 2 checks instead of one


urseberry (approved):

comp/core/autodiscovery/listeners/snmp.go
```diff
@@ -54,14 +59,82 @@ const (
...
+	services                        map[string]*SNMPService
+	devicesFoundByFullDeviceHash    map[string]DeviceHashInfo
+	fullDeviceHashByBasicDeviceHash map[string][]string
+	ipsCounter                      IPCounter
+	pendingServicesByFullDeviceHash map[string]*pendingService
+}
+
+type ipAuthenticationCounter struct {
+	sync.RWMutex
+	counter map[string]int
```
FlorianVeaux, line 95: What about using `sync.Map` here ?

FlorianVeaux (changes requested):

comp/core/autodiscovery/listeners/snmp.go
```diff
@@ -54,14 +59,82 @@ const (
...
+type IPCounter interface {
+	Inc(ip string)
+	Dec(ip string)
+	Get(ip string) int
+	Set(ip string, count int)
+	Len() int
+	GetAll() map[string]int
+}
+
+type DeviceHashInfo struct {
```
FlorianVeaux, line 72: It doesn't look like we ever read Name, Description or IP.

I feel like we should be able to keep this more streamlined with a simple interface like:

```golang
interface {
    AddDevice(Name, Description, IP, BootTimeMs)
    Contains(Name, Description, IP, BootTimeMs)
}
```

which can be backed by either a simple list of struct (but then contains is a for-loop, likely ok but not optimal). Or a map with "basicHash" as keys, and a list of `{Name, Description, BootTimeMs, IP}` as the values

comp/core/autodiscovery/listeners/snmp.go
```diff
@@ -54,14 +59,82 @@ const (
...
-	services   map[string]*SNMPService
+	newService                      chan<- Service
+	delService                      chan<- Service
+	stop                            chan bool
+	config                          snmp.ListenerConfig
+	services                        map[string]*SNMPService
+	devicesFoundByFullDeviceHash    map[string]DeviceHashInfo
+	fullDeviceHashByBasicDeviceHash map[string][]string
+	ipsCounter                      IPCounter
+	pendingServicesByFullDeviceHash map[string]*pendingService
```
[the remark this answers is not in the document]
jedupau: yes indeed, I did a complete refacto of this to have most of the logic in a separate file

comp/core/autodiscovery/listeners/snmp.go
jedupau: changed to use it !

comp/core/autodiscovery/listeners/snmp_deduper.go
```diff
@@ -0,0 +1,188 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-present Datadog, Inc.
+
+package listeners
```
FlorianVeaux, line 6: This could benefit from being it's own lib/package, separated from the listener so that not everything is accessible and we can focus on the API between the two regardless of the implementation

comp/core/autodiscovery/listeners/snmp_deduper.go
```diff
@@ -0,0 +1,188 @@
...
+import (
+	"math"
+	"net"
+	"sync"
+
+	"github.com/DataDog/datadog-agent/pkg/snmp"
+	"github.com/DataDog/datadog-agent/pkg/util/log"
+)
+
+const uptimeDiffTolerance = 50
```
FlorianVeaux, line 17: - Add the unit, like 50 * time.Milliseconds
- 50ms seems a tiny bit small to be. What about 5 seconds ?

comp/core/autodiscovery/listeners/snmp_deduper.go
```diff
@@ -0,0 +1,188 @@
...
+
+type deviceInfo struct {
+	Name        string
+	Description string
+	BootTimeMs  int64
+	IP          string
```
FlorianVeaux, line 23: This doesn't seem needed to be in this struct.

comp/core/autodiscovery/listeners/snmp_deduper.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func (d *deviceDeduperImpl) addDevice(device deviceInfo) {
+	d.Lock()
+	defer d.Unlock()
+	d.devices = append(d.devices, device)
+}
+
+func (d *deviceDeduperImpl) contains(device deviceInfo) bool {
+	d.Lock()
```
FlorianVeaux, line 109: This could be a RLock

comp/core/autodiscovery/listeners/snmp_deduper.go
```diff
@@ -0,0 +1,188 @@
...
+func newDeviceDeduper(config snmp.ListenerConfig) deviceDeduper {
+	deduper := &deviceDeduperImpl{
+		devices:         make([]deviceInfo, 0),
+		pendingServices: make([]pendingService, 0),
+	}
+	deduper.initIPCounter(config)
+	return deduper
+}
+
+func (d *deviceDeduperImpl) initIPCounter(config snmp.ListenerConfig) {
```
FlorianVeaux, line 68: Can we move this in `newDeviceDeduper` as this is not thread-safe ?

comp/core/autodiscovery/listeners/snmp_deduper.go
```diff
@@ -0,0 +1,188 @@
...
+		}
+	}
+	d.pendingServices = newPendingServices
+	return svcToActivate
+}
+
+func (d *deviceDeduperImpl) removeIP(ip string) bool {
+	d.Lock()
+	defer d.Unlock()
+	count, _ := d.ipsCounter.Load(ip)
```
FlorianVeaux, line 166: Can we use a map of string to atomic.UInt32 here ? This could help remove the lock here.
We can use the Uint32.Add function https://pkg.go.dev/sync/atomic#Uint32.Add which returns a copy of the new value, and use that new value for comparing to zero

FlorianVeaux: Another option would be to continue storing int, but using atomic.AddUint32 which acts directly on integer fields https://pkg.go.dev/sync/atomic#AddUint32

FlorianVeaux (changes requested):

comp/core/autodiscovery/listeners/snmp.go
```diff
@@ -218,48 +215,42 @@ func (l *SNMPListener) checkDevice(job snmpJob) {
 	autodiscoveryStatusBySubnetVar.Set(GetSubnetVarKey(job.subnet.config.Network, job.subnet.cacheKey), &autodiscoveryStatus)
 }
 
-func (l *SNMPListener) checkDeviceForParams(params *gosnmp.GoSNMP, deviceIP string) bool {
+func (l *SNMPListener) checkDeviceInfo(authentication snmp.Authentication, port uint16, deviceIP string, skipGetNext bool) (bool, devicededuper.DeviceInfo) {
```
FlorianVeaux, line 218: Should it be (devicededuper.DeviceInfo, error) instead and leverage err != nil?

FlorianVeaux (approved):

TCheruy (commented):
i like that this fits in 2 files 👏

comp/core/autodiscovery/listeners/snmp.go
jedupau: as discussed, I splited the method in two 
77848cd7b22d04f06b499f016fb98cca79fe3ceb

comp/core/autodiscovery/listeners/snmp.go
```diff
@@ -211,28 +215,55 @@ func (l *SNMPListener) checkDevice(job snmpJob) {
...
-	if len(value.Variables) < 1 || value.Variables[0].Value == nil {
-		log.Debugf("SNMP get to %s no data", deviceIP)
-		return false
+	if len(value.Variables) < 4 || value.Variables[0].Value == nil || value.Variables[1].Value == nil || value.Variables[2].Value == nil || value.Variables[3].Value == nil {
+		return true, devicededuper.DeviceInfo{}
 	}
 
-	log.Debugf("SNMP get to %s success: %v", deviceIP, value.Variables[0].Value)
-	return true
+	sysName := string(value.Variables[0].Value.([]byte))
```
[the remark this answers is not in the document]
jedupau: good catch
77848cd7b22d04f06b499f016fb98cca79fe3ceb

pkg/snmp/devicededuper/device_deduper.go
```diff
@@ -0,0 +1,206 @@
...
+		}
+	}
+
+	return deduper
+}
+
+func (d *deviceDeduperImpl) checkPreviousIPs(deviceIP string) bool {
+	previousIPsDiscovered := true
+
+	for ip, counter := range d.ipsCounter {
```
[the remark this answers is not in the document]
jedupau: indeed the PR might be risky, as discussed IRL I added a commit to gate the feature behind a config 
f765a1f53088eb0afe8cdcf04ecc31b46df3bb0a

TCheruy (approved):

dustmop (approved):

PR #35559, opened 2025-03-27
Change FailNow() in Fail() to make sure the error is displayed in the panic

KevinFairise2:
### What does this PR do?

Call Fail() instead of FailNow() otherwise the execution of the function is interrupted and panic(err) is never called, preventing displaying the actual error. 
The panic(err) will then interrupt execution anyway

### DIFF
--- a/test/new-e2e/pkg/e2e/suite.go
+++ b/test/new-e2e/pkg/e2e/suite.go
@@ -250,7 +250,7 @@ func (bs *BaseSuite[Env]) UpdateEnv(newProvisioners ...provisioners.Provisioner)
 		targetProvisioners[provisioner.ID()] = provisioner
 	}
 	if err := bs.reconcileEnv(targetProvisioners); err != nil {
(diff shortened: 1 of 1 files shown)


test/new-e2e/pkg/e2e/suite.go
```diff
@@ -584,7 +584,7 @@ func (bs *BaseSuite[Env]) BeforeTest(string, string) {
 	// In `Test` scope we can `panic`, it will be recovered and `AfterTest` will be called.
 	// Next tests will be called as well
 	if err := bs.reconcileEnv(bs.originalProvisioners); err != nil {
-		bs.T().FailNow() // We need to call FailNow otherwise bs.T().Failed() will be false in AfterTest
+		bs.T().Fail() // We need to call Failotherwise bs.T().Failed() will be false in AfterTest
```
pgimalac, line 587:
```suggestion
		bs.T().Fail() // We need to call Fail otherwise bs.T().Failed() will be false in AfterTest
```

pgimalac (approved):

chouetz (approved):

PR #30266, opened 2024-10-18
[clusteragent] Remove the CPU/memory limits by default for the init container

iamluc:
### What does this PR do?

+ Set the CPU/memory limits and requests for the init container to the maximum that will be used for the pod
+ Block injection if the memory would have been too low and cause OOMs on init containers

### Motivation

The current limits are too low for the SSI init container.
Customer pods will usually allocate a lot more than we are requiring. Since init containers are run sequentially, adding limits/requirements set to what the pod would have needed wihtout us doesn't change the way it is scheduled and gives us a lot more of resources to work with (in most cases at least). 

### Describe how to test/QA your changes

How to test:
* Enable SSI on the cluster agent
Case 1:
* Deploy a new pod with combinations of init-containers/sidecar containers/containers.
* Verify that the resource limits and requests on injected init containers are set to the effective request/limit for the pod

Case 2:
* Set auto_instrumentation.init_resources.(cpu|memory)
* Verify that the value is applied to injected init container resources, regardless of the pod limits and requests

Case 3:
* Deploy a pod with a container with a very small memory footprint (<100Mi)
* Verify that we are not injecting anything into it

### Additional Notes

Associated documentation PR: https://github.com/DataDog/documentation/pull/25873

PR #35588, opened 2025-03-28
[clusteragent] Fix wrong computation of the init container resources

### LINKED ISSUE #35170: [BUG] Limit for autoinstrumentation init container calculated wrong
**Agent Environment**
```
Agent 7.63.3 - Commit: eab078f2e7 - Serialization version: v5.0.141 - Go version: go1.23.5
```

**Describe what happened:**
Datadog webhook adds init containers with invalid resource spec. CPU limit can be smaller than request which is not allowed. This happens if there are multiple container in a pod where the sum all limits is smaller than the sum of all requests (e.g. because on container does not have a limit at all).

Datadog Admission controller does not create invalid spec.

Create a pod with two containers like this:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: demo
spec:
  template:
    spec:
      containers:
        - name: container1
          resources:
            limits:
              cpu: 1000m
              memory: 512Mi
            requests:
              cpu: 1000m
              memory: 512Mi
        - name: container2
          resources:
            limits:
              memory: 128Mi
            requests:
              cpu: 50m
              memory: 128Mi
```

GKE.

The code that does this calculation wrong is probably here: https://github.com/DataDog/datadog-agent/blob/main/pkg/clusteragent/admission/mutate/autoinstrumentation/auto_instrumentation.go#L288

Seems to have been introduced with #30266

iamluc:
### What does this PR do?

Fix wrong computation of the init container resources, introduced in https://github.com/DataDog/datadog-agent/pull/30266

If a request was set but not the limit, the computed request for the init containers could be higher than the computed limit. This is invalid and causes the deployment to fail.

To fix that, in case there is a request but no limit, we use the request as the limit.

### Motivation

Fix https://github.com/DataDog/datadog-agent/issues/35170 / https://datadoghq.atlassian.net/browse/CONS-7154 / https://datadoghq.atlassian.net/browse/INPLAT-529

Unit test
+
manually using
```
# values.yaml
datadog:
  logLevel: DEBUG
  clusterName: my-test-cluster
  logs:
    enabled: true
    containerCollectAll: true
    containerCollectUsingFiles: true
  kubelet:
    tlsVerify: false
  apm:
    instrumentation:
      enabled: true
clusterAgent:
  admissionController:
    enabled: true


# deploy.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: demo
  namespace: init
  labels:
    app: init-container-test
spec:
  replicas: 1
  selector:
    matchLabels:
      app: init-container-test
  template:
    metadata:
      labels:
        app: init-container-test
    spec:
      containers:
        - name: container1
          image: busybox
          command: ["sleep", "3600"]
          resources:
            limits:
              cpu: 100m
              memory: 128Mi
            requests:
              cpu: 100m
              memory: 128Mi
        - name: container2
          image: busybox
          command: ["sleep", "3600"]
          resources:
            limits:
              # cpu: 50m    <- uncommenting this allows the pod to be created
              memory: 128Mi
            requests:
              cpu: 50m
              memory: 128Mi
```

With cluster agent 7.64, the deployment of the app results in a failure:
```
Pod "demo-57667dd664-q8h7z" is invalid: [spec.initContainers[0].resources.requests: Invalid value: "150m": must be less than or equal to cpu limit of 100m, spec.initContainers[1].resources.requests: Invalid value: "150m": must be less than or equal to cpu limit of 100m,
```

With this PR, the deployment is successful.

### DIFF
--- a/pkg/clusteragent/admission/mutate/autoinstrumentation/auto_instrumentation.go
+++ b/pkg/clusteragent/admission/mutate/autoinstrumentation/auto_instrumentation.go
@@ -293,6 +293,8 @@ func podSumRessourceRequirements(pod *corev1.Pod) corev1.ResourceRequirements {
 
 	for _, k := range [2]corev1.ResourceName{corev1.ResourceMemory, corev1.ResourceCPU} {
 		// Take max(initContainer ressource)
+		maxInitContainerLimit := resource.Quantity{}
+		maxInitContainerRequest := resource.Quantity{}
 		for i := range pod.Spec.InitContainers {
 			c := &pod.Spec.InitContainers[i]
 			if initContainerIsSidecar(c) {
@@ -300,45 +302,28 @@ func podSumRessourceRequirements(pod *corev1.Pod) corev1.ResourceRequirements {
 				// we need to add it's resources to the main container's resources
 				continue
 			}
-
-			request, requestFound := c.Resources.Requests[k]
-			limit, limitFound := c.Resources.Limits[k]
-
-			if requestFound {
-				maxRequest := ressourceRequirement.Requests[k]
-				if request.Cmp(maxRequest) == 1 {
-					ressourceRequirement.Requests[k] = request
+			if limit, ok := c.Resources.Limits[k]; ok {
+				if limit.Cmp(maxInitContainerLimit) == 1 {
+					maxInitContainerLimit = limit
 				}
 			}
-
-			if limitFound || requestFound {
-				if !limitFound {
-					limit = request
-				}
-				maxLimit := ressourceRequirement.Limits[k]
-				if limit.Cmp(maxLimit) == 1 {
-					ressourceRequirement.Limits[k] = limit
+			if request, ok := c.Resources.Requests[k]; ok {
+				if request.Cmp(maxInitContainerRequest) == 1 {
+					maxInitContainerRequest = request
 				}
 			}
 		}
 
+		// Take sum(container resources) + sum(sidecar containers)
 		limitSum := resource.Quantity{}
 		reqSum := resource.Quantity{}
 		for i := range pod.Spec.Containers {
 			c := &pod.Spec.Containers[i]
-
-			request, requestFound := c.Resources.Requests[k]
-			limit, limitFound := c.Resources.Limits[k]
-
-			if requestFound {
-				reqSum.Add(request)
+			if l, ok := c.Resources.Limits[k]; ok {
+				limitSum.Add(l)
 			}
-
-			if requestFound || limitFound {
-				if !limitFound {
-					limit = request
-				}
-				limitSum.Add(limit)
+			if l, ok := c.Resources.Requests[k]; ok {
+				reqSum.Add(l)
 			}
 		}
 		for i := range pod.Spec.InitContainers {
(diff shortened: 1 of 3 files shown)


releasenotes/notes/clusteragent-fix-resources-computation-0b59c193205483a5.yaml
```diff
@@ -0,0 +1,4 @@
+---
+fixes:
+  - |
```
adel121, line 3: this should go under releasenotes-dca

michaelcretzman (approved):

pkg/clusteragent/admission/mutate/autoinstrumentation/auto_instrumentation.go
```diff
@@ -300,16 +300,24 @@ func podSumRessourceRequirements(pod *corev1.Pod) corev1.ResourceRequirements {
...
-				if limit.Cmp(existing) == 1 {
-					ressourceRequirement.Limits[k] = limit
+
+			request, requestFound := c.Resources.Requests[k]
+			limit, limitFound := c.Resources.Limits[k]
+
+			if requestFound {
+				maxRequest := ressourceRequirement.Requests[k]
+				if request.Cmp(maxRequest) == 1 {
+					ressourceRequirement.Requests[k] = request
```
paullegranddc, line 310: The comment is not on this block of code specifiquely, but rather than putting all these branches during computation, would it not be easier to compute requests and limits independently, and then at the end do:
```
if  limitSum != 0 && reqSum > limitSum {
  reqSum = limitSum
}
```

This way we're sure that we never have requests > limits, and don't have to look at every individual containers.
Because to be honest, I struggle already enough to read this code as it is, and I wrote it....

iamluc: Updated with a similar suggestion that also covers init containers. I've also tried to improve readability a bit.
WDYT?

paullegranddc (approved):

pkg/clusteragent/admission/mutate/autoinstrumentation/auto_instrumentation.go
paullegranddc: Perfect 👌

adel121 (approved):
LGTM

Thanks for adding a QA section. 🙇 

* Would be good to include the expected outcome in the QA to make it easier during the QA process later. 
* I also removed `team/container-platform` label from the labels because it will generate a QA card for us too, while this should be owned only by #injection-platform.
* Additionally, `qa/done` means that there is no need to QA during the release process once this is merged on main. I think this should still be QA'ed after being merged on main. For this we need the label `qa/rc-required`. But I will leave it to your to decide on this.

iamluc:
Thanks!

> * Would be good to include the expected outcome in the QA to make it easier during the QA process later.

QA section updated

> * Additionally, `qa/done` means that there is no need to QA during the release process once this is merged on main. I think this should still be QA'ed after being merged on main. For this we need the label `qa/rc-required`. But I will leave it to your to decide on this.

Sounds good. I changed `qa/done` to `qa/rc-required`.

PR #35594, opened 2025-03-28
Ignore go.mod files for tests and lints in pre-commit

pgimalac:
### What does this PR do?
Ignore go.mod and go.sum files to determine which tests and lints to run in pre-commit.

### Motivation
Before this change, any dependency update runs a lot of tests and lints in pre-commit, which is really inconvenient.

### Additional Notes

`get_modified_packages` is only used by the `test` and `linter.go` tasks when running in pre-commits. 
It shouldn't have any impact on CI.

### DIFF
--- a/tasks/gotest.py
+++ b/tasks/gotest.py
@@ -446,6 +446,9 @@ def get_modified_packages(ctx, build_tags=None, lint=False) -> list[GoModule]:
     go_mod_modified_modules = set()
 
     for modified_file in modified_go_files:
+        if modified_file.endswith(".mod") or modified_file.endswith(".sum"):
+            continue
+
         best_module_path = Path(get_go_module(modified_file))
 
         # Check if the package is in the target list of the module we want to test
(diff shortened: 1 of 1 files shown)


tasks/gotest.py
```diff
@@ -437,7 +437,9 @@ def e2e_tests(ctx, target="gitlab", agent_image="", dca_image="", argo_workflow=
 def get_modified_packages(ctx, build_tags=None, lint=False) -> list[GoModule]:
     modified_files = get_go_modified_files(ctx)
 
-    modified_go_files = [f"./{file}" for file in modified_files]
+    modified_go_files = [
```
KevinFairise2, line 440: You should rather remove that part I think:
https://github.com/DataDog/datadog-agent/pull/35594/files#diff-701454bcbc727fba4131e72d5be303fd60c33ce65d815abfdd6e816e871188edR471-R475

pgimalac: Yep good catch !

KevinFairise2 (approved):

PR #35612, opened 2025-03-28
[E2E test] k8s test with DCA leader election

zhuminyi:
### What does this PR do?
Adds an E2E test to verify the leader election status using the `cluster-agent status --json` output.

1. Get leader pod name from any dca pod
2. Connect to leader pod
3. Use leader pod for the rest of E2E tests on dca status check. 

### Motivation
The current [integration test](https://github.com/DataDog/datadog-agent/blob/e0c94ee83df20740e1289b7bb757fe5a247a1e9a/test/integration/util/leaderelection/leaderelection_test.go#L255) for leader election is being deprecated. This PR introduces a new E2E test to validate the leader election status.

Currently, the test infrastructure only launches a single Cluster Agent pod, which will always assume leadership. In the next version of test-infra-definitions, the number of Cluster Agent pods will increase to two. This allows us to retrieve the leader pod name from any DCA pod status and then verify that pod's leadership status.

Note: Validation of re-election is not supported in either the integration test or the new E2E test, as triggering re-election would interfere with other parts of the E2E suite.

We don't support validation on reelection in integration test or new e2e test yet. Doing so will cause other part of e2e test fail. 

E2e test

### DIFF
--- a/test/new-e2e/tests/containers/k8s_test.go
+++ b/test/new-e2e/tests/containers/k8s_test.go
@@ -388,8 +388,11 @@ func (suite *k8sSuite) testClusterAgentCLI() {
 	suite.Require().NoError(err)
 	suite.Require().Len(pod.Items, 1)
 
+	leaderDcaPodName := suite.testDCALeaderElection(false) //check cluster agent leaderelection without restart
+	suite.Require().NotEmpty(leaderDcaPodName, "Leader DCA pod name should not be empty")
+
 	suite.Run("cluster-agent status", func() {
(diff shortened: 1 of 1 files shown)


test/new-e2e/tests/containers/k8s_test.go
```diff
@@ -388,8 +388,11 @@ func (suite *k8sSuite) testClusterAgentCLI() {
 	suite.Require().NoError(err)
```
gabedos, line 388: Nit: Can we remove this redundancy since we are already testing that a LinuxClusterAgent exists in `suite.testDCALeaderElection`

zhuminyi: Yes, I totally agree. This part is no longer needed.

gabedos (approved):

PR #35615, opened 2025-03-28
[CWS] CWS-4206 Add DNS response event / NXDOMAIN

kovagsm:
### What does this PR do?

This PR allows the user the possibility to capture an entire DNS packet and create and event based on it, with the entire process and network context. The user can filter based on the DNS response_code, can be especially interesting when wishing to capture a DNS response that returned an NXDOMAIN, but can be used for all the response codes.

In order to be more efficient, a discarder mechanism was added.

* Unit and manual tests


pkg/security/secl/model/model.go
```diff
@@ -548,19 +548,29 @@ type ExitEvent struct {
...
-	Name  string `field:"question.name,opts:length" op_override:"eval.CaseInsensitiveCmp"` // SECLDoc[question.name] Definition:`the queried domain name`
-	Type  uint16 `field:"question.type"`                                                   // SECLDoc[question.type] Definition:`a two octet code which specifies the DNS question type` Constants:`DNS qtypes`
-	Class uint16 `field:"question.class"`                                                  // SECLDoc[question.class] Definition:`the class looked up by the DNS question` Constants:`DNS qclasses`
-	Size  uint16 `field:"question.length"`                                                 // SECLDoc[question.length] Definition:`the total DNS request size in bytes`
-	Count uint16 `field:"question.count"`                                                  // SECLDoc[question.count] Definition:`the total count of questions in the DNS request`
+	ID       uint16      `field:"id"`       // SECLDoc[id] Definition:`[Experimental] the DNS request ID`
+	Question DNSQuestion `field:"question"` // SECLDoc[id] Definition:`DNS Question`
```
YoannGh, line 562: You don't need to add a SECLDoc comment on `question` fields here because these aren't not terminal fields.

docs/cloud-workload-security/linux_expressions.md
```diff
@@ -38,6 +38,7 @@ Triggers are events that correspond to types of activity seen by the system. The
 | `chown` | File | A file’s owner was changed | 7.27 |
 | `connect` | Network | A connect was executed | 7.60 |
 | `dns` | Network | A DNS request was sent | 7.36 |
+| `dns_response` | Network | A DNS response was received | 7.6x |
```
safchain, line 41: could we use `dns.response` instead of dns_response ? to have a more consistent syntax. In case of a response both dns.question and dns.response will be filled :thinking:

jhgilbert (approved):

pkg/security/secl/model/unmarshallers_linux.go
```diff
@@ -12,6 +12,7 @@ import (
 	"encoding/binary"
 	"encoding/json"
 	"fmt"
+	"github.com/google/gopacket/layers"
```
safchain, line 15: nit: move this line around line 23

pkg/security/tests/dns_response_test.go
```diff
@@ -0,0 +1,127 @@
...
+	"net"
+	"os"
+	"strconv"
+	"testing"
+	"time"
+)
+
+const (
+	OK       = uint8(0)
+	NXDOMAIN = uint8(3)
```
safchain, line 27: I would expose the error code in SECL using constants, like `dns.reponse.code == NXDOMAIN`

kovagsm: Addressed

pkg/security/secl/model/unmarshallers_linux.go
kovagsm: Addressed

docs/cloud-workload-security/linux_expressions.md
```diff
@@ -2891,6 +2892,13 @@ Constants: [DNS qtypes](#dns-qtypes)
 
 
 
+### `dns.response.response_code` {#dns-response-response_code-doc}
```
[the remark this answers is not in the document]
kovagsm: Addressed

safchain (approved):

PR #35617, opened 2025-03-28
use glibc buildimage for custom clang

brycekahle:
### What does this PR do?

Uses the glibc buildimages to build the custom version of clang we use for ebpf compilation

### Motivation

Standardizing on a buildimage

Ran pipelines of full KMT suite using this new built version of clang.

### Additional Notes

Added versioning to the built clang, so we don't accidentally clobber anything. Currently doesn't force a `.` separator, so that it is backwards compatible with the current filenames. A future PR will update to using only the versioned files.


spencergilbert (approved):

gjulianm (approved):
LGTM, but I'd rebase on `main` before merging to ensure that https://github.com/DataDog/datadog-agent/pull/35549 doesn't break this PR

brycekahle:
> to ensure that #35549 doesn't break this PR

This PR doesn't use the new built clang yet. I'll do that in a future PR, so I'll make sure it has that change and doesn't break.

ofek (approved):
Thanks for updating the other tags as well, looks great to me!

PR #35619, opened 2025-03-28
Ensure arguments are properly parsed through `dda inv`

ofek:
### Motivation

I became paranoid since [this PR](https://github.com/DataDog/datadog-agent/pull/35596) earlier today so this changes only commands that are executed within the CI and anything user facing (code comments, command output, etc.) does not use `--`.


ofek:
Going to merge since this is a no-op.

PR #34800, opened 2025-03-05
Add loadbalancingexporter to the default package

liustanley:
### What does this PR do?
This PR adds the [loadbalancingexporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/loadbalancingexporter) to the default list of OTel Agent components. It also adds a new e2e test to verify that traces and metrics get routed to the same backend if they belong to the same service.

The loadbalancingexporter only allows log routing by trace id and the calendar go examples doesn't have log trace correlation yet. So logs will be added to the e2e test in a separate PR.

### Motivation
OTAGENT-181

### Describe how you validated your changes

E2E tests

PR #35622, opened 2025-03-28
Add e2e tests for loadbalancingexporter logs

liustanley:
### What does this PR do?
Follow up to https://github.com/DataDog/datadog-agent/pull/34800 to add logs tests.


test/new-e2e/tests/otel/utils/pipelines_utils.go
```diff
@@ -499,11 +500,48 @@ func getLoadBalancingSpans(t assert.TestingT, traces []*aggregator.TracePayload)
...
+	log1 := logs1[0]
+	log1Tags := getTagMapFromSlice(c, log1.Tags)
+	attrs := make(map[string]interface{})
+	err = json.Unmarshal([]byte(log1.Message), &attrs)
+	require.NoError(c, err)
+	for k, v := range attrs {
+		log1Tags[k] = fmt.Sprint(v)
+	}
+
+	logs2, err := s.Env().FakeIntake.Client().FilterLogs(service, fakeintake.WithMessageContaining(log2Body))
```
songy23, line 520: QQ: log2Body is a sub-string of log1Body, so if you filter on log2Body would it be possible to still get log1 rather than log2?

truthbk: Yup, I think that would be the case. Good catch.

liustanley: True, just updated to use `WithMessageMatching` instead

test/new-e2e/tests/otel/utils/pipelines_utils.go
```diff
@@ -499,11 +494,36 @@ func getLoadBalancingSpans(t assert.TestingT, traces []*aggregator.TracePayload)
...
-	assert.Equal(t, 12, spans)
+	require.Equal(t, 12, spans)
 	return spanMap
 }
 
-func getLoadBalancingMetrics(t assert.TestingT, metrics []*aggregator.MetricSeries) map[string][]map[string]string {
+func getLoadBalancingLogs(c require.TestingT, s OTelTestSuite, service string) {
+	logs1, err := s.Env().FakeIntake.Client().FilterLogs(service, fakeintake.WithMessageMatching(log1Body))
+	require.NoError(c, err)
+	require.NotEmpty(c, logs1)
```
songy23, line 504: should this be `require.Len(c, logs1, 1)` ?

test/new-e2e/tests/otel/utils/pipelines_utils.go
```diff
@@ -499,11 +494,36 @@ func getLoadBalancingSpans(t assert.TestingT, traces []*aggregator.TracePayload)
...
+	log1Tags := getLogTags(c, log1)
+
+	logs2, err := s.Env().FakeIntake.Client().FilterLogs(service, fakeintake.WithMessageMatching(log2Body))
+	require.NoError(c, err)
+	require.NotEmpty(c, logs2)
+	matchedLog := false
+	for _, log2 := range logs2 {
+		// Find second log with the same trace id
+		log2Tags := getLogTags(c, log2)
+		if log1Tags["dd.trace_id"] == log2Tags["dd.trace_id"] {
```
songy23, line 515: I think logs2 can still contains log1 so you may be just getting log1 here, so you may want something like
`if log1.Message != log2.Message && log1Tags["dd.trace_id"] == log2Tags["dd.trace_id"]`

test/new-e2e/tests/otel/utils/pipelines_utils.go
```diff
@@ -783,6 +807,17 @@ func getContainerTags(t *testing.T, tp *trace.TracerPayload) (map[string]string,
...
 }
 
+func getLogTags(t require.TestingT, log *aggregator.Log) map[string]string {
+	tags := getTagMapFromSlice(t, log.Tags)
+	attrs := make(map[string]interface{})
+	err := json.Unmarshal([]byte(log.Message), &attrs)
+	require.NoError(t, err)
+	for k, v := range attrs {
+		tags[k] = fmt.Sprint(v)
+	}
```
songy23, line 817: Just to make sure: the func is called getLogTags but here is looping through the log message?

test/new-e2e/tests/otel/utils/pipelines_utils.go
liustanley: I tried this at first but since there are lots of logs being generated, `FilterLogs` will always return more than 1 log. We only need the first log here so it's sufficient to check `NotEmpty`.

test/new-e2e/tests/otel/utils/pipelines_utils.go
liustanley: Good point, just renamed to `getLogTagsAndAttrs`

songy23 (approved):
LGTM

truthbk (approved):

PR #35625, opened 2025-03-28
[AGNTLOG-9] Add compression config for different pipelines 

DDuongNguyen:
### What does this PR do?
This PR makes several improvements to compression settings and logging across different agent pipelines:

1. Changes the default log agent compression to zstd for better performance and reduced bandwidth usage

2. Improves logging clarity for compression settings by modifying the compressionKind() method to clearly indicate the source of compression settings:
- Pipeline-specific configuration (e.g., service_discovery.forwarder.compression_kind)
- Global logs configuration (logs_config.compression_kind)
- Default fallback settings

3. Ensures backward compatibility by automatically falling back to gzip compression when additional endpoints are configured

4. Fixes the event platform forwarder to use correct pipeline-specific compression settings instead of log endpoint settings

5. Sets zstd as the default compression for non-log pipelines unless configured otherwise

### Motivation
This change improves observability and debugging of compression settings across different agent pipelines. 

By providing clearer logging messages, operators can better understand which compression settings are being used and where they're coming from. This helps prevent confusion when different pipelines might be using different compression.

**Test 1:**
1. Turn on agent debug logging
2. Configure different compression settings:
   - Set `logs_config.compression_kind` to a value (e.g., "zstd")
   - Set pipeline-specific compression settings (e.g., `service_discovery.forwarder.compression_kind`)
3. Verify the debug logs show the correct information:
   ```
   "Pipeline {prefix} compression settings - configured: {value}"
   "Agent is using configured compression: {value}"
   ```
Note: The agent by default is using logs_config.compression_kind inside the config.yaml (currently zstd), and it will take a while before new config is applied to it so its normal to see in the log the agent going from zstd to gzip or vice versa depending on the configured vs default value

**Test 2:** 
1. Turn on agent debug logging
2. Configure different compression settings:
   - Set `logs_config.compression_kind` to zstd
3. Configure a logs_config.additional_endpoints 
4. Verify the debug logs show the correct information:
```
   "Pipeline {prefix} compression settings - configured: zstd"
   "Agent is using configured compression: gzip"
```

### Additional Notes
- Epforwarder and the intake team approves that all events are supported:
image
image


comp/forwarder/eventplatform/eventplatformimpl/epforwarder.go
```diff
@@ -50,6 +50,10 @@ const (
 	eventTypeDBMMetrics  = "dbm-metrics"
 	eventTypeDBMActivity = "dbm-activity"
 	eventTypeDBMMetadata = "dbm-metadata"
+	gzipCompressionKind  = "gzip"
+	gzipCompressionLevel = 6
+	zstdCompressionKind  = "gzip"
```
gh123man, line 55:
```suggestion
	zstdCompressionKind  = "zstd"
```

comp/forwarder/eventplatform/eventplatformimpl/epforwarder.go
```diff
@@ -123,6 +135,8 @@ var passthroughPipelineDescs = []passthroughPipelineDesc{
 		defaultBatchMaxContentSize:    pkgconfigsetup.DefaultBatchMaxContentSize,
 		defaultBatchMaxSize:           pkgconfigsetup.DefaultBatchMaxSize,
 		defaultInputChanSize:          pkgconfigsetup.DefaultInputChanSize,
+		compressionKind:               "gzip",
```
gh123man, line 138:
```suggestion
		compressionKind:               gzipCompressionKind,
```
Same for below?

comp/logs/agent/config/config_keys.go
```diff
@@ -110,15 +110,22 @@ func (l *LogsConfigKeys) devModeUseProto() bool {
...
-		log.Debugf("Logs agent is using: %s compression", compressionKind)
+	configKey := l.getConfigKey("compression_kind")
+	compressionKind := l.getConfig().GetString(configKey)
+
+	if l.prefix != "logs_config." {
+		pipelineCompressionKind := l.getConfig().GetString(l.prefix + "compression_kind")
+		log.Debugf("Pipeline %s compression settings - configured: %s", l.prefix, pipelineCompressionKind)
+	}
+
+	if compressionKind == "zstd" || compressionKind == "gzip" {
```
gh123man, line 121: NIT does it make sense to use the compression kind constants here/ 
```suggestion
	if compressionKind == zstdCompressionKind || compressionKind == gzipCompressionKind {
```

gh123man (approved):

gh123man:
As we discussed offline - let's keep the default gzip for now unitl we change the agent wide default.

joepeeples (approved):

hush-hush (approved):

comp/forwarder/eventplatform/eventplatformimpl/epforwarder.go
```diff
@@ -149,7 +149,7 @@ var passthroughPipelineDescs = []passthroughPipelineDesc{
 		// Each NetFlow flow is about 500 bytes
 		// 10k BatchMaxSize is about 5Mo of content size
 		defaultBatchMaxSize: 10000,
-
```
gh123man, line 152: NIT: Can revert this file - no code changes.

comp/logs/agent/config/config_keys.go
```diff
@@ -110,23 +119,39 @@ func (l *LogsConfigKeys) devModeUseProto() bool {
...
 func (l *LogsConfigKeys) compressionKind() string {
-	compressionKind := l.getConfig().GetString(l.getConfigKey("compression_kind"))
-	switch compressionKind {
-	case "zstd", "gzip":
-		log.Debugf("Logs agent is using: %s compression", compressionKind)
+	// Check if additional endpoints are configured
+	endpoints, _ := l.getAdditionalEndpoints()
+	if len(endpoints) > 0 {
+		log.Debugf("Additional endpoints detected, pipeline: %s falling back to gzip compression for compatibility", l.prefix)
+		return GzipCompressionKind
```
gh123man, line 126: I think we still want to use `compression_kind` here - but only if it's explicitly set (ex not the default). 
[You can do this with `IsConfigured`](https://github.com/DataDog/datadog-agent/blob/05e00bf3dd1da372b71d9d422056d09c43c572c4/pkg/config/model/types.go#L174). 

This way (for example in org2) we can still use zstd with dual shipping if we explicitly opt-in to it.

gh123man: So the control flow would be something like:

```
if IsConfigured("compression_kind"):
    return config.Get("compression_kind")
else:
    if using addtional_endpoints:
        return gzip
    else:
        return config.Get("compression_kind")
```

gh123man: Ideally let's also add a UT to cover this specific case.

gh123man (approved):

PR #35634, opened 2025-03-31
usm: tests: Cleanup protocol map before running the tests

guyarb:
### What does this PR do?

Clenaup entries from `connection_protocol` map before running each classification test

### Motivation

Ensure no "leaks" between tests, as we run a single monitor/tracer for all classification tests

### DIFF
--- a/pkg/network/usm/tests/tracer_usm_linux_test.go
+++ b/pkg/network/usm/tests/tracer_usm_linux_test.go
@@ -1155,6 +1155,12 @@ func testMySQLProtocolClassificationInner(t *testing.T, tr *tracer.Tracer, clien
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			originalPostTracer := tt.postTracerSetup
+			wrapperPostTracer := func(t *testing.T, ctx testContext) {
+				cleanProtocolMapByProtocol(t, tr, protocols.MySQL)
+				originalPostTracer(t, ctx)
+			}
+			tt.postTracerSetup = wrapperPostTracer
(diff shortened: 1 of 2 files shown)


pkg/network/usm/tests/tracer_usm_linux_test.go
```diff
@@ -1155,6 +1155,12 @@ func testMySQLProtocolClassificationInner(t *testing.T, tr *tracer.Tracer, clien
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			originalPostTracer := tt.postTracerSetup
```
amitslavin, line 1158: Can we initialize it once for all the tests at the beginning of the test?

pkg/network/usm/tests/tracer_usm_linux_test.go
```diff
@@ -1401,6 +1407,12 @@ func testPostgresProtocolClassification(t *testing.T, tr *tracer.Tracer, clientH
 				serverAddress: serverAddress,
 				extras:        make(map[string]interface{}),
 			}
+			originalPostTracer := tt.postTracerSetup
+			wrapperPostTracer := func(t *testing.T, ctx testContext) {
+				cleanProtocolMapByProtocol(t, tr, protocols.Postgres)
+				originalPostTracer(t, ctx)
+			}
```
amitslavin, line 1414: Can we pass a `wrapperPostTracer` to the test itself, which will accept the protocol type and create it for each test?

amitslavin (changes requested):

amitslavin (approved):

PR #35641, opened 2025-03-31
[OTAGENT-331] Stop DDOT if otelCollector.enabled is false

ogaca-dd:
### What does this PR do?
Stop DDOT if `otelCollector.enabled` is false

Set `otelCollector.enabled` at false and make sure DDOT is not run in the full image.


truthbk (approved):
🍰

truthbk:
Some e2e tests seem to be failing, perhaps because they are not properly setting the `otelCollector.enabled` option to true prior to launching the test, just assuming the collector would be there.

truthbk (approved):

PR #35644, opened 2025-03-31
SUSM-94: Handle Same-Node K8s Service

DanielLavie:
### What does this PR do?
- Reorders connection key callback sequence to prioritize direct connection matches before NAT translations
- Adds test case `TestKubernetesLocalNATScenario` that reproduces and validates the fix for SUSM-94
- Adds detailed debug tracing capabilities for connection handling when trace logging is enabled, so we can use it for future customer investigations

### Motivation
In SUSM-94, a customer reported an issue where in a K8s environment, when a client communicates with a K8s service that NATs traffic to a server on the same node, USM suffered from accuracy loss. We reproduced the issue locally on a K8s cluster and noticed the following issue:

image

NPM captures 2 connections:

1. **Connection 1** (Pre-NAT, client's perspective):
   - Source: 172.29.161.37:53792 (Client)
   - Dest: 10.100.103.122:7778 (K8s Service)
   - Direction: OUTGOING
   - PID: 1784258 (client process)
   - IPTranslation: Contains post-NAT info mapping to Client->Server

2. **Connection 2** (Post-NAT, server's perspective):
   - Source: 172.29.191.94:7777 (Server)
   - Dest: 172.29.161.37:53792 (Client)
   - Direction: INCOMING
   - PID: 334392 (server process)
   - No IPTranslation

USM captures 2 connection keys:
- **Key 1**: Client->K8s Service (pre-NAT)
- **Key 2**: Client->Server (post-NAT, normalized)

**Before this change**, the same USM aggregation (client -> server) would match both NPM connections because: 
- For Connection 1, it would check IPTranslation first and match with Key 2
- For Connection 2, it would also match with Key 2 

Because we have a PID filtering mechanism that prevents matching the same USM connection to different PIDs, we would drop the second matching entirely, resulting in accuracy loss.

**The change introduced in this PR** fixes this issue by prioritizing direct connection matches (Source and Dest fields) before NAT translations. This means:
- Connection 1 now correctly matches with Key 1 (Client->K8s Service)
- Connection 2 correctly matches with Key 2 (Client->Server)

This ensures both connections retain their appropriate USM data.

- Added `TestKubernetesLocalNATScenario` that exactly reproduces the customer scenario from SUSM-94:
  * Client (172.29.161.37:53792) -> K8s Service (10.100.103.122:7778) -> Server (172.29.191.94:7777)
  * The test validates both pre-NAT and post-NAT connection handling
- Manually tested in a Kubernetes environment that matches the customer's setup
- All existing tests continue to pass
- Tested the change in the [entei](https://ddstaging.datadoghq.com/dashboard/kfn-zy2-t98/agent-system-probe-overview?fromUser=false&refresh_mode=paused&tpl_var_kube_cluster_name%5B0%5D=entei&from_ts=1744559253393&to_ts=1744704602000&live=false&tile_focus=8822127419025042) and [parent12](https://ddstaging.datadoghq.com/dashboard/kfn-zy2-t98/agent-system-probe-overview?fromUser=false&refresh_mode=paused&tpl_var_kube_cluster_name%5B0%5D=parent12&from_ts=1744542825486&to_ts=1744704824000&live=false&tile_focus=8822127419025042) staging clusters, observed no degradation in the orphan connections and aggregations metrics.
- Tested Haproxy scenario using the load test ([results](https://dddev.datadoghq.com/dashboard/gqu-2wz-asy/usm-performance-evaluation-20?fromUser=true&refresh_mode=paused&tpl_var_base_agent-env%5B0%5D=unclaimed-http-haproxy-base&tpl_var_client-service%5B0%5D=golang-k6-client-http&tpl_var_compare_agent-env%5B0%5D=unclaimed-http-haproxy-new&tpl_var_http.version%5B0%5D=http%2F1.1&tpl_var_kube_cluster_name%5B0%5D=usm2&tpl_var_server-service%5B0%5D=golang-httpbin&from_ts=1745136330736&to_ts=1745148478000&live=false))


cmd/system-probe/modules/network_tracer.go
```diff
@@ -378,3 +383,57 @@ func writeConntrackTable(table *tracer.DebugConntrackTable, w http.ResponseWrite
...
+					traceLog("    Dynamic Tags: %v", stat.DynamicTags)
+				}
+			}
+		}
+	} else {
+		traceLog("No HTTP data found in connections")
+	}
+
+	// Log all connections for debugging
+	traceLog("Found %d total connections", len(cs.Conns))
```
amitslavin, line 417: what is the difference between this log and ling 394?

cmd/system-probe/modules/network_tracer.go
```diff
@@ -378,3 +383,57 @@ func writeConntrackTable(table *tracer.DebugConntrackTable, w http.ResponseWrite
...
+	for i, conn := range cs.Conns {
+		traceLog("Connection %d:", i)
+		traceLog("  Source: %s:%d", conn.Source.String(), conn.SPort)
+		traceLog("  Destination: %s:%d", conn.Dest.String(), conn.DPort)
+		traceLog("  Protocol Stack: %+v", conn.ProtocolStack)
```
amitslavin, line 422: Any reason why not logging all the data in a same traceLog function?

cmd/system-probe/modules/network_tracer.go
```diff
@@ -378,3 +383,57 @@ func writeConntrackTable(table *tracer.DebugConntrackTable, w http.ResponseWrite
...
+		if conn.IPTranslation != nil {
+			traceLog("  IP Translation:")
+			traceLog("    ReplSrcIP: %s, ReplSrcPort: %d", conn.IPTranslation.ReplSrcIP.String(), conn.IPTranslation.ReplSrcPort)
+			traceLog("    ReplDstIP: %s, ReplDstPort: %d", conn.IPTranslation.ReplDstIP.String(), conn.IPTranslation.ReplDstPort)
+		} else {
+			traceLog("  No IP Translation")
+		}
+		traceLog("  Direction: %s", conn.Direction)
```
amitslavin, line 430: same as above

pkg/network/encoding/marshal/usm_http_test.go
```diff
@@ -307,6 +310,201 @@ func TestLocalhostScenario(t *testing.T) {
...
+
+	// Create IP addresses and ports
+	clientIP := util.AddressFromString("172.29.161.37")
+	k8sServiceIP := util.AddressFromString("10.100.103.122") // K8s Service IP (ClusterIP)
+	serverIP := util.AddressFromString("172.29.191.94")
+	clientPort := uint16(53792)
+	servicePort := uint16(7778)
+	serverPort := uint16(7777)
+
+	// Create both connections exactly as seen from SUSM-94 reproduction environment
```
amitslavin, line 329: Not sure the comment with SUSM-94 is relevant  let's explain the purpose without referencing the ticket number

pkg/network/encoding/marshal/usm_http_test.go
```diff
@@ -307,6 +310,201 @@ func TestLocalhostScenario(t *testing.T) {
...
+			// No IPTranslation for this connection
+		},
+	}
+
+	// Create HTTP stats with a sample request - 4 requests as seen in logs
+	httpStats := http.NewRequestStats()
+	httpStats.AddRequest(200, 15.0, 0, nil)
+	httpStats.AddRequest(200, 15.0, 0, nil)
+	httpStats.AddRequest(200, 15.0, 0, nil)
+	httpStats.AddRequest(200, 15.0, 0, nil)
```
amitslavin, line 372: maybe create a loop for that with a const

pkg/network/encoding/marshal/usm_http_test.go
```diff
@@ -307,6 +310,201 @@ func TestLocalhostScenario(t *testing.T) {
...
+	// will always be in the saddr field (this is what we've seen when reproducing SUSM-94)
+	postNATKey := http.NewKey(
+		clientIP,   // Client is still the source in the HTTP key (172.29.161.37)
+		serverIP,   // Server is the destination (172.29.191.94)
+		clientPort, // Client port (53792)
+		serverPort, // Server port (7777)
+		[]byte("/delay/5"),
+		true,
+		http.MethodGet,
+	)
```
amitslavin, line 396: If the only difference between the `preNATKey` and the `postNATKey` is the `serverIP`, let's structure it more cleanly by creating a `http.newKey` and updating just that field.

Plus IMO the comments are not really relevant

pkg/network/encoding/marshal/usm_http_test.go
```diff
@@ -307,6 +310,201 @@ func TestLocalhostScenario(t *testing.T) {
...
+				Conns: connections,
+			},
+			HTTP: httpData,
+		}
+
+		// Create HTTP encoder
+		httpEncoder := newHTTPEncoder(payload.HTTP)
+
+		// Test post-NAT connection (server → client) - should have HTTP data
+		aggregations, _, _ := getHTTPAggregations(t, httpEncoder, payload.Conns[1])
```
amitslavin, line 493: shouldn't you validate that payload.Conns > 0?

pkg/network/encoding/marshal/usm_http_test.go
```diff
@@ -307,6 +310,201 @@ func TestLocalhostScenario(t *testing.T) {
...
+		assert.NotNil(aggregations)
+		assert.Equal("/delay/5", aggregations.EndpointAggregations[0].Path)
```
amitslavin, line 495: shouldn't you validate that aggregations.EndpointAggregations > 0?

cmd/system-probe/modules/network_tracer.go
```diff
@@ -378,3 +383,57 @@ func writeConntrackTable(table *tracer.DebugConntrackTable, w http.ResponseWrite
 		w.WriteHeader(500)
 	}
 }
+
+// traceLogConnections logs detailed connection information to help investigate NPM/USM
+// customer issues by providing visibility into network connections and their metadata.
+func traceLogConnections(id string, cs *network.Connections) {
```
hmahmood, line 389: I think this can result in an enormous number of logs. I have also not really had a customer send us trace level logs; usually we ask for debug level logs. 

There are already endpoints like `/connections` and others for USM that we can use to get similar info. We can add them to the agent flare if we need them from customers; this seems like a better alternative to me than logs.

amitslavin (changes requested):

amitslavin:
@DanielLavie, I understand the problem, but I'm not sure how the proposed change solves it. The description isn't fully clear to me on how the fix addresses the issue. I think that the description needs to be updated.

cmd/system-probe/modules/network_tracer.go
DanielLavie: 1. "I think this can result in an enormous number of logs" – This should only be printed when trace logs are enabled, meaning we’re actively investigating a customer issue and need detailed context.
I'm also seeing [this](https://github.com/DataDog/datadog-agent/blob/ec797eabcea52b0f3b009f4a98242689054b9819/cmd/system-probe/modules/network_tracer.go#L366), which suggests we may already be emitting a high volume of trace logs.

2. "I have also not really had a customer send us trace level logs; usually we ask for debug level logs" – Why is that? It makes sense to begin with debug logs, but in cases like this one where more granularity is required, escalating to trace level should be an option.

3. "There are already endpoints like /connections and others for USM that we can use to get similar info" – That’s valid, but logs offer full historical context, while /connections only reflects the most recent 30 seconds or so. Logs also simplify correlation: they allow us to directly compare the NPM connection tuple with the USM connection key and quickly identify mismatches, without needing to parse aggregated HTTP data.

hmahmood: I am not sure how useful this will be even if we somehow manage to get these logs from the customer. This could be a pretty large number of logs and parsing through them would be challenging to say the least. The endpoints will give you JSON, which is parse-able and query-able with tools like `jq`. Another problem would be that this obscures other trace logs since the context from this one log is going to be much larger than other trace logs.

Content like this is more appropriate for the flare. We were already considering adding this, but de-prioritized it; maybe this can be incentive enough to resurrect it.

DanielLavie: The issue with the /connections endpoint is that it doesn’t include all the HTTP connection key details—it relies on writeConnections, which is where the underlying problem was hidden. As a result, we wouldn’t observe the issue when using this endpoint.

We could try using the HTTP debug endpoint instead, but it only returns the USM portion of the HTTP data, without NPM context, and it consumes all connections in the process - making it unusable for our needs.

So at this point, I’m not exactly sure how the endpoints could help us retrieve the necessary information.

hmahmood: A combination of the two doesn't give the same results? You could also modify or add another endpoint. Another option could be to do more limited trace logs at the point where you think the problem can happen.

guyarb, line 389: One small suggestion as enabling trace logs is going to spam system-probe - 
Either add a prefix to the logs, or maybe change the feature enabling the trace logs, as you might not be able to see only your trace logs when needed
I do generally recommend (unless you've already done that) to simulate the scenario in which system-probe is enabled with trace logs, and ensure we can easily obtain the information we need

pkg/network/encoding/marshal/usm_http_test.go
guyarb, line 1: I'm surprised (maybe disappointed) that not a single existing test was broken due to your change
How can we guarantee any other change/regression will be detected?
Can you try and "tweak" the code to ensure your new test is covering the cases?

cmd/system-probe/modules/network_tracer.go
guyarb: @DanielLavie, eventually, will we need a snapshot of the logs (say, the last iteration) or a massive dump (i.e. an hour)?

pkg/network/encoding/marshal/usm_http_test.go
```diff
@@ -307,6 +310,201 @@ func TestLocalhostScenario(t *testing.T) {
 	assert.Equal(uint32(1), aggregations.EndpointAggregations[0].StatsByStatusCode[int32(103)].Count)
 }
 
+// TestKubernetesLocalNATScenario tests how USMConnectionIndex handles Kubernetes-style local NAT connections
+// with both pre-NAT and post-NAT connections as seen in Kubernetes environments
+func TestKubernetesLocalNATScenario(t *testing.T) {
+	if runtime.GOOS == "windows" || os.Getenv("CI") == "true" {
+		t.Skip("Skipping test on Windows or CI")
+	}
```
guyarb, line 318: Consider moving the test to a new file with `linux` go build on it (as this test might run on macos)
also, why do we skip CI?

pkg/network/usm_connection_keys.go
```diff
@@ -75,13 +75,13 @@ func WithKey(connectionStats ConnectionStats, f func(key types.ConnectionKey) (s
 		clientIPNAT, clientPortNAT, serverIPNAT, serverPortNAT = serverIPNAT, serverPortNAT, clientIPNAT, clientPortNAT
 	}
 
-	// Callback 1: NATed (client, server)
-	if hasNAT && f(types.NewConnectionKey(clientIPNAT, serverIPNAT, clientPortNAT, serverPortNAT)) {
+	// Callback 1: (client, server)
+	if f(types.NewConnectionKey(clientIP, serverIP, clientPort, serverPort)) {
```
amitslavin, line 79: I suggest adding a comment here explaining the reason for the change, to help prevent future breaking changes

amitslavin (changes requested):

cmd/system-probe/modules/network_tracer.go
DanielLavie: @guyarb a snapshot should suffice

pkg/network/encoding/marshal/usm_http_test.go
DanielLavie: The `TestKubernetesLocalNATScenario` initially failed. Only when I made the modification in this PR, it passed

PR #35646, opened 2025-03-31
[AGTMETRICS-16]API Key refresh for metric endpoints

StephenWakely:
### What does this PR do?

Implements API key refresh for metric endpoints.

### Motivation

Allows the agent to update the API keys used if an API key returned by a secret changes.

Have manually tested the key specified by the main endpoint, `additional_endpoints`, `multi_region_failover` and `vector` endpoints, with the following config:

```yaml
api_key: ENC[mainkee]
dd_url: http://localhost:8080

additional_endpoints:
  "http://localhost:9081":
  - ENC[kee1]
  "http://localhost:9082":
  - ENC[kee2]

multi_region_failover:
  api_key: ENC[kee3]
  dd_url: http://localhost:9081
  site: datad0g.com
  enabled: true
 
vector:
  metrics:
    enabled: true
    url: http://localhost:8900

secret_backend_command: /home/stephenwakely/src/secret-resolver.py
secret_refresh_interval: 10
```

I hope to write E2E tests in a follow up PR.


comp/forwarder/defaultforwarder/default_forwarder.go
```diff
@@ -358,20 +358,6 @@ func NewDefaultForwarder(config config.Component, log log.Component, options *Op
 		}
 	}
 
-	config.OnUpdate(func(setting string, oldValue, newValue any) {
```
StephenWakely, line 361: `OnUpdate` is now handled in each domain resolver.

releasenotes/notes/metrics_api_key_refresh-6e4a0f95099b5ac5.yaml
```diff
@@ -0,0 +1,13 @@
...
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+enhancements:
+  - |
+    Metric endpoints (main endpoint, ``additional_endpoints``,
+    ``multi_region_failover`` and ``vector``) will refresh the API
```
urseberry, line 12:
```suggestion
    ``multi_region_failover`` and ``vector``) refresh the API
```

urseberry (approved):

pkg/status/endpoints/status.go
```diff
@@ -23,10 +23,11 @@ func PopulateStatus(stats map[string]interface{}) {
 		return
 	}
 
+	deduped := utils.DedupEndpoints(endpoints)
 	endpointsInfos := make(map[string]interface{})
 
 	// obfuscate the api keys
-	for endpoint, keys := range endpoints {
+	for endpoint, keys := range deduped {
 		for i, key := range keys {
```
vickenty, line 31: With this and change in `domain_resolver.go` we would only need one dedup function.

```suggestion
	for endpoint, eps := range endpoints {
		for i, key := range utils.DedupEndpoint(eps) {
```

pkg/diagnose/connectivity/core_endpoint.go
```diff
@@ -71,7 +71,8 @@ func getLogsUseTCP() bool {
 func Diagnose(diagCfg diagnose.Config, log log.Component) []diagnose.Diagnosis {
 
 	// Create domain resolvers
-	keysPerDomain, err := utils.GetMultipleEndpoints(pkgconfigsetup.Datadog())
+	config := pkgconfigsetup.Datadog()
+	keysPerDomain, err := utils.GetMultipleEndpoints(config)
```
vickenty, line 75: Doesn't seem like anything has changed here. Would it make sense to remove this hunk?

comp/forwarder/defaultforwarder/default_forwarder.go
```diff
@@ -122,20 +121,31 @@ func ToggleFeature(features, flag Features) Features { return features ^ flag }
...
 			vectorMetricsURL,
 		)
 	}
+
+	// Add a hook into the config for each resolver to track API key changes.
+	// Do this after the OP resolver may have replaced the original resolver so we don't leak a callback in the config.
+	for domain := range resolvers {
+		resolver := resolvers[domain]
+		pkgresolver.OnUpdateConfig(resolver, log, config)
+		resolvers[domain] = resolver
```
vickenty, line 146: `resolver` value can not change in this segment of code, there should be no need to re-assign it back into the map.

pkg/config/utils/endpoints.go
```diff
@@ -90,29 +88,98 @@ func GetMainEndpointBackwardCompatible(c pkgconfigmodel.Reader, prefix string, d
...
+		trimmed := []string{}
+		for _, key := range keys {
+			trimmedAPIKey := strings.TrimSpace(key)
+			if trimmedAPIKey != "" {
+				trimmed = append(trimmed, trimmedAPIKey)
+			}
+		}
+
+		if len(trimmed) > 0 {
+			result[url] = []Endpoint{{
```
vickenty, line 108: Would it make sense to use `NewEndpoint` constructor here?

pkg/config/utils/endpoints.go
```diff
@@ -90,29 +88,98 @@ func GetMainEndpointBackwardCompatible(c pkgconfigmodel.Reader, prefix string, d
...
 		return nil, fmt.Errorf("could not parse main endpoint: %s", err)
 	}
 
-	keysPerDomain := map[string][]string{
-		ddURL: {
-			c.GetString("api_key"),
-		},
+	keysPerDomain := map[string][]Endpoint{
+		ddURL: {{
+			ConfigSettingPath: "api_key",
```
vickenty, line 166: Would it make sense to use `NewEndpoint` here and also at line 179?

pkg/config/utils/endpoints.go
```diff
@@ -90,29 +88,98 @@ func GetMainEndpointBackwardCompatible(c pkgconfigmodel.Reader, prefix string, d
...
+			if trimmedAPIKey != "" {
+				trimmed = append(trimmed, trimmedAPIKey)
+			}
+		}
+
+		if len(trimmed) > 0 {
+			result[url] = []Endpoint{{
+				ConfigSettingPath: root,
+				APIKeys:           trimmed,
+			}}
```
vickenty, line 111: Would it make sense to keep the info message that we skipped a url due to an empty api key (the one on old line 71).

vickenty (commented):
Left a couple of questions, but overall looks good.

pkg/config/utils/endpoints.go
StephenWakely: I hesitated to do so because the function parameters for `NewEndpoint` is just a bunch of strings. That function was mainly useful for tests. Creating the struct directly here allows us to use the struct field names as documentation.

comp/forwarder/defaultforwarder/resolver/domain_resolver.go
```diff
@@ -33,38 +37,98 @@ type DomainResolver interface {
...
+				break
+			}
+		}
+
+		if !found {
+			return
+		}
+
+		if setting == "additional_endpoints" {
+			// Updating additional endpoints don't give us the exact key that has been updated so we reload the whole config section.
```
daniel-taf, line 84: The live process checks also make use of the default forwarder and want to leverage it for api key refreshing, but our `additional_endpoints` config is different (`process_config.additional_endpoints`) so this won't work for us. It seems like we can easily add support here for it just by checking if the setting contains `additional_endpoints` as a substring. Can we do that here? If it out of scope, it can be done in another PR. I would mostly like to know if that approach makes sense.

StephenWakely: Yes, good catch. I have updated it to check it contains `additional_endpoints`.

StephenWakely: It still doesn't seem to be working correctly for `process_config`. Am investigating...

StephenWakely: Ok. This should be good now.

comp/forwarder/defaultforwarder/resolver/domain_resolver.go
```diff
@@ -33,38 +38,98 @@ type DomainResolver interface {
 	// Resolve returns the domain to be used to send data for a given `transaction.Endpoint` along with a
 	// destination type
 	Resolve(endpoint transaction.Endpoint) (string, DestinationType)
+	// GetEndpoints returns the list of endpoints associated with this `DomainResolver`
+	GetEndpoints() []utils.APIKeys
```
hush-hush, line 42: This doesn't returns endpoints but the API keys used by this resolver, no ? Should we rename this to `GetAPIKeysInfo` or something similar ?

comp/forwarder/defaultforwarder/resolver/domain_resolver.go
```diff
@@ -146,6 +235,18 @@ func NewMultiDomainResolver(baseDomain string, apiKeys []string) *MultiDomainRes
...
 // GetAPIKeys returns the slice of API keys associated with this SingleDomainResolver
 func (r *MultiDomainResolver) GetAPIKeys() []string {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	return r.dedupedAPIKeys
+}
+
+// SetAPIKeys updates the deduped list of API keys associated with this `DomainResolver`
+func (r *MultiDomainResolver) SetAPIKeys(keys []string) {
+	r.dedupedAPIKeys = keys
```
hush-hush, line 245: Same here, should we lock ?

comp/forwarder/defaultforwarder/resolver/domain_resolver.go
```diff
@@ -33,38 +38,98 @@ type DomainResolver interface {
...
+			additionalEndpoints := utils.MakeEndpoints(config.GetStringMapStringSlice(setting), setting)
+			endpoints, ok := additionalEndpoints[resolver.GetBaseDomain()]
+			if !ok {
+				log.Errorf("error: the domain in additional_endpoints changed at runtime for '%s', discarding update.", resolver.GetBaseDomain())
+				return
+			}
+			keys := utils.DedupAPIKeys(endpoints)
+			resolver.SetAPIKeys(keys)
+
+			log.Infof("rotating API key for '%s'", setting)
```
hush-hush, line 95: Could we print the new old and new api keys like we do line 107 ? I think it could be very useful when troubleshooting.

comp/forwarder/defaultforwarder/resolver/domain_resolver.go
```diff
@@ -33,38 +38,98 @@ type DomainResolver interface {
...
+			log.Infof("rotating API key for '%s': %s -> %s",
+				setting,
+				scrubber.HideKeyExceptLastFiveChars(oldAPIKey),
+				scrubber.HideKeyExceptLastFiveChars(newAPIKey),
+			)
+
+			return
+		}
+
+		log.Errorf("new API key for '%s' is invalid (not a string) ignoring new value", setting)
```
hush-hush, line 114: We could also hit this case if the old value was not a string (like nil for example).
Should we error out if the old API key is invalid ?

hush-hush (changes requested):

daniel-taf (approved):

comp/forwarder/defaultforwarder/resolver/domain_resolver.go
StephenWakely: I've added more logging to capture if old and new has changed. 

I'm not sure if it would be wise to just force set a key if the old is invalid as there can be multiple API keys set and without  knowing the old value we won't know if it is a new api key or a changed one. It would be best to revisit when we know how things will work in future..

hush-hush (approved):

duncanista (approved):

comp/forwarder/defaultforwarder/resolver/domain_resolver.go
```diff
@@ -84,6 +162,42 @@ func (r *SingleDomainResolver) GetBaseDomain() string {
...
+	defer r.mu.Unlock()
+	return r.dedupedAPIKeys
+}
+
+// missing returns a list of elements that are in list a, but not in list b.
+// This is inefficient for large lists, but the assumption is that a config
+// will only have a very small number of API keys specified.
+// NOTE, this scrubs the API key to avoid leaking the key when logging.
+func missing(a []string, b []string) []string {
+	added := []string{}
```
vickenty, line 175: Added or missing?

comp/forwarder/defaultforwarder/resolver/domain_resolver.go
```diff
@@ -33,38 +38,111 @@ type DomainResolver interface {
 	// Resolve returns the domain to be used to send data for a given `transaction.Endpoint` along with a
 	// destination type
 	Resolve(endpoint transaction.Endpoint) (string, DestinationType)
+	// GetAPIKeysInfo returns the list of API Keys and config paths associated with this `DomainResolver`
+	GetAPIKeysInfo() []utils.APIKeys
 	// GetAPIKeys returns the list of API Keys associated with this `DomainResolver`
 	GetAPIKeys() []string
+	// SetAPIKeys updates the deduped list of API keys associated with this `DomainResolver`
+	// Returns the list of keys that have changed - the returned keys are scrubbed as these will be used to log
+	// which keys have been changed.
```
vickenty, line 47: Not a huge deal, but I think this is suboptimal abstraction: the resolver shouldn't be tasked with preparing api keys for logging if it isn't doing the logging itself, and it causes code the same bit of code to be duplicated that could be just written once inline with the logging call once.

vickenty (approved):

PR #35651, opened 2025-03-31
[NPM-4279] Add socket filter to linux TCP traceroutes

pimlu:
### What does this PR do?
This PR adds a socket filter to linux TCP traceroutes, which should reduce the amount of packets we capture to userspace.

### Motivation
We want to reduce system-probe CPU usage of dynamic path

`sudo go test -tags test,linux,linux_bpf github.com/DataDog/datadog-agent/pkg/networkpath/traceroute/filter -count 10`

TCP traceroutes should still work in e2e tests:
```
aws-vault exec sso-agent-sandbox-account-admin -- inv new-e2e-tests.run --targets=tests/netpath --run=TestFakeTracerouteSuite
```

When the SOCK_RAW socket is first created, it has no filter and will still capture all packets for a brief moment, before the filter gets attached. There's truly no way around this besides holding onto SOCK_RAW sockets in a pool, which I don't think is worth the complexity.

### DIFF
--- a/pkg/networkpath/traceroute/filter/attach_linux.go
+++ b/pkg/networkpath/traceroute/filter/attach_linux.go
@@ -1,0 +1,84 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-present Datadog, Inc.
+
+//go:build linux
+
+package filter
+
+import (
+	"errors"
+	"fmt"
+	"math"
+	"syscall"
+	"unsafe"
+
+	"golang.org/x/net/bpf"
+	"golang.org/x/sys/unix"
+)
+
+// SetBPF attaches a BPF filter to the underlying socket
+func SetBPF(c syscall.RawConn, filter []bpf.RawInstruction) error {
+	var p unix.SockFprog
+	if len(filter) > math.MaxUint16 {
+		return errors.New("filter too large")
+	}
+	p.Len = uint16(len(filter))
+	p.Filter = (*unix.SockFilter)(unsafe.Pointer(&filter[0]))
+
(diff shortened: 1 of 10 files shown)


pkg/networkpath/traceroute/filter/attach_linux.go
```diff
@@ -0,0 +1,84 @@
...
+	"math"
+	"syscall"
+	"unsafe"
+
+	"golang.org/x/net/bpf"
+	"golang.org/x/sys/unix"
+)
+
+// SetBPF attaches a BPF filter to the underlying socket
+func SetBPF(c syscall.RawConn, filter []bpf.RawInstruction) error {
```
pimlu, line 22: This ought to be combined with `(*Socket).SetBPF` in `pkg/network/netlink/socket.go` but I'm not sure where to put it lol... Any ideas for places that would be a good shared location?

hmahmood (approved):

pkg/networkpath/traceroute/filter/attach_nolinux.go
```diff
@@ -0,0 +1,24 @@
...
+package filter
+
+import (
+	"syscall"
+
+	"golang.org/x/net/bpf"
+)
+
+// SetBPF is not supported
+func SetBPF(_c syscall.RawConn, _filter []bpf.RawInstruction) error {
```
brycekahle, line 17: Is this used somewhere without the `linux` build tag? I didn't see a usage.

brycekahle (approved):

pkg/networkpath/traceroute/filter/attach_nolinux.go
pimlu: Currently it's not supported but in the future we might try to support TCP in `traceroute_darwin.go` -- if so, it will pull in `tcpv4_unix.go` and this will be relevant

PR #35657, opened 2025-03-31
[test/integration] Delete docker listener test

davidor:
### What does this PR do?

This PR removes an old integration test as part of the ongoing effort to migrate legacy tests to the new framework.

I think that the deleted test can be safely removed because its coverage is now redundant.

More specifically, the deleted test:
- Starts some docker containers
- Instantiates a container listener manually
- Checks that the expected autodiscovery services associated with the containers deployed are there

The ECS tests that use the new framework (https://github.com/DataDog/datadog-agent/blob/main/test/new-e2e/tests/containers/ecs_test.go) test the full pipeline, including metrics for autodiscovery checks, without directly instantiating components, so they're more complete.

The removed test also included a check for excluded containers. While we already have some coverage for that, it wasn’t in the autodiscovery listeners, so I added a test to cover that scenario for completeness.

Tests in CI should pass.


clamoriniere (approved):

CelianR (approved):
Thanks!

PR #29435, opened 2024-09-18
Refactor ReceiveResourceSpans to share code with otelSpanToDDSpan fro…

IbraheemA:
…m Concentrator

### What does this PR do?

JIRA Link: https://datadoghq.atlassian.net/browse/OTEL-1726

Refactor `ReceiveResourceSpans` such that it uses the same mapping logic as `otelSpanToDDSpan` from `pkg/trace/stats/otel_util.go`  . Also leverage other helpers from `pkg/trace/traceutil/otel_util.go`. 

Moved “otelSpanToDDSpan” to new `trace/transform` module because putting it in `traceutil` module resulted in circular import between `trace/config` and `trace/traceutil` 

Add a new feature flag `enable_receive_resource_spans_v2` under `apm_config.features` to gate this refactor.

Removed functionality compared to the old `ReceiveResourceSpans` :

- No longer populate these fields in `tagstats`:
    
    `LangVersion, Interpreter, LangVendor`

- No longer check for `lang` and `containerID` in HTTP header

- No longer check for resource-related values (container, env, hostname) in span attributes. We had support for it before, but that behaviour doesn't follow OTel spec

Added functionality:

- Leverage `o.conf.Ignore` to ignore spans with specified resource names

### Motivation

The goals of this refactor were to make the code more readable, and to make the logic for setting `span.Resource` and `span.Type` easier to change in a future PR.

### Describe how to test/QA your changes

Add `enable_receive_resource_spans_v2` flag in apm_config.features in datadog.yaml. Build agent locally, run it, and send through traces (I tested using calendar app). Verify in the frontend that the data received is the same before and after the refactor.

Benchmarks run in `otlp_test.go`:
```
BenchmarkProcessRequestV1-16    	     129	   9138596 ns/op	15494826 B/op	   11583 allocs/op
BenchmarkProcessRequestV2-16    	     150	   8102461 ns/op	10221881 B/op	   11312 allocs/op

BenchmarkProcessRequestTopLevelV1-16    	     134	   8847806 ns/op	15701001 B/op	   12577 allocs/op
BenchmarkProcessRequestTopLevelV2-16    	     145	   8197172 ns/op	10222064 B/op	   11313 allocs/op
```

PR #35664, opened 2025-03-31
[OTEL-2526] Fix "ignore resources" logic in OTLP ingest

jade-guiton-dd:
### What does this PR do?

The `apm_config.ignore_resources` Agent config discards traces whose root spans matches a list of regexes. By mistake, #29435 introduced logic which discards individual spans in the OTLP receiver, which leads to the root span behind discarded but not the rest of the trace. Since spans received in the OTLP receiver are passed to the trace agent anyway, this PR removes this part of the code.

### Motivation

A customer encountered this bug, which led to an escalation.

The structure of the OTLP receiver does not make it easy to write a test of the ignore_resources feature with that ingest method, unfortunately. I wrote a terrible and invasive test to validate that 1. the current behavior is broken, and 2. my change fixes it, but I'd rather not commit it. Since the behavior of the ignore resources logic in the trace agent already has a test, the only thing remaining to test would be that we... no longer do anything in the OTLP receiver, which doesn't seem worth a test. If you think we need such a regression test, I can try to add one anyway.

### DIFF
--- a/pkg/trace/api/otlp.go
+++ b/pkg/trace/api/otlp.go
@@ -244,18 +244,6 @@ func (o *OTLPReceiver) receiveResourceSpansV2(ctx context.Context, rspans ptrace
 		libspans := rspans.ScopeSpans().At(i)
 		for j := 0; j < libspans.Spans().Len(); j++ {
 			otelspan := libspans.Spans().At(j)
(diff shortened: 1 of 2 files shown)


urseberry (approved):

songy23 (approved):
LGTM

PR #35685, opened 2025-03-31
AGNTLOG-4 Refactor tcp/http destination support for logs pipeline

ddrthall:
### What does this PR do?
This PR adds nonfunction tcp/http sender constructors, with each split in to its own sub package to prevent consumers of sender from being forced to pull in the entirety of either the http or tcp dependency stack if they do not require both.

### Motivation
This MR serves as part of the RTT Fairness/Distributed Sends chain, with the overall feature set enablement MR found at https://github.com/DataDog/datadog-agent/pull/35484. This feature requires destination construction in the core pipeline to be moved, as the sender object is no longer constructed at runtime within a new pipeline. Additionally, this refactor itself allows us to abstract away the concept of destinations and how to create/manipulate them from users of the log pipeline.

No QA required for this change, as the entities added are either testing additions or currently unreachable code.

### DIFF
--- a/pkg/logs/sender/http/http_sender.go
+++ b/pkg/logs/sender/http/http_sender.go
@@ -1,0 +1,109 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+// Package http manages creation of http-based senders
+package http
+
+import (
+	"strconv"
+	"sync"
+
+	"github.com/DataDog/datadog-agent/comp/logs/agent/config"
+	pkgconfigmodel "github.com/DataDog/datadog-agent/pkg/config/model"
+	"github.com/DataDog/datadog-agent/pkg/logs/auditor"
+	"github.com/DataDog/datadog-agent/pkg/logs/client"
+	"github.com/DataDog/datadog-agent/pkg/logs/client/http"
+	"github.com/DataDog/datadog-agent/pkg/logs/metrics"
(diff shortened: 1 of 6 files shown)


pkg/logs/sender/http/http_sender.go
```diff
@@ -0,0 +1,109 @@
...
+		bufferSize,
+		senderDoneChan,
+		flushWg,
+		queueCount,
+		workersPerQueue,
+		pipelineMonitor,
+	)
+}
+
+func httpDestinationFactory(
```
ddrthall, line 76: Original code locations, which will be supplanted with this logic:
[epforwarder](https://github.com/DataDog/datadog-agent/blob/main/comp/forwarder/eventplatform/eventplatformimpl/epforwarder.go#L418)
[pipeline](https://github.com/DataDog/datadog-agent/blob/0dd8d837a73267435c6dd847cd5c8894dc0569a0/pkg/logs/pipeline/pipeline.go#L133)

pkg/logs/sender/tcp/tcp_sender.go
```diff
@@ -0,0 +1,74 @@
...
+		bufferSize,
+		senderDoneChan,
+		flushWg,
+		queueCount,
+		workersPerQueue,
+		pipelineMonitor,
+	)
+}
+
+func tcpDestinationFactory(
```
ddrthall, line 56: Original code location which will be supplanted with this logic:  [pipeline](https://github.com/DataDog/datadog-agent/blob/0dd8d837a73267435c6dd847cd5c8894dc0569a0/pkg/logs/pipeline/pipeline.go#L152)

gh123man (approved):
Nice, simple, and clean PR. Thanks for breaking down this refactor and making it easy to review! 👏

PR #35686, opened 2025-04-01
Go DI capture depth backoff

grantseltzer:
### What does this PR do?

Adds logic to BPF program generation/compilation/attachment where the capture depth is decremented and the steps are retried in cases of compilation, verifier, or attachment errors. This is to increase reliability of partial snapshots. The idea is that if one layer breaks something (for some unpredicted reason), we can hopefully try to capture as much data as possible.

### Motivation

Increased reliability of capturing at least partial snapshots.

Added test.

### Additional Notes

We may also want to add logic for capturing one parameter but not another?

### DIFF
--- a/pkg/dynamicinstrumentation/codegen/expression_templates.go
+++ b/pkg/dynamicinstrumentation/codegen/expression_templates.go
@@ -117,3 +117,14 @@ var setParameterIndexText = `
 bpf_printk("Setting param index %d to %d", {{.Arg1}}, context.output_offset);
 event->base.param_indicies[{{.Arg1}}] = context.output_offset;
 `
+
+var compilerErrorText = `
+!@#$%^
+`
+
+var verifierErrorText = `
+for (int i=0; i==0;) {
+    i++;
+    i--;
+}
+`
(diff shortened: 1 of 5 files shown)


pkg/dynamicinstrumentation/codegen/expression_templates.go
```diff
@@ -117,3 +117,14 @@ var setParameterIndexText = `
 bpf_printk("Setting param index %d to %d", {{.Arg1}}, context.output_offset);
 event->base.param_indicies[{{.Arg1}}] = context.output_offset;
 `
+
+var compilerErrorText = `
+!@#$%^
+`
+
+var verifierErrorText = `
```
tylfin, line 125: Can you explain what these templates are doing?

grantseltzer: Yes i've added comments above the texts. They're just for testing purposes (but need to be in this package for the sake of package correctness, I don't want to define an interface just for this)

grantseltzer:
I somehow picked up a couple commits from outside my branch. Fixing.

tylfin (approved):

PR #35692, opened 2025-04-01
Move t.Fatal after TearDownSuite

KevinFairise2:
### What does this PR do?

We need to call `Fatal` after `TearDownSuite` otherwise the suite is never destroyed on SetupFailure.
Calling only `Error` would not work because it does prevent test from being executed after Tearing down the suite

### DIFF
--- a/test/new-e2e/pkg/e2e/suite.go
+++ b/test/new-e2e/pkg/e2e/suite.go
@@ -246,7 +246,6 @@ func (bs *BaseSuite[Env]) CleanupOnSetupFailure() {
 		defer func() {
 			bs.TearDownSuite()
 			bs.T().Fatal("SetupSuite failed, aborted test suite and called TearDownSuite")
-
 		}()
 
 		// run environment diagnose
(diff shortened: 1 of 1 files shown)


test/new-e2e/pkg/e2e/suite.go
```diff
@@ -244,8 +244,9 @@ func (bs *BaseSuite[Env]) CleanupOnSetupFailure() {
 	if err := recover(); err != nil || bs.T().Failed() {
 		bs.firstFailTest = "Initial provisioning SetupSuite" // This is required to handle skipDeleteOnFailure
 		defer func() {
-			bs.T().Fatalf("SetupSuite failed, aborting test suite and calling TearDownSuite")
 			bs.TearDownSuite()
+			bs.T().Fatal("SetupSuite failed, aborted test suite and called TearDownSuite")
+
```
CelianR, line 249:
```suggestion
```

CelianR (approved):

PR #35693, opened 2025-04-01
gitlab-ci.yml: Use a variable for compare_to

albertvaka:
### What does this PR do?
Remove a TODO by replacing the hardcoded branch name with a variable.

### Motivation
It's supported since Gitlab 17.2, as stated in the linked issue https://gitlab.com/gitlab-org/gitlab/-/issues/369916


KevinFairise2:
I think we should update the tooling we have to handle updating the compare_to branch: https://github.com/DataDog/datadog-agent/blob/main/tasks/release.py#L809-L818
It would need to update only the global variable now, can you do it as well?

chouetz:
> .gitlab/.pre/ci_configuration.yml

@albertvaka I think it's because the test on `ci_configuration.yml` is a recent one and we missed to related it to the `compare_to` updates. So I think with your change we are now OK, thanks a lot for this contribution!

chouetz (approved):

FlorentClarret (approved):

PR #35710, opened 2025-04-01
[SBOM] Improve disk space check for image scanning

lebauce:
### What does this PR do?

Checks against the real image size instead of a static value and only when uncompressing the image is required.

### Motivation

On hosts with less than the static value defined (15G on prod), no scan can be done.

### DIFF
--- a/pkg/sbom/scanner/scanner.go
+++ b/pkg/sbom/scanner/scanner.go
@@ -125,7 +125,7 @@ func (s *Scanner) Scan(request sbom.ScanRequest) error {
 	return nil
 }
 
-func (s *Scanner) enoughDiskSpace(opts sbom.ScanOptions) error {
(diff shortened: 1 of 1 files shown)


pkg/sbom/scanner/scanner.go
```diff
@@ -139,6 +139,17 @@ func (s *Scanner) enoughDiskSpace(opts sbom.ScanOptions) error {
 		return fmt.Errorf("not enough disk space to safely collect sbom, %d available, %d required", usage.Available, opts.MinAvailableDisk)
 	}
 
+	if imgMeta == nil || opts.OverlayFsScan || opts.UseMount {
```
paulcacheux, line 142: Shouldn't this be move on top of the previous check ?

lebauce: I wanted to keep the ability to have a minimal size check to handle things like cache. The default is pretty small (1G)

paulcacheux (approved):

PR #35726, opened 2025-04-01
[EBPF] gpu: tag limit metrics too when sending closing values

gjulianm:
### What does this PR do?

Adds entity (PID/container) tags to the `gpu.{core,memory}.limit` metrics when we send the zero closing values once we stop seeing activity for a given PID.

### Motivation

Usage metrics should always have a closing value.

### DIFF
--- a/pkg/collector/corechecks/gpu/gpu.go
+++ b/pkg/collector/corechecks/gpu/gpu.go
@@ -202,14 +202,6 @@ func (c *Check) Run() error {
 }
 
 func (c *Check) emitSysprobeMetrics(snd sender.Sender, gpuToContainersMap map[string][]*workloadmeta.Container) error {
-	sentMetrics := 0
-
-	// Always send telemetry metrics
-	defer func() {
-		c.telemetry.metricsSent.Add(float64(sentMetrics), "system_probe")
-		c.telemetry.activeMetrics.Set(float64(len(c.activeMetrics)))
-	}()
-
 	if err := c.ensureInitDeviceCache(); err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


pkg/collector/corechecks/gpu/gpu.go
```diff
@@ -265,11 +265,22 @@ func (c *Check) emitSysprobeMetrics(snd sender.Sender, gpuToContainersMap map[st
...
 
+			// Here we also need to mark these entities as active, if we don't the limit metrics won't have
+			// the tags and utilization will not be reported for them, as the limit metric won't match
+			if _, ok := activeEntitiesPerDevice[key.DeviceUUID]; !ok {
+				activeEntitiesPerDevice[key.DeviceUUID] = common.NewStringSet()
+			}
+
+			for _, t := range processTags {
+				activeEntitiesPerDevice[key.DeviceUUID].Add(t)
+			}
```
val06, line 282: consider adding a UT to test that edge case.

pkg/collector/corechecks/gpu/gpu.go
```diff
@@ -265,11 +265,22 @@ func (c *Check) emitSysprobeMetrics(snd sender.Sender, gpuToContainersMap map[st
...
 	for key, active := range c.activeMetrics {
 		if !active {
-			tags := append(c.getProcessTagsForKey(key), c.deviceTags[key.DeviceUUID]...)
+			processTags := c.getProcessTagsForKey(key)
+			tags := append(processTags, c.deviceTags[key.DeviceUUID]...)
 			snd.Gauge(metricNameMemoryUsage, 0, "", tags)
 			snd.Gauge(metricNameCoreUsage, 0, "", tags)
 			sentMetrics += 2
 
+			// Here we also need to mark these entities as active, if we don't the limit metrics won't have
```
val06, line 274:
```suggestion
			// Here we also need to mark these entities as active, if we don't, the limit metrics won't have
```

val06 (approved):
left suggestion

PR #35732, opened 2025-04-01
pkg/util/log: allow `log.Limiter` to work without a goroutine

paulcacheux:
### What does this PR do?

This PR refactors (but keeps the behavior unchanged) the log limiter to remove the need for a background goroutine (currently every log limiter requires an additional goroutine to reset the counter periodically) and instead use /x/time utilities (here the rate limiter).


guyarb (approved):

BaptisteFoy (approved):
lgtm for files owned by fleet

pkg/util/log/log_limit.go
pgimalac, line 1: For most uses `n` will be 0 so I think we can improve perfs by read-locking first, and only if n is non-zero we write-lock (`rate.Limiter` is thread-safe)
I don't think this is used by anything on a critical path so up to you

pgimalac, line 1: Alternatively this could be implemented using https://pkg.go.dev/golang.org/x/time@v0.11.0/rate#Sometimes, it supports allowing the first N calls and then calls at every interval, we barely would have any code to write 😄

pgimalac (approved):
Suggested some possible changes, but I believe this is already an improvement so approving anyway

pkg/util/log/log_limit.go
paulcacheux: I don't expect a lot of contention on this, so if I can do without the "read lock to write lock" dance which is never easy I think I would prefer to keep it as is. if this really ends up being a problem in the future we can revisit

paulcacheux: Ok, I ended up rewriting it using this, it's much simpler now

pgimalac (approved):
❤️

hmahmood (approved):

PR #35737, opened 2025-04-02
[AGENTCFG-172] Adding support to UnmarshalKey in handling scalar pointers.

rahulkaukuntla:
### What does this PR do?
The agent configuration system is being changed from Viper to Nodetreemodel. The [UnmarshalKey](https://github.com/DataDog/datadog-agent/blob/66541050bde7cde83e76e84b5c6ae2b03e7de6cb/pkg/config/model/types.go#L261) function in Nodetreemodel does not support fields with type *bool, so this PR adds support for that.

While testing this new support, we noticed that the test suites `comp/core/autodiscovery/listeners` and `pkg/collector/corechecks/snmp/status` were passing structs into the method [SetWithoutSource](https://github.com/DataDog/datadog-agent/blob/66541050bde7cde83e76e84b5c6ae2b03e7de6cb/pkg/config/model/types.go#L208), which is not allowed and also causing the new *bool support to fail. Viper was being too lenient in permitting this, but for Nodetreemodel, we don't want tests to do this, as in the actual productionized agent, config data always comes from unmarshalling text. This means that the kind of data held by the config's store will always be of specific kinds: either scalar primitives (bool, int, float, string, etc) or slices or maps of other such data. So, this pr also fixes these tests, so that only simple objects are being passed into calls to SetWithoutSource.

This code's relevant unit tests are passing, both with viper enabled and with nodetreemodel enabled.


TCheruy (approved):

wdhif (approved):

pkg/config/nodetreemodel/config.go
```diff
@@ -236,6 +236,11 @@ func (c *ntmConfig) Set(key string, newValue interface{}, source model.Source) {
 
 // SetWithoutSource assigns the value to the given key using source Unknown
 func (c *ntmConfig) SetWithoutSource(key string, value interface{}) {
+	v := reflect.ValueOf(value)
+	if v.Kind() == reflect.Struct {
+		log.Debugf("Structs are not supported as values for key %q", key)
```
dustmop, line 241: No need to `log.Debug` as well, the `panic` should be loud enough for our needs.

pkg/config/structure/unmarshal.go
```diff
@@ -369,6 +370,32 @@ func copyList(target reflect.Value, inputList []nodetreemodel.Node, currPath []s
 
 func copyAny(target reflect.Value, input nodetreemodel.Node, currPath []string, fs *featureSet) error {
 	if target.Kind() == reflect.Pointer {
+		if target.Type().Elem().Kind() != reflect.Invalid && isScalarKind(reflect.New(target.Type().Elem()).Elem()) {
```
dustmop, line 373: I don't think this block of code is necessary. It is performing extra allocations and adds lots of complexity that isn't needed. If it's removed, the tests that used to be failing (before this PR) will still succeed because the actual issue was the usage of structs being passed to `SetWithoutSource`.

It does fix a real problem, that of assigning nil to fields of a struct that aren't already nil, but I think we should hold off on that, and save it for another PR, because we want to ensure the behavior is the same as what viper did.

pkg/config/structure/unmarshal_test.go
```diff
@@ -1482,3 +1482,173 @@ func TestMapGetChildNotFound(t *testing.T) {
 	assert.True(t, ok)
 	assert.Equal(t, inner.ChildrenKeys(), []string{"a", "b"})
 }
+
+func TestCopyAny_ScalarPointers(t *testing.T) {
+	t.Run("Nil source to nil bool pointer", func(t *testing.T) {
```
dustmop, line 1487: It's great to have this breadth of testing, but I feel these tests are too low-level. Instead of exercising `copyAny` and writing tests against it directly, we should be looking higher and test `UnmarshalKey`. It's better to have tests against the actual interface that external packages are using, instead of implementation details.

dustmop (changes requested):
Thanks for the fixes here! There's a couple of fixes and then I think we'll be good to merge.

pkg/config/nodetreemodel/config.go
rahulkaukuntla: removed.

pkg/config/structure/unmarshal_test.go
rahulkaukuntla: Ah, makes sense. Since I'm removing some code due to [this](https://github.com/DataDog/datadog-agent/pull/35737#discussion_r2091921780), I don't need this test suite anyway. I've replaced it with a standalone test that just makes sure that *bool fields are unmarshalled by UnmarshalKey correctly.

dustmop: Great, having just the single test to validate `*bool` fields work is a good idea.

dustmop (approved):
Thanks for fixing, LGTM!

PR #35741, opened 2025-04-02
[EBPF]: changed gpu deviceCache to be a global singleton

val06:
### What does this PR do?

changes the deviceCache API to be a global singleton.

### Motivation

Simplify the code by removing some parameters propagation

modified the tests to use a new mock API for the deviceCache
existing tests should pass

### Additional Notes
using the double-lock pattern to prevent locks contentions on a hot-path, but also provide thread safety when accessed from different go-routines during the init phase

### DIFF
--- a/pkg/gpu/nvml/devices.go
+++ b/pkg/gpu/nvml/devices.go
@@ -10,6 +10,8 @@ package nvml
 
 import (
 	"fmt"
+	"sync"
+	"sync/atomic"
 
 	"github.com/NVIDIA/go-nvml/pkg/nvml"
 
@@ -65,6 +67,13 @@ func NewDevice(dev nvml.Device) (*Device, error) {
 	}, nil
 }
 
+var (
+	// globalDeviceCache is the singleton instance
+	globalDeviceCache atomic.Pointer[DeviceCache]
+	// initMutex protects initialization
+	initMutex sync.Mutex
+)
+
 // DeviceCache is a cache of GPU devices, with some methods to easily access devices by UUID or index
 type DeviceCache interface {
 	// GetByUUID returns a device by its UUID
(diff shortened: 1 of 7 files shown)


pkg/gpu/nvml/devices.go
```diff
@@ -65,6 +67,13 @@ func NewDevice(dev nvml.Device) (*Device, error) {
 	}, nil
 }
 
+var (
+	// globalDeviceCache is the singleton instance
+	globalDeviceCache atomic.Pointer[DeviceCache]
+	// initMutex protects initialization
+	initMutex sync.Mutex
+)
+
```
gjulianm, line 76: I'd put this in a struct as in the `nvmlLib`, just to avoid confusion between this singleton and the one for the library

pkg/gpu/nvml/devices_mock.go
```diff
@@ -0,0 +1,42 @@
...
+
+// testingT is an interface matching the necessary testing.T methods we need
+type testingT interface {
+	Cleanup(func())
+	Errorf(format string, args ...any)
+}
+
+// WithMockDeviceCache sets up a mock device cache for testing and ensures cleanup
+// This should only be used in tests
+func WithMockDeviceCache(t testingT, mock nvml.Interface) {
```
gjulianm, line 20:
```suggestion
func WithMockDeviceCache(t testing.TB, mock nvml.Interface) {
```

I think `testing.TB` should be enough, that way we avoid another interface

gjulianm (commented):
Overall I'm not sure if this simplifies things, before we had code that received a initialized, correct `DeviceCache` that now has to get it from the singleton and check the error. And we were initializing the cache only once in system-probe and once in the agent, so I don't think we're reducing the number of run-time instances.

PR #35744, opened 2025-04-02
[ADXT-1000] Make sure go test command remain under the maximum length on Windows

KevinFairise2:
### What does this PR do?

Now that all the tests are executed with a single command, we are very likely to hit the max CLI length on Windows. Add a safeguard to make sure that we disable fast test if the command become too long.

Tested here: https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/876879287

### Motivation

Avoid broken tests because of too long CLI command

### DIFF
--- a/tasks/gotest.py
+++ b/tasks/gotest.py
@@ -835,7 +835,7 @@ def format_packages(ctx: Context, impacted_packages: set[str], build_tags: list[
 
     # We need to make sure the CLI length is not too long
     packages = compute_gotestsum_cli_args(modules_to_test.values())
-    if sys.platform "win32" and len(packages) > WINDOWS_MAX_CLI_LENGTH:
+    if sys.platform == "win32" and len(packages) > WINDOWS_MAX_CLI_LENGTH:
         print("CLI length is too long, skipping fast tests")
         return get_default_modules().values()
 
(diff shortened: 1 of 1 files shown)


tasks/gotest.py
```diff
@@ -842,6 +833,12 @@ def format_packages(ctx: Context, impacted_packages: set[str], build_tags: list[
         for module in modules_to_test:
             print(f"- {module}: {modules_to_test[module].test_targets}")
 
+    # We need to make sure the CLI length is not too long
+    packages = compute_gotestsum_cli_args(modules_to_test.values())
+    if len(packages) > WINDOWS_MAX_CLI_LENGTH:
```
chouetz, line 838:
```suggestion
    if sys.platform "win32" and len(packages) > WINDOWS_MAX_CLI_LENGTH:
```
`sys` is already included in this file. I suppose we only need to limit this on windows (which is unfortunate because it's the slowest environment 😭 )

chouetz (approved):
I suppose like this it will not impact overall duration but it's maybe better to really limit this on windows, wdyt?

tasks/gotest.py
```diff
@@ -842,6 +833,12 @@ def format_packages(ctx: Context, impacted_packages: set[str], build_tags: list[
         for module in modules_to_test:
             print(f"- {module}: {modules_to_test[module].test_targets}")
 
+    # We need to make sure the CLI length is not too long
+    packages = compute_gotestsum_cli_args(modules_to_test.values())
+    if sys.platform "win32" and len(packages) > WINDOWS_MAX_CLI_LENGTH:
```
KevinFairise2:
```suggestion
    if sys.platform == "win32" and len(packages) > WINDOWS_MAX_CLI_LENGTH:
```

PR #35746, opened 2025-04-02
support process tags in apm stats and spans 

raphaelgavache:
### What does this PR do?
Propagates process tags received by connected tracer in spans and stats through the header `X-Datadog-Process-Tags`.

E2E test coverage
- [x] trace payload
- [x] stats payload
- [ ] client stats payload (will add after go tracer code is shipped, it's not easy to add with the current stats protocol and curl)

#### Spans
The tags are transmitted straight in the TracerPayload.Tags["_dd.tags.process"]

#### Stats
Stats aggregation by process tags follows the same pattern as container tags. A specific process tags field is added to the stats payload.


pkg/trace/stats/span_concentrator.go
```diff
@@ -244,6 +251,7 @@ func (sc *SpanConcentrator) Flush(now int64, force bool) []*pb.ClientStatsPayloa
 			ImageTag:     k.ImageTag,
 			Stats:        s,
 			Tags:         containerTagsByID[k.ContainerID],
+			ProcessTags:  processTagsByHash[k.ProcessTagsHash],
```
raphaelgavache, line 254: todo: add ProcessTagsHash too

pkg/trace/stats/concentrator.go
```diff
@@ -113,28 +116,46 @@ type Input struct {
...
-		c.addNow(&trace, t.ContainerID, t.ContainerTags)
+		c.addNow(&trace, tags)
 	}
 }
 
+type infraTags struct {
+	containerID     string
+	containerTags   []string
+	processTagsHash uint64
+	processTags     string
```
dumontg, line 147: nit: a bit stange to have `containerTags` being a slice and `processTags` a string, is it because you don't want to perform useless conversions? I assume converting `containerTags` to a string (since they will end up concatenated anyway) would cause too many changes?

dumontg (approved):

pkg/trace/stats/concentrator.go
raphaelgavache: given that other products use the trace-agent as a proxy, I wanted to keep the same encoding like this we can do same normalization/treatments in different ingestion points

pkg/trace/api/payload.go
```diff
@@ -29,6 +29,9 @@ type Payload struct {
 
 	// ClientDroppedP0s specifies the number of P0 traces chunks dropped by the client.
 	ClientDroppedP0s int64
+
+	// ProcessTags is a list of tags describing an instrumented process.
+	ProcessTags string
```
ichinaski, line 34: Process Tags are already passed within the `TracerPayload.Tags["_dd.tags.process"]`, why have this dedicated field with the same data?

pkg/proto/datadog/trace/stats.proto
```diff
@@ -49,6 +49,10 @@ message ClientStatsPayload {
 	string git_commit_sha = 13;
 	// The image tag is obtained from a container's set of tags.
 	string image_tag = 14;
+        // The process tags hash is used as a key for agent stats agregation.
+        uint64 process_tags_hash = 15;
+        // The process tags contains a list of tags that are specific to the process.
+        string process_tags = 16;
```
ichinaski, line 55: why not have a `repeated string` field with split tags, same as container tags? This way you don't have to split the tags when computing the hash.

ichinaski: I was also going to suggest using camelCase for consistency but noticed we already have a mix of cases :(

pkg/trace/api/api.go
```diff
@@ -657,13 +658,21 @@ func (r *HTTPReceiver) handleTraces(v Version, w http.ResponseWriter, req *http.
...
+		tp.Tags[tagProcessTags] = ptags
+	}
 
 	payload := &Payload{
 		Source:                 ts,
 		TracerPayload:          tp,
 		ClientComputedTopLevel: isHeaderTrue(header.ComputedTopLevel, req.Header.Get(header.ComputedTopLevel)),
 		ClientComputedStats:    isHeaderTrue(header.ComputedStats, req.Header.Get(header.ComputedStats)),
 		ClientDroppedP0s:       droppedTracesFromHeader(req.Header, ts),
+		ProcessTags:            ptags,
```
ichinaski, line 675: see other comment, this data is already present within the `TracerPayload` passes here.

pkg/trace/stats/span_concentrator.go
```diff
@@ -184,24 +184,28 @@ func (sc *SpanConcentrator) addSpan(s *StatSpan, aggKey PayloadAggregationKey, c
...
+	if tags.containerID != "" && len(tags.containerTags) > 0 {
+		b.containerTagsByID[tags.containerID] = tags.containerTags
+	}
 	b.HandleSpan(s, weight, origin, aggKey)
 }
 
 // AddSpan to the SpanConcentrator, appending the new data to the appropriate internal bucket.
-func (sc *SpanConcentrator) AddSpan(s *StatSpan, aggKey PayloadAggregationKey, containerID string, containerTags []string, origin string) {
-	sc.addSpan(s, aggKey, containerID, containerTags, origin, 1)
+func (sc *SpanConcentrator) AddSpan(s *StatSpan, aggKey PayloadAggregationKey, tags infraTags, origin string) {
```
ichinaski, line 199: Note that this is also accessed from `dd-trace-go`, thus its code needs to be updated after upgrading the agent version.

pkg/trace/api/payload.go
raphaelgavache: it's used to pass it to stats without doing a map access, agreed though it makes a dupe, I'll move stats to pick it from this field

pkg/proto/datadog/trace/stats.proto
raphaelgavache: since we're propagating this payload to multiple intakes, to keep the same normalisation I preferred to not touch what's provided by the tracing library and normalize in the backend
Tracer library will have it's own normalisation at the source

pkg/trace/stats/span_concentrator.go
stanistan: If this is accessed by `dd-trace-go` does the `infraTags` struct needs to be public?

pkg/trace/api/api.go
raphaelgavache: Played with it, but in the end it's a lot cleaner to keep the processTags as a dedicated field available for the stats concentrator

pkg/proto/datadog/trace/stats.proto
truthbk: > This way you don't have to split the tags when computing the hash.

I think this can actually be a pretty big deal when it comes to performance, so we should really think twice about not doing this and be very intentional about it. It's probably fine, but it would be nice to understand what perf penalty we might be eating here with this decision.

raphaelgavache: What performance cost do you expect? This should be actually the most efficient
1. tracers normalise once their process tags in their lifetime
2. trace-agent is a passthrough on them
3. backends have a second normalization to cover further renormalization

dumontg (approved):

pkg/trace/stats/span_concentrator.go
```diff
@@ -184,24 +184,28 @@ func (sc *SpanConcentrator) addSpan(s *StatSpan, aggKey PayloadAggregationKey, c
...
+	if tags.containerID != "" && len(tags.containerTags) > 0 {
+		b.containerTagsByID[tags.containerID] = tags.containerTags
+	}
 	b.HandleSpan(s, weight, origin, aggKey)
 }
 
 // AddSpan to the SpanConcentrator, appending the new data to the appropriate internal bucket.
 func (sc *SpanConcentrator) AddSpan(s *StatSpan, aggKey PayloadAggregationKey, containerID string, containerTags []string, origin string) {
-	sc.addSpan(s, aggKey, containerID, containerTags, origin, 1)
+	sc.addSpan(s, aggKey, infraTags{containerID: containerID, containerTags: containerTags}, origin, 1)
```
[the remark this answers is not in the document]
raphaelgavache: I was looking to treat this as a follow-up starting with dd-trace-go work on how to handle process tags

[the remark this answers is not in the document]
raphaelgavache: for example I suspect that the API will actually be processTags only there, as container info isn't available,
I'll look more into the go tracer just after

pkg/trace/stats/aggregation.go
```diff
@@ -99,6 +100,13 @@ func NewAggregationFromSpan(s *StatSpan, origin string, aggKey PayloadAggregatio
 	return agg
 }
 
+func processTagsHash(processTags string) uint64 {
+	if processTags == "" {
+		return 0
+	}
+	return peerTagsHash(strings.Split(processTags, ","))
```
kitfre, line 107: nit: can we rename `peerTagsHash` to be more general now that we're using it in both places?

kitfre (approved):
LGTM from the stats backend

pkg/trace/stats/span_concentrator.go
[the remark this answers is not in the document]
raphaelgavache: Unfortunately there's a chicken and the egg problem with changing that import as the agent imports the tracer too and we get this failure. This means that we have to add a new API, transition the tracer to it and rename etc..
```
pkg/collector/corechecks/servicediscovery/apm/testutil/instrumented/instrumented.go:16:2: could not import gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer (-: # gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer
/go/pkg/mod/gopkg.in/!data!dog/dd-trace-go.v1@v1.72.2/ddtrace/tracer/stats.go:174:59: cannot use "" (untyped string constant) as uint64 value in argument to c.spanConcentrator.AddSpan
/go/pkg/mod/gopkg.in/!data!dog/dd-trace-go.v1@v1.72.2/ddtrace/tracer/stats.go:174:63: cannot use nil as string value in argument to c.spanConcentrator.AddSpan) (typecheck)
	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
```

I'd rather re-work the dependency of the go tracer separately 
- remove container piece that is unused
- figure out of we want process tags there or at the end of the export of ClientStatsPayload in the tracer

There's absolutely no risk as long as process tags are not implemented in the go tracer, and to be implemented they need this rework

pkg/trace/stats/client_stats_aggregator.go
```diff
@@ -298,13 +301,15 @@ func (b *bucket) aggregationToPayloads() []*pb.ClientStatsPayload {
...
-			Stats:        clientBuckets,
+			Hostname:        payloadKey.Hostname,
+			Env:             payloadKey.Env,
+			Version:         payloadKey.Version,
+			ImageTag:        payloadKey.ImageTag,
+			GitCommitSha:    payloadKey.GitCommitSha,
+			ContainerID:     payloadKey.ContainerID,
+			Stats:           clientBuckets,
+			ProcessTagsHash: payloadKey.ProcessTagsHash,
+			ProcessTags:     b.processTags[payloadKey.ProcessTagsHash],
```
ichinaski, line 312: could you add one test on `client_stats_aggregator_test.go` to ensure the flushed payload contains these fields? 🙏

ichinaski (approved):
Leaving the PR approved to unblock it, but would be nice to have a bit more test coverage, at least the client stats aggregation one.

pkg/trace/stats/client_stats_aggregator.go
raphaelgavache: it's actually covered by the test TestConvertStats in trace/agent/agent_test.go where most of client_stats_aggregator tests are, I'll check if I can find a meaningful additional test to add in `lient_stats_aggregator_test.go`

pkg/proto/datadog/trace/stats.proto
truthbk: So, maybe I'm missing something but from this comment:

>This way you don't have to split the tags when computing the hash.

The associated cost I'm alluding to is related to that precisely, the need to do string allocations due to the string split, and then calculating the hash on that split.

raphaelgavache: that split would still be needed if we stored this as an array of string here, the process tags are received through a http header (string to split)

truthbk (approved):

PR #35750, opened 2025-04-02
[kubelet/podwatcher] Report changes in container states

davidor:
### What does this PR do?

Fixes an issue in the Kubelet pod watcher where changes in container state (Waiting, Running, or Terminated) were not being reported.

This affected the KSM check when running with the `pod_collection_mode: node_kubelet` option, because it relies on the Kubelet pod watcher as a data source.

As a result, the `kubernetes_state.container.status_report.count.waiting` metric could report incorrect values. For example, in cases where a container is in a crash loop backoff, its state frequently transitions between Waiting, Running, and Terminated, but these changes were not consistently detected by the pod watcher.

I added a unit test and also manually verified that `kubernetes_state.container.status_report.count.waiting` is now correct when KSM is configured with `pod_collection_mode: node_kubelet` and there's a pod in CrashLoopBackOff.

### DIFF
--- a/pkg/util/kubernetes/kubelet/podwatcher.go
+++ b/pkg/util/kubernetes/kubelet/podwatcher.go
@@ -18,19 +18,26 @@ import (
 	"github.com/DataDog/datadog-agent/pkg/util/log"
 )
 
-const unreadinessTimeout = 30 * time.Second
+const (
+	unreadinessTimeout = 30 * time.Second
+
+	containerRunning    = "Running"
+	containerWaiting    = "Waiting"
+	containerTerminated = "Terminated"
+)
 
 // PodWatcher regularly pools the kubelet for new/changed/removed containers.
 // It keeps an internal state to only send the updated pods.
 type PodWatcher struct {
 	sync.Mutex
-	kubeUtil       KubeUtilInterface
-	expiryDuration time.Duration
-	lastSeen       map[string]time.Time
-	lastSeenReady  map[string]time.Time
-	tagsDigest     map[string]string
-	oldPhase       map[string]string
-	oldReadiness   map[string]bool
+	kubeUtil          KubeUtilInterface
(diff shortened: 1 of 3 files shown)


releasenotes/notes/fix-kubernetes-podwatcher-container-state-6c469104778275a2.yaml
```diff
@@ -0,0 +1,13 @@
...
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+fixes:
+  - |
+    Fixed an issue in the KSM check when running with the ``pod_collection_mode:
+    node_kubelet`` option, where the Agent was reporting incorrect values for
+    the ``kubernetes_state.container.status_report.count.waiting`` metric.
```
drichards-87, line 13:
```suggestion
    Fixed an issue in the KSM check where, when using ``pod_collection_mode: node_kubelet``, the Agent reported incorrect values for the ``kubernetes_state.container.status_report.count.waiting`` metric.
```

drichards-87 (approved):

pkg/util/kubernetes/kubelet/podwatcher.go
```diff
@@ -212,3 +224,19 @@ func digestMapValues(m map[string]string) string {
 	}
 	return strconv.FormatUint(h.Sum64(), 16)
 }
+
+func containerStateStr(containerState ContainerState) string {
+	if containerState.Running != nil {
+		return "Running"
```
clamoriniere, line 230: should we use `const string` here?

davidor: Updated :+1:

clamoriniere (approved):

PR #35763, opened 2025-04-02
[EBPF] gpu: increase timeouts for finding workload container in e2e test

gjulianm:
### What does this PR do?

Increases the timeout to wait for workload container to come up, to avoid flaky failures. It also improves the error message output so that, if it still fails, it will be easier to debug.

### Motivation

Reduce flaky failures in e2e tests.

E2E tests passing.

### DIFF
--- a/test/new-e2e/tests/gpu/capabilities.go
+++ b/test/new-e2e/tests/gpu/capabilities.go
@@ -7,7 +7,6 @@ package gpu
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"math/rand"
 	"strconv"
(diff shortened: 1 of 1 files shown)


test/new-e2e/tests/gpu/capabilities.go
```diff
@@ -29,8 +28,8 @@ import (
 
 const agentNamespace = "datadog"
 const podSelectorField = "app"
-const jobQueryInterval = 250 * time.Millisecond
-const jobQueryTimeout = 30 * time.Second // Might take some time to pull the image
+const jobQueryInterval = 500 * time.Millisecond
```
val06, line 31: why do you increase the query interval? isn't it enough to increase the timeout interval?

gjulianm: It's just to avoid querying too much, if the wait interval is this long it doesn't matter if we're querying every quarter or half a second

val06: but don't we stop query once we get a successful response? anyway up to you, approving it

val06 (approved):

PR #35773, opened 2025-04-02
feat(autoinstrumentation): dont mutate istio-proxy

### LINKED ISSUE #33296: [BUG]  APM Inject fails when explicitely mentioned istio container
**Agent Environment**

Agent 7.61.0 - Commit: 202f54bcf5 - Serialization version: v5.0.137 - Go version: go1.22.8

**Describe what happened:**
Deployed app with admission.datadoghq.com/enabled
Pod couldn't be created because of error in container `datadog-init-apm-inject`
```
/bin/sh: can't create /datadog-etc/ld.so.preload: Is a directory
```

Create pod without issue

It seems it's related to another init container, in our case istio. If I don't specify container explicitly everything works as expected, but if I want to modify something then init-apm-inject fails.

```
apiVersion: v1
kind: Pod
metadata:
  name: nginx-pod
spec:
  containers:
  - name: istio-proxy
    image: auto
  - name: nginx
    image: nginx:latest
    ports:
    - containerPort: 80
```

K8s Rev: v1.30.8-gke.1051000   
Istio: 1.22.1

stanistan:
### What does this PR do?

This excludes `istio-proxy` from mutation via auto-instrumentation.

The technical approach here is to _stop using_ the `mutatecommon` package for environment variable mutations and _instead_ make `containerMutator`s that can use a `containerFilter` built on top of configuration so that something that mutates a pod can be composed of a filter and container mutators.

### Motivation

- https://datadoghq.atlassian.net/browse/INPLAT-454
- Closes #33296

There are cases with auto-instrumentation where you want to skip entire containers in a pod.

An alternative here would be to build container targeting into workload selection. That feels a lot heavier and if you're running a something like `istio` (something with another mutating controller) you'll need to change every target to explicitly choose a container.

There are _known_ container names (like `istio-proxy`) that will not be supported for auto-instrumentation (C++) in the near future and when conflating with an issue like #33296 where the `istio` controller will take a sidecar container and move to be a new kubernetes-native "sidecar container" (https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/).

This doesn't fix the issue _fully_ (IE there are probably other containers where it's possible for another mutating controller to cause this issue) but will give people the flexibility to disable injection at a container+cluster level.

We can expose this kind of option in the future but for now having something that prevents the issue from happening is a good start.

### Including Changes

- Introduces a concept of a `containerPredicate` for the `containerMutator`
- Changes how we parameterize both the namespace and target mutators so that they are `containerMutator` instead of `podMutator`. This means we can a `containerPredicate`
- migrate places that use `mutatecommon.InjectEnv` (which currently operates on the pod wholly) to use `predicate(envVar)` 
- update `envVar` to be the basic and common building block for the ways we do mutation in the autoinstrumentation webhook (+ tests) since we had two ways of approaching this problem. This is complicated because of in-tree support for instrumentationv1 (which is deprecated and slated for removal/cleanup) https://datadoghq.atlassian.net/browse/INPLAT-489

- [x] Added unit tests for the cases we support.
- [ ] an end to end test with injector dev

### DIFF
--- a/pkg/clusteragent/admission/mutate/autoinstrumentation/auto_instrumentation.go
+++ b/pkg/clusteragent/admission/mutate/autoinstrumentation/auto_instrumentation.go
@@ -130,26 +130,6 @@ func initContainerName(lang language) string {
 	return fmt.Sprintf("datadog-lib-%s-init", lang)
 }
 
-func injectApmTelemetryConfig(pod *corev1.Pod) {
-	// inject DD_INSTRUMENTATION_INSTALL_TIME with current Unix time
-	instrumentationInstallTime := os.Getenv(instrumentationInstallTimeEnvVarName)
-	if instrumentationInstallTime == "" {
-		instrumentationInstallTime = common.ClusterAgentStartTime
-	}
-	instrumentationInstallTimeEnvVar := corev1.EnvVar{
-		Name:  instrumentationInstallTimeEnvVarName,
-		Value: instrumentationInstallTime,
-	}
-	_ = mutatecommon.InjectEnv(pod, instrumentationInstallTimeEnvVar)
-
-	// inject DD_INSTRUMENTATION_INSTALL_ID with UUID created during the Agent install time
-	instrumentationInstallIDEnvVar := corev1.EnvVar{
-		Name:  instrumentationInstallIDEnvVarName,
-		Value: os.Getenv(instrumentationInstallIDEnvVarName),
-	}
-	_ = mutatecommon.InjectEnv(pod, instrumentationInstallIDEnvVar)
-}
-
 type libInfoLanguageDetection struct {
 	libs             []libInfo
 	injectionEnabled bool
@@ -242,15 +222,38 @@ func (s libInfoSource) isFromLanguageDetection() bool {
 	return s == libInfoSourceSingleStepLangaugeDetection
 }
 
-func (s libInfoSource) mutatePod(pod *corev1.Pod) error {
-	_ = mutatecommon.InjectEnv(pod, corev1.EnvVar{
-		Name:  instrumentationInstallTypeEnvVarName,
-		Value: s.injectionType(),
-	})
+func (s libInfoSource) instrumentationInstallTime() string {
+	instrumentationInstallTime := os.Getenv(instrumentationInstallTimeEnvVarName)
+	if instrumentationInstallTime == "" {
+		instrumentationInstallTime = common.ClusterAgentStartTime
+	}
 
-	injectApmTelemetryConfig(pod)
+	return instrumentationInstallTime
+}
 
-	return nil
+// containerMutator creates a containerMutator for
+// telemetry environment variables pertaining to:
+//
+// - installation_time
+// - install_id
+// - injection_type
+func (s libInfoSource) containerMutator() containerMutator {
+	return containerMutators{
+		// inject DD_INSTRUMENTATION_INSTALL_TIME with current Unix time
+		envVarMutator(corev1.EnvVar{
+			Name:  instrumentationInstallTimeEnvVarName,
+			Value: s.instrumentationInstallTime(),
+		}),
+		// inject DD_INSTRUMENTATION_INSTALL_ID with UUID created during the Agent install time
+		envVarMutator(corev1.EnvVar{
+			Name:  instrumentationInstallIDEnvVarName,
+			Value: os.Getenv(instrumentationInstallIDEnvVarName),
+		}),
+		envVarMutator(corev1.EnvVar{
+			Name:  instrumentationInstallTypeEnvVarName,
+			Value: s.injectionType(),
+		}),
+	}
 }
 
 type extractedPodLibInfo struct {
--- a/pkg/clusteragent/admission/mutate/autoinstrumentation/config.go
+++ b/pkg/clusteragent/admission/mutate/autoinstrumentation/config.go
@@ -37,22 +37,47 @@ type Config struct {
 	// containerRegistry is the container registry to use for the autoinstrumentation logic
 	containerRegistry string
 
-	// precomputed mutators for the security and profiling products
-	securityClientLibraryPodMutators  []podMutator
-	profilingClientLibraryPodMutators []podMutator
+	// precomputed containerMutators for the security and profiling products
+	securityClientLibraryMutator  containerMutator
+	profilingClientLibraryMutator containerMutator
 
-	// initResources is the resource requirements for the init container
+	// containerFilter is a pre-computed filter function
+	// to be passed to the different mutators that dictates
+	// whether or not given pod containers should be mutated by
+	// this webhook.
+	//
+	// At the moment it is based on [[excludedContainerNames]].
+	// See [[excludedContainerNamesContainerFilter]].
+	//
+	// This should be used in conjunction with containerMutators
+	// via [[filteredContainerMutator]] and [[mutatePodContainers]]
+	// to make sure we aren't applying mutations to containers we
+	// should be skipping.
+	containerFilter containerFilter
+
+	// initResources is the resource requirements for the init containers.
 	initResources initResourceRequirementConfiguration
 
-	// initSecurityContext is the security context for the init container
+	// initSecurityContext is the security context for the init containers.
 	initSecurityContext *corev1.SecurityContext
 
-	// defaultResourceRequirements is the default resource requirements for the init container
+	// defaultResourceRequirements is the default resource requirements
+	// for the init containers.
 	defaultResourceRequirements initResourceRequirementConfiguration
 
-	// version is the version of the autoinstrumentation logic to use. We don't expose this option to the user, and V1
+	// version is the version of the autoinstrumentation logic to use.
+	// We don't expose this option to the user, and [[instrumentationV1]]
 	// is deprecated and slated for removal.
 	version version
+}
+
+var excludedContainerNames = map[string]bool{
+	"istio-proxy": true, // https://datadoghq.atlassian.net/browse/INPLAT-454
+}
+
+func excludedContainerNamesContainerFilter(c *corev1.Container) bool {
+	_, exclude := excludedContainerNames[c.Name]
+	return !exclude
 }
 
 // NewConfig creates a new Config from the datadog config. It returns an error if the configuration is invalid.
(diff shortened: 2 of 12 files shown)


pkg/clusteragent/admission/mutate/autoinstrumentation/config.go
```diff
@@ -165,6 +181,15 @@ type InstrumentationConfig struct {
 	// used. If no target matches, the auto instrumentation will not be applied. Full config key:
 	// apm_config.instrumentation.targets
 	Targets []Target `mapstructure:"targets" json:"targets"`
+	// ExcludeContainers configuration for containers to exclude from auto instrumentation.
+	// By default, istio-proxy containers are excluded, but this can be overwritten.
+	ExcludeContainers *ExcludeContainers `mapstructure:"exclude_containers" json:"exclude_containers"`
```
betterengineering, line 186: Any reason not to make `ExcludeContainers []string`? Is there value for adding a custom struct type?

pkg/clusteragent/admission/mutate/autoinstrumentation/target_mutator.go
```diff
@@ -178,24 +181,28 @@ func (m *TargetMutator) MutatePod(pod *corev1.Pod, ns string, _ dynamic.Interfac
...
 	}
 
-	// Inject the tracer configs. We do this before lib injection to ensure DD_SERVICE is set if the user configures it
-	// in the target.
+	// Inject the tracer configs.
+	// We do this before lib injection to ensure DD_SERVICE is set
+	// if the user configures it in the target.
 	for _, envVar := range target.envVars {
-		mutatecommon.InjectEnv(pod, envVar)
+		_ = allContainersPodMutator(
```
betterengineering, line 202: If I'm being honest, I really find that these mutators hard to reason about. I find it much more straightforward when we have a methods like `InjectEnv` that add the env var to a pod. Reading the `allContainersPodMutator`, it's really not clear what is happening. 

Would it be more clear to simply pass a map into `mutatecommon.InjectEnv(pod, envVar, containerFilter)`?

pkg/clusteragent/admission/mutate/autoinstrumentation/config.go
stanistan: The thing I was thinking about here was that if we ever wanted to add _anything_ else it would make it difficult to do it. Like container image names, or patterns, or anything else.

pkg/clusteragent/admission/mutate/autoinstrumentation/target_mutator.go
stanistan: Agree this name is bad and the mutators are a bit much here.

I can try that-- the one other thing that occurred to me about this function was that it doesn't need to exist :) alternatively a `mutatePodContainers(pod, containerMutator)` would be clearer

pkg/clusteragent/admission/mutate/autoinstrumentation/mutators.go
```diff
@@ -160,30 +182,80 @@ func appendOrPrepend[T any](item T, toList []T, prepend bool) []T {
...
+
+func boolEnvValue(in bool) string {
+	return strconv.FormatBool(in)
+}
+
+func stringEnvValue(in string) string {
+	return in
+}
 
+func newConfigEnvVarFromStringMutator(key string, val *string) configKeyEnvVarMutator {
```
stanistan, line 257: I think this can be removed here.

pkg/clusteragent/admission/mutate/autoinstrumentation/namespace_mutator.go
```diff
@@ -205,14 +226,8 @@ func (m *mutatorCore) injectTracers(pod *corev1.Pod, config extractedPodLibInfo)
...
 
-func (m *mutatorCore) newContainerMutators(requirements corev1.ResourceRequirements) containerMutators {
-	return containerMutators{
-		containerResourceRequirements{requirements},
-		containerSecurityContext{m.config.initSecurityContext},
-	}
-}
-
-func (m *mutatorCore) newInjector(startTime time.Time, pod *corev1.Pod, opts ...injectorOption) podMutator {
+func (m *mutatorCore) newInjector(startTime time.Time, pod *corev1.Pod, libOpts libRequirementOptions) podMutator {
```
stanistan, line 229: Did some work to remove the indirection here so that when we set up the injector there's _less stuff_.

pkg/clusteragent/admission/mutate/autoinstrumentation/namespace_mutator.go
```diff
@@ -205,14 +226,8 @@ func (m *mutatorCore) injectTracers(pod *corev1.Pod, config extractedPodLibInfo)
 	return lastError
 }
 
-func (m *mutatorCore) newContainerMutators(requirements corev1.ResourceRequirements) containerMutators {
```
stanistan, line 208: This is just moved renamed to `initContainerMutators`

pkg/clusteragent/admission/mutate/autoinstrumentation/auto_instrumentation.go
```diff
@@ -130,26 +130,6 @@ func initContainerName(lang language) string {
 	return fmt.Sprintf("datadog-lib-%s-init", lang)
 }
 
-func injectApmTelemetryConfig(pod *corev1.Pod) {
```
stanistan, line 133: Moved this into `libInfoSource.mutatePod()` since these both are telemetry metadata related and should be filtered to the same _container rules_.

pkg/clusteragent/admission/mutate/autoinstrumentation/config.go
stanistan: Dropped configuration / UX for this.

pkg/clusteragent/admission/mutate/autoinstrumentation/auto_instrumentation.go
```diff
@@ -242,15 +222,32 @@ func (s libInfoSource) isFromLanguageDetection() bool {
...
+	if instrumentationInstallTime == "" {
+		instrumentationInstallTime = common.ClusterAgentStartTime
+	}
 
-	injectApmTelemetryConfig(pod)
+	return instrumentationInstallTime
+}
 
-	return nil
+func (s libInfoSource) podMutator(filter containerPredicate) podMutator {
```
knusbaum, line 234: Should document what the `podMutator` mutates. What are it's responsibilities?

pkg/clusteragent/admission/mutate/autoinstrumentation/config.go
```diff
@@ -38,8 +38,12 @@ type Config struct {
...
 
 	// precomputed mutators for the security and profiling products
-	securityClientLibraryPodMutators  []podMutator
-	profilingClientLibraryPodMutators []podMutator
+	securityClientLibraryMutator  containerMutator
+	profilingClientLibraryMutator containerMutator
+
+	// precomputed containerPredicate to use with filteredContainerMutator
+	// and other mutators
+	containerFilter containerPredicate
```
knusbaum, line 46: This isn't super descriptive. What kind of filter? What are we filtering and why? When should I use this?

pkg/clusteragent/admission/mutate/autoinstrumentation/env_vars.go
```diff
@@ -75,21 +101,32 @@ func (e envVar) mutateContainer(c *corev1.Container) error {
...
 	}
 
 	return nil
 }
 
+func (e envVar) mutatePod(pod *corev1.Pod) error {
+	return mutatePodContainers(pod, e)
+}
+
 type envValFunc func(string) string
```
knusbaum, line 130: I know you didn't introduce this, but it would be helpful to document what this *is*. It makes the API difficult to understand. 

Is this returning an environment variable value? What do we pass in? 

I'd be in favor of renaming this thing, but doesn't have to be in this PR.

pkg/clusteragent/admission/mutate/autoinstrumentation/mutators.go
```diff
@@ -17,21 +17,31 @@ import (
 
 // containerMutator describes something that can mutate a container.
 type containerMutator interface {
+	podMutator
+
```
knusbaum, line 21: I don't think this is the right abstraction. It would requires every `containerMutator` to reimplement pod mutation, which requires duplicating this everywhere:
```
func (...) mutatePod(pod *corev1.Pod) error {
	return mutatePodContainers(pod, r)
}
```
and all it gets is turning this:
```
mutatePodContainers(pod, r)
```
into this:
```
r.mutatePod(pod)
```

It's fine to just have `mutatePodContainers(pod, r)` take care of executing a `containerMutator` across a pod. If we absolutely needed it, we could add a `podContainerMutator(c containerMutator) podMutator` function that would wrap it, but I don't know that it's necessary.

pkg/clusteragent/admission/mutate/autoinstrumentation/env_vars.go
```diff
@@ -75,21 +101,32 @@ func (e envVar) mutateContainer(c *corev1.Container) error {
...
+	}
+
+	if found {
+		c.Env[index] = nextEnvVar
 	} else {
-		if c.Env[index].ValueFrom != nil {
-			return fmt.Errorf("%q is defined via ValueFrom", e.key)
-		}
-		c.Env[index].Value = e.valFunc(c.Env[index].Value)
+		c.Env = appendOrPrepend(nextEnvVar, c.Env, e.prepend)
```
knusbaum, line 120: I'm having trouble following this logic. Might be Friday-brain. I'll take another look, but I'm convinced we can make this simpler.

pkg/clusteragent/admission/mutate/autoinstrumentation/env_vars.go
```diff
@@ -134,3 +171,25 @@ func dotnetProfilingLdPreloadEnvValFunc(predefinedVal string) string {
...
+}
+
+func stringValFromPointer[T any](pointer *T, format func(T) string) string {
+	var val T
+	if pointer != nil {
+		val = *pointer
+	}
+
+	return format(val)
+}
```
knusbaum, line 191: Suggesting we drop the `format` argument to generalize.
```suggestion
func valueOrZero[T any](pointer *T) T {
	var val T
	if pointer != nil {
		val = *pointer
	}
	return val
}
```

Then we can do:
```
strconv.FormatBool(valueOrZero(ptr))
```
instead of 
```
stringValFromPointer(ptr, strconv.FormatBool)
```

And we can drop the `stringValue` function.

I don't see a reason to tie this to a particular format signature.

knusbaum (changes requested):
This is looking good.

A lot of this we talked about in our sync, but wanted to reproduce here.

pkg/clusteragent/admission/mutate/autoinstrumentation/mutators.go
stanistan: Removed this.

pkg/clusteragent/admission/mutate/autoinstrumentation/env_vars.go
stanistan: Documented it a bit better.

pkg/clusteragent/admission/mutate/autoinstrumentation/auto_instrumentation.go
stanistan: Updated this to produce a `containerMutator` so we can use `m.core.mutatePodContainers` which applies the filter _for us_.

knusbaum (approved):
Looks good, thanks for making those changes.

joepeeples (approved):

betterengineering (approved):
🚀

davidor (approved):
:+1: (only looked at files owned only by container-platform)

florianmutter:
Does this also prevent env variables like DD_SERVICE or DD_ENV from being added to the istio-proxy container?

PR #35790, opened 2025-04-03
Try to avoid reporting errors during system-probe startup

vitkyrka:
### What does this PR do?

During startup of the agent, the checks which need system-probe start before system-probe is up and this leads to error-level logs from the agent and the checks temporarily being listed as broken in Fleet Automation, and the errors reported are visible to the user.

To improve this, first check whether system-probe is up by hitting a generic endpoint, and if it's not, for the first few minutes, only report warnings, with a clearer message referring to the startup of system-probe.

### Motivation

Remove confusing error status and logs
https://datadoghq.atlassian.net/browse/DSCVR-93

Automated test added.


cmd/system-probe/api/client/client.go
```diff
@@ -63,31 +78,87 @@ func get(socketPath string) *http.Client {
...
+
+	req, err := http.NewRequest("GET", "http://sysprobe/debug/stats", nil)
+	if err != nil {
+		return err
+	}
+
+	_, err = doReq(c.client, req, "status")
+	if err != nil {
+		minutes := time.Since(c.uptime).Minutes()
+		if minutes < 5 {
```
justin-lesko, line 125: Would we ever want this grace period to be configurable?

L3n41c (approved):

justin-lesko (approved):

cmd/system-probe/api/client/client.go
```diff
@@ -63,31 +78,87 @@ func get(socketPath string) *http.Client {
...
+	c.started = true
+	return nil
+}
+
+// GetCheck returns data unmarshalled from JSON to T, from the specified module at the /<module>/check endpoint.
+func GetCheck[T any](check *corechecks.CheckBase, client *CheckClient, module types.ModuleName) (T, error) {
+	var data T
+	err := client.ensureStarted(check)
+	if err != nil {
+		return data, err
```
alexn-dd, line 148: Should this return nil data?

pkg/util/crashreport/crashreport.go
```diff
@@ -42,7 +42,7 @@ func NewWinCrashReporter(hive registry.Key, key string) (*WinCrashReporter, erro
 	wcr := &WinCrashReporter{
 		hive:           hive,
 		baseKey:        key,
-		sysProbeClient: sysprobeclient.Get(pkgconfigsetup.SystemProbe().GetString("system_probe_config.sysprobe_socket")),
+		sysProbeClient: sysprobeclient.GetCheckClient(pkgconfigsetup.SystemProbe().GetString("system_probe_config.sysprobe_socket")),
```
alexn-dd, line 45: Did these changes go through the Windows tests?

cmd/system-probe/api/client/client.go
```diff
@@ -15,9 +15,11 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/DataDog/datadog-agent/cmd/system-probe/config/types"
+	"github.com/DataDog/datadog-agent/pkg/collector/corechecks"
```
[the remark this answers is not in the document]
brycekahle: Maybe something like `pkg/system-probe/api/remoteclient`?

cmd/system-probe/api/client/client.go
```diff
@@ -63,31 +78,87 @@ func get(socketPath string) *http.Client {
...
 	if resp.StatusCode != http.StatusOK {
 		checkTelemetry.responseErrors.IncWithTags(map[string]string{checkLabelName: string(module)})
-		return data, fmt.Errorf("non-ok status code: url %s, status_code: %d, response: `%s`", req.URL, resp.StatusCode, string(body))
+		return nil, fmt.Errorf("non-ok status code: url %s, status_code: %d, response: `%s`", req.URL, resp.StatusCode, string(body))
+	}
+
+	return body, err
+}
+
+func (c *CheckClient) ensureStarted(check *corechecks.CheckBase) error {
```
brycekahle, line 109: There are also system-probe clients that are not checks. Do you intend to solve this for them as well?

pkg/util/crashreport/crashreport.go
vitkyrka: Yes, [TestWinCrashReporting](https://app.datadoghq.com/ci/test/runs?query=test_level%3Atest%20%40test.name%3ATestWinCrashReporting%20%40git.branch%3Avincent.whitchurch%2Fsp-notyet&agg_m=count&agg_m_source=base&agg_t=count&currentTab=overview&eventStack=&fromUser=false&graphType=flamegraph&index=citest&start=1743148633267&end=1743753433267&paused=true) was run automatically on the PR.

cmd/system-probe/api/client/client.go
vitkyrka: Do you mean that the return statement could be `return nil, err`? That doesn't work since `nil` is not valid for the generic type used in this function.

cmd/system-probe/api/client/client.go
vitkyrka: I was only planning on handling it for the checks, since the checks are special since the status get reported up to the UI.

cmd/system-probe/api/client/client.go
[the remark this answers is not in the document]
vitkyrka: I rebased it. I got rid of the corecheck usage in this package for now and just logged the warning as normal. If we want to log the warning via the corecheck (which would mean they get visible in the integration status in the UI), we can move the IgnoreStartupError to a separate package which is not imported from system-probe.

pkg/system-probe/api/client/client.go
```diff
@@ -63,31 +82,87 @@ func get(socketPath string) *http.Client {
...
+	}
+
+	return body, err
+}
+
+func (c *CheckClient) ensureStarted(module types.ModuleName) error {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+
+	if c.started {
```
brycekahle, line 118: minor nit, but it looks like each client is doing its own "started" determination, when we could probably do that for all clients.

brycekahle: It would need to properly vary on the socket path though, so maybe not worth the complication.

brycekahle:
Can you add a comment to where the `/debug/stats` handler is registered indicating that it is now load-bearing and required for core agent checks that talk to system-probe.

dustmop (approved):

pkg/system-probe/api/client/client.go
vitkyrka: All users with the same socket path are already using a single client thanks to the memoization in GetCheckClient, so this should be already taken care of?

vitkyrka:
> Can you add a comment to where the `/debug/stats` handler is registered indicating that it is now load-bearing and required for core agent checks that talk to system-probe?

Done.

Yumasi (approved):

brycekahle (approved):

alexn-dd (approved):

PR #35793, opened 2025-04-03
[CONTP-698] fix bug in EntityID.Empty method

adel121:
### What does this PR do?

Fixes EntityID.Empty method. 

It should return `true` if prefix is empty or if the id is empty. 

Previously, it was returning true only if both of them are empty.

### Motivation

The entityID is composed of two parts:
- Prefix: ideally, should never be empty.
- ID: a unique identifier of the entity within the related prefix (example: a pod ui unique identifies an entityID given that the prefix is `kubernetes_pod_uid`)

Updated unit tests.

We already have e2e covering the tagger.

### Additional Notes

I checked where this method is used, and it should never be posing an issue because we have no case where we try tagging with an empty entity id by mistake. 

For example, [here](https://github.com/DataDog/datadog-agent/blob/6b94a52dc19bf92e1874592b8eccc660e2b0bedc/pkg/collector/corechecks/ebpf/oomkill/oom_kill.go#L107-L113) and [here](https://github.com/DataDog/datadog-agent/blob/6b94a52dc19bf92e1874592b8eccc660e2b0bedc/pkg/collector/corechecks/ebpf/tcpqueuelength/tcp_queue_length.go#L101-L108) we are getting container tags. But if the container id is empty, an error is returned even before calling the tagger.


comp/core/tagger/types/entity_id.go
```diff
@@ -29,7 +29,7 @@ type EntityID struct {
 
 // Empty returns true if prefix and id are both empty strings
```
gabedos, line 30: nit: method comment should replace "and" with "or"

gabedos (approved):
LGTM, small comment on test

comp/core/tagger/types/entity_id.go
```diff
@@ -27,9 +27,9 @@ type EntityID struct {
 	id     string
 }
 
-// Empty returns true if prefix and id are both empty strings
+// Empty returns true if either the prefix or id empty strings
 func (eid EntityID) Empty() bool {
-	return eid.prefix == "" && eid.id == ""
+	return eid.prefix == "" || eid.id == ""
```
wdhif, line 32: Although I agree with the change, we might want to be careful about it as this could lead to unforeseen changes. WDYT cc @davidor ?

wdhif (changes requested):

comp/core/tagger/types/entity_id.go
wdhif: Blocked the PR for now to avoid merging until we discuss this.

adel121: @wdhif 

To explain a bit, `eid.prefix` should follow the enum defined [here](https://github.com/DataDog/datadog-agent/blob/24b2bd028c7167a72c1268b24073bfa9567dde5d/comp/core/tagger/types/entity_id.go#L60-L81). 

In theory, the prefix should never be empty, and we already impose this constraint [here](https://github.com/DataDog/datadog-agent/blob/24b2bd028c7167a72c1268b24073bfa9567dde5d/comp/core/tagger/types/entity_id.go#L53).

So in essence, the condition `eid.prefix == "" || eid.id == ""` is equivalent to `eid.id == ""`. 

Additionally, I have just realised we can make this even more strict by asserting that the prefix belongs to the enum, instead of just verifying that it is not empty. I will push a commit to have this in place.

wdhif: Seen together, we decided to wait for the next release to have this change run in nightly for as long as possible.

comp/core/tagger/types/entity_id.go
```diff
@@ -27,9 +27,9 @@ type EntityID struct {
 	id     string
 }
 
-// Empty returns true if prefix and id are both empty strings
+// Empty returns true if either the prefix or id empty strings
 func (eid EntityID) Empty() bool {
```
wdhif, line 31: nit: We might also want to add unit tests for the `Empty` method in case someone would try to create an `EntityID` without using the `NewEntityID` method.

wdhif (approved):

comp/core/tagger/types/entity_id.go
adel121: I think it is ok not to have unit tests for this considering that it is only a simple one-line function, we will be testing the OR operator.

PR #35796, opened 2025-04-03
[AGENTRUN-201] Update RTLoader to use PyConfig

louis-cqrl:
### What does this PR do?

This PR aims to refactor RTLoader to use PyConfig to support python 3.13+ versions

### Motivation

Remove any usage of deprecated function in python 3.13+
Using PyConfig also enable the usage of python library such as `multiprocessing`

tests and CI
Manual QA (on windows and linux):
- Created a python check:
```
import os
import time
from multiprocessing import Process

def write_to_file(file_path, data, write_count=5, delay=1):
    """
    Writes the specified data to the given file multiple times with delays,
    printing process information to demonstrate concurrent execution.
    """
    start_time = time.time()
    print(f"[Process {os.getpid()}] Starting to write to {file_path} at {time.strftime('%X')}")
    
    try:
        with open(file_path, 'a') as f:
            for i in range(write_count):
                message = f"{data} - Line {i+1} (at {time.strftime('%X')} to PID: {os.getpid()})"
                f.write(message + "\n")
                f.flush()
                print(f"[Process {os.getpid()}] {message}")
                time.sleep(delay)
        elapsed = time.time() - start_time
        print(f"[Process {os.getpid()}] Finished writing to {file_path} at {time.strftime('%X')} (Elapsed: {elapsed:.2f} seconds)")
    except Exception as e:
        print(f"[Process {os.getpid()}] Error writing to {file_path}: {str(e)}")

class MultiFileCheck(AgentCheck):
    def check(self, instance):
        """
        Spawns 5 processes that concurrently write to 5 different files.
        """
        file_paths = [
            "/tmp/file1.txt",
            "/tmp/file2.txt",
            "/tmp/file3.txt",
            "/tmp/file4.txt",
            "/tmp/file5.txt"
        ]
        processes = []
        for idx, file_path in enumerate(file_paths, start=1):
            data = f"Datadog Check Data for File {idx}"
            p = Process(target=write_to_file, args=(file_path, data))
            processes.append(p)
            p.start()
        
        for p in processes:
            p.join()

        self.log.info("MultiFileCheck completed writing to all files.")
```
- Run the Agent
- Ensure the check is ran with `datadog-agent` status
- Ensure PID is different in each files
```
Datadog Check Data for File 2 - Line 5 (at 10:19:05 for PID: 364)
Datadog Check Data for File 4 - Line 5 (at 10:19:05 for PID: 2288)
Datadog Check Data for File 3 - Line 5 (at 10:19:05 for PID: 4260)
Datadog Check Data for File 1 - Line 5 (at 10:19:05 for PID: 5020)
Datadog Check Data for File 5 - Line 5 (at 10:19:05 for PID: 4712)
```

This should not have any effect on the existing processes

### Additional Notes

For the QA, run a VM on windows and linux:
- Ensure python is loaded
- Ensure python checks are running and working 
- Create a python check using `multiprocessing` library
- Ensure the custom check is working as expected


rtloader/three/three.cpp
```diff
@@ -34,19 +34,34 @@ extern "C" DATADOG_AGENT_RTLOADER_API void destroy(RtLoader *p)
...
     , _pythonPaths()
     , _pymallocPrev{ 0 }
     , _pymemInuse(0)
     , _pymemAlloc(0)
 {
-    initPythonHome(python_home);
+    PyStatus status;
+    // Initialize the configuration with default values
+    PyConfig_InitPythonConfig(&_config);
+    _config.install_signal_handlers = 1;
```
pgimalac, line 46: I think `1` is already the default value for the non-isolated mode ? https://docs.python.org/3/c-api/init_config.html#c.PyConfig.install_signal_handlers
But we can set that explicitly, not a big deal

rtloader/three/three.cpp
```diff
@@ -110,15 +96,13 @@ bool Three::init()
...
-    if (!Py_IsInitialized()) {
-        setError("Python not initialized");
-        return false;
+    // Initialize Python with our configuration
+    status = Py_InitializeFromConfig(&_config);
+    if (PyStatus_Exception(status)) {
+        setError("Failed to initialize Python: " + std::string(status.err_msg));
     }
+    // Clean up the configuration
+    PyConfig_Clear(&_config);
```
pgimalac, line 105: Why do we clear the config here ?

rtloader/three/three.cpp
pgimalac, line 1: Should we use the isolated mode ? https://docs.python.org/3/c-api/init_config.html#isolated-configuration

rtloader/three/three.cpp
louis-cqrl: it was set explicitly before, I would rather stick to the same information in the code

rtloader/three/three.cpp
louis-cqrl: I was wondering the same thing, I think the reviewer from `agent-metric-pipelines` will have a stronger opinion on this 🙇🏻

rtloader/three/three.cpp
```diff
@@ -58,50 +73,21 @@ Three::~Three()
...
-bool Three::init()
-{
-    // we want the checks to be runned with the standard encoding utf-8
-    // setting this var to 1 forces the UTF8 mode for CPython >= 3.7
-    // See:
-    //	- PEP UTF8 mode https://www.python.org/dev/peps/pep-0540/
-    //	- about this var https://github.com/python/cpython/pull/12589
-    // This has to be set before the Py_Initialize() call.
-    Py_UTF8Mode = 1;
+    status = Py_PreInitialize(&preconfig);
```
vickenty, line 84: Pe-init would be done earlier implicitly by a call to [PyConfig_SetBytesString](https://docs.python.org/3/c-api/init_config.html#c.PyConfig_SetBytesString). Is this call still effective?

louis-cqrl: Good catch, applied it

rtloader/three/three.cpp
```diff
@@ -34,19 +34,44 @@ extern "C" DATADOG_AGENT_RTLOADER_API void destroy(RtLoader *p)
...
-    initPythonHome(python_home);
+    // Pre-Initialize Python with UTF-8 mode
+    PyStatus status;
+    PyPreConfig preconfig;
+    PyPreConfig_InitPythonConfig(&preconfig);
+    preconfig.utf8_mode = 1;
+
+    status = Py_PreInitialize(&preconfig);
+    if (PyStatus_Exception(status)) {
+        Py_ExitStatusException(status);
```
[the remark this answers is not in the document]
louis-cqrl: I was questioning my self on that as well, @pgimalac thoughts are to mot exit the agent in case of failure, I will remove it

rtloader/three/three.cpp
```diff
@@ -34,19 +34,50 @@ extern "C" DATADOG_AGENT_RTLOADER_API void destroy(RtLoader *p)
...
-    initPythonHome(python_home);
+    // Pre-Initialize Python with UTF-8 mode
+    PyStatus status;
+    PyPreConfig preconfig;
+    PyPreConfig_InitIsolatedConfig(&preconfig);
+    preconfig.utf8_mode = 1;
+
+    status = Py_PreInitialize(&preconfig);
+    if (PyStatus_Exception(status)) {
+        setError("Failed to pre-initialize Python: " + std::string(status.err_msg));
```
vickenty, line 51: I don't think anything checks for an error set this weay when we call the constructor. Would it be possible to add a test for this code path?

rtloader/three/three.cpp
louis-cqrl: We might consider using an environment variable to disable the GIL in the Agent. However, since we may want customers to have the option to enforce the GIL by overriding our setting, it’s best to stick with the default mode.

rtloader/three/three.cpp
```diff
@@ -58,47 +56,64 @@ Three::~Three()
...
+    _config.install_signal_handlers = 1;
+
+    // Set Python home
+    if (!status.check(PyConfig_SetBytesString(&_config, &_config.home, _pythonHome), "Failed to set python home")) {
+        PyConfig_Clear(&_config);
+        return false;
+    }
+
+    // Configure Python executable if provided
+    if (_pythonExe) {
```
pgimalac, line 108: I think it's supposed to always be provided ? (or at least in the previous version it didn't check if it was ?)

pgimalac (approved):
Some nitpicks but LGTM overall

rtloader/three/three.cpp
```diff
@@ -58,47 +56,64 @@ Three::~Three()
...
-    } else {
-        _pythonHome = Py_DecodeLocale(pythonHome, NULL);
-    }
+    class StatusHandler
+    {
+    public:
+        explicit StatusHandler(Three *ctx)
+            : context(ctx)
+        {
+        }
```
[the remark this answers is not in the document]
GustavoCaso: I agree with @pgimalac I had to check the code multiple times to understand what are we doing. I think a private `check` method would simplify things

rtloader/three/three.cpp
GustavoCaso: That is correct 

```c++
extern "C" DATADOG_AGENT_RTLOADER_API RtLoader *create(const char *python_home, const char *python_exe,
                                                       cb_memory_tracker_t memtrack_cb)
{
    return new Three(python_home, python_exe, memtrack_cb);
}
```

We call that from:

```c++
rtloader_t *make3(const char *python_home, const char *python_exe, char **error)
{
    if (rtloader_backend != NULL) {
        *error = strdupe("RtLoader already initialized!");
        return NULL;
    }

    create_t *create_three = loadAndCreate(DATADOG_AGENT_THREE, python_home, error);
    if (!create_three) {
        return NULL;
    }
    return AS_TYPE(rtloader_t, create_three(python_home, python_exe, _get_memory_tracker_cb()));
}
```

on the Go side 
```go
if pythonVersion == "3" {
	log.Infof("Initializing rtloader with Python 3 %s", PythonHome)
	rtloader = C.make3(csPythonHome, csPythonExecPath, &pyErr)
} else {
	return addExpvarPythonInitErrors(fmt.Sprintf("unsuported version of python: %s", pythonVersion))
}
```

GustavoCaso (changes requested):
Is going on a good direction, but I think the code for checking if this are initialize correctly is a bit to complex, I think we can simplify it a bit 😄

rtloader/three/three.cpp
```diff
@@ -42,11 +42,9 @@ Three::Three(const char *python_home, const char *python_exe, cb_memory_tracker_
     , _pymemInuse(0)
     , _pymemAlloc(0)
 {
-    initPythonHome(python_home);
-
-    // If not empty, set our Python interpreter path
+    _pythonHome = (python_home && strlen(python_home) > 0) ? strdup(python_home) : strdup(_defaultPythonHome);
     if (python_exe && strlen(python_exe) > 0) {
-        initPythonExe(python_exe);
+        _pythonExe = strdup(python_exe);
```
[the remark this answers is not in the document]
louis-cqrl: used std::string for smoother memory management

rtloader/test/rtloader/rtloader_test.go
```diff
@@ -96,6 +96,57 @@ with open(r'%s', 'w') as f:
...
+
+	code := fmt.Sprintf(`
+import sys
+with open(r'%s', 'w') as f:
+ f.write(sys.prefix)`, tmpfile.Name())
+
+	output, err := runString(code)
+	if err != nil {
+		t.Fatalf("`test_sys_home` error: %v", err)
+	}
```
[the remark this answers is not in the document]
louis-cqrl: let's stick to the same format across all the tests for rtloader

rtloader/three/three.cpp
louis-cqrl: it was [check previously](https://github.com/DataDog/datadog-agent/blob/95c9aa1dedd13585682db851a3b1ea77b53df1b5/rtloader/three/three.cpp#L48), let's stick on the same code behaviour as before.

(Note: i think this is due to the mock for the test which I do not really understand, without this check it makes the tests crash for a reason I do not know)

pgimalac (approved):

rtloader/three/three.cpp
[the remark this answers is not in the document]
GustavoCaso: Thanks for removing the custom class.

rtloader/three/three.cpp
```diff
@@ -58,47 +55,50 @@ Three::~Three()
...
+        return true;
+    };
 
-void Three::initPythonExe(const char *python_exe)
-{
-    // Py_SetProgramName stores a pointer to the string we pass to it, so we must keep it in memory
-    wchar_t *oldPythonExe = _pythonExe;
-    _pythonExe = Py_DecodeLocale(python_exe, NULL);
+    // Initialize UTF-8 mode
+    {
```
GustavoCaso, line 70: Do we need to open a new scope here?

louis-cqrl: i like the idea of having a different scope for the preinit since it makes the code easier to read between the init and the preinit, but not mandatory at all i can remove it ?

rtloader/three/three.cpp
```diff
@@ -58,47 +55,50 @@ Three::~Three()
...
+        PyPreConfig preconfig;
+        PyPreConfig_InitPythonConfig(&preconfig);
+        preconfig.utf8_mode = 1;
 
-    Py_SetProgramName(_pythonExe);
+        if (!checkStatus(Py_PreInitialize(&preconfig), "Failed to pre-initialize Python")) {
```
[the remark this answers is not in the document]
GustavoCaso: > I named it checkStatus since the config functions return a PyStatus

Fair point. What about `checkConfigStatus` ?

rtloader/three/three.cpp
GustavoCaso: Sounds good

GustavoCaso (approved):
LGTM

rtloader/three/three.cpp
vickenty: Do we need to clear it on success as well? Python docs seem to imply so.

rtloader/three/three.cpp
```diff
@@ -58,47 +55,50 @@ Three::~Three()
...
-    // See:
-    //	- PEP UTF8 mode https://www.python.org/dev/peps/pep-0540/
-    //	- about this var https://github.com/python/cpython/pull/12589
-    // This has to be set before the Py_Initialize() call.
-    Py_UTF8Mode = 1;
+    // Configure Python executable if provided
+    if (!_pythonExe.empty()) {
+        if (!checkConfigStatus(PyConfig_SetBytesString(&_config, &_config.executable, _pythonExe.c_str()),
+                               "Failed to set executable path")
+            || !checkConfigStatus(PyConfig_SetBytesString(&_config, &_config.program_name, _pythonExe.c_str()),
```
vickenty, line 96: Would it be possible to document why we need to set both? Python examples seem to only set `program_name`, but not `executable`.

rtloader/three/three.cpp
louis-cqrl: > Do we need to clear it on success as well?

I think so, python doc seems to indicate to free it whenever we are not using pyconfig anymore

rtloader/three/three.cpp
```diff
@@ -58,47 +55,50 @@ Three::~Three()
...
-    if (pythonHome == NULL || strlen(pythonHome) == 0) {
-        _pythonHome = Py_DecodeLocale(_defaultPythonHome, NULL);
-    } else {
-        _pythonHome = Py_DecodeLocale(pythonHome, NULL);
-    }
-
-    Py_SetPythonHome(_pythonHome);
-    PyMem_RawFree((void *)oldPythonHome);
-}
+    std::function<bool(const PyStatus &, const std::string &)> checkConfigStatus
```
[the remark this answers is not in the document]
GustavoCaso: I agree with Vikenty. I find a private method cleaner, and it provides the same functionality

rtloader/three/three.cpp
louis-cqrl: In the code before, both Py_SetProgramName and _Py_SetProgramFullPath are used to ensure that the interpreter receives the correct path for both program name and executable.

```
void Three::initPythonExe(const char *python_exe)
{
    // Py_SetProgramName stores a pointer to the string we pass to it, so we must keep it in memory
    wchar_t *oldPythonExe = _pythonExe;
    _pythonExe = Py_DecodeLocale(python_exe, NULL);

    Py_SetProgramName(_pythonExe);

    // HACK: This extra internal API invocation is due to the workaround for an upstream bug on
    // Windows (https://bugs.python.org/issue34725) where just using `Py_SetProgramName` is
    // ineffective. The workaround API call will be removed at some point in the future (Python
    // 3.12+) so we should convert this initialization to the new`PyConfig API`
    // (https://docs.python.org/3.11/c-api/init_config.html#c.PyConfig) before then.
    _Py_SetProgramFullPath(_pythonExe);

    PyMem_RawFree((void *)oldPythonExe);
}
```

With the new PyConfig API, we explicitly set both the executable and the program_name fields in the config structure, achieving the same effect.

I can add a comment in the same style as we did before if you want too ?

rtloader/three/three.cpp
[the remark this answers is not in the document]
vickenty: There is nothing wrong with a method that is useful only in some context. `PyStatus` is already specific to python's initializtion, there is little chance to (mis)use the method for anything else.

> We can always not make it with a lambda or a method, but just check it after each PyConfig function call, it would also works.

Personally, I think that would be the most preferred option. This would be consistent with surrounding code, and would look more familiar to go developers. There are only 4 calls that need to be checked, I don't think we need to optimize for brevity here.

rtloader/three/three.cpp
vickenty: The comment above `_Py_SetProgramFullPath` says it was a workaround for an issue. This issue was fixed by the new configuration API, so IIUIC we don't need the workaround any more, and use simpler init code. So if we still need to set both, the comment should address why the new api is still not working as expected.

vickenty:
This needs manual QA, since there aren't e2e tests on windows that validate python as far as I could see, and the code being modified was a workaround for a windows issue. 

The `no-code-change` label is probably wrong too.

louis-cqrl:
> This needs manual QA, since there aren't e2e tests on windows that validate python as far as I could see, and the code being modified was a workaround for a windows issue.
> 
> The `no-code-change` label is probably wrong too.

Manual QA is done and described in the description, let's keep a QA in the RC for a double check. @vickenty

rtloader/three/three.cpp
louis-cqrl: made some tests, it seems to have issue on windows unit test, let's for now keep it since it does not change any behaviour and see in an other PR if we can remove this workaround, are you okay with that ?

vickenty (approved):

louis-cqrl:

PR #35805, opened 2025-04-03
[CWS] reduce footprint of SetFieldValue

safchain:


pkg/security/generators/accessors/accessors.tmpl
```diff
@@ -297,7 +298,129 @@ func (ev *Event) GetFieldMetadata(field eval.Field) (eval.EventType, reflect.Kin
 	return "", reflect.Invalid, "", &eval.ErrFieldNotFound{Field: field}
 }
 
+func (ev *Event) setStringArray(field string, fv *[]string, value interface{}) error {
```
paulcacheux, line 301: Do we really need those in the template ? They could be moved to a real go file no?

paulcacheux (approved):
Nice ! Thanks

spikat (approved):

PR #35833, opened 2025-04-04
[AGENTCFG-192] Warnings in config should be displayed during startup sequence

rahulkaukuntla:
### What does this PR do?
We actually want to report the errors that we receive as warnings instead of returning nil.

Added a test


paulcacheux (approved):

pkg/config/nodetreemodel/config_test.go
```diff
@@ -886,3 +887,18 @@ func TestEnvVarOrdering(t *testing.T) {
 		assert.Equal(t, "https://app.datadoghq.dd_url.eu", config.GetString("dd_url"))
 	})
 }
+
+func TestWarningLogged(t *testing.T) {
+	cfg := NewNodeTreeConfig("test", "TEST", strings.NewReplacer(".", "_"))
+	t.Setenv("DD_CONF_NODETREEMODEL", "enable")
```
hush-hush, line 893: I don't think we need this inside the nodetreemodel package

hush-hush (changes requested):

pkg/config/nodetreemodel/config_test.go
rahulkaukuntla: removed that line.

hush-hush (approved):

PR #35836, opened 2025-04-04
Show static quality gates relative size and provide a tool to handle exceptions

Pythyu:
### What does this PR do?
Allow the PR notification to show the relative size variation of each gates.
Add a tool to edit static quality gates to their correct value when a PR is exempted of static quality gates.

### Additional Notes
This PR needs #36110 merged on main to function


tasks/quality_gates.py
```diff
@@ -245,3 +250,56 @@ def notify_threshold_update(pr_url):
...
+
+
+@task()
+def exception_threshold_bump(ctx):
+    """
+    When a PR is exempt of static quality gates, they have to use this invoke task to adjust the quality gates thresholds accordingly to the exempted added size.
+
+    Note: This invoke task must be run on a pipeline that has finished running static quality gates
+    :param ctx:
+    :param pipeline_id: Pipeline ID of the latest commit of the exempted PR
```
Pythyu, line 262:
```suggestion
```
Will be removed after the current pipeline ran once

tasks/quality_gates.py
```diff
@@ -32,6 +41,17 @@
...
 
+def get_formated_relative_size(relative_disk_size, relative_wire_size):
+    if relative_disk_size == "DataNotFound" or relative_wire_size == "DataNotFound":
+        return "N/A", "N/A"
+
+    if string_to_byte(relative_disk_size) >= 0:
+        relative_disk_size = f"+{relative_disk_size}"
+    if string_to_byte(relative_wire_size) >= 0:
+        relative_wire_size = f"+{relative_wire_size}"
+    return relative_disk_size, relative_wire_size
```
chouquette, line 52: nit: I think this can be done with built in formatters: `print(f'{0:+}', relative_disk_size)`

tasks/quality_gates.py
```diff
@@ -247,3 +277,57 @@ def notify_threshold_update(pr_url):
...
+
+    Note: This invoke task must be run on a pipeline that has finished running static quality gates
+    :param ctx:
+    :return:
+    """
+    current_branch_name = ctx.run("git rev-parse --abbrev-ref HEAD", hide=True).stdout.strip()
+    ancestor_commit = get_common_ancestor(ctx, "HEAD")
+    repo = get_gitlab_repo()
+    with tempfile.TemporaryDirectory() as extract_dir, ctx.cd(extract_dir):
+        with open(f"{extract_dir}/gate_archive.zip", "wb") as f:
```
chouquette, line 295: Is the `extract_dir` needed? We've just `cd`'ed into it in the line above, so this is generating `extract_dir/extract_dir/gate_archive.zip` if I'm not mistaking, which is a bit confusing IMO

tasks/static_quality_gates/lib/gates_lib.py
```diff
@@ -136,6 +136,31 @@ def _add_gauge(self, timestamp, common_tags, gate, metric_name, metric_key):
...
+                    if not ancestor_gate:
+                        continue
+                    for metric_key in ["current_on_wire_size", "current_on_disk_size"]:
+                        if self.metrics.get(metric_key) and ancestor_gate.get(metric_key):
+                            relative_metric_size = ancestor_gate[metric_key] - self.metrics[gate][metric_key]
+                            self.register_metric(gate, metric_key.replace("current", "relative"), relative_metric_size)
+            else:
+                print(
+                    color_message(
+                        f"[WARN] Unable to fetch quality gates {filename} from {ancestor} !",
```
chouquette, line 159: I think it would be better to include the error logs to ease debugging in case this happens

tasks/quality_gates.py
Pythyu: I thought so too at first but after getting some errors I saw that `ctx.cd` is a wrapper to `ctx.run` which add a `cd /some/true/path &&` to the run command. Every other functions that are unrelated to `invoke` will still need to be provided the path directly hence giving `extract_dir`.

tasks/quality_gates.py
Pythyu: Finally I can't use it because it works only on integers not string and I don't want to do 
`f"{string_to_byte(relative_wire_size):+}"` into `byte_to_string` to convert it back into `XY.ZZ MiB`

tasks/quality_gates.py
chouquette: Given there's a single `ctx.run` that's affected by the `ctx.cd`, I think it would be clearer to just cd before it, or include the cd/factore in the subfolder as part of the command directly

chouquette (approved):

tasks/quality_gates.py
Pythyu: metric_handler.generate_relative_size(ctx, ancestor=ancestor_commit)` also uses it. I could do a with within a with to just have this part within the `cd` but I this point it feels like currently is pretty much the same
```python
    with ctx.cd(extract_dir):
        ctx.run("unzip gate_archive.zip", hide=True)
        if os.path.isfile(f"{extract_dir}/static_gate_report.json"):
            metric_handler = GateMetricHandler(
                git_ref=current_branch_name, bucket_branch="dev", filename=f"{extract_dir}/static_gate_report.json"
            )
            metric_handler.generate_relative_size(ctx, ancestor=ancestor_commit)
            with open("test/static/static_quality_gates.yml") as f:
                file_content, total_size_saved = generate_new_quality_gate_config(f, metric_handler, True)

            if total_size_saved == 0:
                print(color_message("[WARN] No gates needs to be changed.", "orange"))

            with open("test/static/static_quality_gates.yml", "w") as f:
                f.write(yaml.dump(file_content))

            print(
                color_message(
                    f"[SUCCESS] Static Quality gate have been updated ! Total gate threshold impact : {byte_to_string(-total_size_saved)}",
                    "green",
                )
            )
        else:
            print(
                color_message(
                    "[ERROR] Unable to find static_gate_report.json inside of the last artifact of the static_quality_gates job",
                    "red",
                )
            )
```

chouquette: I don't think the dedicated `with` changes much since `extract` dir is still mentionned in other paths, but a comment is probably enough to clarify

tasks/quality_gates.py
```diff
@@ -260,3 +286,58 @@ def manual_threshold_update(self, filename="static_gate_report.json"):
...
+@task()
+def exception_threshold_bump(ctx):
+    """
+    When a PR is exempt of static quality gates, they have to use this invoke task to adjust the quality gates thresholds accordingly to the exempted added size.
+
+    Note: This invoke task must be run on a pipeline that has finished running static quality gates
+    :param ctx:
+    :return:
+    """
+    current_branch_name = ctx.run("git rev-parse --abbrev-ref HEAD", hide=True).stdout.strip()
```
chouetz, line 300:
```suggestion
    current_branch_name = get_current_branch()
```

tasks/quality_gates.py
```diff
@@ -260,3 +286,58 @@ def manual_threshold_update(self, filename="static_gate_report.json"):
...
+            try:
+                f.write(repo.artifacts.download(ref_name=current_branch_name, job="static_quality_gates"))
+            except gitlab.exceptions.GitlabGetError as e:
+                print(
+                    color_message(
+                        "[ERROR] Unable to fetch the last artifact of the static_quality_gates job. Details :", "red"
+                    )
+                )
+                print(repr(e))
+                return
```
chouetz, line 314:
```suggestion
                raise Exit(code=1)
```
If we want to ignore the error I suppose we should say it, we should not silent the error otherwise.

tasks/static_quality_gates/lib/gates_lib.py
```diff
@@ -30,14 +30,27 @@ def argument_extractor(entry_args, **kwargs) -> SimpleNamespace:
     return SimpleNamespace(**kwargs)
 
 
-def byte_to_string(size):
```
chouetz, line 33: Could we use [this module](https://github.com/ofek/binary) to replace partially `byte_to_string` and `string_to_byte`?

tasks/static_quality_gates/lib/gates_lib.py
```diff
@@ -30,14 +30,27 @@ def argument_extractor(entry_args, **kwargs) -> SimpleNamespace:
...
     s = round(size / p, 2)
-    return f"{s} {size_name[i]}"
+    # If s is not exactly 0 but rounded like (0.0 or -0.0)
+    # Goal is to output +0 / -0 for very small changes and 0 for no changes at all
+    if id(s) != id(0) and s == 0:
+        s = 0
+    return f"{sign}{s}{' '+size_name[unit_power] if with_unit else ''}"
+
+
+def string_to_latex_color(text, latex_color):
```
chouetz, line 52: why do we need this?

chouetz (commented):
Can you please confirm about the silent errors before I approve? Thanks in advance.

tasks/static_quality_gates/lib/gates_lib.py
Pythyu: Because the PR notification now uses latex colors (it saves horizontal space)

tasks/quality_gates.py
```diff
@@ -260,3 +286,58 @@ def manual_threshold_update(self, filename="static_gate_report.json"):
...
+                    "green",
+                )
+            )
+        else:
+            print(
+                color_message(
+                    "[ERROR] Unable to find static_gate_report.json inside of the last artifact of the static_quality_gates job",
+                    "red",
+                )
+            )
```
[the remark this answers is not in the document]
Pythyu: I don't remember why I didn't want to raise errors in this function but since I don't I'm adding them back 👍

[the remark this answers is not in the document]
chouetz: I hope this will not trigger an unexpected side effect, thanks a lot

chouetz (approved):

PR #35838, opened 2025-04-04
Add telemetry around IPC authentication

misteriaud:
### What does this PR do?
This PR:
1. updates the IPC client-side helper to use IPC certificate
2. Adds a new tag to the cross-org telemetry (`api_server.request_duration_seconds.auth`)
3. updates the API Telemetry Middleware to set `mTLS` tag if the client certificate is the same as the server certificate 

### Motivation
This PR introduces a new `auth` tag to the  Agent telemetry. The tag will help evaluate the impact of renewing IPC authentication system, while ensuring minimal disruption to end-users.

This change is covered by unit-tests.


comp/api/api/apiimpl/observability/telemetry.go
```diff
@@ -49,26 +54,56 @@ func (th *telemetryMiddlewareFactory) Middleware(serverName string) mux.Middlewa
 			next.ServeHTTP(w, r)
 
 			path := extractPath(r)
-			th.requestDuration.Observe(duration.Seconds(), serverName, strconv.Itoa(statusCode), r.Method, path)
+
+			durationSeconds := duration.Seconds()
+
+			// We can assert that the auth is at least a token because it have been checked previously by the validateToken middleware
```
pgimalac, line 60:
```suggestion
			// We can assert that the auth is at least a token because it has been checked previously by the validateToken middleware
```

comp/api/api/apiimpl/observability/telemetry.go
```diff
@@ -49,26 +54,56 @@ func (th *telemetryMiddlewareFactory) Middleware(serverName string) mux.Middlewa
...
+
+	if serverTLSConfig.Certificates[0].Leaf != nil {
+		ipcCert = serverTLSConfig.Certificates[0].Leaf.Raw
+	} else {
+		cert, err := x509.ParseCertificate(serverTLSConfig.Certificates[0].Certificate[0])
+		if err != nil {
+			return nil, fmt.Errorf("error parsing IPC certificate: %v", err)
+		}
+		ipcCert = cert.Raw
+	}
```
pgimalac, line 94: I think this logic should be in the tls config (possibly using a wrapper to add a new method), it seems way too specific and low-level for an observability middleware

misteriaud: I agree, I moved the logic to `server.go` to avoid having it in telemetry.
We may move it in a util/helper file for a broader use if the need arise.

michaelcretzman (approved):

comp/api/api/apiimpl/observability/telemetry.go
```diff
@@ -49,26 +51,36 @@ func (th *telemetryMiddlewareFactory) Middleware(serverName string) mux.Middlewa
...
 			path := extractPath(r)
-			th.requestDuration.Observe(duration.Seconds(), serverName, strconv.Itoa(statusCode), r.Method, path)
+
+			durationSeconds := duration.Seconds()
+
+			// We can assert that the auth is at least a token because it has been checked previously by the validateToken middleware
+			auth := "token"
+			if th.cert != nil && r.TLS != nil && len(r.TLS.PeerCertificates) > 0 && th.cert.Equal(r.TLS.PeerCertificates[0]) {
+				auth = "mTLS"
+			}
```
pgimalac, line 61: **🥜 nitpick:** ‏This logic makes assumptions on what happens in the server (ie. a previous middleware checked the token), and does some low level checks
I think instead of taking the expected cert as parameter, the middleware could just take a `func(http.Request) string` which would return the auth kind for a request
This func would be created in the API, which makes more sense (no more assumptions here, and no dependency on `crypto/x509`)

pgimalac (approved):
Just one nitpick but overall LGTM

iglendd (approved):
Left a minor comment

comp/api/api/apiimpl/observability/telemetry.go
pgimalac: This middleware emits a metric for each request, it has a dependency on the telemetry component, it doesn't handle the telemetry itself (it just emits some)
Maybe the name is not the best haha, if you have another proposal we can update 😄

pgimalac: I mean the package is named `observability`, it contains two files, one for emitting metrics and one for emitting logs (two middlewares)
Maybe replacing "telemetry" with "metric" would be clearer ?

misteriaud: I agree that having the telemetry middleware agnostic as much as possible of the API implementation is better! 
I moved the certificate verification into `server.go` to have clearer separation of concerns. At the same time I moved the related tests to a new file called `server_test.go`.

PR #35840, opened 2025-04-04
fix(assess_permissions): Rework the assess_permissions workflow

chouetz:
### What does this PR do?
Rework the assess_permissions workflow:
- Run on main repository and only list contributors and teams.
- Add an additional run on `agent-all`:
  - List the teams belonging to `agent-all`
  - List the maintainers
  - List sub-teams with no contributor: these teams can be removed from agent-all
  - List members of the `agent-contributor` team with no contribution: these contributors can be removed from this team
  - List other members with no contribution. These ones must be assessed case by case.

### Motivation
https://datadoghq.atlassian.net/browse/ACIX-577

Manual tests in the #agent-notification-testing slack channel, and unit tests in the PR

### DIFF
--- a/tasks/libs/common/slack.py
+++ b/tasks/libs/common/slack.py
@@ -1,3 +1,6 @@
+from tasks.libs.types.types import PermissionCheck
+
+
 def header_block(text: str) -> dict:
     """
     Create a Slack block with a header.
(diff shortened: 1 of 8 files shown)


tasks/libs/common/slack.py
```diff
@@ -0,0 +1,81 @@
+from tasks.libs.types.types import PermissionCheck
```
CelianR, line 1: Perfect 👌

tasks/libs/common/slack.py
```diff
@@ -0,0 +1,81 @@
...
+
```
CelianR, line 2: Should we add a `send_slack_message` function here such that we don't have to do the nested import in our functions?

CelianR (approved):

PR #35842, opened 2025-04-04
APM: Begin introducing new Trace Payload format

ajgajg1134:
### What does this PR do?
This PR begins the process of implementing the Efficient Trace Payload. This initial PR implements the Unmarshalling code for these new payloads - note here that the wire format is different from the new Internal* types. NOTE: I made use of the Cursor IDE here but all code was reviewed and modified by me.

### Motivation
See the [RFC](https://docs.google.com/document/d/1hNS6anKYutOYW-nmR759UlKXUdT6H0mRwVt7_L70ESc/edit?tab=t.0#heading=h.yzp9l1fml007) for full motivation on the changes.

For the motivation of why this PR is the way it is - to avoid a single MASSIVE PR we have decided to implement and review changes in as small of chunks as we can.

There are a number of unit tests here as well as fuzz tests. I ran each of the fuzz tests for at least 5 minutes (after fixing the issues they initially found) And no subsequent issues were found.

This code does nothing yet but adds a bit of binary size, this increase in size will be well worth the improvements here and should be small overall.


pkg/proto/datadog/trace/idx/span.proto
```diff
@@ -0,0 +1,153 @@
+syntax = "proto3";
+
+package datadog.trace.idx;
+
+option go_package="pkg/proto/pbgo/trace/idx";
+
+message AnyValue {
+	oneof value {
+		// @gotags: json:"string_value" msg:"string_value"
+		uint32 stringValue = 1;
```
ichinaski, line 10: Maybe we can be more descriptive about this field name, making it more obvious that it's a String reference/key/othername? (eg `stringRef` or similar)

It would also leave room, if needed in the future, to have an actual `string stringValue` (just in case)

pkg/proto/datadog/trace/idx/span.proto
```diff
@@ -0,0 +1,153 @@
...
+    fixed64 start = 6;
+    // duration is the time length of this span in nanoseconds.
+    // @gotags: json:"duration" msg:"duration"
+    uint64 duration = 7;
+    // if there is an error associated with this span
+    // @gotags: json:"error" msg:"error"
+    bool error = 8;
+    // meta is a mapping from tag name to tag value for string-valued tags.
+    // @gotags: json:"attributes,omitempty" msg:"attributes,omitempty"
+    repeated KeyValue attributes = 9;
```
ichinaski, line 101: I recall talking about this at some point, but forgot the exact discussion 😅 

Isn't there a way to preserve the map-like structure of the existing `meta`/`metrics` fields? (eg map of keys to list of values if multiple values per key are needed). I worry about linear data access the repeated field would enforce.

pkg/proto/datadog/trace/idx/tracer_payload.proto
```diff
@@ -0,0 +1,70 @@
...
+	// @gotags: json:"app_version" msg:"app_version"
+	uint32 appVersion = 10;
+	// a collection of key to value pairs common in all `chunks`
+	// @gotags: json:"attributes" msg:"attributes"
+	repeated KeyValue attributes = 11;
+	// chunks specifies list of containing trace chunks.
+	// @gotags: json:"chunks" msg:"chunks"
+	repeated TraceChunk chunks = 12;
+
+}
```
ichinaski, line 70: We probably need an additional `AgentPayload` message to wrap the payloads we'll send to the Intake?

pkg/proto/pbgo/trace/idx/internal_span.go
```diff
@@ -0,0 +1,148 @@
...
+
+// Add adds a string to the string table if it doesn't already exist and returns the index of the string
+func (s *StringTable) Add(str string) uint32 {
+	if idx, ok := s.lookup[str]; ok {
+		return idx
+	}
+	return s.addUnchecked(str)
+}
+
+// Get returns the string at the given index - panics if out of bounds
```
ichinaski, line 37: to ease the access, maybe the signature and implementation could account for missing strings by returning `(string, error)` or `(string, bool)`?

pkg/proto/datadog/trace/idx/span.proto
```diff
@@ -0,0 +1,153 @@
...
+   uint32 tracestate = 4;                      // Optional. W3C tracestate.
+   // @gotags: json:"flags" msg:"flags"
+   uint32 flags = 5;                           // Optional. W3C trace flags. If set, the high bit (bit 31) must be set.
+}
+
+message SpanEvent {
+  // @gotags: json:"time" msg:"time"
+  fixed64 time = 1; // time is the number of nanoseconds between the Unix epoch and this event.
+  // @gotags: json:"name" msg:"name"
+  uint32 name = 2; // name is this event's name.
```
ichinaski, line 64: Similar to the other comment, I think all fields that are string references could be more descriptive to avoid confusion. It should be very clear when a `uint32` is a string key, and when an actual integer. Something like `nameKey`, `nameRef`, or `nameIdx` maybe?

(This suffix suggestion is not for this field only, just a general observation applicable to all related fields)

pkg/proto/datadog/trace/idx/span.proto
ajgajg1134: Ahh good catch here, I meant to move this to map but missed it!

pkg/proto/datadog/trace/idx/span.proto
ajgajg1134: Yeah definitely, this also frees up those names for helper methods like `func Name() string` for ease of use

pkg/proto/pbgo/trace/idx/internal_span.go
ajgajg1134: Yeah I think that's a good idea - I think most usages outside this package will not call this method directly, but use higher level functions that handle interacting with the string table for you

ichinaski (approved):
LGTM! 💯

CelianR (approved):

PR #35844, opened 2025-04-04
[SVLS-6247][serverless] Support metric origins in Serverless agents

DylanLovesCoffee:
### What does this PR do?
Adds metric origins for Serverless agents (Lambda Extension and serverless-init/sidecar). These Serverless agent variations are not [built with zstd or zlib](https://github.com/DataDog/datadog-lambda-extension/blob/94cebb9037c28b17c086c049adb04d77db250a5b/scripts/build_binary_and_layer_dockerized.sh#L29) and purposefully [send timeseries data](https://github.com/DataDog/datadog-agent/blob/e0c94ee83df20740e1289b7bb757fe5a247a1e9a/cmd/serverless/main.go#L271) to `/api/v1/series`, so they have not benefited from the work previously done to support metric origins for the agent.

### Motivation
SVLS-6247 - we must support metric origins going forward. Primarily for GCP and Azure and the legacy AWS Lambda Extension

**`SendSketch()`:**
Since we do not build the Serverless agents with zlib or zstd, the StreamCompressor component is not created and we fall into this `else` condition in the serializer where no metric origin metadata is added: https://github.com/DataDog/datadog-agent/blob/e0c94ee83df20740e1289b7bb757fe5a247a1e9a/pkg/serializer/serializer.go#L469

**`SendIterableSeries`:**
Something similar happens for series data, where we do not use v2 of the series endpoint so we fall into this specific block that also does not have logic that adds metric origin metadata to the v1 request: https://github.com/DataDog/datadog-agent/blob/e0c94ee83df20740e1289b7bb757fe5a247a1e9a/pkg/serializer/serializer.go#L347

Local build of the Go Lambda Extension, where my Lambda function sends both a distribution metric and a timeseries metric and then validated the correct metric origin metadata exists in-app.
Screenshot 2025-04-04 at 3 19 51 PM  Screenshot 2025-04-04 at 3 19 57 PM


comp/dogstatsd/server/server.go
```diff
@@ -773,6 +773,14 @@ func (s *server) parseMetricMessage(metricSamples []metrics.MetricSample, parser
...
 			metricSamples[idx].Tags = metricSamples[0].Tags
 		}
+
+		// If we're in serverless mode, we need to determine the metric source from extra tags
+		if s.enrichConfig.serverlessMode {
+			for _, tag := range s.extraTags {
+				metricSamples[idx].Source = getMetricSourceFromTag(metricSamples[idx].Source, tag)
+			}
+		}
+
```
DylanLovesCoffee, line 783: Context: https://github.com/DataDog/datadog-agent/blob/e0c94ee83df20740e1289b7bb757fe5a247a1e9a/cmd/serverless-init/main.go#L202

pkg/serializer/internal/metrics/origin_mapping.go
```diff
@@ -10,6 +10,7 @@ import (
 )
 
 func metricSourceToOriginProduct(ms metrics.MetricSource) int32 {
+	const serieMetadataOriginOriginProductServerlessType = 1
```
DylanLovesCoffee, line 13: Ref: https://github.com/DataDog/dd-source/blob/821d1cf2f104b01d55ff4714f83f9098c9b20a76/domains/metrics/shared/libs/proto/origin/origin.proto#L166

pkg/serializer/internal/metrics/origin_mapping.go
```diff
@@ -332,6 +340,14 @@ func metricSourceToOriginCategory(ms metrics.MetricSource) int32 {
...
 	case metrics.MetricSourceGPU:
 		return 72 // ref: https://github.com/DataDog/dd-source/blob/276882b71d84785ec89c31973046ab66d5a01807/domains/metrics/shared/libs/proto/origin/origin.proto#L427
+	case metrics.MetricSourceAzureAppService:
+		return 35
+	case metrics.MetricSourceGoogleCloudRun:
+		return 36
+	case metrics.MetricSourceAzureContainerApp:
+		return 37
+	case metrics.MetricSourceAwsLambda:
+		return 38
```
DylanLovesCoffee, line 350: Ref: https://github.com/DataDog/dd-source/blob/821d1cf2f104b01d55ff4714f83f9098c9b20a76/domains/metrics/shared/libs/proto/origin/origin.proto#L376-L382

pkg/serializer/internal/metrics/iterable_series.go
```diff
@@ -500,17 +500,35 @@ func (pb *PayloadsBuilder) finishPayload() error {
...
+		serieWithMetadata := SeriesAlias{
+			Serie: serie,
+		}
+
+		if serie.Source != 0 {
+			serieWithMetadata.Metadata = map[string]interface{}{
+				"origin": map[string]interface{}{
+					"origin_product":        metricSourceToOriginProduct(serie.Source),
+					"origin_sub_product":    metricSourceToOriginCategory(serie.Source),
+					"origin_product_detail": metricSourceToOriginService(serie.Source),
```
hghotra, line 523: Can we also figure out a mapping for product detail? We need to set product detail to something which helps us identify whether it's an enhanced/runtime/standard (non-billable) or custom (billable) metric. This mapping will live in https://github.com/DataDog/dd-analytics/blob/master/luigiscripts/billing/usage/standard_metric_allowlist.json and help inform billing decisions in the future along with metric prefix.

pkg/serializer/internal/metrics/iterable_series.go
```diff
@@ -500,17 +500,35 @@ func (pb *PayloadsBuilder) finishPayload() error {
 // FIXME(maxime): to be removed when v2 endpoints are available
 func (series *IterableSeries) MarshalJSON() ([]byte, error) {
 	// use an alias to avoid infinite recursion while serializing a Series
-	type SeriesAlias Series
+	type SeriesAlias struct {
```
vickenty, line 503: Now that this is no longer a type alias, would it make sense to rename the type? The comment above about recursion is probably not applicable too.

```suggestion
	type SeriesWithMetadata struct {
```

pkg/serializer/internal/metrics/iterable_series.go
```diff
@@ -500,17 +500,35 @@ func (pb *PayloadsBuilder) finishPayload() error {
...
+		*metrics.Serie
+		Metadata map[string]interface{} `json:"metadata,omitempty"`
```
vickenty, line 505: Would it make sense to define a real struct for metadata instead of using `map[string]any`? All values are numeric, it should be possible to do this without the extra allocations.

comp/dogstatsd/server/enrich.go
```diff
@@ -62,9 +67,8 @@ func extractTagsMetadata(tags []string, originFromUDS string, processID uint32,
 		} else if strings.HasPrefix(tag, CardinalityTagPrefix) && origin.Cardinality == "" {
 			origin.Cardinality = tag[len(CardinalityTagPrefix):]
 			continue
-		} else if strings.HasPrefix(tag, jmxCheckNamePrefix) {
-			checkName := tag[len(jmxCheckNamePrefix):]
-			metricSource = metrics.JMXCheckNameToMetricSource(checkName)
+		} else if newSource := getMetricSourceFromTag(metricSource, tag); newSource != metricSource {
```
vickenty, line 70: `newSource != metricSource` means that non-duplicate tags with the same prefix would not be processed and would remain attached to the metric. Is there any reason to do so?

This is also different to how jmxfetch tag was handled before.

DylanLovesCoffee: Oof I missed the fact we are extracting the found tag. The new vs old source was an attempt to fix some overwriting, but this is not necessary. I will redo this to be more serverless specific and leave the JMX logic untouched

pkg/serializer/internal/metrics/iterable_series.go
DylanLovesCoffee: Agreed, makes sense 👍

pkg/serializer/internal/metrics/iterable_series.go
DylanLovesCoffee: https://github.com/DataDog/dd-source/pull/208062 - I'll make a small follow-up PR to map the details properly

comp/dogstatsd/server/server.go
```diff
@@ -773,6 +773,17 @@ func (s *server) parseMetricMessage(metricSamples []metrics.MetricSample, parser
 		} else {
 			metricSamples[idx].Tags = metricSamples[0].Tags
 		}
+
+		// If we're in serverless mode, we need to determine the metric source from extra tags
+		if s.enrichConfig.serverlessMode {
+			isRuntime := strings.HasPrefix(metricSamples[idx].Name, "runtime.")
```
hghotra, line 779: There's a case where prefix is `.jvm` for `java` right?

DylanLovesCoffee: There's an existing case for `jvm.*` metrics that I don't think is necessary to cover again: https://github.com/DataDog/datadog-agent/pull/35844/files#diff-a16207429aa26872091b358ada19dbad52797ca17561c5a5eee4fbacd5518315R71-R73

steveny91 (approved):

hghotra (approved):
✅

vickenty (approved):

comp/dogstatsd/server/enrich_serverless_test.go
```diff
@@ -35,3 +36,90 @@ func TestConvertParseDistributionServerless(t *testing.T) {
 	// an empty string and the parser / enricher should keep the host that way.
 	assert.Equal(t, "", parsed.Host, "In serverless mode, the hostname should be an empty string")
 }
+
+func TestGetServerlessMetricSource(t *testing.T) {
```
tobz, line 40: Are these tags actually relevant anymore now that the logic isn't tags-based?

tobz (commented):
One small question.

comp/dogstatsd/server/enrich_serverless_test.go
DylanLovesCoffee: Aha you're right, the tags aren't necessary 🧼

tobz (approved):

PR #35847, opened 2025-04-04
[prometheus] Add support for including/excluding k8s annotations using regexp

rossdylan:
### What does this PR do?
I've added the ability to configure Prometheus auto-discovery to include or exclude pods based on a regexp over Kubernetes annotations. This is done via a new set of options that mirror the existing include/exclude fields.

### Motivation
We've run into a situation where we need to exclude a set of deployments from auto-discovery but they don't have common annotations, and we don't control them (GKE addons). In order to prevent Datadog from collecting these metrics we don't care about we want a way to specify inclusion or exclusion rules via regexp.

I've added unit tests the specifically exercise this behavior, and ran the existing ones to ensure I didn't regress the current behavior.


michaelcretzman (approved):

rossdylan:
I'm waiting on legal approval for the CLA, should hopefully get that signed this week

rossdylan:
alright cla, signed should be good to run CI now

rossdylan:
any chance this could get looked at?

PR #35849, opened 2025-04-04
add source tag for log byte sent telemetry

DDuongNguyen:
### What does this PR do?
Adds proper telemetry source tagging for destinations based on component type. 

Destinations now dynamically set the source tag to either "logs" or "epforwarder" based on the telemetry origin, improving metric attribution and observability.

### Motivation
Previously, all destinations used the same source tag for telemetry metrics, making it difficult to distinguish between logs and non-logs traffic in our monitoring systems. This change allows us to properly attribute metrics to their source component, enabling better monitoring and alerting for different data flows.

I've added new tests in pkg/logs/client/http/destination_test.go that verify both source tag scenarios:
When the telemetry name contains "logs", the source tag is set to "logs"
When the telemetry name doesn't contain "logs", the source tag is set to "epforwarder"

QA wise, you can check by sending some logs and some dmb metric events and check for the telemetry taggings.

The telemetry logs_byte_sent an cardinality is increased by 2, which are `source:logs` and `source:epforwarder`


pkg/logs/client/http/destination.go
```diff
@@ -301,7 +302,15 @@ func (d *Destination) unconditionalSend(payload *message.Payload) (err error) {
 		return err
 	}
 	metrics.BytesSent.Add(int64(payload.UnencodedSize))
-	metrics.TlmBytesSent.Add(float64(payload.UnencodedSize))
+	var sourceTag string
+
+	if strings.Contains(d.Metadata().TelemetryName(), "logs") {
+		sourceTag = "logs"
+	} else {
```
soberpeach, line 309: Are "logs" and "epforwarder" the only two possibilities for TelemetryName?

DDuongNguyen: No, each telemetry has its own name but theres 10+ telemetries so to decrease tag cardinality until we need specific epforwarders telemetry names, i just categorize it into 2.

DDuongNguyen: From my understanding its only either epforwarder events or logs that go through here. A full list of the events can be found here:
https://github.com/DataDog/datadog-agent/blob/yoon/log-telemetry/comp/forwarder/eventplatform/eventplatformimpl/epforwarder.go#L55-L222

soberpeach: Ohhh okay. Makes sense to me then!

soberpeach (approved):

PR #35895, opened 2025-04-07
Go DI e2e test improvements

grantseltzer:
### What does this PR do?

- Each e2e test fixture is run as a subtest, giving us the ability to run each test individually which is useful for development.
- Fixes the format/naming of many fixtures/test-cases
- Adds more fixtures/test-cases

### Motivation

Need to improve e2e testing to avoid regressions like in #35839

Ran the e2e tests.

### Additional Notes

There's a lot more improvements to be made. A lot more cases, the ability to run multiple samples at once. There's also still some flakiness that needs to be fixed so that these tests are run in CI reliably.


tylfin (approved):
🔥

PR #35902, opened 2025-04-07
[CIP] feat: Use s3 clone strategy for datadog-agent CI jobs

BaptisteLalanne:
### What does this PR do?

This PR is adding a new job at the beginning of the CI pipeline called `clone` that is responsible to upload the shallow clone checked-out on the pipeline's commit into S3. The clone will have the depth that the `clone` job has.

Every job with `GIT_STRATEGY: s3` will look into s3 for the clone first and if present, will download it and extract it. In case it's missing, it will fallback to Gitaly and therefore clone from GitLab as usual. 

The clone job is running in parallel of the current pipeline config thus should only have a positive impact on the pipeline duration as not supposed to be part of the critical path. For some repositories we set that job as needed by all the next jobs to have something unified and have an _O(Npipeline)_ fetch to Gitaly instead of _O(Njobs)_ to reduce the impact on Gitaly. This does not seem necessary for now for datadog-agent as we currently have no traffic issues.

### Motivation

The critical path of a datadog-agent pipeline on a main pipeline varies depending on the pipeline between 1 to 70 jobs but seems to be roughly around 10
[CI Visibility provide this insights](https://app.datadoghq.com/ci/pipeline-executions?query=ci_level%3Ajob%20env%3Aprod%20%40git.repository.id%3Agitlab.ddbuild.io%2FDataDog%2Fdatadog-agent%20%40ci.pipeline.name%3ADataDog%2Fdatadog-agent%20%40ci.provider.instance%3Agitlab-ci%20%40git.branch%3Amain%20%40ci.on_critical_path%3Atrue%20%40ci.status%3Asuccess&agg_m=%40duration&agg_m_source=base&agg_q=%40ci.pipeline.id%2C%40ci.on_critical_path&agg_q_source=base%2Cbase&agg_t=avg&colorBy=meta%5B%27ci.stage.name%27%5D&colorByAttr=meta%5B%27ci.stage.name%27%5D&currentTab=trace&filterCriticalPath=true&fromUser=false&graphType=flamegraph&index=cipipeline&sort_m=%2C&sort_m_source=%2C&sort_t=%2C&spanViewType=metadata&top_n=250%2C30&top_o=top%2Ctop&viz=toplist&x_missing=true%2Ctrue&start=1742982968083&end=1744188968083&paused=false)

When `clone` job is finished there might be most likely still ~10 jobs to do in the critical path

image

For the jobs that will be scheduled after the `clone` job, they will be able to pick up the shallow clone from s3 which has a more stable clone time (p95 ~20s instead of ~1mn). We thus could expect a performance win, more or less depending on the amount of jobs in the critical that starts after that clone job.

`env:prod @git.repository.id:gitlab.ddbuild.io/DataDog/datadog-agent @ci.pipeline.name:DataDog/datadog-agent @ci.provider.instance:gitlab-ci @user.name:devflow-worker @gitlab.pipeline_source:api @ci.param.run_all_builds:true @ci.status:success -@gitlab.result:"passed with warnings"`

There are only 1 pipeline that run all the tests and succeeded in the last 3 months with roughly the same jobs and since the biggest time spent is inside the job's script, there is not much interest of comparing. 
- `1h54` - https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/59612167
- `1h50` - https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/61569252  (this pr)

Not only is it supposed to reduce the critical path, it would also reduce the cumulated CI time on the pipeline for every jobs using it which should translate to cost win (minus the s3 price for 1 day of shallow clones).

More detail on the feature and how it works: https://datadoghq.atlassian.net/wiki/spaces/DEVX/pages/3622535770/GitLab+Runners+Fork+Features

Examples: 
- https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/885658873
- https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/888208781

I triggered the `Test all` pipeline but it might be missing jobs that only run on main. I would defer to agent devx folks to know if we can consider it enough or not.

### Additional Notes

That strategy is currently used in the biggest internal repos and provided improvement of speed and reliability. 

Right now the s3 strategy is only available to the fork version of our gitlab runners. There are some runners like the `windows` one that do not work with it yet and just use the classic strategy which for them seems to be even longer (once windows runner are migrated to the fork, it might also improve even further)

Extra note: Remainder for CI Platforms team: If DevX Agent folks want to continue with this, we need to set the lifecycle to 1d for the bucket not to accumulate shallow clones. As mentioned above if the cache is missing we fallback to classic Gitaly - no impact on retried job after 1d.


BaptisteLalanne:

CelianR (approved):

PR #35923, opened 2025-04-08
migrate authtoken to comp/core

misteriaud:
### What does this PR do?
This PR migrates the `comp/api/authtoken` component to `comp/core/ipc`. In the process, it introduces several improvements:
- Transitions the component to a single implementation flavor.
- Migrates it to the new component layout.
- Adds two new helpers to simplify and secure IPC communications:
  - **`GetClient`**: Provides an HTTPS client preconfigured with secure default settings for IPC communication.
  - **`HTTPMiddleware`**: Supplies an HTTP middleware for securing internal IPC servers.

### Motivation
Enforcing the initialization of authentication artifacts at command startup eliminates the use of insecure TLS configurations, which previously allowed communication with the Agent API server in degraded contexts. This approach enables the gradual removal of calls to the `SetAuthToken` function (`pkg/api/util/util.go`), reducing reliance on uninitialized global variables.

Additionally, by consolidating and improving the IPC communication framework, this change enhances the security, maintainability, and usability of the Agent's inter-process communication for developers.

To ensure the correctness of these changes:
- Unit tests have been added to cover the new features provided by `GetClient`.

### Additional Notes
- Any existing references to the old `GetClient()` function will need to be refactored to use the new `ipc.HTTPClient`. Once this is completed, we will be able to remove globals and helpers in `pkg/api/util`.
- **Review Advice**: To make the review process easier, it is recommended to review this PR commit by commit:
  - The first 4 commits move and migrate the `comp/api/authtoken` component to the `comp/core/ipc` location and adopt the new component layout.
  - The 5th commit introduces the new client (`GetClient`) and middleware (`HTTPMiddleware`) features.
  - The 6th and subsequent commits update the rest of the codebase to integrate these changes.


vickenty (approved):

comp/core/ipc/http/client.go
```diff
@@ -0,0 +1,348 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
```
louis-cqrl, line 1: all the http method functions are duplicating a lot of simple code, would it be possible to make it shorter ?

wdhif (approved):
LGTM for @DataDog/container-platform files.

louis-cqrl (approved):
Most of the logic look good to me, I did not caught any typing error either, the code is also clearer than before and more unified.

Nice PR 🎉

paulcacheux (approved):

gjulianm (approved):
LGTM for ebpf-platform file (1)

ichinaski (approved):

mellon85 (approved):

BaptisteFoy (approved):
lgtm for the only fleet file 👍

julien-lebot (approved):
Looks good for Windows Agent changes

rahulkaukuntla (approved):

liustanley (approved):

daniel-taf (approved):

comp/core/ipc/http/client.go
```diff
@@ -0,0 +1,348 @@
...
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2023-present Datadog, Inc.
+
+// Package http implements helpers for HTTP communication between Agent processes
+package http
```
hush-hush, line 7: Could we rename this folder `utils` or something similar ? `comp/core/ipc/http` make is sound like an HTTP implementation of the IPC comp.

comp/core/ipc/impl-none/ipc.go
```diff
@@ -0,0 +1,55 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2023-present Datadog, Inc.
+
+// Package noneimpl implements a noop version of the auth_token component
+package noneimpl
```
hush-hush, line 7: What is the use case for the noneimpl ?

comp/core/ipc/mock/mock.go
```diff
@@ -0,0 +1,109 @@
...
+// inMemoryIPCComponent is a mock for the IPC component
+// It is used to set the auth token, client TLS config and server TLS config in memory
+type inMemoryIPCComponent struct {
+	t      testing.TB
+	conf   config.Component
+	client ipc.HTTPClient
+}
+
+// Mock returns a mock for ipc component.
+func Mock(t testing.TB) Component {
```
hush-hush, line 48: Let's rename this to `New`. According to Go guideline we should avoid repeating the package name in the method. `mock.Mock(t)` is less explicit than `mock.New(t)`.

comp/core/ipc/mock/mock.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/DataDog/datadog-agent/pkg/util/option"
+)
+
+// Component is the mocked component type.
+type Component interface {
+	ipc.Component
+	// NewMockServer allows to create a mock server that use the IPC certificate
+	NewMockServer(handler http.Handler) *httptest.Server
+	Optional() option.Option[ipc.Component]
+}
```
hush-hush, line 37: Do we need an interface here ? Are we expecting multiple implementation for this mock ? If not, we could remove the interface and make `inMemoryIPCComponent` public (maybe with a rename to `IPCMock` ?).

hush-hush (changes requested):

comp/core/ipc/impl/ipc.go
```diff
@@ -0,0 +1,92 @@
...
+// Requires defines the dependencies for the ipc component
+type Requires struct {
+	Conf   config.Component
+	Log    log.Component
+	Params ipc.Params
+}
+
+// Provides defines the output of the ipc component
+type Provides struct {
+	Comp option.Option[ipc.Component]
```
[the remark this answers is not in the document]
misteriaud: Here's the reasoning behind the current approach:
The goal is to accommodate commands that optionally require the `ipc.Component` without failing if it fails to initialize. Using `ProvideOptional` would cause an `option.Option[ipc.Component]` dependency to fail at the fx initialization stage because fx would attempt to construct the `ipc.Component` and stop on failure, instead of returning `option.None`.

To address this, I opted for a design where the `ipc.Component` is wrapped into an `option.Option[ipc.Component]` during construction. This allows fx helpers to unwrap it automatically when needed. Here's how it works in practice:

1. Optional dependency example: `diagnose` Command

Without any flags, diagnose requires IPC requests to the Agent CMD API. In this case, it depends on `option.Option[ipc.Component]`, unwraps it as needed, and returns an error if it's `None`.
When used with the `--local` flag, `diagnose` operates autonomously and does not rely on the initialization state of `ipc.Component`. This ensures the command won't fail unnecessarily if IPC artifacts does not exists.

2. Hard dependency example: `status` Command

For commands like `status`, which always require the `ipc.Component`, the `unwrapIPCComp` helper converts the optional dependency into the raw dependency during the fx initialization step.

I didn't saw any other alternative to tackled the ease of use I wanted for this component. Nevertheless I'm open to suggestion !

comp/core/ipc/impl/ipc.go
```diff
@@ -0,0 +1,92 @@
...
+	return util.GetTLSServerConfig()
+}
+
+func (ipc *ipcComp) HTTPMiddleware(next http.Handler) http.Handler {
+	return ipchttp.NewHTTPMiddleware(func(format string, params ...interface{}) {
+		ipc.logger.Errorf(format, params...)
+	}, ipc.GetAuthToken())(next)
+}
+
+func (ipc *ipcComp) GetClient() ipc.HTTPClient {
```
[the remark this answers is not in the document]
misteriaud: I added a comment on the `ipc.Component` interface. This function is intended to be used in API servers as authorization middleware. For now it's basically comparing the `authorization` HTTP header against the local loaded `auth_token`. It might change in the future.
Having this feature centralized in the `ipc.Component` makes it easier to assess and update IPC security.

comp/core/ipc/fx/fx.go
```diff
@@ -0,0 +1,61 @@
...
+func Module() fxutil.Module {
+	return fxutil.Component(
+		fxutil.ProvideComponentConstructor(
+			ipcimpl.NewComponent,
+		),
+		fx.Provide(unwrapIPCComp),
+		fx.Provide(newIPCClient),
+		fx.Provide(newOptionalIPCClient),
+	)
+}
```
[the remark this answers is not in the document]
misteriaud: Good idea ! Thanks ! I'll make the change this afternoon

comp/core/ipc/impl/ipc.go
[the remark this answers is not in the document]
hush-hush: I think its a convoluted solution to this problem. The real issue here is that the diagnose command should not require a dependency it doesn't need.

In general, having the edge case of a command bleed into the design of a component is not a good idea. We should aim   at solving the problem within the diagnose command. This would be possible by a conditional include of the ipc comp [here](https://github.com/DataDog/datadog-agent/blob/main/cmd/agent/subcommands/diagnose/command.go#L98-L121).

Something like
```
extaDeps := []fxutil.Module{}
if !cliParams.runLocal {
    extrDeps = append(extraDeps, ipc.Module())
}

return fxutil.OneShot(cmdDiagnose,
				// [...]
				extraDeps....,
			)
```

WDYT ?

comp/core/ipc/impl/ipc.go
```diff
@@ -0,0 +1,92 @@
...
+func NewComponent(reqs Requires) Provides {
+	var initFunc func(model.Reader) error
+
+	if reqs.Params.AllowWriteArtifacts {
+		reqs.Log.Infof("Load or create IPC artifacts")
+		initFunc = util.CreateAndSetAuthToken
+	} else {
+		reqs.Log.Infof("Load IPC artifacts")
+		initFunc = util.SetAuthToken
+	}
```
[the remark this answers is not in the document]
hush-hush: The issue with the bundle is that people copy/paste code everywhere. We often have the same issue with `secrets`, where many CLI command include the secret comp even when not needed (resulting in unnecessary resolution of secret).

We could have 2 implementations and include only the "default" one in the core Bundle. This is what we do for metadata for example. By default the bundle include everything, and if something require another behavior, then they can cherry-pick what they want.

Once of the issue with the current state of component is that they are complex to understand for new devs as they expose a lot of complexity. We should aime at component with very explicit behavior.

comp/core/ipc/impl-none/ipc.go
hush-hush: I see, ideally the tracer would not depends on `remote-config` in serverless mode. But that's for another PR !

comp/core/ipc/mock/mock.go
```diff
@@ -0,0 +1,109 @@
...
+		m.conf.UnsetForSource("cmd_port", pkgconfigmodel.SourceAgentRuntime)
+	})
+
+	return ts
+}
+
+// New returns a new authtoken mock
+func (m *inMemoryIPCComponent) Optional() option.Option[ipc.Component] {
+	return option.New[ipc.Component](m)
+}
```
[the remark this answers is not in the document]
misteriaud: This was a helper to get the optional version of the `ipc.Component` if needed.
I resolve the comment since the `option.Option` type is no longer used for the `ipc.Component`.

comp/core/ipc/http/client.go
misteriaud: Sure ! I renamed it to `httphelpers` (we plan to also have `grpchelpers` in a following PR).
Done in 22f36c128dbe0f9ea509164d463b3adfcb7f31af.

comp/core/ipc/mock/mock.go
misteriaud: Good point ! Done in f578be1a5e562f76cdce110dcb3f5b0a8812f7af.

comp/core/ipc/mock/mock.go
misteriaud: No need of the interface indeed, I also renamed the struct to `IPCMock` as you suggested. Done in f578be1a5e562f76cdce110dcb3f5b0a8812f7af.

comp/core/ipc/impl/ipc.go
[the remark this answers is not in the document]
misteriaud: Thanks to your feedback I followed a new approach to be closer to the way the other components work:
- as you suggested I removed the usage of `option.Option` type and return an error if the component cannot be build successfully (instead of returning `option.None`)
- to cover the edge-cases such as the `flare` command, where we need the command to run even if the auth artifacts cannot be found, I introduced a specific `debug` constructor (and `fx.Module`) that might return empty authtoken or default `tls.Config` objects.

comp/core/ipc/impl/ipc.go
```diff
@@ -0,0 +1,92 @@
...
+
+	if err := initFunc(reqs.Conf); err != nil {
+		reqs.Log.Errorf("could not load IPC artifacts: %s", err)
+		return Provides{
+			Comp: option.None[ipc.Component](),
+		}
```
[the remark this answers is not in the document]
misteriaud: With the new implementation mentioned above, the constructors return an error in these cases, except for the `debug` constructor.

comp/core/ipc/impl/ipc.go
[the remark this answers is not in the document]
misteriaud: Since the only difference between `ReadOnly` and the `ReadWrite` versions of the components are the way they are initialized, I created multiple constructors to cover the different use cases. Each constructor have its own `fx.Module`. We end up with this type of usage:
```
ipcfx.ModuleForOneshot()
```
or 
```
ipcfx.ModuleForDaemon()
```

I think it's more explicit than previous version and avoid duplication of code (compared to a multi implementation approach). WDYT ?

comp/core/ipc/fx/fx.go
```diff
@@ -0,0 +1,46 @@
...
+
+// ModuleForDebug defines the fx options for this component for commands that should work
+// even if the agent is not running or IPC artifacts are not initialized.
+// WARNING: This module should not be used outside of the flare and diagnose commands.
+// This module covers cases where it is acceptable to not have initialized IPC component.
+// This is typically the case for commands that MUST work no matter the coreAgent is running or not, if the auth artifacts are not found/initialized.
+// A good example is the `flare` command, which should return a flare even if the IPC component is not initialized.
+func ModuleForDebug() fxutil.Module {
```
hush-hush, line 40: nit-pick: `Debug` doesn't reflect well what the documentation above says. With `Debug` in the name I'm expecting something that will log more, or send telemetry for example.

comp/core/ipc/fx/fx.go
```diff
@@ -0,0 +1,46 @@
...
+package fx
+
+import (
+	ipcimpl "github.com/DataDog/datadog-agent/comp/core/ipc/impl"
+	"github.com/DataDog/datadog-agent/pkg/util/fxutil"
+)
+
+// ModuleForDaemon defines the fx options for this component for the daemon commands.
+// It is using the NewReadWriteComponent constructor under the hood.
+func ModuleForDaemon() fxutil.Module {
```
hush-hush, line 16: Why not use `ModuleReadWrite` instead ? It's often better to name things after what they are rather than what they're meant for.

comp/core/ipc/def/component.go
```diff
@@ -0,0 +1,50 @@
...
+
+// HTTPClient is a HTTP client that abstracts communications between Agent processes
+type HTTPClient interface {
+	Do(req *http.Request, opts ...RequestOption) (resp []byte, err error)
+	Get(url string, opts ...RequestOption) (resp []byte, err error)
+	Head(url string, opts ...RequestOption) (resp []byte, err error)
+	Post(url string, contentType string, body io.Reader, opts ...RequestOption) (resp []byte, err error)
+	PostChunk(url string, contentType string, body io.Reader, onChunk func([]byte), opts ...RequestOption) (err error)
+	PostForm(url string, data url.Values, opts ...RequestOption) (resp []byte, err error)
+	NewIPCEndpoint(endpointPath string) (Endpoint, error)
```
hush-hush, line 44: Could we document those method ?

comp/core/ipc/impl/ipc.go
```diff
@@ -0,0 +1,126 @@
...
+type ipcComp struct {
+	logger log.Component
+	conf   config.Component
+	client ipc.HTTPClient
+}
+
+// NewReadOnlyComponent creates a new ipc component by trying to read the auth artifacts on filesystem.
+// If the auth artifacts are not found, it will return an error.
+func NewReadOnlyComponent(reqs Requires) (Provides, error) {
+	reqs.Log.Infof("Load IPC artifacts")
```
hush-hush, line 42: Debug would be better, I don't think we should show this by default.

hush-hush (changes requested):

comp/core/ipc/fx/fx.go
misteriaud: I finally renamed it `ModuleInsecure()` which makes it evident that it should be used over `ModuleReadWrite()` or `ModuleReadOnly()`.

comp/core/ipc/fx/fx.go
misteriaud: Good point, updated in 6b689d2999a6803c4013bd0b7e98ed9036953e1c.

comp/core/ipc/def/component.go
misteriaud: Sure ! Done in 6b689d2999a6803c4013bd0b7e98ed9036953e1c.

comp/core/ipc/impl/ipc.go
misteriaud: I turned logs level down to `Debug` in 6b689d2999a6803c4013bd0b7e98ed9036953e1c.

comp/core/ipc/fx/fx.go
```diff
@@ -0,0 +1,46 @@
...
+	return fxutil.Component(
+		fxutil.ProvideComponentConstructor(
+			ipcimpl.NewReadWriteComponent,
+		),
+	)
+}
+
+// ModuleForOneshot defines the fx options for this component for the one-shot commands.
+// It is using the NewReadOnlyComponent constructor under the hood.
+func ModuleForOneshot() fxutil.Module {
```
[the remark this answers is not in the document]
misteriaud: Done in 6b689d2999a6803c4013bd0b7e98ed9036953e1c.

hush-hush (approved):

duncanista (approved):

iliakur (approved):

PR #35930, opened 2025-04-08
[Backport 7.64.x] autodiscovery: only detect NVML library if a feature flag is enabled

gjulianm:
### What does this PR do?
This PR changes the NVML feature detection to have a feature flag gate, so that for now all GPU monitoring features are gated and only enabled explicitly by the customer.

### Motivation

Avoid enabling by default preview features. Backported to reduce risk for customers. #incident-37075

Manually validated that the agent does not try to load the NVML library by default.

Manually corrupted the `libnvidia-ml.so.1` library, checked that it crashes the agent with the feature flag enabled and doesn't crash it with it disabled.


justin-lesko (approved):

jhgilbert (approved):
Approved with a style suggestion, thanks!

PR #35939, opened 2025-04-08
fix(autoinstrumentation): DD_SERVICE in some init containers was inconsistent

stanistan:
### What does this PR do?

Addresses: https://datadoghq.atlassian.net/browse/INPLAT-550

### Motivation

This isn't a complete fix since we would also want to remove the code that sets the environment variable based on the "owner name" at some point (the source of this confusion).

This makes it so that the likelyhood of inconsistent env-vars when this is called is minimized and it's more obvious what the behavior does (by not tying it to lib-config).

Changes:

1. Moves the implicit service naming out of `libConfig` so this behavior is less hidden and more obvious that it's happening.

2. Adds behavior to find an already set `corev1.EnvVar` for `DD_SERVICE` and use that if it is found before setting one based on the owner name.

Behavior changes:

- For users where `DD_SERVICE` was set by `tagsFromLabels` this would make the init containers injected consistent with the application container. `tagsFromLabels` can use the labels on the pod OR the labels on the owner, and if thats set they would propagate as strings.

- For users where it was not set, this would find the owner name and then propagate it.

Added unit tests for this behavior, will do some more manual QA and add steps to reproduce.

??

### Additional Notes

There is a related issue https://datadoghq.atlassian.net/browse/INPLAT-458 where service name can be overridden for some customers that are providing it the `DD_SERVICE` environment variable in their containers and we _always_ set it to the deployment name. We are investigating ways of deprecating this behavior (and the function `getServiceNameFromPodOwnerName` entirely).

### DIFF
--- a/pkg/clusteragent/admission/mutate/autoinstrumentation/namespace_mutator.go
+++ b/pkg/clusteragent/admission/mutate/autoinstrumentation/namespace_mutator.go
@@ -208,6 +208,32 @@ func (m *mutatorCore) injectTracers(pod *corev1.Pod, config extractedPodLibInfo)
 
 	if m.filter.IsNamespaceEligible(pod.Namespace) {
 		_ = basicLibConfigInjector{}.mutatePod(pod)
+
+		// N.B.
+		//
+		// This is kind of gross, and would ideally not happen more than in
+		// one place but we made a decision to infer DD_SERVICE in the auto-instrumentation
+		// webhook a while ago and customers might be relying on this behavior.
+		//
+		// We have another webhook that does something really similar: tagsFromLabels and
+		// it this is where the responsibility should generally.
+		//
+		// The big difference between the two is that tagsFromLabels looks at the label
+		// metadata and we might override it and this one will look for the _name_ of the
+		// owner resource.
+		//
+		// The intention is to have this always run last so that we fallback to the owner
+		// name in cases of missing labels coming from the resource or its owner.
+		//
+		// mutatecommon.InjectEnv will _only_ change an env var if it is present so this code
+		// will add `DD_SERVICE` to all containers on the pod if it wasn't already set.
+		//
+		// We want to get rid of the behavior when we are triggering the fallback _and_
+		// it applies: https://datadoghq.atlassian.net/browse/INPLAT-458
+		serviceNameEnv, found := findServiceNameInPod(pod)
+		if found {
+			_ = mutatecommon.InjectEnv(pod, serviceNameEnv)
+		}
 	}
 
 	return lastError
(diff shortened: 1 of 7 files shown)


pkg/clusteragent/admission/mutate/autoinstrumentation/service_name_mutator.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/DataDog/datadog-agent/pkg/util/log"
+)
+
+func encodeEnvVar(e corev1.EnvVar) string {
+	out, _ := json.Marshal(e)
+	return string(out)
+}
+
+type serviceEnvVarData struct {
+	env        corev1.EnvVar
```
stanistan, line 26: Note to self (and thing to change):

This is still a draft, but I was originally thinking about being _smarter_ here which is where the container pointers come from, and potentially storing whether or not this was an init container or not. 

I'm no longer _also_ iterating through init containers so that's not necessary. 

This function could have potentially output a candidate list of containers missing the env var and we would have a `mutator` that then be executed but it is much simpler to do a different approach, output candidate env vars and pick the best one.

pkg/clusteragent/admission/mutate/autoinstrumentation/namespace_mutator.go
```diff
@@ -200,6 +200,32 @@ func (m *mutatorCore) injectTracers(pod *corev1.Pod, config extractedPodLibInfo)
...
 		_ = basicLibConfigInjector{}.mutatePod(pod)
+
+		// N.B.
+		//
+		// This is kind of gross, and would ideally not happen more than in
+		// one place but we made a decision to infer DD_SERVICE in the auto-instrumentation
+		// webhook a while ago and customers might be relying on this behavior.
+		//
+		// We have another webhook that does something really similar: tagsFromLabels and
+		// it this is where the responsibility should generally.
```
knusbaum, line 211:
```suggestion
		// that is where the responsibility for service tagging should generally be.
```

pkg/clusteragent/admission/mutate/autoinstrumentation/namespace_mutator.go
```diff
@@ -200,6 +200,32 @@ func (m *mutatorCore) injectTracers(pod *corev1.Pod, config extractedPodLibInfo)
...
+		// owner resource.
+		//
+		// The intention is to have this always run last so that we fallback to the owner
+		// name in cases of missing labels coming from the resource or its owner.
+		//
+		// mutatecommon.InjectEnv will _only_ change an env var if it is present so this code
+		// will add `DD_SERVICE` to all containers on the pod if it wasn't already set.
+		//
+		// We want to get rid of the behavior when we are triggering the fallback _and_
+		// it applies: https://datadoghq.atlassian.net/browse/INPLAT-458
```
knusbaum, line 224: So this should go away completely, and rely on the Tracers to make service-naming decisions when we can't, correct?

knusbaum (commented):
First round, looks like a good change. Would be good to get the telemetry in on this too, though, since we need that to eventually remove this.

pkg/clusteragent/admission/mutate/autoinstrumentation/namespace_mutator.go
stanistan: Ideally this is gone so the only thing that can set the service name is `tagsFromLabels`, workload selection, and the user themselves.

knusbaum (approved):
Looks great.

Thanks for the refactor and the additional testing.

git-thuerk-done (approved):

PR #35947, opened 2025-04-09
Use prebuilt setuptools

alopezz:
### What does this PR do?

Installs setuptools from the wheel rather than from source.

### Motivation

This avoids needing `wheel` as a requirement.


alopezz:
We're favoring https://github.com/DataDog/datadog-agent/pull/35952 for now.

PR #35269, opened 2025-03-19
[OTAGENT-331] Publish Agent docker image `7-full-amd64` to SMP registry

ogaca-dd:
### What does this PR do?
This PR publishes Agent docker image `7-full-amd64` to SMP registry

### Motivation

https://github.com/DataDog/datadog-agent/pull/35185

PR #35953, opened 2025-04-09
single_machine_performance-full-amd64-a7

ogaca-dd:
### What does this PR do?
This PR publishes Agent docker image `7-full-amd64` to SMP registry

This is the first step in setting up writing experiments for the otel-agent:
1.	Create a new job `single_machine_performance-full-amd64-a7` that pushes the full agent image to the SMP registry (this PR).
2.	Update regression_detector.yml to use the full image. See [this PR](https://github.com/DataDog/datadog-agent/pull/35185).
3.	Remove the old single_machine_performance-amd64-a7 job and keep only the -full- version, ensuring there’s a single job responsible for pushing to the SMP registry.
4.	Add OTEL experiments, which rely on the full image.

Note: There was a previous attempt to add `single_machine_performance-full-amd64-a7` in this [PR](https://github.com/DataDog/datadog-agent/pull/35269). This PR was reverted because the otel agent was enabled by default in the `full` image leading to performance difference in the quality gate jobs. This is not the case since [this PR](https://github.com/DataDog/datadog-agent/pull/35641).
### Motivation

https://github.com/DataDog/datadog-agent/pull/35185


L3n41c (approved):

.gitlab/container_build/docker_linux.yml
```diff
@@ -73,6 +73,19 @@ single_machine_performance-amd64-a7:
     IMG_SOURCES: ${SRC_AGENT}:v${CI_PIPELINE_ID}-${CI_COMMIT_SHORT_SHA}-7-amd64
     IMG_DESTINATIONS: 08450328-agent:${CI_COMMIT_SHA}-7-amd64
 
+single_machine_performance-full-amd64-a7:
+  extends: .docker_publish_job_definition
+  stage: container_build
```
FlorentClarret, line 78: Any reason to publish images in the `container_build` stage and not in `container_deploy`?

FlorentClarret: Same question for `single_machine_performance-amd64-a7` that's already there

ogaca-dd: I don't know the rational of this choice but I wanted to keep the exact same behaviour as I will remove `single_machine_performance-amd64-a7` in a follow-up PR and only keep `single_machine_performance-full-amd64-a7`.

FlorentClarret: If there's no strong reason to keep it in the build stage, I'd move it to the deploy stage to keep yes so we have all the deploy jobs in the same stage

goxberry: Moving the `single_machine_performance-full-amd64-a7` job to the `deploy_containers` stage would be fine for Single Machine Performance. 

I support moving that job to `.gitlab/deploy_containers/deploy_containers_a7.yml` and following the naming conventions of that YAML file so the job configuration in that file looks something like:

```yaml
deploy_containers-single_machine_performance-full-amd64-a7:
  extends: .docker_publish_job_definition
  stage: deploy_containers
  rules:
    - !reference [.except_mergequeue]
    - when: on_success
  variables:
    IMG_REGISTRIES: internal-aws-smp
    IMG_SOURCES: ${SRC_AGENT}:v${CI_PIPELINE_ID}-${CI_COMMIT_SHORT_SHA}-7-full-amd64
    IMG_DESTINATIONS: 08450328-agent:${CI_COMMIT_SHA}-7-full-amd64
```

The `needs` key should be unnecessary because GitLab CI stages run sequentially, and I am guessing the `deploy_containers` stage depends on the `container_build` stage.

I also support making similar changes to the `single_machine_performance-amd64-a7` job in a later PR.

goxberry: As long as the `deploy_containers` stage precedes the `functional_test` stage -- I imagine this condition is true -- I don't anticipate the configuration change I've proposed in the comment above would cause any issues. However, if there are difficulties, I'd be happy to help with review and remediation.

ogaca-dd: > I also support making similar changes to the single_machine_performance-amd64-a7 job in a later PR.

I'll remove `single_machine_performance-amd64-a7` in a later PR to only keep `deploy_containers-single_machine_performance-full-amd64-a7`, so there is no need to migrate `single_machine_performance-amd64-a7`

.gitlab/deploy_containers/deploy_containers_a7.yml
```diff
@@ -354,3 +354,13 @@ deploy_containers_latest-a7-fips:
       - JMX:
           - ""
           - "-jmx"
+
+deploy_containers-single_machine_performance-full-amd64-a7:
+  extends: .docker_publish_job_definition
+  stage: deploy_containers
+  rules:
+    !reference [.on_internal_final]
```
spencergilbert, line 362: I expect this should match the rules for the regression detector job which is expected to use this image.

```suggestion
    - !reference [.on_dev_branches]
    - when: on_success
```

ogaca-dd: This seems to create an issue for [compute_gitlab_ci_config](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/891684145)

spencergilbert (approved):

PR #35954, opened 2025-04-09
[process-agent] Switch to remote workloadmeta

davidor:
### What does this PR do?

This PR makes the process-agent always use the remote workloadmeta.

There used to be a config option to choose between remote and local, but now we want to use remote only. This is already how it works in the security-agent and system-probe.

The benefits are the same as for the other agents: the process-agent binary is smaller, and now only one agent will fetch data from sources like container runtime sockets.

@DataDog/container-intake how should we test this?

I did a basic test on a local cluster. I set up the process-agent to run the process checks instead of the core agent. Everything looks fine, but I’m not sure what else we should check to make sure this change is safe.


hush-hush (approved):

daniel-taf (approved):
I feel fairly confident about the safety of the change. I think our e2e tests should be good coverage as they cover our uses cases for workloadmeta that would be affected and they both passed (`process` and `language_detection`).

PR #35966, opened 2025-04-09
Build system-probe on macos

alopezz:
### What does this PR do?

Adds system-probe to the macOS package, supporting the traceroute feature.

~This PR only enables system-probe on "system-wide" installs for now; what to do for "single-user" installs is still being discussed and could be added later~. This enables system-probe regardless of the install type.

### Motivation

[BARX-935](https://datadoghq.atlassian.net/browse/BARX-935)

- Functionally: @AlexandreYang first verified that, building and running the agent with system-probe manually on macOS supported the traceroute feature correctly. I then confirmed that it worked on an installed package by running a deploy pipeline and installing the Agent via the install script as updated in this PR, on a VM.
- I tested downgrading to a version without system-probe and upgrading back to the one with it in several combinations and the behavior fulfills expectations: system-probe gets removed when downgrading, it comes back on when installing the new version. All of this on system wide installation mode.

Given that we can't have e2e tests for macOS and that it's packaging / installation related, I'm still tagging this as rc-required for us to double check that this works properly at release time.

### Additional Notes

- The plist file for system probe, as implemented, doesn't start system probe until the config file is created, and it will start it automatically when that file is created.
- We've confirmed that startup order between core agent and system probe doesn't affect behavior.

[BARX-935]: https://datadoghq.atlassian.net/browse/BARX-935?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


louis-cqrl (approved):

AlexandreYang (approved):

hmahmood (approved):

estherk15 (approved):

cmd/agent/install_mac_os.sh
```diff
@@ -376,12 +378,23 @@ printf "\033[34m\n    - Mounting the DMG installer...\n\033[0m"
 $sudo_cmd hdiutil attach "$dmg_file" -mountpoint "/Volumes/datadog_agent" >/dev/null
 if [ "$systemdaemon_install" != false ] && [ -f "$systemwide_servicefile_name" ]; then
     printf "\033[34m\n    - Stopping systemwide Datadog Agent daemon ...\n\033[0m"
-    # we use "|| true" because if the service is not started/loaded, the commands fail
+    # we use "|| true" because if the sIervice is not started/loaded, the commands fail
```
FlorentClarret, line 381:
```suggestion
    # we use "|| true" because if the service is not started/loaded, the commands fail
```

nit

Pythyu, line 381:
```suggestion
    # we use "|| true" because if the service is not started/loaded, the commands fail
```

cmd/agent/install_mac_os.sh
```diff
@@ -455,6 +468,15 @@ else
     $sudo_cmd launchctl kickstart "system/$service_name"
 fi
 
+# Set up and start the system-probe service if this version includes support for it
+if [ -f "${etc_dir}/com.datadoghq.sysprobe.plist.example" ]; then
```
FlorentClarret, line 472: Super nit question: should we use the `sysprobe_service_name` variable there instead? (same question for two lines after)

cmd/agent/install_mac_os.sh
```diff
@@ -455,6 +468,15 @@ else
...
+    printf "\033[34m\n* Setting up system-probe ($sysprobe_service_name) as a systemwide LaunchDaemon ...\n\n\033[0m"
+    $sudo_cmd mv "${etc_dir}/com.datadoghq.sysprobe.plist.example" "${sysprobe_servicefile_name}"
+    $sudo_cmd chown "0:0" "$sysprobe_servicefile_name"
+    $sudo_cmd launchctl load -w "$sysprobe_servicefile_name"
```
FlorentClarret, line 476: We have a mix of `${sysprobe_servicefile_name}` and `$sysprobe_servicefile_name` there, should we try to be consistent?

Pythyu (approved):
Otherwise LGTM

cmd/system-probe/modules/traceroute.go
```diff
@@ -3,7 +3,7 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2016-present Datadog, Inc.
 
-//go:build linux || windows
+//go:build linux || windows || darwin
```
FlorentClarret, line 6: Probably a stupid question but: in `[sysprobe.go](https://github.com/DataDog/datadog-agent/pull/35966/files#diff-d4497ddb96f01b74afcbc2d08a7c1feb7d1e421732da596ea65c6d1a875121c7)`, we simply remove the line. Why here we add `darwin` and not just drop the line?

cmd/agent/install_mac_os.sh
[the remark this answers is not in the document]
alopezz: It is intentional that single-user installations also set up system-probe.

cmd/system-probe/modules/traceroute.go
alopezz: I think it wasn't intentional (I modified one of the files and it was @AlexandreYang who changed the one you're pointing at I think) and it should be possible to drop it here without harm, too, I'll do that.

cmd/agent/install_mac_os.sh
alopezz: We definitely could, this was just based on https://github.com/DataDog/datadog-agent/blob/a5b59ed4f7ad2e77227af8ef2766691e8783792e/omnibus/package-scripts/agent-dmg/postinst#L103

This could definitely be better set as a variable too as an improvement.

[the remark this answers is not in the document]
Pythyu: I see so even in single-user it they will have a system-wide process for system-probe at least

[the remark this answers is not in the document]
alopezz: Yup, though it will only start if they actually configure system-probe.

FlorentClarret (approved):

PR #35967, opened 2025-04-09
Fix clusterchecks dispatching on Cloud Foundry Cluster Agent

NouemanKHAL:
### What does this PR do?

Fixes the `CloudFoundryBBS` listener name to its previous valid state to match the config provider's name `cloudfoundry-bbs`.

Listener Name:
https://github.com/DataDog/datadog-agent/blob/5c073c1e4d8c7129579897a0b404d5475f64c912/comp/core/autodiscovery/listeners/listeners.go#L10

Provider Name:
https://github.com/DataDog/datadog-agent/blob/5c073c1e4d8c7129579897a0b404d5475f64c912/comp/core/autodiscovery/providers/names/provider_names.go#L14

These two values are hardcoded in the datadog-cluster.yaml we generate to be used on Cloud Foundry.

### Motivation

Fixes the clusterchecks dispatching due to an incorrect listener_name that was modified during a refactor.

Deployed the change on a PCF environment and validated that it works after this fix using two environments with running nginx apps.

None.


comp/core/autodiscovery/listeners/listeners.go
```diff
@@ -7,7 +7,7 @@
 package listeners
 
 const (
-	cloudFoundryBBSListenerName = "cloudfoundry_bbs"
+	cloudFoundryBBSListenerName = "cloudfoundry-bbs"
```
davidor, line 10: I think we should support both names to maintain compatibility.
Otherwise, this will break for users who have already adopted the new name, since it’s been there for several releases.

NouemanKHAL: @davidor  That's a good point, but not in the case of Cloud Foundry because users install the DCA through our own packaged BOSH release. And In that package, we hardcode the listener-name's value:

https://github.com/DataDog/datadog-cluster-agent-boshrelease/blob/master/jobs/datadog-cluster-agent/templates/datadog-cluster.yaml.erb#L50-L51.

We're gonna change this value temporarily to make a bugfix release and revert it to this "original" name by the next agent release.

https://github.com/DataDog/datadog-cluster-agent-boshrelease/compare/master...noueman/fix-cloud-foundry-listener-name

buraizu (approved):

comp/core/autodiscovery/listeners/listeners.go
davidor: Thanks for clarifying. I'm not familiar with how the cloudfoundry integration works. If that's the only way of using it, it should be fine :+1:

davidor (approved):

comp/core/autodiscovery/listeners/listeners.go
NouemanKHAL: Thanks @davidor, I will try to clarify these information in the description in the future.

PR #32611, opened 2025-01-02
Lazy load rtloader when scheduling a python check

pgimalac:
### What does this PR do?
Initialize Python and rtloader when a first Python check is scheduled, rather than when starting the agent.

### Motivation
Eventually this will allow not loading Python if no Python check is scheduled.
Right now this doesn't change anything (in particular no change in resource usage), since we need to load Python before loading any check anyway (including Go checks).
 
### Describe how you validated your changes

We have some e2e tests running custom Python checks and ensuring they behave as expected, so if they pass then this change works properly.

### Additional Notes

This only makes sense if we can avoid loading Python for many users. 
In particular next steps are:
- moving default Python checks to Go (eg. https://github.com/DataDog/datadog-agent/pull/33116 for `disk`) 
- avoid needing to load Python to load a Go check (which is the case now)

PR #35972, opened 2025-04-09
Avoid caching python version in status header

KevinFairise2:
### What does this PR do?

Now that RTLoader is lazy loaded, if the python version is retrieved only at startup it remains `n/a` all the time.
Without that PR the `datadog-agent status` command always return `python_version: n/a`

Likely broken since: https://github.com/DataDog/datadog-agent/pull/32611


pgimalac (approved):
Thanks for fixing !

hush-hush (approved):

PR #35974, opened 2025-04-09
Remove global Datadog config accessor from most agent tests

dustmop:
### What does this PR do?

Instead of using pkgconfigsetup.Datadog() in the tests, prefer the mock constructor from pkg/config/mock. This makes testing easier as we rely more on the mock for testing the config's behavior.

### Motivation

Better testability to use the mock everywhere possible.

Unit tests all still pass. Running with DD_CONF_NODETREEMODEL=enable fixes a few previously failing tests.


GustavoCaso (approved):
Amazing work!!!

Should we add a rule in our [`forbidigo`](https://github.com/DataDog/datadog-agent/blob/5d3bc1db22621205848564c84e90a882ddb4fe0b/.golangci.yml#L737)  linter to ensure people do not default to using `pkgconfigsetup.Datadog()` in the future?

dustmop:
> Amazing work!!!
> 
> Should we add a rule in our [`forbidigo`](https://github.com/DataDog/datadog-agent/blob/5d3bc1db22621205848564c84e90a882ddb4fe0b/.golangci.yml#L737) linter to ensure people do not default to using `pkgconfigsetup.Datadog()` in the future?

I started looking into this, is there a way to make such a `forbidigo` rule only run over test files? Couldn't find anything in my searching. Removing this accessor from the non-test part of the codebase is probably not currently possible.

davidor (approved):

hmahmood (approved):

hush-hush (approved):

ogaca-dd (approved):

robertjli (approved):

soberpeach (approved):
LGTM 👍

antoninferrand (approved):
Approving for changes in `pkg/clusteragent/telemetry/collector_test.go`

mellon85 (approved):

pkg/collector/corechecks/cluster/kubernetesapiserver/kubernetes_openshift_test.go
```diff
@@ -30,9 +30,10 @@ func TestReportClusterQuotas(t *testing.T) {
 	json.Unmarshal(raw, &list)
 	require.Len(t, list.Items, 1)
 
-	prevClusterName := pkgconfigsetup.Datadog().GetString("cluster_name")
-	pkgconfigsetup.Datadog().SetWithoutSource("cluster_name", "test-cluster-name")
-	defer pkgconfigsetup.Datadog().SetWithoutSource("cluster_name", prevClusterName)
+	mockConfig := configmock.New(t)
+	prevClusterName := mockConfig.GetString("cluster_name")
+	mockConfig.SetWithoutSource("cluster_name", "test-cluster-name")
+	defer mockConfig.SetWithoutSource("cluster_name", prevClusterName)
```
L3n41c, line 36: Is it still needed to save the previous `cluster_name` and to restore it at the end of the test execution?
Shouldn’t the new `test-cluster-name` value of `cluster_name` be scoped to the new `mockConfig`, the lifetime of which is limited to the execution of one test?

pkg/util/containerd/namespaces_test.go
```diff
@@ -68,16 +68,17 @@ func TestNamespacesToWatch(t *testing.T) {
...
 		t.Run(test.name, func(t *testing.T) {
-			pkgconfigsetup.Datadog().SetWithoutSource("containerd_namespaces", test.containerdNamespaceVal)
-			defer pkgconfigsetup.Datadog().SetWithoutSource("containerd_namespaces", originalContainerdNamespacesOpt)
+			cfg.SetWithoutSource("containerd_namespaces", test.containerdNamespaceVal)
+			defer cfg.SetWithoutSource("containerd_namespaces", originalContainerdNamespacesOpt)
 
-			pkgconfigsetup.Datadog().SetWithoutSource("containerd_exclude_namespaces", test.excludeNamespaceVal)
-			defer pkgconfigsetup.Datadog().SetWithoutSource("containerd_exclude_namespaces", originalExcludeNamespacesOpt)
+			cfg.SetWithoutSource("containerd_exclude_namespaces", test.excludeNamespaceVal)
+			defer cfg.SetWithoutSource("containerd_exclude_namespaces", originalExcludeNamespacesOpt)
```
L3n41c, line 81: Same remark here: if the `cfg` mock config has a lifetime limited to the execution of the test, do we still need to restore previous config parameters value in `defer` statements?

pkg/util/containerd/namespaces_test.go
```diff
@@ -154,16 +155,17 @@ func TestFiltersWithNamespaces(t *testing.T) {
...
```
L3n41c, line 168: Same here.

jmw51798 (approved):

pkg/util/containers/filter_test.go
```diff
@@ -483,11 +485,12 @@ func TestNewMetricFilterFromConfig(t *testing.T) {
 }
 
 func TestNewAutodiscoveryFilter(t *testing.T) {
-	resetConfig()
+	cfg := mock.New(t)
+	resetConfig(cfg)
```
L3n41c, line 489: As the `cfg` mock config is freshly created. It should already be “clean” and `resetConfig` shouldn’t be needed.

pkg/util/containers/filter_test.go
```diff
@@ -700,14 +703,14 @@ func TestParseFilters(t *testing.T) {
...
+func resetConfig(cfg pkgconfigmodel.Config) {
+	cfg.SetWithoutSource("exclude_pause_container", true)
+	cfg.SetWithoutSource("container_include", []string{})
+	cfg.SetWithoutSource("container_exclude", []string{})
+	cfg.SetWithoutSource("container_include_metrics", []string{})
+	cfg.SetWithoutSource("container_exclude_metrics", []string{})
+	cfg.SetWithoutSource("container_include_logs", []string{})
+	cfg.SetWithoutSource("container_exclude_logs", []string{})
+	cfg.SetWithoutSource("ac_include", []string{})
+	cfg.SetWithoutSource("ac_exclude", []string{})
```
L3n41c, line 715: This functions aims at reverting the config changes that used to be done on the config singleton.
Now that we have a mock config object, it should be smarter to just drop the mock config object and re-create a fresh one rather trying to reset one.

L3n41c (commented):
* Because each test may requires a specific configuration, potentially different between all tests, and
* because we used to have to modify the config singleton for that purpose,

tests used to implement the following pattern:
```
* Get the current value of the config parameter we want to modify.
* Modify this config parameter for the purpose of this test.
* Restore the previous value of the config parameter value to avoid impacting next tests
```

But the change introduced by this PR seems to replace the use of a config singleton by a per-test mock object the lifetime of which is limited to each test.

So, can’t we clean the above-mentioned pattern from everywhere?

zoedt (approved):

AlexandreYang (approved):

NouhaManai96 (approved):

rayz (approved):

hghotra (approved):
LGTM - Serverless ✅

dustmop:
> * Because each test may requires a specific configuration, potentially different between all tests, and
> * because we used to have to modify the config singleton for that purpose,
> 
> tests used to implement the following pattern:
> 
> ```
> * Get the current value of the config parameter we want to modify.
> * Modify this config parameter for the purpose of this test.
> * Restore the previous value of the config parameter value to avoid impacting next tests
> ```
> 
> But the change introduced by this PR seems to replace the use of a config singleton by a per-test mock object the lifetime of which is limited to each test.
> 
> So, can’t we clean the above-mentioned pattern from everywhere?

Because of the history you mention, the config mock is not acting as cleanly as it may appear. For example, the mock will often return a [cached version of the previous mock](https://github.com/DataDog/datadog-agent/blob/e8215e0b3273636d0522756712790445f3ca0d9f/pkg/config/mock/mock.go#L38), as tests rely on the global-state aspect of the config object. This is difficult to fix in general because some of the tests also construct a `config.Component` using `fxutil.Test`, while also using the global accessor in their implementation. It's admittedly messy, and there's probably a lot that can be cleaned up here, but the main goal of this PR is to: (1) limit the use of the global accessor and move towards banishing it from test code, and (2) adding the mock so that we have a predictable place where config initialization happens for tests.

pkg/collector/corechecks/cluster/kubernetesapiserver/kubernetes_openshift_test.go
dustmop: Removing the defer is a good idea, done.

pkg/util/containerd/namespaces_test.go
dustmop: Same here, removed the defer.

pkg/util/containers/filter_test.go
dustmop: Because the mock config caches its reference to the global config, it's not possible to reset its state by making a new local object. So this function to manually reset the state is still required. Arguably, this test could be split into smaller individual tests, but that's a larger change than I'd like to make in a PR that's already touching so many files.

pkg/util/containers/filter_test.go
dustmop: This instance can be removed since the config is new here, but it's necessary for the later calls.

L3n41c (approved):

PR #35975, opened 2025-04-09
[EBPF-696] gpu: limit kernel launches and memory allocations held in memory

gjulianm:
### What does this PR do?

This PR adds limits to the number of kernel launches and memory allocations that are temporarily held in memory until the respective sync or free events are seen. This is achieved by two different ways:

* For the kernel launches, a hard limit on the length of the list. Whenever that limit is reached, a sync event is triggered and all the kernel launches are included in a kernel span.
* For the memory allocations, a LRU map that will evict the oldest entry if the map is full. 

The reasoning for having two different methods of "release" is to mirror how they are released in normal operation: kernel launches between two sync events are all grouped in the same span, but memory allocations are accounted one-by-one once we see the corresponding free.

### Motivation

Avoid accumulating memory when we are not receiving or not taking into account sync or memory free events.

Added unit tests to validate the new scenarios.

Note that the list of spans and the list of finished allocations do not have limits. This is consistent with the rest of system-probe modules, where we don't flush until the agent check performs a query.


hush-hush (approved):

pkg/gpu/stream.go
```diff
@@ -162,37 +183,79 @@ func (sh *StreamHandler) handleKernelLaunch(event *gpuebpf.CudaKernelLaunch) {
...
 
 	sh.allocations = append(sh.allocations, &data)
-	delete(sh.memAllocEvents, event.Addr)
+}
+
+func (sh *StreamHandler) onEvictedAlloc(_ uint64, value gpuebpf.CudaMemEvent) {
+	// If we are suspending the evict callback, don't create memory allocations.
+	// The reason is that the LRU cache calls the eviction callback whenever an item is removed for
+	// any cause (full cache or manual removal). In handleMemEvent, we manually remove items from the cache,
+	// so we don't want to create another memory allocation in that case.
```
val06, line 242: that sounds complicated, do we really need to use LRU here? why not some simpler data structure like circular buffer?

pkg/gpu/stream.go
```diff
@@ -162,37 +183,79 @@ func (sh *StreamHandler) handleKernelLaunch(event *gpuebpf.CudaKernelLaunch) {
...
+	if err != nil {
+		if logLimitStreams.ShouldLog() {
+			log.Warnf("Error getting current time: %v", err)
+		}
+		return
+	}
+
+	// This is probably a leak, it's an allocation that was never freed and it got to the point
+	// where our cache was full. But we don't know for sure, so don't mark it as a leak.
+	sh.createMemoryAllocation(uint64(nowTs), &value, false)
```
val06, line 257: that is also very confusing and feels like it should be here.
evict handler should have only one job: clean the data structure and not create additional entries

pkg/gpu/stream.go
```diff
@@ -308,7 +371,7 @@ func (sh *StreamHandler) getPastData(flush bool) *streamData {
 // getCurrentData returns the current state of the stream (kernels that are still running, and allocations that haven't been freed)
 // as this data needs to be treated differently from past/finished data.
 func (sh *StreamHandler) getCurrentData(now uint64) *streamData {
-	if len(sh.kernelLaunches) == 0 && len(sh.memAllocEvents) == 0 {
+	if len(sh.kernelLaunches) == 0 && sh.memAllocEvents.Len() == 0 {
```
val06, line 374: why one is using a built-in `len` and the other one has `Len()` method?

val06 (changes requested):
reviewed, still need to check the tests more deeply

pkg/gpu/config/config.go
```diff
@@ -36,6 +36,10 @@ type Config struct {
 	KernelCacheQueueSize int
 	// RingBufferSizePagesPerDevice is the number of pages to use for the ring buffer per device.
 	RingBufferSizePagesPerDevice int
+	// MaxKernelLaunchesPerStream is the maximum number of kernel launches to process per stream before forcing a sync.
+	MaxKernelLaunchesPerStream int
+	// MaxMemAllocEventsPerStream is the maximum number of memory allocation events to process per stream before evicting the oldest events.
+	MaxMemAllocEventsPerStream int
```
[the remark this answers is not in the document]
gjulianm: Added limit to the number of streams too.

pkg/gpu/stream.go
```diff
@@ -9,28 +9,42 @@ package gpu
...
-	metadata       streamMetadata
-	kernelLaunches []enrichedKernelLaunch
-	memAllocEvents map[uint64]gpuebpf.CudaMemEvent
-	kernelSpans    []*kernelSpan
-	allocations    []*memoryAllocation
-	processEnded   bool // A marker to indicate that the process has ended, and this handler should be flushed
-	sysCtx         *systemContext
+	metadata             streamMetadata
+	kernelLaunches       []enrichedKernelLaunch
+	memAllocEvents       *lru.LRU[uint64, gpuebpf.CudaMemEvent]
```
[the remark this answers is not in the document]
gjulianm: Added

pkg/gpu/stream.go
gjulianm: We need to know the event based on the address, so we need some kind of map, and also some way to evict the oldest element, so an LRU seemed to fit well.

pkg/gpu/stream.go
gjulianm: `kernelLaunches` is still a slice, but the other one is not a map anymore so we can't use the built-in len

pkg/gpu/stream_collection.go
```diff
@@ -38,10 +40,12 @@ type globalStreamKey struct {
...
 type streamCollection struct {
-	streams       map[streamKey]*StreamHandler
-	globalStreams map[globalStreamKey]*StreamHandler
-	sysCtx        *systemContext
-	telemetry     *streamCollectionTelemetry
+	streams         map[streamKey]*StreamHandler
+	globalStreams   map[globalStreamKey]*StreamHandler
+	sysCtx          *systemContext
+	telemetry       *streamCollectionTelemetry
+	streamTelemetry *streamTelemetry
```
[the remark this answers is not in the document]
gjulianm: Renamed both items, but the idea is that the `streamTelemetry` is only passed to the streams, to avoid having to create new objects for each (we can't create multiple counters with the same name)

pkg/gpu/config/config.go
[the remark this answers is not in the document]
val06: can we simplify this somehow. too complex to track 3 different parameters. can we have one limit number in the config for total amount of events?

pkg/gpu/stream.go
val06: lets keep it as simple as possible. let's loose the data (maybe "log" it via some internal telemetry to track the frequency of this edge-case

pkg/gpu/stream_collection.go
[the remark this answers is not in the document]
val06: let's refactor that into a separate struct then that will have 2 fields for different type of telemetry.

 I still think that it is too complicated and not worth the gain. would be much simpler to have one object that we operate and each data struct will update the relevant counters

pkg/gpu/stream.go
val06: perhaps once we remove the evictCallback we can also update the comment

pkg/gpu/stream.go
```diff
@@ -162,37 +183,79 @@ func (sh *StreamHandler) handleKernelLaunch(event *gpuebpf.CudaKernelLaunch) {
...
 	if !ok {
-		log.Warnf("Invalid free event: %v", event)
+		if logLimitStreams.ShouldLog() {
+			log.Warnf("Invalid free event: %v", event)
+		}
+		sh.telemetry.invalidFreeEvents.Inc()
 		return
 	}
 
+	sh.createMemoryAllocation(event.Header.Ktime_ns, &alloc, false)
```
val06, line 217: why do we report memory allocation event only after getting the free event?
we shouldn't depend on the "proper" lifecycle of memory management. what if free is never called?

pkg/gpu/config/config.go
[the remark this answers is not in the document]
gjulianm: The problem with the total number of events is that then we'd need to know how many streams are there, so it complicates the code as each stream needs to know about how much do the other ones have. I was thinking about having "max events" and then "max streams", but because the LRU for the memory allocations needs a fixed size beforehand, we'd need to hardcode how much of the "max events" limit we want to assign to each event type.

pkg/gpu/stream_collection.go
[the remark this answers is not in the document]
gjulianm: Changed to have only one telemetry struct.

pkg/gpu/stream.go
gjulianm: If free is never called we keep on reporting the allocation as part of the "current data" set. We use the free only to determine when the allocation is no longer active.

pkg/gpu/stream.go
```diff
@@ -200,7 +199,10 @@ func (sh *StreamHandler) handleKernelLaunch(event *gpuebpf.CudaKernelLaunch) {
 
 func (sh *StreamHandler) handleMemEvent(event *gpuebpf.CudaMemEvent) {
```
val06, line 200: this is more clear now tbh.
i would consider renaming some related variables (or even types) to `completedAllocation`, `activeAllocation`

gjulianm: I can give it another check in a separate PR refactoring a bit that part, I agree that it's confusing

val06 (approved):

PR #35983, opened 2025-04-09
[CWS] Take ttl into account for non-array variables

YoannGh:


paulcacheux (changes requested):
NACK

lebauce:
@YoannGh Could you please rebase it now that we have only one go routine ?

paulcacheux (approved):

PR #35990, opened 2025-04-09
[AGENTRUN-213] Add disk check e2e test comparing versions

pgimalac:
### What does this PR do?
Add an e2e test running the old and new version of the disk check and comparing emitted metrics.

### Motivation
Catch differences between the two versions as well as regressions. 

This is an e2e test so CI is enough.


test/new-e2e/tests/agent-runtimes/check_disk/disk_test.go
```diff
@@ -0,0 +1,183 @@
...
+func (v *baseCheckSuite) runDiskCheck(agentConfig string, checkConfig string, useNewVersion bool) []metricPayload {
+	if useNewVersion {
+		agentConfig += "\nuse_diskv2_check: true"
+		checkConfig += "\n  loader: core"
+	}
+
+	v.UpdateEnv(awshost.Provisioner(
+		awshost.WithAgentOptions(
+			agentparams.WithAgentConfig(agentConfig),
+			agentparams.WithFile("/etc/datadog-agent/conf.d/conf.yaml", checkConfig, true),
```
GustavoCaso, line 131: Should we write to `/etc/datadog-agent/conf.d/disk.d/conf.yaml` ?

pgimalac: Ah yes, I'm not sure how that worked 😕

test/new-e2e/tests/agent-runtimes/check_disk/disk_test.go
```diff
@@ -0,0 +1,167 @@
...
+	for i := range metrics {
+		// remove the disk_check_version tag
+		tagLen := len(metrics[i].Tags)
+		metrics[i].Tags = slices.DeleteFunc(metrics[i].Tags, func(tag string) bool {
+			return tag == diskCheckVersionTag
+		})
+		if !assert.Equalf(v.T(), 1, tagLen-len(metrics[i].Tags), "expected tag %s once in metric %s, but found %d times", diskCheckVersion, metrics[i].Metric, tagLen-len(metrics[i].Tags)) {
+			v.T().Logf("metric: %+v", metrics[i])
+		}
+	}
```
GustavoCaso, line 164: I looks a bit complex the check for the custom tag 😄 

```go
for i := range metrics {
	tagsWithoutCustomTags := []string{}
	foundCustomTags := false
	// remove the disk_check_version tag
	for _, t := range metrics[i].Tags {
		if t == diskCheckVersionTag {
			foundCustomTags = true
			continue
		}
		tagsWithoutCustomTags = append(tagsWithoutCustomTags, t)
	}

	if !foundCustomTags {
		v.T().Logf("metric %s does not have the disk_check_version tag", metrics[i].Metric)
	}
	metrics[i].Tags = tagsWithoutCustomTags
}
```

Do this alternatives do the same thing?

pgimalac: This wouldn't detect if it's twice in the list I believe ?

Do you really find the code complex though ? It just removes the tag and asserts only one element was removed
I can store `tagLen-len(metrics[i].Tags)` in a variable to make it clear it's the number of elements removed if you want (and maybe shorten the assert message)

pgimalac: I updated the code a bit, let me know if it's better

GustavoCaso: > This wouldn't detect if it's twice in the list I believe ?

You are right about that 😄

The updated code looks much readable 👏

GustavoCaso (approved):
LGreatTM 🎉

GustavoCaso (approved):
LGTM

jose-manuel-almaza (approved):

PR #35993, opened 2025-04-09
[OTAGENT-355] Improve config in serializer exporter

songy23:
### What does this PR do?
1. Add static validations on configs, copied from https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/datadog/config
2. Move queue configs to a dedicated sub-section `sending_queue`. In the new versions queue configs can no longer be top-level exporter configs. 

### Motivation
related to http://github.com/open-telemetry/opentelemetry-collector/issues/12709#issuecomment-2790524593

unit tests

### DIFF
--- a/comp/otelcol/otlp/components/exporter/serializerexporter/factory.go
+++ b/comp/otelcol/otlp/components/exporter/serializerexporter/factory.go
@@ -7,6 +7,7 @@ package serializerexporter
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sync"
 	"time"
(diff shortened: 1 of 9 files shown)


liustanley (approved):

comp/otelcol/otlp/components/exporter/serializerexporter/factory.go
```diff
@@ -156,10 +156,21 @@ func (f *factory) Reporter(params exp.Settings, forwarder defaultforwarder.Forwa
 	return f.reporter, reporterErr
 }
 
+// checkAndCastConfig checks the configuration type and its warnings, and casts it to
+// the Datadog Config struct.
+func checkAndCastConfig(c component.Config, logger *zap.Logger) *ExporterConfig {
+	cfg, ok := c.(*ExporterConfig)
+	if !ok {
+		panic("programming error: config structure is not of type *ExporterConfig")
```
truthbk, line 164: This is fine, but I normally prefer to avoid panics... If this is a generalized pattern, then it's OK, but I'm wondering if it wouldn't be better to return `nil`, or change the function to also return an `error`, and then return the error with `createMetricExporter` (which I think would have the same effect and prevent the process from running as well)?

truthbk (commented):
Looks good, just a quick question on the panicking code.

songy23:
going to hold off the PR until 7.67, to be merged together with the version upgrades

PR #35994, opened 2025-04-09
chore(autoinstrumentation): simpler injector mutator

stanistan:
### Motivation

While working on #35773 there are a few unrelated small changes that I wanted to make to how the injector/mutator is created so that it's less clearer.

Added more unit tests and existing tests pass.

### DIFF
--- a/pkg/clusteragent/admission/mutate/autoinstrumentation/injector.go
+++ b/pkg/clusteragent/admission/mutate/autoinstrumentation/injector.go
@@ -196,7 +196,7 @@ func injectorDebug(boolean string) injectorOption {
 	}
 }
 
-func newInjector(startTime time.Time, registry, imageTag string, opts ...injectorOption) *injector {
+func newInjector(startTime time.Time, registry string, opts ...injectorOption) *injector {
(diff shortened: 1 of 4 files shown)


pkg/clusteragent/admission/mutate/autoinstrumentation/injector.go
```diff
@@ -196,7 +196,7 @@ func injectorDebug(boolean string) injectorOption {
 	}
 }
 
-func newInjector(startTime time.Time, registry, imageTag string, opts ...injectorOption) *injector {
+func newInjector(startTime time.Time, registry string, opts ...injectorOption) *injector {
```
betterengineering, line 199: Should registry and startTime also be opts? 🤔

betterengineering (approved):

pkg/clusteragent/admission/mutate/autoinstrumentation/injector.go
stanistan: Thinking about dropping `newInjector` entirely but that might be for another follow up. Thanks for the review.

PR #35995, opened 2025-04-09
[CONTP-437] Support "check_tag_cardinality" change in e2e

zhuminyi:
### What does this PR do?
Following this test-infra [change](https://github.com/DataDog/test-infra-definitions/pull/1246) , nginx integration test has a customized config on `check_tag_cardinality`
```yaml
init_config:
check_tag_cardinality: high
```
which overrides the global Agent-level tag cardinality for a specific check. 
The high cardinality tags will appear in new metrics "nginx.net.connections", ref [doc](https://datadoghq.atlassian.net/wiki/spaces/TS/pages/328438514/Container+Tags+on+Integration+Metrics+or+Custom+Metrics+and+Origin+Detection#Tagging-Integration-Metrics)


adel121 (approved):

adel121 (approved):

test/new-e2e/tests/containers/k8s_test.go
```diff
@@ -542,6 +542,41 @@ func (suite *k8sSuite) TestNginx() {
 		},
 	})
 
+	suite.testMetric(&testMetricArgs{
+		Filter: testMetricFilterArgs{
+			Name: "nginx.net.connections",
```
L3n41c, line 547: If DataDog/test-infra-definitions#1246 had an impact on the tags cardinality of the `nginx` check, it should already be visible on the metric assertion for `nginx.net.request_per_s` done above, at line 491.
This new assertion doesn’t increase the test coverage as we only see that, for the same check on the same pods, we get exactly the same tag set.

PR #36003, opened 2025-04-10
Bump the timeout to 2h 30m for the release jobs

FlorentClarret:
### What does this PR do?

Bump the timeout to 2h 30m for the release jobs

### Motivation

- We have way to many artefacts in the db and the job timeout on nightly build. We need to clean them but it will take some time and effort. Bumping the timeout for now
- [This PR](https://github.com/DataDog/datadog-agent/pull/35642) was not enough because the default timeout for the command is [2h](https://github.com/DataDog/datadog-agent/blob/06fc75a956283bd78afc9f0ea7b08f2da3ccd50d/tasks/pipeline.py#L331), which can be seen [here](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/890044559)


kacper-murzyn (approved):

chouetz (approved):
We can maybe remove the `timeout: 2h 30m` is it's useless.

PR #36004, opened 2025-04-10
Optimize unit tests

KSerrania:
### What does this PR do?

Adds the `-parallel` option to `go test` commands, to allow parallel runs of tests within a package (if tests specify the `t.Parallel()` line).

Updates a few long-running unit tests to be shorter.

The changes for each impacted test are as follows:
- `pkg/api/util.TestDeadline`: lasts for 10s because it waits for a timeout, set at the default `auth_init_timeout` value of 10 seconds. Reduced to 1s, could likely be reduced further?
  - 9s improvement
- `pkg/collector/worker.TestWorkerUtilization`: contains various arbitrary sleep values (500ms, 2s, 2s -> 200ms, 1.2s, 1.2s). Updated to use `EventuallyWithT` instead, ensuring the test completes as soon as possible.
  - 2s improvement
- `pkg/trace/stats/oteltest`: two tests that run for 20s each. Made them run in parallel, reducing the total duration of the package from 40s to 20s.
  - 20s improvement
- `test/fakeintake/client.TestClient`: tests that validate error cases run for 20s because they use the default retry parameters (4 retries, 5 second backoff). Updated the client to allow specifying retry settings, and changed them in the tests to last for 20ms instead
  - 20s improvement
- `pkg/trace/api`: several tests last for 5s because the `ReceiverTimeout` is set to 5s, which makes the test wait 5s in the `defer` block for the receiver to stop. Updated the timeout to 1s (I would have liked to reduce it further, but the timeout is an int value in seconds).
  - 10s improvement
- `comp/core/workloadmeta/collectors/internal/kubeapiserver`: contains a few tests that run for 5s each. Made them run in parallel.
  - 10s improvement
- `pkg/logs/client/http`: the test server created for most tests in this package and in `pkg/logs/sender` has backoff values that cause long wait times when retries are triggered (up to 18s for the longest tests). Reduced the base to 0.01 instead of 1 to make all tests significantly faster.
  - 1m improvement across both packages
- `comp/otelcol/ddprofilingextension/impl`: each test starts the extension with a 5s polling period. Reduced the polling value to 1s.
  - 15s improvement

Each test is changed in its own commit.
 
### Motivation

After investigating the length of our unit tests and optimizing their runs (see https://github.com/DataDog/datadog-agent/pull/35176), I started looking into the tests themselves (see [this doc](https://docs.google.com/document/d/1FUKlq8Faw5obqyEuIFXLZ_cjQ0JnPLwzFZZMFW1sQ88/edit)).

What I detected is that a significant number of packages run for very long durations (104 tests or packages with run time >= 5s, 20+ tests or packages with run time >= 15s).
Some of these can be parallelized (using the `t.Parallel()` option), others can be rewritten to not wait for unwarranted amounts of time.

This PR only improves a fraction of the long-running tests. More efforts would be needed to improve all of them, sometimes requiring specific knowledge on the area covered by the tests. For instance, there were several packages I could not improve with parallelization because of the use of globals.

The main goal of this PR is to showcase the kinds of issues we have in unit tests that make them inherently slow, as well as potential solutions.

This PR should also slightly improve the total unit test run time in CI (though only by a few seconds, since in CI test packages are run in parallel already, 16 at a time), and strongly improve local test runs (with `--only-impacted-packages`) that run the packages I improved (because runs against a limited number of packages will not make efficient use of package-level parallelism).

All unit tests pass in CI.

n/a


pgimalac (approved):
🚀 :shipit:

pkg/collector/worker/worker_test.go
```diff
@@ -331,14 +331,14 @@ func TestWorkerUtilizationExpvars(t *testing.T) {
...
-	time.Sleep(500 * time.Millisecond)
+	time.Sleep(200 * time.Millisecond)
 	require.InDelta(t, getWorkerUtilizationExpvar(t, "worker_2"), 0, 0)
 
 	// High util checks should be reflected in expvars
 
 	pendingChecksChan <- blockingCheck
 
-	time.Sleep(2000 * time.Millisecond)
+	time.Sleep(1200 * time.Millisecond)
```
KSerrania, line 341: This change is the one I am least confident in. Locally and in CI, I did not get any failures after several dozens of attempts ([CI Visibility](https://app.datadoghq.com/ci/test/runs?query=test_level%3Atest%20env%3Aprod%20%40git.repository.id%3Agitlab.ddbuild.io%2FDataDog%2Fdatadog-agent%20%40test.service%3Adatadog-agent%20%40git.branch%3Akserrania%2Ftest-optim%20%40test.name%3ATestWorkerUtilizationExpvars&agg_m=count&agg_m_source=base&agg_t=count&citest_explorer_sort=time%2Cdesc&fromUser=false&index=citest&mode=sliding&saved-view-id=2721651&start=1743775027596&end=1744379827596&paused=false)), but I did get flakes with lower values like 1s.

ajgajg1134 (approved):
nice!

pkg/collector/worker/worker_test.go
vickenty: Would it make sense to replace with `EventuallyWithT`?

CelianR (approved):

chouetz (approved):

jose-manuel-almaza (approved):

ddrthall (approved):

pkg/collector/worker/worker_test.go
KSerrania: Good idea, I'll push a commit to use it.

vickenty (approved):

davidor (approved):

truthbk (approved):

PR #36005, opened 2025-04-10
langdetect: Only include privileged detectors when needed

vitkyrka:
### What does this PR do?

Move the privileged detectors to a separate package so that they do not get included where they are not needed (such as in the core agent).

### Motivation

Remove unnecessary code in core-agent.
Noticed in https://github.com/DataDog/datadog-agent/pull/35865.

Covered by automated tests.


smeso-ddg (approved):
Awesome! Thanks!

PR #36006, opened 2025-04-10
Olivierg/smp use full image

ogaca-dd:
### What does this PR do?
This PR use the `full` docker image to run SMP experiments

### Motivation
The `full` docker image contains ddot, so it  allows to write experiments for ddot.


sblumenthal (approved):

goxberry:
Note to other reviewers and for my future self: this PR is follow-up from #35953. That PR adds the "full (Agent) image" that includes DDOT. Before this PR, we were using an image without DDOT.

goxberry (approved):
Looks good to me. Thank you for working on adding DDOT experiments!

PR #36024, opened 2025-04-10
[AGENTRUN-214] Add `pkg/template` containing patched code from Go stdlib

pgimalac:
### What does this PR do?
Add patched version of `text/template` and `html/template` in `pkg/template`.

### Motivation
This version of templates doesn't support methods, so that we don't disable dead code elimination when using it.

`pkg/template` code still uses `"text/template/parse"` from the stdlib, and some exported types are aliases of the types in the standard library to avoid typing conflicts (this can be removed once everything in the agent uses the patched version).

### Additional Notes

Other PRs will be opened on top of this one.
This is a step to enable dead code elimination in agent binaries.


CelianR (approved):

pkg/template/generate.sh
```diff
@@ -0,0 +1,44 @@
...
+
+set -e
+
+# This script generates the code for the template package.
+# It takes the code from the Go standard library and applies the patches.
+
+if ! command -v gopatch &> /dev/null; then
+    echo "gopatch could not be found, installing it..."
+    go install github.com/uber-go/gopatch@latest
+fi
```
GustavoCaso, line 11: Should we move this to the `internal/tools/tools.go` file? That way when calling `inv install-tools` hey get install for everyone 😄

pkg/template/generate.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
+set -e
```
GustavoCaso, line 3: Should we move this script to a invoke task?

pkg/template/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/DataDog/datadog-agent/pkg/template
```
GustavoCaso, line 1: Would this module be available for download from outside the repo? If so, should we make it an internal module? That way we do not expose it to outside of this Agent repo 😄

pkg/template/generate.sh
GustavoCaso: Also, should we take into consideration Windows as well in the script?

pkg/template/generate.sh
pgimalac: That's a good idea yes !

pgimalac: Right now this is not used in CI, but I'd like to make this automated when we do a Go update, so that would be better

pkg/template/generate.sh
pgimalac: Yeah I'll rewrite this as an invoke task, will be cleaner than a shell script, and more portable too

pkg/template/go.mod
pgimalac: I don't think you can have an internal module, internal is only about packages
TBH it's like any module in the agent repo, you can use it at your own risks 😄

GustavoCaso: > TBH it's like any module in the agent repo, you can use it at your own risks 😄

We should put that as a disclaimer somewhere 🤣

tasks/pkg_template.py
```diff
@@ -0,0 +1,78 @@
...
+    if not goroot:
+        print("Could not find Go's source code path")
+        sys.exit(1)
+    print(f"Using code from {goroot}")
+
+    print("Removing previous code...")
+    remove_dirs = ["text", "html", "internal"]
+    for dir_name in remove_dirs:
+        if os.path.exists(dir_name):
+            shutil.rmtree(dir_name)
```
GustavoCaso, line 37: Should the deletion be in the `pkg/template/..`  folder?

GustavoCaso: Same for the creation of directories 😄

pgimalac: Yeah that's the case, the current directory is `pkg/template` (there is a `chdir` at the top)

pgimalac: It avoids having to prefix every file with `pkg/template` or doing `ctx.cd`

internal/tools/tools.go
```diff
@@ -16,6 +16,7 @@ import (
 	_ "github.com/golangci/golangci-lint/cmd/golangci-lint"
 	_ "github.com/goware/modvendor"
 	_ "github.com/stormcat24/protodep"
+	_ "github.com/uber-go/gopatch"
```
GustavoCaso, line 19: You need to add the `gopatach` to the list as well https://github.com/DataDog/datadog-agent/blob/494f98eb5db09652258c0e0aac936a35d69d7b62/tasks/install_tasks.py#L15-L16

tasks/pkg_template.py
GustavoCaso: Nevermind I just saw `chdir("pkg/template"):` 🤦

internal/tools/tools.go
pgimalac: Good catch 👍

tasks/pkg_template.py
GustavoCaso: Should we move the `chdir("pkg/template"):` inside generate?

pgimalac: The whole function would have to run in the `with` block, that's why everything is in a `_generate` function, it avoids one indentation level everywhere

pkg/template/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Code changes
+The only functional change is that you can't call methods.
+The types like `FuncMap` or `HTML` are aliases of standard library types, so there is no conflict when using one with the other.
+
+## Organization
+The [text](./text) directory contains the code from `text/template`, the [html](./html) directory contains the code from `html/template`.
+
+## Code Generation
+The code in this directory can be re-generated using `invoke -e pkg-template.generate` from the root of the repository.
```
GustavoCaso, line 12: When I run the task I get an error regarding file permissions
```
inv -e pkg-template.generate
go version
Generating code for Go version go version go1.23.7 darwin/arm64
go env GOROOT
Using code from /Users/gustavo.caso/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.7.darwin-arm64
Removing previous code...
Creating directories...
Copying code from Go standard library...
Applying patches...
git apply no-method.patch
gopatch -p imports.gopatch ./...
open /Users/gustavo.caso/go/src/github.com/DataDog/datadog-agent/pkg/template/html/escape.go: permission denied; open /Users/gustavo.caso/go/src/github.com/DataDog/datadog-agent/pkg/template/html/template.go: permission denied
```

Also, it seems that some files are changed after running the task:
image

I'm not sure if that is the intended behavior

pgimalac: The change is because the task failed, one patch could not be applied

Can you `ls -l /Users/gustavo.caso/go/src/github.com/DataDog/datadog-agent/pkg/template/html/escape.go` to check the permissions ?
I'm guessing the files don't have write permissions, but I don't have this issue on my side
I can add a `chmod +w` in the task to be safe 🤷

GustavoCaso:
```
ls -l /Users/gustavo.caso/go/src/github.com/DataDog/datadog-agent/pkg/template/html/escape.go
-r--r--r-- 1 gustavo.caso staff 33163 Mar  6 09:38 /Users/gustavo.caso/go/src/github.com/DataDog/datadog-agent/pkg/template/html/escape.go
```

GustavoCaso (approved):
Validated that things are working on my machine 
image

Nice work 🎉 

Should validate with someone from the Windows team, that things are working as expected for them?

PR #36032, opened 2025-04-10
Fix various string/array bugs in Go DI

grantseltzer:
### What does this PR do?

Handles a few bugs in Go DI:

- There were issues where in some test cases strings would cause a verifier issue that's resolved by having an extra register assignment for string addresses.
- Fixes an issue where we weren't subtracting from the `seenTypes` map counters when returning up from depth first traversal of types, causing false positives of recursive types. For example, structs with 5+ strings would have issues.
- Various index/nil checks 

Also adds more e2e test fixtures

### Motivation

Bug fixes that I observed.

More e2e tests.


tylfin (approved):
Nice to see these incremental improvements

PR #36040, opened 2025-04-11
[AGENTRUN-214] Use pkg/template where easily possible

pgimalac:
### What does this PR do?
Update some uses of `text/template` or `html/template` to use `pkg/template` instead.

Also update the status component to use both versions, with `pkg/template` in "shadow-mode", warning when there is an error or a difference in output. This will help detect templates using methods.

The functional change is that the new package doesn't support method calls in templates, so any template which uses method calls will fail.

### Motivation
Eventually all uses of templates which end up in agent binaries should use `pkg/template` to enable dead code elimination.

I checked each impacted template manually. I only changed the ones where I could easily determine that there was no method call, or where I could easily remove the method calls. CI helped a little but we don't have that many tests around status templates.

#### QA

If you own any template used in the status of the agent, please check that they work correctly, in particular there should be no warning about "shadow pkg/template" having errors or different outputs.
This includes both the text template and the GUI html template.
If you can ensure that your templates don't use methods, you don't have to run the commands from the agent.

Tips to determine statically if a template uses methods:
- The syntax is the same to access a key, a struct field, and call a method, so pattern matching is impossible. However, you can still detect potential cases by searching for a dot followed by an uppercase (only exported methods can be called).
- Check if any object passed to the template has exported methods, eg. if you pass a `map[string]string` then it's also safe
- List exported methods of given objects and search for uses in the template

Note that templates can have branching, and objects are passed dynamically, so running the agent once is not enough.
To determine dynamically whether a template uses methods, you need to test the template with the same objects as used in the agent, and for each code path of the template.

### Additional Notes

Remaining uses of templates are:
- comp/core/status
- pkg/dynamicinstrumentation/codegen
- pkg/dynamicinstrumentation/ebpf
- pkg/ebpf/bytecode/runtime

There is no need to update tests, mocks, `main` packages outside of the agent ones, or any package which is not imported by any agent.

The static quality gates break because we now duplicate the template library in each agent binary.
It will decrease once every use of template is migrated (and decrease a lot once DCE is fully enabled).


paulcacheux (approved):

ichinaski (approved):

BaptisteFoy (approved):
Tested on the installer -- lgtm for Fleet files!

daniel-taf (approved):

rahulkaukuntla (approved):

comp/core/status/render_helpers.go
```diff
@@ -101,8 +105,25 @@ func RenderHTML(templateFS embed.FS, template string, buffer io.Writer, data any
...
+		return err
+	}
+
+	var pkgbuff bytes.Buffer
+	tt := pkghtmltemplate.Must(pkghtmltemplate.New(template).Funcs(HTMLFmap()).Parse(string(tmpl)))
+	terr := tt.Execute(&pkgbuff, data)
+	if terr != nil {
+		log.Warnf("Error executing shadow pkg/template/html for %s: %s", template, terr)
+	} else if pkgbuff.String() != stdbuff.String() {
+		log.Infof("Shadow pkg/template/html output does not match html/template output for %s", template)
```
GustavoCaso, line 123: Should we log the difference? Maybe as DEBUG, so customers do not see it, also showing a the does not match message as Info, might trip customers. Since the change is internal to them, maybe we can log both of the messages as warn or debug

GustavoCaso:
Should we add a `depguard` setting on our golangci configuration to ensure people do not use the Go template libraries? 

```
- pkg: "text/template"
  desc: "use the custom package github.com/DataDog/datadog-agent/pkg/template/text. This package ensures dead code elimination works in the Agent"
- pkg: "html/template"
  desc: "use the custom package github.com/DataDog/datadog-agent/pkg/template/html. This package ensures dead code elimination works in the Agent"
```

comp/core/status/render_helpers.go
pgimalac: The issue is that not all differences are unexpected, for example some templates read atomic variables, they can change between two executions, so we can't just warn about it
I think what we care most about is which template has a change, so that we can manually check whether it is using methods, logging a potentially very big template would be spammy...

go.mod
pgimalac, line 1: @GustavoCaso for the question of `depguard`, I considered it but actually importing the template package is not necessarily an issue.
For example in (unit / e2e) tests, or in any package not used by the agent binaries (we have a few which are used to generate code for example).

I'm thinking we can leave it allowed, and if people get it wrong then we can consider adding it to the denylist
If they try to use it in code which is imported by an agent then it will increase the size and fail in CI anyway

What do you think ?

comp/core/status/render_helpers.go
pgimalac: What's the issue with logs ? It's more convenient to print the error 😄 
The main issue we expect here is that templates using methods will fail
Having an info log when there is a diff is just to be safe, but that's not what should happen with this change

go.mod
GustavoCaso: That sounds great to me. We can also create a custom rule for only targeting folders that the Agent uses.
```
templates:
  files:
    - "pkg/**"
    - "comp/**"
    - "cmd/**"
  deny:
    - pkg: "text/template"
      desc: "use the custom package github.com/DataDog/datadog-agent/pkg/template/text. This package ensures dead code elimination works in the Agent"
    - pkg: "html/template"
      desc: "use the custom package github.com/DataDog/datadog-agent/pkg/template/html. This package ensures dead code elimination works in the Agent"
```

I will leave the choice up to you 😄

comp/core/status/render_helpers.go
GustavoCaso: There is nothing wrong with logging, but how are you planning on monitoring once this gets released? 
Since we can emit internal metrics, we can use those to monitor for these mismatches in a more holistic way than just logging. 

We can create a dashboard, etc.

go.mod
pgimalac: > we have a few which are used to generate code for example

The ones I was referring to here are in `pkg`, it's pretty much impossible to have a static rule for code actually used by the agent binaries...

But we can just add some exceptions for the specific uses which are fine, I'll add the rule 👍

comp/core/status/render_helpers.go
pgimalac: Just search for logs in the app ? (from internal deployments)
The metric wouldn't help with customer cases either, unless we add it to instrumentation telemetry (is that what you had in mind ?)

But even then we can't tag with the whole error message, at best we can tag with the name of the template file
We can discuss this separately but I'd rather having doing that in this PR anyway

GustavoCaso: We agreed that we would use the `agent telemetry` to send metrics regarding discrepancies on the customer side of things. @pgimalac Would add it on a separate PR

GustavoCaso (approved):
LGTM

julien-lebot (approved):

KevinFairise2 (approved):

PR #36041, opened 2025-04-11
Export function and types to parse check command output and use float64 for points

pgimalac:
### What does this PR do?
1. Use float64 as type for metric points
2. Export the function and types used to parse the check subcommand output

### Motivation
Points are really float64 (it happens that the ones in this test are integers, but it's still not the correct type to unmarshal the output).
This will be used by https://github.com/DataDog/datadog-agent/pull/35990, where metrics are really float64.

CI.

### DIFF
--- a/test/new-e2e/tests/agent-subcommands/check/parse.go
+++ b/test/new-e2e/tests/agent-subcommands/check/parse.go
@@ -13,25 +13,29 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-type root struct {
-	Aggregator aggregator `json:"aggregator"`
+// Root is the root object of the check command output
+type Root struct {
+	Aggregator Aggregator `json:"aggregator"`
 }
 
-type aggregator struct {
-	Metrics []metric `json:"metrics"`
+// Aggregator contains the metrics emitted by a check
+type Aggregator struct {
+	Metrics []Metric `json:"metrics"`
 }
 
-type metric struct {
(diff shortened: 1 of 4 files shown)


GustavoCaso (approved):
LGTM

test/new-e2e/tests/agent-subcommands/check/parse.go
```diff
@@ -13,33 +13,37 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-type root struct {
-	Aggregator aggregator `json:"aggregator"`
+// Root is the root object of the check command output
```
vickenty, line 16: Would it make sense to move this code to pkg? I don't think creating dependencies between tests are a good thing.

vickenty (approved):

test/new-e2e/tests/agent-subcommands/check/parse.go
pgimalac: If we do that we have to make a new module so that it can be imported...
I could create a shared package somewhere in `test/new-e2e` though 🤔

vickenty: I meant `test/new-e2e/pkg/`

pgimalac: I'm going to do that 👍

KevinFairise2 (approved):

PR #36046, opened 2025-04-11
trace-agent: collect process tags from first span

raphaelgavache:
### What does this PR do?
Support extracting process tags not only from headers but also from the first span. We're still discussing the favoured vector of transmission and the first span is a likely one.
Making sure that this agent transmits process tags, will cleanup unused method after implementation on tracer side

Added dedicated e2e test

### DIFF
--- a/pkg/trace/api/api.go
+++ b/pkg/trace/api/api.go
@@ -710,21 +710,21 @@ func droppedTracesFromHeader(h http.Header, ts *info.TagStats) int64 {
 
 // todo:raphael cleanup unused methods of extraction once implementation
 // in all tracers is completed
+// order of priority:
+// 1. tags in the v06 payload
+// 2. tags in the first span of the first chunk
+// 3. tags in the header
 func getProcessTags(h http.Header, p *pb.TracerPayload) string {
-	// Check if the process tags are not already set in the payload
 	if p.Tags != nil {
 		if ptags, ok := p.Tags[tagProcessTags]; ok {
 			return ptags
 		}
 	}
-	// Check if the process tags are set in the first span
-	span, ok := getFirstSpan(p)
-	if ok {
+	if span, ok := getFirstSpan(p); ok {
 		if ptags, ok := span.Meta[tagProcessTags]; ok {
 			return ptags
 		}
 	}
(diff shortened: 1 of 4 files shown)


amarziali (approved):

pkg/trace/api/api.go
```diff
@@ -709,10 +708,42 @@ func droppedTracesFromHeader(h http.Header, ts *info.TagStats) int64 {
...
+// in all tracers is completed
+func getProcessTags(h http.Header, p *pb.TracerPayload) string {
+	// Check if the process tags are not already set in the payload
+	if p.Tags != nil {
+		if ptags, ok := p.Tags[tagProcessTags]; ok {
+			return ptags
+		}
+	}
+	// Check if the process tags are set in the first span
+	span, ok := getFirstSpan(p)
```
stanistan, line 721: nit for consistency

```go
if span, ok := getFirstSpan(p); ok {
    // rest...
}
```

stanistan: This does prefers process tags on the tracer payload to the header, would we want that, or to merge them instead?

Is there any scenario where both things might be sent? If there is in the future, we would be disregarding the header and that feels like something that would be confusing.

raphaelgavache: yes that's on purpose, since for a given payload there's only one sender (tracer) it should have only one implementation. I'm pre-shooting multiple of them to avoid losing an agent, one will be picked and I'll cleanup the others after

pkg/trace/api/api.go
```diff
@@ -709,10 +708,42 @@ func droppedTracesFromHeader(h http.Header, ts *info.TagStats) int64 {
 	return dropped
 }
 
-func getProcessTagsFromHeader(h http.Header) string {
+// todo:raphael cleanup unused methods of extraction once implementation
+// in all tracers is completed
+// order of priority:
+// 1. tags in the v06 payload
```
ajgajg1134, line 714:
```suggestion
// 1. tags in the v07 payload
```

ajgajg1134 (approved):
small nit but overall this looks good (and it sounds like we'll be able to have an answer soon as to which method of transport the process tags will take!)

PR #36052, opened 2025-04-11
[gitlab] remove deprecated `ot-beta` images 

truthbk:
### What does this PR do?

This PR removes `ot-beta` images from the build pipeline as it is now deprecated and no longer required. It also makes sure the `full` image which supersedes the old one is not used everywhere.

### Motivation

`ot-beta` image has been deprecated.

### DIFF
--- a/.gitlab/e2e/e2e.yml
+++ b/.gitlab/e2e/e2e.yml
@@ -750,7 +750,7 @@ new-e2e-otel-eks:
     - !reference [.needs_new_e2e_template]
     - qa_dca
     - qa_agent
-    - qa_agent_ot
+    - qa_agent_full
     - new-e2e-otel-eks-init
   variables:
     TARGETS: ./tests/otel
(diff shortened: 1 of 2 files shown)


.gitlab/e2e/e2e.yml
```diff
@@ -750,7 +750,7 @@ new-e2e-otel-eks:
     - !reference [.needs_new_e2e_template]
     - qa_dca
     - qa_agent
-    - qa_agent_ot
+    - qa_agent_full
```
L3n41c, line 753: This change alone looks suspicious to me.
I didn’t think that the e2e tests framework was able to check the Gitlab jobs dependencies to asses which images need to be tested.
So, with only this change, I would have expected the e2e tests to fail because of missing `-ot-beta` images.
If the e2e tests are not failing, then, it would mean that they’re not using the `-ot-beta`/`-full` images and this line can simply be deleted.
Or am I missing something subtle?

L3n41c (commented):
This change looks suspicious to me because it updates Gitlab jobs dependencies but I cannot see any change making the e2e test job use the `-full` images instead of the `-ot-beta`.
So, if the tests are passing, doesn’t that mean those images are not used and the dependency can simply be removed?

.gitlab/e2e/e2e.yml
truthbk: [This](https://github.com/DataDog/test-infra-definitions/pull/1505) and the bump made [here](https://github.com/DataDog/datadog-agent/pull/36033) was already merged, so the tests should already be using the `full` images.

L3n41c (approved):

CelianR (approved):

PR #36054, opened 2025-04-11
[FA] Add span wrapping every check

coignetp:
### What does this PR do?

Add a wrapping span for every python check.
These span are noop most of the time

### DIFF
--- a/comp/collector/collector/collectorimpl/internal/middleware/check_wrapper.go
+++ b/comp/collector/collector/collectorimpl/internal/middleware/check_wrapper.go
@@ -7,22 +7,26 @@
 package middleware
 
 import (
+	"fmt"
 	"sync"
 	"time"
 
+	agenttelemetry "github.com/DataDog/datadog-agent/comp/core/agenttelemetry/def"
 	"github.com/DataDog/datadog-agent/comp/core/autodiscovery/integration"
 	diagnose "github.com/DataDog/datadog-agent/comp/core/diagnose/def"
 	"github.com/DataDog/datadog-agent/pkg/aggregator/sender"
 	"github.com/DataDog/datadog-agent/pkg/collector/check"
 	checkid "github.com/DataDog/datadog-agent/pkg/collector/check/id"
 	"github.com/DataDog/datadog-agent/pkg/collector/check/stats"
+	"github.com/DataDog/datadog-agent/pkg/util/option"
 )
 
 // CheckWrapper cleans up the check sender after a check was
 // descheduled, taking care that Run is not executing during or after
 // that.
 type CheckWrapper struct {
-	senderManager sender.SenderManager
+	senderManager  sender.SenderManager
+	agentTelemetry option.Option[agenttelemetry.Component]
 
 	inner check.Check
 	// done is true when the check was cancelled and must not run.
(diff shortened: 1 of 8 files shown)


BaptisteFoy (approved):

davidor (approved):
Nothing really to review from container-platform. It's just a small change in 1 file.

rahulkaukuntla (approved):

NouemanKHAL (approved):
LGTM!

GustavoCaso (changes requested):
I request a change to force a conversation, but not because the changes are wrong or anything like that. 

I think that the approach taken is good, but the fact that the Python side of things know includes agenttelemetry and pkg/fleet/installer/telemetry shows that we are leaking information at possible wrong levels of abstraction. Also, that fact that we need to pass the agent telemetry to a lot of places. 

I suggest a different approach. I suggest making the `option.Option[agenttelemetry.Component]` a dependency in the Collector component. The Collector component already wrapps every check with its custom wrapper https://github.com/DataDog/datadog-agent/blob/99a38c9cbee504baf557de63bbf7a49c80dfda69/comp/collector/collector/collectorimpl/collector.go#L233. We could add the logic in the Wrapper Run function. That way:
- We do not have to pass the agentetelemetry all over the codebase 
- We do not include agentetelemetry and fleet related information in the python wrapper 

Looking forward to your thoughts

coignetp:
> I request a change to force a conversation, but not because the changes are wrong or anything like that.
> 
> I think that the approach taken is good, but the fact that the Python side of things know includes agenttelemetry and pkg/fleet/installer/telemetry shows that we are leaking information at possible wrong levels of abstraction. Also, that fact that we need to pass the agent telemetry to a lot of places.
> 
> I suggest a different approach. I suggest making the `option.Option[agenttelemetry.Component]` a dependency in the Collector component. The Collector component already wrapps every check with its custom wrapper
> 
> https://github.com/DataDog/datadog-agent/blob/99a38c9cbee504baf557de63bbf7a49c80dfda69/comp/collector/collector/collectorimpl/collector.go#L233
> 
> . We could add the logic in the Wrapper Run function. That way:
> * We do not have to pass the agentetelemetry all over the codebase
> * We do not include agentetelemetry and fleet related information in the python wrapper
> 
> Looking forward to your thoughts

@GustavoCaso I updated the PR to take your recommendations into account, so we avoid passing the component over the codebase

comp/collector/collector/collectorimpl/internal/middleware/check_wrapper.go
```diff
@@ -32,20 +36,29 @@ type CheckWrapper struct {
...
-		senderManager: senderManager,
+		inner:          inner,
+		senderManager:  senderManager,
+		agentTelemetry: agentTelemetry,
 	}
 }
 
 // Run implements Check#Run
-func (c *CheckWrapper) Run() error {
+func (c *CheckWrapper) Run() (err error) {
```
GustavoCaso, line 48: For my understanding. Why add the `err` part to the return value?

pkg/collector/python/check.go
```diff
@@ -90,7 +90,7 @@ func NewPythonCheck(senderManager sender.SenderManager, name string, class *C.rt
 	return pyCheck, nil
 }
 
-func (c *PythonCheck) runCheckImpl(commitMetrics bool) error {
+func (c *PythonCheck) runCheckImpl(commitMetrics bool) (checkErr error) {
```
GustavoCaso, line 93: For my understanding. Why add the `checkErr ` part to the return value?

comp/collector/collector/collectorimpl/internal/middleware/check_wrapper.go
```diff
@@ -32,20 +36,29 @@ type CheckWrapper struct {
...
 	defer c.runM.Unlock()
 	if c.done {
 		return nil
 	}
+
+	// Start telemetry span if telemetry is enabled
+	if telemetry, isSet := c.agentTelemetry.Get(); isSet {
+		span, _ := telemetry.StartStartupSpan(fmt.Sprintf("check.%s", c.inner.String()))
+		defer span.Finish(err)
+	}
```
GustavoCaso, line 59: In the first implementation, we only added to the Python checks, now we are adding to every check. Is that intended?

GustavoCaso:
@coignetp, thank you so much for taking into account my suggestion. The code looks much better and cleaner. 

I left a few questions, and before approving it. I believe we need to add unit test validating that span is created when agent-telemetry is enabled 😄

comp/collector/collector/collectorimpl/internal/middleware/check_wrapper.go
coignetp: That way we can easily attach the error to the current span

comp/collector/collector/collectorimpl/internal/middleware/check_wrapper.go
coignetp: Yes, we get more observability data there

GustavoCaso (approved):

carlosroman (approved):

arbll:
had to squash to pass the signed commits check

PR #36055, opened 2025-04-11
Avoid restarting the agent too often in step by step tests

KevinFairise2:
### What does this PR do?

The fix is not beautiful but the main constraint is that the agent must not be restarted more than 5 times in 10 seconds, otherwise it breaks the service, according to our systemd file: https://github.com/DataDog/datadog-agent/blob/main/omnibus/config/templates/init-scripts-agent/systemd.service.erb#L16-L17

### DIFF
--- a/test/new-e2e/tests/agent-platform/step-by-step/step_by_step_test.go
+++ b/test/new-e2e/tests/agent-platform/step-by-step/step_by_step_test.go
@@ -12,6 +12,7 @@ import (
 	"strconv"
 	"strings"
 	"testing"
+	"time"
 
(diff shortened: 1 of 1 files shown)


CelianR (approved):
NIT: I think using `EventuallyWithT` would be more appropriate, isn't it?

KevinFairise2:
> NIT: I think using `EventuallyWithT` would be more appropriate, isn't it?

Actually, the agent should restart properly on the first try, but we need to make sure we do not try too often.
Using `EventuallyWithT` could hide some issue where the agent always crash on the first restart and then run properly

PR #36058, opened 2025-04-11
[Closed] Use Core check by default instead of Python check for SNMP integration

Pierre-L42:

### DIFF
--- a/pkg/collector/corechecks/snmp/snmp.go
+++ b/pkg/collector/corechecks/snmp/snmp.go
@@ -218,11 +218,12 @@ func (c *Check) IsHASupported() bool {
 	return true
 }
 
-// SetPythonCheck sets the python check to use as a fallback
+// SetPythonCheck sets the Python check to use as a fallback
 func (c *Check) SetPythonCheck(pythonCheck check.Check) {
 	c.pythonCheck = pythonCheck
 }
 
+// RunPythonCheck runs the Python check if the Core check failed
 func (c *Check) RunPythonCheck(coreCheckErr error) error {
 	if c.pythonCheck == nil {
 		return coreCheckErr
--- a/pkg/collector/corechecks/snmp/snmp.go
+++ b/pkg/collector/corechecks/snmp/snmp.go
@@ -49,6 +49,7 @@ type Check struct {
 	sessionFactory             session.Factory
 	workerRunDeviceCheckErrors *atomic.Uint64
 	agentConfig                config.Component
+	pythonCheck                check.Check
 }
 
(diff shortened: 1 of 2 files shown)


pkg/collector/corechecks/snmp/snmp.go
```diff
@@ -49,19 +49,34 @@ type Check struct {
...
+	pythonCheck                check.Check
 }
 
 // Run executes the check
 func (c *Check) Run() error {
 	var checkErr error
 	sender, err := c.GetSender()
 	if err != nil {
-		return err
+		return c.RunPythonCheck(err)
```
GustavoCaso, line 60: If the sender fails, would it be possible to get it in Python? I recall it uses the same sender manager

pkg/collector/scheduler.go
```diff
@@ -190,6 +191,45 @@ func (s *CheckScheduler) getChecks(config integration.Config) ([]check.Check, er
...
+						log.Errorf("Unable to load a check from instance of config '%s': %s", config.Name, strings.Join(errors, "; "))
+					} else {
+						checks = append(checks, pythonCheck)
+					}
+					continue
+				}
+
+				if pythonErr == nil {
+					coreCheck.(*snmp.Check).SetPythonCheck(pythonCheck)
+				}
```
pgimalac, line 226: I'm not sure that's the best way to do this, but anyway if you're going this way then you should avoid importing the implementation of a check, and just cast to an interface instead, eg.:
```suggestion
				if pythonErr == nil {
					coreCheck.(interface { SetPythonCheck(check.Check) }).SetPythonCheck(pythonCheck)
				}
```
This way you don't import the impl.

(disclaimer I haven't tested this, it might not be quite the correct syntax)

pkg/collector/corechecks/snmp/snmp.go
GustavoCaso: We do  call `GetSender` within the Go <-> Python bridge:

https://github.com/DataDog/datadog-agent/blob/3ba0257d8eadd13d7dcf0b5616751a2aa5042e90/pkg/collector/python/aggregator.go#L31-L68

We call this code from the Python integration via the the `aggregator.submit_metric` 

https://github.com/DataDog/integrations-core/blob/98186cfdf5231a4efab3fbd05cae8f105da5430c/datadog_checks_base/datadog_checks/base/checks/base.py#L768

Pierre-L42:
We are not planning to fallback to Python on the check's run level anymore, but only on the check's init (load) level.

PR #36060, opened 2025-04-11
[NPM-4304] - adding the struct to tracer

AmitaiBl:
### What does this PR do?
moves out from the global struct

### DIFF
--- a/pkg/network/tracer/tracer.go
+++ b/pkg/network/tracer/tracer.go
@@ -64,18 +64,12 @@ const tracerModuleName = "network_tracer"
 //
 // If we want to have a way to track the # of active TCP connections in the future we could use the procfs like here: https://github.com/DataDog/datadog-agent/pull/3728
 // to determine whether a connection is truly closed or not
-var tracerTelemetry = struct {
+type tracerTelemetry struct {
 	skippedConns         telemetry.Counter
 	expiredTCPConns      telemetry.Counter
 	closedConns          *telemetry.StatCounterWrapper
 	connStatsMapSize     telemetry.Gauge
 	payloadSizePerClient telemetry.Gauge
-}{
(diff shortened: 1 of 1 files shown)


hmahmood:
Can we make the title/description a little more ... descriptive?

pkg/network/tracer/tracer.go
```diff
@@ -152,13 +147,20 @@ func newTracer(cfg *config.Config, telemetryComponent telemetryComponent.Compone
 			log.Warnf("NPM is explicitly enabled, so system-probe will continue with only NPM features enabled")
 		}
 	}
-
+	tt := tracerTelemetry{
+		skippedConns:         telemetryComponent.NewCounter(tracerModuleName, "skipped_conns", []string{"ip_proto"}, "Counter measuring skipped connections"),
+		expiredTCPConns:      telemetryComponent.NewCounter(tracerModuleName, "expired_tcp_conns", []string{}, "Counter measuring expired TCP connections"),
+		closedConns:          telemetry.NewStatCounterWrapper(tracerModuleName, "closed_conns", []string{"ip_proto"}, "Counter measuring closed TCP connections"),
```
hmahmood, line 153: This would still use the global telemetry object instead of `telemetryComponent`. I think you can add a parameter to this to pass in `telemetryComponent`, and in `NewStatCounterWrapper` check if it is `nil` or not; if not `nil` use it to create the wrapper counter.

hmahmood:
There are other global telemetry struct's as well under `pkg/network` that need to also not use the global telemetry component.

AmitaiBl:
> There are other global telemetry struct's as well under `pkg/network` that need to also not use the global telemetry component.

is it not better to do them 1 by 1?

charan-manjunath (approved):

PR #36064, opened 2025-04-11
[EBPF-694] gpu: remove inactive streams

gjulianm:
### What does this PR do?

This PR changes the clean logic to include also removal of stream handlers based on inactivity. In order to keep the logic consistent for the aggregator cleanup, it also changes it so that aggregators are cleaned whenever they don't have any stream to aggregate. 

### Motivation

Reduce reporting errors when we miss events that would let us close the stream.

Unit tests included to test the logic.


hush-hush (approved):

pkg/gpu/stats.go
```diff
@@ -54,6 +55,8 @@ func newStatsGeneratorTelemetry(tm telemetry.Component) *statsGeneratorTelemetry
 // TODO: consider removing this parameter and encapsulate it inside the function (will affect UTs as they rely on precise time intervals)
 func (g *statsGenerator) getStats(nowKtime int64) *model.GPUStats {
 	g.currGenerationKTime = nowKtime
+	// reset the active aggregators map
+	g.isAggregatorActive = make(map[model.StatsKey]bool)
```
val06, line 59: can't we add `isActive` field to the aggregator object instead of another map?`statsGenerator` already have a map of aggregators by the same key

val06 (commented):
left some comments

pkg/config/setup/system_probe.go
```diff
@@ -442,7 +442,8 @@ func InitSystemProbeConfig(cfg pkgconfigmodel.Config) {
 	cfg.BindEnvAndSetDefault(join(gpuNS, "configure_cgroup_perms"), false)
 	cfg.BindEnvAndSetDefault(join(gpuNS, "enable_fatbin_parsing"), false)
 	cfg.BindEnvAndSetDefault(join(gpuNS, "fatbin_request_queue_size"), 100)
-	cfg.BindEnvAndSetDefault(join(gpuNS, "ring_buffer_pages_per_device"), 32) // 32 pages = 128KB by default per device
+	cfg.BindEnvAndSetDefault(join(gpuNS, "ring_buffer_pages_per_device"), 32)  // 32 pages = 128KB by default per device
+	cfg.BindEnvAndSetDefault(join(gpuNS, "max_stream_inactivity_seconds"), 30) // 30 seconds by default, includes two checks at the default interval of 15 seconds
```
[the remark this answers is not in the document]
gjulianm: Based on the activity we were seeing this seems to be enough, I haven't seen cases of workloads running for 30 seconds without any CUDA events.

pkg/gpu/stats.go
gjulianm: Changed

val06 (approved):

PR #36065, opened 2025-04-11
feat(gha): Signed commits in go mod tidy

chouetz:
### What does this PR do?
- Move from `stefanzweifel/git-auto-commit-action` to `asana/push-signed-commits` to have signed commits on the mod_tidy generation for `dependabot` PR
- Use `[dependabot skip]` in the commit message ([source](https://github.com/dependabot/dependabot-core/issues/2798#issuecomment-737779277)) to try to fool dependabot, and allow it to auto-close PRs

### Motivation
Enabler to force signed commits in our branch protection rule
https://datadoghq.atlassian.net/browse/ACIX-515

Tested on a personal repo on [this PR](https://github.com/petit-suisse/datadog-agent/pull/63):
- the `@dependabot rebase` works even if we have an additional commit. It flushes the `go_mod_tidy` workflow, but we can force it by re-applying the `dependencies-go` label
- we have a [signed commit](https://github.com/petit-suisse/datadog-agent/pull/23/commits/6002432a5e6388d8ddf3230a5f69f6ff10d660cf)
- A [go-dep PR with an additional commit](https://github.com/petit-suisse/datadog-agent/pull/23) is auto-closed if the dependency is [updated](https://github.com/petit-suisse/datadog-agent/pull/93)

### DIFF
--- a/.github/workflows/go_mod_tidy.yml
+++ b/.github/workflows/go_mod_tidy.yml
@@ -3,6 +3,7 @@ on:
   pull_request:
     types:
       - labeled
+      - synchronize # As the commit is ignored by dependabot we need to recreate it if we rebase.
 
 permissions: {}
 jobs:
--- a/.github/workflows/go_mod_tidy.yml
+++ b/.github/workflows/go_mod_tidy.yml
@@ -4,16 +4,18 @@ on:
     types:
       - labeled
 
+permissions: {}
 jobs:
   mod_tidy_and_generate_licenses:
     if: ${{ github.repository == 'DataDog/datadog-agent' && github.event.pull_request.user.login == 'dependabot[bot]' && contains(github.event.pull_request.labels.*.name, 'dependencies-go') }}
-    runs-on: ubuntu-latest
     permissions:
       contents: write
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           ref: ${{ github.head_ref }}
+          fetch-depth: 0
       - name: Install go
         uses: actions/setup-go@3041bf56c941b39c61721a86cd11f3bb1338122a # v5.2.0
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/go_mod_tidy.yml
```diff
@@ -37,12 +39,15 @@ jobs:
...
-        env:
-          CHANGES: ${{ steps.autocommit.outputs.changes_detected }}
+      - name: Create commit
         run: |
-          echo "Changes detected: $CHANGES"
+          git config --global user.name "Login will be determined by the Github API based on the creator of the token"
+          git config --global user.email ""
+          git commit -am "[dependabot skip] Auto-generate go.sum and LICENSE-3rdparty.csv changes"
+      - name: "Push signed commits"
+        uses: asana/push-signed-commits@d615ca88d8e1a946734c24970d1e7a6c56f34897 # v1
```
KevinFairise2, line 48: How confident are we about using that repo that has 0 stars, and that seems not really active? What effort would it be to have the implementation of the action on our side?

GustavoCaso (approved):
LGTM

.github/workflows/go_mod_tidy.yml
chouetz: This was given as advice [here](https://docs.google.com/document/d/1JXty7M6SocN3fWRqMO7eyohmgdIhzeAQXHB1R36T0hw/edit?usp=sharing). I've reviewed the code and using this precise commit sha is "just" the implementation of [this gist](https://github.com/orgs/community/discussions/50055), so I think it can be trusted.
But as we will probably need to do the same on other places without using predefined actions, we might need to have our own implementation of it.

CelianR (approved):
Lgtm if we are sure that asana/push-signed-commits is reliable

.github/workflows/go_mod_tidy.yml
```diff
@@ -4,16 +4,18 @@ on:
     types:
       - labeled
 
+permissions: {}
```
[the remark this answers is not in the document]
chouetz: Yes because we still grand GITHUB_TOKEN permissions to create PR. When we will remove this, it will become unnecessary.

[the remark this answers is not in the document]
GustavoCaso: Sounds good. Thanks for the clarification

.github/workflows/go_mod_tidy.yml
KevinFairise2: I see, if that's a recommendation that was made by security team, that should be fine.

Agree with your second point too, we might move out of that Action if we need to reimplement the logic as an invoke task, so it ca be used with Gitlab pipelines too

KevinFairise2 (approved):

.github/workflows/go_mod_tidy.yml
avidal: For clarity, it's not a recommendation by the security team. I'm in our internal DX org. The linked doc is a work in progress where I'm trying to find sources of unsigned commits and come up with solutions to address.

I don't have a concrete plan yet, but my thoughts were all leading towards our own implementation both for actions and for gitlab once I have a better view of the landscape (to make sure we can make something that'll work for 90% of the use cases to avoid having to migrate, etc)

avidal: In support of `asana/push-signed-commits`, it doesn't really matter that it has 0 stars or any activity. The implementation is simple and easily readable and Asana is a verified creator (even though this particular action hasn't been published by them on the marketplace).

PR #36066, opened 2025-04-11
restart installer on config changes

jack0x2:
### What does this PR do?
This adds functionality to stop and start the installer so you can follow the same update config logic:
1. update config
2. stop agent
3. start agent (**now will try to start `Datadog Installer` when enabled**) 

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1322

Validated manually and with added e2e test that verify that it starts and stops when it is supposed to be enabled on restarts.


cmd/agent/subcommands/run/dependent_services.go
```diff
@@ -36,3 +41,19 @@ func startDependentServices() {
...
+		if svc.IsEnabled() && svc.ShouldStop() {
+			log.Debugf("Attempting to stop service: %s", svc.name)
+			err := svc.Stop()
+			if err != nil {
+				log.Warnf("Failed to stop services %s: %s", svc.name, err.Error())
+			} else {
+				log.Debugf("Stopped service %s", svc.name)
+			}
+		} else {
+			log.Infof("Service %s is disabled, not stopping", svc.name)
```
dustmop, line 56: This log message will be deceptive in the case that the service is enabled but `ShouldStop` returns false. Maybe instead:

```suggestion
		} else if svc.IsEnabled() {
			log.Infof("Service %s is enabled and doesn't need to be stopped", svc.name)
		} else {
			log.Infof("Service %s is disabled, not stopping", svc.name)
```

cmd/agent/subcommands/run/command.go
```diff
@@ -677,6 +679,9 @@ func stopAgent() {
 	_, cancel := pkgcommon.GetMainCtxCancel()
 	cancel()
 
+	// shutdown depentent services
```
dustmop, line 682:
```suggestion
	// shutdown dependent services
```

dustmop (approved):
Some small suggestions but LGTM overall

cmd/agent/subcommands/run/dependent_services.go
```diff
@@ -36,3 +41,19 @@ func startDependentServices() {
 		}
 	}
 }
+
+func stopDependentServices() {
+	for _, svc := range subservices {
+		if svc.IsEnabled() && svc.ShouldStop() {
```
alexn-dd, line 47: `ShouldStop()` should encapsulate the `IsEnabled()` within it, so that the decision is entirely centralized by a single function.

I have seen in some Go patterns that some places prefer to encapsulate this type of variable decision logic with an interface.  IMO, an interface is not necessary at the moment.

alexn-dd (approved):

sblumenthal (approved):

clarkb7 (approved):

jack0x2:

PR #36069, opened 2025-04-11
[EBPF-694] gpu: normalize memory usage to avoid overreporting

gjulianm:
### What does this PR do?

This PR adds normalization factors for the memory usage, similar to what was done for core usage.  The normalization logic is also moved away from the aggregators into the general stats generator, in that way we avoid having to compute multiple times both core and memory usage.

### Motivation

Improve coherency of the metrics, so that we don't report more memory used than what's actually available.

Unit tests added to validate the normalization, existing tests also ensure that the logic remains correct.


pkg/gpu/probe.go
```diff
@@ -231,7 +231,12 @@ func (p *Probe) GetAndFlush() (*model.GPUStats, error) {
 	if err != nil {
 		return nil, fmt.Errorf("error getting current time: %w", err)
 	}
-	stats := p.statsGenerator.getStats(now)
+
+	stats, err := p.statsGenerator.getStats(now)
+	if err != nil {
+		return nil, fmt.Errorf("error getting stats: %w", err)
```
val06, line 237:
```suggestion
		return nil, err
```

pkg/gpu/stats.go
```diff
@@ -78,25 +80,45 @@ func (g *statsGenerator) getStats(nowKtime int64) *model.GPUStats {
 		}
 	}
 
-	normFactors := g.getNormalizationFactors()
-
-	stats := &model.GPUStats{
-		Metrics: make([]model.StatsTuple, 0, len(g.aggregators)),
-	}
+	// Compute unnormalized stats first for each aggregator
+	unnormalizedStats := make([]model.StatsTuple, 0, len(g.aggregators))
```
val06, line 84: let's call it `rawStats`
(you can keep the comment mentioning those stats are not normalized yet)

pkg/gpu/stats.go
```diff
@@ -78,25 +80,45 @@ func (g *statsGenerator) getStats(nowKtime int64) *model.GPUStats {
...
+
+	// Apply normalization to each stats entry
+	stats := &model.GPUStats{
+		Metrics: make([]model.StatsTuple, 0, len(unnormalizedStats)),
+	}
+	for _, entry := range unnormalizedStats {
+		factors, ok := normFactors[entry.Key.DeviceUUID]
+		if !ok {
+			return nil, fmt.Errorf("cannot find normalization factors for device %s", entry.Key.DeviceUUID)
 		}
```
val06, line 108: what does this error mean? are we throwing away all stats if we don't have a normalizationFactor for a device?
(i don't remember all the relations, are all those stats belong to one device or rather those are global raw stats slice?)

val06 (commented):
left comments

pkg/gpu/probe.go
gjulianm: I'd rather keep this as-is, I feel passing the errors without wrapping makes things harder to debug

pkg/gpu/stats.go
gjulianm: It's more of a safeguard as this case should never happen, we're computing the normalizationFactors map based on the entries in the `rawStats` struct. If this happens it's a programming error.

pkg/gpu/probe.go
val06: but this message doesn't add any extra info on the error

gjulianm: It specifies where it came from, in this case from the `getStats` call. Makes it easier to locate, specially if for some reason we forget wrapping some error down below in the stack or if we add more functions to the `GetAndFlush` function.

val06 (approved):

PR #36075, opened 2025-04-11
feat(gha): Use signed commits for otel collector dependencies

chouetz:
### What does this PR do?
Use signed commits for otel collector dependecies workflow:
- Save the new OCB_VERSION as env variable and push it in GITHUB_OUTPUT
- Reuse this variable in the `peter-evans/create-pull-request` action
- Remove the no more used python code

### Motivation
https://datadoghq.atlassian.net/browse/ACIX-515

Ran the workflow on a personal fork TBD

### DIFF
--- a/tasks/collector.py
+++ b/tasks/collector.py
@@ -14,16 +14,6 @@ from invoke.tasks import task
 from tasks.go import tidy
 from tasks.libs.ciproviders.github_api import GithubAPI
 from tasks.libs.common.color import Color, color_message
-from tasks.libs.common.constants import (
-    GITHUB_REPO_NAME,
-)
-from tasks.libs.common.git import (
-    check_clean_branch_state,
-    check_uncommitted_changes,
-    get_git_config,
-    revert_git_config,
-    set_git_config,
-)
 
 LICENSE_HEADER = """// Unless explicitly stated otherwise all files in this repository are licensed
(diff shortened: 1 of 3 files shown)


tasks/collector.py
```diff
@@ -524,45 +515,3 @@ def update(_):
     updater = CollectorVersionUpdater()
     updater.update()
     print("Update complete.")
-
-
-@task()
```
KevinFairise2, line 529: Is that task used locally? The fact that we skip pre-commit hooks make me feel that the task was used locally

CelianR (approved):

songy23 (approved):
cc @jackgopack4

tasks/collector.py
songy23: It is supposed to be used both locally and in automations

truthbk: I don't think we need to keep it around if we're going to do the deps update via automation. Can you confirm removing this is OK @songy23?

truthbk (approved):

tasks/collector.py
songy23: I think when automations fail we need to able to run it locally too. Without this inv task it is painful to do the version upgrade manually

chouetz: I've put back the task and changed it to restrict its usage for a local run.

PR #36099, opened 2025-04-14
extract memory monitor into separate package

paulcacheux:
### What does this PR do?

The memory controller imports `github.com/containerd/cgroups/v3/cgroup1` that ends up importing some dbus dependency. But the controller is only used by system-probe while `pkg/util/cgroup` is more broadly used. This PR extracts the memory controller to its own package allowing non-system-probe agents to not import a few dependencies.


brycekahle (approved):
I know this is behind a config flag. Do we actually enable it anywhere?

sblumenthal (approved):

PR #36114, opened 2025-04-14
system-probe: Use longer timeouts for check clients

vitkyrka:
### What does this PR do?

Currently, the timeout in the system-probe API is only 5 seconds.  While
this makes sense for clients which need to access system-probe as part
of some other work (e.g. for flare generation), it may be too small for
core agent checks which rely on system-probe to do most of the work.
Spuriously timing out early in this case is not helpful since the check
does not perform its main function, and system-probe actually continues
to compute whatever it had to compute in the background (since the http
handler will run to completion), except that the result is not returned
to whoever asked for it.

Increase the timeout to handle this.  During startup, to prevent waiting
for unnecessarily long time if system-probe has issues starting up, use
the shorter timeout since the stats endpoint should return quickly.

### Motivation

https://datadoghq.atlassian.net/browse/DSCVR-94

The existing code being touched is covered by automated tests.

The increased timeouts were [tested](https://ddstaging.datadoghq.com/dashboard/cc9-qxb-fsx/agent-discovery?fromUser=true&refresh_mode=paused&tpl_var_kube_cluster_name%5B0%5D=oddish-c&from_ts=1744624944571&to_ts=1744645238000&live=false) on one cluster in staging.

### DIFF
--- a/pkg/config/setup/config.go
+++ b/pkg/config/setup/config.go
@@ -1126,6 +1126,7 @@ func agent(config pkgconfigmodel.Setup) {
 	config.BindEnvAndSetDefault("check_runners", int64(4))
 	config.BindEnvAndSetDefault("check_cancel_timeout", 500*time.Millisecond)
 	config.BindEnvAndSetDefault("check_system_probe_startup_time", 5*time.Minute)
+	config.BindEnvAndSetDefault("check_system_probe_timeout", 60*time.Second)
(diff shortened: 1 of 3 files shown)


dustmop (approved):

brycekahle (approved):

smeso-ddg (approved):

PR #36117, opened 2025-04-14
[netpath] SACK: Add socket filter for CPU perf

pimlu:
### What does this PR do?
This is a follow up to [\[NPM-4279\] Add socket filter to linux TCP traceroutes](https://github.com/DataDog/datadog-agent/pull/35651), applying the same thing to SACK.

### Motivation
This improves CPU usage of TCP traceroutes because system-probe captures vastly less packets.

New tests should pass:
```
sudo go test -tags test,linux,linux_bpf github.com/DataDog/datadog-agent/pkg/networkpath/traceroute/filter -count 10
```
SACK traceroute should still work:
```
go build -race github.com/DataDog/datadog-agent/pkg/networkpath/traceroute/sack/portable_sack
LOG_LEVEL=trace sudo -E ./portable_sack 205.251.242.103:443
```

### DIFF
--- a/pkg/networkpath/traceroute/filter/tcp.go
+++ b/pkg/networkpath/traceroute/filter/tcp.go
@@ -72,3 +72,35 @@ func (c TCP4FilterConfig) GenerateTCP4Filter() ([]bpf.RawInstruction, error) {
 		bpf.RetConstant{Val: 0},
 	})
 }
+
+// GenerateSynack4Filter creates a classic BPF filter for TCP SOCK_RAW sockets.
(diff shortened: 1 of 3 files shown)


AlexandreYang (approved):
LGTM once tested on staging with static and dynamic paths (if not already done)

ken-schneider (approved):

pimlu:
I tested this on staging with static path, but not dynamic. I don't think SACK is compatible with dynamic paths until we figure out how to filter out agent traceroutes to avoid an infinite loop

PR #36122, opened 2025-04-14
[OTAGENT-296] Fix otel grpc max recv msg size config

songy23:
### What does this PR do?
1. The OTLP receiver gRPC server in the trace agent now respects the config `otlp_config.receiver.protocols.grpc.max_recv_msg_size_mib` or env var `DD_OTLP_CONFIG_RECEIVER_PROTOCOLS_GRPC_MAX_RECV_MSG_SIZE_MIB`.
2. Fixes a bug that Agent OTLP ingestion fails to start when the config `otlp_config.receiver.protocols.grpc.max_recv_msg_size_mib` or env var `DD_OTLP_CONFIG_RECEIVER_PROTOCOLS_GRPC_MAX_RECV_MSG_SIZE_MIB` is set to a string.

### Motivation
1. Make the gRPC message size limit consistent in OTLP trace ingestion between core agent and trace agent
2. Bug fix for the `DD_OTLP_CONFIG_RECEIVER_PROTOCOLS_GRPC_MAX_RECV_MSG_SIZE_MIB` env var

Added & modified tests, also tested E2E with the CI docker image

For manual QA:
1. Set `DD_OTLP_CONFIG_RECEIVER_PROTOCOLS_GRPC_MAX_RECV_MSG_SIZE_MIB` to a string, e.g. `export DD_OTLP_CONFIG_RECEIVER_PROTOCOLS_GRPC_MAX_RECV_MSG_SIZE_MIB="20"` and verify OTLP ingestion still starts with no error
2. Adjust `DD_OTLP_CONFIG_RECEIVER_PROTOCOLS_GRPC_MAX_RECV_MSG_SIZE_MIB` and send traces whose byte size is below the limit, verify traces go through; then send traces with sizes too large, verify traces are dropped


joepeeples (approved):

comp/otelcol/otlp/configcheck/configcheck.go
```diff
@@ -52,12 +52,27 @@ func ReadConfigSection(cfg config.Reader, section string) *confmap.Conf {
...
+				val = deepcopy.Copy(cfg.Get(key))
+			}
+			stringMap[mapKey] = val
 		}
 	}
 	return confmap.NewFromStringMap(stringMap)
 }
 
+// intConfigs has the known config keys that may need a type cast to int by calling cfg.GetInt
+var intConfigs = map[string]struct{}{
```
liustanley, line 68: We might want to double check with Agent Configuration, but I believe this type cast to int can be done with `config.BindEnvAndSetDefault` if we set the defaults to ints. For example, I added the following change in pkg/config/setup/otlp.go and didn't see the int conversion error:
```
config.BindEnvAndSetDefault("otlp_config.receiver.protocols.grpc.max_recv_msg_size_mib", 10)
```

comp/otelcol/otlp/configcheck/configcheck.go
```diff
@@ -52,12 +52,27 @@ func ReadConfigSection(cfg config.Reader, section string) *confmap.Conf {
 		if strings.HasPrefix(key, prefix) && cfg.IsSet(key) {
 			mapKey := strings.ReplaceAll(key[len(prefix):], ".", confmap.KeyDelimiter)
 			// deep copy since `cfg.Get` returns a reference
-			stringMap[mapKey] = deepcopy.Copy(cfg.Get(key))
+			var val interface{}
+			if _, ok := intConfigs[key]; ok {
+				val = deepcopy.Copy(cfg.GetInt(key)) // ensure to get an int even if it is set as a string in env vars
+			} else {
+				val = deepcopy.Copy(cfg.Get(key))
```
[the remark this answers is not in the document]
songy23: This grpc max recv msg size needs to be an int, otherwise confmap unmarshal will panic. Config reader cannot guarantee to return an int with `cfg.Get` (it may return a string) vs. always returns an int with `cfg.GetInt`

comp/otelcol/otlp/configcheck/configcheck.go
songy23: This seems to have some strange side effects. Once I do this, the other otel gRPC & HTTP configs break. Will take a closer look tmr

ichinaski (approved):

comp/otelcol/otlp/configcheck/configcheck.go
truthbk: I'll add @hush-hush to the list of reviewers as he might have suggestions.

truthbk (approved):
Looks good to me. I've added a reviewer from @DataDog/agent-configuration just to make sure we're aligned with current best principles.

comp/otelcol/otlp/configcheck/configcheck.go
hush-hush: @liustanley is correct. When using `Get` the config will try, as best as possible, to cast to the default type. The implementation from [viper](https://github.com/DataDog/viper/blob/master/viper.go#L651) only support basic types (int, float, time, slice of string, ..).

hush-hush (changes requested):

comp/otelcol/otlp/configcheck/configcheck.go
songy23: > Adding a default to all those setting

What if we don't have defaults for some of those configs? https://github.com/open-telemetry/opentelemetry-collector/blob/main/receiver/otlpreceiver/config.md It seems OTel requires some of them to not have a default

songy23: @hush-hush unfortunately we cannot do this because OTLP ingestion relies on a check on whether the `otlp_config.receiver` section is present to determine whether to enable the OTLP ingestion pipeline http://github.com/DataDog/datadog-agent/blob/5c3c17908079421d4b663c6dea60b2965efe623f/comp/otelcol/otlp/configcheck/configcheck.go#L72

As soon as default values are added to any config under `otlp_config.receiver`, it will force the OTLP ingestion pipeline to be always enabled, that will be a big breaking change

hush-hush: I agree this could be taken care in another PR. Can we create a card to track this effort ? We're starting to accumulate a lot of workaround around the OTLP initalization logic.

hush-hush (approved):
Added a last comment, feel free to merge after.

pkg/config/setup/otlp.go
```diff
@@ -43,22 +43,24 @@ func OTLP(config pkgconfigmodel.Setup) {
 		"DD_OTLP_CONFIG_TRACES_PROBABILISTIC_SAMPLER_SAMPLING_PERCENTAGE")
 	config.BindEnvAndSetDefault("otlp_config.traces.internal_port", 5003)
 
+	// Do NOT add default values for any config under otlp_config.receiver, that will force the OTLP ingestion pipelines to always start
```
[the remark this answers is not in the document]
songy23: Yes, I left a todo with tix for the future fix

PR #36123, opened 2025-04-14
build custom agent from 7.64.x with 7.65.x integrations-core

lu-zhengda:
### What does this PR do?

incident-37187

Build a custom agent based on 7.64.x with integrations-core from 7.65.0-rc.8.


paulcacheux:
Is this still needed ?

PR #36132, opened 2025-04-14
Change structure of remote-tagger timeout to cancel when receiving Done signal

grantseltzer:
### What does this PR do?

This PR changes the structure of closing the remote tagger to recognize the write to the done channel immediately by using the backoff via timer instead of using the imported backoff library's structure.

### Motivation

After #35818 merged I started observing that the system-probe was very slow to shutdown during local development when I interrupted the process. I've concluded that the new wait group blocks the process from exiting until the remote tagger properly closes. While that PR  makes sense, it exposes a very slow implementation of exponential backoff where it can take up to 5 minutes for the process to close even after the tagger has been closed.

When locally developing the dynamicinstrumentation system-probe module the remote tagger was failing to initialize (as expected as i'm not connecting to platform). Hitting control-c would show a delay in killing the process that only gets longer as the system-probe runs for longer. After this change it does not occur. 

### Additional Notes

@misteriaud please verify that this doesn't affect the bug/race you fixed in your PR.

### DIFF
--- a/comp/core/tagger/impl-remote/remote.go
+++ b/comp/core/tagger/impl-remote/remote.go
@@ -568,7 +568,7 @@ func (t *remoteTagger) startTaggerStream(maxElapsed time.Duration) error {
 	for {
 		select {
 		case <-t.ctx.Done():
-			return &backoff.PermanentError{Err: errTaggerStreamNotStarted}
+			return errTaggerStreamNotStarted
 		case <-timer.C:
 			// Check the auth token
 			if t.token == "" {
--- a/comp/core/tagger/impl-remote/remote.go
+++ b/comp/core/tagger/impl-remote/remote.go
@@ -561,48 +561,55 @@ func (t *remoteTagger) startTaggerStream(maxElapsed time.Duration) error {
 	expBackoff.MaxInterval = 5 * time.Minute
 	expBackoff.MaxElapsedTime = maxElapsed
 
-	return backoff.Retry(func() error {
+	var err error
+	timer := time.NewTimer(0) // immediate first attempt
+	defer timer.Stop()
+
+	for {
 		select {
 		case <-t.ctx.Done():
 			return &backoff.PermanentError{Err: errTaggerStreamNotStarted}
-		default:
+		case <-timer.C:
+			// Check the auth token
+			if t.token == "" {
+				return errors.New("RemoteTagger initialization failed: auth token is unset")
+			}
+
+			// Cancel any existing stream context before creating a new one
+			if t.streamCancel != nil {
+				t.streamCancel()
+			}
+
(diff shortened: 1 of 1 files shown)


comp/core/tagger/impl-remote/remote.go
```diff
@@ -561,48 +561,55 @@ func (t *remoteTagger) startTaggerStream(maxElapsed time.Duration) error {
...
 
-	return backoff.Retry(func() error {
+	var err error
+	timer := time.NewTimer(0) // immediate first attempt
+	defer timer.Stop()
+
+	for {
 		select {
 		case <-t.ctx.Done():
 			return &backoff.PermanentError{Err: errTaggerStreamNotStarted}
```
misteriaud, line 571: **🥜 nitpick:** Since this part of the code is no longer in the `backoff.Retry`‏ callback function, we can directly return the root error.
```suggestion
			return errTaggerStreamNotStarted
```

misteriaud (approved):
Nice job, indeed it's a better design !
I checked in a VM, still working as expected with the coreAgent (not showing TLS error).
I just left a nit comment.

comp/core/tagger/impl-remote/remote.go
```diff
@@ -561,48 +561,55 @@ func (t *remoteTagger) startTaggerStream(maxElapsed time.Duration) error {
...
-			return &backoff.PermanentError{Err: errTaggerStreamNotStarted}
-		default:
-		}
-
-		var err error
+			return errTaggerStreamNotStarted
+		case <-timer.C:
+			// Check the auth token
+			if t.token == "" {
+				return errors.New("RemoteTagger initialization failed: auth token is unset")
```
gabedos, line 575: Previously, the `startTaggerStream` method would still retry if there were any auth token issues. It can be dynamically updated [here](https://github.com/DataDog/datadog-agent/blob/main/comp/core/tagger/impl-remote/remote.go#L215). Can you update your implementation to still allow retries if there is an auth token error?

Now:
```
	for {
		select {
		case <-t.ctx.Done():
			return errTaggerStreamNotStarted
		case <-timer.C:
			// Check the auth token
			if t.token == "" {
				return errors.New("RemoteTagger initialization failed: auth token is unset")
			}
```

Previously:
```
return backoff.Retry(func() error {
		select {
		case <-t.ctx.Done():
			return &backoff.PermanentError{Err: errTaggerStreamNotStarted}
		default:
		}

		var err error

		// Check the auth token
		if t.token == "" {
			return errors.New("RemoteTagger initialization failed: auth token is unset")
		}
```

grantseltzer: Done! Good catch.

gabedos (approved):
Changes LGTM!

PR #36135, opened 2025-04-14
[CONTP-539] Agent internal telemetry command

gabedos:
### What does this PR do?

Implements a  command`agent-full-telemetry` as a subcommand of `diagnose`. This prints all the telemetry on the agent exposed on the telemetry endpoint (typically `"http://#{cmd_host}:#{expvar_port}/telemetry)`

### Motivation

Improve tooling on agent telemetry for easier visibility into agent. Use cases for seeing telemetry that the Agent exposes but isn't necessarily sent over to Datadog. Useful for debugging the Agent.

Deploy agent and run `agent diagnose show-metadata agent-full-telemetry`. You should expect a long list of agent telemetry metrics.

N/A

### Additional Notes

N/A


GustavoCaso (changes requested):
We currently have the `diagnose` command to expose internal information. I don't think creating a new subcommand is the way to go. Could we use the diagnose command? Also, once we add new commands is pretty hard to deprecated them, could you explain the use cases this subcommands helps to solve? Mostly to understand if there are already exiting tool that we can leverage to solve the problem 

@hush-hush @iglendd @sgnn7

iglendd:
@GustavoCaso and @gabedos I already had similar discussion with Gustavo ... when I have trying to do something like this. And that being said that I was trying to "dump" Agent internal telemetry ... and felt very strongly for adding a new command since it was not of "diagnose" category. But after zoom discussion with Gustavo and contemplation I have decided against it.

While I do agree that the term "diagnose" itself semantics "normally" have a quite a narrow and specific meaning, it does not actually applied to the existing agent diagnose command, which had been used traditionally by developer to show internal telemetry via `show-metadata` sub-command. Here is the list

```
agent.exe diagnose show-metadata -h
Print metadata payloads sent by the agent

Usage:
  agent.exe diagnose show-metadata [flags]
  agent.exe diagnose show-metadata [command]

Available Commands:
  agent-telemetry  [internal] Print agent telemetry payloads sent by the agent.
  gohai            [internal] Print the gohai payload for the agent.
  ha-agent         Print the HA Agent Metadata payload.
  host-gpu         [internal] Print the host-gpu agent metadata payload.
  inventory-agent  [internal] Print the Inventory agent metadata payload.
  inventory-checks [internal] Print the Inventory checks metadata payload.
  inventory-host   [internal] Print the Inventory host metadata payload.
  inventory-otel   Print the Inventory otel metadata payload.
  package-signing  [internal] Print the Inventory package signing payload.
  security-agent   [internal] Print the security-agent process metadata payload.
  system-probe     [internal] Print the inventory systemprobe metadata payload.
  v5               [internal] Print the metadata payload for the agent.
```
It is probably not perfect to be very precise semantic-wise but it is used for diagnostic and it is what devs, support and many customers maybe using and its change perhaps would not be simple, even though we may discuss it.

What is being missing IMHO is its documentation although it is not clear it is value for non-Agent dev users

iglendd:
we can further discuss it in a meeting

gabedos:
> To be clear, just in case, in simple term, I am in strong favor of adding Agent's new internal information display in CLI using `agent diagnose show-metadata` commands

Hey @iglendd, thanks for your opinion. I honestly had no idea that the `show-metadata` subcommand had further subcommands within it. I agree it could be added in there.

As you pointed out there's already a subcommand `agent-telemetry  [internal] Print agent telemetry payloads sent by the agent`. However, I'm more interested in telemetry **not** sent by the agent. I could add another subcommand like `agent-full-telemetry [internal] Print all agent telemetry.`

iglendd:
sounds good to me.

cmd/agent/subcommands/diagnose/command.go
```diff
@@ -466,3 +482,13 @@ func requestDiagnosesFromAgentProcess(diagCfg diagnose.Config) (*diagnose.Result
...
+
+// queryAgentTelemetry gets the full telemetry payload exposed by the agent
+func queryAgentTelemetry() ([]byte, error) {
+	r, err := http.Get(fmt.Sprintf("http://localhost:%s/telemetry", pkgconfigsetup.Datadog().GetString("expvar_port")))
+	if err != nil {
+		return nil, err
+	}
+	defer r.Body.Close()
+	return io.ReadAll(r.Body)
+}
```
GustavoCaso, line 494: I would change this function signature to take Fx components and use `fxutil.OneShot` inthe cobra command.

```go
// printAgentTelemetry gets the full telemetry payload exposed by the agent
func printAgentTelemetry(config config.Component) error {
	c := util.GetClient()
	ipcAddress, err := pkgconfigsetup.GetIPCAddress(config)
	if err != nil {
		return err
	}

	r, err := c.Get(fmt.Sprintf("http://%s:%s/telemetry", ipcAddress, config.GetString("expvar_port")))
	if err != nil {
		return err
	}
	defer r.Body.Close()
	result, err := io.ReadAll(r.Body)
	if err != nil {
		return err
	}
	fmt.Println(string(result))
	return nil
}
```

Noticeable changes. 
- Use  our HTTP client `c := util.GetClient()`
- Handle the case the customer has modify `cmd_host` and we do not hard code `localhost` in the URL `ipcAddress, err := pkgconfigsetup.GetIPCAddress(config)`

cmd/agent/subcommands/diagnose/command.go
```diff
@@ -321,6 +323,19 @@ This command print the security-agent metadata payload. This payload is used by
...
+	agentFullTelemetryCmd := &cobra.Command{
+		Use:   "agent-full-telemetry",
+		Short: "[internal] Print the full agent telemetry metrics exposed by the agent",
+		RunE: func(_ *cobra.Command, _ []string) error {
+			payload, err := queryAgentTelemetry()
+			if err != nil {
+				return err
+			}
+			fmt.Print(string(payload))
+			return nil
```
GustavoCaso, line 335: With the other comment I would change it to use `fxutil.OneShot`

```go
return fxutil.OneShot(printAgentTelemetry,
    fx.Supply(command.GetDefaultCoreBundleParams(cliParams.GlobalParams)),
    core.Bundle(),
) 
```

cmd/agent/subcommands/diagnose/command.go
gabedos: Thank you! Good catch on the `cmd_host` for the query.

cmd/agent/subcommands/diagnose/command.go
```diff
@@ -321,6 +322,19 @@ This command print the security-agent metadata payload. This payload is used by
 		},
 	}
 
+	agentFullTelemetryCmd := &cobra.Command{
+		Use:   "agent-full-telemetry",
+		Short: "[internal] Print the full agent telemetry metrics exposed by the agent",
+		Long:  `.`,
+		RunE: func(_ *cobra.Command, _ []string) error {
+			return fxutil.OneShot(printAgentFullTelemetry,
+				fx.Supply(payloadName("agent-full-telemetry")),
```
GustavoCaso, line 331: No need to supply this as it is not used 😄

GustavoCaso (approved):
LGTM

hush-hush (approved):

PR #30086, opened 2024-10-14
[BARX-645] tasks: remove --python-runtimes parameters

chouquette:
### What does this PR do?

Remove `--python-runtimes` invoke tasks parameters.

### Motivation

Dropping python2 support. There parameters are now never passed by the CI, and never read by the build system

PR #36143, opened 2025-04-15
[BARX-661] remove --use-release-entries and release_version from everywhere 

sabrina-datadog:
### What does this PR do?
- unifies the release and nightly entries in the release.json to one (renamed it to dependencies)
- removes `RELEASE_VERSION` environment variable from all the build related jobs
- removes `--use-release-entries` flag from everywhere

*note:
keeping the `release_version` parameter in the `omnibus.build` function for now because it's still used in the `datadog-agent-macos-build`repo ([here](https://github.com/DataDog/datadog-agent-macos-build/blob/eea8feab18a8f8e13e02bd1f959d02e59cf7902a/scripts/build_script.sh#L73)). Although it's no longer needed, we're preserving it temporarily to maintain compatibility/no downtime between the two repositories (they rely on each other). 

The plan is to:
1. first merge this PR with the parameter left in place
2. [open a PR](https://github.com/DataDog/datadog-agent-macos-build/pull/256) in the `macos-build` repo to remove release_version there
3. once that change is merged, we'll follow up with a third PR to remove the `release_version` parameter and the print statement from this file

### Motivation
since we now only support Agent 7 and always build release versions from the release branch, we no longer need to have two separate entries for nightly and release. main branch will always have the "nightly" version and the release branch will always have the "release" version.

made sure CI passes --> [successful pipeline example](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/63134324)


tasks/github_tasks.py
```diff
@@ -90,16 +89,11 @@ def trigger_macos(
...
         conclusion = _trigger_macos_workflow(
-            # Provide the release version to be able to fetch the associated
-            # macos-build branch from release.json for all workflows...
-            release_version,
             destination,
             retry_download,
             retry_interval,
             workflow_name="macos.yaml",
             datadog_agent_ref=datadog_agent_ref,
-            # ... And provide the release version as a workflow input when needed
```
sabrina-datadog, line 101: ended up removing the `release_version` because in the `datadog-agent-macos-build` repo there is a `DEFAULT_RELEASE_VERSION` env variable ([here](https://github.com/DataDog/datadog-agent-macos-build/blob/eea8feab18a8f8e13e02bd1f959d02e59cf7902a/.github/workflows/macos.yaml#L9)), so we don’t need to pass it explicitly from the datadog-agent side

tasks/omnibus.py
```diff
@@ -199,6 +198,9 @@ def build(
     Build the Agent packages with Omnibus Installer.
     """
 
+    if release_version:
```
sabrina-datadog, line 201: keeping the `release_version` parameter in this function for now because it's still used in the `datadog-agent-macos-build`repo ([here](https://github.com/DataDog/datadog-agent-macos-build/blob/eea8feab18a8f8e13e02bd1f959d02e59cf7902a/scripts/build_script.sh#L73)). Although it's no longer needed in the `datadog-agent` omnibus.py file, we're preserving it temporarily to maintain compatibility/no downtime between the two repositories. 

The plan is to:
1. first merge this PR with the parameter left in place
2. open a PR in the `macos-build` repo to remove release_version there
3. once that change is merged, we'll follow up with a third PR to remove the `release_version` parameter and the print statement from this file

tasks/libs/releasing/json.py
```diff
@@ -117,17 +118,12 @@ def _get_windows_driver_info(release_json_version_data, driver_name):
     return driver_value, version_value, shasum_value
 
 
-def _get_release_json_info_for_next_rc(release_json, is_first_rc=False):
+def _get_release_json_info_for_next_rc(release_json):
```
sabrina-datadog, line 121: not sure if I changed this correctly. I removed the `is_first_rc` parameter because the initial release.json for the `7.XX.x` branch would be branched off from the "nightly" release.json from main. the first rc would be using the "nightly" to modify the release.json fields. from then on, it would then be the "release" fields.

.gitlab/notify/notify.yml
```diff
@@ -54,39 +54,6 @@ send_pipeline_stats:
...
-    - python3 -m dda self dep sync -f legacy-tasks
-    - messagefile="$(mktemp)"
-    - echo "Use this command from [test-infra-definitions](https://github.com/DataDog/test-infra-definitions) to manually test this PR changes on a VM:" >> "$messagefile"
-    - echo '```sh' >> "$messagefile"
-    - echo "dda inv aws.create-vm --pipeline-id=$CI_PIPELINE_ID --os-family=ubuntu" >> "$messagefile"
-    - echo '```' >> "$messagefile"
-    - 'echo "Note: This applies to commit **$CI_COMMIT_SHORT_SHA**" >> "$messagefile"'
-    - dda inv -- -e github.pr-commenter --title "Test changes on VM" --body "$(cat "$messagefile")" --echo
-    - rm "$messagefile"
-
```
spencergilbert, line 89: **❓ question:** ‏why's this fully removed? Is it not used at all anymore?

release.json
```diff
@@ -5,7 +5,7 @@
         "6": "6.53.1",
         "7": "7.64.3"
     },
-    "nightly": {
+    "dependencies":{
```
spencergilbert, line 8: **🥜 nitpick:** ‏ alphabetizing would be nice

spencergilbert: Also I'm not sure how to check, but it would be good to confirm we're keeping the right values here.

tasks/libs/releasing/json.py
spencergilbert: Logically it seems correct, we don't differentiate anymore so it seems like dropped it is right - but I'm not too familiar with the functional usage.

tasks/omnibus.py
```diff
@@ -199,6 +198,9 @@ def build(
...
+        print("The release_version argument is deprecated and will be removed soon.")
```
spencergilbert, line 202: **🥜 nitpick:** ‏ there's _probably_ a "log" function to use here but given the short term nature of the change it's not a big deal.

.gitlab/notify/notify.yml
sabrina-datadog: I talked to @chouetz and he said this job can be removed since the generated message ([example](https://github.com/DataDog/datadog-agent/pull/36064#issuecomment-2797203351)) isn't used by anyone. originally the message was to encourage qa before the release

clarkb7 (approved):

release.json
```diff
@@ -19,48 +15,8 @@
...
-        "SECURITY_AGENT_POLICIES_VERSION": "v0.66.0",
-        "MACOS_BUILD_VERSION": "7.56.0-rc.3",
-        "WINDOWS_DDNPM_DRIVER": "release-signed",
-        "WINDOWS_DDNPM_VERSION": "2.7.1",
-        "WINDOWS_DDNPM_SHASUM": "0f4665761324e1fef1c21651be5b70e79c72b5e7e5662d74619e7db2b27d5bc2",
-        "WINDOWS_DDPROCMON_DRIVER": "release-signed",
-        "WINDOWS_DDPROCMON_VERSION": "1.2.0",
-        "WINDOWS_DDPROCMON_SHASUM": "e5525959521254ba8218f64f644ad19cdeb720aa48ee83eb5e402a8d1643e32e"
-    },
-    "dca-1.17.0": {
```
sabrina-datadog, line 36: talked to the agent-security team and they confirmed that they don't use these values anymore

spencergilbert (approved):

Dockerfiles/agent/windows/README.md
```diff
@@ -6,7 +6,7 @@ How to build the Agent docker image
 From the root of the repository, run the following command:
 
 ```
-docker run --rm -it -v "${pwd}:c:\mnt" -e OMNIBUS_TARGET=main -e MAJOR_VERSION=7 -e RELEASE_VERSION=nightly -e PY_RUNTIMES=3 datadog/agent-buildimages-windows_x64:1809 powershell -C "c:\mnt\tasks\winbuildscripts\Build-AgentPackages.ps1 -BuildOutOfSource 1 -InstallDeps 1 -CheckGoVersion 1"
+docker run --rm -it -v "${pwd}:c:\mnt" -e OMNIBUS_TARGET=main -e MAJOR_VERSION=7 -e PY_RUNTIMES=3 datadog/agent-buildimages-windows_x64:1809 powershell -C "c:\mnt\tasks\winbuildscripts\Build-AgentPackages.ps1 -BuildOutOfSource 1 -InstallDeps 1 -CheckGoVersion 1"
```
chouetz, line 9:
```suggestion
docker run --rm -it -v "${pwd}:c:\mnt" -e OMNIBUS_TARGET=main -e MAJOR_VERSION=7 datadog/agent-buildimages-windows_x64:1809 powershell -C "c:\mnt\tasks\winbuildscripts\Build-AgentPackages.ps1 -BuildOutOfSource 1 -InstallDeps 1 -CheckGoVersion 1"
```
I think this is not used anymore, cf https://github.com/DataDog/datadog-agent/pull/30086

tasks/libs/releasing/json.py
chouetz: Regarding the content of this method we can probably get rid of the method as well, as `release_json_version_data = release_json[RELEASE_JSON_DEPENDENCIES]` is probably easier to read than `release_json_version_data = _get_release_json_info_for_next_rc(release_json)`

tasks/libs/releasing/version.py
```diff
@@ -239,40 +241,28 @@ def _get_highest_repo_version(
...
-            if match:
-                release_component_version = _create_version_from_match(match)
-            else:
-                print(
-                    f"{release_entry_name} does not have a valid {release_json_key} ({release_json_entry.get(release_json_key, '')}), ignoring"
-                )
-
-    if not release_version:
-        raise Exit(f"Couldn't find any matching {release_version} version.", 1)
+    if not dependencies_entry:
```
chouetz, line 252:
```suggestion
    if dependencies_entry is None:
```
As you explicitly returns `None` from the `dict.get` method when this key is not found you should test if it is `None`

tasks/libs/releasing/version.py
```diff
@@ -239,40 +241,28 @@ def _get_highest_repo_version(
...
+        raise Exit(f"release.json is missing a {RELEASE_JSON_DEPENDENCIES} entry.", 1)
 
+    # Check that the component's version is defined in the dependencies entry
     if release_json_key is not None:
```
chouetz, line 256: it seems this argument is always passed when you call that function, and you remove the default value to `None` so I'm not sure this test is relevant anymore.

chouetz (approved):
Minor comments, LGTM.
I think you should backport this PR to agent6 branch

tasks/libs/releasing/json.py
alopezz: +1 on removing this function (inlining it) if the name is not meaningful in the current release process.

PR #36156, opened 2025-04-15
[SVLS] Improve serverless-init log tailing in sidecar mode

jcstorms1:
### What does this PR do?
Our current implementation did not take into account the fact that AAS shares a log volume with all of the instances running serverless-init. This resulted in sending duplicate logs from every running sidecar. This update adds an option to enable logging based on the `$COMPUTERNAME` or  instanceID of the container running serverless-init. Enabling the per instance logging is possible by setting `DD_AAS_INSTANCE_LOGGING_ENABLED=false`.
image

### Motivation
[SLES-2168](https://datadoghq.atlassian.net/browse/SLES-2168)

Unit tests
Application testing in AAS and GCP

### Additional Notes

[SLES-2168]: https://datadoghq.atlassian.net/browse/SLES-2168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/serverless-init/log/log.go
```diff
@@ -54,30 +57,61 @@ func CreateConfig(origin string) *Config {
...
-			Path:    filePath,
+			Path:    os.ExpandEnv("/home/LogFiles/*$COMPUTERNAME*.log"),
 			Service: os.Getenv("DD_SERVICE"),
 			Tags:    tags,
 			Source:  source,
 		})
 		logsAgent.GetSources().AddSource(src)
 	}
+
+	if filePaths, set := os.LookupEnv(envVarTailFilePath); set {
```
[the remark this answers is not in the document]
jcstorms1: @hghotra Do you mean for non-AAS platforms? None of the other use cases for sidecar have a default volume, so users define their own volume when setting up their environment.

cmd/serverless-init/log/log_test.go
```diff
@@ -37,3 +37,12 @@ func TestIsEnabledFalse(t *testing.T) {
...
 	assert.False(t, isEnabled("FALSE"))
 }
+
+func TestIsInstanceTailingEnabled(t *testing.T) {
+	t.Setenv("DD_AAS_INSTANCE_LOGGING_ENABLED", "false")
+	assert.False(t, isInstanceTailingEnabled())
+	t.Setenv("DD_AAS_INSTANCE_LOGGING_ENABLED", "True")
+	assert.True(t, isInstanceTailingEnabled())
+	t.Setenv("DD_AAS_INSTANCE_LOGGING_ENABLED", "")
+	assert.True(t, isInstanceTailingEnabled())
```
duncanpharvey, line 47: Should `DD_AAS_INSTANCE_LOGGING_ENABLED` be False be default? I thought this was something we would give to customers if they were having issues with instance specific logging but would be off by default.

cmd/serverless-init/log/log.go
```diff
@@ -54,30 +57,61 @@ func CreateConfig(origin string) *Config {
...
 
-func addFileTailing(logsAgent logsAgent.ServerlessLogsAgent, source string, tags []string) {
-	if filePath, set := os.LookupEnv(envVarTailFilePath); set {
-		src := sources.NewLogSource("serverless-file-tail", &logConfig.LogsConfig{
+func addFileTailing(logsAgent logsAgent.ServerlessLogsAgent, source string, tags []string, origin string) {
+	// The Azure App Service log volume is shared across all instances. This leads to every instance tailing the same files.
+	// To avoid this, we want to add the azure instance ID to the filepath so each instance tails their respective system log files.
+	appServiceDefaultLoggingEnabled := origin == "appservice" && isInstanceTailingEnabled()
+
+	if appServiceDefaultLoggingEnabled {
```
hghotra, line 76: It seems like if instance tailing is disabled and `DD_SERVERLESS_LOG_PATH` is set to something other than `/home/LogFiles/*.log`, we'll still end up missing system logs?

jcstorms1: Yeah that's correct.

cmd/serverless-init/log/log_test.go
jcstorms1: We can change it, but we discussed releasing this as V2 of serverless-init. If we want to make `false` the default we won't need to do that.

cmd/serverless-init/log/log.go
```diff
@@ -81,3 +96,9 @@ func addFileTailing(logsAgent logsAgent.ServerlessLogsAgent, source string, tags
 func isEnabled(envValue string) bool {
 	return strings.ToLower(envValue) == "true"
 }
+
+// enabled by default
```
duncanpharvey, line 100: It looks like the behavior was changed to disabled by default but this comment still says enabled by default

cmd/serverless-init/log/log_test.go
```diff
@@ -37,3 +37,14 @@ func TestIsEnabledFalse(t *testing.T) {
 	assert.False(t, isEnabled("1"))
 	assert.False(t, isEnabled("FALSE"))
 }
+
+func TestIsInstanceTailingEnabled(t *testing.T) {
```
duncanpharvey, line 41: Is it worth adding a unit test where `DD_AAS_INSTANCE_LOGGING_ENABLED` is not set at all?

hghotra (approved):
👍🏼

cmd/serverless-init/log/log.go
```diff
@@ -54,19 +55,33 @@ func CreateConfig(origin string) *Config {
...
 	return logsAgent
 }
 
-func addFileTailing(logsAgent logsAgent.ServerlessLogsAgent, source string, tags []string) {
-	if filePath, set := os.LookupEnv(envVarTailFilePath); set {
+func addFileTailing(logsAgent logsAgent.ServerlessLogsAgent, source string, tags []string, origin string) {
+
+	appServiceDefaultLoggingEnabled := origin == "appservice" && isInstanceTailingEnabled()
+	// The Azure App Service log volume is shared across all instances. This leads to every instance tailing the same files.
+	// To avoid this, we want to add the azure instance ID to the filepath so each instance tails their respective system log files.
```
duncanpharvey, line 73: It's all files correct? Not just the system log files

duncanpharvey (changes requested):

cmd/serverless-init/log/log_test.go
jcstorms1: It already is with `""`. An unset env var in Go returns `""`

cmd/serverless-init/log/log.go
jcstorms1: Only if they set their custom log filename to contain the `instanceID`

cmd/serverless-init/log/log_test.go
duncanpharvey: Yes, an unset environment variable and an environment variable set to an empty string both return empty strings. But they are test different cases.

PR #35815, opened 2025-04-04
CAP-2292 add e2e test for api key refresh

kangyili:
### What does this PR do?

Add a test to verify api key refresh on orchestrator endpoint.

PR #36160, opened 2025-04-15
[AGTMETRICS-167]API Key refresh for metric endpoints (take 2)

StephenWakely:
### What does this PR do?

This PR was originally merged as #35646, but unfortunately on merging it caused the `TestZzzClusterAgentAPIKeyRefresh` e2e test to fail (#35815). So the PR had to be rolled back. #36146

The failure was because the the original PR had neglected to set the `ConfigSettingPath` for the orchestrator endpoints, so they weren't being updated on a key refresh.

This PR:
- Implements key refresh for metric endpoints as per the original.
- Updates CI settings so the orchestrator e2e tests are run when there are changes in the default forwarder
- Updates the orchestrator endpoints to set the `ConfigSettingPath`.

See #35646.

In addition I have now run both the agent and the cluster agent in minikube and tested under a kubernetes environment that keys are still being refreshed - with keys being set both through the config file and via environment variables.


StephenWakely:

vickenty (approved):

hush-hush (approved):

KevinFairise2 (approved):

CelianR (approved):

kangyili (approved):
thank you. I can see the test is green now https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/898796373

daniel-taf (approved):

DylanLovesCoffee (approved):

hghotra (approved):

PR #36174, opened 2025-04-15
Revert "Backport OpenSSL update (#35875)"

lu-zhengda:
This reverts commit 1d0e10a49592cd74e02a216a77c12cf7e033f028.

#incident-37187

We've observed increased Postgres integration connection premature close and increased query time in `7.65.0-rc` in staging and production clusters, result in inflated postgres query metrics (i.e. `postgresql.queries.count`). 
After reverting openssl3 version bump PR (bump from 3.4.0 to 3.4.1), we no longer seeing the issue. 
Why openssl3 version bump caused the issue requires additional investigation. 

Postgres async job inactive stops after reverting
image


spencergilbert (approved):

git-thuerk-done (approved):
good for docs!

PR #36182, opened 2025-04-16
Fix ECS Fargate check logic

aquiladayc:
Proposing a change to a logic that check if it's ecs fargate.
internal: CONS-7235

### DIFF
--- a/Dockerfiles/agent/cont-init.d/50-ecs.sh
+++ b/Dockerfiles/agent/cont-init.d/50-ecs.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
  
-if [[ "${ECS_FARGATE}" == "true" ]]; then
+if [[ "${AWS_EXECUTION_ENV}" == "AWS_ECS_FARGATE" ]]; then
     exit 0
 fi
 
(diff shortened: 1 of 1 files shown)


Dockerfiles/agent/cont-init.d/50-ecs.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
-
-if [[ -z "${ECS_FARGATE}" ]]; then
+ 
+if [[ "${ECS_FARGATE}" == "true" ]]; then
```
L3n41c, line 3: We would need to update https://github.com/DataDog/datadog-agent/blob/c5717efe0244d7fc8ebbd2ab5ee93f07e6c3fee8/pkg/config/env/environment.go#L65 as well to keep the same logic everywhere.

L3n41c (commented):
Wasn’t it simpler to just define `ECS_FARGATE` on ECS Fargate hosts and not define it elsewhere?

The problem is what should we do if someone sets `ECS_FARGATE` to `True` instead of `true`. Or `TRUE`. What about `yes` or `Yes`, etc.

aquiladayc:
Discussed internally in linked jira

PR #36190, opened 2025-04-16
feat(devx) Merge team names

chouetz:
### What does this PR do?
Merge `agent-devx-infra` and `agent-devx-loops` in a same team

### Motivation
Org update

I'm keeping the link to `#agent-devx-infra-reviews` for now, I might use the channel ID to be able to rename without impact


.ddqa/config.toml
```diff
@@ -50,20 +50,12 @@ github_team = "agent-configuration"
...
-jira_project = "ADXT"
-jira_issue_type = "QA"
-jira_statuses = ["To Do", "In Progress", "Done"]
-github_team = "agent-devx-loops"
-github_labels = ["team/agent-devx-loops"]
-exclude_members = []
-
-[teams."Agent CI Experience"]
+[teams."Agent Developer Experience"]
 jira_project = "ACIX"
```
CelianR, line 54: Would it be still the same board name?

CelianR (approved):

.ddqa/config.toml
chouetz: yes we will keep the same to have less updates to do

PR #36195, opened 2025-04-16
[Fleet] Stop Agent service before MSI install

julien-lebot:
### What does this PR do?
Stops the Agent service before running the MSI installer on Windows.

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1330

Due to MSIs InstallValidate issue, we should stop the services ourself before running the MSI

- Ran an upgrade with the Agent service running.
- Confirmed confirmed with Procmon that the service stops before install.
- Verified install succeeds and service restarts.
- Checked logs for errors.


pkg/fleet/installer/packages/datadogagent/interface_windows.go
```diff
@@ -303,6 +303,18 @@ func installAgentPackage(env *env.Env, target string, args []string, logFileName
 	}
 	logFile := path.Join(tempDir, logFileName)
 
+	// Stop the Datadog Agent services before running MSI command
+	log.Infof("stopping the datadogagent service")
+	err = winutil.StopService("datadogagent")
```
clarkb7, line 308: the service should be stopped before `removeAgentIfInstalled` is called. I think `installAgentPackage` is always called after that. It might even make sense to just add it into `removeAgentIfInstalled`, I can't think of a scenario where we wouldn't want to stop the services, the MSI is always going to try.

julien-lebot: ✅

jack0x2 (approved):

BaptisteFoy (approved):

PR #36207, opened 2025-04-16
[NDMII-3397] Make the Core loader the loader by default for the SNMP integration and check for SNMP legacy profiles when loader is not specified

Pierre-L42:
### What does this PR do?

This PR makes the Core loader the loader by default when the loader config is not specified for SNMP.

To ensure backward compatibility, we added logic to validate that the SNMP profiles written by the user are not written in the legacy Python way during the initialization of the Core check. In case there are profiles written in the legacy way and the loader is not specified, we fallback to the Python loader by returning an error in the Configure() function. (If the loader is specified, we do not return an error, the behavior should be the same as before).

A legacy profile is one that have a metric where MIB+name is specified but not the OID, or one where we write `symbol: someString`.

- Unit tests
- Tested locally and on Mimic3:
  - `loader: core`
    - With ok profiles: ok, Core check is ran.
    - With a legacy profile: ok, Core check is ran (legacy profiles are detected and skipped, we do not return legacy error during check init).
  - `loader: python`
    - ok, Python check is ran.
  - `loader not specified`
    - With ok profiles: the Core check runs by default instead of the Python check.
    - With a legacy profile: we try to load Core check but a legacy profile is detected and we fallback to Python check.
- Maybe write an E2E test?


misteriaud (approved):
LGTM for Agent Runtimes code 🎉

pkg/collector/scheduler.go
```diff
@@ -190,7 +190,15 @@ func (s *CheckScheduler) getChecks(config integration.Config) ([]check.Check, er
 			log.Debugf("Loading check instance for check '%s' using default loaders", config.Name)
 		}
 
-		for _, loader := range s.loaders {
+		// TODO: Remove this special case to use Core loader by default for SNMP
```
FlorianVeaux, line 193: That's a sensitive part, can we add a unit test for this?

pkg/collector/scheduler.go
```diff
@@ -190,7 +190,15 @@ func (s *CheckScheduler) getChecks(config integration.Config) ([]check.Check, er
...
+		loaderList := s.loaders
```
pgimalac, line 194: **🔨 warning:** ‏I believe you're changing the loaders list for every check, you need to duplicate the slice first to avoid this

```suggestion
		loaderList := slices.Clone(s.loaders)
```

pgimalac (changes requested):

pkg/collector/scheduler.go
pgimalac: Note you can just clone in the `if` block to avoid allocations for every check

pkg/collector/scheduler.go
Pierre-L42: Yes, done in [f7af4b6](https://github.com/DataDog/datadog-agent/pull/36207/commits/f7af4b6c27f5514cc197657567f560246cc8c230)

pkg/collector/scheduler.go
Pierre-L42: Nice catch, done in [f7af4b6](https://github.com/DataDog/datadog-agent/pull/36207/commits/f7af4b6c27f5514cc197657567f560246cc8c230)

pkg/collector/scheduler.go
```diff
@@ -190,7 +190,16 @@ func (s *CheckScheduler) getChecks(config integration.Config) ([]check.Check, er
...
+		if config.Name == "snmp" && selectedInstanceLoader == "" {
+			loaderList = slices.Clone(s.loaders)
+			if len(loaderList) == 2 && loaderList[0].Name() == "python" && loaderList[1].Name() == "core" {
+				loaderList[0], loaderList[1] = loaderList[1], loaderList[0]
+			}
```
pgimalac, line 199: You can make the code a bit simpler
```suggestion
			if len(loaderList) == 2 && loaderList[0].Name() == "python" && loaderList[1].Name() == "core" {
				loaderList = []check.Loader{loaderList[1], loaderList[0]}
			}
```

pgimalac (approved):
LGTM considering we plan on changing order of the loaders soon

pkg/collector/scheduler.go
Pierre-L42: Done in [b6e48e3](https://github.com/DataDog/datadog-agent/pull/36207/commits/b6e48e355d7a0b29516ff07677482af65c035cda)

maycmlee (approved):

pkg/collector/corechecks/snmp/internal/profile/profile_initconfig.go
```diff
@@ -25,8 +30,11 @@ func loadInitConfigProfiles(rawInitConfigProfiles ProfileConfigMap) (ProfileConf
 		initConfigProfiles[name] = profConfig
 	}
 
-	userProfiles := mergeProfiles(getYamlUserProfiles(), initConfigProfiles)
-	resolvedProfiles := resolveProfiles(userProfiles, getYamlDefaultProfiles())
+	userProfiles, haveLegacyUserProfile := getYamlUserProfiles()
+	userProfiles = mergeProfiles(userProfiles, initConfigProfiles)
+
+	defaultProfiles, haveLegacyDefaultProfile := getYamlDefaultProfiles()
```
TCheruy, line 36: nit: I think we can assume our default profiles are written in a non-legacy syntax. we probably don't need to check `haveLegacyDefaultProfile`

pkg/collector/corechecks/snmp/internal/profile/profile_resolver.go
```diff
@@ -42,6 +43,13 @@ func normalizeProfiles(pConfig ProfileConfigMap, defaultProfiles ProfileConfigMa
 			log.Warnf("failed to expand profile %q: %v", name, err)
 			continue
 		}
+
+		isLegacyMetrics := profiledefinition.IsLegacyMetrics(newProfileConfig.Definition.Metrics)
```
TCheruy, line 47: can this be checked in `readProfileDefinition` ? that way I believe we don't have to check anything in `normalizeProfiles`

pkg/networkdevice/profile/profiledefinition/yaml_utils.go
```diff
@@ -62,3 +64,25 @@ func (mtcl *MetricTagConfigList) UnmarshalYAML(unmarshal func(interface{}) error
 	*mtcl = multi
 	return nil
 }
+
+// UnmarshalYAML unmarshalls MetricsConfig
+func (mc *MetricsConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	type metricsConfig MetricsConfig
```
TCheruy, line 70: nit: Do we need this type copy?

releasenotes/notes/snmp-core-check-by-default-4e30981d1252f594.yaml
```diff
@@ -0,0 +1,11 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+features:
+  - |
+    Validate SNMP profiles when loader config is not specified and make Core loader the loader by default.
```
TCheruy, line 11: I would focus only on the fact that we're making core loader the default

TCheruy (commented):
This is great! 👏 Small comments

releasenotes/notes/snmp-core-check-by-default-4e30981d1252f594.yaml
NouhaManai96: seconding this ➕

NouhaManai96 (approved):

pkg/collector/corechecks/snmp/internal/profile/profile_resolver.go
Pierre-L42: Done in [b67943c](https://github.com/DataDog/datadog-agent/pull/36207/commits/b67943cd6f5591c7bbc795a1d25693b735676be6)

releasenotes/notes/snmp-core-check-by-default-4e30981d1252f594.yaml
Pierre-L42: Done in [727a439](https://github.com/DataDog/datadog-agent/pull/36207/commits/727a439944c1a928a19656726b511bf07ccf3a89)

pkg/networkdevice/profile/profiledefinition/yaml_utils.go
Pierre-L42: Yes because the `unmarshal` parameter represents the `UnmarshalYAML` function of the type we unmarshal.
Since we use the parameter to unmarshal into `mc` which is a `MetricsConfig` we need to first convert `mc` into a `metricsConfig` (which doesn't have a custom unmarshal function) to avoid infinite recursion

TCheruy: makes sense!

TCheruy (approved):
LGTM! Small update to the release note

git-thuerk-done (approved):
good for docs

PR #36214, opened 2025-04-16
[CXP-2073] Add e2e tests for process_config.additional_endpoints

daniel-taf:
### What does this PR do?

Adds Linux and Windows E2E tests for secret refreshing of `process_config.additional_endpoints` introduced in https://github.com/DataDog/datadog-agent/pull/36160. 

Ran the tests and they passed

### DIFF
--- a/test/fakeintake/client/client.go
+++ b/test/fakeintake/client/client.go
@@ -733,7 +733,7 @@ func (c *Client) GetAllProcessPayloadAPIKeys() (map[string]struct{}, error) {
 		return nil, err
 	}
 
-	keys := make(map[string]struct{}, 0)
+	keys := make(map[string]struct{})
 	for _, payload := range payloads {
 		keys[payload.APIKey] = struct{}{}
 	}
--- a/test/fakeintake/client/client.go
+++ b/test/fakeintake/client/client.go
@@ -727,15 +727,19 @@ func (c *Client) GetLastProcessPayloadAPIKey() (string, error) {
 
 // GetAllProcessPayloadAPIKeys fetches fakeintake on `/api/v1/collector` endpoint and returns
 // a list of unique API keys of the received process payloads
-func (c *Client) GetAllProcessPayloadAPIKeys() (map[string]struct{}, error) {
+func (c *Client) GetAllProcessPayloadAPIKeys() ([]string, error) {
 	payloads, err := c.getFakePayloads(processesEndpoint)
 	if err != nil {
 		return nil, err
 	}
 
-	keys := make(map[string]struct{})
+	keysFound := make(map[string]struct{})
+	keys := make([]string, 0)
 	for _, payload := range payloads {
-		keys[payload.APIKey] = struct{}{}
+		if _, ok := keysFound[payload.APIKey]; !ok {
+			keysFound[payload.APIKey] = struct{}{}
+			keys = append(keys, payload.APIKey)
+		}
 	}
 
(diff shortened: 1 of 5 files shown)


test/fakeintake/client/client.go
```diff
@@ -725,6 +725,26 @@ func (c *Client) GetLastProcessPayloadAPIKey() (string, error) {
...
+// GetAllProcessPayloadAPIKeys fetches fakeintake on `/api/v1/collector` endpoint and returns
+// a list of unique API keys of the received process payloads
+func (c *Client) GetAllProcessPayloadAPIKeys() ([]string, error) {
+	payloads, err := c.getFakePayloads(processesEndpoint)
+	if err != nil {
+		return nil, err
+	}
+
+	keysFound := make(map[string]struct{})
+	keys := make([]string, 0)
```
Stephanie0829, line 737: Curious what you think about pre-allocating capacity of `keys` to `length(payloads)`. I'm not too sure what the distribution of duplicate apiKeys looks like (i.e if there are many duplicates this may be overkill since you end up using less space than capacity, but if there are rarely duplicates this may save on allocations during append).

test/new-e2e/tests/process/linux_test.go
```diff
@@ -105,7 +109,65 @@ func (s *linuxTestSuite) TestAPIKeyRefreshCoreAgent() {
...
+				secretsutils.WithUnixSetupScript("/tmp/test-secret/secret-resolver.py", false),
+				agentparams.WithSkipAPIKeyInConfig(),
+			),
+		),
+	)
+
+	fakeIntakeClient := s.Env().FakeIntake.Client()
+	agentClient := s.Env().Agent.Client
+
+	fakeIntakeClient.FlushServerAndResetAggregators()
```
Stephanie0829, line 146: Curious if you could explain a bit what this FlushServerAndResetAggregators() function does (or any documentation to learn more)! It isn't obvious what the "aggregators" refers to.

Stephanie0829 (approved):

test/fakeintake/client/client.go
daniel-taf: I thought about it, but it would be overkill. I would say the normal case is to have one api key for all payloads, but the test I use it in, I would expect two.  Since the assert clauses run for about 20 minutes, there would be 20 payloads in total. Also, this function is used pretty sparingly, so I don't think it would make a difference regardless.

test/new-e2e/tests/process/linux_test.go
daniel-taf: So it performs two actions:

- Flushes the server to delete any payloads being stored in the server. I do this to start with a clean slate of process payloads because I expect there to be some changes (api key in this case) and don't want to pick up payloads from before that. 
- Reseting the aggregators is not relevant here but some teams have defined Aggregators following this [interface](https://github.com/DataDog/datadog-agent/blob/main/test/fakeintake/aggregator/common.go#L43) to store payloads in memory and perform any other actions that are specific to a payload type. We have a [processAggregator](https://github.com/DataDog/datadog-agent/blob/main/test/fakeintake/aggregator/processAggregator.go#L38) for example that is used to parse the processes data from a payload. But, we don't really use the storage functionality because we overwrite the stored payloads everytime we ask for new ones so its not affected by this function :sweat_smile:

KevinFairise2 (approved):

PR #36217, opened 2025-04-16
Optimize allocations in cgroup parsing

blt:
### What does this PR do?

This commit is an attempt to improve cgroup parsing -- especially concerned with v2 -- with reference to incident-37244. I have introduced new benchmarks and especially a "large" one that is also rigged up to do memory profiling. Baseline results prior to any code changes, other than introduction of the benchmark:

```
      768B    32.12MB (flat, cum) 93.77% of Total
         .          .     44:func (r *readerV2) parseCgroups() (map[string]Cgroup, error) {
      768B       768B     45:   res := make(map[string]Cgroup)
         .          .     46:
         .    32.12MB     47:   err := filepath.WalkDir(r.cgroupRoot, func(fullPath string, de fs.DirEntry, err error) error {
         .          .     48:           if err != nil {
         .          .     49:                   return err
         .          .     50:           }
         .          .     51:           if !de.IsDir() {
         .          .     52:                   return nil
         .          .     53:           }
         .          .     54:
         .          .     55:           id, err := r.filter(fullPath, de.Name())
         .          .     56:           if id != "" {
         .          .     57:                   relPath, err := filepath.Rel(r.cgroupRoot, fullPath)
         .          .     58:                   if err != nil {
         .          .     59:                           return err
         .          .     60:                   }
    2.54MB     9.57MB     61:                   res[id] = newCgroupV2(id, r.cgroupRoot, relPath, r.cgroupControllers, r.pidMapper)
         .          .     62:           }
         .          .     63:
         .          .     64:           return err
         .          .     65:   })
         .          .     66:   return res, err
```

and after:

```
  435.55kB    20.55MB (flat, cum) 91.79% of Total
         .          .     47:func (r *readerV2) parseCgroups(res map[string]Cgroup) (map[string]Cgroup, error) {
         .          .     48:   // Mark all existing cgroups for deletion. Those that are found in the DFS
         .          .     49:   // will be retained ultimately in this map, the rest will be removed. This
         .          .     50:   // allows us to elide expensive calls to newCgroupV2 -- one obligatory
         .          .     51:   // syscall per call, string allocation per -- for cgroups we have already
         .          .     52:   // seen.
         .          .     53:   for id := range res {
         .          .     54:           if cg, ok := res[id].(*cgroupV2); ok {
         .          .     55:                   cg.markedForDeletion = true
         .          .     56:           }
         .          .     57:   }
         .          .     58:
         .          .     59:   // Avoid the use of filepath.WalkDir which is allocation hungry, at least as
         .          .     60:   // of Go 1.23.8. Instead we perform our own depth-first search.
         .          .     61:   stack := []string{r.cgroupRoot}
         .          .     62:
         .          .     63:   for len(stack) > 0 {
         .          .     64:           // Pop the directory to search from the stack, alter said stack.
         .          .     65:           dir := stack[len(stack)-1]
         .          .     66:           stack = stack[:len(stack)-1]
         .          .     67:
         .          .     68:           id, err := r.filter(dir, filepath.Base(dir))
         .          .     69:           if err != nil {
         .          .     70:                   // If SkipDir is returned, record the cgroup if valid, then skip
         .          .     71:                   // pushing subdirectories.
         .          .     72:                   if errors.Is(err, filepath.SkipDir) {
         .          .     73:                           if id != "" {
         .          .     74:                                   relPath, err := filepath.Rel(r.cgroupRoot, dir)
         .          .     75:                                   if err != nil {
         .          .     76:                                           return nil, err
         .          .     77:                                   }
         .          .     78:                                   // Reuse existing cgroup if available
         .          .     79:                                   if existing, ok := res[id]; ok {
         .          .     80:                                           if cg, ok := existing.(*cgroupV2); ok {
         .          .     81:                                                   cg.relativePath = relPath
         .          .     82:                                                   cg.markedForDeletion = false
         .          .     83:                                           }
         .          .     84:                                   } else {
         .          .     85:                                           res[id] = newCgroupV2(id, r.cgroupRoot, relPath, r.cgroupControllers, r.pidMapper)
         .          .     86:                                   }
         .          .     87:                           }
         .          .     88:                           continue
         .          .     89:                   }
         .          .     90:                   return nil, err
         .          .     91:           }
         .          .     92:           if id != "" {
         .          .     93:                   relPath, err := filepath.Rel(r.cgroupRoot, dir)
         .          .     94:                   if err != nil {
         .          .     95:                           return nil, err
         .          .     96:                   }
         .          .     97:                   // Reuse existing cgroup if available
         .          .     98:                   if existing, ok := res[id]; ok {
         .          .     99:                           if cg, ok := existing.(*cgroupV2); ok {
         .          .    100:                                   cg.relativePath = relPath
         .          .    101:                                   cg.markedForDeletion = false
         .          .    102:                           }
         .          .    103:                   } else {
  326.95kB   892.86kB    104:                           res[id] = newCgroupV2(id, r.cgroupRoot, relPath, r.cgroupControllers, r.pidMapper)
         .          .    105:                   }
         .          .    106:           }
         .          .    107:
         .          .    108:           // Now read all entries and push only directories onto the stack. Note
         .          .    109:           // this is allocation hungry itself. A direct call to unix.Open is
         .          .    110:           // feasible but would require us to handle C-style strings.
         .          .    111:           //
         .          .    112:           // Might be worthwhile eventually.
         .    17.43MB    113:           entries, err := os.ReadDir(dir)
         .          .    114:           if err != nil {
         .          .    115:                   return nil, err
         .          .    116:           }
         .          .    117:           for _, entry := range entries {
         .          .    118:                   if entry.IsDir() {
         .     2.14MB    119:                           subPath := filepath.Join(dir, entry.Name())
  108.59kB   108.59kB    120:                           stack = append(stack, subPath)
         .          .    121:                   }
         .          .    122:           }
         .          .    123:   }
         .          .    124:
```

Flat costs are up but cummulative costs are down by 12Mb. In particular I have:

* changed the field ordering of the cgroupv2 struct so it is more compact in memory,
* updated the parseCgroup interface to reuse the passed `res` map,
* added a mark/sweep mechanism to the gathered v2 cgroups to avoid calling newCgroupV2 for already discovered cgroups, eliding allocation and syscall expense and
* removed a call in `inodeForPath` to `os.Stat` in favor of `unix.Fstatat` which elides some allocation overhead in `os.Stat`.

I have relied on existing unit tests, in addition to introducing new micro-benchmarks. My goal is to see the improvement in the Regression Detector results, not in micro.

The implementation is now more complicated and distinctly relies on this code being only built for Linux.

### Additional Notes

The current allocation hot-spot in the v2 parseCgroup is in `os.ReadDir` as the above demonstrates. I left notes to the effect that `unix.Open` might be profitable to explore next.

### DIFF
--- a/.gitlab/functional_test/regression_detector.yml
+++ b/.gitlab/functional_test/regression_detector.yml
@@ -112,6 +112,7 @@ single-machine-performance-regression_detector:
       --baseline-sha ${BASELINE_SHA} \
       --comparison-sha ${CI_COMMIT_SHA} \
       --target-config-dir test/regression/ \
+      --replicas 250 \
       --submission-metadata submission_metadata \
       --tags ${SMP_TAGS} || {
         exit_code=$?
--- a/.gitlab/functional_test/regression_detector.yml
+++ b/.gitlab/functional_test/regression_detector.yml
@@ -112,7 +112,7 @@ single-machine-performance-regression_detector:
       --baseline-sha ${BASELINE_SHA} \
       --comparison-sha ${CI_COMMIT_SHA} \
       --target-config-dir test/regression/ \
-      --replicas 250 \
+      --replicas 100 \
       --submission-metadata submission_metadata \
       --tags ${SMP_TAGS} || {
         exit_code=$?
--- a/.gitlab/functional_test/regression_detector.yml
+++ b/.gitlab/functional_test/regression_detector.yml
@@ -112,7 +112,7 @@ single-machine-performance-regression_detector:
       --baseline-sha ${BASELINE_SHA} \
       --comparison-sha ${CI_COMMIT_SHA} \
       --target-config-dir test/regression/ \
-      --replicas 100 \
+      --replicas 50 \
       --submission-metadata submission_metadata \
       --tags ${SMP_TAGS} || {
         exit_code=$?
--- a/pkg/util/cgroups/cgroupv2.go
+++ b/pkg/util/cgroups/cgroupv2.go
@@ -12,13 +12,19 @@ import (
 )
 
 type cgroupV2 struct {
+	// 8-byte aligned fields first
+	controllers map[string]struct{}
+	pidMapper   pidMapper
+	fr          fileReader
+	inode       uint64
+
+	// string fields (16 bytes each on 64-bit systems)
 	identifier   string
 	cgroupRoot   string
 	relativePath string
-	controllers  map[string]struct{}
-	fr           fileReader
-	pidMapper    pidMapper
-	inode        uint64
+
+	// bool field last
+	markedForDeletion bool
 }
 
 func newCgroupV2(identifier, cgroupRoot, relativePath string, controllers map[string]struct{}, pidMapper pidMapper) *cgroupV2 {
--- a/pkg/util/cgroups/go.mod
+++ b/pkg/util/cgroups/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/containerd/cgroups/v3 v3.0.5
 	github.com/google/go-cmp v0.7.0
 	github.com/stretchr/testify v1.10.0
+	golang.org/x/sys v0.32.0
 )
 
 require (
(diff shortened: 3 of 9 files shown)


blt:
Well the profiler suggests there's less allocation pressure after this change but the Regression Detector doesn't indicate any macro-level improvement. Closing.

PR #36221, opened 2025-04-16
Add hostname fallback to ddot connector

IbraheemA:
### What does this PR do?

In the datadogconnector component in DDOT, if a hostname can't be computed from the span/resource attributes on an incoming trace, it falls back to the default hostname resolved from core-agent.

Ran otel-agent locally and sent traces + metrics to sandbox (where host cannot be computed from span/resource attributes). Verified that host tags are applied to the metrics. Ran otel-agent on main (without the changes from this PR) with all else equal, and verified that `host` is missing from metrics tags.
image


songy23 (commented):
Why do you need to regenerate the pb go files?

comp/otelcol/otlp/components/connector/datadogconnector/connector_test.go
```diff
@@ -153,7 +160,7 @@ func newTranslatorWithStatsChannel(t *testing.T, logger *zap.Logger, ch chan []b
 	return tr
 }
 
-func TestContainerTagsNative(t *testing.T) {
+func TestContainerTagsAndHostnameNative(t *testing.T) {
```
songy23, line 163: Can you add another test that if OTLP payloads have a host, the host is kept?

pkg/proto/pbgo/core/api.pb.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.36.6
+// 	protoc-gen-go v1.36.5
```
songy23, line 3: This does not look correct, maybe you are using an older version of protoc

songy23:
I think you can simply reset all the .pb.go files to main e.g. `git checkout origin/main -- '**/*.pb.go'`

comp/otelcol/otlp/components/connector/datadogconnector/factory.go
```diff
@@ -19,15 +19,25 @@ import (
...
+	tagger   testutil.TaggerClient
+	hostname option.Option[string]
 }
 
+// SourceProviderFunc is a function that returns the source of the host.
+type SourceProviderFunc func(context.Context) (string, error)
+
 // NewFactoryForAgent creates a factory for datadog connector for use in OTel agent
-func NewFactoryForAgent(tagger tagger.Component) connector.Factory {
+func NewFactoryForAgent(tagger testutil.TaggerClient, hostGetter SourceProviderFunc) connector.Factory {
```
songy23, line 30: Implementation code should not depend on testutil, can you move the TaggerClient to e.g. traceutil/otel_util ?

songy23: Actually I don't quite understand why you need to change it from tagger.Component to testutil.TaggerClient, is it a typo?

comp/otelcol/otlp/testutil/testutil.go
```diff
@@ -610,3 +611,34 @@ func ProcessLogsAgentRequest(w http.ResponseWriter, r *http.Request) JSONLogs {
 	handleError(w, err, 0)
 	return jsonLogs
 }
+
+// TaggerClient provides client for tagger interface,
+// see comp/core/tagger for tagger functions; client for tagger interface
+type TaggerClient interface {
```
songy23, line 617: This may also be moved to github.com/DataDog/datadog-agent/comp/core/tagger/types if import works

songy23: hmmm actually I think you can just move this TaggerClient back to infra attribute processor. DD connector should just use tagger.Component

IbraheemA: I moved TaggerClient to a common module so that I could use `NewTestTaggerClient` in `connector_test`

songy23, line 617: This still needs cleanup

comp/core/tagger/types/types.go
```diff
@@ -225,3 +225,12 @@ type Subscription interface {
 	// Unsubscribe is used cancel subscription to the tagger
 	Unsubscribe()
 }
+
+// TaggerClient provides client for tagger interface,
+// see comp/core/tagger for tagger functions; client for tagger interface
+type TaggerClient interface {
```
songy23, line 231: OK so I think this is just a simplified interface of tagger.Component with fewer funcs. Ideally we should just remove this and have `TestTaggerClient` implement tagger.Component, but that can go to a follow-up PR.

cc @jackgopack4 can correct me

songy23 (approved):

comp/core/tagger/types/types.go
gabedos: Hey all, we also provide a test tagger client [here](https://github.com/DataDog/datadog-agent/blob/main/comp/core/tagger/fx-mock/fx.go) that can be easily set up using fx via the `SetupFakeTagger` method.

We're continually trying to simplify the tagger interface and make it easier for others to use. Please let us know if there's any issues setting this up on otel.

gabedos (approved):
Approval for cutoff v7.66. @IbraheemA is going to look into using the mockTagger and removing this `TaggerClient` on OTEL-2546. They need to deal with issues with go mods first.

PR #36223, opened 2025-04-16
remove show-error traceback

clarkb7:
### What does this PR do?
Replace `Write-Error` with `Write-Host` so we don't display a PowerShell traceback of the Show-Error call.
The traceback is always the same, just the `Show-Error` call, so it doesn't give any useful info but is verbose and distracts from the actual error output.

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1364

### Additional Notes

With an error from bootstrap executable
Before
image
After
image

with a testing `throw "banana phone"` line, which shows we don't get the real traceback now anyway
Before
image
After
image


julien-lebot (approved):

iglendd:
nice

PR #36232, opened 2025-04-17
compute client stats hash in agent

raphaelgavache:
### What does this PR do?

Compute the client stats hash in the agent as it's used only here for aggregation. This will avoid consistency issues on the tracer implementation side.

[matching PR](https://github.com/DataDog/dd-trace-java/pull/8704)


amarziali (approved):
Thanks!

ajgajg1134 (approved):

PR #36233, opened 2025-04-17
[AGTMETRICS-167]Fix key refresh when additional endpoints and main endpoint shares domain

StephenWakely:
### What does this PR do?

Fixes a bug from #36160 when the main endpoint domain is the same as the additional endpoint. When an api key was updated in the additional endpoint, this was dropping the main endpoints api key.

This PR ensures that only the additional endpoints api keys are updated in this situation.

Have updated the tests to test additional_endpoint key updates.

Tested manually with this config:

```yaml
dd_url: http://localhost:8080
api_key: ENC[mainkee]

secret_backend_command: /home/stephenwakely/src/secret-resolver.py

additional_endpoints:
  "http://localhost:8080":
  - ENC[kee1]
```

### DIFF
--- a/comp/forwarder/defaultforwarder/default_forwarder_test.go
+++ b/comp/forwarder/defaultforwarder/default_forwarder_test.go
@@ -61,3 +61,43 @@ func TestDefaultForwarderUpdateAPIKey(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, expectData, string(data))
 }
+
+func TestDefaultForwarderUpdateAdditionalEndpointAPIKey(t *testing.T) {
+	mockConfig := config.NewMock(t)
+	mockConfig.Set("api_key", "api_key1", pkgconfigmodel.SourceAgentRuntime)
+	log := logmock.New(t)
+
+	// starting API Keys, before the update
+	// main api_key is a duplicate of the additional_endpoints one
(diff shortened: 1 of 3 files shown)


rayz (approved):
lgtm, small question: why do we store both `apiKeys` and `dedupedAPIKeys` vs just storing `dedupedAPIKeys` for example? Do we lose information from `apiKeys` after deduping that we need for something?

StephenWakely:
> lgtm, small question: why do we store both `apiKeys` and `dedupedAPIKeys` vs just storing `dedupedAPIKeys` for example? Do we lose information from `apiKeys` after deduping that we need for something?

Awesome thanks!

Yeah, we need to be able to store the config location of the API keys so that we know if they get updated. But we need the deduped version so we don't send the same transaction to the same domain/endpoint pair twice.

eg, We could start with this config:

```yaml
dd_url: example.com
api_key: key1

additional_endpoints:
  - example.com
    key1
```

Obviously they are the same domain and Url, so we would only want to sent to `key1` once.

But then supposing the additional_endpoint key was updated to:

```yaml
dd_url: example.com
api_key: key1

additional_endpoints:
  - example.com
    key2
```

We need to keep track of both locations in the config, but now we need to update one of them to `key2` and now we need to send to `key1` and `key2`.

PR #36235, opened 2025-04-17
[CWS] Fix disabling of multiple rules

YoannGh:
### What does this PR do?

Fixes:
- a bug where no more than one runtime security rules could be disabled through custom ones.
- a bug where the policy of all default rules was changed to the policy of the first rule that disabled/override a default one, and thus `ruleset_loaded` events reporting wrong policy information.

### DIFF
--- a/releasenotes/notes/cws-fix-disable-rules-c34db2eedfa31681.yaml
+++ b/releasenotes/notes/cws-fix-disable-rules-c34db2eedfa31681.yaml
@@ -1,0 +1,13 @@
+# Each section from every release note are combined when the
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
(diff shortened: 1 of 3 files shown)


spikat (approved):

releasenotes/notes/cws-fix-disable-rules-c34db2eedfa31681.yaml
```diff
@@ -0,0 +1,13 @@
...
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+fixes:
+  - |
+    CWS: Fix a bug preventing more than one runtime security rule 
+    from being disabled, and causing wrong policy information to 
+    be reported in ruleset loaded events.
```
janine-c, line 13:
```suggestion
    CWS: Fixed a bug preventing more than one runtime security rule 
    from being disabled, which caused incorrect policy information to 
    be reported in ruleset loaded events.
```

janine-c (approved):

lebauce (approved):

PR #36241, opened 2025-04-17
Add instrumentation telemetry in status rendering for dce

pgimalac:
### What does this PR do?
Add a new metric `status.dce_render_errors` to the instrumentation telemetry, emitted when `pkg/template` would fail while std template would succeed (which most likely means the given template is using methods). 

### Motivation
We have too many templates in the agent to know for sure if some use methods, this telemetry will allow us to know if something might break when we do the switch of template library.

Tested manually:
Added a method call in a template, ran `status`, and checked that the metric was emitted.

### Additional Notes

See https://github.com/DataDog/datadog-agent/pull/36024 and https://github.com/DataDog/datadog-agent/pull/36040 for details.


iglendd (approved):
Looks good for Agent Telemetry

comp/core/agenttelemetry/impl/agenttelemetry_test.go
```diff
@@ -476,7 +477,8 @@ func TestRun(t *testing.T) {
 	// The order is not deterministic
 	profile0Len := len(r.(*runnerMock).jobs[0].profiles)
 	profile1Len := len(r.(*runnerMock).jobs[1].profiles)
-	assert.True(t, (profile0Len == 1 && profile1Len == 3) || (profile0Len == 3 && profile1Len == 1))
+	t.Logf("%+v", r.(*runnerMock).jobs)
```
jeremy-hanna, line 480: is this a left over debug statement or intentional?

jeremy-hanna (approved):

comp/core/agenttelemetry/impl/agenttelemetry_test.go
pgimalac: Kind of both ?
I had a hard time debugging this test (it's not clear which entries of the config end up in which "profile"), the log helps a lot 😄
So I think it's worth keeping

dustmop (approved):

janine-c (approved):

PR #36243, opened 2025-04-17
[SBOM] Generate SBOMs for running containers

lebauce:
### What does this PR do?

SBOMs are already generated in the security module of system-probe. This PR introduces
a remote collector to serve and fetch SBOMs over gRPC, similar to the remote process collector.
SBOMs are associated to the workloadmeta `container` type. These SBOMs are then forwarded to
the intake through the SBOM check.

### Motivation

This allows vulnerabilities to be generated for software not part of the base image.

Enable runtime security in `system_probe.yaml`:

```
runtime_security_config:
  enabled: true

  sbom:
    enabled: true
```

Enable SBOM collection in `datadog.yaml`

```
sbom:
  enabled: true
```

- Restart the agent.
- Start a docker container with `docker run -ti ubuntu /bin/bash`
- Make sure the status of SBOM for the container is "SUCCESS" in `agent workload-list --verbose"


comp/core/workloadmeta/collectors/internal/remote/sbomcollector/sbom_collector.go
```diff
@@ -0,0 +1,180 @@
...
+	if err := decoder.Decode(&bom); err != nil {
+		return workloadmeta.Event{}, fmt.Errorf("failed to unmarshal SBOM: %w", err)
+	}
+
+	if event.Kind != string(workloadmeta.KindContainer) {
+		return workloadmeta.Event{}, fmt.Errorf("expected KindContainer, got %s", event.Kind)
+	}
+
+	if event.ID == "" {
+		return workloadmeta.Event{}, fmt.Errorf("expected container ID, got empty")
```
GustavoCaso, line 85:
```suggestion
		return workloadmeta.Event{}, fmt.Errorf("expected event ID, got empty")
```

Maybe?

comp/core/workloadmeta/collectors/internal/remote/sbomcollector/sbom_collector.go
```diff
@@ -0,0 +1,180 @@
...
+)
+
+const (
+	collectorID       = "sbom-collector"
+	cacheValidityNoRT = 2 * time.Second
+)
+
+type client struct {
+	cl              sbompb.SBOMCollectorClient
+	parentCollector *streamHandler
```
GustavoCaso, line 40: It seems we are not using this field. Do we need it?

comp/core/workloadmeta/impl/dump_test.go
```diff
@@ -190,6 +194,8 @@ Health:
 Created At: 0001-01-01 00:00:00 +0000 UTC
 Started At: 0001-01-01 00:00:00 +0000 UTC
 Finished At: 0001-01-01 00:00:00 +0000 UTC
+----------- SBOM -----------
+SBOM is nil
```
GustavoCaso, line 198: Should we exercise a test in which the SBOM is not nil?

comp/core/workloadmeta/collectors/internal/remote/processcollector/process_collector.go
```diff
@@ -145,6 +145,10 @@ func (s *streamHandler) Port() int {
 	return s.port
 }
 
+func (s *streamHandler) Address() string {
+	return fmt.Sprintf(":%d", s.Port())
```
GustavoCaso, line 149: Should we make the `Port` function private? Since it is no longer part of the generic interface

comp/core/workloadmeta/collectors/internal/remote/generic.go
```diff
@@ -113,9 +113,17 @@ func (c *GenericCollector) Start(ctx context.Context, store workloadmeta.Compone
 		opts = append(opts, grpc.WithTransportCredentials(creds))
 	}
 
+	opts = append(opts, grpc.WithContextDialer(func(_ context.Context, url string) (net.Conn, error) {
+		if strings.HasPrefix(url, "/") {
+			return net.Dial("unix", url)
+		}
+
+		return net.Dial("tcp", url)
+	}))
```
GustavoCaso, line 122: I suggest making the generic implementation branch based on the URL provided via the address. I think we could probably create a new public function in the generic interface that each collector has to implement to return the correct connection struct

pkg/security/resolvers/sbom/resolver.go
```diff
@@ -366,53 +381,63 @@ func (r *Resolver) removePendingScan(containerID containerutils.ContainerID) {
...
+	}
+	scanDuration := time.Since(scanTime)
+
+	scanResult := &sbom.ScanResult{
+		CreatedAt: scanTime,
+		Duration:  scanDuration,
+		RequestID: string(sb.ContainerID),
+		Report:    report,
 	}
+	r.NotifyListeners(SBOMComputed, scanResult)
```
GustavoCaso, line 423: Are we adding a test for this new functionality?

comp/core/workloadmeta/collectors/internal/remote/sbomcollector/sbom_collector.go
lebauce: Indeed. Removed

comp/core/workloadmeta/impl/dump_test.go
lebauce: Added test

comp/core/workloadmeta/collectors/internal/remote/processcollector/process_collector.go
lebauce: :+1: 
Done

comp/core/workloadmeta/collectors/internal/remote/generic.go
lebauce: I can definitely add a `NewConnection` or `Connect` function to the generic interface.
Just to make sure I understand, why not modifying the generic implementation ? The socket vs tcp is an implementation detail.

GustavoCaso: > Just to make sure I understand, why not modify the generic implementation? The socket vs tcp is an implementation detail.

The way I see it, the generic wrapper simply calls the function that the different implementations have. So instead of checking the return value of `Address()` we can abstract that and use something like:

```go
opts := []grpc.DialOption{grpc.WithContextDialer(func(_ context.Context, url string) (net.Conn, error) {
    return c.StreamHandler.NewConnection(url)
})}
```

The way the logic of which type of connection is pushed to each specific implementation.

Of course this is a possible way of doing it, and the way you've done it is valid as well 😄 But If I were making the change, I would go for `NewConnection`

pkg/security/resolvers/sbom/resolver.go
lebauce: I added a test regarding the notification mechanism.
That being said, this functionality would deserve a bigger e2e test that I suggest to put in a dedicated PR.

comp/core/workloadmeta/collectors/internal/remote/sbomcollector/sbom_collector.go
lebauce: In this case, the event ID is really the container ID so I suggest to keep it as it as it would be easier to troubleshoot if that happens, does it ok to you ?

GustavoCaso: Absolutely 😄

pkg/proto/datadog/sbom/workloadmeta_sbom.proto
```diff
@@ -0,0 +1,20 @@
...
+option go_package = "pkg/proto/pbgo/sbom"; // golang
+
+message SBOMStreamParams {
+}
+
+message SBOMMessage {
+    bytes Data = 1;
+    string Kind = 2;
+    // ID is the ID for this entity, in a format specific to the entity Kind.
+    string ID = 3;
```
GustavoCaso, line 14: Should we rename it to KindID to avoid confusion in the future?

pkg/security/resolvers/sbom/resolver.go
GustavoCaso: Thanks. Agree on a new e2e test on separate to avoid increasing the scope of this one 😄

comp/core/workloadmeta/collectors/internal/remote/generic.go
GustavoCaso: We talked over Slack and agree that @lebauce point are correct and we do not to duplicate the logic in the NewConnection 😄

comp/core/workloadmeta/collectors/internal/remote/generic.go
```diff
@@ -113,9 +113,17 @@ func (c *GenericCollector) Start(ctx context.Context, store workloadmeta.Compone
 		opts = append(opts, grpc.WithTransportCredentials(creds))
 	}
 
+	opts = append(opts, grpc.WithContextDialer(func(_ context.Context, url string) (net.Conn, error) {
+		if strings.HasPrefix(url, "/") {
+			return net.Dial("unix", url)
```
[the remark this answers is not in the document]
GustavoCaso: I was mostly asking because I saw this in the codebase 

https://github.com/DataDog/datadog-agent/blob/6cd58aa800614fba17cb7c455b84aac36843717f/pkg/security/agent/client.go#L184-L197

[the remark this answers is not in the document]
lebauce: Based on offline discussions: UNIX sockets are supported on newer Windows so keeping it as it but using filepath.IsAbs instead

pkg/proto/datadog/sbom/workloadmeta_sbom.proto
lebauce: Shamelessly stole it from https://github.com/DataDog/datadog-agent/blob/main/comp/core/workloadmeta/def/types.go#L227-L228

comp/core/workloadmeta/collectors/internal/remote/sbomcollector/sbom_collector.go
```diff
@@ -0,0 +1,179 @@
...
+		return nil, fmt.Errorf("incorrect response type")
+	}
+
+	var collectorEvents []workloadmeta.CollectorEvent
+	collectorEvents = handleEvents(collectorEvents, []*sbompb.SBOMMessage{response}, workloadmetaEventFromSBOMEventSet)
+	log.Tracef("collected [%d] events", len(collectorEvents))
+	return collectorEvents, nil
+}
+
+func handleEvents[T any](collectorEvents []workloadmeta.CollectorEvent, sbomEvents []T, convertFunc func(T) (workloadmeta.Event, error)) []workloadmeta.CollectorEvent {
```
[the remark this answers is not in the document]
lebauce: No, because I was just lazy and copy/pasted the code from the existing remote collector.
Removed the generic

GustavoCaso (approved):
Awesome work

PR #36253, opened 2025-04-17
Stop reporting fake Ready status in WLM based on AD annotation

vboulineau:
### What does this PR do?

Remove AD logic from the PodWatcher as the PodWatcher has not been related to AD for a long time now.
The AD logic is put back in AD so that the `Ready` status in WLM is accurate and will allow `PodWatcher` to detect readiness changes to Pods with `tolerate-unready` in all other WLM subscribers.

### Motivation

Bugfix.

Pick a Pod with `tolerate-unready` and a readiness probe on a container that takes some time (like >10s) to be ready. The reported Container heath in `agent workload-list` for the container will stay `unhealthy` even if the container readiness has changed to `ready`.

### DIFF
--- a/comp/core/autodiscovery/listeners/container.go
+++ b/comp/core/autodiscovery/listeners/container.go
@@ -26,8 +26,9 @@ import (
 )
 
 const (
-	newIdentifierLabel    = "com.datadoghq.ad.check.id"
-	legacyIdentifierLabel = "com.datadoghq.sd.check.id"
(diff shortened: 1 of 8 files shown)


sblumenthal (approved):

gabedos (approved):
lgtm. Makes sense to have wlm reflect true pod state and later AD can resolve it's [cases](https://docs.datadoghq.com/containers/guide/container-discovery-management/?tab=datadogoperator#tolerate-unready-pods) with tolerating unready pods

PR #36254, opened 2025-04-17
[ AGNTLOG-56 ] -Add go module support for auditor def, impl-none, and mock

ddrthall:
### What does this PR do?
Registers various components of the new logs auditor as go modules. There should be no functional changes whatsoever in this MR, as the comp/logs/auditor path is not yet in use in the code base.


gh123man (approved):
ALP owned file LGTM

GustavoCaso (approved):
LGTM

ddrthall:
After some code alteration we no longer need to export the auditor as a go module

PR #36257, opened 2025-04-17
[fentry] Make config to enable CNM fentry

pimlu:
### What does this PR do?
This PR adds a `network_config.enable_fentry` option to make the tracer try to use fentry.

### Motivation
We are evaluating making another push towards fentry to fix some performance issues.

```
TEST_FENTRY_OVERRIDE=true sudo -E go test -tags=linux,linux_bpf,npm,process,test ./pkg/network/tracer -v --run  TestTracerSuite/fentry -count 1
```

This doesn't fall back to kprobes when fentry fails to load, but at least for now that's what I want to happen.

### DIFF
--- a/pkg/config/setup/system_probe.go
+++ b/pkg/config/setup/system_probe.go
@@ -309,6 +309,8 @@ func InitSystemProbeConfig(cfg pkgconfigmodel.Config) {
 
 	cfg.BindEnvAndSetDefault(join(netNS, "enable_ebpfless"), false, "DD_ENABLE_EBPFLESS", "DD_NETWORK_CONFIG_ENABLE_EBPFLESS")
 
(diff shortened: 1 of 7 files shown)


AyyLam (approved):

hmahmood (approved):

hmahmood:
There may be other places where we have tied to fentry to fargate like this that also need updating: https://github.com/datadog/datadog-agent/blob/main/pkg/network/tracer/tracer_linux_test.go#L1969

usamasaqib:
I would suggest adding this check
https://github.com/DataDog/datadog-agent/blob/main/pkg/ebpf/kernel_bugs.go#L11

hush-hush (approved):

brycekahle (approved):

PR #36261, opened 2025-04-17
[Backport 7.65.x] [CWS] Fix disabling of multiple rules

YoannGh:
Backport 8a66955d3143820e2d5baa7a40211e3bcf660f6b from #36235.

___

### What does this PR do?

Fixes:
- a bug where no more than one runtime security rules could be disabled through custom ones.
- a bug where the policy of all default rules was changed to the policy of the first rule that disabled/override a default one, and thus `ruleset_loaded` events reporting wrong policy information.

### DIFF
--- a/releasenotes/notes/cws-fix-disable-rules-c34db2eedfa31681.yaml
+++ b/releasenotes/notes/cws-fix-disable-rules-c34db2eedfa31681.yaml
@@ -1,0 +1,13 @@
+# Each section from every release note are combined when the
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+fixes:
+  - |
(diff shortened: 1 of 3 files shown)


releasenotes/notes/cws-fix-disable-rules-c34db2eedfa31681.yaml
```diff
@@ -0,0 +1,13 @@
...
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+fixes:
+  - |
+    CWS: Fixed a bug preventing more than one runtime security rule
+    from being disabled, and caused incorrect policy information to
+    be reported in ruleset loaded events.
```
hestonhoffman, line 13:
```suggestion
    CWS: Fixed a bug that prevented more than one runtime security rule
    from being disabled, and caused incorrect policy information to
    be reported in ruleset loaded events.
```

hestonhoffman (approved):
Small release note edit (same as the other one)

I just realized these are backports. This is minor, so if fixing this is an issue, you can ignore this edit

PR #36271, opened 2025-04-18
omnibus: wipe empty folders before packaging

chouquette:
### What does this PR do?

Purge empty directory before adding them to our packages

### Motivation

Fixing the size discrepancies between build with & without omnibus cache

### DIFF
--- a/omnibus/config/software/datadog-agent-dependencies.rb
+++ b/omnibus/config/software/datadog-agent-dependencies.rb
@@ -46,3 +46,12 @@ if windows_target?
   end
 end
 
+do build
+    # Delete empty folders that can still be present when building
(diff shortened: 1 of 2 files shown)


Pythyu (approved):

omnibus/config/software/datadog-agent-dependencies.rb
```diff
@@ -46,3 +46,12 @@
   end
 end
 
+do build
```
FlorentClarret, line 49:
```suggestion
build do
```

FlorentClarret (changes requested):

FlorentClarret (approved):

PR #36277, opened 2025-04-18
Use std os instead of external dependency for executable

pgimalac:
### What does this PR do?
Use the `Executable` function from the standard library package `os` instead of an external dependency.

### Motivation
We were using the external dependency before `Executable` was added to `os`. Now even the dependency just calls `os.Executable`, so we might as well use that directly https://github.com/kardianos/osext/blob/master/osext_go18.go

There should be no functional change, but anyway CI is enough.


GustavoCaso (approved):
LGTM

PR #36280, opened 2025-04-18
merge windows-kernel-integrations into windows-agent

clarkb7:
### What does this PR do?
Replace references to `windows-kernel-integrations` with `windows-agent`

### Motivation
Teams are in the process of merging, so point automation/scripts at windows-agent instead.

e.g. release management script uses CODEOWNERS to generate list of teams


julien-lebot (approved):

pgimalac (approved):

chouquette (approved):

tbavelier (approved):
LGTM for container-ecosystems single file (windows e2e)

L3n41c (approved):
LGTM for the single file owned by @DataDog/container-integrations.

GustavoCaso (approved):
LGTM 

I noticed there is still a reference to the kernel team in: 

https://github.com/DataDog/datadog-agent/blob/77e07ac0324b819bbbe40a28e2b9f14de523ac22/tasks/unit_tests/release_tests.py#L435

It is a unit test for a Python task. Is this something we want to rename as well?

clarkb7:
> It is a unit test for a Python task. Is this something we want to rename as well?

I think it would be nice to change it but there's no comment on how that HTML blob is generated/formatted so I don't want to fiddle with it. Maybe @DataDog/agent-devx can update it?

chouetz (approved):

chouetz:
> > It is a unit test for a Python task. Is this something we want to rename as well?
> 
> I think it would be nice to change it but there's no comment on how that HTML blob is generated/formatted so I don't want to fiddle with it. Maybe @DataDog/agent-devx (CODEOWNER) can update it?

Yes the idea was to be independent of the direct codebase for the tests (unlike on [this test](https://github.com/DataDog/datadog-agent/pull/36280/files?w=1#diff-2ac5b9edac3923d5b1d54ca2d1cc45322aac98a6fc31ca27689cc4b30f84a7a8R58)), so to me we could let a reference here

alexn-dd (approved):

PR #36285, opened 2025-04-18
Updated precommit install to avoid go vcs issues

CelianR:
### What does this PR do?

It was possible for go to crash when installing packages due to vcs build flags, removed them.

### DIFF
--- a/docs/public/setup.md
+++ b/docs/public/setup.md
@@ -214,7 +214,7 @@ To install it, run:
 
 ```sh
 python3 -m pip install pre-commit
-pre-commit install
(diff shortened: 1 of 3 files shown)


docs/public/setup.md
```diff
@@ -214,7 +214,7 @@ To install it, run:
 
 ```sh
 python3 -m pip install pre-commit
-pre-commit install
+GOFLAGS=-buildvcs=false pre-commit install
```
KevinFairise2, line 217: Can we add a comment somewhere explaining why it is needed?

KevinFairise2 (approved):

PR #33521, opened 2025-01-29
[AI-5082] Add FIPS logic to Python init

dkirov-dd:
### What does this PR do?

Adds logic to modify the `OPENSSL_CONF` and `OPENSSL_MODULES` environment variables whenever the Agent is running in FIPS mode.

### Motivation

For Python integrations, we want to set the above mentioned environment variables in order to have all libraries that bundle their own OpenSSL (like `cryptography`) benefit from the configuration and module included in the Agent.
This needs to be done as early as possible in the process in order to guarantee no "wrong" config could have been loaded beforehand.

### Describe how you validated your changes

Manually tested using the `dd-gitlab/docker_build_fips_agent7` image.
E2E tests added to confirm desired functionality.
Additional automated tests are being worked on to be added in the `integrations-core` CI. ([PR](https://github.com/DataDog/integrations-core/pull/19404))

PR #36293, opened 2025-04-18
[Backport 7.65.x] Resolve PythonHome in the initialization call outside of init hook

jeremy-hanna:
### What does this PR do?

Investigation into incident-37187 of increased query time and metric inflation of the postgres dbm check. Current hypothesis is that the `PythonHome` resolution was moved [into an init hook](https://github.com/DataDog/datadog-agent/pull/33521/files#diff-443184ab5594b2887f4fd9e68df5c2375d1d8f473818f5e33a5a74385b327d5eR236) which may cause a race condition with the python runtime initialization. This PR addresses this by resolving the `PythonHome` again during the Initialization(...) call

### DIFF
--- a/pkg/collector/python/init.go
+++ b/pkg/collector/python/init.go
@@ -233,8 +233,6 @@ func init() {
 	expvarPyInit = expvar.NewMap("pythonInit")
 	expvarPyInit.Set("Errors", expvar.Func(expvarPythonInitErrors))
 
-	resolvePythonHome()
-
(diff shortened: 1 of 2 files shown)


jeremy-hanna:
It looks like this resolves the issue seen in #incident-37187: https://dd.datad0g.com/notebook/297096/incident-37187?range=29821936&start=1744985799096&live=false

improvement

rtrieu (approved):

PR #36311, opened 2025-04-18
Resolve PythonHome in the initialization call outside of init hook

jeremy-hanna:
### What does this PR do?

see https://github.com/DataDog/datadog-agent/pull/36293 for details

Investigation into incident-37187 of increased query time and metric inflation of the postgres dbm check. Current hypothesis is that the PythonHome resolution was moved [into an init hook](https://github.com/DataDog/datadog-agent/pull/33521/files#diff-443184ab5594b2887f4fd9e68df5c2375d1d8f473818f5e33a5a74385b327d5eR236) which may cause a race condition with the python runtime initialization. This PR addresses this by resolving the PythonHome again during the Initialization(...) call


rtrieu (approved):

releasenotes/notes/bugfix-python-home-init-hook-3bce41cc40f4b91d.yaml
```diff
@@ -0,0 +1,12 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+fixes:
+  - |
+    Fixes a race condition introduced in PR #33521 around the PythonHome getting
```
pgimalac, line 11:
```suggestion
    Fixes a race condition introduced in 7.65.0 around the PythonHome getting
```

pgimalac (approved):
Not sure how that can fix it but it shouldn't make it worse
Although we now write the same global variable twice without any sync...

Considering the PythonHome variable is only used in two places basically (`Initialize` and `initFIPS`), maybe `resolvePythonHome` should just return it and we can get rid of the global ?

PR #36318, opened 2025-04-20
Add jq to agent

guyarb:
### What does this PR do?

Adds `jq` to the agent packaging 

### Motivation

Enabling formatting and querying of results coming from agent endpoints


omnibus/config/software/jq.rb
```diff
@@ -0,0 +1,56 @@
...
+    "--prefix=#{install_dir}/embedded",
+    "--disable-maintainer-mode",
+    "--disable-dependency-tracking",
+    "--disable-silent-rules",
+    "--disable-docs",
+    "--disable-valgrind",
+    "--with-oniguruma=builtin"
+  ]
+
+  command "./configure #{configure_options.join(' ')}", env: env
```
chouquette, line 51: Could you use the `configure` function instead? See https://github.com/DataDog/datadog-agent/blob/main/omnibus/config/software/curl.rb#L57 for an example

The `prefix` will be provided automatically so you can remove it from `configure_options` while doing so

chouquette (commented):
LGTM overall with just a small change needed

chouquette (approved):
LGTM, but I suppose we'll need to dig a bit more on the size increase

chouquette:
Sorry for adding this to the MQ, wrong PR tab :facepalm:

PR #36321, opened 2025-04-20
[CNM] Flush closed connections dynamically

akarpz:
### What does this PR do?

- Keeps a per-client flag in the tracer state that tracks if the closed connection buffer is at 50% capacity. 
- Adds a new endpoint to fetch this status, and process-agent will now periodically check this endpoint to eagerly run the connections check if we're on track to hit capacity.
- The periodic capacity check will operate if a new binary config is set to true (default false).
- The connections check will still _always_ run at 30s (was always configurable) intervals regardless of the capacity check. 
- If a user configures the connection interval, then dynamic interval will be disabled.
- More info [here](https://docs.google.com/document/d/1646571V0f2PbAXyGCriiYFaL9n1WjAfM6zHfG8ZKUps/edit?tab=t.0)

### Motivation

https://datadoghq.atlassian.net/browse/NPM-4369

More info [here](https://docs.google.com/document/d/1646571V0f2PbAXyGCriiYFaL9n1WjAfM6zHfG8ZKUps/edit?tab=t.0)

Pros: Hosts that previously dropped connections will do this less or not at all. Only run checks more frequently when necessary.

Cons: Increase in CPU on nodes that need to run checks more often


pkg/network/state.go
```diff
@@ -570,6 +581,14 @@ func (ns *networkState) storeClosedConnection(c *ConnectionStats) {
 			continue
 		}
 		client.closed.insert(c, ns.maxClosedConns)
+
+		// Check if the buffer is near capacity using the configured threshold
+		if float64(len(client.closed.conns)) >= float64(ns.maxClosedConns)*ns.closedConnectionsBufferThresholdRatio {
```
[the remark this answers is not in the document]
akarpz: I don't think this makes sense as it checks if the flag is already true before emitting this log

pkg/network/state_test.go
```diff
@@ -2590,6 +2590,158 @@ func TestDNSPIDCollision(t *testing.T) {
...
+	ns := NewState(nil, 1*time.Minute, maxClosedConns, closedConnsThresholdRatio, 100, 100, 100, 100, 100, 100, false, false).(*networkState)
+	require.NotNil(t, ns)
+
+	clientID := "test-client-flag-set"
+	ns.RegisterClient(clientID)
+
+	// Add connections below threshold
+	ns.Lock()
+	for i := 0; uint32(i) < capacityThreshold-1; i++ { // Add 8 connections
+		ns.storeClosedConnection(createTestConnectionStats(StatCookie(i)))
```
hmahmood, line 2627: You can use `networkState.StoreClosedConnection` in the tests which acquires the lock.

pkg/network/tracer/tracer.go
```diff
@@ -917,3 +920,14 @@ func setupConnectionProtocolMapCleaner(connectionProtocolMap *ebpf.Map, name str
 
 	return mapCleaner, nil
 }
+
+// IsClosedConnectionsNearCapacity checks if the closed connections buffer is near capacity for a specific client.
+// It returns true if near capacity, false otherwise, and an error if the state is not initialized.
+func (t *Tracer) IsClosedConnectionsNearCapacity(clientID string) (bool, error) {
+	if t.state == nil {
```
hmahmood: How is this possible?

pkg/config/setup/process.go
```diff
@@ -205,6 +205,9 @@ func setupProcesses(config pkgconfigmodel.Setup) {
 
 	procBindEnvAndSetDefault(config, "process_config.language_detection.grpc_port", DefaultProcessEntityStreamPort)
 
+	// Connection Capacity Check configuration
+	procBindEnvAndSetDefault(config, "process_config.connections_capacity_check_interval", 30*time.Second)
```
guyarb, line 209: we check the capacity every 30s? not 10s?

pkg/network/config/config.go
```diff
@@ -327,15 +331,16 @@ func New() *Config {
...
-		MaxFailedConnectionsBuffered:   uint32(cfg.GetInt64(sysconfig.FullKeyPath(netNS, "max_failed_connections_buffered"))),
-		ClosedConnectionFlushThreshold: cfg.GetInt(sysconfig.FullKeyPath(netNS, "closed_connection_flush_threshold")),
-		ClosedChannelSize:              cfg.GetInt(sysconfig.FullKeyPath(netNS, "closed_channel_size")),
-		ClosedBufferWakeupCount:        cfg.GetInt(sysconfig.FullKeyPath(netNS, "closed_buffer_wakeup_count")),
-		MaxConnectionsStateBuffered:    cfg.GetInt(sysconfig.FullKeyPath(spNS, "max_connection_state_buffered")),
-		ClientStateExpiry:              2 * time.Minute,
+		TCPFailedConnectionsEnabled:           cfg.GetBool(sysconfig.FullKeyPath(netNS, "enable_tcp_failed_connections")),
+		MaxTrackedConnections:                 uint32(cfg.GetInt64(sysconfig.FullKeyPath(spNS, "max_tracked_connections"))),
+		MaxClosedConnectionsBuffered:          uint32(cfg.GetInt64(sysconfig.FullKeyPath(spNS, "max_closed_connections_buffered"))),
+		ClosedConnectionsBufferThresholdRatio: cfg.GetFloat64(sysconfig.FullKeyPath(spNS, "closed_connections_buffer_threshold_ratio")),
```
guyarb, line 337: Where is `closed_connections_buffer_threshold_ratio` defined? 
also why do you use `spNS`? I do think it is polluting the global namespace of system-probe

pkg/network/state.go
```diff
@@ -570,6 +583,14 @@ func (ns *networkState) storeClosedConnection(c *ConnectionStats) {
...
+			if !client.closedConnectionsNearCapacity.Load() {
+				log.Warnf("Closed connections buffer for client %s is nearing capacity (%d/%d, threshold: %.2f%%). Setting flag.", clientID, len(client.closed.conns), ns.maxClosedConns, ns.closedConnectionsBufferThresholdRatio*100)
+				client.closedConnectionsNearCapacity.Store(true)
+			}
```
guyarb, line 592: isn't it a racy code? two instances can reach to line 589 at the same time

pkg/process/runner/runner.go
```diff
@@ -393,6 +398,53 @@ func (l *CheckRunner) basicRunner(c checks.Check) func() {
...
 }
 
+// getConnectionsCheckInterval validates 'process_config.connections_capacity_check_interval' (the "capacity interval").
+// It must be between 10s-30s (inclusive) and be a factor of defaultInterval (the main connections check interval).
+// If both are true, the capacity interval is returned.
+// If explicitly configured by the user but fails validation, a specific warning is logged and defaultInterval is returned.
+// Otherwise (e.g. not configured, or derived value is invalid), defaultInterval is returned.
+func (l *CheckRunner) getConnectionsCheckInterval(checkName string, defaultInterval time.Duration) time.Duration {
+	const minAllowedInterval = 10 * time.Second
+	const maxAllowedInterval = 30 * time.Second
```
guyarb, line 408: the max must match the configuration value of connection check.
We can set the connections check to report every 1m if we want, but the current code ignore that and forces us to use 30s or lower).

Also, the minimum interval can cause a violation as well, as if we set a connection check interval to 10s (we have that in a single customer due to high load), then we have 2 issues -
1. Race between the new endpoint and the connections check endpoint (cc: see my comment in the RFC about reusing the same endpoint)
2. you hard-coded prevent the connection check to be less than 10s

pkg/process/runner/runner.go
```diff
@@ -393,6 +398,53 @@ func (l *CheckRunner) basicRunner(c checks.Check) func() {
 	}
 }
 
+// getConnectionsCheckInterval validates 'process_config.connections_capacity_check_interval' (the "capacity interval").
+// It must be between 10s-30s (inclusive) and be a factor of defaultInterval (the main connections check interval).
+// If both are true, the capacity interval is returned.
+// If explicitly configured by the user but fails validation, a specific warning is logged and defaultInterval is returned.
+// Otherwise (e.g. not configured, or derived value is invalid), defaultInterval is returned.
+func (l *CheckRunner) getConnectionsCheckInterval(checkName string, defaultInterval time.Duration) time.Duration {
```
guyarb, line 406: do we have UTs for this?

pkg/process/checks/net.go
```diff
@@ -533,3 +598,47 @@ func getNetworkID(sysProbeClient *http.Client) (string, error) {
...
+	}
+
+	capacityCheckStart := time.Now()
+	url := sysprobeclient.ModuleURL(sysconfig.NetworkTracerModule, "/connections/check_capacity?client_id="+ProcessAgentClientID)
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return false, fmt.Errorf("error creating capacity check request %s: %w", url, err)
+	}
+
+	log.Tracef("Checking connections capacity endpoint: %s", url)
```
guyarb, line 614: perhaps you should use `shouldLog` as if the log level is not trace (99% of the time), then this is a redundant cpu consumption

guyarb (changes requested):
Blocking the PR on:

1. [comment](https://github.com/DataDog/datadog-agent/pull/36321/files#r2115351471) regarding breaking today's status and possible race conditions
2. The code is enabled even when USM is enabled.

pkg/system-probe/config/adjust_npm.go
```diff
@@ -82,6 +83,17 @@ func adjustNetwork(cfg model.Config) {
...
 
+	ratioPath := spNS("closed_connections_buffer_threshold_ratio")
+	if !cfg.IsConfigured(ratioPath) {
+		cfg.Set(ratioPath, defaultClosedConnectionsBufferThresholdRatioSystemProbe, model.SourceDefault)
+	} else {
+		ratio := cfg.GetFloat64(ratioPath)
+		if ratio <= 0 || ratio > 1.0 {
+			log.Warnf("Invalid value %f for %s, using default %f. Value must be > 0 and <= 1.0.", ratio, ratioPath, defaultClosedConnectionsBufferThresholdRatioSystemProbe)
+			cfg.Set(ratioPath, defaultClosedConnectionsBufferThresholdRatioSystemProbe, model.SourceAgentRuntime)
+		}
```
brycekahle: suggestion: create `validateFloat64` which mirrors the other `validate*` functions in `adjust.go` and use that.

brycekahle (approved):

pkg/network/config/config.go
hmahmood: Is the default here `1.0`?

hmahmood: How does it change if we decrease the check interval from the default of 30s? Do we have to set both configs?

pkg/network/state.go
hmahmood: I think this code only called when the lock for the state is held, so likely no need to for atomics.

pkg/system-probe/config/adjust_npm.go
```diff
@@ -17,10 +17,11 @@ import (
...
 const (
-	maxConnsMessageBatchSize     = 1000
-	maxOffsetThreshold           = 3000
-	defaultMaxProcessesTracked   = 1024
-	defaultMaxTrackedConnections = 65536
+	maxConnsMessageBatchSize                                = 1000
+	maxOffsetThreshold                                      = 3000
+	defaultMaxProcessesTracked                              = 1024
+	defaultMaxTrackedConnections                            = 65536
+	defaultClosedConnectionsBufferThresholdRatioSystemProbe = 0.75
```
hmahmood: Why is this 0.75? Don't we want this to be 1.0 to keep the status quo?

hmahmood (commented):
How the two new configs interact with each and the connection interval is not clear to me. Also what happens if one is set and the other is not? It would be great if we can figure out one config and define some sane defaults.

dustmop (approved):

pkg/config/setup/process.go
robertjli: Could we move this to a child of `process_config`, like `process_config.connections` or similar?

pkg/process/runner/runner.go
```diff
@@ -368,7 +368,12 @@ func (l *CheckRunner) basicRunner(c checks.Check) func() {
 			l.runCheck(c)
 		}
 
-		ticker := time.NewTicker(checks.GetInterval(l.config, c.Name()))
+		tickerInterval := checks.GetInterval(l.config, c.Name())
+		if c.Name() == checks.ConnectionsCheckName {
```
robertjli: Could you move this logic into `GetInterval()`? I see there's other check-specific logic in there already

pkg/system-probe/config/adjust_npm.go
akarpz: removed this config entirely. also removed the atomic

pkg/network/config/config.go
akarpz: removed this config

pkg/network/tracer/tracer.go
akarpz: it shouldn't be. removed

pkg/system-probe/config/adjust_npm.go
akarpz: removed the block in question

pkg/process/checks/net.go
akarpz: added, thanks

pkg/process/runner/runner.go
akarpz: I simplified this greatly and moved to `interval.go`. added UTs as well

pkg/process/runner/runner.go
akarpz: removed this and made it driven by a simple binary config. basically now:
- if user sets the interval, use that (unless it's less than 10s, then it will default to 10s)
- if dynamic interval is enabled, we will check capacity every 10s and run full check if capacity is breached. otherwise run every 30s
- else, existing behavior (30s fixed interval)

you can look at the UTs which validate this behavior

pkg/system-probe/config/adjust_npm.go
```diff
@@ -82,6 +83,17 @@ func adjustNetwork(cfg model.Config) {
...
+	if !cfg.IsConfigured(ratioPath) {
+		cfg.Set(ratioPath, defaultClosedConnectionsBufferThresholdRatioSystemProbe, model.SourceDefault)
+	} else {
+		ratio := cfg.GetFloat64(ratioPath)
+		if ratio <= 0 || ratio > 1.0 {
+			log.Warnf("Invalid value %f for %s, using default %f. Value must be > 0 and <= 1.0.", ratio, ratioPath, defaultClosedConnectionsBufferThresholdRatioSystemProbe)
+			cfg.Set(ratioPath, defaultClosedConnectionsBufferThresholdRatioSystemProbe, model.SourceAgentRuntime)
+		}
+	}
+
```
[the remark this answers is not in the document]
akarpz: removed this logic

cmd/system-probe/modules/network_tracer.go
```diff
@@ -97,6 +97,31 @@ func (nt *networkTracer) Register(httpMux *module.Router) error {
...
+			return
+		}
+
+		if isNearCapacity {
+			log.Debugf("Responding to capacity check request for client %s: Near capacity (200 OK)", clientID)
+			w.WriteHeader(http.StatusOK)
+		} else {
+			log.Tracef("Responding to capacity check request for client %s: Not near capacity (204 No Content)", clientID)
+			w.WriteHeader(http.StatusNoContent)
+		}
```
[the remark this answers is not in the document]
akarpz: added some telemetry for tracking % full during capacity checks

pkg/network/config/config.go
akarpz: thanks, removed this config entirely after discussion with hasan to simplify the code

pkg/network/state.go
```diff
@@ -570,6 +583,14 @@ func (ns *networkState) storeClosedConnection(c *ConnectionStats) {
 			continue
 		}
 		client.closed.insert(c, ns.maxClosedConns)
+
+		// Check if the buffer is near capacity using the configured threshold
+		if float64(len(client.closed.conns)) >= float64(ns.maxClosedConns)*ns.closedConnectionsBufferThresholdRatio {
+			if !client.closedConnectionsNearCapacity.Load() {
+				log.Warnf("Closed connections buffer for client %s is nearing capacity (%d/%d, threshold: %.2f%%). Setting flag.", clientID, len(client.closed.conns), ns.maxClosedConns, ns.closedConnectionsBufferThresholdRatio*100)
```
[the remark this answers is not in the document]
akarpz: this could only be hit every 10s, as that's the max interval the `closedConnectionsNearCapacity` flag is reset to false. once it's set to true, the log will be skipped

pkg/config/setup/process.go
akarpz: done, good suggestion

pkg/process/runner/runner.go
akarpz: moved, thanks!

pkg/network/state.go
```diff
@@ -551,6 +565,31 @@ func (ns *networkState) mergeByCookie(conns []ConnectionStats) ([]ConnectionStat
 	return conns, connsByKey
 }
 
+// getClientCapacityPercentage calculates the current capacity percentage (0-100) for a client's closed connections buffer
+func (ns *networkState) getClientCapacityPercentage(client *client) float64 {
+	if ns.maxClosedConns == 0 {
```
hmahmood: I think this would be a fatal condition, we shouldn't ever see this.

pkg/network/state.go
```diff
@@ -551,6 +565,31 @@ func (ns *networkState) mergeByCookie(conns []ConnectionStats) ([]ConnectionStat
...
+		client.closedConnectionsNearCapacity = false
+		return
+	}
+
+	// Calculate current capacity percentage and check threshold
+	capacityPct := ns.getClientCapacityPercentage(client)
+	client.closedConnectionsNearCapacity = capacityPct >= ClosedConnectionsCapacityThreshold
+
+	if client.closedConnectionsNearCapacity {
+		log.Warnf("Closed connections buffer for client %s is nearing capacity (%d/%d, %.2f%%). Setting flag.",
```
hmahmood: This could get noisy.

pkg/network/state.go
```diff
@@ -228,6 +237,9 @@ type client struct {
 	lastFetch time.Time
 	closed    *closedConnections
 	stats     map[StatCookie]StatCounters
+
+	// Flag for handling closed connection capacity pressure for this client
+	closedConnectionsNearCapacity bool
```
hmahmood: Do we need to keep this around? Why not just calculate this every time we need to?

pkg/network/state.go
```diff
@@ -570,6 +609,10 @@ func (ns *networkState) storeClosedConnection(c *ConnectionStats) {
 			continue
 		}
 		client.closed.insert(c, ns.maxClosedConns)
+
+		if !client.closedConnectionsNearCapacity {
+			ns.updateClientCapacity(clientID, client)
```
hmahmood: Do we need to update it here? We could just calculate this on demand.

pkg/process/checks/net.go
```diff
@@ -114,6 +122,20 @@ func (c *ConnectionsCheck) Init(syscfg *SysProbeConfig, hostInfo *HostInfo, _ bo
 	c.localresolver = resolver.NewLocalResolver(sharedContainerProvider, clock.New(), maxResolverAddrCacheSize, maxResolverPidCacheSize)
 	c.localresolver.Run()
 
+	// Initialize state for the capacity-based run logic only when dynamic interval is enabled
+	if c.useDynamicInterval {
+		c.lastFullRunTime = time.Time{} // Ensure the first run is a full run
+		// Guaranteed run interval is driven by the standard connections check interval (30s)
+		c.guaranteedRunInterval = ConnectionsCheckDefaultInterval
```
hmahmood: Isn't this always `ConnectionsCheckDefaultInterval`? Why keep around `guaranteedRunInterval`?

pkg/process/checks/net.go
```diff
@@ -172,13 +203,74 @@ func (c *ConnectionsCheck) Run(nextGroupID func() int32, _ *RunOptions) (RunResu
...
+			if c.statsd != nil {
+				_ = c.statsd.Count("datadog.process.connections.capacity_check_errors", 1, []string{}, 1)
+			}
+		}
+	} else {
+		log.Debug("system probe client not available, skipping capacity check.")
+	}
+
+	// Decide whether to run the full check
+	shouldRunFullCheck := isTimeForGuaranteedRun || isNearCapacity
```
hmahmood: What's a full check vs just a check? This could be confusing; I would just call this `shouldRunCheck`.

pkg/network/state.go
akarpz: this can only be called once every time the flag is flipped to true (every 10s at most)

pkg/process/checks/net.go
akarpz: agreed, updated

pkg/process/checks/net.go
akarpz: yeah we don't need it anymore after the simplifications

pkg/network/state.go
akarpz: agreed, changed it to fully on demand and no extra state variables

pkg/network/state.go
akarpz: removed

pkg/network/state.go
akarpz: removed anyway

pkg/network/state.go
akarpz: removed

pkg/process/checks/net.go
```diff
@@ -227,10 +225,10 @@ func (c *ConnectionsCheck) shouldRunCheck(start time.Time) bool {
 	}
 
 	// Dynamic interval mode: use capacity-aware logic
-	timeSinceLastRun := start.Sub(c.lastFullRunTime)
+	timeSinceLastRun := start.Sub(c.lastRunTime)
 	// Add a small tolerance (250ms) to account for differences in the check interval and the guaranteed run interval
-	guaranteedIntervalWithTolerance := c.guaranteedRunInterval - (time.Millisecond * 250)
-	isTimeForGuaranteedRun := c.lastFullRunTime.IsZero() || timeSinceLastRun >= guaranteedIntervalWithTolerance
+	guaranteedIntervalWithTolerance := ConnectionsCheckDefaultInterval - (time.Millisecond * 250)
```
hmahmood: Will do the math correctly? i.e., would `ConnectionsCheckDefaultInterval.Sub(250 * time.Millisecond)` be more correct here?

akarpz: I will update it, it worked in the e2e tests though

jack0x2 (approved):
LGTM for the file owned by windows-agent

pkg/process/checks/net.go
akarpz: I think since it's a duration doing the direct math is the correct approach

PR #36329, opened 2025-04-21
Make system-probe client test use real transport

vitkyrka:
### What does this PR do?

Currently, the client tests use TCP sockets, but this is not the actual
transport used by system-probe (UDS on Linux and named pipes on
Windows).  Modify the tests to use the correct transports to match the
real scenarios closer and test the `Get` function.  For this, move and reuse
some test support code used in the flare tests.

### Motivation

Better test coverage.
Review comment on https://github.com/DataDog/datadog-agent/pull/35790.


pkg/system-probe/api/client/client_test.go
```diff
@@ -6,17 +6,31 @@
...
+func startTestServer(t *testing.T, handler http.Handler) (string, *httptest.Server) {
+	t.Helper()
+
+	socketPath := testutil.SystemProbeSocketPath(t)
+	server, err := testutil.NewSystemProbeTestServer(handler, socketPath)
+	require.NoError(t, err)
+	if server != nil {
+		server.Start()
+		t.Cleanup(server.Close)
+	}
```
val06, line 29: nit: `require.NotNil(t, server)`

gjulianm (approved):

pkg/system-probe/api/server/testutil/server_windows.go
```diff
@@ -0,0 +1,41 @@
...
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	sysprobeserver "github.com/DataDog/datadog-agent/pkg/system-probe/api/server"
+)
+
+const (
+	// systemProbeTestPipeName is the test named pipe for system-probe
+	systemProbeTestPipeName = `\\.\pipe\dd_system_probe_flare_test`
```
val06, line 21: are we ok with this path being fixed for **all** tests? this is a different behavior than in the unix implementation

val06 (approved):
approved, with a few small comments

hush-hush (approved):

pkg/system-probe/api/server/testutil/server_windows.go
alexn-dd: Good point.  If by any chance tests are run in parallel, some will have random conflicts and access denied.
Each component area should have their own dedicated named pipe path to make the tests independent, easier to isolate and troubleshoot.

vitkyrka: Fixed.

pkg/system-probe/api/client/client_test.go
vitkyrka: Fixed.

pkg/system-probe/api/server/testutil/server_windows.go
vitkyrka: Fixed.

PR #36332, opened 2025-04-21
[EBPF] attacher: do not log unnecessary errors

gjulianm:
### What does this PR do?

Change how registry errors are logged by the uprobe attacher, to ensure that we don't report errors that aren't really errors and are instead part of the normal operation of the attacher.

### Motivation

Reduce log noise.

Ran manually with GPU and USM enabled, ensured that log wasn't polluted.

### DIFF
--- a/pkg/gpu/probe.go
+++ b/pkg/gpu/probe.go
@@ -383,7 +383,7 @@ func getAttacherConfig(cfg *config.Config) uprobes.AttacherConfig {
 		SharedLibsLibset:               sharedlibraries.LibsetGPU,
 		ScanProcessesInterval:          cfg.ScanProcessesInterval,
 		EnablePeriodicScanNewProcesses: true,
-		EnableDetailedLogging:          true,
+		EnableDetailedLogging:          false,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


pkg/gpu/probe.go
```diff
@@ -383,7 +383,7 @@ func getAttacherConfig(cfg *config.Config) uprobes.AttacherConfig {
 		SharedLibsLibset:               sharedlibraries.LibsetGPU,
 		ScanProcessesInterval:          cfg.ScanProcessesInterval,
 		EnablePeriodicScanNewProcesses: true,
-		EnableDetailedLogging:          true,
+		EnableDetailedLogging:          false,
```
guyarb, line 386: when do we enable detailed logging?

guyarb (approved):

brycekahle (approved):

pkg/gpu/probe.go
gjulianm: I want to make a separate PR to add a config option, but personally I've enabled manually when debugging and in unit tests.

PR #35626, opened 2025-03-28
use glibc buildimage for system-probe

brycekahle:
### What does this PR do?

Uses the glibc buildimage for all system-probe purposes, including local KMT compilation.

### Motivation

Standardization of build images

### Describe how you validated your changes

Tested with CI and local KMT launches.

### Additional Notes

This changes local system-probe builds to use our custom built clang for eBPF compilation. This relaxes the requirement of a specific `clang` version installed in the development environment.

PR #36333, opened 2025-04-21
[EBPF] kmt: use dynamic compiler user

gjulianm:
### What does this PR do?

This PR changes the KMT code to be able to deal with the case where the compiler container already has an user with the same UID as the host user.

### Motivation

In https://github.com/DataDog/datadog-agent/pull/35626, the base build image was changed and now it includes a `ubuntu` user with UID 1000. In cases where the developer's user has the same UID, this would make our code thinking the `compiler` user was already created, and would then find errors when it didn't have the expected `compiler` name.

Ran `kmt.prepare --compile-only` locally, forced a situation where the host UID was the 1000 and ensured it was working. Also tested with a different, non-matching UID.

### DIFF
--- a/tasks/kernel_matrix_testing/compiler.py
+++ b/tasks/kernel_matrix_testing/compiler.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from functools import cached_property, lru_cache
 import json
 import os
 import sys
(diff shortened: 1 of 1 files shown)


tasks/kernel_matrix_testing/compiler.py
```diff
@@ -151,21 +197,12 @@ def start(self) -> None:
...
-        gid = cast('Result', self.ctx.run("id -g")).stdout.rstrip()
-
-        if uid == 0:
-            # If we're starting the compiler as root, we won't be able to create the compiler user
-            # and we will get weird failures later on, as the user 'compiler' won't exist in the container
-            raise ValueError("Cannot start compiler as root, we need to run as a non-root user")
-
-        # Now create the compiler user with same UID and GID as the current user
-        self.exec(f"getent group {gid} || groupadd -f -g {gid} compiler", user="root")
-        self.exec(f"getent passwd {uid} || useradd -m -u {uid} -g {gid} compiler", user="root")
```
brycekahle, line 164: I think we can just add `-o` to create a non-unique user, and that would solve the problem.

gjulianm: TIL, I didn't know we could do that. However, I've tried and it's a bit messy still, as it seems you'll log in not as `compiler` but as `ubuntu` even when you tell Docker to use the user `compiler`. See the output from these two commands:

``` 
$ docker exec -u compiler -i -t kmt-compiler-arm64 whoami
ubuntu
``` 

and  using `su` from the root user gives the same result

``` 
docker exec -u root -i -t kmt-compiler-arm64 bash
(ddpy3) root@07643502758b:/# su compiler
$ whoami
ubuntu
``` 

It's a bit problematic because then the home directory is `/home/ubuntu`, which gets a bit confusing because any changes done in `/home/compiler` config files (e.g., the `.bashrc`) don't get loaded, which causes some fun issues. In this case I'd rather keep the previous solution where we just switch the user and the rest is explicit, seems less complicated.

brycekahle (approved):

PR #36335, opened 2025-04-21
net path windows clients 

jack0x2:
### What does this PR do?
Uses socket options on windows to implement `syn` netpath on windows clients. Windows clients do not allow for the use of raw sockets. This implementation gives us the ability to do NetPath on widows clients, but is not as reliable as raw socket options used for windows servers. 

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1343
https://datadoghq.atlassian.net/browse/WINA-1507

Tested via manual testing on client and server versions of windows. 

Added `tcp_socket` options to existing e2e tests. 

Future implementations may need to enable to use of raw sockets to allow for other options for net path. This requires a new socket to be created for each hop. It also misses some hops due to the microsoft kernel not matching all ICMP packets to the TCP stream. Thus it should only be used when raw packets are not available.


releasenotes/notes/net-path-windows-clients-82d54dab7ca7dc7e.yaml
```diff
@@ -0,0 +1,12 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+features:
+  - |
+    Enables support for NetPath on Window Client versions.
```
urseberry, line 11:
```suggestion
    Enables support for NetPath on Windows client versions.
```

pimlu:
I have a small TCP testbed I ran this with on Windows Server 2019 Datacenter, and got this error:
```
Traceroute failed: failed to create raw socket: failed to set TCP_FAIL_CONNECT_ON_ICMP_ERROR: An unknown, invalid, or unsupported option or level was specified in a getsockopt or setsockopt call.
```
Was wondering if this depends on your version of Windows

jack0x2:
Yes it only works on Windows 10/ Windows server 2019 after version 2004. The docs for it are here: https://learn.microsoft.com/en-us/windows/win32/api/ws2ipdef/ns-ws2ipdef-icmp_error_info

pkg/networkpath/traceroute/winconn/winconn_nonraw_windows.go
```diff
@@ -0,0 +1,292 @@
...
+		Addr: [4]byte{dst[0], dst[1], dst[2], dst[3]},
+	}
+
+	return connect(r.Socket, sa)
+}
+
+// getHoppAddress gets the address of the hop
+// this will only work if errorInfo is set
+// otherwise it will fail
+func (r *Conn) getHoppAddress() (net.IP, error) {
```
pimlu, line 161: nit: single p spelling

pkg/networkpath/traceroute/winconn/winconn_nonraw_windows.go
```diff
@@ -0,0 +1,292 @@
...
+	var errorInfo ICMP_ERROR_INFO
+	var errorInfoSize = int32(unsafe.Sizeof(errorInfo))
+	// getsockopt for ICMP_ERROR_INFO
+	// this will have the address of the hop
+	err := getsockopt(
+		r.Socket,
+		windows.IPPROTO_TCP,
+		windows.TCP_ICMP_ERROR_INFO,
```
pimlu, line 169: Does windows clear `TCP_ICMP_ERROR_INFO` when you try to dial again? That way, if a hop is truly missing there will be no address instead of an old address repeated?

test/new-e2e/tests/netpath/network-path-integration/netpath_int_win_test.go
```diff
@@ -46,13 +60,19 @@ func (s *windowsNetworkPathIntegrationTestSuite) TestWindowsNetworkPathIntegrati
...
-		// s.checkGoogleDNSUDP(c, hostname)
+		s.checkGoogleTCPSocket(c, hostname)
+		s.checkGoogleDNSUDP(c, hostname)
 
 	}, 5*time.Minute, 3*time.Second)
 }
+
+// disable defender firewall for windows
+func (s *windowsNetworkPathIntegrationTestSuite) disableFirewall() error {
+	_, err := s.Env().RemoteHost.Host.Execute("Set-NetFirewallProfile -Profile Domain,Public,Private -Enabled False")
```
pimlu, line 76: Maybe we should double check with the e2e team that this is fine before merging

pkg/networkpath/traceroute/tcp/tcpv4_windows.go
```diff
@@ -12,13 +12,88 @@ import (
...
+
+	rtt := time.Duration(0)
+	if !hopIP.Equal(net.IP{}) {
+		rtt = end.Sub(start)
+	}
+
+	return &common.Hop{
+		IP:       hopIP,
+		Port:     0, // TODO: fix this
+		ICMPType: 0, // TODO: fix this
```
pimlu, line 91: nit: Forward this out of `ICMP_ERROR_INFO`?

pimlu (approved):

pkg/networkpath/traceroute/winconn/winconn_nonraw_windows.go
jack0x2: It is a new socket each time. So there will be no old error there.

pkg/networkpath/traceroute/tcp/tcpv4_windows.go
jack0x2: I should be able to do that. Let me check and add that

jack0x2: Okay I added the Type and Code from the packet, but I don't have the port so I will leave that blank for now

urseberry (approved):

clarkb7 (approved):

PR #36349, opened 2025-04-22
[ACIX-648] create unique qualification tags for agent 6

sabrina-datadog:
### What does this PR do?
This PR updates the way we manage qualification tags for Agent 6 releases. Instead of reusing the same qualification tag and force-updating it:

- now we create unique timestamped qualification tags in this format: `qualification-<timestamp>`
- these tags will not be moved once created (which will fix the need to force-fetch tags and eliminate clobbering errors)
- this means that `# of RCs = the # of qualification tags` per qualification phase
- we always use `git ls-remote` to get the qualification tags so that we can use the remote as the source of truth
- at the end of the qualification phase we delete all generated qualification tags

This removes the need to force-fetch tags and eliminates clobbering errors.
### Motivation
Previously, we reused a single tag called qualification and force moved it to a new commit for a new rc during the qualification phase. This caused Git errors for developers when pulling or fetching changes since Git doesn't automatically overwrite local tags and it will throw an error since local and remote are out of sync.

tested by calling functions locally by using them to create qualification tags, delete them, check if it's in qualification phase

### DIFF
--- a/tasks/libs/common/git.py
+++ b/tasks/libs/common/git.py
@@ -360,7 +360,6 @@ def push_tags_in_batches(ctx, tags, force_option="", delete=False):
 
     for idx in range(0, len(tags), TAG_BATCH_SIZE):
         batch_tags = tags[idx : idx + TAG_BATCH_SIZE]
-        print(f"git {command} origin {' '.join(batch_tags)}{force_option}")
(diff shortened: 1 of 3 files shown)


chouetz (approved):
Minor comments LGTM, thanks!

spencergilbert (approved):
Agree with @chouetz's feedback

PR #36356, opened 2025-04-22
Revert "[CIP] feat: Use s3 clone strategy for datadog-agent CI jobs (#35902)"

alopezz:
### What does this PR do?

Reverts #35902 

### Motivation

#incident-37509. We suspect this might be interfering with the way the version is computed due to a newly introduced caching mechanism, causing public images to be tagged incorrectly.


chouquette (approved):
LGTM, thanks!

PR #36371, opened 2025-04-22
[ha-agent] Add ha_agent.enabled and config_id to template

AlexandreYang:
### What does this PR do?

[ha-agent] Add ha_agent.enabled and config_id to template

### Motivation

Needed for HA Agent is in Preview.

### DIFF
--- a/pkg/config/config_template.yaml
+++ b/pkg/config/config_template.yaml
@@ -4345,9 +4345,9 @@ api_key:
 
 ## @param config_id - string - optional
 ## The config_id configuration is used by Fleet Automation and High Availability Agent
-## to assign specific Config ID to an Agent.
-## For High Availability, all agents with the same config_id with be part of the same group of agents;
-## meaning that, one agent within this group will be designated as primary and others as standby.
+## to assign a specific config ID to an Agent.
+## For High Availability, all Agents with the same config_id with be part of the same group of Agents;
+## meaning that, one Agent within this group is designated as primary and others as standby.
 #
 # config_id:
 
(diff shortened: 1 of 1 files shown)


jedupau (approved):

pkg/config/config_template.yaml
```diff
@@ -4332,6 +4332,26 @@ api_key:
...
+  ## Set to true to enable High Availability Agent feature.
+  #
+  # enabled: false
+
+
+## @param config_id - string - optional
+## The config_id configuration is used by Fleet Automation and High Availability Agent
+## to assign specific Config ID to an Agent.
+## For High Availability, all agents with the same config_id with be part of the same group of agents;
+## meaning that, one agent within this group will be designated as primary and others as standby.
```
maycmlee, line 4350:
```suggestion
## to assign a specific config ID to an Agent.
## For High Availability, all Agents with the same config_id with be part of the same group of Agents;
## meaning that, one agent within this group is designated as primary and others as standby.
```

maycmlee (approved):
Just a couple of small suggestions, but approving to not block merge.

pkg/config/config_template.yaml
```diff
@@ -4332,6 +4332,25 @@ api_key:
...
+  #
+  # enabled: false
+
+
+## @param config_id - string - optional
+## The config_id configuration is used by High Availability Agent to assign a specific config ID to an Agent.
+## When used with `ha_agent.enabled: true`, all Agents with the same config_id will be part of the same group of Agents;
+## meaning that, one Agent within this group is designated as primary and others as standby.
+#
+# config_id:
```
hush-hush, line 4351: Should this be within the `ha_agent` section ?

AlexandreYang: @hush-hush Right now, the only feature setting config_id manually is HA Agent, so we can probably keep it here for now and move if needed.

hush-hush (approved):

PR #36373, opened 2025-04-22
Properly set python version for macos runners

CelianR:
### What does this PR do?

Sets python version on macos runners using pyenv according to the major version within `.python-version`.

The `.python-version` file contains only the major version and the minor but not the patch, my way of selecting the version is simple (given M found within the file):
- If there is a 3.M.x version (or multiple matches) that is already installed on the runner, select the latest version in this list
- If there is no match, download and install the latest one

Example [job](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/906816685).
Example [job](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/908223868) with custom `PYTHON_VERSION` set.

Won't look for future minor versions (e.g. `3.12.6` is installed but `3.12.7` could be used). Will look at this when the AMI is updated however.


chouetz (approved):
Like the previous PR I would probably put all this shell code in a dedicated script but this is a minor comment. Thanks

PR #36374, opened 2025-04-22
[CWS] change log level of sysctl error

safchain:
### What does this PR do?

Allow to generate event even if some "collector" failed. Reports the error as warn.


pkg/security/probe/sysctl/snapshot.go
```diff
@@ -51,14 +52,12 @@ type SnapshotEvent struct {
...
+func NewSnapshotEvent(ignoredBaseNames []string, kernelCompilationFlags map[string]uint8) (*SnapshotEvent, *multierror.Error, error) {
 	se := &SnapshotEvent{
 		Sysctl: NewSnapshot(),
 	}
-	if err := se.Sysctl.Snapshot(ignoredBaseNames, kernelCompilationFlags); err != nil {
-		return nil, err
-	}
-	return se, nil
+	err := se.Sysctl.Snapshot(ignoredBaseNames, kernelCompilationFlags)
+	return se, err, nil
```
paulcacheux, line 60: the error part is always nil now, do we really need it ?

spikat (approved):

PR #36377, opened 2025-04-22
[EBPF-702] gpu: add preliminary support for handling MIG devices

gjulianm:
### What does this PR do?

This PR adds preliminary support for handling GPU devices:

- Adds some MIG-related fields to the `Device` type.
- Adds support in the `DeviceCache` for MIG devices.
- Ensures the code for enumerating devices supports MIG devices.
- Ensures we understand `nvidia.com/mig-*` resources in k8s containers.

### Motivation

Preliminary support for MIG devices in GPU monitoring.

Unit tests added. Tested manually on instances with MIG devices.


pkg/gpu/context.go
```diff
@@ -20,6 +21,7 @@ import (
 )
 
 const nvidiaResourceName = "nvidia.com/gpu"
```
val06, line 23: consider consolidating those consts with `pkg/util/gpu/common.go` (perhaps we can change them to exported and use it)

pkg/gpu/context.go
```diff
@@ -178,7 +180,7 @@ func (ctx *systemContext) filterDevicesForContainer(devices []*ddnvml.Device, co
 	numContainerGPUs := 0
 	for _, resource := range container.AllocatedResources {
 		// Only consider NVIDIA GPUs
-		if resource.Name != nvidiaResourceName {
+		if resource.Name != nvidiaResourceName && !strings.HasPrefix(resource.Name, nvidiaMigResourceName) {
```
val06, line 183: _nit:_ consider introducing a helper func (as you are using similar conditional logic here and in gpu.go)

pkg/gpu/safenvml/cache.go
```diff
@@ -25,15 +25,18 @@ type DeviceCache interface {
 	SMVersionSet() map[uint32]struct{}
 	// All returns all devices in the cache
 	All() []*Device
+	// AllRootDevices returns all root devices in the cache
+	AllRootDevices() []*Device
```
val06, line 29: I don't like the Root name, let's call it differently:

`Parent` / `Physical`

pkg/gpu/safenvml/cache.go
```diff
@@ -75,7 +78,13 @@ func NewDeviceCacheWithOptions(lib SafeNVML) (DeviceCache, error) {
 
 		cache.uuidToDevice[dev.UUID] = dev
 		cache.allDevices = append(cache.allDevices, dev)
+		cache.allRootDevices = append(cache.allRootDevices, dev)
```
val06, line 81: if we maintain a separate list of physical devices, why do we need to add them also to the Devices? why not having phyiscal and mig devices lists and if needed to return a full list, just return an `append` of those 2?

pkg/gpu/safenvml/device.go
```diff
@@ -85,13 +88,29 @@ type SafeDevice interface {
 type Device struct {
```
val06, line 88: we should keep this struct as simple as possible.
I don't like the fact that this serves both mig and physical device, while some fields of this struct are relevant just for one type. we need to find a better design

pkg/gpu/cuda/env.go
```diff
@@ -78,7 +91,42 @@ func getVisibleDevices(systemDevices []*ddnvml.Device, cudaVisibleDevices string
...
+// or the list of all devices if it is not. This replicates the behavior of CUDA with MIG devices, where
+// if any MIG device is present, only the first MIG child is visible and all other devices are hidden.
+func keepEitherFirstMIGOrAllDevices(devices []*ddnvml.Device) []*ddnvml.Device {
+	for _, device := range devices {
+		if device.IsMIG {
+			return []*ddnvml.Device{device}
+		}
+	}
+
+	return devices
```
val06, line 129: the name of the function already suggests that something is really wrong :)

val06 (changes requested):
reviewed. let's talk about MIG. I feel very uncomfortable with the proposed changes

pkg/gpu/context.go
gjulianm: Done with the helper function

pkg/gpu/safenvml/cache.go
gjulianm: The main issue is that order is important for CUDA so we need to keep that as much as possible. Also, we're either returning all devices or all physical devices, so I thought it's better to have those two lists pre-computed rather than appending every time we use them.

pkg/gpu/safenvml/device.go
gjulianm: On one hand I agree that having a device that serves two types of devices is not that good, but on the other I think it's the best option we have. It replicates the same abstraction as in NVML (the `nvml.Device` type doesn't distinguish between MIG or non-MIG) and it also fits with the idea that most consumers of the `Device` type don't actually care whether it's MIG or not. I think introducing a separate `MIGDevice` type and then having to add an interface with the common fields/methods would be more complex and we wouldn't gain too much.

pkg/gpu/cuda/env.go
gjulianm: I'm not sure, to me it makes the code that uses it more understandable straight away. The main weird thing for me is the CUDA logic, but we can't change that.

pkg/gpu/safenvml/cache.go
gjulianm: Changed to Physical

davidor (approved):
:+1: for files owned by container-platform

pkg/gpu/cuda/env.go
```diff
@@ -78,13 +92,53 @@ func getVisibleDevices(systemDevices []ddnvml.Device, cudaVisibleDevices string)
...
+
+// adjustVisibleDevicesForMig adjusts the list of visible devices taking into account how CUDA
+// presents MIG devices in order
+func adjustVisibleDevicesForMig(visibleDevices []ddnvml.Device) []ddnvml.Device {
+	// CUDA removes all devices with MIG feature enabled and then appends at the
+	// end of the list the first MIG child of those devices. So we split the
+	// list into two: one with devices without MIG feature enabled (which
+	// includes physical devices and MIG children) and one with MIG children of
+	// MIG-enabled physical devices.
+	var nonMigEnabledDevices []ddnvml.Device
```
val06, line 106: `migDisabledDevices`

pkg/gpu/cuda/env.go
```diff
@@ -78,13 +92,53 @@ func getVisibleDevices(systemDevices []ddnvml.Device, cudaVisibleDevices string)
...
+	}
+
+	// Now merge the two lists
+	return append(nonMigEnabledDevices, migChildDevices...)
+}
+
+// keepEitherFirstMIGOrAllDevices returns the list of devices with only the first MIG child if it is present,
+// or the list of all devices if it is not. This replicates the behavior of CUDA with MIG devices, where
+// if any MIG device is present, only the first MIG child is visible and all other devices are hidden.
+func keepEitherFirstMIGOrAllDevices(devices []ddnvml.Device) []ddnvml.Device {
```
val06, line 127: no need to express the whole algorithm of the function in its name.
just call it `selectMigDevice` and leave the detailed comment to help understand the logic of the code

pkg/gpu/cuda/env.go
```diff
@@ -78,13 +92,53 @@ func getVisibleDevices(systemDevices []ddnvml.Device, cudaVisibleDevices string)
...
+	for _, device := range devices {
+		migDevice, ok := device.(*ddnvml.MIGDevice)
+		if ok {
+			return []ddnvml.Device{migDevice}
+		}
+	}
+
+	return devices
```
val06, line 135: in which case this function could be called with a list of devices where non of them is a mig device?

pkg/gpu/cuda/env.go
```diff
@@ -78,13 +92,53 @@ func getVisibleDevices(systemDevices []ddnvml.Device, cudaVisibleDevices string)
...
 }
 
 // getDeviceWithMatchingUUIDPrefix returns the first device with a UUID that
 // matches the given prefix. If there are multiple devices with the same prefix
 // or the device is not found, an error is returned.
-func getDeviceWithMatchingUUIDPrefix(systemDevices []ddnvml.Device, uuidPrefix string) (ddnvml.Device, error) {
+func getDeviceWithMatchingUUIDPrefix(systemDevices []ddnvml.Device, uuidPrefix string, searchMigChildren bool) (ddnvml.Device, error) {
```
val06, line 141: `getDeviceByUUIDPrefix`

pkg/gpu/cuda/env.go
```diff
@@ -78,13 +92,53 @@ func getVisibleDevices(systemDevices []ddnvml.Device, cudaVisibleDevices string)
...
+			return []ddnvml.Device{migDevice}
+		}
+	}
+
+	return devices
 }
 
 // getDeviceWithMatchingUUIDPrefix returns the first device with a UUID that
 // matches the given prefix. If there are multiple devices with the same prefix
 // or the device is not found, an error is returned.
```
val06, line 140: can we really face a situation with multiple devices matching the prefix? why do we have such a flow in our code at all? sounds like an entry into a rabbit hole, what should we do with an error that will be returned?

pkg/gpu/safenvml/cache.go
```diff
@@ -25,15 +25,18 @@ type DeviceCache interface {
 	SMVersionSet() map[uint32]struct{}
 	// All returns all devices in the cache
 	All() []Device
+	// AllPhysicalDevices returns all root devices in the cache
+	AllPhysicalDevices() []Device
```
val06, line 29: consider adding also AllMigDevice to the interface to return just the mig devices

val06 (changes requested):
reviewed

pkg/gpu/context.go
```diff
@@ -189,6 +188,17 @@ func (ctx *systemContext) filterDevicesForContainer(devices []ddnvml.Device, con
...
 				break
 			}
+
+			// If the device has MIG children, check if any of them matches the resource ID
+			physicalDevice, ok := device.(*ddnvml.PhysicalDevice)
+			if ok {
+				for _, migChild := range physicalDevice.MIGChildren {
+					if resource.ID == migChild.UUID {
+						filteredDevices = append(filteredDevices, migChild)
+						break
```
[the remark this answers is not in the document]
gjulianm: You're right, fixed.

pkg/gpu/cuda/env.go
```diff
@@ -60,16 +66,24 @@ func getVisibleDevices(systemDevices []ddnvml.Device, cudaVisibleDevices string)
...
 		case strings.HasPrefix(visibleDevice, "GPU-"):
-			matchingDevice, err = getDeviceWithMatchingUUIDPrefix(systemDevices, visibleDevice)
+			matchingDevice, err = getDeviceWithMatchingUUIDPrefix(systemDevices, visibleDevice, false)
+			if err != nil {
+				return filteredDevices, err
+			}
+
+			// Selecting a parent device when there are MIG children just returns the first MIG child,
+			physicalDevice, ok := matchingDevice.(*ddnvml.PhysicalDevice)
+			if ok && physicalDevice.HasMIGFeatureEnabled && len(physicalDevice.MIGChildren) > 0 {
```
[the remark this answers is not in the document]
gjulianm: We just select the physical device itself, without changing `matchingDevice`

pkg/gpu/cuda/env.go
```diff
@@ -78,13 +92,53 @@ func getVisibleDevices(systemDevices []ddnvml.Device, cudaVisibleDevices string)
...
-	return filteredDevices, nil
+	return keepEitherFirstMIGOrAllDevices(filteredDevices), nil
+}
+
+// adjustVisibleDevicesForMig adjusts the list of visible devices taking into account how CUDA
+// presents MIG devices in order
+func adjustVisibleDevicesForMig(visibleDevices []ddnvml.Device) []ddnvml.Device {
+	// CUDA removes all devices with MIG feature enabled and then appends at the
+	// end of the list the first MIG child of those devices. So we split the
+	// list into two: one with devices without MIG feature enabled (which
```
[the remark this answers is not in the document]
gjulianm: Clarified the comment. The idea is that here we include physical devices with mig disabled, and then also mig children of other physical devices that do have MIG enabled.

pkg/gpu/cuda/env.go
gjulianm: But `selectMigDevice` is a different thing, I'd rather keep the name that describes what it does so that it's easier to understand the code that calls this function.

pkg/gpu/cuda/env.go
gjulianm: For example, in systems without MIG devices.

pkg/gpu/cuda/env.go
gjulianm: CUDA allows users to specify just part of the UUID. If the users specifies a prefix that matches multiple GPUs, we'll run into this issue. CUDA will also fail to execute in this case, so we're matching that behavior.

pkg/gpu/safenvml/device.go
```diff
@@ -108,10 +109,27 @@ type Device interface {
 type PhysicalDevice struct {
 	SafeDevice
 	DeviceInfo
+
+	// HasMIGFeatureEnabled is true if the device has the MIG feature enabled. Note that a device might
+	// have the MIG feature enabled but no MIG devices created yet.
+	HasMIGFeatureEnabled bool
```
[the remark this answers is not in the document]
gjulianm: I wanted to have a strong distinction between "is mig device" and "has mig feature".

pkg/gpu/safenvml/device.go
```diff
@@ -140,25 +158,97 @@ func NewPhysicalDevice(dev nvml.Device) (*PhysicalDevice, error) {
...
 
-	// Get the number of cores and memory size. This is not part of the fillBasicDataFromNVML
-	// because it might not work for MIG devices.
-	cores, err := device.SafeDevice.GetNumGpuCores()
+	migEnabled, _, err := safeDev.GetMigMode()
+	if err == nil && migEnabled == nvml.DEVICE_MIG_ENABLE {
+		device.HasMIGFeatureEnabled = true
+
+		if err := device.fillMigChildren(); err != nil {
+			return nil, err
```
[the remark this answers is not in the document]
gjulianm: I think so, as we use the MIG children to get the memory and core count of the device. Added the comment.

pkg/gpu/cuda/env.go
val06: maybe perform the logic inside `adjustVisibleDevicesForMig` function and remove this function altogether?

pkg/gpu/cuda/env.go
val06: resolving this, to continue the conversation around this method in [this](https://github.com/DataDog/datadog-agent/pull/36377#discussion_r2080328419) thread

pkg/gpu/cuda/env.go
gjulianm: There's two places where this function is used, and in both cases it's in the return, not sure if that'll look better. I think that even if it's a longer name, considering it's a helper function only used here it should be ok, no?

val06 (approved):

PR #36382, opened 2025-04-22
Add container-integrations as codeowners to wlm collectors and tag list

sblumenthal:
### What does this PR do?

Add container-integrations as codeowners to wlm collectors and tag list

### Motivation

The container-integrations team is responsible for the collection of data from various container-based sources

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -587,16 +587,16 @@
 /pkg/tagger/                            @DataDog/container-platform
 /pkg/windowsdriver/                     @DataDog/windows-kernel-integrations
 /comp/core/workloadmeta/collectors/internal/cloudfoundry  @DataDog/agent-integrations
-/comp/core/workloadmeta/collectors/internal/containerd    @DataDog/container-integrations
-/comp/core/workloadmeta/collectors/internal/crio          @DataDog/container-integrations
-/comp/core/workloadmeta/collectors/internal/docker        @DataDog/container-integrations
-/comp/core/workloadmeta/collectors/internal/ecs           @DataDog/container-integrations
-/comp/core/workloadmeta/collectors/internal/ecsfargate    @DataDog/container-integrations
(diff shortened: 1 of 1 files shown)


wdhif (commented):
Approved, as discussed, it would be worth it to have @DataDog/container-platform as co-owner of the collectors.

wdhif (approved):

.github/CODEOWNERS
```diff
@@ -586,8 +586,18 @@
...
+/comp/core/workloadmeta/collectors/internal/crio          @DataDog/container-integrations
+/comp/core/workloadmeta/collectors/internal/docker        @DataDog/container-integrations
+/comp/core/workloadmeta/collectors/internal/ecs           @DataDog/container-integrations
+/comp/core/workloadmeta/collectors/internal/ecsfargate    @DataDog/container-integrations
+/comp/core/workloadmeta/collectors/internal/kubeapiserver @DataDog/container-integrations
+/comp/core/workloadmeta/collectors/internal/kubelet       @DataDog/container-integrations
+/comp/core/workloadmeta/collectors/internal/kubemetadata  @DataDog/container-integrations
+/comp/core/workloadmeta/collectors/internal/nvml          @DataDog/ebpf-platform
+/comp/core/workloadmeta/collectors/internal/podman        @DataDog/container-integrations
+/comp/core/tagger/collectors            @DataDog/container-platform @DataDog/container-integrations
```
davidor, line 600: To reduce the number of people pinged for reviews, it could be better to keep only container-platform here. We own the tagger, and this directory contains a core part of its functionality.

What do you think?

sblumenthal: It does, that is why you are still in the codeowners. It also contains the full mapping of every container and kubernetes related tags which we extract and which the container-integrations team is responsible for maintaining

davidor (approved):

PR #36386, opened 2025-04-22
CXP-2078 add warning about pidmode misconfiguration on fargate

gengnamstyle:
### What does this PR do?
- adds a one-off warning when `pidMode` is likely misconfigured on linux ecs fargate
    - experimentation was done to determine that an extra container `aws-fargate-pause` exists only when `"pidMode":"task"` on the task definition. This was verified via personal image testing and when querying for all ecs containers on dddev and prod. 
        - [prod query](https://app.datadoghq.com/containers?query=orchestrator%3A%28%22ecs%22%20OR%20%22fargate%22%29%20container_name%3Aaws-fargate-pause&selectedTopGraph=none)
        - [dddev query](https://dddev.datadoghq.com/containers?query=orchestrator%3A%28%22ecs%22%20OR%20%22fargate%22%29%20container_name%3Aaws-fargate-pause&selectedTopGraph=timeseries)

Warning message:
```
Process collection is likely misconfigured. Please ensure your task definition has "pidMode":"task" set. See https://docs.datadoghq.com/integrations/ecs_fargate/?tab=webui#process-collection for more information.
```

### Motivation

Customers frequently misconfigure or omit pidMode when running the agent on ECS Fargate which can result in a support escalation. We would like to emit a log to warn that it should be set to task.

I decided to go against having an E2E test for this since the cost/time of spinning up a misconfigured ecs instance to test a log didn't seem worth it. Open to creating one if there are strong opinions.

1. Built image based off of this branch and deployed to personal docker repo via these docs https://datadoghq.atlassian.net/wiki/spaces/agent/pages/4540892047/Building+the+Datadog+Agent+using+a+Dev+Container
2. Created an ecs fargate cluster manually with this aws account: `sso-agent-sandbox-account-admin`
```aws-vault login sso-agent-sandbox-account-admin```
3. Created 2 task definitions, one with `"pidMode":"task"` and the other missing it

Get a shell with aws creds:
```aws-vault exec sso-agent-sandbox-account-admin```

Register definition 
```aws ecs register-task-definition --cli-input-json file://<FILEPATH>```

```
{
  "pidMode": "task", (ONE DEFINITION WITH THIS LINE AND ONE NOT)
  "containerDefinitions": [
    {
      "name": "agent",
      "image": "<IMAGE REFERENCE>",
      "cpu": 0,
      "portMappings": [],
      "essential": true,
      "environment": [
        {
          "name": "ECS_FARGATE",
          "value": "true"
        },
        {
          "name": "DD_API_KEY",
          "value": "<DDDEV_API_KEY>"
        },
        {
          "name": "DD_PROCESS_AGENT_PROCESS_COLLECTION_ENABLED",
          "value": "true"
        }
      ],
      "mountPoints": [],
      "volumesFrom": [],
      "logConfiguration": {
        "logDriver": "awslogs",
        "options": {
          "awslogs-group": "<CREATE LOG GROUP>",
          "mode": "non-blocking",
          "awslogs-create-group": "true",
          "max-buffer-size": "25m",
          "awslogs-region": "us-east-1",
          "awslogs-stream-prefix": "ecs"
        }
      },
      "systemControls": []
    },
    {
      "name": "nginx",
      "image": "nginx:latest",
      "cpu": 256,
      "memory": 256,
      "portMappings": [
        {
          "containerPort": 80,
          "hostPort": 80,
          "protocol": "tcp"
        }
      ],
      "essential": true,
      "environment": [],
      "mountPoints": [],
      "volumesFrom": [],
      "logConfiguration": {
        "logDriver": "awslogs",
        "options": {
          "awslogs-group": "<CREATE LOG GROUP>",
          "mode": "non-blocking",
          "awslogs-create-group": "true",
          "max-buffer-size": "25m",
          "awslogs-region": "us-east-1",
          "awslogs-stream-prefix": "ecs"
        }
      },
      "systemControls": []
    }
  ],
  "family": "matthew-geng-nginx-agent-pidmode-task",
  "taskRoleArn": "arn:aws:iam::376334461865:role/ecsTaskRole",
  "executionRoleArn": "arn:aws:iam::376334461865:role/ecsTaskExecutionRole",
  "networkMode": "awsvpc",
  "volumes": [],
  "placementConstraints": [],
  "requiresCompatibilities": [
    "FARGATE"
  ],
  "cpu": "1024",
  "memory": "3072",
  "runtimePlatform": {
    "cpuArchitecture": "ARM64",
    "operatingSystemFamily": "LINUX"
  },
  "enableFaultInjection": false
}
```
4. Created a service and alternated referenced task definition without `pidMode` and with.

If you would like to be able to run `aws ecs execute` on the datadog-agent pod, you must create the service via the CLI and the datadog-agent must be defined first in your task definition.

See this ongoing issue, reported by our @robertjli https://github.com/aws/containers-roadmap/issues/2268

5. Checked cloudwatch logs for single warning
image

6. verified dddev.datadoghq.com had no nginx processes 
7. verified log does not print on local linux devcontainer using these docs https://datadoghq.atlassian.net/wiki/spaces/agent/pages/3171582307/Creating+your+Dev+Container

- there doesn't seem to be documentation on this `aws-fargate-pause` container that only exists when `"pidMode":"task"` is set in the task definition. So this warning could be triggered incorrectly if this indirect dependency is changed somehow.

### CODE REVIEW

File: pkg/process/checks/process.go [hunk 1]
```diff
@@ -296,6 +303,12 @@ func (p *ProcessCheck) run(groupID int32, collectRealTime bool) (RunResult, erro
 	procsByCtr := fmtProcesses(p.scrubber, p.disallowList, procs, p.lastProcs, pidToCid, cpuTimes[0], p.lastCPUTime, p.lastRun, p.lookupIdProbe, p.ignoreZombieProcesses, p.serviceExtractor, pidToGPUTags)
 	messages, totalProcs, totalContainers := createProcCtrMessages(p.hostInfo, procsByCtr, containers, p.maxBatchSize, p.maxBatchBytes, groupID, p.networkID, collectorProcHints)
 
+	once.Do(func() {
+		if !isECSFargatePidModeSetToTask(totalProcs, containers) {
+			log.Warn(`Process collection is likely misconfigured. Please check if your task definition has "pidMode":"task"`)
```
robertjli, line 308: I would change the message to be more clear about what the user should do, and potentially also link to public documentation.

Maybe something like `Please ensure your task definition has "pidMode": "task" set. See <doc> for more information.`

When you say "please check" could mean "check if it's there because it shouldn't be" or "check that it's there because it should be" and I personally read it the wrong way at first glance.

pkg/process/checks/process.go
```diff
@@ -30,20 +33,24 @@ import (
...
+	configPrefix                          = "process_config."
+	configCustomSensitiveWords            = configPrefix + "custom_sensitive_words"
+	configScrubArgs                       = configPrefix + "scrub_args"
+	configStripProcArgs                   = configPrefix + "strip_proc_arguments"
+	configDisallowList                    = configPrefix + "blacklist_patterns"
+	configIgnoreZombies                   = configPrefix + "ignore_zombie_processes"
+	awsECSFargatePidModeTaskContainerName = "aws-fargate-pause"
 )
 
+var once sync.Once // used for one-off warning
```
robertjli, line 52: I would give it a more descriptive name, ideally it should be named so that you don't need a comment at all 🙂 

Also, this likely makes more sense inside the ProcessCheck struct as it's called from a method of ProcessCheck

pkg/process/checks/process.go
```diff
@@ -30,20 +33,24 @@ import (
...
-	configDisallowList         = configPrefix + "blacklist_patterns"
-	configIgnoreZombies        = configPrefix + "ignore_zombie_processes"
+	emptyCtrID                            = ""
+	configPrefix                          = "process_config."
+	configCustomSensitiveWords            = configPrefix + "custom_sensitive_words"
+	configScrubArgs                       = configPrefix + "scrub_args"
+	configStripProcArgs                   = configPrefix + "strip_proc_arguments"
+	configDisallowList                    = configPrefix + "blacklist_patterns"
+	configIgnoreZombies                   = configPrefix + "ignore_zombie_processes"
+	awsECSFargatePidModeTaskContainerName = "aws-fargate-pause"
```
robertjli, line 49: IMO this isn't worth making a `const` for. You only use it once and it has very limited intended scope.

You also reference the value in a comment, so changing the value here could lead to discrepancy/confusion in the function below

pkg/process/checks/process.go
```diff
@@ -702,3 +715,20 @@ func isDisallowListed(cmdline []string, disallowList []*regexp.Regexp) bool {
...
+
+func isECSFargatePidModeSetToTask(totalProcs int, containers []*model.Container) bool {
+	if !fargate.IsFargateInstance() {
+		return false
+	}
+
+	// aws-fargate-pause container only exists when "pidMode"" is set to "task" on ecs fargate
+	pidModeTaskContainerExists := false
+	for _, c := range containers {
+		// TODO: consider revising this check since it doesn't seem very robust
```
robertjli, line 727: IMO this comment isn't helpful, do we have a plan to revisit or improve this? If not, then it seems like a TODO that will live in the codebase forever, and we should avoid that

releasenotes/notes/ecs-fargate-pidmode-misconfiguration-f7887b42ad48af54.yaml
```diff
@@ -0,0 +1,4 @@
+---
+enhancements:
+  - |
+    ECS Linux Fargate: added one-time warning when "pidMode" is not set to "task"
```
robertjli, line 4: Probably add in "for process collection", it's possible that customers are running the agent without process collection.

pkg/process/checks/process.go
gengnamstyle: interesting got it. will update

pkg/process/checks/process.go
gengnamstyle: will update

pkg/process/checks/process.go
gengnamstyle: gotcha. I assumed the visibility would be good, but I do agree it has limited scope. It does feel iffy to have hardcoded values in the functions they are used, but I will update.

pkg/process/checks/process.go
gengnamstyle: agreed. Removed!

releasenotes/notes/ecs-fargate-pidmode-misconfiguration-f7887b42ad48af54.yaml
gengnamstyle: good catch

pkg/process/checks/process.go
wiyu: Also, vars and const are pkg scoped, so consider how much visibility is truly needed. It's generally better to limit scope as it's easier to increase scope, but harder to reduce it.

wiyu: > It does feel iffy to have hardcoded values in the functions they are used, but I will update. 

So you don't have to "hardcode the value" per-se. 
You can just do something like this, which gives the variable name some context.
```go
func myFunc() {
	awsECSFargatePidModeTaskContainerName := "aws-fargate-pause"
    foo(awsECSFargatePidModeTaskContainerName)
    ...
}
```

pkg/process/checks/process.go
```diff
@@ -702,3 +715,20 @@ func isDisallowListed(cmdline []string, disallowList []*regexp.Regexp) bool {
...
 	return false
 }
+
+func isECSFargatePidModeSetToTask(totalProcs int, containers []*model.Container) bool {
+	if !fargate.IsFargateInstance() {
+		return false
+	}
+
+	// aws-fargate-pause container only exists when "pidMode"" is set to "task" on ecs fargate
+	pidModeTaskContainerExists := false
```
wiyu, line 725: [minor] You can just use the var style declaration as the default value is already false.
```suggestion
	var pidModeTaskContainerExists bool
```

pkg/process/checks/process.go
```diff
@@ -702,3 +715,20 @@ func isDisallowListed(cmdline []string, disallowList []*regexp.Regexp) bool {
...
+	for _, c := range containers {
+		// TODO: consider revising this check since it doesn't seem very robust
+		if slices.Contains(c.Tags, fmt.Sprintf("%s:%s", tags.EcsContainerName, awsECSFargatePidModeTaskContainerName)) {
```
wiyu, line 728: Let's lift this out into a variable
```go
fmt.Sprintf("%s:%s", tags.EcsContainerName, awsECSFargatePidModeTaskContainerName)
```
Something like:
```go
fargatePauseContainerTag := fmt.Sprintf("%s:%s", tags.EcsContainerName, awsECSFargatePidModeTaskContainerName)
```

pkg/process/checks/process.go
gengnamstyle: I feel explicitly setting the variable to false is more clear. However, should we prefer the var declaration?

pkg/process/checks/process.go
gengnamstyle: will update

pkg/process/checks/process.go
wiyu: Check out the [zero values of variables](https://go.dev/tour/basics/12). 

From the [uber style guide](https://github.com/uber-go/guide/blob/master/style.md#local-variable-declarations)

Usually when initializing the variable with `:=` means you want something that is not the zero value.

pkg/process/checks/process.go
```diff
@@ -702,3 +715,20 @@ func isDisallowListed(cmdline []string, disallowList []*regexp.Regexp) bool {
...
+			pidModeTaskContainerExists = true
+			break
+		}
+	}
+	return pidModeTaskContainerExists && totalProcs > 0
```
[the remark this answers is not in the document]
gengnamstyle: I'm checking the collected processes count as an extra level of confidence that the user has correctly set `pidMode` to `task`. The case the container exists and we have 0 processes is weird. We could potentially remove the process check here, but in the chunking code https://github.com/DataDog/datadog-agent/blob/8add1e009ce760dee03d589e88899a9459839860/pkg/process/checks/chunking.go#L16-L16

We also do a processes = 0 check which covers the ecs case as well. Perhaps @daniel-taf has some thoughts?

I don't think we can return early since, it's possible to have `pidMode` set correctly but still have process collection disabled. So technically, pidMode can be correct and I think returning false would be slightly misleading.

pkg/process/checks/process.go
```diff
@@ -702,3 +715,20 @@ func isDisallowListed(cmdline []string, disallowList []*regexp.Regexp) bool {
 	}
 	return false
 }
+
+func isECSFargatePidModeSetToTask(totalProcs int, containers []*model.Container) bool {
+	if !fargate.IsFargateInstance() {
+		return false
+	}
```
daniel-taf, line 722: Won't this result in the warning being logged in non-ECS fargate environments?

gengnamstyle: yup you're right thanks for the catch.

pkg/process/checks/process.go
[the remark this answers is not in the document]
daniel-taf: To me, it seems like the total process count doesn't really have any significance here. Checking for the container seems certain enough to me at the moment. Also, if process collection is disabled, this code won't ever run so we don't need to worry about that case. Additionally, the chunking code is per container so the process count has more relevancy there.

pkg/process/checks/process.go
gengnamstyle: I definitely would agree with defining empty slices with the var format, but I don't feel fully convinced about the bool since I think it reads clearer. I did find some other references in the repo owned by us, but I can change it if we can decide as a team this is the standard we want to continue with. Thoughts?

pkg/process/checks/process.go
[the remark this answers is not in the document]
wiyu: So in the `totalProcs` value you're passing it is the count of all [processes](https://github.com/DataDog/datadog-agent/blob/d84840abe8e635729662e77919935c312084fbcb/pkg/process/checks/process.go#L427-L452) collected. The code in the chunking you referenced can potentially see a subset of those processes which is the processes per container. 

As for returning early if `totalProcs <= 0` then you'll always return false since this is an `&&` condition. Why even bother checking for the container if that is the case, because you'll return false?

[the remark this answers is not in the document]
gengnamstyle: gotcha. Okay I'll remove the references to using process count here.

pkg/process/checks/process.go
```diff
@@ -296,6 +304,14 @@ func (p *ProcessCheck) run(groupID int32, collectRealTime bool) (RunResult, erro
 	procsByCtr := fmtProcesses(p.scrubber, p.disallowList, procs, p.lastProcs, pidToCid, cpuTimes[0], p.lastCPUTime, p.lastRun, p.lookupIdProbe, p.ignoreZombieProcesses, p.serviceExtractor, pidToGPUTags)
 	messages, totalProcs, totalContainers := createProcCtrMessages(p.hostInfo, procsByCtr, containers, p.maxBatchSize, p.maxBatchBytes, groupID, p.networkID, collectorProcHints)
 
+	// warn customer if "pidMode":"task" is not set in ecs linux fargate
+	p.warnOnceECSLinuxFargateMisconfig.Do(func() {
+		// pidMode is currently not supported on windows, see docs: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#task_definition_pidmode
+		if fargate.IsFargateInstance() && runtime.GOOS == "linux" && !isECSLinuxFargatePidModeSetToTask(containers) {
+			log.Warn(`Process collection is likely misconfigured. Please ensure your task definition has "pidMode":"task" set. See https://docs.datadoghq.com/integrations/ecs_fargate/?tab=webui#process-collection for more information.`)
+		}
```
wiyu, line 312: let's avoid checking the OS directly like this. We use BuildFlags for OS specific features.

check out 
- https://github.com/DataDog/datadog-agent/blob/5be2e4941eb41823740ee118aaa16c531186921e/pkg/process/checks/process_nix.go
- https://github.com/DataDog/datadog-agent/blob/ba08dfcc0bb4b569fc5a59e65652b46eeeba42a4/pkg/process/checks/process_windows.go

releasenotes/notes/ecs-fargate-pidmode-misconfiguration-f7887b42ad48af54.yaml
```diff
@@ -0,0 +1,4 @@
+---
+enhancements:
+  - |
+    ECS Linux Fargate: added one-time warning when "pidMode" is not set to "task" for process collection
```
domalessi, line 4:
```suggestion
    ECS Fargate (Linux): added a one-time warning when `pidMode` is not set to `task`, which may prevent process collection.
```

domalessi (approved):
Left a minor suggestion that you can take or leave!

releasenotes/notes/ecs-fargate-pidmode-misconfiguration-f7887b42ad48af54.yaml
gengnamstyle: thanks :)

pkg/process/checks/process.go
gengnamstyle: move check to another function in separate build flag files

pkg/process/checks/process_nix.go
```diff
@@ -89,3 +92,27 @@ func calculatePct(deltaProc, deltaTime, numCPU float64) float32 {
 	pct = math.Max(pct, 0.0)
 	return float32(pct)
 }
+
+// warnECSLinuxFargateMisconfig pidMode is currently not supported on fargate windows, see docs: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#task_definition_pidmode
+func warnECSFargateMisconfig(containers []*model.Container) {
+	if fargate.IsFargateInstance() && !isECSFargatePidModeSetToTask(containers) {
```
wiyu, line 98: You can just remove this fargate check since your other function checks it.

pkg/process/checks/process_nix.go
```diff
@@ -89,3 +92,27 @@ func calculatePct(deltaProc, deltaTime, numCPU float64) float32 {
...
+		log.Warn(`Process collection is likely misconfigured. Please ensure your task definition has "pidMode":"task" set. See https://docs.datadoghq.com/integrations/ecs_fargate/?tab=webui#process-collection for more information.`)
+	}
+}
+
+// isECSLinuxFargatePidModeSetToTask checks if pidMode is set to task in task (which will prevent process collection) definition using specific container existence
+func isECSFargatePidModeSetToTask(containers []*model.Container) bool {
+	if !fargate.IsFargateInstance() {
+		return false
+	}
```
daniel-taf, line 107: This condition will never happen since you added a check on line 98. Go uses short circuit evaluation for conditionals so if `fargate.IsFargateInstance()` is false, `isECSFargatePidModeSetToTask()` will never be called.

pkg/process/checks/process_nix.go
```diff
@@ -89,3 +92,27 @@ func calculatePct(deltaProc, deltaTime, numCPU float64) float32 {
 	pct = math.Max(pct, 0.0)
 	return float32(pct)
 }
+
+// warnECSLinuxFargateMisconfig pidMode is currently not supported on fargate windows, see docs: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#task_definition_pidmode
```
wiyu, line 96: [minor] update the comment as the method name changed

pkg/process/checks/process_nix.go
```diff
@@ -89,3 +92,27 @@ func calculatePct(deltaProc, deltaTime, numCPU float64) float32 {
...
+func warnECSFargateMisconfig(containers []*model.Container) {
+	if fargate.IsFargateInstance() && !isECSFargatePidModeSetToTask(containers) {
+		log.Warn(`Process collection is likely misconfigured. Please ensure your task definition has "pidMode":"task" set. See https://docs.datadoghq.com/integrations/ecs_fargate/?tab=webui#process-collection for more information.`)
+	}
+}
+
+// isECSLinuxFargatePidModeSetToTask checks if pidMode is set to task in task (which will prevent process collection) definition using specific container existence
```
wiyu, line 103: same here regarding method name

pkg/process/checks/process_nix.go
```diff
@@ -89,3 +92,27 @@ func calculatePct(deltaProc, deltaTime, numCPU float64) float32 {
 	pct = math.Max(pct, 0.0)
 	return float32(pct)
 }
+
+// warnECSLinuxFargateMisconfig pidMode is currently not supported on fargate windows, see docs: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#task_definition_pidmode
+func warnECSFargateMisconfig(containers []*model.Container) {
+	if fargate.IsFargateInstance() && !isECSFargatePidModeSetToTask(containers) {
+		log.Warn(`Process collection is likely misconfigured. Please ensure your task definition has "pidMode":"task" set. See https://docs.datadoghq.com/integrations/ecs_fargate/?tab=webui#process-collection for more information.`)
```
robertjli, line 99:
```suggestion
		log.Warn(`Process collection may be misconfigured. Please ensure your task definition has "pidMode":"task" set. See https://docs.datadoghq.com/integrations/ecs_fargate#process-collection for more information.`)
```

Suggestion: removed `?tab=webui` query parameter because it's not relevant here, and rephrased to be "softer" (IMO at least, feel free to reject)

domalessi (approved):

pkg/process/checks/process_nix.go
robertjli, line 103: Two comments:

> checks if pidMode is set to task in task (which will prevent process collection) definition

I think the "(which will ...)" is in the wrong place. Also as I read it, it seems to say if pidMode is set to task then process collection will be prevented, which is the opposite?

> using specific container existence

Suggestion: Function comments should say what the function does but not how it does it, i.e. avoid implementation details. This allows us to change the implementation in the future without breaking any implied contract.

pkg/process/checks/process_nix.go
wiyu: He did this for unit testing purposes, so we should just keep the fargate check here so it's covered by the tests.

pkg/process/checks/process_nix.go
```diff
@@ -89,3 +92,27 @@ func calculatePct(deltaProc, deltaTime, numCPU float64) float32 {
...
+	}
+}
+
+// isECSLinuxFargatePidModeSetToTask checks if pidMode is set to task in task (which will prevent process collection) definition using specific container existence
+func isECSFargatePidModeSetToTask(containers []*model.Container) bool {
+	if !fargate.IsFargateInstance() {
+		return false
+	}
+
+	// aws-fargate-pause container only exists when "pidMode"" is set to "task" on ecs fargate
```
robertjli, line 109:
```suggestion
	// aws-fargate-pause container only exists when "pidMode" is set to "task" on ecs fargate
```

pkg/process/checks/process_nix.go
daniel-taf: Why not call `warnECSFargateMisconfig()` instead?

wiyu: There's no return. 😆

pkg/process/checks/process_nix.go
robertjli: Actually can't remove this because then non-Fargate envs will emit the log

pkg/process/checks/process_nix.go
daniel-taf: Fair 😆 but it does seem odd to have a check that will never happen just for testing purposes because then why are we testing it?

daniel-taf: ultimately, I think we can just remove that aspect of the test since it doesnt seem to add anything then.

robertjli: Discussed offline, @gengnamstyle will remove the `isFargateInstance` check from this method and update the function comment to be clear that we are assuming the env is Fargate if the method is called.

pkg/process/checks/process_windows.go
```diff
@@ -67,3 +65,11 @@ func calculatePct(deltaProc, deltaTime, numCPU float64) float32 {
 	overalPct = math.Max(overalPct, 0.0)
 	return float32(overalPct)
 }
+
+//nolint:revive // ecs fargate checks only support linux
```
wiyu, line 69: Just noticed this, but don't use `//nolint:revive` that's for the linter to ignore the method. We have it around are code base because the core team migrated linters at some point which triggered a lot of failures so they put those comments so individual teams would fix the issues on their own time.

gengnamstyle: renamed to `_` 👍

pkg/process/checks/process_nix.go
gengnamstyle: done!

pkg/process/checks/process_nix.go
gengnamstyle: updated comment to: 
`isECSFargatePidModeSetToTask checks if pidMode is set to task in task definition to allow for process collection and assumes the method is called in a fargate environment`

pkg/process/checks/process_nix.go
gengnamstyle: fixed

pkg/process/checks/process.go
```diff
@@ -128,6 +129,9 @@ type ProcessCheck struct {
 	statsd         statsd.ClientInterface
 
 	gpuSubscriber gpusubscriber.Component
+
+	// used for one-off warning
```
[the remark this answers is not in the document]
gengnamstyle: removed

pkg/process/checks/process_nix.go
gengnamstyle: I initially had it in the case that someone else wanted to use the function which, admittedly, is low.  Therefore, I believe the function had the responsibility to check.

robertjli (approved):

wiyu (approved):

gengnamstyle:
error from merge
image
will retry

PR #36412, opened 2025-04-23
[CWS] process cache entry refcount fixes

safchain:
### What does this PR do?

Add checks for the cache entry release.


YoannGh (approved):
Do you think it makes sense to move the increment on the `cacheSize` counter variable from `insertEntry` to the `NewProcessCacheEntry` method instead? Should this be done in this PR?

PR #36416, opened 2025-04-23
[ebpfless] Fix handling of retransmits in TIME_WAIT state

pimlu:
### What does this PR do?
This PR adds a `recentlyClosed` LRU map which keeps track of recently closed connection tuples. Previously, if a retransmit occurred _after_ a connection closed, it would get re-opened as a preexisting connection. This PR makes TCPProcessor consider such connections to remain closed.

### Motivation
There have been reports of a spurious error in ebpfless, which I couldn't reproduce myself, but I fixed in [\[NPM-4334\] Silence misleading error in ebpfless](https://github.com/DataDog/datadog-agent/pull/35419). However, I found the reason I never saw the error myself was because it only occurred if a TCP connection expired via timeout (_not_ via normal connection closing). This got me suspicious that I was missing an edge case with TCP. It turns out there is one such edge case -- I wasn't handling retransmits after connection close properly.

New test for this case, `TestIncomingRetransmitAfterClose` should pass in the TCP processor suite:
```
go test -tags linux,linux_bpf,npm,process,test github.com/DataDog/datadog-agent/pkg/network/tracer/connection/ebpfless
```

Ebpfless test suite should still pass:
```
sudo -E go test -race -tags=linux,linux_bpf,npm,process,test ./pkg/network/tracer -v --run TestTracerSuite/eBPFless
```

### Additional Notes

I also added a couple log warnings that I also want to get backported into 7.66.

### DIFF
--- a/pkg/network/tracer/connection/ebpfless/tcp_processor.go
+++ b/pkg/network/tracer/connection/ebpfless/tcp_processor.go
@@ -15,6 +15,7 @@ import (
 	"golang.org/x/sys/unix"
 
 	"github.com/google/gopacket/layers"
+	"github.com/hashicorp/golang-lru/v2/expirable"
 
 	"github.com/DataDog/datadog-agent/pkg/network"
 	"github.com/DataDog/datadog-agent/pkg/network/config"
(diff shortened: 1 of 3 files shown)


pkg/network/tracer/connection/ebpfless/tcp_processor.go
```diff
@@ -77,18 +78,26 @@ type TCPProcessor struct {
 	pendingConns map[PCAPTuple]*connectionState
 	// establishedConns contains connections with tcpState == connStatEstablished
 	establishedConns map[PCAPTuple]*connectionState
+	// recentlyClosed is analogous to the TIME_WAIT state after a connection closes. we use this to
+	// suppress "pre-existing connection" logic on retransmits after a connection closes
+	recentlyClosed *expirable.LRU[PCAPTuple, struct{}]
```
hmahmood, line 83: `struct{}` -> `any`

hmahmood (approved):

pkg/network/tracer/connection/ebpfless/tcp_processor.go
pimlu: Wait, why not `struct{}`? It saves 16 bytes per entry no?

hmahmood: Indeed it does.

PR #36417, opened 2025-04-23
[clusteragent/autoscaling] Add CLI command to dump autoscaling store

jennchenn:
### What does this PR do?

Adds a new CLI command to allow users to list all installed `DatadogPodAutoscalers`. Format looks something like:
<details>
<summary>Sample output</summary>

```
=== PodAutoscaler default/better-cyclic-burner-server ===
----------- PodAutoscaler ID -----------
default/better-cyclic-burner-server

----------- PodAutoscaler Meta -----------
Creation Timestamp: 2025-04-10 17:13:41 +0000 UTC
Generation: 2
Settings Timestamp: 0001-01-01 00:00:00 +0000 UTC

----------- PodAutoscaler Spec -----------
Target Ref: {Deployment better-cyclic-burner-server apps/v1}
Owner: Local
Apply Policy Mode: Apply

----------- PodAutoscaler Constraints -----------
Min Replicas: 1
Max Replicas: 5

----------- PodAutoscaler Objectives -----------
Objective Type: PodResource
Resource Name: cpu
Utilization: 85

----------- PodAutoscaler Scaling Values -----------
[Horizontal]
Horizontal Error: <nil>
Source: Autoscaling
Timestamp: 2025-04-23 20:42:00 +0000 UTC
Replicas: 6
--------------------------------
[Vertical]
Vertical Error: <nil>
Source: Autoscaling
Timestamp: 2025-04-23 12:00:00 +0000 UTC
ResourcesHash: 1415926535
Container Name: server
Container Resources: map[memory:32Mi]
Container Limits: map[]
--------------------------------
Error: <nil>

----------- PodAutoscaler Main Scaling Values -----------
[Horizontal]
Horizontal Error: <nil>
Source:
Timestamp: 0001-01-01 00:00:00 +0000 UTC
Replicas: 0
--------------------------------
[Vertical]
Vertical Error: <nil>
Source:
Timestamp: 0001-01-01 00:00:00 +0000 UTC
ResourcesHash:
--------------------------------
Error: <nil>

----------- PodAutoscaler Fallback Scaling Values -----------
[Horizontal]
Horizontal Error: <nil>
Source:
Timestamp: 0001-01-01 00:00:00 +0000 UTC
Replicas: 0
--------------------------------
[Vertical]
Vertical Error: <nil>
Source:
Timestamp: 0001-01-01 00:00:00 +0000 UTC
ResourcesHash:
--------------------------------
Error: <nil>

----------- PodAutoscaler Status -----------
Current Replicas: 5
--------------------------------
Horizontal Last Action: Timestamp: 2025-04-23 19:56:35 +0000 UTC
From Replicas: 4
To Replicas: 5
Recommended Replicas: 6
Limited Reason: desired replica count limited to 5 (originally 6) due to max replicas constraint
--------------------------------
Vertical Last Action Error: no scaling values available


===
```
</details>

### Motivation

We want to give users more visibility so they are able to debug issues with autoscaling

1. Deploy changes; exec into the cluster agent and run `agent autoscaler-list`
2. Verify that output is well-formatted, matches the number of active `DatadogPodAutoscalers` in your environment

Output is quite verbose, so may be difficult to parse especially with a large number of autoscalers. Adding a `verbose` flag may be useful to allow users to control how much information they would like to see. Also open to removing any fields that may not be useful/of interest.


zhuminyi:
LGTM (container-platform). Todo list:
1. We also need to convert GetAutoscalingStore to component framework once it is done. 
2. Add optional subcommand argument to support dumping local failover store entities.

zhuminyi (approved):

pkg/cli/subcommands/autoscalerlist/command.go
```diff
@@ -0,0 +1,128 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-present Datadog, Inc.
+
+// Package autoscalerlist implements 'agent autoscaler-list'.
+package autoscalerlist
```
dustmop, line 7: Do you think it makes sense to add this folder to CODEOWNERS so that your team owns it? CLI commands default to agent-configuration's ownership, but specific commands can be owned by teams, for example `/pkg/cli/subcommands/clusterchecks` is owned by `container-platform`.

dustmop (approved):
LGTM, with one question concerning code ownership.

pkg/cli/subcommands/autoscalerlist/command.go
jennchenn: I added this line here which should update the ownership going forward (but let me know if it looks misconfigured!): https://github.com/DataDog/datadog-agent/pull/36417/files#diff-3d36a1bf06148bc6ba1ce2ed3d19de32ea708d955fed212c0d27c536f0bd4da7R371

dustmop: LGTM!

pkg/clusteragent/autoscaling/workload/model/pod_autoscaler_string.go
```diff
@@ -0,0 +1,417 @@
...
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	datadoghqcommon "github.com/DataDog/datadog-operator/api/datadoghq/common"
+	"github.com/DataDog/datadog-operator/api/datadoghq/v1alpha2"
+)
+
+func (p *PodAutoscalerInternal) String(verbose bool) string {
```
vboulineau, line 24: I know it's the pattern we have in WLM, but given the smaller scope, I am wondering if it's worth having an auto printer instead? Like https://github.com/sanity-io/litter

pkg/clusteragent/autoscaling/workload/model/pod_autoscaler_string.go
```diff
@@ -0,0 +1,417 @@
...
+func toResourceQuantityMap(resourceList corev1.ResourceList) map[string]string {
+	quantities := make(map[string]string)
+	for name, quantity := range resourceList {
+		quantities[string(name)] = quantity.String()
+	}
+	return quantities
+}
+
+// MarshalJSON implements the json.Marshaler interface for PodAutoscalerInternal
+func (p *PodAutoscalerInternal) MarshalJSON() ([]byte, error) {
```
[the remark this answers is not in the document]
jennchenn: I added the custom JSON unmarshalling/marshaling logic since the `PodAutoscalerInternal` fields are not exported.  I get a linter error if I try to add JSON annotations as well (because the fields are unexported). Would changing the fields to be exported be a better option?

pkg/clusteragent/autoscaling/workload/model/pod_autoscaler_string.go
```diff
@@ -0,0 +1,417 @@
...
+func formatVerticalAction(action *datadoghqcommon.DatadogPodAutoscalerVerticalAction) string {
+	var sb strings.Builder
+	_, _ = fmt.Fprintln(&sb, "Timestamp:", action.Time)
+	_, _ = fmt.Fprintln(&sb, "Version:", action.Version)
+	_, _ = fmt.Fprintln(&sb, "Type:", action.Type)
+	return strings.TrimRight(sb.String(), "\n")
+}
+
+// TODO: move to common util, shared with orchestrator
+func toResourceQuantityMap(resourceList corev1.ResourceList) map[string]string {
```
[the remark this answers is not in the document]
jennchenn: Without converting it I think the format would look like this:
```
_, _ = fmt.Fprintln(&sb, "Requests Min Allowed:", container.Requests.MinAllowed)
...

-RequestsMinAllowed:map[cpu:100mmemory:100Mi]
+RequestsMinAllowed:map[cpu:{{100-3}{<nil>}100mDecimalSI}memory:{{1048576000}{<nil>}100MiBinarySI}]
```
or were you thinking of another way to call it? 

I reused the logic from https://github.com/DataDog/datadog-agent/blob/94bd6ccf65fdc9f91bd5e1f86faa737b0b8f444b/pkg/collector/corechecks/cluster/orchestrator/transformers/k8s/cluster.go#L51-L57 (still needs to be moved to common util if we want to keep it).

pkg/clusteragent/autoscaling/workload/model/pod_autoscaler_string.go
jennchenn: Started playing around with it, having some trouble getting it formatted nicely (e.g. for timestamps), will keep looking into it!

pkg/clusteragent/autoscaling/workload/dump.go
```diff
@@ -0,0 +1,78 @@
...
+
+var defaultDumper storeDumper
+
+// InitDumper initializes the default dumper with a given store
+func InitDumper(store *store) {
+	defaultDumper.store = store
+}
+
+// AutoscalersInfo is used to dump the autoscaling store content
+type AutoscalersInfo struct {
```
vboulineau, line 33: nit: `type` and member functions should go together and free functions together

pkg/clusteragent/autoscaling/workload/model/pod_autoscaler_string.go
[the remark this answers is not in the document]
vboulineau: What I mean is that you don't have to build a `map[string]string` and pass it to Fprintf, you can write directly the string format you want

zhuminyi:
Do we have plan to support this information dump in dca [flare](https://github.com/DataDog/datadog-agent/blob/main/pkg/flare/clusteragent/archive_dca.go)?

pkg/clusteragent/autoscaling/workload/model/pod_autoscaler_string.go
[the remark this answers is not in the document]
vboulineau: I prefer to keep the private fields for now. You can add `json` to `ScalingValues` to simplify the wrappers and merge as-is.

pkg/clusteragent/autoscaling/workload/controller_test.go
```diff
@@ -45,6 +45,8 @@ const testMaxAutoscalerObjects int = 2
 
 func newFixture(t *testing.T, testTime time.Time) *fixture {
 	store := autoscaling.NewStore[model.PodAutoscalerInternal]()
+	InitDumper(store)
```
vboulineau, line 48: Why do we need it? Except in `dump_test.go` I don't expect us to use the variable?

jennchenn: `dump_test.go` is currently reusing this code to generate the fixture, but I can create a new test helper function to keep them separate

vboulineau: You can use the fixture and call `InitDumper()` in your test

vboulineau (approved):

PR #36419, opened 2025-04-24
Avoid copying paths that are not needed

alopezz:
### What does this PR do?

Try to exclude folders that are not necessary to build the Agent from the copy that Omnibus does of the repo.

### Motivation

See if there's anything to gain from copying less.


alopezz:
Closing for now as this barely shaves off 10 seconds and still has issues to solve (for example this fails because go.work refers to modules that I wanted to skip).

PR #36441, opened 2025-04-24
Remove channel buffers for callback queue in process monitor

grantseltzer:
### Motivation

When starting to deploy the dynamic instrumentation module I observed hard to track down behavior where services that we want to instrument were not being picked up by the process monitor. Through manual investigation @tylfin and I concluded that a potential cause was the buffered channels used in the process monitor.

When `SubscribeExec()` gets called, the callback that's passed gets add to a collection of callbacks. Then, every time an exec occurs, the collection of callbacks is iterated over and each one is passed to a worker pool where individual go routines will call the callback on the exec'd process' pid. However, on startup the process monitor iterates over the systems running process tree and treats each running process as if it were an exec, passing the callbacks for each on the channel. Depending on how many subscribers, this can overwhelm the channels buffer, filling it up.

When the channel buffer is filled, current logic just drops the callback from being sent on it instead of waiting for the buffer to be emptied. I'm not sure why this was done this way. Regardless, every time a new subscriber is added to the process monitor, the need to increase the buffer size will come again. Therefore I don't see a reason to keep this unbuffered and let the workers complete all callbacks.

Deploying custom image to staging cluster to test manually. Need input from other teams that rely on process monitor to ensure it works as expected for them.

### DIFF
--- a/pkg/process/monitor/process_monitor.go
+++ b/pkg/process/monitor/process_monitor.go
@@ -25,13 +25,6 @@ import (
 	"github.com/DataDog/datadog-agent/pkg/util/log"
 )
 
-const (
-	// The size of the process events queue of netlink.
-	processMonitorEventQueueSize = 2048
-	// The size of the callbacks queue for pending tasks.
-	pendingCallbacksQueueSize = 5000
-)
-
 var (
 	processMonitor = &ProcessMonitor{
 		// Must initialize the sets, as we can register callbacks prior to calling Initialize.
@@ -171,15 +164,7 @@ func (pm *ProcessMonitor) handleProcessExec(pid uint32) {
 
 	for callback := range pm.processExecCallbacks {
 		temporaryCallback := callback
-		select {
-		case pm.callbackRunner <- func() { (*temporaryCallback)(pid) }:
-			continue
-		default:
-			pm.tel.processExecChannelIsFull.Add(1)
-			if log.ShouldLog(log.DebugLvl) && pm.oversizedLogLimit.ShouldLog() {
-				log.Debug("can't send exec callback to callbackRunner, channel is full")
-			}
-		}
+		pm.callbackRunner <- func() { (*temporaryCallback)(pid) }
 	}
 }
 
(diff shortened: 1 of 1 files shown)


tylfin (approved):

guyarb (changes requested):
Why do you use process monitor? you should be using process-event-data-streams
A couple of months ago I wrote your team in slack about migrating to PEDS, as we know the process-monitor is unstable on one hand, and it cannot be used by multiple modules (usm + dynamic instrumentation) at the same time
your team promised to make the migration

The reason for the buffering came from process-churn that caused the entire code to be slow and blocked

grantseltzer:
> Why do you use process monitor? you should be using process-event-data-streams A couple of months ago I wrote your team in slack about migrating to PEDS, as we know the process-monitor is unstable on one hand, and it cannot be used by multiple modules (usm + dynamic instrumentation) at the same time your team promised to make the migration
> 
> The reason for the buffering came from process-churn that caused the entire code to be slow and blocked

The plan remains to use PEDS, but it will take some rearchitecting of code. As far as I understand this affects PEDS as well though. Does USM operate alright with dropped callbacks?

guyarb:
> > Why do you use process monitor? you should be using process-event-data-streams A couple of months ago I wrote your team in slack about migrating to PEDS, as we know the process-monitor is unstable on one hand, and it cannot be used by multiple modules (usm + dynamic instrumentation) at the same time your team promised to make the migration
> > The reason for the buffering came from process-churn that caused the entire code to be slow and blocked
> 
> The plan remains to use PEDS, but it will take some rearchitecting of code. As far as I understand this affects PEDS as well though. Does USM operate alright with dropped callbacks?

You shouldn't be using process monitor, but PEDS directly (unlike USM today), so the code you've changed shouldn't impact peds
USM has periodic process scanning to fill the gaps
The current proposal is problematic as it makes the entire code base synchronous.
The issue is not we event being dropped, the issue is with yours/USM callbacks being slow, so the callback runnners are too slow to process

grantseltzer:
Closing in favor of https://github.com/DataDog/datadog-agent/pull/36775

PR #36442, opened 2025-04-24
fix(manual_gate_threshold_update): wrong rules on threshold update

Pythyu:


spencergilbert (approved):

Pythyu:
Force merging to avoid spam due to wrong gitlab job rules

PR #36444, opened 2025-04-24
[CWS] Fix `proc_cache_t` marshalling

YoannGh:
### What does this PR do?

Fix `proc_cache_t` marshalling that was corrupted for processes with an empty cgroup path key due to an incorrect write offset.


pkg/security/secl/model/marshallers_linux.go
```diff
@@ -212,3 +211,10 @@ func (p *PathKey) MarshalBinary() ([]byte, error) {
 
 	return buff, nil
 }
+
+// MarshalBinaryAllowEmpty returns the binary representation of a path key with could be empty
```
spikat, line 215: which *

spikat (approved):
just a small nit but LGTM

safchain (approved):

PR #36448, opened 2025-04-24
[NDINT-226] Versa Appliance, Controller, and Director Metadata

ken-schneider:
### What does this PR do?
Retrieves all Appliance, Controller, and Director metadata from Versa HTTP Basic Authentication. Even if we switch to OAuth, the endpoints and structure of the responses should be the same.

This PR removes the previous auth logic since the logic around basic auth is minimal.

### Motivation
Add NDM support for Versa

Run against our own Versa cluster using the agent.


pkg/collector/corechecks/network-devices/cisco-sdwan/client/client_test.go
```diff
@@ -3,6 +3,8 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2024-present Datadog, Inc.
 
+//go:build test
+
```
ken-schneider, line 7: This file was missing the the test tag

pkg/collector/corechecks/network-devices/versa/versa.go
```diff
@@ -67,26 +67,58 @@ func (v *VersaCheck) Run() error {
...
-	client, err := client.NewClient(v.config.DirectorEndpoint, v.config.AnalyticsEndpoint, v.config.Username, v.config.Password, v.config.UseHTTP)
+	c, err := client.NewClient(v.config.DirectorEndpoint, v.config.Username, v.config.Password, v.config.UseHTTP)
 	if err != nil {
 		return fmt.Errorf("error creating Versa client: %w", err)
 	}
 
-	appliances, err := client.GetAppliancesLite()
+	// Get all the organizations, so we can get the appliances
+	// for each. We should only need to get the top level organizations
+	// which act as tenants. Appliances of child tennants should be part
```
zoedt, line 77: very weird but i guess tenant and organization is interchangeable based on their [docs](https://docs.versa-networks.com/Management_and_Orchestration/Versa_Director/Configuration/Configure_Multitenancy#:~:text=a%20distinction%20is%20made%20between,that%20belong%20to%20their%20organization.)!

zoedt: [future question for customers] maybe as part of the "toggle" list, if customers only want to support a specific list of orgs/tenants?

pkg/collector/corechecks/network-devices/versa/payload/devices.go
```diff
@@ -77,43 +102,28 @@ func GetDevicesStatus(devices []client.ApplianceLite) map[string]float64 {
...
+	id := buildDeviceID(namespace, device.IPAddress)
 
 	return devicemetadata.DeviceMetadata{
 		ID:           id,
 		IPAddress:    device.IPAddress,
 		Vendor:       "versa",
 		Name:         device.Name,
 		Tags:         append(buildApplianceDeviceTags(namespace, device), "source:versa"),
-		IDTags:       []string{"device_namespace:" + namespace, "system_ip:" + device.IPAddress},
+		IDTags:       []string{"device_namespace:" + namespace, "device_ip:" + device.IPAddress},
```
zoedt, line 114: should this be using the `GetApplianceDevicesTags` method?

zoedt: or rather the map from it?

zoedt: 🤦 sorry i read this wrong for IDTags and not `Tags`

ken-schneider: Good catch, I missed this. I'm actually not doing anything with the appliance tags right now other than adding the ones we create on our end.

I've added the `device.ApplianceTags` to the `buildApplianceDeviceTags` function.

`GetApplianceDevicesTags` will drive the tags for device metrics like it does for Cisco SD-WAN and the value of that will include the appliance tags now too

pkg/collector/corechecks/network-devices/versa/versa.go
ken-schneider: I've added this as well, there's now an `includedTenants` and `excludedTenants` in the config which will filter down the appliances we pull in and that should propagate to other calls.

zoedt (approved):
lgtm!!

PR #36450, opened 2025-04-24
[OTAGENT-389] fix a race condition in serializer exporter

songy23:
### What does this PR do?
Move `forwarder.Start()` in serializer exporter to be synchronous

### Motivation
Fix a race condition when  serializer exporter is up and receiving metrics, the forwarder has not yet started.

N/A this will be tested in OSS datadog exporter: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/39601 & https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/39669


mackjmr (approved):
LGTM, a regression test would be good, but I'm assuming it would need to be in contrib ?

PR #36455, opened 2025-04-25
service discovery: Get working directory from proc

vitkyrka:
### What does this PR do?

Currently, the code only gets the working directory from the PWD
environment variable, but this does not necessarily need to be present
nor up-to-date, so try to get the working directory from the cwd symlink
in proc. Unfortunately, we can't know which source is the correct one
for the command line which we're looking at (since the working directory
can be changed at anytime during execution of the processes), so we try
both when trying to resolve relative paths.

### Motivation

https://datadoghq.atlassian.net/browse/DSCVR-113

Unit tests added.  Also test the generation on a process on staging where the
name was not being generated correctly due to a mismatch in the working
directory.


pkg/collector/corechecks/servicediscovery/usm/nodejs.go
```diff
@@ -9,11 +9,12 @@ import (
 	"io"
 	"io/fs"
 	"path"
-	"path/filepath"
 	"strings"
 
 	"github.com/valyala/fastjson"
 
+	"path/filepath"
+
```
Yumasi, line 17: Is this diff intentional?

Yumasi (approved):

pkg/collector/corechecks/servicediscovery/usm/nodejs.go
vitkyrka: No. Fixed.

PR #36458, opened 2025-04-25
[EBPF] attacher: add unit test to ensure attacher works from a container

gjulianm:
### What does this PR do?

This PR adds testing code that allows us to see the behavior of the uprobe attacher when running inside of a container. It adds code to execute the `fmapper` test util inside of a Docker container, runs tests with the target running inside and outside of a container in both cases. 

This is done by creating a "standalone attacher" binary, that just runs the attachment of simple probes to any binaries, and then can be queried through HTTP to get the list of attached probes.

### Motivation

Ensure we have unit test coverage for the case of attaching to processes when system-probe runs in a container.

The changes themselves are tests which should pass.


pkg/network/usm/sharedlibraries/testutil/testutil.go
```diff
@@ -27,19 +27,29 @@ import (
...
-	require.NoError(t, err, "failed to create pattern scanner")
+	patternScanner := BuildFmapperScanner(t)
 	cmd.Stdout = patternScanner
 	cmd.Stderr = patternScanner
 
 	require.NoError(t, cmd.Start())
 	log.Infof("exec prog=%s, paths=%v | PID = %d", programExecutable, paths, cmd.Process.Pid)
 
 	t.Cleanup(func() {
+		patternScanner.PrintLogs(t)
```
guyarb, line 51: I'm afraid it will spam the test log

pkg/util/testutil/docker/run.go
```diff
@@ -82,6 +85,11 @@ func run(t testing.TB, cfg LifecycleConfig) (context.Context, error) {
 	}
 	//register cleanup function to kill the instances upon finishing the test
 	t.Cleanup(func() {
+		if t.Failed() {
+			// Get the logs of the container
+			logs, _ := exec.Command("docker", "logs", cfg.Name()).Output()
+			t.Logf("outputting container logs for %s\n%s", cfg.Name(), logs)
+		}
```
guyarb, line 92: why don't we use - `cfg.PatternScanner().PrintLogs(t)`?

pkg/util/testutil/docker/config.go
```diff
@@ -20,6 +20,9 @@ const (
 
 	// DefaultRetries is the default number of retries for starting a container/s.
 	DefaultRetries = 3
+
+	// MinimalDockerImage is the minimal docker image, just used for running a binary
+	MinimalDockerImage = "alpine:3.20.3"
```
guyarb, line 25: do we have that in the ami-builder repo?
also, do we have that in the `invoke` command that reports docker images used by our tests?

guyarb (approved):

pkg/ebpf/uprobes/testutil/standalone_attacher/standalone_attacher.go
```diff
@@ -0,0 +1,70 @@
...
+	"os/signal"
+	"syscall"
+	"testing"
+
+	"github.com/DataDog/datadog-agent/pkg/ebpf/uprobes"
+)
+
+var configName = flag.String("config", "", "config name")
+
+func main() {
```
brycekahle, line 28: Couldn't you write a normal test and then compile it to a binary using `go test -c`?

tasks/kmt.py
```diff
@@ -1236,17 +1317,27 @@ def test(
...
+            d.run_cmd(
+                ctx,
+                f"/opt/micro-vm-init.sh {' '.join(args)}",
+                verbose=verbose,
+                allow_fail=True,
+            )
 
             info(f"[+] Showing summary of results for {d}")
-            d.run_cmd(ctx, "/opt/testing-tools/test-json-review", verbose=verbose, allow_fail=True)
+            d.run_cmd(
```
brycekahle, line 1335: There are a lot of formatting changes for lines that did not change. It makes it hard to see what actually changed.

pkg/ebpf/uprobes/testutil_attacher_runner.go
brycekahle, line 1: We've got this file, `testutil.go` in this package, and a `testutil` package. Should everything be in the `testutil` package?

pkg/network/usm/sharedlibraries/testutil/testutil.go
gjulianm: Removed

pkg/util/testutil/docker/run.go
gjulianm: The scanner only stores logs until the start pattern is found, so we miss anything after that. I added a comment to clarify here.

pkg/util/testutil/docker/config.go
gjulianm: It's in [ami-builder](https://github.com/DataDog/ami-builder/blob/84de57480aa04d390a92a0db5f475d5f2fc8e0d0/qemu-images/docker/docker.ls#L14), I added it to the inv command.

pkg/network/usm/sharedlibraries/testutil/testutil.go
gjulianm: Removed

pkg/ebpf/uprobes/testutil/standalone_attacher/standalone_attacher.go
gjulianm: I tried to do that, but then I had to change how the helper binaries are built to include this specific command line, and also I was afraid that it would be included as a regular test in other jobs.

tasks/kmt.py
gjulianm: I didn't notice these, apparently the pre-commit hook modified all of this. I'm making a separate PR formatting the KMT files so that this PR gets cleaned up.

gjulianm: Formatting fixed now, I had something misconfigured apparently.

brycekahle (approved):

PR #36464, opened 2025-04-25
[AGENTRUN-269] Remove core-agent binary from heroku build

pgimalac:
### What does this PR do?
Remove the core-agent binary from the Heroku build.

### Motivation
Reduce Heroku package size.

The package also contains a bundled core-agent + process-agent binary, we can ship this one only.
Both binaries used to have a size difference so it made sense to have both and pick the one which should be used, but now they have pretty much the same size, so we can just drop the "core-agent only" and always used the bundled one.

### Additional Notes

This has been discussed with Heroku agent owners.
https://github.com/DataDog/heroku-buildpack-datadog/blob/master/bin/compile will be updated before the next release of the Heroku agent.


arapulido (approved):
LGTM! Please, ping me when there is a nightly available, so we can test as soon as possible, in case anything goes wrong. Thanks for working on this!

chouquette:
Isn't that equivalent to https://github.com/DataDog/heroku-buildpack-datadog/blob/master/bin/compile#L252-L254?
Is there any observable change from this PR?

pgimalac:
@chouquette The code in `compile` runs after downloading the package, but the package itself still contains both binaries.
The only change in this PR is that the package will now contain a single binary, so the package is 60MiB smaller (uncompressed).
It won't change much at runtime (since the binary is already removed on install), but will reduce network cost and disk usage at the very beginning.

The code in `compile` will be updated once this PR is merged.

chouquette (approved):

PR #36467, opened 2025-04-25
[autodiscovery/configresolver] Use allow-list when resolving env vars

davidor:
### What does this PR do?

This PR adds a new configuration option, `ad_allowed_env_vars`, which allows users to restrict which environment variables can be resolved in Autodiscovery check configurations using the `%%env_SOME_ENV%%` notation. When set, only the environment variables listed will be resolved.

I added unit tests. I also checked the following cases in a local cluster:
- Used an env in the allow-list => env resolved
- Used an env not in the allow-list => env not resolved
- Defined an empty allow-list => env resolved
- No allow-list => env resolved


pkg/config/config_template.yaml
```diff
@@ -2843,6 +2843,17 @@ api_key:
 #
 # ad_config_poll_interval: 10
 
+## @param ad_allowed_env_vars - list of strings - optional
+## @env DD_AD_ALLOWED_ENV_VARS - list of strings - optional
+## The list of environment variables that are allowed to be resolved in check
+## configurations.
+## The default is to allow all environment variables. If the list is set, but
+## empty, no environment variables will be resolved.
```
jhgilbert, line 2851:
```suggestion
## empty, no environment variables are resolved.
```

releasenotes/notes/ad-envs-allow-list-7df22a6dbcfdc009.yaml
```diff
@@ -0,0 +1,14 @@
...
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+enhancements:
+  - |
+    Added a new configuration option, ``ad_allowed_env_vars``, which allows
+    users to restrict which environment variables can be resolved in
+    Autodiscovery check configurations. When set, only the environment variables
+    listed will be resolved.
```
jhgilbert, line 14:
```suggestion
    listed are resolved.
```

jhgilbert (approved):
Approved with a very minor suggestion for style. Thank you!

comp/core/autodiscovery/configresolver/configresolver.go
```diff
@@ -622,6 +624,14 @@ func getEnvvar(_ context.Context, envVar string, svc listeners.Service) (string,
 		}
 		return "", fmt.Errorf("envvar name is missing")
 	}
+
+	if !allowEnvVar(envVar) {
```
adel121, line 628: `allowEnvVar` results in a call to the config component in order to get the list of allowed env vars. 

The obtained list doesn't change across subsequent calls to `allowEnvVar`. Would it make sense to extract the call to the config component so that we get the slice only once and store it at the level of the configresolver object instead of recalculating it each time in the function call?

adel121 (approved):
Approved with a minor suggestion.

pkg/config/setup/config.go
```diff
@@ -1236,6 +1236,7 @@ func autoconfig(config pkgconfigmodel.Setup) {
 	config.BindEnvAndSetDefault("container_exclude_logs", []string{})
 	config.BindEnvAndSetDefault("container_exclude_stopped_age", DefaultAuditorTTL-1) // in hours
 	config.BindEnvAndSetDefault("ad_config_poll_interval", int64(10))                 // in seconds
+	config.BindEnv("ad_allowed_env_vars")
```
hush-hush, line 1239: Please add a default to any new settings. a `nil` slices would to the trick. Types are required is we want to have a schema at some point.

pkg/config/config_template.yaml
hush-hush, line 2851: Having different behavior depending on the value type is usually not a great idea since it's not obvious to newcomers. Also, once loaded in the code the diff in Go between an empty slice and a nil slice is error prone.

I think a dedicated `ad_disable_env_vars` would be more explicit and less error prone. It would also allow us to add a default to both in the code. Not having default means the config can't know what to expect for a setting which makes validation impossible.

hush-hush (changes requested):

pkg/config/config_template.yaml
davidor: I think this should be an allow-list instead of a block-list. The goal is to let users control what envs can be used. It’s easier to list what you want to allow than to block everything you don’t need.

The main question is what to do with an empty allow-list. I think we can set a default like `[]string{}`, as you mentioned in the other comment. We should only use this setting if the user has set it (we can check with `IsConfigured`). If it’s not configured, we keep the current behavior. If the list is set but empty, it means no environment variables should be resolved. Since this is an allow-list, an empty list should block everything if set explicitly.

hush-hush: I agree that an allow-list is better. I'm saying that an empty list to disable env vars resolution is not a intuitive behavior for customers and is error prone. We had such behavior in the past and it's very easy when working with YAML to comment a few line and create en empty list. We also had different YAML lis produce different results when loading an empty YAML list.

I would ignore a nil or empty list for `ad_allowed_env_vars` and add another dedicated setting to disable env var resolution.

hush-hush (approved):

PR #36469, opened 2025-04-25
Support selecting systemd units with regular expressions

iliakur:
### What does this PR do?
Add a new config field `unit_regexes` that can be a list of regular expressions to select Systemd units.

### Motivation
Customer requests.

Unit tests.


pkg/collector/corechecks/systemd/systemd.go
```diff
@@ -515,6 +519,11 @@ func (c *SystemdCheck) isMonitored(unitName string) bool {
 			return true
 		}
 	}
+	for _, pattern := range c.unitPatterns {
+		if pattern.MatchString(unitName) {
+			return true
+		}
+	}
```
AAraKKe, line 526: I was thinking what would be the expected behavior if the user configures a unit name that contradicts a provided regexp? For example, using the ones in your test below, what happens if the user provides `unit1.service` as unit name but also a regexp like `^[^u].*`? If we keep it like this, names take preference, maybe we should write this down somewhere?

Also, I was thinking about an example with negative look ahead and I realized that go does not support look arounds which made me think: should we add these somewhere in the documentation or the example config so the users knows that the regexp to be used are the ones supported by the [RE2 syntax](https://github.com/google/re2/wiki/syntax)? It seems a bit more proactive to let the user knows before the agent fails to load the regexp in the config.

iliakur: both very good points, thanks!

I realized I missed adding a test for both `unit_names` and `unit_regexes` specified. I'll do that and will defo include your specific example in there. Will also add it to docs.

As for regex features, lemme read up real quick on what golang `regexp` does and I'll see what makes sense to add to the docs.

NouemanKHAL: https://pkg.go.dev/regexp#pkg-overview

NouemanKHAL: As @AAraKKe mentioned, we need to document the order of precedence between `unit_names` and `unit_regexes`, same as you did in the design document.
It would be good to have a different test function for the different cases in the design document.

jhgilbert (approved):

pkg/collector/corechecks/systemd/systemd.go
```diff
@@ -546,8 +555,17 @@ func (c *SystemdCheck) Configure(senderManager sender.SenderManager, integration
...
-	if len(c.config.instance.UnitNames) == 0 {
-		return fmt.Errorf("instance config `unit_names` must not be empty")
+	if len(c.config.instance.UnitNames) == 0 && len(c.config.instance.UnitRegexes) == 0 {
+		return fmt.Errorf("please set either `unit_names` or `unit_regexes` in the instance config")
+	}
+
+	for _, regex := range c.config.instance.UnitRegexes {
+		pattern, err := regexp.Compile(regex)
+		if err != nil {
+			return errors.Wrapf(err, "cannot compile regular expression %q to monitor systemd units", regex)
```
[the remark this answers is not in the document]
iliakur: Good thing to check this! afaict `%q` does precisely that: it's the same as `%s` but wraps the argument in quotes.

rahulkaukuntla (approved):

AAraKKe (approved):
Thanks for the update @iliakur !

pkg/collector/corechecks/systemd/systemd_test.go
```diff
@@ -661,6 +675,114 @@ unit_names:
...
+
+			// Then
+			mockSender.AssertCalled(t, "ServiceCheck", canConnectServiceCheck, servicecheck.ServiceCheckOK, "", []string(nil), mock.Anything)
+			for _, unitName := range test.monitoredUnits {
+				tags := []string{fmt.Sprintf("unit:%s", unitName)}
+				metrics := unitsToMetrics[unitName]
+				for _, metric := range metrics {
+					mockSender.AssertCalled(t, "Gauge", metric, mock.Anything, "", tags)
+				}
+			}
```
NouemanKHAL, line 781: I think that we need to assert that we don't report metrics for other units that are not in `test.monitoredUnits` otherwise, this test would pass if we're monitoring everything as well.

iliakur: sg, I pushed a tweak to do this, lmk what you think!

NouemanKHAL (approved):

pkg/collector/corechecks/systemd/systemd_test.go
NouemanKHAL: Nice!

PR #36474, opened 2025-04-25
Ignore empty env vars when building the configuration

hush-hush:
### What does this PR do?

Ignore empty env vars when building the configuration.

### Motivation

This mirror the behavior from viper in nodetreemodel.

Running the CI tests is enough

### DIFF
--- a/pkg/config/nodetreemodel/config_test.go
+++ b/pkg/config/nodetreemodel/config_test.go
@@ -387,6 +387,21 @@ func TestEnvVarMultipleSettings(t *testing.T) {
 	assert.Equal(t, 0, cfg.GetInt("c"))
 }
 
+func TestEmptyEnvVarSettings(t *testing.T) {
+	cfg := NewNodeTreeConfig("test", "TEST", nil)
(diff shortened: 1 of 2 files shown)


dustmop (approved):
LGTM

pkg/config/nodetreemodel/config_test.go
```diff
@@ -386,6 +386,19 @@ func TestEnvVarMultipleSettings(t *testing.T) {
 	assert.Equal(t, 0, cfg.GetInt("c"))
 }
 
+func TestEmptyEnvVarSettings(t *testing.T) {
+	cfg := NewNodeTreeConfig("test", "TEST", nil)
+	cfg.SetDefault("a", -1)
+	cfg.BindEnv("a")
+	t.Setenv("TEST_A", "")
```
dustmop, line 393: I think it's worth adding a comment here explaining exactly what this test is demonstrating. The semantics are not completely obvious otherwise.
```suggestion
	// This empty string is ignored, so the default value of -1 will be returned by GetInt
	t.Setenv("TEST_A", "")
```

PR #36475, opened 2025-04-25
Add VolumeMount to all containers even if volume already exists

florianmutter:
### What does this PR do?

Depending on the order of webhooks it can happen that datadog adds the volume in the first step. If another webhook (e.g. istio sidecar injection) adds another container to the pod datadog will inject all env variables but the volume mount will not be added.

This PR fixes this.


adel121:
Hello @florianmutter 

Thanks for opening this PR. 

The fix you shared looks good makes sense. However, we have to modify more details in the function to ensure that it still informs its callers when something has been injected (even if we injected a volume mount, while not injecting a volume).

We also need to update the unit tests. 

We've open [another PR](https://github.com/DataDog/datadog-agent/pull/37091) including the idea used in your fix along with some additional refactors.

Feel free to check it and leave any comments or questions and we would be happy to answer and discuss.

I think you can close the current PR if you have no other questions.

Thank you again for being proactive on fixing this issue!

PR #36478, opened 2025-04-25
[Draft] Capture live KM dump when driver service in E2E is stuck

alexn-dd:
### What does this PR do?
This PR captures a live KM dump if during the Windows E2E start-stop tests, a driver service has not transitioned to its expected state (e.g. stuck trying to stop).  The live dump is generated under the existing dump folder.  The command used is originally intended for storage clusters and therefore the dump will be under a "localhost" subdirectory.

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1331 
This is needed to investigate intermittent E2E test failures, where the driver service does not stop on shutdown.

With a test npm driver that 1) holds a thread with busy spinning, 2) yields to the scheduler, verified that the dump does contain the stuck thread.
-TBD: verify that the dump is accessible.

The live KM dumps can be quite large: 500MB - 1GB, larger depending on the VM memory.

### DIFF
--- a/test/new-e2e/tests/windows/service-test/startstop_test.go
+++ b/test/new-e2e/tests/windows/service-test/startstop_test.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/DataDog/test-infra-definitions/components/datadog/agentparams"
 
+	fs "github.com/DataDog/datadog-agent/pkg/util/filesystem"
 	"github.com/DataDog/datadog-agent/pkg/util/testutil/flake"
 	"github.com/DataDog/datadog-agent/test/new-e2e/pkg/components"
 	"github.com/DataDog/datadog-agent/test/new-e2e/pkg/e2e"
@@ -61,9 +62,6 @@ var securityAgentConfig string
 
 //go:embed fixtures/security-agent-disabled.yaml
 var securityAgentConfigDisabled string
-
-// driver services that start up and shut down with system probe.
-var driverServices := [2]string{"ddnpm", "ddprocmon"}
 
 // TestServiceBehaviorAgentCommandNoFIM tests the service behavior when controlled by Agent commands
 func TestNoFIMServiceBehaviorAgentCommand(t *testing.T) {
@@ -664,13 +662,11 @@ func (s *baseStartStopSuite) assertAllServicesState(expected string) {
 }
 
 func (s *baseStartStopSuite) assertServiceState(expected string, serviceName string) {
-	stateAttained := false
 	host := s.Env().RemoteHost
 
 	s.Assert().EventuallyWithT(func(c *assert.CollectT) {
 		status, err := windowsCommon.GetServiceStatus(host, serviceName)
 		if !assert.NoError(c, err) {
-			stateAttained = true
 			return
 		}
 		if !assert.Equal(c, expected, status, "%s should be %s", serviceName, expected) {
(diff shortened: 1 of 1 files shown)


test/new-e2e/tests/windows/service-test/startstop_test.go
```diff
@@ -62,6 +62,9 @@ var securityAgentConfig string
 //go:embed fixtures/security-agent-disabled.yaml
 var securityAgentConfigDisabled string
 
+// driver services that start up and shut down with system probe.
+var driverServices := [2]string{"ddnpm", "ddprocmon"}
```
clarkb7, line 66: we have a list already in `getInstalledKernelServices`

test/new-e2e/tests/windows/service-test/startstop_test.go
```diff
@@ -776,3 +788,17 @@ func (s *baseStartStopSuite) sendHostMemoryMetrics(host *components.RemoteHost)
 		s.T().Logf("posted memory metrics")
 	}
 }
+
+// captureLiveKernelDump sends a commnad to the host to create a live kernel dump.
+func (s *baseStartStopSuite) captureLiveKernelDump(dumpDir string) {
+	getSubsystemCmd := `$ss = Get-CimInstance -ClassName MSFT_StorageSubSystem -Namespace Root\Microsoft\Windows\Storage`
+    createLiveDumpCmd := fmt.Sprintf(`Invoke-CimMethod -InputObject $ss -MethodName "GetDiagnosticInfo" -Arguments @{DestinationPath="%s"; IncludeLiveDump=$true}"`, dumpDir)
```
clarkb7, line 795: it looks like this command puts the dump file in a subdirectory, `C:\dumps\localhost\LiveDump.dmp`, but the [dump collection isn't recursive](https://github.com/DataDog/datadog-agent/blob/0145beac3b8895657bd2bd2a726fdbe119a2556d/test/new-e2e/tests/windows/common/crashdump.go#L107-L110), so it won't be detected.

It seems like the command fails when run twice because the files already exist. So maybe it'd be better to run the command with a temp directory target and then copy only the dump file into the provided `dumpDir`.

test/new-e2e/tests/windows/service-test/startstop_test.go
```diff
@@ -661,16 +664,25 @@ func (s *baseStartStopSuite) assertAllServicesState(expected string) {
...
 		if !assert.Equal(c, expected, status, "%s should be %s", serviceName, expected) {
 			s.T().Logf("waiting for %s to be %s, status %s", serviceName, expected, status)
 		}
 	}, 1*time.Minute, 1*time.Second, "%s should be in the expected state", serviceName)
+
+	if !stateAttained && slices.Contains(driverServices. serviceName)
+		// if a driver service failed to get to the expected state, capture a kernel dump for debugging.
+		s.T().Logf("Capturing live kernel dump due to %s not in %s state", serviceName, expected)
+		captureLiveKernelDump(s.dumpFolder)
+	}
```
clarkb7, line 685: I think the above `stateAttained = true` is in the wrong spot, `GetServiceStatus` is just getting the status not comparing it, that happens with the assertion below. I think `GetServiceStatus` would only fail if the service didn't exist.

you can do without manually tracking `stateAttained` if you use `s.T().Failed()` instead.

```suggestion
	if s.T().Failed() && slices.Contains(driverServices. serviceName)
		// if a driver service failed to get to the expected state, capture a kernel dump for debugging.
		s.T().Logf("Capturing live kernel dump due to %s not in %s state", serviceName, expected)
		captureLiveKernelDump(s.dumpFolder)
	}
```

PR #36479, opened 2025-04-25
Use debug level for the ddurl resolver log when there is an override of DD_SITE

jeremy-hanna:
### What does this PR do?

This is a very noisy log internally and is only useful for debugging purposes.


GustavoCaso (approved):

jeremy-hanna:
preference from @DataDog/agent-configuration is to log this only once and not reduce the log level, closing the PR

hush-hush (approved):

PR #36494, opened 2025-04-25
[CWS] refactor security profiles remote storage forwarder

paulcacheux:
### What does this PR do?

This PR refactors the security profiles remote storage (and forwarder) with the main goal being removing the dependency on the security profile data structure (and associated imports) from the security agent.

The main redesign is that we now have a common remote storage forwarder, that can either forward to an abstracted interface. This allows the forwarder to be used both to send the profiles to the security agent through gRPC and to the backend directly.

Data is interchanged in serialized form, allowing the security agent to directly forward blobs and not have to re-create a security profile in memory. This allows the security-agent to not depend on actual CWS code at all.


spikat (approved):
noice :)

PR #36507, opened 2025-04-28
Make rc_kubernetes_deploy wait for docker_trigger_internal-full

kacper-murzyn:
### What does this PR do?

This PR makes r`c_kubernetes_deploy` job to need `docker_trigger_internal-full` before it starts. As in most cases we deploy the `-full` flavor image now we need to make sure that those images are available before we start the deployment.

### Motivation

Improve the Agent release process.

### DIFF
--- a/.gitlab/internal_kubernetes_deploy/rc_kubernetes_deploy.yml
+++ b/.gitlab/internal_kubernetes_deploy/rc_kubernetes_deploy.yml
@@ -11,8 +11,11 @@ rc_kubernetes_deploy:
       when: manual
   needs:
     - job: docker_trigger_internal
+      artifacts: false
     - job: docker_trigger_cluster_agent_internal
+      artifacts: false
     - job: docker_trigger_internal-full
+      artifacts: false
     - job: k8s-e2e-main # Currently only require container Argo workflow
       artifacts: false
       optional: true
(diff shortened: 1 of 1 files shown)


.gitlab/internal_kubernetes_deploy/rc_kubernetes_deploy.yml
```diff
@@ -11,9 +11,8 @@ rc_kubernetes_deploy:
       when: manual
   needs:
     - job: docker_trigger_internal
-      artifacts: false
```
chouquette, line 14: Is there an artifact generated by this job? I thought it only pushed images

chouquette (commented):
LGTM as far as the new job is concerned, but I'd have expected `artifacts: false` to stay and be added for the new jobs dependencies

.gitlab/internal_kubernetes_deploy/rc_kubernetes_deploy.yml
kacper-murzyn: There are none, that's why I removed this line. It was there from the past, copy-paste probably.

chouquette: But then it will now attempt to download artifacts, even though there are none.
It doesn't change anything functionally, but I find the intent to be clearer with the explicit `artifacts: false`

chouquette (approved):
LGTM, the `artifacts` open question isn't blocking

.gitlab/internal_kubernetes_deploy/rc_kubernetes_deploy.yml
kacper-murzyn: Oh, I actually could have made a mistake here. I though that skipping that field in the needs section is like setting it to false. But I reread the documentation and it's actually opposite. I will change it back. Sorry for the confusion.

PR #36523, opened 2025-04-28
[USMON-1074] kafka kernel telemetry report api versions

guymor4:
### What does this PR do?
Adds kafka telemetry metric `datadog.system_probe.usm.kafka.classified_hits` by `operation` and `api_version` so we can have better visibility into the kafka versions running


pkg/network/ebpf/c/protocols/kafka/defs.h
```diff
@@ -50,4 +50,7 @@
 // The size of each kafka telemetry topic name bucket
 #define KAFKA_TELEMETRY_TOPIC_NAME_BUCKET_SIZE 10
 
+// The maximum version we'll support in telemetry.
+#define KAFKA_TELEMETRY_API_VERSIONS_NUM_OF_BUCKETS 30
```
guyarb, line 54: not sure why `30`, and how is that the max we can support
as we support fetch (v17) and produce (v12), why do we need 30 entries?

pkg/network/ebpf/c/protocols/kafka/kafka-parsing.h
```diff
@@ -1804,4 +1808,22 @@ static __always_inline void update_topic_name_size_telemetry(kafka_telemetry_t *
     __sync_fetch_and_add(&kafka_tel->topic_name_size_buckets[bucket_idx], 1);
 }
 
+// update_classified_api_version_hits_telemetry updates arrays keeping track of api versions hits in the telemetry
+static __always_inline void update_classified_api_version_hits_telemetry(kafka_telemetry_t *kafka_tel, s16 api_key, s16 api_version) {
+    __u8 bucket_idx = api_version - 1;
```
guyarb, line 1813: there's a type mismatch here between s16 and _u8
if you assume the api_version is positive and smaller than 255, then let's use __u8 all the way

pkg/network/ebpf/c/protocols/kafka/kafka-parsing.h
```diff
@@ -12,6 +12,7 @@ static __always_inline bool kafka_allow_packet(skb_info_t *skb_info);
 static __always_inline bool kafka_process(conn_tuple_t *tup, kafka_info_t *kafka, pktbuf_t pkt, kafka_telemetry_t *kafka_tel);
 static __always_inline bool kafka_process_response(void *ctx, conn_tuple_t *tup, kafka_info_t *kafka, pktbuf_t pkt, skb_info_t *skb_info);
 static __always_inline void update_topic_name_size_telemetry(kafka_telemetry_t *kafka_tel, __u64 size);
+static __always_inline void update_classified_api_version_hits_telemetry(kafka_telemetry_t *kafka_tel, s16 api_key, s16 api_version);
```
guyarb, line 15: as we care only for produce (api 0) and fetch (api 1), and the versions are 0-12 to produce, and 0-17 to consume, I'm think we should be using `__u8` for `api_key` and `api_version`

pkg/network/ebpf/c/protocols/kafka/kafka-parsing.h
```diff
@@ -1618,12 +1619,15 @@ static __always_inline bool kafka_process(conn_tuple_t *tup, kafka_info_t *kafka
     kafka_header.correlation_id = bpf_ntohl(kafka_header.correlation_id);
     kafka_header.client_id_size = bpf_ntohs(kafka_header.client_id_size);
 
-    log_debug("kafka: kafka_header.api_key: %d api_version: %d", kafka_header.api_key, kafka_header.api_version);
 
     if (!is_valid_kafka_request_header(&kafka_header)) {
         return false;
     }
```
guyarb, line 1625: wouldn't this condition drop any unsupport-by-decoding produce/fetch messages?

pkg/network/ebpf/c/protocols/kafka/kafka-parsing.h
```diff
@@ -1804,4 +1808,22 @@ static __always_inline void update_topic_name_size_telemetry(kafka_telemetry_t *
...
+
+    switch (api_key) {
+    case KAFKA_FETCH:
+        __sync_fetch_and_add(&kafka_tel->classified_fetch_api_version_hits[bucket_idx], 1);
+        break;
+    case KAFKA_PRODUCE:
+        __sync_fetch_and_add(&kafka_tel->classified_produce_api_version_hits[bucket_idx], 1);
+        break;
+    }
+}
```
guyarb, line 1827: you took the implementation from the topic_name telemetry, I think that's incorrect, or at least, needs to be modified per-fetch and produce to adjust the number of buckets to the correct size

pkg/network/ebpf/c/protocols/kafka/types.h
```diff
@@ -121,6 +121,10 @@ typedef struct {
     // The array topic_name_size_buckets maps a bucket index to the number of occurrences observed for topic name lengths
     __u64 topic_name_size_buckets[KAFKA_TELEMETRY_TOPIC_NAME_NUM_OF_BUCKETS];
     __u64 produce_no_required_acks;
+
+    // Arrays to keep track of the number of occurrences of each API version that were classified as kafka, before decoding
+    __u64 classified_fetch_api_version_hits[KAFKA_TELEMETRY_API_VERSIONS_NUM_OF_BUCKETS];
+    __u64 classified_produce_api_version_hits[KAFKA_TELEMETRY_API_VERSIONS_NUM_OF_BUCKETS];
```
guyarb, line 127: can reuse the already existing consts for max-<produce/fetch>-for-classification

pkg/network/ebpf/c/protocols/kafka/kafka-parsing.h
guymor4: It will, but the next PR will split classification and decoding, it's already tested and approved so I didn't want to move it to this PR

pkg/network/ebpf/c/protocols/kafka/kafka-parsing.h
guymor4: Not sure I got you, it's not really buckets it's just counters where index = api_verison so I'm just doing +1 to the correct counter (maybe I should rename bucket -> counter?)
But I changed the implementation to use index 0 for unknowns

pkg/network/ebpf/c/protocols/kafka/defs.h
```diff
@@ -50,7 +50,12 @@
...
 
-// The maximum version we'll support in telemetry.
-#define KAFKA_TELEMETRY_API_VERSIONS_NUM_OF_BUCKETS 30
+// The maximum version we'll support in telemetry, take the maximum supported version in classification
+// Adds 1 because it's zero based and we use index 0 to count unknown versions
+#if KAFKA_MAX_SUPPORTED_FETCH_REQUEST_API_VERSION > KAFKA_MAX_SUPPORTED_PRODUCE_REQUEST_API_VERSION
+#define KAFKA_TELEMETRY_API_VERSIONS_NUM_OF_BUCKETS (KAFKA_MAX_SUPPORTED_FETCH_REQUEST_API_VERSION + 1)
+#else
+#define KAFKA_TELEMETRY_API_VERSIONS_NUM_OF_BUCKETS (KAFKA_MAX_SUPPORTED_PRODUCE_REQUEST_API_VERSION + 1)
+#endif
```
guyarb, line 59: why having a single const for produce and fetch? it seems to actually complicating the code

pkg/network/ebpf/c/protocols/kafka/kafka-parsing.h
guyarb: i think you should split it per api key, as the shared const, makes it a little bit messy

pkg/network/ebpf/c/protocols/kafka/kafka-parsing.h
```diff
@@ -1620,6 +1622,16 @@ static __always_inline bool kafka_process(conn_tuple_t *tup, kafka_info_t *kafka
...
 
+    // Report api version hits telemetry
+    switch (kafka_header.api_key) {
+        case KAFKA_PRODUCE:
+            update_classified_produce_api_version_hits_telemetry(kafka_tel, kafka_header.api_version);
+            break;
+        case KAFKA_FETCH:
+            update_classified_fetch_api_version_hits_telemetry(kafka_tel, kafka_header.api_version);
+            break;
+    }
```
guyarb, line 1633: I'd move that after the condition in line 1635, to ensure we're dealing with valid kafka requests

pkg/network/ebpf/c/protocols/kafka/kafka-parsing.h
```diff
@@ -1804,4 +1816,26 @@ static __always_inline void update_topic_name_size_telemetry(kafka_telemetry_t *
     __sync_fetch_and_add(&kafka_tel->topic_name_size_buckets[bucket_idx], 1);
 }
 
+// update_classified_fetch_api_version_hits_telemetry updates the array keeping track of fetch api versions hits in the telemetry
+static __always_inline void update_classified_fetch_api_version_hits_telemetry(kafka_telemetry_t *kafka_tel, __u8 api_version) {
+    __u8 bucket_idx = api_version - 1;
```
guyarb, line 1821: fetch can have api_version of `0` (produce as well). so why do we need the `-1`?

pkg/network/ebpf/c/protocols/kafka/kafka-parsing.h
guymor4: The verifier complains about `update_classified_fetch_api_version_hits_telemetry` if I move it :( it shouldn't have any effect because it is being checked before getting here (IIUC)

guyarb: > it shouldn't have any effect because it is being checked before getting here (IIUC)

Where? the condition that should verify the traffic represents a valid kafka request is in the condition after this code

> The verifier complains about update_classified_fetch_api_version_hits_telemetry if I move it

What does it complain about? we should be able to address it

guymor4: I have no idea what's going on but it works now
and I meant `is_valid_kafka_request_header` is called both in the classification and again at the start of the decoding so I think the one in the decoding is obsolete

pkg/network/ebpf/c/protocols/kafka/types.h
```diff
@@ -121,6 +121,12 @@ typedef struct {
     // The array topic_name_size_buckets maps a bucket index to the number of occurrences observed for topic name lengths
     __u64 topic_name_size_buckets[KAFKA_TELEMETRY_TOPIC_NAME_NUM_OF_BUCKETS];
     __u64 produce_no_required_acks;
+
+    // Arrays to keep track of the number of occurrences of each API version that were classified as kafka, before decoding
+    // Not limited to supported api versions
+    // Make them have KAFKA_TELEMETRY_MAX_API_VERSION + 1 length because we want to support api version 0.
+    __u64 classified_fetch_api_version_hits[KAFKA_TELEMETRY_MAX_API_VERSION+1];
+    __u64 classified_produce_api_version_hits[KAFKA_TELEMETRY_MAX_API_VERSION+1];
```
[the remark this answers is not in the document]
guyarb: The reason we should be tied to the classification limits, is that we cannot guarantee we don't miss anything in the classification for "new and unknown" versions. Seeing produce v15, while there are only 12 known versions, can indicate a bug rather than providing useful information

guyarb (approved):

guyarb (approved):

amitslavin (approved):

PR #36524, opened 2025-04-28
use remove-experiment for rollback

clarkb7:
### What does this PR do?
Use the `remove-experiment` subcommand to rollback to stable version during an experiment. This updates the repository state (removes experiment symlink) during rollback, which makes `stop-experiment` received from the backend in the daemon [a no-op](https://github.com/DataDog/datadog-agent/blob/21fe278399eae9c60bb93256a4cf3e2d3fd9c6b1/pkg/fleet/installer/installer.go#L402-L410).

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1354

Avoid running Agent MSI twice during rollback (once in watchdog, once from daemon)

Update E2E tests to ensure daemon service isn't restarted when stop-experiment is called after a rollback (compares service PID).

relies on the repository state ([presence of experiment symlink](https://github.com/DataDog/datadog-agent/blob/21fe278399eae9c60bb93256a4cf3e2d3fd9c6b1/pkg/fleet/installer/installer.go#L402-L410)) for detection. It may be better to compare the installed MSI version with the stable MSI version, but its not clear how to do this since those versions don't map directly, MSI uses a 4 part number version. That's probably fine for releases and release candidates but might not work for dev testing versions like `7.66.0-devel`.


iadjivon (approved):
LGTM!

BaptisteFoy (approved):

julien-lebot (approved):

PR #36546, opened 2025-04-29
[ACIX-655] Add a client to download file from a artifact bucket in e2e tests

KevinFairise2:
### What does this PR do?

This PR add a client to download artifact from a s3 bucket in e2e tests. The goal of this bucket is to provide a solution where artifacts can be hosted and used in e2e tests.

The solution is not perfect as it requires installing the AWS cli on the remote host, but we do not really have other options since we want to use the instance role from the VM to download the artifact directly on the remote host.
The AWS CLI is installed from AWS infrastructure which should not bring too much flakiness. The installation of the CLI is lazy, if it is not used the CLI will never be installed, if needed it will be installed on the first fetch of an artifact

For now artifacts can only be uploaded on agent-e2e-tests-s3-bucket by an agent-qa admin. In the future each team should have its own subpath that they can use.

### Motivation

Provide a solution to teams that want to upload artifact to be used in e2e tests.


test/new-e2e/pkg/utils/e2e/client/host_cache.go
```diff
@@ -0,0 +1,179 @@
...
+	return err
+}
+
+func (c *windowsAWSCLI) check() bool {
+	_, err := c.sshExecutor.Execute("& \"c:\\Program Files\\Amazon\\AWSCLIV2\\aws.exe\" --version")
+	return err == nil
+}
+
+func (c *windowsAWSCLI) download(path string, destPath string) error {
+	_, err := c.sshExecutor.Execute(fmt.Sprintf("& \"c:\\Program Files\\Amazon\\AWSCLIV2\\aws.exe\" s3 cp %s %s", path, destPath))
```
CelianR, line 105: I suppose it's safer like this:
```suggestion
	_, err := c.sshExecutor.Execute(fmt.Sprintf("& \"c:\\Program Files\\Amazon\\AWSCLIV2\\aws.exe\" s3 cp \"%s\" \"%s\"", path, destPath))
```

test/new-e2e/pkg/utils/e2e/client/host_cache.go
```diff
@@ -0,0 +1,179 @@
...
+	return nil
+}
+
+func (c *unixAWSCLI) check() bool {
+	_, err := c.sshExecutor.Execute("aws --version")
+	return err == nil
+}
+
+func (c *unixAWSCLI) download(path string, destPath string) error {
+	_, err := c.sshExecutor.Execute(fmt.Sprintf("aws s3 cp %s %s", path, destPath))
```
CelianR, line 141:
```suggestion
	_, err := c.sshExecutor.Execute(fmt.Sprintf("aws s3 cp \"%s\" \"%s\"", path, destPath))
```

CelianR (approved):

PR #36547, opened 2025-04-29
BYOC Reporting

mackjmr:
### What does this PR do?
This PR reports metrics `datadog.otel.byoc` to `1` if the otel agent was built via BYOC, else it reports `0`. It is only applied in otel agent path, and is ignored in OTLP ingest path.


truthbk:
Closing this PR, as I believe you will be using the REDAPL method.

PR #36557, opened 2025-04-29
[APR-284] Create TCP MRF endpoints

soberpeach:
### What does this PR do?

Creates TCP MRF endpoints.

### Motivation

Customers that send logs over TCP will have the option to use `multi_region_failover` endpoints.

I tested by routing logs to a failover endpoint that I set up locally:

1. Create a logs intake TCP receiver that can accept incoming TCP connections.
I built the Docker image here https://github.com/DataDog/datadog-agent-benchmarks/tree/main/docker/logs-intake to with the command 
```bash
docker build . -t logs-intake:v1
```
to create the TCP logs intake. I ran it with the command: 
```bash
docker run -p 9999:9999 logs-intake:v1 -m "tcp" -d --bind ":9999"
```
2. Enable `multiregion_failover_logs` in `datadog.yaml` and configure the agent to sent logs
```yaml
multi_region_failover:
  enabled: true
  site: 127.0.0.1:9999
  api_key: api_key
  failover_logs: true
```

Caveat: in order to test this, I also had to temporarily change the output of this [this](https://github.com/DataDog/datadog-agent/blob/main/pkg/logs/client/tcp/connection_manager.go#L147) function to return a hard coded value of `127.0.0.1:9999`.
```go
// address returns the address of the server to send logs to.
func (cm *ConnectionManager) address() string {
	// return net.JoinHostPort(cm.endpoint.Host, strconv.Itoa(cm.endpoint.Port))
	return "127.0.0.1:9999"
}
```
The point of testing here was to see if the agent properly switched to the `multi_region_failover` endpoint.


maycmlee (approved):

comp/logs/agent/config/config.go
```diff
@@ -172,6 +172,36 @@ func buildTCPEndpoints(coreConfig pkgconfigmodel.Reader, logsConfig *LogsConfigK
...
+		if err != nil {
+			return nil, fmt.Errorf("cannot construct MRF endpoint: %s", err)
+		}
+
+		mrfHost, mrfPort, mrfUseSSL, err := parseAddressWithScheme(mrfURL, defaultNoSSL, parseAddressAsHost)
+		if err != nil {
+			return nil, fmt.Errorf("could not parse %s: %v", mrfURL, err)
+		}
+
+		e := NewEndpoint(coreConfig.GetString("multi_region_failover.api_key"), "multi_region_failover.api_key", mrfHost, mrfPort, mrfUseSSL)
```
gh123man, line 188: Does it make sense ot use `newTCPEndpoint` here? 
If so I think you can slim down the code below applying all the overrides.

comp/logs/agent/config/config.go
```diff
@@ -172,6 +172,36 @@ func buildTCPEndpoints(coreConfig pkgconfigmodel.Reader, logsConfig *LogsConfigK
...
+		e.CompressionLevel = main.CompressionLevel
+		e.BackoffBase = main.BackoffBase
+		e.BackoffMax = main.BackoffMax
+		e.BackoffFactor = main.BackoffFactor
+		e.RecoveryInterval = main.RecoveryInterval
+		e.RecoveryReset = main.RecoveryReset
+		e.Version = main.Version
+		e.TrackType = intakeTrackType
+		e.Protocol = intakeProtocol
+		e.Origin = intakeOrigin
```
gh123man, line 200: I don't think TrackType, protocol, or Origin are used in TCP mode. 
If this is the case, you can omit these.

comp/logs/agent/config/config.go
soberpeach: Ah yes, that would make sense.

soberpeach: After looking at it, I'm actually not sure it would make that much sense. The `newTCPEndpoint()` function seems to take a `logsConfigKeys` object as it's object, however the MRF endpoints don't use the `logs_config` preface so I don't think they wouldn't be available in the this object. The additional TCP endpoints don't use this function either: https://github.com/DataDog/datadog-agent/blob/main/comp/logs/agent/config/endpoints.go#L149C18-L149C19

comp/logs/agent/config/config.go
soberpeach: Sounds good. They're included in [`loadTCPAdditionalEndpoints()`](https://github.com/DataDog/datadog-agent/blob/main/comp/logs/agent/config/endpoints.go#L149), is that also an anomaly?

gh123man: They are copied from the main endpoint, but the main endpoint never sets them. So I don't think they are ever actually used.

comp/logs/agent/config/config.go
gh123man: FWIW `logsConfigKeys` can be configured wiht a prefix, so it's not just for `logs_config` configs. 
But copying it from `main` like you already are is fine. Was just trying to think of ways to reuse code.

maycmlee (approved):

comp/logs/agent/config/config.go
soberpeach: Removed!

comp/logs/agent/config/config.go
```diff
@@ -234,6 +233,35 @@ func buildTCPEndpoints(coreConfig pkgconfigmodel.Reader, logsConfig *LogsConfigK
...
 	additionals := loadTCPAdditionalEndpoints(main, logsConfig)
+
+	// Add in the MRF endpoint if MRF is enabled.
+	if coreConfig.GetBool("multi_region_failover.enabled") {
+		mrfURL, err := pkgconfigutils.GetMRFLogsEndpoint(coreConfig, tcpEndpointPrefix)
+		if err != nil {
+			return nil, fmt.Errorf("cannot construct MRF endpoint: %s", err)
+		}
+
+		mrfHost, mrfPort, mrfPathPrefix, mrfUseSSL, err := parseAddressWithScheme(mrfURL, defaultNoSSL, parseAddressAsHost)
```
gh123man: Should we be using `parseAddress` here? it looks like `parseAddressWithScheme` is only used for HTTP endpoints.

gh123man (approved):
Non blocking comment. 
Otherwise LGTM

comp/logs/agent/config/config.go
soberpeach: Ah good point, changing it to only use `parseAddress()`

soberpeach: Changed!

PR #36566, opened 2025-04-30
[autodiscovery] Delete unnecessary funcs from interface

davidor:
### What does this PR do?

This PR removes 2 unnecessary functions from the autodiscovery interface:
- `MapOverLoadedConfigs`, which is only used in a unit test.
- `LoadedConfigs`, which is a duplicate of the existing `GetAllConfigs` function.

CI.

### DIFF
--- a/comp/core/autodiscovery/autodiscoveryimpl/autoconfig_test.go
+++ b/comp/core/autodiscovery/autodiscoveryimpl/autoconfig_test.go
@@ -399,11 +399,10 @@ func TestResolveTemplate(t *testing.T) {
 }
 
 func countLoadedConfigs(ac *AutoConfig) int {
-	count := -1 // -1 would indicate f was not called
-	ac.MapOverLoadedConfigs(func(loadedConfigs map[string]integration.Config) {
-		count = len(loadedConfigs)
-	})
-	return count
+	if ac == nil || ac.store == nil {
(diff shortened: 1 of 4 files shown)


comp/core/autodiscovery/autodiscoveryimpl/autoconfig_test.go
```diff
@@ -399,11 +399,10 @@ func TestResolveTemplate(t *testing.T) {
 }
 
 func countLoadedConfigs(ac *AutoConfig) int {
-	count := -1 // -1 would indicate f was not called
-	ac.MapOverLoadedConfigs(func(loadedConfigs map[string]integration.Config) {
-		count = len(loadedConfigs)
-	})
-	return count
+	if ac == nil || ac.store == nil {
+		return 0
```
adel121, line 403: With the previous implementation I assume it was possible to return `-1`. After your change, I think it can only return non-negative values. 

maybe we should return -1 here? 

Am I missing something?

davidor: I don’t think the previous implementation ever returned -1, because the condition mentioned in the comment (`would indicate f was not called`) couldn’t occur.

adel121 (approved):

PR #36578, opened 2025-04-30
[autodiscovery/configresolver] Make NoServiceError constructor private

davidor:
### What does this PR do?

This PR is a small cleanup in the autodiscovery code.

The `NoServiceError` constructor is made private. The struct needs to be public so other packages can check for the error, but it should only be created internally.

CI


gabedos (approved):
lgtm

PR #36584, opened 2025-04-30
Add multiline JSON aggregation support to auto multiline detection

gh123man:
### What does this PR do?

Adds support for processing pretty printed/multi line JSON into single line JSON. 

ex: 

```json
{
  "foo": "bar",
  "abc":  123
}
```

becomes

```json
{"foo":"bar","abc":123}
```

This allows Datadog to treat pretty printed multi-line json a a structured log (automatically parsing it's attributes) 

This PR is roughly broken into two piece to make it easier to review: 
1. Introduce the new JSON validator and aggregator (and tests): https://github.com/DataDog/datadog-agent/pull/36584/commits/e0d656c94c858578d43e97671c3aad2c6182da86
2. Integrate it with auto multiline detection: https://github.com/DataDog/datadog-agent/pull/36584/commits/8700206e47c97102d5bb3df0b69e3a6428063aa9
- Also moves the buffer timeout logic into the aggregator so It can be shared. 

### Motivation

Historically, recombining multi-line JSON was extremely challenging to do correctly using `multi_line` regex rules. 
This change introduces a new incremental JSON validator, and a JSON aggregator build on top of it. 

Together these two components allow detecting, processing, and reformatting JSON in stream. 

- Each segment of JSON is buffered
- If the next segment is valid, the buffer grows
- when a JSON object ends, and the JSON is valid, the buffer is flushed as a single line of JSON (removing all whitespace and newlines)
- If a segment represents invalid JSON, or breaks previously valid but incomplete JSON, the buffer is flushed as the original individual log events. 

#### Happy path
1. Enable `logs_config.auto_multi_line_detection`
4. emit some logs with mixed pretty printed JSON. 
```
2024-08-13 17:21:34 INFO Main main Some log line 10
2024-08-13 17:21:34 INFO Main main Some log line 11
{
  "foo": "bar",
  "abc":  123
}
2024-08-13 17:21:34 INFO Main main Some log line 12
```
5. The JSON should be correctly aggregated (parsed as a structured log)

#### Tagging
1. Enable `logs_config.auto_multi_line_detection` and `logs_config.auto_multi_line.tag_aggregated_json`
4. emit some logs with mixed pretty printed JSON. 
```
2024-08-13 17:21:34 INFO Main main Some log line 10
2024-08-13 17:21:34 INFO Main main Some log line 11
{
  "foo": "bar",
  "abc":  123
}
2024-08-13 17:21:34 INFO Main main Some log line 12
```
6. Query for `aggregated_json:true` and the log should appear. 

#### broken JSON
1. Enable `logs_config.auto_multi_line_detection`
2. emit some logs with mixed with incomplete JSON
```
2024-08-13 17:21:34 INFO Main main Some log line 10
2024-08-13 17:21:34 INFO Main main Some log line 11
{
  "foo": "bar",
  "abc":  123
2024-08-13 17:21:34 INFO Main main Some log line 12
```
3. incomplete JSON should be joined to the log above it because it gets aggregated as a multi-line log since it was not detected as valid JSON. ex:
```
2024-08-13 17:21:34 INFO Main main Some log line 11
{
  "foo": "bar",
  "abc":  123
```

#### broken JSON flush timeout
1. Enable `logs_config.auto_multi_line_detection`
2. emit some logs with mixed with incomplete JSON, but no terminating log line to make the payload incomplete. 
```
2024-08-13 17:21:34 INFO Main main Some log line 10
2024-08-13 17:21:34 INFO Main main Some log line 11
{
  "foo": "bar",
  "abc":  123
```
NOTE: The omission of the last log line in this test is important! (compared to the previous test)

3. Wait a few seconds. Incomplete JSON should be joined to the log above it because it gets aggregated as a multi-line log since it was not detected as valid JSON. In this case, the buffer timed out and the logs were flushed and not aggregated. 
```
2024-08-13 17:21:34 INFO Main main Some log line 11
{
  "foo": "bar",
```

#### Feature disabled
1. Enable `logs_config.auto_multi_line_detection`
2. Set `logs_config.logs_config.auto_multi_line.enable_json_aggregation: false`
3. emit some logs with mixed pretty printed JSON. 
```
2024-08-13 17:21:34 INFO Main main Some log line 10
2024-08-13 17:21:34 INFO Main main Some log line 11
{
  "foo": "bar",
  "abc":  123
}
2024-08-13 17:21:34 INFO Main main Some log line 12
```
5. JSON should be joined to the log above it because it gets aggregated as a multi-line log. ex:
```
2024-08-13 17:21:34 INFO Main main Some log line 11
{
  "foo": "bar",
  "abc":  123
}
```

### Additional Notes

[](https://datadoghq.atlassian.net/browse/AGNTLOG-164)


releasenotes/notes/json-aggregation-fb6a85c1731f69e2.yaml
```diff
@@ -0,0 +1,13 @@
...
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+features:
+  - |
+    Pretty printed/multi-line JSON messages are now aggregated into a single line when auto multiline
+    detection is enabled. This ensures the log is treated a structured log when processed by Datadog.
+    This can be disabled by setting ``logs_config.auto_multi_line.enable_json_aggregation`` to ``false``.
```
urseberry, line 13:
```suggestion
    Aggregation can be disabled by setting ``logs_config.auto_multi_line.enable_json_aggregation`` to ``false``.
```

urseberry (approved):
Left wording nit feedback. Approved so as not to hold up release.

pkg/logs/internal/decoder/auto_multiline_detection/incremental_json_validator.go
```diff
@@ -0,0 +1,88 @@
...
+				d.objCount--
+			}
+		}
+	}
+	if !isValid {
+		return Invalid
+	}
+
+	if d.objCount <= 0 {
+		return Complete
```
ddrthall, line 78: We'll return Complete for a raw string (`"hi"`), a root-level array (`["hi"]`), and an initial opening array delimiter (`[`). I'm yet to see any direct negative impact from this outside of an errant tag, but it probably makes sense to fail everything with Invalid if we don't start with a `{` delimiter if ingest doesn't support those entities as valid json.

gh123man: Nice catch. 
Agreed that this should not cause problems. However I fixed it anyway and added test cases for these.

pkg/logs/internal/decoder/auto_multiline_detection/incremental_json_validator_test.go
```diff
@@ -0,0 +1,220 @@
...
+			name:     "Incomplete nested array",
+			inputs:   []string{`{ "arr":`, `[`, `[`, `]`, `}`},
+			expected: []JSONState{Incomplete, Incomplete, Incomplete, Incomplete, Invalid},
+		},
+		{
+			name:     "Complete nested array",
+			inputs:   []string{`{ "arr":`, `[`, `[`, `]`, `]`, `}`},
+			expected: []JSONState{Incomplete, Incomplete, Incomplete, Incomplete, Incomplete, Complete},
+		},
+		{
```
soberpeach, line 146: Maybe another test case with a complete JSON object plus extra text might be worth adding?

pkg/logs/message/message.go
```diff
@@ -19,11 +19,8 @@ import (
 // or/and at the end of every trucated lines.
 var TruncatedFlag = []byte("...TRUNCATED...")
 
-// TruncatedTag is added to truncated log messages (if enabled).
-const TruncatedTag = "truncated"
```
soberpeach, line 23: Is the truncated tag no longer added to truncated logs?

pkg/logs/internal/decoder/auto_multiline_detection/incremental_json_validator.go
```diff
@@ -0,0 +1,88 @@
...
+			if delim.String() == "}" {
+				d.objCount--
+			}
+		}
+	}
+	if !isValid {
+		return Invalid
+	}
+
+	if d.objCount <= 0 {
```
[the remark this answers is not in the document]
gh123man: In reality, it shouldn't and this would be a bug. 
However using `<=` is a sanity check in the off chance that something does go horribly wrong, this code will still do the right thing because a `-1` object count would still represent an invalid state.

pkg/logs/internal/decoder/auto_multiline_detection/incremental_json_validator_test.go
```diff
@@ -0,0 +1,205 @@
...
+			expected: []JSONState{Incomplete, Incomplete, Incomplete, Complete},
+		},
+		{
+			name:     "Invalid JSON with array",
+			inputs:   []string{`{`, `[`, `]`, `}`},
+			expected: []JSONState{Incomplete, Invalid, Invalid, Invalid},
+		},
+		{
+			name:     "Incomplete array",
+			inputs:   []string{`{ "arr":`, `[`, `[`, `}`},
```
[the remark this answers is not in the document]
gh123man: > Does the Token() function have internal state

Yes! the `json.Decoder` has it's own limited internal state representing the context of the json parser. 
Unfortunately, it doesn't count object open/closes so we have to do that manually.

pkg/logs/internal/decoder/auto_multiline_detection/incremental_json_validator_test.go
gh123man: Added an additional test 😄

pkg/logs/internal/decoder/auto_multiline_detection/json_aggregator.go
```diff
@@ -0,0 +1,89 @@
...
+		inBuf := &bytes.Buffer{}
+		for _, m := range r.messageBuf {
+			inBuf.Write(m.GetContent())
+		}
+		err := json.Compact(outBuf, inBuf.Bytes())
+		if err != nil {
+			return r.Flush()
+		}
+		r.messageBuf = r.messageBuf[:0]
+		r.currentSize = 0
```
ddrthall, line 62: Should we be setting the message's RawDataLen to currentSize or outBuf length?

gh123man: Good catch - yes. 
This does need to be `currentSize` to preserve any stripped whitespace in json parsing to ensure we maintain the correct file offset in the auditor! 

Added a unit test to cover this.

ddrthall (approved):

hush-hush (approved):

soberpeach (approved):

PR #36590, opened 2025-04-30
always set local account password

clarkb7:
### What does this PR do?
Sets the Agent account password to the provided password (or the password found in the LSA) when the account is a local account.

Moves the "last stable" configuration from e2e.yml to the test itself, to allow tests to override the value and to make it easier to run tests locally.

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1469
Fix issue caused by installing 7.65 or lower using fleet automation after upgrading to 7.66 or later, with default credentials. The  7.65 or lower install changes the account password, causing the password stored in the LSA to be incorrect (LSA feature was added in 7.66), subsequent attempts to install 7.66 or later will fail, as it uses the password from the LSA without setting the account password again.

Add new E2E tests

### Additional Notes

The password in the LSA is removed on manual full uninstall, but this is skipped during remote upgrade operations since we need the password.
This issue only applies to local accounts and installs with an empty/default password, since this configuration causes 7.65 or earlier to generate a new random password.


sblumenthal (approved):

releasenotes/notes/win-set-localaccount-password-c44389ee895acb4a.yaml
```diff
@@ -0,0 +1,7 @@
+---
+enhancements:
+  - |
+    The Windows Agent MSI now sets the Agent account password to the
+    provided ``DDAGENTUSER_PASSWORD`` value when it is a local account.
+    Previously, if the provided password did not match the account password
+    then the Agent would fail to start.
```
brett0000FF, line 7:
```suggestion
    Previously, if the provided password did not match the account password,
   the Agent would fail to start.
```

brett0000FF (approved):

jack0x2 (approved):

KevinFairise2 (approved):

BaptisteFoy (approved):

PR #36591, opened 2025-05-01
CEL Filter base [CONTP-809]

gabedos:
### What does this PR do?

Creates a new component for storing filtering programs to be eventually used across various agent components as part of a multiphase process.

Step 1. Create the filtering component with CEL filters (this PR)
Step 2. Replace existing containers.Filters with the component
Step 3. Expose never config options to users to create CEL filters directly

### Motivation

Centralize filtering logic. Improve filtering expressiveness through a rule engine (cel-go)

Unit-tests

Slight increase in static binary size to pull in additional cel-go packages. Although cel-go is already an indirect dependency in the Datadog Agent repo, it's completely removed from dead code elimination. The direct use of cel-go in this PR causes the binary size to increase.

Other rule engine options include Expr but that increases the static binary size by 35MB+ and causes SMP memory usage to increase by 7MB+ due to disabling dead code elimination completely as a result of the reflection methods used in expr.


comp/core/filter/def/types.go
```diff
@@ -0,0 +1,151 @@
...
+	ContainerType ResourceType = "container"
+	PodType       ResourceType = "pod"
+)
+
+//
+// Container Definition
+//
+
+// Container represents a filterable container object.
+type Container struct {
```
gabedos, line 45: You can look through the previous commits to see the iterations that this `Container` struct has looked like.

What I eventually landed on in the end was using an embedded proto struct. I thought of using a proto struct because it allows third party libraries like CEL-go to be able to understand the object and its attributes. Thus, the CEL parser at compile time is able to throw precise errors if users ever misconfigure any filter. We still have access to all the attributes if we ever decide to use them in some other way.

Currently I am thinking of providing functions to help other users of this component to create the Filterable structs from the wmeta structs. However, this is something that we could theoretically do within the component itself as well. For the case of `IsContainerExcluded` we need to have the container's owner. This could be either a pod, task, or maybe something else. We could store a reference to `wmeta` internally and query for the container's owner. My only thought is that this could be repetitive work if the client already has the owner pod for other reasons but on the flip side would make the interface simpler.

pgimalac:
Is this component supposed to be imported in the heroku and iot agents ?
Also does it need to be imported on Windows ?

gabedos:
> Is this component supposed to be imported in the heroku and iot agents ? Also does it need to be imported on Windows ?

Hey @pgimalac,

In regards to Windows, I think this component would still be used in that environment because all of the core agent functionalities like Autodiscovery and Container corechecks still run on windows and we would want to give the same experience for users regardless of OS.

For heroku and iot agents... I'm honestly not as familiar what the Agent does in these environments. It seems like autodiscovery is pulled into all of these flavors still? and one of the primary use cases of these filters would be to restrict what is being autodiscovered. I know in the next quarter we also want to start autodiscovering log collection on host processes (but once again unfamiliar with heroku/iot if we even collect logs there). Would appreciate if you have any significant missing context related to these 2 flavors!

rahulkaukuntla (approved):

comp/core/filter/catalog/utils.go
```diff
@@ -0,0 +1,124 @@
...
+			continue
+		}
+
+		// Split the filter into key and value using the first colon
+		parts := strings.SplitN(filter, ":", 2)
+		if len(parts) != 2 {
+			return "", fmt.Errorf("invalid filter format: %s", filter)
+		}
+
+		key, value := parts[0], parts[1]
```
pgimalac: **🥜 nitpick:** ‏You can use `strings.Cut`
```suggestion
		key, value, ok := strings.Cut(filter, ":")
		if !ok {
			return "", fmt.Errorf("invalid filter format: %s", filter)
		}
```

comp/core/filter/def/types.go
```diff
@@ -0,0 +1,141 @@
...
+func CreatePod(pod workloadmeta.KubernetesPod) *Pod {
+	return &Pod{
+		FilterPod: &typedef.FilterPod{
+			Id:          pod.ID,
+			Name:        pod.Name,
+			Namespace:   pod.Namespace,
+			Annotations: pod.Annotations,
+		},
+	}
+}
```
pgimalac, line 119: **💬 suggestion:** ‏I believe this is only used in tests right now, if it's going to stay that way it should be moved to a test file, otherwise maybe a `workloadmetautils` subpackage would be best ?
That way we avoid a direct dependency from the filter component to workloadmeta types by default

Same for `CreateContainer`

comp/core/filter/def/utils.go
```diff
@@ -0,0 +1,85 @@
...
+	pkgconfigsetup "github.com/DataDog/datadog-agent/pkg/config/setup"
+)
+
+const (
+	highPrecedence = 0
+	lowPrecedence  = 1
+)
+
+// GetSharedMetricsFilters identifies the filtering component's individual Container Filters for container metrics.
+func GetSharedMetricsFilters() [][]ContainerFilter {
```
pgimalac, line 18: Is this used anywhere ?

pgimalac: Same for everything in this file actually ?

pgimalac (approved):
LGTM, just some nitpicks

As discussed I think the size increase for the IoT Agent is acceptable, and for msi it's actually not such a big increase, the comment has some known issues

comp/core/filter/def/utils.go
gabedos: In hindsight this file would make more sense in the next PR, but the general purpose of this filter component (initially) is to replace the pre-existing `containers.Filters`. That filtering package has these same helper methods called [GetSharedMetricFilter](https://github.com/DataDog/datadog-agent/blob/ea730f9d06a9b58e0e36d60610b82380f10be27b/pkg/util/containers/filter.go#L172-L173) and [NewAutodiscoveryFilter](https://github.com/DataDog/datadog-agent/blob/ea730f9d06a9b58e0e36d60610b82380f10be27b/pkg/util/containers/filter.go#L297-L298) that are used to centralize the setup for the filters.

comp/core/filter/def/proto/filter.proto
```diff
@@ -0,0 +1,28 @@
...
+  string name = 2;
+  string namespace = 3;
+  map<string, string> annotations = 4;
+}
+
+message FilterECSTask {
+  string id = 1;
+  string arn = 2;
+  // Add other ECS task fields here
+}
```
wdhif: This EOF is broken

comp/core/filter/def/proto/filter.proto
```diff
@@ -0,0 +1,28 @@
...
+  string id = 1;
+  string name = 2;
+  string namespace = 3;
+  map<string, string> annotations = 4;
+}
+
+message FilterECSTask {
+  string id = 1;
+  string arn = 2;
+  // Add other ECS task fields here
```
wdhif: Was this a TODO that is supposed to be done now or in a subsequent PR?
You might want to actually add `TODO: ` in front of it if it would be done later to keep track of it 👍

comp/core/filter/impl/filter.go
```diff
@@ -0,0 +1,148 @@
...
+	filterSets [][]T, // Generic filter types
+) filterdef.Result {
+	for _, filterSet := range filterSets {
+		var setResult = filterdef.Unknown
+		for _, filter := range filterSet {
+
+			// 1. Retrieve the filtering program
+			prg := f.getProgram(resource.Type(), int(filter))
+			if prg == nil {
+				f.log.Warnf("No program found for filter %d on resource %s", filter, resource.Type())
```
wdhif, line 122: Could this log become too noisy if the resource is only part of a single program, or not associated with any program at all?

comp/core/filter/program/cel_program.go
```diff
@@ -0,0 +1,62 @@
...
+	if p.Include != nil {
+		out, _, err := p.Include.Eval(map[string]any{string(entity.Type()): entity.Serialize()})
+		if err == nil {
+			res, ok := out.Value().(bool)
+			if ok {
+				if res {
+					return filterdef.Included, nil
+				}
+			} else {
+				errs = append(errs, fmt.Errorf("Include (%s) result not bool: %v", p.Name, out.Value()))
```
wdhif: nit: Error string should not be capitalized or end with punctuation mark

comp/core/filter/program/cel_program.go
```diff
@@ -0,0 +1,62 @@
...
+			}
+		} else {
+			errs = append(errs, fmt.Errorf("Include (%s) eval error: %w", p.Name, err))
+		}
```
wdhif, line 42: nit: Error string should not be capitalized or end with punctuation mark

comp/core/filter/program/cel_program.go
```diff
@@ -0,0 +1,62 @@
...
+	if p.Exclude != nil {
+		out, _, err := p.Exclude.Eval(map[string]any{string(entity.Type()): entity.Serialize()})
+		if err == nil {
+			res, ok := out.Value().(bool)
+			if ok {
+				if res {
+					return filterdef.Excluded, nil
+				}
+			} else {
+				errs = append(errs, fmt.Errorf("Exclude (%s) result not bool: %v", p.Name, out.Value()))
```
wdhif: nit: Error string should not be capitalized or end with punctuation mark

comp/core/filter/program/cel_program.go
```diff
@@ -0,0 +1,62 @@
...
+			}
+		} else {
+			errs = append(errs, fmt.Errorf("Exclude (%s) eval error: %w", p.Name, err))
```
wdhif: nit: Error string should not be capitalized or end with punctuation mark

comp/core/filter/impl/filter.go
```diff
@@ -0,0 +1,148 @@
...
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+// Package filterimpl contains the implementation of the filter component.
+package filterimpl
+
+import (
+	"github.com/DataDog/datadog-agent/comp/core/config"
+	catalog "github.com/DataDog/datadog-agent/comp/core/filter/catalog"
```
wdhif: nit: This alias is not needed

wdhif (approved):

comp/core/filter/def/proto/filter.proto
gabedos: The `FilterECSTask` was included in the proto definitions as a proof of concept that we could associate a container with multiple different entities for filtering and make sure the implementation would be flexible to support that in the future.

I can add a `TODO:` in front of it, but it definitely is not on our short term radar to support filtering based on the attributes of the task definition for ECS. Maybe a KR to be added in Q3 or an onboarding ramp-up work.

comp/core/filter/impl/filter.go
gabedos: This log warning would only execute if there is a bug in the use/setup of the filtering component. Through the interface, I am forcing users to only provide filters that are applicable for that resource:

```
IsContainerExcluded(container *Container, containerFilters [][]ContainerFilter)
```

If the filter has been registered properly in the initialization of the component in `newFilter` using `filter.registerProgram`, then this warning would never execute.

Essentially the intention behind the log was to be noisy and alert us before ever releasing the Agent if a bug is introduced into the filtering component initialization. Do you have any other recommendations as to how this could be done?

L3n41c (approved):
👍🏻 LGTM for the single file owned by @DataDog/container-integrations.

comp/core/filter/catalog/utils.go
```diff
@@ -0,0 +1,119 @@
...
+	filter "github.com/DataDog/datadog-agent/comp/core/filter/def"
+	log "github.com/DataDog/datadog-agent/comp/core/log/def"
+	legacyFilter "github.com/DataDog/datadog-agent/pkg/util/containers"
+)
+
+// createProgramFromOldFilters handles the conversion of old filters to new filters and creates a CEL program.
+func createProgramFromOldFilters(oldFilters []string, key filter.ResourceType, logger log.Component) cel.Program {
+	filterString, err := convertOldToNewFilter(oldFilters)
+	if err != nil {
+		logger.Warnf("Error converting filters: %v", err)
```
adel121, line 25: Is it sufficient to only log a warning here? 

Should we modify this function to return `(cel.Program, error)` ? 

My main concern is that if we silently fail, we might end up silently including stuff that the user wanted to exclude. 
For autodiscovery for example, IIRC, any issues in the filters are shown in the `agent status` command. Not sure if this is preserved with this change.

Maybe it was like this already, but just thinking out loud.

WDYT?

comp/core/filter/catalog/utils.go
```diff
@@ -0,0 +1,119 @@
...
+		return nil, nil
+	}
+	env, err := cel.NewEnv(
+		cel.Types(&filter.Container{}, &filter.Pod{}),
+		cel.Variable(string(key), cel.ObjectType(convertTypeToProtoType(key))),
+	)
+	if err != nil {
+		return nil, err
+	}
+	ast, issues := env.Compile(rules)
```
adel121, line 49: nit: can we improve the name here? (`ast` might not be obvious)

gabedos: Noted: I can make the name abstractSyntaxTree

comp/core/filter/catalog/utils.go
gabedos: Currently our AD filters ignore any errors I believe. As seen below, it reads the config input and makes it's best attempt to parse and create filters.

https://github.com/DataDog/datadog-agent/blob/88a97d99bcefee4bbf95b7a039ae61100cc745d8/pkg/util/containers/filter.go#L297-L322

When I look for all references of the `NewFilter` method, you do bring up a good point that there are a few clients in the Agent that do care about the filter being properly parsed/defined. Particularly, the DCA utilizes the error in the admission controller and completely stops the webhook it the user had any bug in the filter definition.

In a subsequent PR I can track the errors while defining filters and store them. Then I can expose these errors for any components that want to make decisions off of them through a new method in the interface. Eventually, I can also add some commands to help view the health of each defined filter as well.

alopezz (approved):

PR #36604, opened 2025-05-02
Remove flaky mark from TestWindowsDiskSuite

jose-manuel-almaza:
### What does this PR do?
Remove flaky test flag from `TestWindowsDiskSuite`.
Increase the accepted percentage difference for some metrics between the Python and Go versions of the check to _5%_, thus reducing the chances of test failures.
Set some metrics (e.g., `system.disk.total`) as constants, since their value should not change between runs of the check (Python and Go)

### Motivation
The mark was introduced temporarily after detecting the test as flaky in our CI.
The test almost always failed on Windows (no failure was detected on Linux) possibly because the processes running in the background (updates, scans, etc.) make greater use of the disk and this caused the metrics `system.disk.used` and `system.disk.utilized` to increase by more than 2% (the previous limit of the difference margin), thus causing the test to fail.
We want the test to always run.

The test should always pass unless the difference in the value of any metric exceeds 5% between both executions of the check (python and Go)

### Additional Notes

It has been verified that the [Python](https://github.com/giampaolo/psutil/blob/077d9c66e18ce9bacd93de98490a1b3999e67eb8/psutil/arch/windows/disk.c#L60) and [Go](https://github.com/shirou/gopsutil/blob/ec85c0b45bcbcc7e53ffa57e4b81ab58d9c933a5/disk/disk_windows.go#L62) code to obtain the failing metrics is very similar and in both cases ended up calling the `GetDiskFreeSpaceExW` primitive, which rules out a test error due to differences in the code executed by the check in Python and Go.

### DIFF
--- a/test/new-e2e/tests/agent-runtimes/check_disk/disk_common_test.go
+++ b/test/new-e2e/tests/agent-runtimes/check_disk/disk_common_test.go
@@ -92,12 +92,12 @@ instances:
 					if !gocmp.Equal(a.Tags, b.Tags, gocmpopts.SortSlices(cmp.Less[string])) {
 						return false
 					}
-					a_value := a.Points[0][1]
-					b_value := b.Points[0][1]
+					aValue := a.Points[0][1]
+					bValue := b.Points[0][1]
 					if isConstantMetric(a.Metric) {
-						return a_value == b_value
+						return aValue == bValue
 					}
-					return compareValuesWithRelativeMargin(a_value, b_value, p, metricCompareFraction)
+					return compareValuesWithRelativeMargin(aValue, bValue, p, metricCompareFraction)
 				}),
 				gocmpopts.SortSlices(metricPayloadCompare), // sort metrics
 			)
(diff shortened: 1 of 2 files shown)


test/new-e2e/tests/agent-runtimes/check_disk/disk_common_test.go
```diff
@@ -50,12 +50,17 @@ func (v *baseCheckSuite) getSuiteOptions() []e2e.SuiteOption {
 	return suiteOptions
 }
 
-// a relative diff considered acceptable when comparing metrics
-const metricCompareFraction = 0.02
+// a relative diff considered acceptable when comparing metrics (5%)
+const metricCompareFraction = 0.05
```
misteriaud, line 54: **💬 suggestion:** ‏Since the difference is higher only on Windows, we should increase it only for this OS test. We could add `metricCompareFraction` as field in `baseCheckSuite` instead of having it as a const. This will let us keep `0.02` for linux check, WDYT ?

misteriaud (approved):

test/new-e2e/tests/agent-runtimes/check_disk/disk_common_test.go
```diff
@@ -81,20 +86,38 @@ instances:
...
+						return false
+					}
+					if !gocmp.Equal(a.Tags, b.Tags, gocmpopts.SortSlices(cmp.Less[string])) {
+						return false
+					}
+					aValue := a.Points[0][1]
+					bValue := b.Points[0][1]
+					if _, ok := constantMetricsSet[a.Metric]; ok {
+						return aValue == bValue
+					}
```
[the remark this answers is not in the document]
jose-manuel-almaza: I'll address this in a new PR

test/new-e2e/tests/agent-runtimes/check_disk/disk_common_test.go
jose-manuel-almaza: I'll address this in a new PR

PR #36638, opened 2025-05-05
Run serverless coldstart on dind shadow

KevinFairise2:
### What does this PR do?

Run a shadow version of serverless test in docker in docker runner. In order to see if there's still flakes in the new environment.
Should replace the old runner once we are confident it is stable enough.


chouetz (approved):
Ok let's try like this. Do you want to add a monitor to count the number of errors, and set a target date to decide if we replace the old job by this one?

PR #36650, opened 2025-05-05
use s3 url instead of cloudfront in e2e tests

KevinFairise2:


BaptisteFoy:

PR #36657, opened 2025-05-05
AGNTLOG-145 - remove sds feature from the logs agent

ddrthall:
### What does this PR do?
The SDS integration with the logs agent has been slated for retirement, with the frontend configuration components officially reaching EOL on the 26th of June. However, there remains a desire to find usages of the SDS library in other components of the agent in the future. As such, this MR will remove
* all usage of the SDS scanner in the core logs pipeline
*  the SDS library build tags in all of the gitlab jobs
* SDS as an omnibus defined dependency of the datadog agent

but will retain

* the SDS go module
* the ability to declare an SDS build tag
* the invoke task to install the SDS library

These retained entities will be subject to further review in six months time to determine if we're still looking to incorporate the SDS components in other features or if it is safe to fully remove the remaining SDS elements.

Validation was performed primarily by the automated regression suite. No new functionality was introduced.


pgimalac (approved):

pkg/logs/status/status.go
```diff
@@ -31,8 +31,6 @@ const (
 	StatusNotStarted = 0
 	// StatusRunning means that the logs agent is running and fully operational
 	StatusRunning = 1
-	// StatusCollectionNotStarted means that the logs agent has not started collecting logs
-	StatusCollectionNotStarted = 2
```
soberpeach, line 35: Was SDS the only reason to have this status?

soberpeach (approved):

pkg/logs/status/status.go
ddrthall: Indeed, the only two usages of this state are locked behind SDS specific functionality

jhgilbert (approved):

omnibus/config/software/datadog-agent-dependencies.rb
```diff
@@ -17,10 +17,6 @@
 # External agents
 dependency 'jmxfetch'
 
-if linux_target?
-  dependency 'sds'
```
chouquette, line 21: Could you also remove the `omnibus/config/software/sds.rb` file since it isn't used anymore?

jeremy-hanna (approved):

KevinFairise2 (approved):
Lgtm for agent deux owned files

chouquette (approved):

PR #21266, opened 2023-12-04
[chore] Remove new-from-rev parameter from golangci-lint

KSerrania:
### What does this PR do?

Removes the `new-from-rev` parameter from golangci-lint.
Adds linter placeholders for all failures reported by revive and other linters.

### Motivation

While it temporarily enabled us to re-enable the `revive` linter a few months ago, the `new-from-rev` parameter has harmful side-effects:
- it hides new linter failures on files not modified after the `new-from-rev` commit, when these linter failures are due to changes in other files (eg. a function becomes unused).
- it makes the linter unpredictable: if there is an operation that the underlying git diff cannot track well, like moving & modifying a file, then `golangci-lint` would throw linter errors that are unrelated to the changes.

### Additional Notes

- Each team has one commit for revive, and one commit for other linters (if applicable). Feel free to only review commits for your team.
- Note for Agent Shared Components: `gofmt` also forced a change in `pkg/gohai/cpu/lscpu_linux_arm64.go`, please verify that it is okay to keep.

### Possible Drawbacks / Trade-offs

n/a

### Describe how to test/QA your changes

Run a full pipeline, verify that it passes.

### Reviewer's Checklist

- [x] If known, an appropriate milestone has been selected; otherwise the `Triage` milestone is set.
- [x] Use the `major_change` label if your change either has a major impact on the code base, is impacting multiple teams or is changing important well-established internals of the Agent. This label will be use during QA to make sure each team pay extra attention to the changed behavior. For any customer facing change use a releasenote.
- [x] A [release note](https://github.com/DataDog/datadog-agent/blob/main/docs/dev/contributing.md#reno) has been added or the `changelog/no-changelog` label has been applied.
- [x] Changed code has automated tests for its functionality.
- [x] Adequate QA/testing plan information is provided if the `qa/skip-qa` label is not applied.
- [x] At least one `team/..` label has been applied, indicating the team(s) that should QA this change.
- [x] If applicable, docs team has been notified or [an issue has been opened on the documentation repo](https://github.com/DataDog/documentation/issues/new).
- [x] If applicable, the `need-change/operator` and `need-change/helm` labels have been applied.
- [x] If applicable, the `k8s/<min-version>` label, indicating the lowest Kubernetes version compatible with this feature.
- [x] If applicable, the [config template](https://github.com/DataDog/datadog-agent/blob/main/pkg/config/config_template.yaml) has been updated.

PR #36658, opened 2025-05-05
Add the Windows Certificate Store check

mrafi97:
### What does this PR do?
This PR adds a check to monitor the expiration of certificates stored in the Windows Certificate Store

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1446

Add the following configuration in `%PROGRAMDATA%\conf.d\windows_certificate.d\conf.yaml` and restart the Agent. It should return a metric called `windows_certificate.days_remaining` and service check `windows_certificate.cert_expiration`:

```
init_config:

instances:
  - certificate_store: ROOT
    certificate_subjects:
      - verisign
      - Microsoft
```


pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
```diff
@@ -0,0 +1,309 @@
...
+		log.Errorf("Error getting certificates: %v", err)
+		return nil, err
+	}
+
+	log.Debugf("Found %d certificates in store", len(certificates))
+	return certificates, nil
+}
+
+// getEnumCertificatesInStore retrieves all certificates in a certificate store
+func getEnumCertificatesInStore(certContext *windows.CertContext, storeHandle windows.Handle) ([]*x509.Certificate, error) {
```
iglendd, line 188: I am not sure how certcontext is passed and freed. (per [getEnumCertificatesInStore](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-certenumcertificatesinstore#remarks). I feel that in your specified flow you probably do not need to pass it in but also you probably need to free it on some (or all?) errors explicitly.

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
```diff
@@ -0,0 +1,309 @@
...
+// findCertificatesInStore finds certificates in a store with a given subject string
+func findCertificatesInStore(certContext *windows.CertContext, storeHandle windows.Handle, subjectFilters []string) ([]*x509.Certificate, error) {
+	var err error
+	certificates := []*x509.Certificate{}
+
+	for _, subject := range subjectFilters {
+		subjectName := windows.StringToUTF16Ptr(subject)
+
+		for {
+			certContext, err = windows.CertFindCertificateInStore(
```
iglendd, line 225: comments for context is probably applied here as well

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
```diff
@@ -0,0 +1,309 @@
...
+	for _, subject := range subjectAttributes {
+		subjectTags = append(subjectTags, fmt.Sprintf("subject_%s:%s", getAttributeTypeName(subject.Type.String()), subject.Value))
+	}
+
+	log.Debugf("Subject tags: %v", subjectTags)
+	return subjectTags
+}
+
+// Returns the human-readable name for a given OID string
+func getAttributeTypeName(oid string) string {
```
iglendd, line 284: Interesting, where did you get this list. I think it is too short. And what about extended oid. Here are some list I found on google
 Abbreviated names for NameOID entries
ABBREVIATED_NAME_OIDS = {
    NameOID.COMMON_NAME: "CN",
    NameOID.COUNTRY_NAME: "C",
    NameOID.LOCALITY_NAME: "L",
    NameOID.STATE_OR_PROVINCE_NAME: "ST",
    NameOID.ORGANIZATION_NAME: "O",
    NameOID.ORGANIZATIONAL_UNIT_NAME: "OU",
    NameOID.SERIAL_NUMBER: "SERIALNUMBER",
    NameOID.SURNAME: "SN",
    NameOID.GIVEN_NAME: "GN",
    NameOID.TITLE: "T",
    NameOID.EMAIL_ADDRESS: "E",
    NameOID.DN_QUALIFIER: "DNQ",
    NameOID.DOMAIN_COMPONENT: "DC",
    NameOID.USER_ID: "UID",
}

ABBREVIATED_EXTENSION_OIDS = {
    ExtensionOID.SUBJECT_ALTERNATIVE_NAME: "SAN",
    ExtensionOID.BASIC_CONSTRAINTS: "BC",
    ExtensionOID.KEY_USAGE: "KU",
    ExtensionOID.EXTENDED_KEY_USAGE: "EKU",
    ExtensionOID.AUTHORITY_KEY_IDENTIFIER: "AKI",
    ExtensionOID.SUBJECT_KEY_IDENTIFIER: "SKI",
    ExtensionOID.CERTIFICATE_POLICIES: "CP",
    ExtensionOID.CRL_DISTRIBUTION_POINTS: "CDP",
    ExtensionOID.AUTHORITY_INFORMATION_ACCESS: "AIA",
    ExtensionOID.INHIBIT_ANY_POLICY: "IAP",
    ExtensionOID.NAME_CONSTRAINTS: "NC",
    ExtensionOID.POLICY_CONSTRAINTS: "PC",
    ExtensionOID.PRECERT_SIGNED_CERTIFICATE_TIMESTAMPS: "SCT",
    ExtensionOID.OCSP_NO_CHECK: "ONC",
    ExtensionOID.ISSUER_ALTERNATIVE_NAME: "IAN",
    ExtensionOID.FRESHEST_CRL: "FCRL",
    ExtensionOID.PRECERT_POISON: "PP",
}

iglendd (changes requested):
There are a few comment, but I am wondering if we have a memory leak in certificate context in some cases.

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
```diff
@@ -0,0 +1,309 @@
...
+	if err := yaml.Unmarshal(data, &w.config); err != nil {
+		return err
+	}
+
+	if w.config.DaysCritical == 0 {
+		w.config.DaysCritical = 7
+	}
+	if w.config.DaysWarning == 0 {
+		w.config.DaysWarning = 14
+	}
```
julien-lebot, line 77: **❓ question:** ‏ I wonder if there is a scenario where the user would want to have one or both of those values as 0.

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
```diff
@@ -0,0 +1,309 @@
...
+func (w *WinCertChk) Run() error {
+	sender, err := w.GetSender()
+	if err != nil {
+		return err
+	}
+	defer sender.Commit()
+
+	if len(w.config.CertificateStores) == 0 {
+		return fmt.Errorf("no certificate stores specified")
+	}
```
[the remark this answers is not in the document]
julien-lebot: I'm thinking if no certificates store are configured we should error out in the `Configure` function - why wait until the check is running to detect that there are not certificates store to check?

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
```diff
@@ -0,0 +1,309 @@
...
+
+	for _, store := range w.config.CertificateStores {
+		certificates, err := getCertificates(store, w.config.CertSubject)
+		if err != nil {
+			return err
+		}
+		if len(certificates) == 0 {
+			log.Warnf("No certificates found in store: %s for subject filters: %v", store, w.config.CertSubject)
```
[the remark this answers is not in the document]
julien-lebot: It should print it like so:
```
[subject A subject B]
```
I agree that perhaps we ought to do some formatting to ensure the subjects are delimited by some commas and surrounded by quotes.

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
julien-lebot: **✏️ suggestion:** ‏ We should also perform some validation on the input, if `DaysWarning` < `DaysCritical` we should print a warning that the `DaysWarning` event will never be emitted.
Perhaps that's what the user wants, as per my comment above, but we should log it.

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
```diff
@@ -0,0 +1,309 @@
...
+// Configure accepts configuration
+func (w *WinCertChk) Configure(senderManager sender.SenderManager, _ uint64, data integration.Data, initConfig integration.Data, source string) error {
+	err := w.CommonConfigure(senderManager, initConfig, data, source)
+	if err != nil {
+		return err
+	}
+
+	if err := yaml.Unmarshal(data, &w.config); err != nil {
+		return err
+	}
```
julien-lebot, line 70: **✏️ suggestion:** ‏ I think the Go error alone makes it difficult to understand what field prevented the un-marshalling from working and we should use something like a `goschema` to validate the configuration before un-marshalling.

pkg/collector/corechecks/system/windowscertificate/windows_certificate_test.go
```diff
@@ -0,0 +1,151 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/mock"
+
+	"github.com/DataDog/datadog-agent/comp/core/autodiscovery/integration"
+	"github.com/DataDog/datadog-agent/pkg/aggregator/mocksender"
+	"github.com/DataDog/datadog-agent/pkg/metrics/servicecheck"
+)
+
+func TestWindowsCertificate(t *testing.T) {
```
julien-lebot, line 21: **✏️ suggestion:** ‏ We're missing some edge case tests here, for instance:
- DaysCritical <= 0 || DaysWarning <= 0
- DaysWarning > DaysCritical
- CertificateStores == []{""} || CertSubject == []{""}

julien-lebot (commented):
left some comments

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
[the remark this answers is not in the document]
mrafi97: The check throwing an error when no certificate stores are specified makes sense since that is a required parameter.  The check has nothing to do if a user provides no certificate stores. Generally, agent integrations do the same thing when a required parameter is missing.

I'll move the this error so that it's thrown in the `Configure` function.

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
```diff
@@ -0,0 +1,309 @@
...
+		0,
+		0,
+		windows.CERT_SYSTEM_STORE_LOCAL_MACHINE|windows.CERT_STORE_READONLY_FLAG,
+		uintptr(unsafe.Pointer(storeName)))
+	if err != nil {
+		log.Errorf("Error opening certificate store: %v", err)
+		return nil, err
+	}
+
+	log.Debugf("Store handle: %v", storeHandle)
```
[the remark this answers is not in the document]
mrafi97: This is a left over debug line from my own debugging, I'll get rid of it.

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
mrafi97: The way the certificates are parsed, the subject is broken down to the OID and its value so we need to manually remap them to the human readable name: 
```
[{0.9.2342.19200300.100.1.25 com} {0.9.2342.19200300.100.1.25 microsoft} {2.5.4.3 Microsoft Root Certificate Authority}]
```

I got this list of OIDS from this GO library that the x509 package uses to parse certificate subjects: https://cs.opensource.google/go/go/+/refs/tags/go1.24.2:src/crypto/x509/pkix/pkix.go;l=26

I added the following line, because its an OID that also has been appearing in some of the default certificates in Windows:
```
case "0.9.2342.19200300.100.1.25":
		return "DC"
```

I'll expand this function to include more OIDS.

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
```diff
@@ -0,0 +1,309 @@
...
+			log.Debugf("Found certificate: %s", cert.Subject.String())
+			daysRemaining := getCertExpiration(cert)
+
+			// Adding Subject and Certificate Store as tags
+			tags := getSubjectTags(cert)
+			tags = append(tags, "certificate_store:"+store)
+
+			sender.Gauge("windows_certificate.days_remaining", daysRemaining, "", tags)
+
+			if daysRemaining <= 0 {
```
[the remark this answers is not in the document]
mrafi97: Yeah that's true, for a situation like that I can check the number of seconds left and report that on the service check.

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
```diff
@@ -0,0 +1,309 @@
...
+					servicecheck.ServiceCheckCritical,
+					"",
+					tags,
+					"Certificate has expired")
+			} else if daysRemaining < float64(w.config.DaysCritical) {
+				sender.ServiceCheck("windows_certificate.cert_expiration",
+					servicecheck.ServiceCheckCritical,
+					"",
+					tags,
+					fmt.Sprintf("Certificate will expire in only %d days", int(daysRemaining)))
```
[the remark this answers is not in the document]
mrafi97: It's not really necessary, I was just mimicking what the TLS check uses for consistency: https://github.com/DataDog/integrations-core/blob/master/tls/datadog_checks/tls/tls.py#L201-L207

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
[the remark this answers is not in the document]
mrafi97: That makes sense. I can use `Strings.join(w.config.CertSubject, ",")` to do this

cmd/agent/dist/conf.d/windows_certificate.d/conf.yaml.example
```diff
@@ -0,0 +1,84 @@
...
+## Every instance is scheduled independently of the others.
+#
+instances:
+
+    ## @param certificate_stores - list of strings - required
+    ## The list of certificate stores to query from when verifying the Windows server certificates.
+    #
+  - certificate_stores:
+      - ROOT
+      - CA
```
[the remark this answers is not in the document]
mrafi97: I was thinking of adding this information as part of the documentation for the integration. I can also add it to the `conf.yaml` as well for consistency so that users have two sources of truth. For the `conf.yaml` I was thinking the following:

    ## @param certificate_stores - list of strings - required
    ## The list of certificate stores to query from when verifying the Windows certificates.
    ## This check can only query from local machine certificates.
    ## Enter the store names as found in `HKEY_LOCAL_MACHINE\Software\Microsoft\SystemCertificates`
    #
    # - certificate_stores:
    #  - MY
    #  - ROOT
    #  - CA

I've been using this MSDN article as my reference: https://learn.microsoft.com/en-us/windows/win32/seccrypto/system-store-locations

cmd/agent/dist/conf.d/windows_certificate.d/conf.yaml.example
```diff
@@ -0,0 +1,84 @@
...
+
+    ## @param certificate_subject - list of strings - optional
+    ## A list of regex patterns to filter by the subject of the certificates. 
+    ## For example if you have `Microsoft Root Authority` in `cert_subject`
+    ## and you have a certificate with subject, 
+    ## "CN=Microsoft Root Authority, OU=Microsoft Corporation, OU=Copyright (c) 1997 Microsoft Corp."
+    ## then the certificate will be matched and metrics will be collected for that certificate.
+    #
+    # certificate_subject:
+    #   - <REGEX_1>
```
alexn-dd, line 31: I am not sure if regex usage is already established in Datadog and YAML, but regex engines are known to have vulnerabilities. This can be potential security attack vector.  If it already a common pattern in Datadog, then it is a larger problem for the company, out of scope of this PR.

mrafi97: Yeah this is my bad for not updating. The original plan was to use regex, but the API actually just does a search for a certificate subject that matches a provided string: https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-certfindcertificateinstore#:~:text=CERT_FIND_SUBJECT_STR,matches%20dwCertEncodingType.

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
mrafi97: Yeah makes sense, I didn't catch that. I can rewrite these functions so that certContext isn't a passed argument and have [CertFreeCertificateContext](https://learn.microsoft.com/en-us/windows/desktop/api/wincrypt/nf-wincrypt-certfreecertificatecontext) be run as a deferred func.

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
```diff
@@ -0,0 +1,389 @@
...
+	for {
+		certContext, err = windows.CertEnumCertificatesInStore(storeHandle, certContext)
+		if err != nil {
+			if certContext == nil {
+				log.Debugf("No more certificates in store")
+				break
+			}
+			log.Errorf("Error enumerating certificates in store: %v", err)
+			return nil, err
+		}
```
iglendd, line 257: I do not think that this block is accurate. If you get an error will certContext be nil in all cases except complete enumeration? I am not sure but I feel that it could be nil in other cases as well. How Go maps native return code ERROR_NO_MORE_FILES and CRYPT_E_NOT_FOUND. They are "Ok" but everything else is a real error and require logging (instead of relying on context value that is). The same for the function below. I wonder if you should use an utility function which handles Cert retrival values and parsing since it is common for the code. Similarly, defer to close context probably can be to move to a dedicated function to make overall code smaller and more readable.

rtrieu (approved):

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
```diff
@@ -0,0 +1,389 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-present Datadog, Inc.
+//go:build windows
+
+//nolint:revive // TODO(WINA) Fix revive linter
```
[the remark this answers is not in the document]
mrafi97: Hey @iglendd, not sure to be honest. I noticed that other checks also had similar lines so I added it. They were added with this specific PR: https://github.com/DataDog/datadog-agent/pull/21266

I added it to be safe but can just go ahead and remove it as well.

cmd/agent/dist/conf.d/windows_certificate.d/conf.yaml.example
```diff
@@ -0,0 +1,88 @@
...
+    ## A list of strings to filter by the subjects of the certificates.
+    ## For example if you have `Microsoft Root Authority` in `certificate_subjects`
+    ## and you have a certificate with subject,
+    ## "CN=Microsoft Root Authority, OU=Microsoft Corporation, OU=Copyright (c) 1997 Microsoft Corp."
+    ## then the certificate will be matched and metrics will be collected for that certificate.
+    ## If `certificate_subjects` is empty then all certificates in `certificate_stores` are collected.
+    #
+    # certificate_subjects:
+    #   - <SUBJECT_1>
+    #   - <SUBJECT_2>
```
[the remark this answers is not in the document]
mrafi97: @iglendd 

My approach was similar to the windows service integration where customers provide a list of services to monitor and then also provide other filtering criteria such as triggers and startup type. 

For your approach are you suggesting the following?
```
- certificate_stores: MY
  certificate_subjects: 
    - <Subject_1>
    - <Subject_2>
- certificate_stores: ROOT
  certificate_subjects: 
    - <Subject_1>
```

Basically have users specify a single store per instance and a list of subjects to get from that store?

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
```diff
@@ -0,0 +1,389 @@
...
+		DaysWarning:  14,
+	}
+
+	if err := yaml.Unmarshal(data, &config); err != nil {
+		return fmt.Errorf("cannot unmarshal configuration: %s", err)
+	}
+	w.config = config
+
+	if w.config.DaysWarning < w.config.DaysCritical {
+		log.Warnf("Days warning (%d) is less than days critical (%d). Warning service checks will not be emitted.", w.config.DaysWarning, w.config.DaysCritical)
```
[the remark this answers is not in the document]
mrafi97: That's a good catch, if they're equal a warning should be emitted to state that Warning service checks will not be emitted.

cmd/agent/dist/conf.d/windows_certificate.d/conf.yaml.example
```diff
@@ -0,0 +1,88 @@
...
+
+    ## @param days_warning - number - optional - default: 14.0
+    ## Number of days before certificate expiration from which the service check
+    ## `tls.cert_expiration` begins emitting WARNING.
+    #
+    # days_warning: 14.0
```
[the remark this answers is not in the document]
mrafi97: The TLS check processes this as a float but I agree that doesn't really make sense. This check actually processes it as an int, the conf.yaml should also reflect that.

cmd/agent/dist/conf.d/windows_certificate.d/conf.yaml.example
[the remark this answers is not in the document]
iglendd: Exactly. It is a bit odd to search for the same subject in different stores.

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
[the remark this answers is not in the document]
iglendd: Let's try to remove it and see what linter would say.

cmd/agent/dist/conf.d/windows_certificate.d/conf.yaml.example
```diff
@@ -0,0 +1,88 @@
...
+    ##
+    ## Overrides any `service` defined in the `init_config` section.
+    #
+    # service: <SERVICE>
+
+    ## @param min_collection_interval - number - optional - default: 15
+    ## This changes the collection interval of the check. For more information, see:
+    ## https://docs.datadoghq.com/developers/write_agent_check/#collection-interval
+    #
+    # min_collection_interval: 15
```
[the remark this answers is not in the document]
lrbhatia: Thanks! Certificates don’t change that often and are not time-sensitive to the second.

rahulkaukuntla (approved):

julien-lebot (approved):
lgtm

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
```diff
@@ -0,0 +1,399 @@
...
+	if err != nil {
+		return err
+	}
+	if len(certificates) == 0 {
+		log.Warnf("No certificates found in store: %s for subject filters: '%s'", w.config.CertificateStore, strings.Join(w.config.CertSubjects, ", "))
+	}
+
+	for _, cert := range certificates {
+		log.Debugf("Found certificate: %s", cert.Subject.String())
+		daysRemaining := getCertExpiration(cert)
```
iglendd, line 160: We now can drop float conversion in the called function and below,

iglendd (commented):
Minor comment about float conversion (leftovers from previous code) which becomes unnesesary

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
mrafi97: @iglendd The float conversions on lines 175 and 181 are needed because GO does not allow comparisons of ints and float64. In this case the config accepts the parameters as `DaysCritical` and `DaysWarning` parameters as int and `daysRemaining` needs to be a float64 for submission as a metric.

iglendd: Right, but you can also return int and use everywhere int except line 167 which you can convert to float (also I am not sure if implicit conversion may work). Also "%.2f" formatting is not necessary, IMHO.

mrafi97: @iglendd 
Ah ok that makes sense. I still think it is worth keeping the original float value of daysRemaining so maybe creating a separate variable called `daysRemainingInt := int(daysRemaining)` and using that to compare with `DaysCritical` and `DaysWarning` is better? The floating point is more accurate than an int, especially for when certificates are less than 1 day left on their expiration.

iglendd: Was wrong about assertion above. Float for the value should stay float indeed for reporting. Please ignore above.

iglendd (approved):

pgimalac (approved):

PR #36670, opened 2025-05-05
[OTAGENT-392] Set EVP origin for logs coming from DDOT

liustanley:
### What does this PR do?
Sets EVP intake origin to `ddot` for logs coming from DDOT and `otel-collector` for logs coming from the OSS OTel Collector (to be implemented in upstream PR). This is done by adding `IntakeOrigin` to the logs agent component dependencies, which is a parameter for `NewLogsAgent` that is called in both DDOT and in [upstream](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/datadogexporter/logs_exporter.go#L190-L195).

### Motivation
To be able to differentiate between logs coming from DDOT and OSS OTel Collector.

Added unit test.

### DIFF
--- a/comp/otelcol/logsagentpipeline/logsagentpipelineimpl/agent_test.go
+++ b/comp/otelcol/logsagentpipeline/logsagentpipelineimpl/agent_test.go
@@ -186,7 +186,8 @@ func TestBuildEndpoints(t *testing.T) {
 		fx.Provide(func() log.Component { return logmock.New(t) }),
 	))
 
-	endpoints, err := buildEndpoints(deps.Config, deps.Log)
+	endpoints, err := buildEndpoints(deps.Config, deps.Log, config.OTelCollectorIntakeOrigin)
 	assert.Nil(t, err)
 	assert.Equal(t, "agent-intake.logs.datadoghq.com", endpoints.Main.Host)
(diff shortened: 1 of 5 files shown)


truthbk (approved):
We'll need a PR upstream to set `config.OtelCollectorIntakeOrigin` in the OSS collector.

ddrthall (approved):

comp/otelcol/logsagentpipeline/logsagentpipelineimpl/agent_test.go
```diff
@@ -185,8 +186,10 @@ func TestBuildEndpoints(t *testing.T) {
 		configComponent.MockModule(),
 		fx.Provide(func() log.Component { return logmock.New(t) }),
 	))
+	deps.Config.Set("logs_config.force_use_http", true, pkgconfigmodel.SourceDefault)
```
truthbk, line 189: Why did this change sneak in? Looks unrelated?

liustanley: This test was failing without this change because TCP doesn't support any headers (including the intake origin header). We also use this setting in both DDOT and OSS Collector

truthbk (approved):

PR #36672, opened 2025-05-05
[CNM-4402] Report interface mac address for routes

hmahmood:
### What does this PR do?

Reports the interface mac address for a route when subnet ID is not available. This is primarily to support gateway resolution on AWS Fargate since on that platform the local ec2 metadata endpoint does not expose subnet or network ID. The mac address can be used to lookup the subnet/network ID in the backend.

Related PR: https://github.com/DataDog/agent-payload/pull/384

### DIFF
--- a/pkg/network/payload/types.go
+++ b/pkg/network/payload/types.go
@@ -8,7 +8,13 @@ package payload
 
 // Via has info about the routing decision for a flow
 type Via struct {
-	Subnet Subnet `json:"subnet,omitempty"`
+	Subnet    Subnet    `json:"subnet,omitempty"`
+	Interface Interface `json:"interface,omitempty"`
+}
+
+// Interface has information about a network interface
+type Interface struct {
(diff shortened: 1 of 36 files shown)


pkg/network/payload/types.go
```diff
@@ -8,7 +8,13 @@ package payload
 
 // Via has info about the routing decision for a flow
 type Via struct {
-	Subnet Subnet `json:"subnet,omitempty"`
+	Subnet    Subnet    `json:"subnet,omitempty"`
```
pimlu, line 11: Now that Via also has Interface in it, is there still a distinction between `network.Via` and `model.Route`?

pimlu (approved):

pkg/network/payload/types.go
hmahmood: I don't think was a distinction even before actually (without interface). I think we can likely get rid of Via and just use `model.Route`, but that likely would be a good follow up.

PR #36688, opened 2025-05-06
[AGENTRUN-286] Add E2E Test for multiprocessing Usage with RTLoader

louis-cqrl:
### What does this PR do?

This PR introduces an end-to-end test that verifies the use of Python's multiprocessing library in conjunction with rtloader. The test ensures that custom checks leveraging multiprocessing continue to function correctly within the Agent's runtime environment.

### Motivation

To validate that the integration between rtloader and the multiprocessing module remains compatible, especially for custom checks that rely on parallel execution.

N/A


test/new-e2e/tests/agent-runtimes/rtloader/python-check/multi_file_check.py
pgimalac, line 1: Why not just emit a metric tagged by the PID ? It would avoid all the logic around writing to files and then reading + parsing

Also note that if you use `agent check --json` subcommand, you get a json containing all the metrics emitted in this run, so very easy to parse

pgimalac: Also I think if we want to test `rtloader` it makes sense to have data flowing back to the agent from the processes, otherwise we're not really testing `rtloader`, just the `multiprocessing` module

test/new-e2e/tests/agent-runtimes/rtloader/python-check/multi_file_check.py
```diff
@@ -0,0 +1,58 @@
...
+        print(f"[Process {os.getpid()}] Error writing to {file_path}: {str(e)}")
+
+
+class MultiFileCheck(AgentCheck):
+    def check(self, instance):
+        """
+        Spawns 3 processes that concurrently write to 3 different files.
+        The files are created in the system's temporary directory.
+        """
+        temp_dir = os.path.join(tempfile.gettempdir(), "multi_file_check")
```
pgimalac, line 39: If we want to go with this implementation, the full path of the tmpdir could be configured in the check config, that way you don't need to hardcode it in Python and Go, you can generate a directory for this specific test and configure the check to use it

test/new-e2e/tests/agent-runtimes/rtloader/python-check/multi_file_check.py
```diff
@@ -0,0 +1,58 @@
...
+    printing process information to demonstrate concurrent execution.
+    """
+    start_time = time.time()
+    print(f"[Process {os.getpid()}] Starting to write to {file_path} at {time.strftime('%X')}")
+    try:
+        with open(file_path, 'a') as f:
+            for i in range(write_count):
+                message = (
+                    f"[Process {os.getpid()}] Finished writing to {file_path} at {time.strftime('%X')} at line {i + 1}"
+                )
```
pgimalac, line 21: **❓ question:** ‏What's the point of writing multiple times ? It doesn't seem to be checked, we just read the first line

test/new-e2e/tests/agent-runtimes/rtloader/multiprocessing_lib_common_test.go
```diff
@@ -0,0 +1,91 @@
...
+	var suiteOptions []e2e.SuiteOption
+	suiteOptions = append(suiteOptions, e2e.WithProvisioner(
+		awshost.Provisioner(
+			awshost.WithAgentOptions(
+				agentparams.WithFile(v.confdPath, `
+init_config:
+
+instances:
+  - name: "default"
+`, true),
```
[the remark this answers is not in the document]
louis-cqrl: did something let me know if this fits in what you were thinking

test/new-e2e/tests/agent-runtimes/rtloader/multiprocessing_lib_common_test.go
```diff
@@ -0,0 +1,91 @@
...
+	))
+
+	return suiteOptions
+}
+
+func (v *baseMultiProcessingLibSuite) TestMultiProcessingLib() {
+	v.T().Log("Running MultiProcessingLib test")
+
+	// Wait for the check to run and verify its status
+	time.Sleep(10 * time.Second) // Give time for the check to run
```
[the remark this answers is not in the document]
louis-cqrl: done

test/new-e2e/tests/agent-runtimes/rtloader/python-check/multi_pid_check.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    pid = os.getpid()
+    queue.put(pid)
+
+
+class MultiPIDCheck(AgentCheck):
+    def check(self, instance):
+        """
+        Spawns 3 processes that emit their PIDs as metrics.
+        """
```
pgimalac, line 19: The processes don't emit anything right ? They just add their PIDs to the queue, and only the main process will emit the metrics ?

Why not pass the check as argument and let them emit the metric directly ?

louis-cqrl: right I will update the comment, the datadog agent check cannot emit metric in subprocess that's why we are adding to the queue to push them after

pgimalac (approved):
LGTM !

PR #36692, opened 2025-05-06
[SBOM] Factorize code that handles host SBOM

lebauce:
### What does this PR do?

Factorize code that handles host SBOM.

### Motivation

Code was duplicated for both regular and procfs filesystem scans.

### DIFF
--- a/pkg/collector/corechecks/sbom/processor.go
+++ b/pkg/collector/corechecks/sbom/processor.go
@@ -50,7 +50,7 @@ type processor struct {
 	hostSBOM              bool
 	procfsSBOM            bool
 	hostname              string
-	hostCache             string
+	hostLastSBOMHash      string
 	hostLastFullSBOM      time.Time
 	hostHeartbeatValidity time.Duration
 }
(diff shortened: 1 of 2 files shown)


pkg/collector/corechecks/sbom/processor.go
```diff
@@ -210,49 +210,60 @@ func (p *processor) processContainerImagesRefresh(allImages []*workloadmeta.Cont
...
-		log.Errorf("Scan error: %v", result.Error)
-		sbom.Sbom = &model.SBOMEntity_Error{
-			Error: result.Error.Error(),
+	defer func() {
+		err := result.Error
+
+		if errCallback != nil {
+			if err = errCallback(err); err != nil {
+				log.Errorf("failed scan: %s", err)
+				return
```
paulcacheux, line 230: doesn't this return prevent the sbom to be put in the queue ?

paulcacheux (changes requested):

pkg/collector/corechecks/sbom/processor.go
```diff
@@ -210,49 +210,60 @@ func (p *processor) processContainerImagesRefresh(allImages []*workloadmeta.Cont
...
+	report, err := result.Report.ToCycloneDX()
+	if err != nil {
+		result.Error = err
+		return
+	}
+
+	sbom.Hash = string(result.Report.ID())
+
+	log.Infof("Successfully generated SBOM for %s: %v, %v", kind, result.CreatedAt, result.Duration)
+	if p.hostLastSBOMHash != "" && p.hostLastSBOMHash == sbom.Hash && result.CreatedAt.Sub(p.hostLastFullSBOM) < p.hostHeartbeatValidity {
```
paulcacheux, line 254: Do we want this host hash stuff in the procfs code path ?

pkg/collector/corechecks/sbom/processor.go
```diff
@@ -210,49 +210,60 @@ func (p *processor) processContainerImagesRefresh(allImages []*workloadmeta.Cont
...
+		sbom.Heartbeat = true
+	} else {
+		sbom.Sbom = &model.SBOMEntity_Cyclonedx{
+			Cyclonedx: convertBOM(report),
 		}
+		p.hostLastSBOMHash = sbom.Hash
```
paulcacheux, line 260: won't this touch the host hash when operating on a SBOM from the procfs ?

pkg/collector/corechecks/sbom/processor.go
lebauce: I would say yes

pkg/collector/corechecks/sbom/processor.go
lebauce: If the errCallback returns an error, yes.
This is to mimic the actual behaviour : prevent the SBOM to be put in the queue when we receive procfs.ErrNotFound

PR #36693, opened 2025-05-06
[ACIX-578] Report versions for macos gitlab runners

CelianR:
### What does this PR do?

This reports current go / python version to Datadog to prepare a job cleaning those versions.


.gitlab/common/macos.yml
```diff
@@ -1,21 +1,11 @@
...
-    echo "Don't forget to regularly delete Python unused versions. Here are the installed Python versions and their disk space on the runner:"
-    echo "Python:"
-    du -sh $(pyenv root)/versions/*
-    echo "To remove a Python version please run:"
-    echo "pyenv uninstall -f <version>"
+.macos_clean_env:
+  # Report current version to datadog
+  - DD_API_KEY="$("$CI_PROJECT_DIR"/tools/ci/fetch_secret.sh "$AGENT_API_KEY_ORG2" token)" || exit $?; export DD_API_KEY
+  - dda inv -- -e macos.report-versions -l python || true
+  - dda inv -- -e macos.report-versions -l go || true
```
CelianR, line 8: Notes:
1. We don't want this to fail due to API failures
2. This section will be updated in #36757, thus the name

spencergilbert: Why not retry instead of forcing non-failure?

spencergilbert (approved):

.gitlab/common/macos.yml
CelianR: Not reporting this has no high impact so I'm not sure it's really necessary, especially if we waste time by doing so

.gitlab/common/macos.yml
```diff
@@ -1,21 +1,11 @@
...
-    echo "gimme uninstall <version>"
-
-.list_python_versions_commands:
-  - |
-    echo "Don't forget to regularly delete Python unused versions. Here are the installed Python versions and their disk space on the runner:"
-    echo "Python:"
-    du -sh $(pyenv root)/versions/*
-    echo "To remove a Python version please run:"
-    echo "pyenv uninstall -f <version>"
+.macos_clean_env:
```
chouetz, line 4: 🥜 
```suggestion
.macos_reporting:
```

.gitlab/common/macos.yml
```diff
@@ -102,9 +92,8 @@
     # Selecting the current Python version
     - !reference [.select_python_env_commands]
     - !reference [.install_python_dependencies]
-    # List Python and Go existing environments and their disk space
-    - !reference [.list_go_versions_commands]
-    - !reference [.list_python_versions_commands]
+    # See there is no virtualization, we need to clean the environment regularly
+    - !reference [.macos_clean_env]
```
chouetz, line 96:
```suggestion
    - !reference [.macos_reporting]
```

.gitlab/package_build/dmg.yml
```diff
@@ -56,6 +64,9 @@ agent_dmg-x64-a7:
     - set -eo pipefail
     - !reference [.vault_login]
     - !reference [.macos_setup_python]
+    - !reference [.macos_setup_go]
+    # List Python and Go existing environments and their disk space
+    - !reference [.macos_clean_env]
```
chouetz, line 69:
```suggestion
    - !reference [.macos_reporting]
```

tasks/macos.py
```diff
@@ -0,0 +1,29 @@
...
+from invoke import task
+
+from tasks.libs.common.datadog_api import create_count, send_metrics
+
+
+@task
+def report_versions(ctx, lang):
+    """Reports the current Python version to Datadog."""
+
+    assert lang in ("python", "go"), "Lang must be either 'python' or 'go'"
```
chouetz, line 12: I'm wondering if this is a good idea to do 2 calls instead of one (considering we can send the 2 metrics for `go` and `python` at once.
Besides I'm wondering if we should also ensure to do this report only once per pipeline or only on `main` or even only once a day, wdyt?

chouetz (approved):
Overall LGTM, and I'm not sure it's a real saving but we can probably limit the number of time we send this metric, let me know what you think, thanks in advance

tasks/macos.py
CelianR: That's true, it's maybe better to do this only on main / release branches. The only drawback is that it will affect the time it takes on main / release branches but I think it is better like this 👍 I'll add an option to send all at once

PR #36697, opened 2025-05-06
Collect EC2 instance information as host tags

hush-hush:
### What does this PR do?

Some customers can't or don't want to enable the AWS integration we now can collect basic instance information as host tags. This new feature is disabled by default behind the `collect_ec2_instance_info` setting.

Start a new EC2 instance, set `collect_ec2_instance_info` to true and check the V5 payload contains the new host tags.

### DIFF
--- a/pkg/config/config_template.yaml
+++ b/pkg/config/config_template.yaml
@@ -397,19 +397,37 @@ api_key:
 ## https://docs.datadoghq.com/integrations/faq/how-do-i-pull-my-ec2-tags-without-using-the-aws-integration/
 #
 # collect_ec2_tags: false
+#
+
+## @param collect_ec2_instance_info - boolean - optional - default: false
+## @env DD_COLLECT_EC2_INSTANCE_INFO - boolean - optional - default: false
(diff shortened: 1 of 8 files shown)


rahulkaukuntla (approved):

pkg/config/config_template.yaml
```diff
@@ -397,19 +397,37 @@ api_key:
...
+## Extend host tags with AWS EC2 instance information. The added tags are:
+##  - region
+##  - instance-type
+##  - aws_account
+##  - image
+##  - availability-zone
+##
+## This should only be enabled when the Datadog AWS integration cannot be enabled (see
+## https://docs.datadoghq.com/integrations/amazon_web_services/ for more information on the AWS integration).
+## Using the AWS integration is recommended as it offers more a better integration with AWS environment.
```
maycmlee, line 413:
```suggestion
## Using the AWS integration is recommended as it offers more features and a better integration with the AWS environment.
```

maycmlee (approved):
Just a couple of small suggestions, but approving to not block merge.

louis-cqrl (approved):
Small comment LGTM otherwise

PR #36698, opened 2025-05-06
Gzip quality gate test

DDuongNguyen:


pkg/config/setup/config.go
```diff
@@ -82,7 +82,7 @@ const (
 	DefaultCompressorKind = "zstd"
 
 	// DefaultLogCompressionKind is the default log compressor. Options available are 'zstd' and 'gzip'
-	DefaultLogCompressionKind = "zstd"
+	DefaultLogCompressionKind = "gzip"
```
hush-hush, line 85: That would change the default log compression for all customers, are we sure we want to do this ?

scottopell: For context, this is testing a revert of https://github.com/DataDog/datadog-agent/commit/db09bbc371b18d3eab5d57b470b8f374a1195da4#diff-8bfed0fac4b33289ce877620604de6aeb9370ffcb6562149627a4a22c222be96 which already landed

PR #36705, opened 2025-05-06
[AGNTLOG-176] Add compression override over config for epforwaders 

DDuongNguyen:
### What does this PR do?
This PR modifies the event platform forwarder to force GZIP compression for DBM (Database Monitoring) events, overriding any existing configuration. This change ensures that DBM events are always compressed using GZIP instead of ZSTD, as the DBM backend currently does not support ZSTD compression.

Changes include:
- Added `forcedCompression` field to `LogsConfigKeys` struct to support compression override
- Added `WithForcedCompression` method to `LogsConfigKeys` to set forced compression
- Modified `compressionKind` method to respect forced compression settings
- Updated all DBM pipeline descriptions to use GZIP compression
- Added appropriate debug logging for compression type changes

### Motivation
The DBM backend currently does not support ZSTD compression, but our default configuration uses ZSTD. 

This causes issues when DBM events are compressed with ZSTD and sent to the backend. 

By forcing GZIP compression specifically for DBM events (and leave options for other events if necessary), we ensure compatibility with the current backend capabilities while allowing other event types to continue using ZSTD compression as configured.

1. Manual Testing:
   - Configured agent with ZSTD compression
   - Verified DBM events are sent with GZIP compression despite ZSTD configuration
   - Verified non-DBM events still use configured compression type
   - Monitored debug logs to confirm correct compression type is being used
   - Verified DBM backend successfully processes the GZIP compressed events

2. Validation Steps:
   ```bash
   # Run the agent with debug logging enabled
   DD_LOG_LEVEL=debug ./bin/agent/agent run

   # Verify compression type in logs
   # Should see "Initialized event platform forwarder pipeline .... eventType=dbm-* compression_kind=gzip, compression_level=6"

   # Instead of "Initialized event platform forwarder pipeline .... eventType=dbm-* compression_kind=zstd, compression_level=1"
   ```

### Additional Notes


comp/forwarder/eventplatform/eventplatformimpl/epforwarder.go
```diff
@@ -388,6 +393,9 @@ func newHTTPPassthroughPipeline(
 ) (p *passthroughPipeline, err error) {
 
 	configKeys := config.NewLogsConfigKeys(desc.endpointsConfigPrefix, coreConfig)
+	if desc.forceCompressionKind != "" {
```
ddrthall, line 396: Instead of extending the config entity for third way to customize values coming in from desc, would it make sense to either mimic desc.BatchMaxConcurrentSend's customization flow or desc.intakeTrackType's usage?

DDuongNguyen: I made some updates and decided to keep the `WithForcedCompression` flow since the pipeline still needs to go through `config_keys.go`. Rather than removing it, I tweaked it to return "none" when forced compression is set (since we'll override the actual compression settings later in `newHTTPPassthroughPipeline` anyway).

Let me know if you think we should remove this path entirely since the compression settings will be overridden.

comp/logs/agent/config/config_keys.go
```diff
@@ -118,7 +118,10 @@ func (l *LogsConfigKeys) devModeUseProto() bool {
 	return l.getConfig().GetBool(l.getConfigKey("dev_mode_use_proto"))
 }
 
-func (l *LogsConfigKeys) compressionKind() string {
+func (l *LogsConfigKeys) compressionKind(forceCompressionKind string) string {
```
ddrthall, line 121: I'm still of the opinion that we don't want the config_keys entity to care about anything but config management. All other overrides of all other config values are handled in other locations.

comp/logs/agent/config/config.go
```diff
@@ -186,11 +195,16 @@ func BuildHTTPEndpointsWithVectorOverride(coreConfig pkgconfigmodel.Reader, inta
 }
 
 // BuildHTTPEndpointsWithConfig uses two arguments that instructs it how to access configuration parameters, then returns the HTTP endpoints to send logs to. This function is able to default to the 'classic' BuildHTTPEndpoints() w ldHTTPEndpointsWithConfigdefault variables logsConfigDefaultKeys and httpEndpointPrefix
-func BuildHTTPEndpointsWithConfig(coreConfig pkgconfigmodel.Reader, logsConfig *LogsConfigKeys, endpointPrefix string, intakeTrackType IntakeTrackType, intakeProtocol IntakeProtocol, intakeOrigin IntakeOrigin) (*Endpoints, error) {
+func BuildHTTPEndpointsWithConfig(coreConfig pkgconfigmodel.Reader, logsConfig *LogsConfigKeys, endpointPrefix string, intakeTrackType IntakeTrackType, intakeProtocol IntakeProtocol, intakeOrigin IntakeOrigin, opts ...EndpointCompressionOptionFunc) (*Endpoints, error) {
```
ddrthall, line 198: While variadic options can be quite powerful in certain contexts, go strongly encourages handing cases like this with default values in structs or by creating multiple functions with different signatures to deal with default arguments. I don't think that this is cleaner than having epforwarder having a desc.EndpointCompressionOption that gets passed in directly.

comp/forwarder/eventplatform/eventplatformimpl/epforwarder.go
```diff
@@ -440,7 +454,7 @@ func newHTTPPassthroughPipeline(
 	var encoder compressioncommon.Compressor
 	encoder = compressor.NewCompressor("none", 0)
 	if endpoints.Main.UseCompression {
-		encoder = compressor.NewCompressor(endpoints.Main.CompressionKind, endpoints.Main.CompressionLevel)
+		encoder = compressor.NewCompressor(endpoints.CompressionKind, endpoints.CompressionLevel)
```
ddrthall, line 457: Neither of these values seem to be set correctly, we appear to still be providing the defaults on endpoints.Main.

comp/logs/agent/config/config.go
```diff
@@ -177,20 +177,20 @@ func buildTCPEndpoints(coreConfig pkgconfigmodel.Reader, logsConfig *LogsConfigK
...
 }
 
 // BuildHTTPEndpointsWithConfig uses two arguments that instructs it how to access configuration parameters, then returns the HTTP endpoints to send logs to. This function is able to default to the 'classic' BuildHTTPEndpoints() w ldHTTPEndpointsWithConfigdefault variables logsConfigDefaultKeys and httpEndpointPrefix
-func BuildHTTPEndpointsWithConfig(coreConfig pkgconfigmodel.Reader, logsConfig *LogsConfigKeys, endpointPrefix string, intakeTrackType IntakeTrackType, intakeProtocol IntakeProtocol, intakeOrigin IntakeOrigin) (*Endpoints, error) {
+func BuildHTTPEndpointsWithConfig(coreConfig pkgconfigmodel.Reader, logsConfig *LogsConfigKeys, endpointPrefix string, intakeTrackType IntakeTrackType, intakeProtocol IntakeProtocol, intakeOrigin IntakeOrigin, compressionOptions EndpointCompressionOptions) (*Endpoints, error) {
 	// Provide default values for legacy settings when the configuration key does not exist
 	defaultNoSSL := logsConfig.logsNoSSL()
 
-	main := newHTTPEndpoint(logsConfig)
+	main := newHTTPEndpoint(logsConfig, compressionOptions)
```
ddrthall, line 193: Nit: Any reason we pass down the compressionOptions instead of overriding the constructed main below like we do with intakeTrackType?

comp/forwarder/eventplatform/eventplatformimpl/epforwarder.go
```diff
@@ -66,7 +66,9 @@ var passthroughPipelineDescs = []passthroughPipelineDesc{
 		defaultBatchMaxContentSize:    10e6,
 		defaultBatchMaxSize:           pkgconfigsetup.DefaultBatchMaxSize,
 		// High input chan size is needed to handle high number of DBM events being flushed by DBM integrations
-		defaultInputChanSize: 500,
+		defaultInputChanSize:  500,
+		forceCompressionKind:  config.GzipCompressionKind,
```
ddrthall, line 70: Excessively ignorable nit: We can bake the EndpointCompressionType here instead of having a string/int that need translation.

ddrthall:
General Question: Since we're opting to keep compression kind as a endpoint level entity (rather than at the aggregate "endpoints" level), we want to make sure that all related endpoints are correctly getting that value. Are we sure that these changes are enough to guarantee all additional endpoints are set with the correct compression kind?

comp/logs/agent/config/endpoints.go
```diff
@@ -177,15 +184,27 @@ func loadTCPAdditionalEndpoints(main Endpoint, l *LogsConfigKeys) []Endpoint {
 func loadHTTPAdditionalEndpoints(main Endpoint, l *LogsConfigKeys, intakeTrackType IntakeTrackType, intakeProtocol IntakeProtocol, intakeOrigin IntakeOrigin) []Endpoint {
 	additionals, configKeyUsed := l.getAdditionalEndpoints()
 
+	// Check if we need gzip compatibility for additional endpoints
+	needsGzipCompatibility := len(additionals) > 0 && !l.config.IsConfigured(l.getConfigKey("compression_kind"))
```
ddrthall, line 188: Do we need this logic? My understanding is that main will already have this built in.

comp/logs/agent/config/config.go
```diff
@@ -177,16 +177,16 @@ func buildTCPEndpoints(coreConfig pkgconfigmodel.Reader, logsConfig *LogsConfigK
...
+	return BuildHTTPEndpointsWithConfig(coreConfig, defaultLogsConfigKeysWithVectorOverride(coreConfig), httpEndpointPrefix, intakeTrackType, intakeProtocol, intakeOrigin, EndpointCompressionOptions{})
 }
 
 // BuildHTTPEndpointsWithConfig uses two arguments that instructs it how to access configuration parameters, then returns the HTTP endpoints to send logs to. This function is able to default to the 'classic' BuildHTTPEndpoints() w ldHTTPEndpointsWithConfigdefault variables logsConfigDefaultKeys and httpEndpointPrefix
-func BuildHTTPEndpointsWithConfig(coreConfig pkgconfigmodel.Reader, logsConfig *LogsConfigKeys, endpointPrefix string, intakeTrackType IntakeTrackType, intakeProtocol IntakeProtocol, intakeOrigin IntakeOrigin) (*Endpoints, error) {
+func BuildHTTPEndpointsWithConfig(coreConfig pkgconfigmodel.Reader, logsConfig *LogsConfigKeys, endpointPrefix string, intakeTrackType IntakeTrackType, intakeProtocol IntakeProtocol, intakeOrigin IntakeOrigin, compressionOptions EndpointCompressionOptions) (*Endpoints, error) {
```
ddrthall, line 189: Instead of overriding the public interface (which requires changing a lot of end users), we can add a new public function with a different signature and have both that new function and BuildHTTPEndpointWithConfig call the same private function. For example:
```
BuildHTTPEndpointsWithCompressionOverride(..., compression *EndpointCompression) {
     buildHTTPEndpoints(..., compression)
}
BuildHTTPEndpointsWithConfig(...) {
    buildHTTPEndpoints(..., nil)
}

buildHTTPEndpoints(..., compression *EndpointCompression) {
// all of the actual logic
}
```

comp/logs/agent/config/config.go
```diff
@@ -201,6 +201,11 @@ func BuildHTTPEndpointsWithConfig(coreConfig pkgconfigmodel.Reader, logsConfig *
 		main.Version = EPIntakeVersion1
 	}
 
+	if compressionOptions.CompressionKind != "" {
+		main.CompressionKind = compressionOptions.CompressionKind
+		main.CompressionLevel = compressionOptions.CompressionLevel
```
ddrthall, line 206: We probably want to carry this main setting over to the mrf endpoint creation below as well.

comp/logs/agent/config/endpoints.go
```diff
@@ -183,7 +190,6 @@ func loadHTTPAdditionalEndpoints(main Endpoint, l *LogsConfigKeys, intakeTrackTy
 
 		newE.isAdditionalEndpoint = true
 		newE.additionalEndpointsIdx = idx
-
 		newE.UseCompression = main.UseCompression
 		newE.CompressionLevel = main.CompressionLevel
```
ddrthall, line 194: We'll want the compression kind override here as well

ddrthall (approved):

PR #36711, opened 2025-05-06
feat: use restrictedSecurityContext by default

stanistan:
### What does this PR do?

https://datadoghq.atlassian.net/browse/INPLAT-492

If we detect that a namespace enforces a restricted security policy we can default to having a restricted securitycontext to apply for the init containers instead of having our users have to provide it as an override.

### Motivation

One fewer issue a customer can run into when setting up SSI in a restricted security environment, we can provide a security context for them.

- Added and updated unit tests. 
- using `injector-dev` to create a namespace with the security policy restriction and testing that pods startup as expected using SSI.

```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: restricted
  labels:
    pod-security.kubernetes.io/enforce: restricted
```

Without this patch, pods will not be started at all.

### Additional Notes

Changed the behavior of `initSecurityContext` to _not be an empty one_, but `nil` when it is not present, this simplifies tests and makes it easier to see if we can override it based on the namespace security policy.

### DIFF
--- a/pkg/clusteragent/admission/mutate/autoinstrumentation/target_mutator.go
+++ b/pkg/clusteragent/admission/mutate/autoinstrumentation/target_mutator.go
@@ -17,7 +17,6 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/dynamic"
 
-	"github.com/DataDog/datadog-agent/comp/core/workloadmeta/collectors/util"
 	workloadmeta "github.com/DataDog/datadog-agent/comp/core/workloadmeta/def"
 	"github.com/DataDog/datadog-agent/pkg/clusteragent/admission/metrics"
 	mutatecommon "github.com/DataDog/datadog-agent/pkg/clusteragent/admission/mutate/common"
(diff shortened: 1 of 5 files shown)


pkg/clusteragent/admission/mutate/autoinstrumentation/target_mutator.go
```diff
@@ -331,15 +330,13 @@ func (m *TargetMutator) getMatchingTarget(pod *corev1.Pod) *targetInternal {
 func (t targetInternal) matchesNamespaceSelector(namespace string) (bool, error) {
 	// If we are using the namespace selector, check if the namespace matches the selector.
 	if t.useNamespaceSelector {
-		// Get the namespace metadata.
-		id := util.GenerateKubeMetadataEntityID("", "namespaces", "", namespace)
-		ns, err := t.wmeta.GetKubernetesMetadata(id)
+		nsLabels, err := getNamespaceLabels(t.wmeta, namespace)
 		if err != nil {
-			return false, fmt.Errorf("could not get kubernetes namespace to match against for %s: %w", namespace, err)
+			return false, fmt.Errorf("could not get lababels to match: %w", err)
```
iamluc, line 335: typo

estherk15 (approved):

betterengineering (approved):

PR #36712, opened 2025-05-06
APM: Improve performance of QuantizeIP (APMSP-1986)

ajgajg1134:
### What does this PR do?
Reduce CPU time and allocations of the QuantizeIPAddresses function. (Spotted by @paullegranddc )

Primarily there are 2 main optimizations here (viewable as separate commits:
1. Remove regexes and instead use strings.Index in a short loop
2. Do not use net.SplitHostPort as it allocates an error when the incoming value isn't parseable, (and we do not care about the exact error just that it exists).

I also investigated moving from strings.Split to using string.Cut and a string builder however it was not a clear-cut improvement. In the case where there is no comma in the incoming string it worsens performance.

### Motivation
A non-zero number of object allocations are done by this code and it is run against for all spans that are used to calculate stats (thus this happens before sampling and so there is a larger impact here compared to post-sampling features).

Existing unit tests have good coverage, to ensure no panics were inadvertently introduced I also added and ran a fuzz test.

I then ran and compared benchmarks using benchstat:
Comparing the regular expression changes shows a large CPU improvement when a match is found and removes an allocation:
```
goos: darwin
goarch: arm64
pkg: github.com/DataDog/datadog-agent/pkg/obfuscate
cpu: Apple M1 Max
                            │ prefixold.out │             prefix2.out             │
                            │    sec/op     │   sec/op     vs base                │
SplitPrefix/matching-10        354.90n ± 0%   13.86n ± 1%  -96.09% (p=0.000 n=10)
SplitPrefix/not_matching-10     37.08n ± 1%   37.26n ± 1%        ~ (p=0.147 n=10)
geomean                         114.7n        22.72n       -80.19%

                            │ prefixold.out │               prefix2.out               │
                            │     B/op      │    B/op     vs base                     │
SplitPrefix/matching-10        32.00 ± 0%      0.00 ± 0%  -100.00% (p=0.000 n=10)
SplitPrefix/not_matching-10    0.000 ± 0%     0.000 ± 0%         ~ (p=1.000 n=10) ¹
geomean                                   ²               ?                       ² ³
¹ all samples are equal
² summaries must be >0 to compute geomean
³ ratios must be >0 to compute geomean

                            │ prefixold.out │               prefix2.out               │
                            │   allocs/op   │ allocs/op   vs base                     │
SplitPrefix/matching-10        1.000 ± 0%     0.000 ± 0%  -100.00% (p=0.000 n=10)
SplitPrefix/not_matching-10    0.000 ± 0%     0.000 ± 0%         ~ (p=1.000 n=10) ¹
```

Then the larger  benchmark comparing using net split instead of our no-error-alloc version shows we reduce allocations when the incoming value isn't a "host:port".
```
goos: darwin
goarch: arm64
pkg: github.com/DataDog/datadog-agent/pkg/obfuscate
cpu: Apple M1 Max
                           │ useNetSplit.out │          ReplaceParse.out          │
                           │     sec/op      │   sec/op     vs base               │
QuantizePeerIpAddresses-10       14.84µ ± 2%   13.94µ ± 0%  -6.05% (p=0.000 n=10)

                           │ useNetSplit.out │           ReplaceParse.out           │
                           │      B/op       │     B/op      vs base                │
QuantizePeerIpAddresses-10      3.492Ki ± 0%   2.617Ki ± 0%  -25.06% (p=0.000 n=10)

                           │ useNetSplit.out │          ReplaceParse.out          │
                           │    allocs/op    │ allocs/op   vs base                │
QuantizePeerIpAddresses-10        150.0 ± 0%   122.0 ± 0%  -18.67% (p=0.000 n=10)
```


releasenotes/notes/ImproveQuantizePeerIPAddressesPerformance-5902d8baa33cda4a.yaml
```diff
@@ -0,0 +1,11 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+enhancements:
+  - |
+    APM: Improve the performance of the trace-agent's QuantizePeerIPAddresses function, this should result in a marginal reduction in CPU usage for most workloads.
```
buraizu, line 11:
```suggestion
    APM: Improve the performance of the Trace Agent's QuantizePeerIPAddresses function, providing a marginal reduction in CPU usage for most workloads.
```

buraizu (approved):
Approving with some updates requested for the release note

ichinaski (approved):

PR #36718, opened 2025-05-07
[cmd/cluster-agent] Fail to start DCA if autoscaling enabled and no cluster name

jennchenn:
### What does this PR do?

Prevents the cluster agent from starting up in the case that autoscaling is enabled and no cluster name is detected.

### Motivation

Not having the cluster name causes some weird behavior with autoscaling - failing to start the cluster agent at this point prevents the user from encountering issues during regular operation. 

1. Deployed to minikube with no `clusterName` set
2. Verified that when autoscaling enabled, cluster agent fails to start with log `Failed to start: autoscaling is enabled but no cluster name detected, exiting`
3. Verified that if autoscaling not enabled, cluster agent starts with log `Failed to auto-detect a Kubernetes cluster name. We recommend you set it manually via the cluster_name config option`
4. Verified that if `clusterName` is set, cluster agent starts even if autoscaling enabled
```
datadog:
  clusterName: jenn-test
  logLevel: DEBUG
  logs:
    containerCollectAll: true
    containerCollectUsingFiles: true
    enabled: true
  kubelet:
    tlsVerify: false
  autoscaling:
    workload:
      enabled: true
...
```

### Additional Notes

A case could be made for failing to start the cluster agent whenever the cluster name is not detected (regardless of whether autoscaling is enabled/not). At this time it is not done.

### DIFF
--- a/cmd/cluster-agent/subcommands/start/command.go
+++ b/cmd/cluster-agent/subcommands/start/command.go
@@ -377,6 +377,9 @@ func start(log log.Component,
 		pkglog.Errorf("Failed to generate or retrieve the cluster ID, err: %v", err)
 	}
 	if clusterName == "" {
+		if config.GetBool("autoscaling.workload.enabled") {
+			return fmt.Errorf("Failed to start: autoscaling is enabled but no cluster name detected, exiting")
+		}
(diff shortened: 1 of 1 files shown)


gabedos (approved):
Looks good to me to keep the failure scope just for when workload autoscaling is enabled

vboulineau (approved):

PR #36727, opened 2025-05-07
omnibus: openssl: disable zlib on windows

chouquette:
### What does this PR do?

Disable OpenSSL's zlib support on Windows

### Motivation
We do not ship the DLL so this is bound to break as soon as compression is needed


alopezz (approved):

alopezz:
Note: I amended the existing commit claiming authorship so that it gets signed and can get properly merged.

PR #36731, opened 2025-05-07
Rename Disk Check Struct to Maintain Consistent Naming

louis-cqrl:
### What does this PR do?

This PR renames the disk check struct to align with the existing naming conventions used across the codebase.

### Motivation

To ensure consistency across check implementations and adhere to internal naming standards.


jose-manuel-almaza (approved):
Question: When you say `existing naming conventions` or `adhere to internal naming standards`, are those conventions documented anywhere?

louis-cqrl:
> Question: When you say `existing naming conventions` or `adhere to internal naming standards`, are those conventions documented anywhere?

I don't think so to be honest, I think it may be something that came naturally for code clarity within the agent, but still I'm not sure

PR #36744, opened 2025-05-07
[DJM-814] Set intaller span tag to true only when len > 0

aboitreaud:
### What does this PR do?

Only set `host_tag_set.<key>` to `true` if the actual value of the host tag `key:value` was not empty, ie of length > 0.

### Motivation

Not consider the tag set if the env var was retrieved without any error but was empty

### DIFF
--- a/pkg/fleet/installer/setup/djm/databricks.go
+++ b/pkg/fleet/installer/setup/djm/databricks.go
@@ -208,7 +208,11 @@ func setClearIfExists(s *common.Setup, envKey, tagKey string, normalize func(str
 
 func setHostTag(s *common.Setup, tagKey, value string) {
 	s.Config.DatadogYAML.Tags = append(s.Config.DatadogYAML.Tags, tagKey+":"+value)
-	s.Span.SetTag("host_tag_set."+tagKey, "true")
+	if len(value) > 0 {
+		s.Span.SetTag("host_tag_set."+tagKey, "true")
+	} else {
+		s.Span.SetTag("host_tag_set."+tagKey, "false")
+	}
 }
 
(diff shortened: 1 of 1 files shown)


pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -208,7 +208,11 @@ func setClearIfExists(s *common.Setup, envKey, tagKey string, normalize func(str
 
 func setHostTag(s *common.Setup, tagKey, value string) {
 	s.Config.DatadogYAML.Tags = append(s.Config.DatadogYAML.Tags, tagKey+":"+value)
```
paul-laffon-dd, line 210: do we still want to set the tag if the value is not present ? As it would add "key:" in the tags array

mobuchowski (approved):

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -208,7 +208,11 @@ func setClearIfExists(s *common.Setup, envKey, tagKey string, normalize func(str
...
-	s.Span.SetTag("host_tag_set."+tagKey, "true")
+	if len(value) > 0 {
+		s.Span.SetTag("host_tag_set."+tagKey, "true")
+	} else {
+		s.Span.SetTag("host_tag_set."+tagKey, "false")
+	}
```
paul-laffon-dd, line 215: nit: writing it this way make the intent a bit more explicit, that only the `isTagPresent` is changed in the if/else

```suggestion
    isTagPresent := "false"
    if value != "" {
        isTagPresent = "true"
    }
    s.Span.SetTag("host_tag_set."+tagKey, isTagPresent)
```

PR #36745, opened 2025-05-07
[EBPF] docker: use WithX options pattern in testutil

gjulianm:
### What does this PR do?

This PR changes the way we pass options to the `NewRunConfig` and `NewComposeConfig` functions of the docker testutils, and adds two new arguments, `WithNetworkMode` and `WithPIDMode` 

### Motivation

Allow us to add more options to the config without having a constructor with a lot of arguments that need to be refactored each time.

Unit tests should still pass.


guyarb (approved):
🚀

pkg/collector/corechecks/servicediscovery/module/impl_linux_test.go
```diff
@@ -921,12 +921,16 @@ func TestDocker(t *testing.T) {
...
-		dockerutils.DefaultRetries,
-		scanner,
-		dockerutils.EmptyEnv,
-		filepath.Join(dir, "testdata", "docker-compose.yml"))
+
+	dockerCfg := dockerutils.NewComposeConfig(
+		dockerutils.WithBaseConfigForCompose(
+			dockerutils.WithName("foo-server"),
+			dockerutils.WithTimeout(dockerutils.DefaultTimeout),
+			dockerutils.WithRetries(dockerutils.DefaultRetries),
```
vitkyrka, line 929: Looks like these two are default in the base config so they could be removed from here

pkg/collector/corechecks/servicediscovery/module/impl_linux_test.go
```diff
@@ -921,12 +921,16 @@ func TestDocker(t *testing.T) {
...
+			dockerutils.WithPatternScanner(scanner),
+			dockerutils.WithEnv(dockerutils.EmptyEnv),
```
vitkyrka, line 931: This could also be default in the base config

vitkyrka (approved):

pkg/util/testutil/docker/config.go
```diff
@@ -154,31 +164,138 @@ func (c composeConfig) commandArgs(t subCommandType) []string {
...
+func NewRunConfig(opts ...RunConfigOption) LifecycleConfig {
+	cfg := &runConfig{}
+	for _, opt := range opts {
+		opt(cfg)
+	}
+	return *cfg
+}
+
+// WithFile sets the path to the docker-compose file.
+func WithFile(file string) ComposeConfigOption {
```
brycekahle, line 219: It looks like `File` is required, so it shouldn't be an optional argument.

gjulianm: Is there a common pattern to use in those cases? I left it like that to avoid having arguments + optionals, specially as it's test code and it will fail quickly if the file is not provided, but maybe there's a better way to have that.

brycekahle: Anything required should be an argument, that way you cannot mis-use it.

guyarb (approved):

pkg/util/testutil/docker/config.go
gjulianm: Changed to use that distinction, let me know if it looks better now.

brycekahle (approved):

PR #36749, opened 2025-05-07
[CWS] Adding a resolver to add file metadata. For now, only on execs

spikat:
### What does this PR do?

It adds a file resolver, enriching files with metadata as:
* size: for the file size
* type: empty, fileless, shell_script, text, compressed, encrypted, binary, linux_executable, windows_executable or macos_executable
* is_executable: true if it is
* architecture: x86, x86_64, arm or arm64
* abi: 32_bit or 64_bit
* compression: none, gzip, zip, zstd, 7z, bzip2 or xz
* is_upx_packets: true if it is
* is_garble_obfuscated: true if it is (produce false positives)

Adding those adds a cost as we need to read the file header (16 bytes for regular ones, or 2MB for executables).
That's why ATM those metadata are ONLY available on execs (as when executed, files are read and should be available in linux cache with less cost).

Related fields can be use on exec secl rules (ie: exec.is_upx_packed == true).

For now linkage metadata is disabled as it comes at a cost of more reads.

Feature disabled by default, can be turn on with `runtime_security_config.file_metadata_resolver.enabled`

### Motivation

Be able to detect ABI, UPX packed and garble obfuscated binaries (needed by PDEs)


udgover (commented):
That's a very cool PR and I'm looking forward to using these capabilities soon! 

I'm adding a general comment about the long term vision and not the review of the code itself. :) 

While I understand we want to keep this implementation as minimal, effective and performant as possible, I can't help but think that every time we want to support a new type, we will need to implement and release a new agent version. For now, it's limited to exec calls (which is already a very good use-case) but what if we would like to support metadata on other specific syscalls in the future?

I'm also thinking about more complex secl rules that could apply file identification on an open syscall if we previously detected a series of specific events ala chained rules + variable (if one variable is true from a previous rule, then check that type of opened file is xxx)

Regarding the implementation, what do you think about trying libmagic bindings like this [one](https://github.com/Velocidex/go-magic) from Velociraptor? As they describe, there are number of existing bindings for libmagic but their implementation aims to be self contained, without requiring libmagic to be installed. Besides, we can also compile our own lightweight version of libmagic compiled patterns. I did this in another open source project to only contain the most relevant patterns so as to removing the very nested logic as well and it was super effective.

We could also think about having a simpler format such as [scalpel carver configuration file](https://github.com/sleuthkit/scalpel/blob/master/scalpel.conf) where only headers are defined so as to the minimum number of bytes to read.

Still related to pattern matching, we also discussed about potentially relying on Yara-X (Rust implementation). Is this something we would like to PoC by limiting to execs as well?

Other than the previous ideas, there's also some other interesting implementations to keep in mind but that won't be the most effective ones for sure :)

* [Google magika](https://github.com/google/magika) is used at scale to help improve Google users' safety by routing Gmail, Drive, and Safe Browsing files to the proper security and content policy scanners, processing hundreds billions samples on a weekly basis.
* [Detect It Easy](https://github.com/horsicq/Detect-It-Easy) and their [key features](https://github.com/horsicq/Detect-It-Easy?tab=readme-ov-file#-key-features)! 
* [TrID](https://mark0.net/soft-trid-e.html) is an utility designed to identify file types from their binary signatures. While there are similar utilities with hard coded logic, TrID has no fixed rules. Instead, it's extensible and can be trained to recognize new formats in a fast and automatic way.

Thanks again for this really cool feature and keep up the good work! :)

docs/cloud-workload-security/linux_expressions.md
```diff
@@ -612,6 +612,14 @@ A process was executed (does not trigger on fork syscalls).
 | [`exec.file.hashes`](#common-fileevent-hashes-doc) | [Experimental] List of cryptographic hashes computed for this file |
 | [`exec.file.in_upper_layer`](#common-filefields-in_upper_layer-doc) | Indicator of the file layer, for example, in an OverlayFS |
 | [`exec.file.inode`](#common-pathkey-inode-doc) | Inode of the file |
+| [`exec.file.metadatas.abi`](#exec-file-metadatas-abi-doc) | [Experimental] ABI of the file (only for executable ones) |
```
cswatt, line 615:
```suggestion
| [`exec.file.metadatas.abi`](#exec-file-metadatas-abi-doc) | [Experimental] ABI of the file (only for executable files) |
```

"ones" has an ambiguous meaning

docs/cloud-workload-security/linux_expressions.md
```diff
@@ -612,6 +612,14 @@ A process was executed (does not trigger on fork syscalls).
...
+| [`exec.file.metadatas.architecture`](#exec-file-metadatas-architecture-doc) | [Experimental] Architecture of the file (only for executable ones) |
```
cswatt, line 616:
```suggestion
| [`exec.file.metadatas.architecture`](#exec-file-metadatas-architecture-doc) | [Experimental] Architecture of the file (only for executable files) |
```

docs/cloud-workload-security/linux_expressions.md
```diff
@@ -2952,6 +2960,74 @@ Definition: Timestamp of the event
 
 
 
+### `exec.file.metadatas.abi` {#exec-file-metadatas-abi-doc}
+Type: int
+
+Definition: [Experimental] ABI of the file (only for executable ones)
```
cswatt, line 2966:
```suggestion
Definition: [Experimental] ABI of the file (only for executable files)
```

docs/cloud-workload-security/linux_expressions.md
```diff
@@ -2952,6 +2960,74 @@ Definition: Timestamp of the event
...
+
+
+Constants: [ABI](#abi)
+
+
+
+### `exec.file.metadatas.architecture` {#exec-file-metadatas-architecture-doc}
+Type: int
+
+Definition: [Experimental] Architecture of the file (only for executable ones)
```
cswatt, line 2976:
```suggestion
Definition: [Experimental] Architecture of the file (only for executable files)
```

docs/cloud-workload-security/secl_linux.json
```diff
@@ -2187,6 +2187,46 @@
           "definition": "Inode of the file",
           "property_doc_link": "common-pathkey-inode-doc"
         },
+        {
+          "name": "exec.file.metadatas.abi",
+          "definition": "[Experimental] ABI of the file (only for executable ones)",
```
cswatt, line 2192:
```suggestion
          "definition": "[Experimental] ABI of the file (only for executable files)",
```

docs/cloud-workload-security/secl_linux.json
```diff
@@ -2187,6 +2187,46 @@
...
+          "property_doc_link": "exec-file-metadatas-abi-doc"
+        },
+        {
+          "name": "exec.file.metadatas.architecture",
+          "definition": "[Experimental] Architecture of the file (only for executable ones)",
```
cswatt, line 2197:
```suggestion
          "definition": "[Experimental] Architecture of the file (only for executable files)",
```

docs/cloud-workload-security/secl_linux.json
```diff
@@ -10347,6 +10387,102 @@
       "constants_link": "",
       "examples": []
     },
+    {
+      "name": "exec.file.metadatas.abi",
+      "link": "exec-file-metadatas-abi-doc",
+      "type": "int",
+      "definition": "[Experimental] ABI of the file (only for executable ones)",
```
cswatt, line 10394:
```suggestion
      "definition": "[Experimental] ABI of the file (only for executable files)",
```

docs/cloud-workload-security/secl_linux.json
```diff
@@ -10347,6 +10387,102 @@
...
+      ],
+      "constants": "ABI",
+      "constants_link": "abi",
+      "examples": []
+    },
+    {
+      "name": "exec.file.metadatas.architecture",
+      "link": "exec-file-metadatas-architecture-doc",
+      "type": "int",
+      "definition": "[Experimental] Architecture of the file (only for executable ones)",
```
cswatt, line 10406:
```suggestion
      "definition": "[Experimental] Architecture of the file (only for executable files)",
```

cswatt (changes requested):

docs/cloud-workload-security/linux_expressions.md
spikat: Fixed on last commit, thank you

cswatt (approved):

paulcacheux (approved):

docs/cloud-workload-security/backend_linux.schema.json
```diff
@@ -525,6 +525,9 @@
         "mount_origin": {
           "type": "string",
           "description": "MountOrigin origin of the mount"
+        },
+        "metadatas": {
```
lebauce: The plural form of metadata is also metadata

pkg/security/serializers/serializers_linux.go
```diff
@@ -703,10 +719,10 @@ func getInUpperLayer(f *model.FileFields) *bool {
 	return &upperLayer
 }
 
-func newFileSerializer(fe *model.FileEvent, e *model.Event, forceInode ...uint64) *FileSerializer {
+func newFileSerializerFull(fe *model.FileEvent, e *model.Event, forceInode uint64, metadatas *model.FileMetadatas) *FileSerializer {
```
safchain: nit: I would keep only the `newFileSerializer`. Full doesn't make really sense

pkg/security/resolvers/file/resolver_linux.go
```diff
@@ -0,0 +1,154 @@
...
+		}, nil
+	}
+
+	// add pid one for hash resolution outside of a container
+	process := event.ProcessContext.Process
+	rootPIDs := []uint32{process.Pid, 1}
+	if event.ProcessCacheEntry != nil {
+		rootPIDs = event.ProcessCacheEntry.GetAncestorsPIDs()
+	}
+	if process.ContainerID != "" && r.cgroupResolver != nil {
```
safchain, line 104: I would check this before the `GetAncestorsPIDs` version. `event.ProcessCacheEntry` will be almost 100% not empty, it will loop over all the ancestors containerized or not and the result is going to be replaced just after in case of cgroup based.

pkg/security/resolvers/file/resolver_linux.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+	}
+
+	for _, pid := range rootPIDs {
+		//  get proc path
+		path := utils.ProcRootFilePath(pid, file.PathnameStr)
+
+		// get file stats
+		fileInfo, err := os.Stat(path)
+		if err != nil {
```
safchain, line 119: We should handle more precisely the error here. A `permissions denied` should stop the iteration for all the pids while a `file not found` should probably continue. Also should we put something in an LRU to not try again and again the same path in case of error ?

pkg/security/resolvers/file/resolver_linux.go
```diff
@@ -0,0 +1,154 @@
...
+		entry, ok := r.cache.Get(key)
+		if ok {
+			r.cacheHit.Inc()
+			return entry, nil
+		}
+
+		// if no result in cache, analyze the file an return result
+		info, err := AnalyzeFile(path, fileInfo, checkLinkage)
+		if err != nil {
+			continue
```
safchain, line 154: same here about keeping track of errors

pkg/security/resolvers/file/resolver_linux.go
```diff
@@ -0,0 +1,154 @@
...
+
+		// cache lookup
+		entry, ok := r.cache.Get(key)
+		if ok {
+			r.cacheHit.Inc()
+			return entry, nil
+		}
+
+		// if no result in cache, analyze the file an return result
+		info, err := AnalyzeFile(path, fileInfo, checkLinkage)
```
safchain, line 149: You need to check that this is a regular file. Named pipe for instance will block

pkg/security/serializers/serializers_linux.go
```diff
@@ -1541,8 +1579,11 @@ func NewEventSerializer(event *model.Event, rule *rules.Rule) *EventSerializer {
 		s.EventContextSerializer.Outcome = serializeOutcome(0)
 		s.IMDSEventSerializer = newIMDSEventSerializer(&event.IMDS)
 	case model.ExecEventType:
+		// only calling the first one, it will resolve all metadata fields
+		event.FieldHandlers.ResolveFileMetadatasSize(event, &event.Exec.FileMetadatas)
```
safchain: I would rather use the : `ResolveFileMetadatas`

pkg/security/resolvers/file/resolver_linux.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+	if file.PathResolutionError != nil {
+		return nil, errors.New("path resolution error")
+	}
+
+	// fileless
+	if file.BasenameStr == "memfd:" && file.PathnameStr == "" {
+		return &model.FileMetadatas{
+			Type:         int(model.FileLess),
+			IsExecutable: true,
```
YoannGh, line 97: I think there can be non-executable memfd files, would do you think about setting the IsExecutable flag if `event.Type() == ExecEventType`?

pkg/security/resolvers/file/resolver_linux.go
```diff
@@ -0,0 +1,154 @@
...
+	for _, pid := range rootPIDs {
+		//  get proc path
+		path := utils.ProcRootFilePath(pid, file.PathnameStr)
+
+		// get file stats
+		fileInfo, err := os.Stat(path)
+		if err != nil {
+			continue
+		}
+
```
YoannGh, line 133: Can we skip the analysis here if `fileInfo.Ino != file.Ino` in order to make sure the file didn't change between the time we generated the event and the time we do the analysis?

pkg/security/serializers/serializers_linux.go
spikat: The main `ResolveFileMetadata` function is not accessible here.

pkg/security/resolvers/file/resolver_linux.go
spikat: The memfd file is kind of a special case where we are not analyzing any file. The `IsExecutable` field does little sense here, but I've put it to be consistent with other exec files. Would you prefer to get it removed completely?

Also, keep in mind that ATM this resolver is called only on execs but it would not be the case eternally. At some files that are not executed can have the execution flag set.

safchain (approved):

docs/cloud-workload-security/linux_expressions.md
```diff
@@ -3625,6 +3701,26 @@ Definition: Path argument of the syscall
 
 Constants are used to improve the readability of your rules. Some constants are common to all architectures, others are specific to some architectures.
 
+### `ABI` {#abi}
+ABI used for binary compilation.
+
+| Name | Architectures |
+| ---- |---------------|
+| `BIT32` | all |
```
lebauce, line 3709: Isn't it redundant with `Architecture` ?

spikat: I don't think, as we could have binaries using the 32bit ABI of 64bit architectures

spikat: But I didn't test this specific use case yet. But IMHO it could make sense to differentiate them

pkg/security/resolvers/file/resolver_linux.go
YoannGh: Indeed the resolver is only called for exec events for now so + I don't think we handle memfd files with other event types so it's ok keeping it like this

YoannGh (approved):

hush-hush (approved):

PR #36750, opened 2025-05-07
[EBPF] ci: fix complexity summary job

gjulianm:
### What does this PR do?

This PR fixes the filter for `main` pipelines, accepting pipelines generated by `api` too. It also adds commands to tag the jobs, to enable better monitoring of the behavior.

### Motivation

With DDCI activation, there are no longer any pipelines generated by `push`. 

Manually checked that the job now finds the complexity data from the `main` pipeline. We will add monitors based on the new tags added to the job via `datadog-ci`.


mbertrone (approved):
The change LGTM!

have you considered adding tests for the 2 new methods introduced?

gjulianm:
Unfortunately I don't think we have unit test infrastructure for the Python tasks code

PR #36757, opened 2025-05-07
Remove python / go inactive versions for macos runners

CelianR:
### What does this PR do?

> [!NOTE]
> This will be merged some time after #36693 to have some metrics and not wipe everything.

This is a follow up of #36693, it removes inactive versions each 20 job run (lint / test / build) for each runner.


.gitlab/common/macos.yml
```diff
@@ -6,6 +6,13 @@
   - DD_API_KEY="$("$CI_PROJECT_DIR"/tools/ci/fetch_secret.sh "$AGENT_API_KEY_ORG2" token)" || exit $?; export DD_API_KEY
   - dda inv -- -e macos.report-versions -l python || true
   - dda inv -- -e macos.report-versions -l go || true
+  # Remove inactive versions
```
chouetz, line 9: Ok now I understand the naming, sorry for the comment on the previous PR (even if I would probably set `macos_maintenance` or something similar instead of `clean_env` as this is not only cleaning and the cleaning is not systematic)

.gitlab/common/macos.yml
```diff
@@ -6,6 +6,13 @@
...
+  - |
+    if [ "$((RANDOM%20))" -eq 0 ]; then
+      echo Trying to remove inactive versions
+      dda inv -- -e macos.remove-inactive-versions -l python -t "$PYTHON_VERSION" --no-dry-run
```
chouetz, line 13:
```suggestion
      dda inv -- -e macos.remove-inactive-versions -l python -t "$PYTHON_VERSION"
```
I suggest to reverse the option (so set `--dry-run` when you need it only)

tasks/macos.py
```diff
@@ -27,3 +29,107 @@ def report_versions(ctx, lang):
...
+            for version in sorted(versions):
+                print(version)
+        else:
+            print(f'No ci active {lang} versions found')
+
+    return versions
+
+
+@task
+def list_runner_active_versions(ctx, lang):
```
chouetz, line 70: Do we need to make a task of this one (as we cannot launch it locally I suppose)?

tasks/macos.py
```diff
@@ -27,3 +29,107 @@ def report_versions(ctx, lang):
...
+
+    # Transform target version from go<version> to <version>
+    if lang == "go":
+        target_version = target_version.removeprefix("go")
+
+    # Transform target version from X.YY.ZZ to X.YY
+    if target_version.count('.') == 2:
+        target_version = '.'.join(target_version.split('.')[:-1])
+
+    # Remove all versions that are not in the CI (except the target version)
```
chouetz, line 120: Can we use `set` operations to get the versions we want to get rid of?

chouetz (approved):
Minor comments

.gitlab/common/macos.yml
CelianR: Let's use `macos_runner_maintenance` 👍

tasks/macos.py
CelianR: We can use it locally to test on our laptop if we use pyenv or gimme. I think it's better to leave this as is if we want to update this later

PR #36760, opened 2025-05-07
CXP-2086 [agent] add permissions check to windows process agent

gengnamstyle:
### What does this PR do?
- introduces a new function `IsProcessProtected` via a system call to determine if the agent has access to reading the memory or not
- refactor command line processing to include executable/image path even if command line processsing fails (existing behaviour, but was impacted by this protection check)
    - align `process_windows` `fillProcessDetails` function to be more similar with `process_windows_toolhelp` `fillFromProcEntry` because they serve the same purpose

### Motivation
- previously log show that we attempt to call `ReadProcessMemory` even when we fail to open the process with `PROCESS_VM_READ`

```
2024-08-21 17:02:31 EDT | PROCESS | DEBUG | (pkg/process/procutil/process_windows.go:545 in OpenProcessHandle) | Couldn't open process with PROCESS_VM_READ 4 Access is denied.
2024-08-21 17:02:31 EDT | PROCESS | DEBUG | (pkg/util/winutil/process.go:101 in ReadProcessMemory) | Access denied error getting process memory
2024-08-21 17:02:31 EDT | PROCESS | DEBUG | (pkg/process/procutil/process_windows_toolhelp.go:247 in fillFromProcEntry) | Error retrieving full command line Access is denied.
```
We shouldn't bother trying to collect the command line if we don't have PROCESS_VM_READ  on the handle.

- added unit tests for the new `IsProcessProtected` function
- updated windows e2e tests to actually check if command line arguments exist for an unprotected process
- benchmarked new `OpenProcessHandle` function to verify performance impact (not significant)
    - thread in pr
> I also just ran some benchmarking just testing the openProcessHandle with 1 process at a time with the following results:
BenchmarkOpenProcessHandleProtected-4             255981              6009 ns/op
BenchmarkOldOpenProcessHandleProtected-4           15889             77775 ns/op
BenchmarkOpenProcessHandleUnprotected-4           197953              6709 ns/op
BenchmarkOldOpenProcessHandleUnprotected-4        358704              3219 ns/op
A single unprotected process takes 6 microseconds vs 3 compared to the old method.
What's interesting is for protected processes, the older method takes ~10x the time likely from handling an error since the older version attempts to get more privileged access initially and on an error would fall back to a less privileged version.
While with the newer version, we check for access or not before attempting.
Here is the testing file as a gist that I used if you'd like to check.
https://gist.github.com/gengnamstyle/8348a4edc51807922fbc90914e63e5aa

- additionally system check in `OpenProcess` handle and opening 2 process handles instead of 1 in the main codepath for an unprotected process leading to 3 microseconds of increased overhead per process so a host having 100 processes may have 0.3 millisecond impact (not substantial)

### Additional Notes

- created a ticket to consolidate testing helper functions in e2e tests since they is a lot of duplicate functionality and refactors to do! (didn't want to increase the scope of this pr)
   - most notably, there is a `assertProcessCollectedNew` and a `assertProcessCollected` and `assertProcessCollectedNew` should be replacing `assertProcessCollected` but this was never done
   - https://datadoghq.atlassian.net/browse/CXP-2229

### DIFF
--- a/pkg/process/procutil/process_windows.go
+++ b/pkg/process/procutil/process_windows.go
@@ -504,7 +504,6 @@ func fillProcessDetails(pid int32, proc *Process) error {
 	proc.Exe = imagePath
 
 	// we cannot read the command line if the process is protected
-	var cmdLine string
 	if !isProtected {
 		processCmdParams, err := winutil.GetCommandParamsForProcess(procHandle, false)
 		if err != nil {
@@ -512,13 +511,11 @@ func fillProcessDetails(pid int32, proc *Process) error {
 		}
 
 		if processCmdParams != nil {
-			cmdLine = processCmdParams.CmdLine
+			proc.Cmdline = ParseCmdLineArgs(processCmdParams.CmdLine)
+			if len(processCmdParams.CmdLine) > 0 && len(proc.Cmdline) == 0 {
+				log.Warnf("Failed to parse the cmdline:%s for pid:%d", processCmdParams.CmdLine, pid)
+			}
 		}
-	}
-
-	proc.Cmdline = ParseCmdLineArgs(cmdLine)
-	if len(cmdLine) > 0 && len(proc.Cmdline) == 0 {
-		log.Warnf("Failed to parse the cmdline:%s for pid:%d", cmdLine, pid)
 	}
 
 	var CPU windows.Rusage
@@ -550,10 +547,6 @@ func OpenProcessHandle(pid int32) (windows.Handle, bool, error) {
 		return procHandle, isProtected, err
 	}
 
-	// 0x1000 is PROCESS_QUERY_LIMITED_INFORMATION, but that constant isn't
-	//        defined in x/sys/windows
-	// 0x10   is PROCESS_VM_READ
-	// if the process is not protected, we should be able to access with PROCESS_VM_READ
 	procHandleMemoryAccess, err := windows.OpenProcess(windows.PROCESS_QUERY_LIMITED_INFORMATION|windows.PROCESS_VM_READ, false, uint32(pid))
 	if err != nil {
 		log.Debugf("Couldn't open unprotected process with PROCESS_VM_READ access. Returning limited process info %v %v", pid, err)
--- a/pkg/process/procutil/process_windows.go
+++ b/pkg/process/procutil/process_windows.go
@@ -552,6 +552,7 @@ func OpenProcessHandle(pid int32) (windows.Handle, bool, error) {
 		log.Debugf("Couldn't open unprotected process with PROCESS_VM_READ access. Returning limited process info %v %v", pid, err)
 		return procHandle, isProtected, err
 	}
+	defer windows.Close(procHandle)
 	return procHandleMemoryAccess, isProtected, err
 }
 
--- a/pkg/process/procutil/process_windows.go
+++ b/pkg/process/procutil/process_windows.go
@@ -485,7 +485,7 @@ func getPIDs() ([]int32, error) {
 }
 
 func fillProcessDetails(pid int32, proc *Process) error {
-	procHandle, err := OpenProcessHandle(pid)
+	procHandle, isProtected, err := OpenProcessHandle(pid)
 	if err != nil {
 		return err
 	}
@@ -497,7 +497,24 @@ func fillProcessDetails(pid int32, proc *Process) error {
 	}
 	proc.Username = userName
 
-	cmdParams := getProcessCommandParams(procHandle)
+	var cmdParams winutil.ProcessCommandParams
+	// we cannot read the command line if the process is protected
+	if !isProtected {
+		processCmdParams, err := winutil.GetCommandParamsForProcess(procHandle, false)
+		if err != nil {
+			log.Debugf("Error retrieving full command line %v", err)
+		}
+
+		if processCmdParams != nil {
+			cmdParams = *processCmdParams
+		}
+	} else {
+		imagePath, err := winutil.GetImagePathForProcess(procHandle)
+		if err != nil {
+			log.Debugf("Error retrieving exe path %v", err)
+		}
+		cmdParams.ImagePath = imagePath
+	}
 
 	proc.Cmdline = ParseCmdLineArgs(cmdParams.CmdLine)
 	if len(cmdParams.CmdLine) > 0 && len(proc.Cmdline) == 0 {
@@ -517,39 +534,31 @@ func fillProcessDetails(pid int32, proc *Process) error {
 	return nil
 }
 
-func getProcessCommandParams(procHandle windows.Handle) *winutil.ProcessCommandParams {
-	var err error
-	if cmdParams, err := winutil.GetCommandParamsForProcess(procHandle, true); err == nil {
-		return cmdParams
-	}
-
-	log.Debugf("Error retrieving command params %v", err)
-	if imagePath, err := winutil.GetImagePathForProcess(procHandle); err == nil {
-		return &winutil.ProcessCommandParams{
-			CmdLine:   imagePath,
-			ImagePath: imagePath,
-		}
-	}
-
-	log.Debugf("Error retrieving exe path %v", err)
-	return &winutil.ProcessCommandParams{}
-}
-
 // OpenProcessHandle attempts to open process handle for reading process memory with fallback to query basic info
-func OpenProcessHandle(pid int32) (windows.Handle, error) {
+func OpenProcessHandle(pid int32) (windows.Handle, bool, error) {
+	procHandle, err := windows.OpenProcess(windows.PROCESS_QUERY_LIMITED_INFORMATION, false, uint32(pid))
+	if err != nil {
+		log.Debugf("Couldn't open process %v %v", pid, err)
+		return windows.Handle(0), false, err
+	}
+
+	// check protection level to determine memory readability for PROCESS_VM_READ
+	// more info: https://learn.microsoft.com/en-us/windows/win32/procthread/process-security-and-access-rights#protected-processes
+	isProtected, err := winutil.IsProcessProtected(procHandle)
+	if isProtected {
+		return procHandle, isProtected, err
+	}
+
 	// 0x1000 is PROCESS_QUERY_LIMITED_INFORMATION, but that constant isn't
 	//        defined in x/sys/windows
 	// 0x10   is PROCESS_VM_READ
-	procHandle, err := windows.OpenProcess(0x1010, false, uint32(pid))
+	// if the process is not protected, we should be able to access with PROCESS_VM_READ
+	procHandleMemoryAccess, err := windows.OpenProcess(windows.PROCESS_QUERY_LIMITED_INFORMATION|windows.PROCESS_VM_READ, false, uint32(pid))
 	if err != nil {
-		log.Debugf("Couldn't open process with PROCESS_VM_READ %v %v", pid, err)
-		procHandle, err = windows.OpenProcess(0x1000, false, uint32(pid))
-		if err != nil {
-			log.Debugf("Couldn't open process %v %v", pid, err)
-			return windows.Handle(0), err
-		}
+		log.Debugf("Couldn't open unprotected process with PROCESS_VM_READ access. Returning limited process info %v %v", pid, err)
+		return procHandle, isProtected, err
 	}
-	return procHandle, nil
+	return procHandleMemoryAccess, isProtected, err
 }
 
 // GetUsernameForProcess returns username for a process
--- a/pkg/process/procutil/process_windows.go
+++ b/pkg/process/procutil/process_windows.go
@@ -529,7 +529,15 @@ func fillProcessDetails(pid int32, proc *Process) error {
 	return nil
 }
 
-// OpenProcessHandle attempts to open process handle for reading process memory with fallback to query basic info
+// OpenProcessHandle attempts to open a process handle with the highest available privilege of memory access
+// Currently, there are performance implications of opening the process handle twice for an unprotected process
+// 1. open with lower privilege
+// 2. check protection level
+// 3. open with higher privilege if not protected
+// Most processes will likely be unprotected and this operation takes ~5-10 microseconds which could be lowered to
+// ~3 microseconds if we open the process with the right access initially.
+// However, the current method was chosen due to security implications from a customer issue where their anti-virus
+// flagged the agent for attempting to open a process with the incorrect permissions
 func OpenProcessHandle(pid int32) (windows.Handle, bool, error) {
 	procHandle, err := windows.OpenProcess(windows.PROCESS_QUERY_LIMITED_INFORMATION, false, uint32(pid))
 	if err != nil {
--- a/pkg/process/procutil/process_windows.go
+++ b/pkg/process/procutil/process_windows.go
@@ -20,22 +20,20 @@ import (
 	"github.com/DataDog/datadog-agent/pkg/util/winutil"
 )
 
-var (
-	counterPaths = []string{
-		pdhutil.CounterAllProcessPID,
-		pdhutil.CounterAllProcessParentPID,
-		pdhutil.CounterAllProcessPctUserTime,
-		pdhutil.CounterAllProcessPctPrivilegedTime,
-		pdhutil.CounterAllProcessWorkingSet,
-		pdhutil.CounterAllProcessPoolPagedBytes,
-		pdhutil.CounterAllProcessThreadCount,
-		pdhutil.CounterAllProcessHandleCount,
-		pdhutil.CounterAllProcessIOReadOpsPerSec,
-		pdhutil.CounterAllProcessIOWriteOpsPerSec,
-		pdhutil.CounterAllProcessIOReadBytesPerSec,
-		pdhutil.CounterAllProcessIOWriteBytesPerSec,
-	}
-)
+var counterPaths = []string{
+	pdhutil.CounterAllProcessPID,
+	pdhutil.CounterAllProcessParentPID,
+	pdhutil.CounterAllProcessPctUserTime,
+	pdhutil.CounterAllProcessPctPrivilegedTime,
+	pdhutil.CounterAllProcessWorkingSet,
+	pdhutil.CounterAllProcessPoolPagedBytes,
+	pdhutil.CounterAllProcessThreadCount,
+	pdhutil.CounterAllProcessHandleCount,
+	pdhutil.CounterAllProcessIOReadOpsPerSec,
+	pdhutil.CounterAllProcessIOWriteOpsPerSec,
+	pdhutil.CounterAllProcessIOReadBytesPerSec,
+	pdhutil.CounterAllProcessIOWriteBytesPerSec,
+}
 
 // NewProcessProbe returns a Probe object
 func NewProcessProbe(...Option) Probe {
@@ -559,7 +557,7 @@ func OpenProcessHandle(pid int32) (windows.Handle, bool, error) {
 		log.Debugf("Couldn't open unprotected process with PROCESS_VM_READ access. Returning limited process info %v %v", pid, err)
 		return procHandle, isProtected, err
 	}
-	defer windows.Close(procHandle)
+	windows.Close(procHandle)
 	return procHandleMemoryAccess, isProtected, err
 }
 
--- a/pkg/process/procutil/process_windows.go
+++ b/pkg/process/procutil/process_windows.go
@@ -222,7 +222,6 @@ func (p *probe) ProcessesByPID(_ time.Time, collectStats bool) (map[int32]*Proce
 		}
 
 		err := fillProcessDetails(pid, proc)
-
 		if err != nil {
 			continue
 		}
@@ -266,7 +265,6 @@ func (p *probe) enumCounters(collectMeta bool, collectStats bool) error {
 	}
 
 	err = p.formatter.Enum(pdhutil.CounterAllProcessPID, p.counters[pdhutil.CounterAllProcessPID], pdhutil.PDH_FMT_LARGE, ignored, valueToUint64(p.mapPID))
-
 	if err != nil {
 		return err
 	}
@@ -505,13 +503,14 @@ func fillProcessDetails(pid int32, proc *Process) error {
 
 	// we cannot read the command line if the process is protected
 	if !isProtected {
-		processCmdParams, err := winutil.GetCommandParamsForProcess(procHandle, false)
+		processCmdParams, err := winutil.GetCommandParamsForProcess(procHandle, true)
 		if err != nil {
 			log.Debugf("Error retrieving full command line %v", err)
 		}
 
 		if processCmdParams != nil {
 			proc.Cmdline = ParseCmdLineArgs(processCmdParams.CmdLine)
+			proc.Exe = processCmdParams.ImagePath
 			if len(processCmdParams.CmdLine) > 0 && len(proc.Cmdline) == 0 {
 				log.Warnf("Failed to parse the cmdline:%s for pid:%d", processCmdParams.CmdLine, pid)
 			}
--- a/pkg/process/procutil/process_windows.go
+++ b/pkg/process/procutil/process_windows.go
@@ -549,7 +549,7 @@ func OpenProcessHandle(pid int32) (windows.Handle, bool, error) {
 	// more info: https://learn.microsoft.com/en-us/windows/win32/procthread/process-security-and-access-rights#protected-processes
 	isProtected, err := winutil.IsProcessProtected(procHandle)
 	if err != nil {
-		log.Debugf("Couldn't access process %v protection info %v", pid, err)
+		log.Debugf("Couldn't access process %v protection info %v. Will attempt to re-open with PROCESS_VM_READ access", pid, err)
 	}
 	if isProtected {
 		return procHandle, isProtected, err
--- a/pkg/process/procutil/process_windows.go
+++ b/pkg/process/procutil/process_windows.go
@@ -497,8 +497,14 @@ func fillProcessDetails(pid int32, proc *Process) error {
 	}
 	proc.Username = userName
 
-	var cmdParams winutil.ProcessCommandParams
+	imagePath, err := winutil.GetImagePathForProcess(procHandle)
+	if err != nil {
+		log.Debugf("Error retrieving exe path for pid %v %v", pid, err)
+	}
+	proc.Exe = imagePath
+
 	// we cannot read the command line if the process is protected
+	var cmdLine string
 	if !isProtected {
 		processCmdParams, err := winutil.GetCommandParamsForProcess(procHandle, false)
 		if err != nil {
@@ -506,22 +512,14 @@ func fillProcessDetails(pid int32, proc *Process) error {
 		}
 
 		if processCmdParams != nil {
-			cmdParams = *processCmdParams
+			cmdLine = processCmdParams.CmdLine
 		}
-	} else {
-		imagePath, err := winutil.GetImagePathForProcess(procHandle)
-		if err != nil {
-			log.Debugf("Error retrieving exe path %v", err)
-		}
-		cmdParams.ImagePath = imagePath
 	}
 
-	proc.Cmdline = ParseCmdLineArgs(cmdParams.CmdLine)
-	if len(cmdParams.CmdLine) > 0 && len(proc.Cmdline) == 0 {
-		log.Warnf("Failed to parse the cmdline:%s for pid:%d", cmdParams.CmdLine, pid)
+	proc.Cmdline = ParseCmdLineArgs(cmdLine)
+	if len(cmdLine) > 0 && len(proc.Cmdline) == 0 {
+		log.Warnf("Failed to parse the cmdline:%s for pid:%d", cmdLine, pid)
 	}
-
-	proc.Exe = cmdParams.ImagePath
 
 	var CPU windows.Rusage
 	if err := windows.GetProcessTimes(procHandle, &CPU.CreationTime, &CPU.ExitTime, &CPU.KernelTime, &CPU.UserTime); err != nil {
@@ -545,6 +543,9 @@ func OpenProcessHandle(pid int32) (windows.Handle, bool, error) {
 	// check protection level to determine memory readability for PROCESS_VM_READ
 	// more info: https://learn.microsoft.com/en-us/windows/win32/procthread/process-security-and-access-rights#protected-processes
 	isProtected, err := winutil.IsProcessProtected(procHandle)
+	if err != nil {
+		log.Debugf("Couldn't access process %v protection info %v", pid, err)
+	}
 	if isProtected {
 		return procHandle, isProtected, err
 	}
--- a/pkg/process/procutil/process_windows_test.go
+++ b/pkg/process/procutil/process_windows_test.go
@@ -106,10 +106,9 @@ func TestWindowsProbe(t *testing.T) {
 			defer cmd.Process.Kill()
 
 			procs, err := probe.ProcessesByPID(time.Now(), true)
-
 			assert.NoError(t, err)
-			p, found := procs[int32(cmd.Process.Pid)]
 
+			p, found := procs[int32(cmd.Process.Pid)]
 			assert.True(t, found)
 			assert.True(t, strings.HasSuffix(p.Exe, "powershell.exe"))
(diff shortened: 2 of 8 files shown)


buraizu (approved):
Approving with a minor suggestion for the release note

pkg/process/procutil/process_windows.go
```diff
@@ -497,14 +497,29 @@ func fillProcessDetails(pid int32, proc *Process) error {
...
-	proc.Cmdline = ParseCmdLineArgs(cmdParams.CmdLine)
-	if len(cmdParams.CmdLine) > 0 && len(proc.Cmdline) == 0 {
-		log.Warnf("Failed to parse the cmdline:%s for pid:%d", cmdParams.CmdLine, pid)
+		if processCmdParams != nil {
+			cmdLine = processCmdParams.CmdLine
+		}
 	}
 
-	proc.Exe = cmdParams.ImagePath
+	proc.Cmdline = ParseCmdLineArgs(cmdLine)
```
robertjli, line 519: Suggestion: since cmdLine is only ever populated if `!isProtected`, we can move this assignment and the following check into the if block above.

pkg/process/procutil/process_windows.go
```diff
@@ -517,39 +532,34 @@ func fillProcessDetails(pid int32, proc *Process) error {
...
 	}
 
-	log.Debugf("Error retrieving exe path %v", err)
-	return &winutil.ProcessCommandParams{}
-}
-
-// OpenProcessHandle attempts to open process handle for reading process memory with fallback to query basic info
-func OpenProcessHandle(pid int32) (windows.Handle, error) {
 	// 0x1000 is PROCESS_QUERY_LIMITED_INFORMATION, but that constant isn't
 	//        defined in x/sys/windows
```
robertjli, line 554: Question: is this comment still true? It looks like you are using the constant above.

pkg/process/procutil/process_windows.go
```diff
@@ -517,39 +532,34 @@ func fillProcessDetails(pid int32, proc *Process) error {
...
+	procHandleMemoryAccess, err := windows.OpenProcess(windows.PROCESS_QUERY_LIMITED_INFORMATION|windows.PROCESS_VM_READ, false, uint32(pid))
 	if err != nil {
-		log.Debugf("Couldn't open process with PROCESS_VM_READ %v %v", pid, err)
-		procHandle, err = windows.OpenProcess(0x1000, false, uint32(pid))
-		if err != nil {
-			log.Debugf("Couldn't open process %v %v", pid, err)
-			return windows.Handle(0), err
-		}
+		log.Debugf("Couldn't open unprotected process with PROCESS_VM_READ access. Returning limited process info %v %v", pid, err)
+		return procHandle, isProtected, err
```
robertjli, line 560: Question: Since `err != nil` here, `fillProcessDetails()` returns early above. Is that expected?

pkg/process/procutil/process_windows_test.go
robertjli, line 1: Chore: this file didn't materially change, please revert to keep the git history clean

pkg/process/procutil/process_windows_toolhelp.go
```diff
@@ -240,14 +241,20 @@ func (cp *cachedProcess) fillFromProcEntry(pe32 *w32.PROCESSENTRY32) (err error)
...
 	}
-	var cmderr error
 	cp.executablePath = winutil.ConvertWindowsString16(pe32.SzExeFile[:])
-	commandParams, cmderr := winutil.GetCommandParamsForProcess(cp.procHandle, false)
-	if cmderr != nil {
-		log.Debugf("Error retrieving full command line %v", cmderr)
-		cp.commandLine = cp.executablePath
-	} else {
-		cp.commandLine = commandParams.CmdLine
+	// set default commandLine value incase we can't read the actual commandLine due to the proteced process
```
robertjli, line 245:
```suggestion
	// set default commandLine value in case we can't read the actual commandLine due to the protected process
```

Suggestion: I don't think this comment is necessary, can remove

pkg/process/procutil/process_windows_toolhelp.go
```diff
@@ -240,14 +241,20 @@ func (cp *cachedProcess) fillFromProcEntry(pe32 *w32.PROCESSENTRY32) (err error)
...
+	cp.commandLine = cp.executablePath
+
+	// we cannot read the command line if the process is protected
+	if !isProtected {
+		var cmderr error
```
robertjli, line 250: This declaration is shadowed by the `:=` below. You can remove it

pkg/util/winutil/process.go
```diff
@@ -80,9 +84,52 @@ func NtQueryInformationProcess(h windows.Handle, class PROCESSINFOCLASS, target,
...
+			PsProtectedTypeNone = 0,
+			PsProtectedTypeProtectedLight = 1,
+			PsProtectedTypeProtected = 2
+		} PS_PROTECTED_TYPE, *PPS_PROTECTED_TYPE;
+
+		if the protection type is anything other than 0, then we cannot open the process with PROCESS_VM_READ
+	*/
+
+	// hex 0x07 == binary 111, so we do a binary AND to determine the protection level
+	if processProtection&0x07 != 0 {
```
robertjli, line 128:
```suggestion
	if processProtection & 0b111 != 0 {
```
Suggestion: This removes the need for the comment

robertjli (commented):
> manual validation should involve verifying that no regressions occur on the normal windows agent

Would it be possible to add an e2e test for this instead of doing manual regression testing?

pkg/util/winutil/process.go
```diff
@@ -80,9 +84,52 @@ func NtQueryInformationProcess(h windows.Handle, class PROCESSINFOCLASS, target,
...
+		target,
+		size,
+		uintptr(0))
+	if r != 0 {
+		err = windows.GetLastError()
+		return err
+	}
+	return nil
+}
+
```
iglendd, line 105: This unnecessary reimplements the same syscall ([NtQueryInformationProcess](https://github.com/DataDog/datadog-agent/blob/058445c5c367ca2afb8a3c074d2853f4e955d951/pkg/util/winutil/process.go#L75)) which easy to miss since either Zw or Nt prefix can be used in user mode

pkg/util/winutil/process.go
```diff
@@ -80,9 +84,52 @@ func NtQueryInformationProcess(h windows.Handle, class PROCESSINFOCLASS, target,
...
+		The first 3 bits contain the type of protected process:
+
+		typedef enum _PS_PROTECTED_TYPE {
+			PsProtectedTypeNone = 0,
+			PsProtectedTypeProtectedLight = 1,
+			PsProtectedTypeProtected = 2
+		} PS_PROTECTED_TYPE, *PPS_PROTECTED_TYPE;
+
+		if the protection type is anything other than 0, then we cannot open the process with PROCESS_VM_READ
+	*/
```
iglendd, line 125: I wonder if [GetProcessInformation](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getprocessinformation) is effectively equivalent and simple to use simply compare it to PROTECTION_LEVEL_NONE

pkg/util/winutil/process.go
```diff
@@ -21,6 +21,7 @@ var (
 	modntdll                       = windows.NewLazyDLL("ntdll.dll")
 	modkernel                      = windows.NewLazyDLL("kernel32.dll")
 	procNtQueryInformationProcess  = modntdll.NewProc("NtQueryInformationProcess")
+	procZwQueryInformationProcess  = modntdll.NewProc("ZwQueryInformationProcess")
```
iglendd, line 24: These two are the same (in user mode)

pkg/process/procutil/process_windows.go
```diff
@@ -517,39 +532,34 @@ func fillProcessDetails(pid int32, proc *Process) error {
...
 	}
-	return procHandle, nil
+	return procHandleMemoryAccess, isProtected, err
```
iglendd, line 562: Is procHandle leaks here?

pkg/process/procutil/process_windows.go
```diff
@@ -517,39 +532,34 @@ func fillProcessDetails(pid int32, proc *Process) error {
 	return nil
 }
 
-func getProcessCommandParams(procHandle windows.Handle) *winutil.ProcessCommandParams {
-	var err error
-	if cmdParams, err := winutil.GetCommandParamsForProcess(procHandle, true); err == nil {
-		return cmdParams
+// OpenProcessHandle attempts to open process handle for reading process memory with fallback to query basic info
+func OpenProcessHandle(pid int32) (windows.Handle, bool, error) {
```
iglendd, line 536: I wonder what is a statistics of protected vs non protected processes on a system. I feel that number of protected process is much smaller than otherwise. Accordingly we are significantly increasing overhead by opening processes twice, which is one of the most expensive operation (compare to other). Why not to go other way around. Attempt to open process normally and if it fails open in limited mode. Technically we do not even need to test if it is protected if in limited it opens successfully. The only thing is to change when to log details what level we have opened.

pkg/util/winutil/process.go
clarkb7: explained on MSDN here
https://learn.microsoft.com/en-us/windows-hardware/drivers/kernel/using-nt-and-zw-versions-of-the-native-system-services-routines

pkg/util/winutil/process.go
clarkb7: nice find, i think it is.
the [ZwQueryInformationProcess](https://learn.microsoft.com/en-us/windows/win32/procthread/zwqueryinformationprocess) docs say "it is better to use public functions mentioned in the description of the ProcessInformationClass parameter instead." but they don't list an alternative API for `ProcessProtectionInformation` 🫠

pkg/process/procutil/process_windows.go
clarkb7: there's a tradeoff with avoiding antimalware detections
> This rule prevents untrusted processes from having direct access to LSASS memory. Whenever a process tries to use the OpenProcess() function to access LSASS, with an access right of PROCESS_VM_READ, the rule specifically blocks that access right.

https://learn.microsoft.com/en-us/defender-endpoint/attack-surface-reduction-faq

iglendd: Ah, that makes a difference. Would be interesting to know relative and absolute overhead of the change for 100 processes.

pkg/util/winutil/process.go
gengnamstyle: I attempted to migrate to using the GetProcessInformation function, but it doesn't seem to return correct values that are actually protected. I've been unit testing using the following process names that are protected

- `csrss`
- `services`
- `smss`
- `wininit`

And all these return 0 which does not seem to be the suggested behaviour. Testing the process I create (unprotected) also returns a non-zero value. Perhaps, the logic can be flipped, but that doesn't seem ideal to me. 

The return value from this function https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/ns-processthreadsapi-process_protection_level_information

doesn't officially indicate how the result is given and I feel iffy about experimenting to the find the behaviour. Perhaps this is why this function is not recommended as an "alternative api". 

I believe I should stick with using NtQuery for now. 🤔

pkg/process/procutil/process_windows.go
gengnamstyle: Yeah I agree. I didn't like the extra overhead of doing extra system calls for a case that doesn't seem to be the most common case.

pkg/util/winutil/process.go
gengnamstyle: using nt as suggested!

pkg/util/winutil/process.go
gengnamstyle: using nt! thanks for the reference

pkg/util/winutil/process.go
gengnamstyle: cleaner thanks

pkg/process/procutil/process_windows_toolhelp.go
gengnamstyle: removed

pkg/process/procutil/process_windows_test.go
gengnamstyle: updated thanks

pkg/process/procutil/process_windows.go
gengnamstyle: Yes that is expected. If `err != nil`, we can't open a process handle with memory access despite having the right protection level so we just return the basic query information instead.

A side effect of this is that if we do get an error and return the basic process handle, we cannot fetch the processes' command line but the function will return that the process is "unprotected" which is true. 

This case can lead to us attempting to read memory but not being able to do so.

I think this is fine for now, since this really should never happen since being able to read memory is officially documented to correspond to the protection level. https://learn.microsoft.com/en-us/windows/win32/procthread/process-security-and-access-rights#protected-processes

If we want to solve this case, we can return `true` for `isProtected` there but this may be slightly misleading because the actual system call said the process was unprotected.

pkg/process/procutil/process_windows.go
gengnamstyle: Removed!

pkg/util/winutil/process.go
iglendd: Yeah, it probably return signer level, ok, marking it as resolved, thank you

pkg/process/procutil/process_windows.go
iglendd: But can you semi-deterministically measure ball-park overall overhead opening one handle vs two handle. What are we talking about I am interested.

gengnamstyle: @clarkb7 do you have suggestions on how to best do this? I am considering doing profiling between the 2 `OpenProcessHandle` versions in a temporary test to see relative differences, but I don't think I have the best tooling as I'm on a mac and sshing into my vm.

iglendd: I did not realize it. I will test it. No worries.

iglendd: Collection of 5 non-protected processes takes 55-70 microseconds. If protection check is dropped for the same processes it takes 40-45 microseconds. Which means overall overhead for 100 processes would be around 1.5-2.5 milliseconds. 

```
OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION, FALSE, pid);
NtQueryInformationProcess(hProcess, ProcessProtectionInformation, &protection...)
OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION|PROCESS_VM_READ, FALSE, pid);
```

vs 

```
OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION|PROCESS_VM_READ, FALSE, pid);
```

I have measured that on a relatively modern laptop.

gengnamstyle: Wow thank you so much! I've actually have not done any profiling on the agent yet, so next time I'll include that as part of the PR. 

Anyhow, seems like the overhead is negligible?

pkg/process/procutil/process_windows.go
gengnamstyle: thanks for the catch. Updated

pkg/process/procutil/process_windows.go
gengnamstyle: I also just ran some benchmarking just testing the openProcessHandle with 1 process at a time with the following results:

```
BenchmarkOpenProcessHandleProtected-4             255981              6009 ns/op
BenchmarkOldOpenProcessHandleProtected-4           15889             77775 ns/op
BenchmarkOpenProcessHandleUnprotected-4           197953              6709 ns/op
BenchmarkOldOpenProcessHandleUnprotected-4        358704              3219 ns/op
```

A single unprotected process takes 6 microseconds vs 3 compared to the old method. 

What's interesting is for protected processes it takes ~10x the time likely from handling an error since the older version attempts to get more privileged access initially and on an error would fall back to a less privileged version.

With the newer version, we check for access or not before attempting.  

Here is the testing file as a gist that I used if you'd like to check.

https://gist.github.com/gengnamstyle/8348a4edc51807922fbc90914e63e5aa

gengnamstyle:
I'll be the release manager for this milestone so I don't mind manually testing, but after looking at the existing e2e tests there seem to be enough, but I'd like to go through the process of manually QA to gain some experience doing that and for my sanity!

pkg/process/procutil/process_windows.go
iglendd: > What's interesting is for protected processes, the older method takes ~10x the time likely from handling an error since the older version attempts to get more privileged access initially and on an error would fall back to a less privileged version.

This is actually a very good/great finding (I did not test it), and overall I think it would lower additional overhead or maybe completely negate it.

Yeas, I think I think that ~0.5 ms overhead (and probably less with your finding) is negligent in the large schema of things while yielding multiple benefits. I would only suggest to add a some performance related comments to this function depicting some aspects performance implications which have discussed in the is thread (without too much details), as well as security. This would help other people and avoid unnecessary or similar inquires or experiments (IMHO).

And yes, thank you for implementing this.

gengnamstyle:
>> manual validation should involve verifying that no regressions occur on the normal windows agent

>Would it be possible to add an e2e test for this instead of doing manual regression testing?

I'll be the release manager for this milestone so I don't mind manually testing, but after looking at the existing e2e tests there seem to be enough, but I'd like to go through the process of manually QA to gain some experience doing that and for my sanity!

iglendd (approved):

test/new-e2e/tests/process/windows_test.go
```diff
@@ -161,9 +165,25 @@ func (s *windowsTestSuite) TestAPIKeyRefreshAdditionalEndpoints() {
 	}, 2*time.Minute, 10*time.Second)
 }
 
-func assertProcessCheck(t *testing.T, env *environments.Host) {
+func assertProcessCMDArgs(t require.TestingT, processes []*agentmodel.Process, processCMDArgs []string) {
```
robertjli, line 168: Suggestion: `CMDArgs` is unclear, replace with `CommandLineArgs` or `CommandLineParams`?

test/new-e2e/tests/process/windows_test.go
```diff
@@ -161,9 +165,25 @@ func (s *windowsTestSuite) TestAPIKeyRefreshAdditionalEndpoints() {
...
+	for _, proc := range processes {
+		// command arguments include the first command/program which can differ depending on the path,
+		// so we compare the user provided arguments starting from index 1
+		assert.Truef(t, slices.Equal(proc.Command.Args[1:], processCMDArgs[1:]), "process args do not match. Expected %+v", processCMDArgs)
+	}
+}
+
+func getMatchingProcessesByName(payloads []*aggregator.ProcessPayload, processName string) []*agentmodel.Process {
```
robertjli, line 176: Thought: This differs from `filterProcesses` in that it accepts payloads instead of process lists. What do you think about renaming it to be more aligned with the existing function, and moving it to be nearby as well? I could see a name like `filterProcessPayloads` fitting well

test/new-e2e/tests/process/windows_test.go
```diff
@@ -6,7 +6,11 @@
 package process
 
 import (
+	"encoding/json"
 	"fmt"
+	agentmodel "github.com/DataDog/agent-payload/v5/process"
```
robertjli, line 11: Chore: please fix import ordering

test/new-e2e/tests/process/windows_test.go
```diff
@@ -161,9 +165,25 @@ func (s *windowsTestSuite) TestAPIKeyRefreshAdditionalEndpoints() {
 	}, 2*time.Minute, 10*time.Second)
 }
 
-func assertProcessCheck(t *testing.T, env *environments.Host) {
+func assertProcessCMDArgs(t require.TestingT, processes []*agentmodel.Process, processCMDArgs []string) {
+	for _, proc := range processes {
+		// command arguments include the first command/program which can differ depending on the path,
+		// so we compare the user provided arguments starting from index 1
+		assert.Truef(t, slices.Equal(proc.Command.Args[1:], processCMDArgs[1:]), "process args do not match. Expected %+v", processCMDArgs)
```
robertjli, line 172: Question: Are we expecting every single process in `processes` to have the exact same args?

Suggestion: `assert.Equal` does slice equality already, you don't need to call `assert.True(slices.Equal())`

test/new-e2e/tests/process/windows_test.go
```diff
@@ -292,14 +339,32 @@ func runDiskSpd(t *testing.T, remoteHost *components.RemoteHost) error {
...
+		"-r",
+		"-Sh",
+		"-w50",
+		"disk-speed-test.dat",
+	}
+	processName, err := runCMD(t, remoteHost, cmd)
+	return processName, cmd, err
+}
+
+func runCMD(t *testing.T, remoteHost *components.RemoteHost, cmd []string) (string, error) {
```
robertjli, line 359:
```suggestion
func runWindowsCommand(t *testing.T, remoteHost *components.RemoteHost, cmd []string) (string, error) {
```

Suggestion: Rename because this is package-private, if it's specific to Windows then we should name it as such

test/new-e2e/tests/process/windows_test.go
```diff
@@ -292,14 +339,32 @@ func runDiskSpd(t *testing.T, remoteHost *components.RemoteHost) error {
...
-		return err
+		return "", err
 	}
 
 	t.Cleanup(func() {
 		_ = session.Close()
 		_ = stdin.Close()
 	})
-	return nil
+	return fmt.Sprintf("%s.exe", cmd[0]), nil
```
robertjli, line 369: Question: Can we actually assume that we should append `.exe` to the command we run?

test/new-e2e/tests/process/windows_test.go
robertjli: Suggestion: This is a generic helper function, should it be moved to `testing.go`?

test/new-e2e/tests/process/windows_test.go
gengnamstyle: I'm not too sure, but this has been the case for all my testing and when I have seen the payloads. I think if issues do arise we can solve it later?

test/new-e2e/tests/process/windows_test.go
gengnamstyle: done!

test/new-e2e/tests/process/windows_test.go
gengnamstyle: I also had a similar thought, but it's been the case for all our current test cases, and I don't mind having this assumption for now until it breaks. 

updated as well.

test/new-e2e/tests/process/windows_test.go
gengnamstyle: done!

test/new-e2e/tests/process/windows_test.go
gengnamstyle: updated. thanks.

test/new-e2e/tests/process/windows_test.go
gengnamstyle: Renamed to CommandLineArgs and moved to `testing.go`.

pkg/process/procutil/process_windows.go
```diff
@@ -517,39 +529,39 @@ func fillProcessDetails(pid int32, proc *Process) error {
...
-	if imagePath, err := winutil.GetImagePathForProcess(procHandle); err == nil {
-		return &winutil.ProcessCommandParams{
-			CmdLine:   imagePath,
-			ImagePath: imagePath,
-		}
+	// check protection level to determine memory readability for PROCESS_VM_READ
+	// more info: https://learn.microsoft.com/en-us/windows/win32/procthread/process-security-and-access-rights#protected-processes
+	isProtected, err := winutil.IsProcessProtected(procHandle)
+	if err != nil {
+		log.Debugf("Couldn't access process %v protection info %v", pid, err)
```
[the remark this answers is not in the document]
gengnamstyle: > 1. Are we ok attempting to open with PROCESS_VM_READ in this case?

I'm okay with attempting to open with PROCESS_VM_READ because in the worst case, that will also fail, and we return the basic process handle which is the equivalent operation of the previous `OpenProcessHandle` function

> 2. Should we add to the debug log that we will attempt to open with PROCESS_VM_READ?
I think we should. I'll add that.

test/new-e2e/tests/process/windows_test.go
iglendd: traditionally all executable and processes have .exe extension. Is it not required formally but in practice other extensions are quite rare.

pkg/process/procutil/process_windows.go
```diff
@@ -517,39 +529,39 @@ func fillProcessDetails(pid int32, proc *Process) error {
...
-		procHandle, err = windows.OpenProcess(0x1000, false, uint32(pid))
-		if err != nil {
-			log.Debugf("Couldn't open process %v %v", pid, err)
-			return windows.Handle(0), err
-		}
+		log.Debugf("Couldn't open unprotected process with PROCESS_VM_READ access. Returning limited process info %v %v", pid, err)
+		return procHandle, isProtected, err
 	}
-	return procHandle, nil
+	defer windows.Close(procHandle)
```
wiyu, line 563: Why `defer` here when you return right after?

pkg/util/winutil/process_test.go
```diff
@@ -0,0 +1,117 @@
...
+		{
+			description: "protected process 3",
+			processName: "smss.exe",
+			expected:    true,
+		},
+		{
+			description: "protected process 4",
+			processName: "wininit.exe",
+			expected:    true,
+		},
```
wiyu, line 83: any reason for 4 of the what looks to be the same tests of verifying a protected process?

pkg/process/procutil/process_windows.go
```diff
@@ -497,14 +497,26 @@ func fillProcessDetails(pid int32, proc *Process) error {
...
 	proc.Username = userName
 
-	cmdParams := getProcessCommandParams(procHandle)
-
-	proc.Cmdline = ParseCmdLineArgs(cmdParams.CmdLine)
-	if len(cmdParams.CmdLine) > 0 && len(proc.Cmdline) == 0 {
-		log.Warnf("Failed to parse the cmdline:%s for pid:%d", cmdParams.CmdLine, pid)
+	imagePath, err := winutil.GetImagePathForProcess(procHandle)
+	if err != nil {
+		log.Debugf("Error retrieving exe path for pid %v %v", pid, err)
```
daniel-taf, line 502: I notice the way image path is retrieved seems different between `winutil.GetImagePathForProcess` ([link](https://github.com/DataDog/datadog-agent/blob/main/pkg/util/winutil/process.go#L345))  and `getCommandParamsForProcess32/getCommandParamsForProcess64` ([link](https://github.com/DataDog/datadog-agent/blob/main/pkg/util/winutil/process.go#L288)). Maybe they are actually equivalent but we should try to be certain of that since this could affect generation of the `command` tag on the backend and thus monitors.

pkg/util/winutil/process_test.go
gengnamstyle: I wanted to demonstrate more confidence that the function was working as intended by using more protected processes.

pkg/process/procutil/process_windows.go
gengnamstyle: Gotcha. Yeah will look into verifying this and or adding a test for this

pkg/process/procutil/process_windows.go
gengnamstyle: I return a different handle with different permissions right after, so I should close the previously opened handle to not unnecessarily "leak" memory (since I think the garbage collector would eventually clean it up, but still)

gengnamstyle: gotcha. Yeah I can remove the defer. I had the wrong assumption of defer occurring after the return when it happens right before!

pkg/process/procutil/process_windows.go
gengnamstyle: I did a bit more digging. There is sometimes a difference in the case for the image path such as
winutil.GetImagePathForProcess:`C:\Windows\System32\PING.EXE`
getCommandParamsForProcess:`C:\Windows\system32\ping.exe`

winutil.GetImagePathForProcess:`C:\tools\msys64\usr\bin\timeout.exe`
getCommandParamsForProcess:`c:\tools\msys64\usr\bin\timeout.exe`

winutil.GetImagePathForProcess:`C:\\Windows\\System32\\svchost.exe`
getCommandParamsForProcess:`C:\\Windows\\system32\\svchost.exe`

This occurs if `getCommandParamsForProcess` fails (due to protection, etc) because then we fallback to using `winutil.GetImagePathForProcess`. 

For the future, I think we should consolidate to using only `winutil.GetImagePathForProcess` since 
- `process_windows_toolhelp` also doesn't using the imagepath from `getCommandParamsForProcess` 
- how image path is parsed from that function is fairly complex
- dependent on being able to read process memory

For now, I replicated the previous behaviour 
1. getCommandParamsForProcess
2. on failure use winutil.GetImagePathForProcess

Now it is:
1. winutil.GetImagePathForProcess
2. override to use getCommandParamsForProcess if available

pkg/util/winutil/process.go
```diff
@@ -80,9 +83,39 @@ func NtQueryInformationProcess(h windows.Handle, class PROCESSINFOCLASS, target,
...
+		return err
 	}
-	return
+	return nil
+}
+
+// IsProcessProtected checks if the process has any level of protection and returns true if any protection is present
+// more info https://learn.microsoft.com/en-us/windows/win32/procthread/zwqueryinformationprocess
+// NOTE: in user mode, NtQueryInformationProcess and ZwQueryInformationProcess behave the same
+// more info https://learn.microsoft.com/en-us/windows-hardware/drivers/kernel/using-nt-and-zw-versions-of-the-native-system-services-routines
```
robertjli, line 94: Are these comments now outdated because you don't use ZwQueryInformationProcess?

pkg/util/winutil/process_test.go
```diff
@@ -0,0 +1,117 @@
...
+			if tc.processName != "lsass.exe" {
+				assert.Equal(t, tc.expected, isProcessProtected)
+			}
+			// CHECKING MEMORY ACCESS PRIVILEGES
+			if tc.expected {
+				_, err := GetCommandParamsForProcess(procHandle, true)
+				assert.Error(t, err)
+			} else {
+				// unprotected process's should allow reading memory with a higher privilege handle
+				procHandle, err := windows.OpenProcess(windows.PROCESS_QUERY_LIMITED_INFORMATION|windows.PROCESS_VM_READ, false, pid)
```
robertjli, line 110: Do we need to close these proc handles?

test/new-e2e/tests/process/windows_test.go
```diff
@@ -268,20 +287,31 @@ func (s *windowsTestSuite) TestManualProcessCheckWithIO() {
...
+		var checkOutput struct {
+			Processes []*agentmodel.Process `json:"processes"`
+		}
+
+		err := json.Unmarshal([]byte(check), &checkOutput)
+		require.NoError(c, err, "failed to unmarshal process check output")
+
+		procs := filterProcesses(process, checkOutput.Processes)
+		require.NotEmpty(c, procs, "'%s' process not found in check:\n%s\n", process, check)
+		assertProcessCommandLineArgs(c, procs, cmd)
```
wiyu, line 308: This unmarshalling is redundant as the `assertManualProcessCheck` already does this. We can deal with it in a follow-up PR to consolidate the assertion needs into the the `assertManualProcessCheck` function instead of copy/pasting the code

pkg/util/winutil/process.go
gengnamstyle: Nope! I am using NtQueryInformationProcess, but the documentation for how to check protection is from the documentation on ZwQueryInformationProcess

test/new-e2e/tests/process/windows_test.go
gengnamstyle: Sounds good, I made another ticket to consolidate a lot of the duplicate testing that already exists there: https://datadoghq.atlassian.net/browse/CXP-2229

pkg/util/winutil/process_test.go
gengnamstyle: Ideally yes, added.

daniel-taf (approved):

gengnamstyle:
For posterity: I created a ticket to handle the test refactoring for the future here: https://datadoghq.atlassian.net/browse/CXP-2229

PR #36761, opened 2025-05-07
[AMCC-8] dogstatsd: metric control remote reconfiguration

remeh:
### What does this PR do?

Add a remote config listener to subscribe to blocklist remote reconfigurations.

This PR is reviewable commit by commit if necessary.

I created one policy, validated that the Agent reconfigures with the metric in this policy.
Then created a second one, validated that the Agent reconfigures with metrics from both policy.
Updated one of the policy adding a metric, validated that the Agent reconfigures with metrics from both policy, with the new one.
Deleted one policy, validated that the Agent reconfigures with only the left policy.
Deleted the last policy, validated that the Agent reconfigures with the local config.

### Additional Notes

AMCC-8


comp/dogstatsd/server/server.go
```diff
@@ -497,16 +501,31 @@ func (s *server) SetExtraTags(tags []string) {
 	s.extraTags = tags
 }
 
+// SetBlocklist updates the metric names blocklist on all running worker.
+func (s *server) SetBlocklist(metricsName []string, matchPrefix bool) {
+	// each worker receives its own copy
+	for _, worker := range s.workers {
+		blocklist := newBlocklist(metricsName, matchPrefix)
```
vickenty, line 508: blocklist can be shared between workers. If `strings.Sort` inside `newBlocklist` temporarily modifies the slice even if it is sorted, it may cause data race or briefly break previous instances of blocklist using the same slice.

comp/dogstatsd/server/rc.go
```diff
@@ -0,0 +1,105 @@
...
+		// unmarshal all configs that can be unmarshalled
+		for configPath, v := range updates {
+			s.log.Info("received:", string(v.Config))
+			var config statsdBlocklistUpdate
+			if err := json.Unmarshal(v.Config, &config); err != nil {
+				applyStateCallback(configPath, state.ApplyStatus{
+					State: state.ApplyStateError,
+					Error: "error unmarshalling payload",
+				})
+				s.log.Error("can't unmarshal received config:", err)
```
vickenty, line 56:
```suggestion
				s.log.Errorf("can't unmarshal received config: %v", err)
```

comp/dogstatsd/server/rc.go
```diff
@@ -0,0 +1,105 @@
...
+			blocklistUpdates = append(blocklistUpdates, config.BlockedMetrics)
+		}
+	}
+
+	// sort by the configuration ID
+	// ------------------------
+
+	slices.SortFunc(blocklistUpdates, func(a, b blockedMetrics) int {
+		return a.ByName.ConfigurationID - b.ByName.ConfigurationID
+	})
```
vickenty, line 72: What is the reason for sorting here? The next block combines everything into a single set, the order is not important for that.

comp/dogstatsd/server/enrich.go
```diff
@@ -28,9 +28,11 @@ var (
 // enrichConfig contains static parameters used in various enrichment
 // procedures for metrics, events and service checks.
 type enrichConfig struct {
-	metricPrefix              string
-	metricPrefixBlacklist     []string
-	metricBlocklist           blocklist
+	// TODO(remy): this metric prefix / prefix blocklist
+	// is independent from the metric name blocklist, that's
+	// confusing and should be merged in the same implemnetation instead.
```
vickenty, line 33: It is confusing, but it's different: `metricPrefixBlacklist` does not block metrics, it blocks addition of  `metricPrefix`.

comp/dogstatsd/server/rc.go
```diff
@@ -0,0 +1,105 @@
...
+	// build a map with all the received metrics
+	// and then use the values as a blocklist
+	// ------------------------
+
+	m := make(map[string]struct{})
+	for _, update := range blocklistUpdates {
+		for _, metric := range update.ByName.Metrics {
+			m[metric.Name] = struct{}{}
+		}
+	}
```
vickenty, line 83: Just to double check, are updates really supposed to be combined into one like this?

comp/dogstatsd/server/rc.go
```diff
@@ -0,0 +1,105 @@
...
+
+	metricsName := make([]string, len(m))
+	i := 0
+	for metricName := range m {
+		metricsName[i] = metricName
+		i++
+	}
```
vickenty, line 90:
```suggestion
	metricsName := slices.Collect(maps.Keys(m))
```

Also, a small nit, but a more grammatically correct name would be `metricNames`.

comp/dogstatsd/server/server.go
```diff
@@ -523,6 +542,11 @@ func (s *server) handleMessages() {
 		go worker.run()
 		s.workers = append(s.workers, worker)
 	}
+
+	// create the metric names blocklist, one per worker
+	// -----------
+
+	s.SetBlocklist(s.config.GetStringSlice("statsd_metric_blocklist"), s.config.GetBool("statsd_metric_blocklist_match_prefix"))
```
tobz, line 549: So we set the initial blocklist to the value of the existing blocklist configuration parameters.. and then any RC-based updates would be sent to update the worker, which makes sense.

However, the main question I'm left with: is it intended that the RC-based blocklist _replaces_ the manually configured one?

comp/dogstatsd/server/rc.go
```diff
@@ -0,0 +1,105 @@
...
+	var blocklistUpdates []blockedMetrics
+
+	// unmarshal all the configurations received from
+	// the RC platform
+	// ------------------------
+
+	if len(updates) > 0 {
+		// unmarshal all configs that can be unmarshalled
+		for configPath, v := range updates {
+			s.log.Info("received:", string(v.Config))
```
tobz, line 49: Info level? Seems like this should be debug or lower, and ideally with a _little_ more detailed of a log message.

comp/dogstatsd/server/rc.go
tobz, line 90: Are you specifically trying to deduplicate here? If so, I think `slices.Sort`/`slices.Compact` would be cleaner.

comp/dogstatsd/server/rc.go
vickenty, line 49: Would it make sense to lower this do debug level?

```suggestion
			s.log.Debugf("received: %q", string(v.Config))
```

comp/dogstatsd/server/rc.go
```diff
@@ -0,0 +1,105 @@
...
+
+	if len(updates) == 0 {
+		return
+	}
+
+	var blocklistUpdates []blockedMetrics
+
+	// unmarshal all the configurations received from
+	// the RC platform
+	// ------------------------
```
tobz, line 44: For what it's worth: I noticed you've added a number of comments in this PR using this format.

While I personally love code comments, I think this format -- in particular, the last line always being `-----------------`, is unique amongst all existing code comments and feels a little out of place.

I would suggest just removing the separator lines, but everything else is fine.

tobz (commented):
Left some feedback.

comp/dogstatsd/server/rc.go
```diff
@@ -0,0 +1,105 @@
...
+			if err := json.Unmarshal(v.Config, &config); err != nil {
+				applyStateCallback(configPath, state.ApplyStatus{
+					State: state.ApplyStateError,
+					Error: "error unmarshalling payload",
+				})
+				s.log.Error("can't unmarshal received config:", err)
+				continue
+			}
+			if len(config.BlockedMetrics.ByName.Metrics) == 0 {
+				s.log.Warn("received a configuration with no metric")
```
vickenty, line 60: Is this unexpected? How one would remove blocklist otherwise?

comp/dogstatsd/server/server.go
remeh: That's why I'm creating a copy of metricNames in `newBlocklist`, am I missing something?

comp/dogstatsd/server/rc.go
remeh: The whole reason is to have a stable behaviour with regards to the processing order, especially in the future since the schema and content of these configurations might change. Do you think I'm being over-zealous? I agree it's not necessary for now.

comp/dogstatsd/server/enrich.go
remeh: Right

comp/dogstatsd/server/rc.go
remeh: Yes. Thanks for checking!

comp/dogstatsd/server/server.go
remeh: Fair, I have to think about this one.

comp/dogstatsd/server/rc.go
remeh: An empty update list. Still, a warn is non-necessary, moving it to debug.

comp/dogstatsd/server/server.go
remeh: Shipped in https://github.com/DataDog/datadog-agent/pull/36761/commits/9d4a81ceb42616d4e2b761eadeebef0326587547

comp/dogstatsd/server/rc.go
tobz: I think to Vickenty's point: we already union the metrics from each update by virtue of how we write them into the map `m`, so their order here doesn't matter.

ameske (approved):

comp/dogstatsd/server/rc.go
```diff
@@ -0,0 +1,92 @@
...
+type byName struct {
+	Metrics []metricEntry `json:"values"`
+}
+
+type metricEntry struct {
+	Name string `json:"metric_name"`
+}
+
+func (s *server) onBlocklistUpdateCallback(updates map[string]state.RawConfig, applyStateCallback func(string, state.ApplyStatus)) {
+	s.log.Debug("onBlocklistUpdateCallback received updates:", len(updates))
```
vickenty, line 34:
```suggestion
	s.log.Debugf("onBlocklistUpdateCallback received updates: %d", len(updates))
```

comp/dogstatsd/server/rc.go
```diff
@@ -0,0 +1,92 @@
...
+
+	// special case: we received a response from RC, but RC didn't have any
+	// configuration for this agent, let's restore the local config and return
+	if len(updates) == 0 {
+		s.restoreBlocklistFromLocalConfig()
+		return
+	}
```
tobz, line 41: You could potentially also skip this and just let the logic at the bottom of the method handle it, since `metricNames` will be empty if there were no updates.

comp/dogstatsd/server/rc.go
```diff
@@ -0,0 +1,92 @@
...
+
+	var blocklistUpdates []blockedMetrics
+
+	// unmarshal all the configurations received from
+	// the RC platform
+	if len(updates) > 0 {
```
vickenty, line 47: We already know this is always true because we checked this on line 38.

tobz (approved):
Looks good to me, but I left some non-blocking nits.

comp/dogstatsd/server/rc.go
```diff
@@ -0,0 +1,92 @@
...
+			if err := json.Unmarshal(v.Config, &config); err != nil {
+				applyStateCallback(configPath, state.ApplyStatus{
+					State: state.ApplyStateError,
+					Error: "error unmarshalling payload",
+				})
+				s.log.Errorf("can't unmarshal received config: %v", err)
+				continue
+			}
+			if len(config.BlockedMetrics.ByName.Metrics) == 0 {
+				s.log.Debug("received a configuration with no metric")
```
vickenty, line 61:
```suggestion
				s.log.Debug("received a blocklist configuration with no metrics")
```

comp/dogstatsd/server/rc.go
```diff
@@ -0,0 +1,92 @@
...
+		// unmarshal all configs that can be unmarshalled
+		for configPath, v := range updates {
+			s.log.Debugf("received: %q", string(v.Config))
+			var config statsdBlocklistUpdate
+			if err := json.Unmarshal(v.Config, &config); err != nil {
+				applyStateCallback(configPath, state.ApplyStatus{
+					State: state.ApplyStateError,
+					Error: "error unmarshalling payload",
+				})
+				s.log.Errorf("can't unmarshal received config: %v", err)
```
vickenty, line 57:
```suggestion
				s.log.Errorf("can't unmarshal received blocklist config: %v", err)
```

Also, would it make sense to report these errors back to the RC?

comp/dogstatsd/server/rc.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	var blocklistUpdates []blockedMetrics
+
+	// unmarshal all the configurations received from
+	// the RC platform
+	if len(updates) > 0 {
+		// unmarshal all configs that can be unmarshalled
+		for configPath, v := range updates {
+			s.log.Debugf("received: %q", string(v.Config))
```
vickenty, line 50:
```suggestion
			s.log.Debugf("received blocklist config: %q", string(v.Config))
```

comp/dogstatsd/server/rc.go
```diff
@@ -0,0 +1,92 @@
...
+	Name string `json:"metric_name"`
+}
+
+func (s *server) onBlocklistUpdateCallback(updates map[string]state.RawConfig, applyStateCallback func(string, state.ApplyStatus)) {
+	s.log.Debug("onBlocklistUpdateCallback received updates:", len(updates))
+
+	// special case: we received a response from RC, but RC didn't have any
+	// configuration for this agent, let's restore the local config and return
+	if len(updates) == 0 {
+		s.restoreBlocklistFromLocalConfig()
```
vickenty, line 39: Is this something that can happen normally or more of a bug? It seems to be to be more of the latter, because normally RC expects a callback for each update. In that case, would it make sense to preserve previously set configuration instead of reverting to startup config?

vickenty (approved):

comp/dogstatsd/server/rc.go
remeh: I'll confirm this but I'm fairly sure it's normal behaviour, meaning that no configuration has to be applied, intentionally.

comp/dogstatsd/server/rc.go
remeh: If you don't mind, I'll prefer to stick with an early return here.

PR #36762, opened 2025-05-07
[AGENTRUN-295] Avoid restarting the agent in disk e2e test

pgimalac:
### What does this PR do?
Update the disk check comparison e2e test to avoid restarting the agent between the runs of the Python and the Go version. 

### Motivation
Restarting the agent takes a long time (we wait until we're sure it's ready), so it happens that the disk metrics between the two runs can have a non-negligible difference.
This change allows running the two version one after the other without any restart, which should help with flakiness.


misteriaud (approved):

test/new-e2e/pkg/utils/e2e/client/host.go
```diff
@@ -438,6 +460,16 @@ func (h *Host) GetLogsFolder() (string, error) {
 	}
 }
 
+// JoinPath joins the path elements with the correct separator for the host
+func (h *Host) JoinPath(path ...string) string {
```
chouetz, line 464: There is not equivalent to the [pathlib](https://docs.python.org/3/library/pathlib.html) module in go? (I don't know if there is a need or it could make sense to enhance [this](https://pkg.go.dev/path#Join)?)

chouetz (approved):

jose-manuel-almaza (approved):

PR #36773, opened 2025-05-07
fix GoDI data races

brycekahle:
### What does this PR do?

Fixes several data races found running `go test -race`

### Motivation

Correct data synchronization


grantseltzer:
When testing your other PR for improving test assertions, a decent amount of tests fail with `Condition never satisfied`, meaning there was an issue instrumenting the new function (at least, instrumenting it in time) which I suspect is why I tried putting sleeps in. Do you expect these races were the culprit?

grantseltzer (approved):

PR #36774, opened 2025-05-07
APM: reduce calls to get container tags to improve performance (APMSP-1986)

ajgajg1134:
### What does this PR do?
Reduce the number of times we call the remote tagger to fetch container tags. Each time we do this we allocate more memory unnecessarily, especially in the case of `GetVersionDataFromContainerTags` which only needs two specific tags. @paullegranddc pointed out this function specifically as a place we could optimize.

In the process of looking at that function and the allocations profiles I realized that we were calling it for every _chunk_ in a payload which is unnecessary as the container ID is set at the payload level so there is no possibility for the tags to change within a payload. This drastically increases the number of allocations for this function compared to the exact same call being made in api.go.

This is a pure refactor and the resulting behavior should not change.

### Motivation
Reducing allocations / cpu time spent fetching container tags is good for performance overall. 

Using existing unit / integration / e2e tests to verify behavior has not changed. I also performed a quick local smoke test to ensure behavior was unchanged when no container ID is provided.


ekump (approved):
LGTM, FWIW

songy23 (approved):

PR #36775, opened 2025-05-07
switch GoDI to use process monitor using PEDS

brycekahle:
### What does this PR do?

Switches GoDI to a process monitor that is using the Process Event Data Stream (PEDS). It will synchronize with the procfs state on a configurable interval, if desired.

### Motivation

- Remove ability to drop process start events on initial sync with procfs
- USM process monitor was never intended to be shared by multiple modules


pkg/ebpf/process/callback.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	"sync"
+	"sync/atomic"
+
+	"github.com/DataDog/datadog-agent/pkg/eventmonitor/consumers"
+)
+
+// callbackMap is a helper struct that holds a map of callbacks and a mutex to protect it
+type callbackMap struct {
```
gjulianm, line 18: Why not export the `callbackMap`in `pkg/eventmonitor/consumers`? (or move everything to a common package)

pkg/ebpf/process/monitor.go
```diff
@@ -0,0 +1,201 @@
...
+	delete(m.knownPIDs, pid)
+	m.mtx.Unlock()
+
+	m.exitCallbacks.call(pid)
+}
+
+// SubscribeExec wraps SubscribeExec from [consumers.ProcessConsumer]
+func (m *Monitor) SubscribeExec(callback consumers.ProcessCallback) func() {
+	m.execCallbacks.add(callback)
+	return m.consumer.SubscribeExec(m.processStart)
```
gjulianm, line 194: It should only be necessary to call `consumer.SubscribeExec` once, no?

pkg/ebpf/process/monitor.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+// SubscribeExit wraps SubscribeExit from [consumers.ProcessConsumer]
+func (m *Monitor) SubscribeExit(callback consumers.ProcessCallback) func() {
+	m.exitCallbacks.add(callback)
+	return m.consumer.SubscribeExit(m.processExit)
```
gjulianm, line 200: Same  issue as `SubsribeExec`

cmd/system-probe/modules/dynamic_instrumentation.go
```diff
@@ -44,3 +52,10 @@ var DynamicInstrumentation = &module.Factory{
 		return true
 	},
 }
+
+// createGoDIProcessEventConsumer creates the process event consumer for the GoDI module. Should be called from the event monitor module
+func createGoDIProcessEventConsumer(evm *eventmonitor.EventMonitor) (err error) {
+	eventTypes := []consumers.ProcessConsumerEventTypes{consumers.ExecEventType, consumers.ExitEventType}
+	godiProcessEventConsumer, err = consumers.NewProcessConsumer("godi", 100, eventTypes, evm)
```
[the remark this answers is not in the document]
grantseltzer: Also why 100?

cmd/system-probe/modules/eventmonitor.go
```diff
@@ -99,5 +100,12 @@ func createEventMonitorModule(_ *sysconfigtypes.Config, deps module.FactoryDepen
 		}
 	}
 
+	if sysconfig.ModuleIsEnabled(config.DynamicInstrumentationModule) {
+		err := createGoDIProcessEventConsumer(evm)
+		if err != nil {
+			return nil, fmt.Errorf("cannot create event consumer for GoDI: %w", err)
```
grantseltzer, line 106:
```suggestion
			return nil, fmt.Errorf("cannot create event consumer for dynamic instrumentation: %w", err)
```

cmd/system-probe/modules/modules.go
```diff
@@ -24,7 +24,7 @@ var moduleOrder = []types.ModuleName{
 	config.OOMKillProbeModule,
 	config.EventMonitorModule, // there is a dependency from EventMonitor -> NetworkTracer, so EventMonitor has to follow NetworkTracer
 	config.ProcessModule,
-	config.DynamicInstrumentationModule,
+	config.DynamicInstrumentationModule, // GoDI needs to be after EventMonitor
```
grantseltzer, line 27:
```suggestion
	config.DynamicInstrumentationModule, // dynamic instrumentation needs to be after EventMonitor
```

grantseltzer (commented):
I'm running manual deployment testing on this now. But code looks great, thank you for contributing!

pkg/ebpf/process/monitor.go
brycekahle: It isn't a problem to call it multiple times. I was returning the wrong remove function though, which is now fixed.

pkg/ebpf/process/callback.go
brycekahle: because some of the internals are accessed within the package, and it is a small amount of code.

cmd/system-probe/modules/dynamic_instrumentation.go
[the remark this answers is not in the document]
brycekahle: 100 is the channel size for real-time events (not startup events)

pkg/ebpf/process/monitor.go
gjulianm: Ah I see, the callbacks use a map, so it just gets overwritten.

gjulianm (approved):

safchain (approved):

grantseltzer (approved):
Thank you very kindly

PR #36783, opened 2025-05-08
[Dynamic Instrumentation] Prevent BPF 512-byte stack limit exhaustion by avoiding forced inlining

GreenMatan:
### What does this PR do?
[Exploration Tests](https://github.com/DataDog/datadog-agent/pull/34423) revealed a regression in Dynamic Instrumentation codegen of BPF programs. The following error started to pop up:
> error: /var/tmp/datadog-agent/system-probe/build/dynamicinstrumentation.c-be86c2e973338ef2f2688a235b67562ef7e4944aa8181070d03dafd9f1ecb171:6132:11: in function google_golang_org_protobuf_TestIntegration i32 (%struct.pt_regs*): Looks like the BPF stack limit of 512 bytes is exceeded. Please move large on stack variables into BPF per-cpu array map.

The root cause is due to our DI's basic block of expressions being inlined with the attribute `__always_inline`.  In Go Dynamic Instrumentation we use several expressions to build up a state machine that is able to serialize arbitrary variables out of memory. The composition of these expressions, when inlined, exceeded the BPF limitation of up to 512 bytes stack usage.
Marking functions with `always_inline` is [no longer a requirement since Kernel v4.16](https://github.com/torvalds/linux/commit/cc8b0b92a1699bc32f7fec71daa2bfc90de43a4d) where function calls were introduced.

### Motivation
Keep the stack usage of Dynamic Instrumentation BPF programs within the 512 bytes limit.

- Exploration tests is the main tool to capture these type of issues and make sure we don't have further regressions. Without this fix, most of the probes are failing with the 512 bytes limit error.
- There's a [separate DRAFT PR](https://github.com/DataDog/datadog-agent/pull/36791) where I aim to introduce new tests that validate this fix.

### DIFF
--- a/pkg/dynamicinstrumentation/codegen/c/expressions.h
+++ b/pkg/dynamicinstrumentation/codegen/c/expressions.h
@@ -1,9 +1,23 @@
 #ifndef DI_EXPRESSIONS_H
 #define DI_EXPRESSIONS_H
 
+// The Linux kernel defines `noinline` as a macro:
+//     #define noinline __attribute__((__noinline__))
+// (reference: https://elixir.bootlin.com/linux/v5.15/source/include/linux/compiler_attributes.h#L231)
+// Our code (via bpf_helpers.h) defines `__noinline` as:
+//     #define __noinline __attribute__((noinline))
+// If the kernel's `noinline` macro is active during preprocessing,
+// it causes `__noinline` to expand into invalid syntax:
+//     __attribute__((__attribute__((__noinline__))))
+//
+// To avoid this conflict, we explicitly undefine `noinline`
+#ifdef noinline
+#undef noinline
+#endif
+
 // read_register reads `element_size` bytes from register `reg` into a u64 which is then pushed to
 // the top of the BPF parameter stack.
-static __always_inline int read_register(expression_context_t *context, __u64 reg, __u32 element_size)
+static int __noinline read_register(expression_context_t *context, __u64 reg, __u32 element_size)
 {
     long err;
     __u64 valueHolder = 0;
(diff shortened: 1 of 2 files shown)


ajwerner (approved):
This makes sense to me. I'm looking forward to learning more about the testing infrastructure you've been working on!

pkg/dynamicinstrumentation/codegen/c/expressions.h
```diff
@@ -1,9 +1,23 @@
...
+//     #define noinline __attribute__((__noinline__))
+// (reference: https://elixir.bootlin.com/linux/v5.15/source/include/linux/compiler_attributes.h#L231)
+// Our code (via bpf_helpers.h) defines `__noinline` as:
+//     #define __noinline __attribute__((noinline))
+// If the kernel's `noinline` macro is active during preprocessing,
+// it causes `__noinline` to expand into invalid syntax:
+//     __attribute__((__attribute__((__noinline__))))
+//
+// To avoid this conflict, we explicitly undefine `noinline`
+#ifdef noinline
```
brycekahle, line 14: Should we pull this into `bpf_helpers.h`?

GreenMatan: That's a great question. I've tried to place it in `bpf_helpers.h` and then I started to encounter weird compilation errors
> /usr/src/linux-headers-5.15.0-73/include/linux/memory_hotplug.h:329:18: error: expected ';' after top level declarator
extern void __ref free_area_init_core_hotplug(int nid);

I did manage to workaround it by redefining `__noinline` as `#define __noinline __attribute__((__noinline__))` (with `__noinline__` instead of `noinline`) in `bpf_helpers.h`, but then I read that it's not encouraged to use `__noinline__` and that it's meant to be used from inside the linux kernel so I decided to scope this change to DI.
This whole problem might be unique to the way we generate bpf programs on the fly and the headers [we pull in](https://github.com/DataDog/datadog-agent/blob/main/pkg/dynamicinstrumentation/codegen/c/dynamicinstrumentation.c).

brycekahle: The `#include <asm/ptrace.h>` include does seem suspect.

GreenMatan: Resolving for now, feel free to ping me if there's any push back

brycekahle (approved):

tylfin (approved):

PR #36786, opened 2025-05-08
cnm: Remove unused telemetry component argument

guyarb:
### What does this PR do?

Removes unused telemetryComponent argument from NewState of the tracer

### Motivation

code cleanup


iglendd (approved):

hmahmood:
We want to keep this around since we do want to use it to register telemetry instead of using the global telemetry object. That work is scheduled but delayed at the moment.

PR #36792, opened 2025-05-08
[Dynamic Instrumentation] Performing per go binary analysis

GreenMatan:
### What does this PR do?
Go DI track processes that have `DD_DYNAMIC_INSTRUMENTATION_ENABLED` set to `true` (along with some extra checks). For an incoming RC request to instrument a function, all these processes are being collectively analyzed to extract DWARF information that is later used to build up the BPF program.
In our testing infra (where `mem_config_manager.go` is used) we haven't handled each process individually for probes installation/deletion which failed the exploration tests. Processes without DWARF (or with incomplete one) sabotaged the installation of probes for other legitimate processes.

The changes in this PR track each process success/fail to read DWARF and use this info later in the decision if probes should be handled for that process or not.

### Motivation
Straightening our testing infra, especially for exploration testing.

[Exploration tests](https://github.com/DataDog/datadog-agent/pull/34423) were used to validate these changes.

### DIFF
--- a/pkg/dynamicinstrumentation/diconfig/binary_inspection.go
+++ b/pkg/dynamicinstrumentation/diconfig/binary_inspection.go
@@ -21,23 +21,28 @@ import (
 // inspectGoBinaries goes through each service and populates information about the binary
 // and the relevant parameters, and their types
 // configEvent maps service names to info about the service and their configurations
-func inspectGoBinaries(configEvent ditypes.DIProcs) error {
+// it returns a map of PID -> bool stating for each PID if the analysis has succeeded or not
+// and an error when none of the processes have succeeded to be analyzed
+func inspectGoBinaries(configEvent ditypes.DIProcs) (map[ditypes.PID]bool, error) {
 	var err error
 	var inspectedAtLeastOneBinary bool
-	for i := range configEvent {
-		err = AnalyzeBinary(configEvent[i])
+	statuses := make(map[ditypes.PID]bool)
+	for pid, proc := range configEvent {
+		err = AnalyzeBinary(proc)
 		if err != nil {
-			log.Infof("inspection of PID %d (path=%s) failed: %s", configEvent[i].PID, configEvent[i].BinaryPath, err)
+			log.Infof("inspection of PID %d (path=%s) failed: %s", proc.PID, proc.BinaryPath, err)
+			statuses[pid] = false
 		} else {
+			statuses[pid] = true
 			inspectedAtLeastOneBinary = true
 		}
 	}
 
 	if !inspectedAtLeastOneBinary {
-		return fmt.Errorf("failed to inspect all tracked go binaries")
+		return statuses, fmt.Errorf("failed to inspect all tracked go binaries")
 	}
 
-	return nil
+	return statuses, nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/dynamicinstrumentation/diconfig/binary_inspection.go
```diff
@@ -21,23 +21,26 @@ import (
 // inspectGoBinaries goes through each service and populates information about the binary
 // and the relevant parameters, and their types
 // configEvent maps service names to info about the service and their configurations
-func inspectGoBinaries(configEvent ditypes.DIProcs) error {
+func inspectGoBinaries(configEvent ditypes.DIProcs) (map[ditypes.PID]bool, error) {
```
andreimatei, line 24: maybe comment on the returned map, and also on when an error is returned?

GreenMatan: Done in 90c911b2c371aed58f206304eb575a2174cf3c3c.

pkg/dynamicinstrumentation/diconfig/binary_inspection.go
```diff
@@ -21,23 +21,28 @@ import (
...
 	var err error
 	var inspectedAtLeastOneBinary bool
-	for i := range configEvent {
-		err = AnalyzeBinary(configEvent[i])
+	statuses := make(map[ditypes.PID]bool)
+	for pid, proc := range configEvent {
+		err = AnalyzeBinary(proc)
 		if err != nil {
-			log.Info("inspection of PID %d (path=%s) failed: %w", configEvent[i].PID, configEvent[i].BinaryPath, err)
+			log.Info("inspection of PID %d (path=%s) failed: %w", proc.PID, proc.BinaryPath, err)
```
grantseltzer, line 33:
```suggestion
			log.Info("inspection of PID %d (path=%s) failed: %s", proc.PID, proc.BinaryPath, err)
```

pkg/dynamicinstrumentation/diconfig/mem_config_manager.go
```diff
@@ -100,6 +100,11 @@ func (cm *ReaderConfigManager) update() error {
 		}
 
 		for pid, procInfo := range updatedState {
+			if statuses[pid] == false {
+				log.Info("Skipped the installation/deletion of probes for pid %d - failed to analyze its binary", pid)
```
grantseltzer, line 104: Will this be a particularly noisy log? Does it only occur when we have a probe configuration for a  proc or it's all procs that we're skipping (i.e. completely irrelevant ones)?

GreenMatan: It's will be as noise as the mutation of probes are. The flow is active when there's a probe configuration change, and the log is for all those processes we track (i.e processes with DI enabled). I don't think it will be that noisy, thus I'd say let's keep and and revisit it later. WDYT?

grantseltzer: Sounds good

grantseltzer (approved):
Works for me

tylfin (approved):

PR #36794, opened 2025-05-08
[EBPF] kmt: reduce memory usage of the complexity calculator

gjulianm:
### What does this PR do?

This PR reduces and limits the memory usage of the complexity calculator. It improves where we call Go to garbage collect, making it more frequent, and it also uses `ulimit` to ensure that the system kills the calculator and not the entire SSH process.

### Motivation

As the eBPF programs we have grow, the memory usage grows too due to the verifier log. Given that the calculator runs inside of the VMs, even if Go ends up releasing the memory, the VM might not do so. 

Additionally, we have seen that if the calculator uses too much memory, the system might kill either the SSH connector program, which makes it harder to distinguish which failures are due to memory usage and which ones due to connectivity issues.

CI passing.

Also ensure that, with lower limits, `ulimit` kills only the calculator process: https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/930975382

### DIFF
--- a/test/new-e2e/system-probe/test/micro-vm-init.sh
+++ b/test/new-e2e/system-probe/test/micro-vm-init.sh
@@ -42,8 +42,8 @@ if [ "${COLLECT_COMPLEXITY:-}" = "yes" ]; then
     test_root=$(echo "$@" | sed 's/.*-test-root \([^ ]*\).*/\1/')
     export DD_SYSTEM_PROBE_BPF_DIR="${test_root}/pkg/ebpf/bytecode/build/${arch}"
 
-    # Limit maximum memory usage of the calculator to 5GB to avoid OOM errors affecting the entire connector
-    ulimit -v $((5 * 1024 * 1024))
+    # Limit maximum memory usage of the calculator to avoid OOM errors affecting the entire connector
+    ulimit -v $((6 * 1024 * 1024))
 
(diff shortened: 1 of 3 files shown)


test/new-e2e/system-probe/test/micro-vm-init.sh
```diff
@@ -38,10 +38,13 @@ if [ "${COLLECT_COMPLEXITY:-}" = "yes" ]; then
...
         arch="arm64"
     fi
-    
+
     test_root=$(echo "$@" | sed 's/.*-test-root \([^ ]*\).*/\1/')
     export DD_SYSTEM_PROBE_BPF_DIR="${test_root}/pkg/ebpf/bytecode/build/${arch}"
-    
+
+    # Limit maximum memory usage of the calculator to avoid OOM errors affecting the entire connector
+    ulimit -v $((6 * 1024 * 1024))
```
usamasaqib, line 46: 6MB feels too conservative. I think we can give 1-2 Gigs.

gjulianm: The value is in kilobytes, so this is 6GB

usamasaqib (approved):

test/new-e2e/system-probe/test/micro-vm-init.sh
gjulianm: Added it, with a bit of a smaller limit than the one in ulimit, I'll check if it works better and then merge this.

PR #36803, opened 2025-05-08
[AGNTLOG-74] Fix Logs File Launcher Blocking AD

soberpeach:
### What does this PR do?

Moves `FilesToTail()` to a go function to run concurrently alongside the rest of the launcher. This allows the launcher to run without being blocked when gathering file tailers for sources that can take a while to fetch (such as those on remote file shares).

### Motivation

In the previous implementation, the `scan` function was called synchronously by the file launcher. If the scan operation took a long time, this synchronous call would block the rest of the launcher leading to missed sourcs.. Now, the launcher will proceed while `FilesToTail()` runs in another thread so it won't miss sources anymore.

1. Benchmarked scan to confirm that increasing numbers of files were the culprit
2. Built and ran the agent

None that I can think of. 

### Additional Notes

Would like more assistance QA'ing as I'm not very familiar with concurrency.


drichards-87 (approved):

pkg/logs/launchers/file/launcher.go
```diff
@@ -117,10 +123,21 @@ func (s *Launcher) run() {
...
-			s.cleanUpRotatedTailers()
-			// check if there are new files to tail, tailers to stop and tailer to restart because of file rotation
-			s.scan()
+			if s.scanMutex.TryLock() {
+				s.wg.Add(1)
+				go func() {
+					defer s.scanMutex.Unlock()
+					defer s.wg.Done()
+
+					s.cleanUpRotatedTailers()
```
gh123man, line 132: IMO this dones't need to be in the goroutine and we can do this synchronously as it's not an expensive operation.

pkg/logs/launchers/file/launcher.go
```diff
@@ -117,10 +123,21 @@ func (s *Launcher) run() {
...
+					// check if there are new files to tail, tailers to stop and tailer to restart because of file rotation
+					s.scan()
```
gh123man, line 134: Instead of granular locking, did you consider pushing 
`files := s.fileProvider.FilesToTail(s.validatePodContainerID, s.activeSources)` to a goroutine instead of the whole scan method? 

AFAIK, `FilesToTail` is the expensive operation, and the majority of `scan` can be done synchronously (to avoid locking). 

`FilesToTail` is mostly stateless, so you could push it's results to a channel, and consume them on the run loop with relatively minimal refactoring. (for example, we may need to copy `s.activeSources`)

pkg/logs/launchers/file/launcher.go
soberpeach: Moved out 👍

pkg/logs/launchers/file/launcher.go
soberpeach: Okay, that makes sense. The latest commit moves the `FilesToTail` function into it's own go routine and changed the functionality of `addSource` back to it's original behavior. A couple of mutexes are needed however. In order to assure that the `scan` doesn't _remove tailers_ from sources that have gone through `addSource` but have not yet  been processed by `getFiles`, a mutex for the `launcher.addedSourceTailers` was created. Additionally, to ensure that `launcher.activeFiles` wasn't changed while it was being used in `getFiles`, a mutex was added around that.

pkg/logs/launchers/file/launcher.go
```diff
@@ -294,6 +334,8 @@ func (s *Launcher) launchTailers(source *sources.LogSource) {
 			continue
 		}
 
+		s.addedSourceTailers = append(s.addedSourceTailers, file)
```
gh123man, line 337: Maybe we only want to do this if `startNewTailer` returns true? (i'd double check)

pkg/logs/launchers/file/launcher.go
gh123man: Left a few more comments

pkg/logs/launchers/file/launcher.go
```diff
@@ -109,18 +115,23 @@ func (s *Launcher) run() {
...
+			go s.getFiles()
+		case files := <-s.filesChan:
 			s.cleanUpRotatedTailers()
-			// check if there are new files to tail, tailers to stop and tailer to restart because of file rotation
-			s.scan()
+
+			s.scan(files)
+			scanTicker.Reset(s.scanPeriod)
 		case <-s.stop:
+			s.wg.Wait()
```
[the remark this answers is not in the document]
soberpeach: Done!

pkg/logs/launchers/file/launcher.go
```diff
@@ -144,21 +155,43 @@ func (s *Launcher) cleanup() {
...
+func (s *Launcher) scan(files []*tailer.File) {
+	tailersFilesCopy := make([]*tailer.File, len(files))
+	copy(tailersFilesCopy, files)
+
+	s.addedSourceTailersMutex.Lock()
+	defer s.addedSourceTailersMutex.Unlock()
+
+	// Union so scan doesn't remove files added by addSource but not yet found by FilesToTail
+	tailersFilesCopy = append(tailersFilesCopy, s.addedSourceTailers...)
+	s.addedSourceTailers = s.addedSourceTailers[:0]
```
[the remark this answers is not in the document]
soberpeach: Sounds good, removed the mutex

pkg/logs/launchers/file/launcher.go
soberpeach: That makes sense, instead of just appending all files.

pkg/logs/launchers/file/launcher.go
```diff
@@ -158,21 +178,26 @@ func (s *Launcher) cleanup() {
...
-// scan checks all the files we're expected to tail, compares them to the currently tailed files,
+// resolveActiveTailers checks all the files we're expected to tail, compares them to the currently tailed files,
 // and triggers the required updates.
 // For instance, when a file is logrotated, its tailer will keep tailing the rotated file.
 // The Scanner needs to stop that previous tailer, and start a new one for the new file.
-func (s *Launcher) scan() {
-	files := s.fileProvider.FilesToTail(s.validatePodContainerID, s.activeSources, s.registry)
+func (s *Launcher) resolveActiveTailers(files []*tailer.File) {
+	tailersFilesCopy := make([]*tailer.File, len(files))
+	copy(tailersFilesCopy, files)
```
gh123man: Why do we need to copy the files slice here?

pkg/logs/launchers/file/launcher.go
```diff
@@ -158,21 +178,26 @@ func (s *Launcher) cleanup() {
...
+
+	// Union so scan doesn't remove files added by addSource but not yet found by FilesToTail
```
gh123man: Can you expand on this comment? Can we describe why scan might remove files added by `addSource`? 
Followup question: Is the opposite possible? can scan add back files that were removed by `removeSource`?

pkg/logs/launchers/file/launcher.go
```diff
@@ -56,6 +57,10 @@ type Launcher struct {
 	scanPeriod             time.Duration
 	flarecontroller        *flareController.FlareController
 	tagger                 tagger.Component
+	filesChan              chan []*tailer.File
+	addedSourceTailers     []*tailer.File
+	ctx                    context.Context
```
gh123man: Since `ctx` and `cancel` are only ever used in the run loop, can we move these into function local variables inside `run()`? this way we dont carry additioal state on the launcher struct (since we don't need it anywhere else)

comp/logs/agent/agentimpl/agent_test.go
```diff
@@ -104,6 +104,8 @@ func (suite *AgentTestSuite) SetupTest() {
 	suite.configOverrides["logs_config.run_path"] = suite.testDir
 	// Shorter grace period for tests.
 	suite.configOverrides["logs_config.stop_grace_period"] = 1
+	// Set a short scan period to allow it to run in the time period of the tcp and http tests
+	suite.configOverrides["logs_config.file_scan_period"] = 1
```
gh123man, line 108: I wonder why this wasn't needed before

pkg/logs/launchers/file/launcher.go
```diff
@@ -56,6 +57,10 @@ type Launcher struct {
 	scanPeriod             time.Duration
 	flarecontroller        *flareController.FlareController
 	tagger                 tagger.Component
+	filesChan              chan []*tailer.File
+	addedSourceTailers     []*tailer.File
```
gh123man: NIT: Can we choose a better name for this var? `sources` isn't really accurate since it's a slice of files.
I'm thinking something like `filesTailedBetweenScans`

pkg/logs/launchers/file/launcher.go
```diff
@@ -120,18 +129,29 @@ func (s *Launcher) run() {
...
 		case source := <-s.addedSources:
 			s.addSource(source)
 		case source := <-s.removedSources:
 			s.removeSource(source)
 		case <-scanTicker.C:
+
+			activeSourcesCopy := make([]*sources.LogSource, len(s.activeSources))
+			copy(activeSourcesCopy, s.activeSources)
+
+			scanTicker.Stop()
```
gh123man, line 143: Should we clear `addedSourceTailers` here? 
As I understand it, `addedSourceTailers` is used to track files that may have been scheduled during a scan. However if a scan is not ongoing and a source is scheduled, it is still added to `addedSourceTailers`. But it's also tracked in `activeSources`, so it will be considered in the next scan. 

by resetting `addedSourceTailers`, we only have to merge additional sources that were added during an active scan, and not between subsiqunt scans. 

(Maybe I am misunderstanding, I think some additional comments around `addedSourceTailers` and how it's supposed to work would be helpful)

gh123man (changes requested):
Mostly questions and clarifications.

> Added a new unit test to test the behavior of the new functionality for addSource and removeSource.

Did this get lost in a merge? I don't see a new test currently.

pkg/logs/launchers/file/launcher.go
soberpeach: I think this is a holdout from a previous version of the feature that used locks instead of the run loop. This isn't necessary anymore so I'll change it back.

pkg/logs/launchers/file/launcher.go
soberpeach: I've updated the comment to reflect clarify this. And good catch, it might be possible actually that scan could add back files that were removed by `removeSource`, I'll update the code to reflect this possibility.

soberpeach: Actually, looking at the `removeSource()` function it seems this is likely already accounted for. The comment mentions that 

> no need to stop the tailer here, it will be stopped in the next iteration of scan.

So with the new change it's _possible_ that the source will be scheduled for one iteration of scan longer than before thought that possibility is small. It seems like we're currently okay with that tradeoff however.

pkg/logs/launchers/file/launcher.go
soberpeach: Yep, changing that now

comp/logs/agent/agentimpl/agent_test.go
soberpeach: Honestly not sure either 🤔 I might change it back if it doesn't affect anything.

pkg/logs/launchers/file/launcher.go
soberpeach: renamed 👍

pkg/logs/launchers/file/launcher.go
soberpeach: Ah yes, that's a better place to clear it. Right now it's only cleared in `scan()` but it can catch duplicates if files are added before `FilesToTail()` is called. Changing it now.

pkg/logs/launchers/file/launcher.go
gh123man: Nice thanks for confirming 👍

pkg/logs/launchers/file/launcher.go
```diff
@@ -162,8 +182,17 @@ func (s *Launcher) cleanup() {
...
-	files := s.fileProvider.FilesToTail(s.validatePodContainerID, s.activeSources, s.registry)
+func (s *Launcher) scan(files []*tailer.File) {
+	// FilesToTail(), which scan() receives the files parameter from, is called in
+	// the main run loop of launcher but runs parallel in a go function. It is
+	// therefore possible that addSource() can be called while FilesToTail() is
+	// still running. Since FilesToTail() is only passed a copy of activeSources
+	// it is therefore possible that it would miss the newly added source and
+	// therefore scan() would unschedule the appropriate tailer. In order to
+	// mitigate that possibility, any tailers started while FilesToTail() is
+	// running need to be added here to prevent scan() from unscheduling them.
```
gh123man: NIT: Slightly reworeded to further increase clarity: 

```
// scan() receives the files parameter from FilesToTail(), which is called in
// the main run loop of launcher. FilesToTail() is always executed concurrently. 
// It is therefore possible that addSource() can be called while FilesToTail() is
// still running. Since FilesToTail() is only passed a copy of activeSources
// it is possible that it would miss new sources added by addsource()
// therefore scan() would unschedule a tailer added during a concurrent scan.
// In order to mitigate that possibility, any tailers started while FilesToTail() is
// running need to be merged with the result of FilesToTail() to prevent scan() from 
// unscheudling them.
```

Feel free to adapt if I missed some context. 

I'm also wondering if we should rename `scan` -> `resolveTailers` or something like that. Since the scan really happens in `FilesToTail` now.

soberpeach: LGTM

I actually renamed `scan()` to `resolveActiveTailers()` earlier in the feature but then changed it back because I felt the behavior was too similar. I'll change it back again.

gh123man (approved):

PR #36804, opened 2025-05-08
[AGENTCFG-249] Adding secret_backend_type and secret_backend_config

rahulkaukuntla:
### What does this PR do?

We need to add these config options to allow customers to use the secret-generic-connector executable that is being shipped as a side binary with the agent--instead of needing to specify an executable script to retrieve their secrets, customers can now just specify the backend they are using (`secret_backend_type`), and any configuration that said backend requires (`secret_backend_config`).


releasenotes/notes/secret_backend_type-and-secret_backend_config-77479d1dc04353f5.yaml
```diff
@@ -0,0 +1,11 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+features:
+  - |
+    Adding the secret_backend_type and secret_backend_config config options to allow for refreshing secrets using Datadog's Secret Backend.
```
brett0000FF:
```suggestion
    Adds `secret_backend_type` and `secret_backend_config` options to simplify secret retrieval. Instead of providing a custom executable script, you can now configure the Agent to use the built-in `secret-generic-connector` to fetch secrets from a specified backend.
```

brett0000FF (approved):

comp/core/secrets/secretsimpl/fetch_secret.go
```diff
@@ -109,6 +109,14 @@ func (r *secretResolver) fetchSecret(secretsHandle []string) (map[string]string,
 		"version": secrets.PayloadVersion,
 		"secrets": secretsHandle,
 	}
+	if r.backendType != "" {
+		payload["type"] = r.backendType
+	}
+	// the backend config map can't be directly added to the payload, because there could be a
+	// "secrets" key in the backend config map, which would override the secrets key in the payload
```
hush-hush: I'm not sure I understand this comment

comp/core/secrets/secretsimpl/secrets.go
```diff
@@ -246,6 +334,35 @@ func (r *secretResolver) Configure(params secrets.ConfigParams) {
...
+				return fmt.Errorf("key '%s' should be a map", key)
+			}
+
+			for nestedKey := range nestedConfig {
+				if _, ok := nestedAllowed[nestedKey]; !ok {
+					return fmt.Errorf("unsupported nested key '%s.%s' for backend '%s'", key, nestedKey, backendType)
+				}
+			}
+		}
+	}
```
hush-hush: Do we need to validate the config ? What is the purpose of it ?

This also prevent customers from leveraging this feature with their custom executable. For example, the following config would not be supported:
```
secret_backend_executable: /path/to/custom/binary
secret_backend_config:
    test: 1234 
```

comp/core/secrets/secretsimpl/secrets.go
```diff
@@ -217,7 +282,30 @@ func (r *secretResolver) Configure(params secrets.ConfigParams) {
...
+	r.backendConfig = params.Config
+	if r.backendType != "" {
+		if err := validateTypeAndConfig(r.backendType, r.backendConfig); err != nil {
+			log.Errorf("%v", err)
+		}
+	}
 	r.backendCommand = params.Command
+	r.defaultBackendUsed = false
+	// only use the backend type option if the backend command is not set
+	if r.backendType != "" && r.backendCommand == "" {
```
hush-hush, line 234: Let's add a warning when both backend type and command are set saying we're ignoring backend type.

comp/core/secrets/secretsimpl/secrets.go
```diff
@@ -217,7 +282,30 @@ func (r *secretResolver) Configure(params secrets.ConfigParams) {
...
+		if runtime.GOOS == "windows" {
+			r.backendCommand = path.Join(defaultpaths.GetInstallPath(), "..", "secret-generic-connector.exe")
+		} else {
+			nonWindowsPath := path.Join(defaultpaths.GetInstallPath(), "..", "..", "embedded", "bin", "secret-generic-connector")
+			permission, _ := filesystem.NewPermission()
+			err := permission.RestrictAccessToUser(nonWindowsPath)
+			if err != nil {
+				log.Warnf("Cannot give the user access to %s: %s", nonWindowsPath, err)
+			}
```
hush-hush: I don't think we should do this here, it should be done either by the installer. If not, we have a window of time between the install of the agent and it's startup where this is not done.

Also, this should already be done. The Unix installer should make all the file from the Agent package owned by the Agent user, no ?

comp/core/secrets/secretsimpl/secrets.go
```diff
@@ -88,10 +150,13 @@ type secretResolver struct {
 	// list of handles and where they were found
 	origin handleToContext
 
+	backendType             string
+	backendConfig           map[string]interface{}
 	backendCommand          string
 	backendArguments        []string
 	backendTimeout          int
 	commandAllowGroupExec   bool
+	defaultBackendUsed      bool
```
hush-hush:
```suggestion
	embeddedBackendUsed      bool
```

comp/core/secrets/secretsimpl/secrets.go
```diff
@@ -672,12 +789,12 @@ func (r *secretResolver) GetDebugInfo(w io.Writer) {
...
-	err = checkRights(r.backendCommand, r.commandAllowGroupExec)
-
 	permissions := "OK, the executable has the correct permissions"
-	if err != nil {
-		permissions = fmt.Sprintf("error: %s", err)
+	if !r.defaultBackendUsed {
+		err = checkRights(r.backendCommand, r.commandAllowGroupExec)
+		if err != nil {
+			permissions = fmt.Sprintf("error: %s", err)
+		}
```
hush-hush: We should have a dedicated message when using the embedded binary since we don't enforce rights in the same way (ie: we should not say that is has correct permissions).

pkg/config/config_template.yaml
```diff
@@ -865,10 +865,33 @@ api_key:
...
+## SET this option if you are using the default secret backend.
+## If secret_backend_config is set, specify the config that is needed to pull secrets from the backend.
+#
+# secret_backend_config:
+#   <KEY_1>: <VALUE_1>
+#   <KEY_2>: <VALUE_2>
+#   <KEY_3>:
+#     <KEY_3_1>: <VALUE_3_1>
+#     <KEY_3_2>: <VALUE_3_2>
+
```
hush-hush:
```suggestion
## @param secret_backend_type - string - optional
## @env DD_SECRET_BACKEND_TYPE - string - optional
## This feature is in preview.
##
## `secret_backend_type` is the type of backend where secrets are stored.
## Supported baclend are: "aws.secrets", "aws.ssm", "azure.keyvault", "hashicorp.vault",
## "file.json", "file.yaml"
## For more information see: https://docs.datadoghq.com/agent/configuration/secrets-management
##
## This option is ignore if 'secret_backend_command' it set.
#
# secret_backend_type: <BACKEND_TYPE>

## @param secret_backend_config - map - optional
## @env DD_SECRET_BACKEND_CONFIG - map - optional
## This feature is in preview.
##
## The section contains information required to resolved secrets.
## For more information see: https://docs.datadoghq.com/agent/configuration/secrets-management
#
# secret_backend_config:
#   <KEY_1>: <VALUE_1>
```

We should also update our official docs and linked to it instead of linking to the datadog-secret-backend repo.

pkg/config/setup/config.go
```diff
@@ -2335,7 +2339,7 @@ func ResolveSecrets(config pkgconfigmodel.Config, secretResolver secrets.Compone
 		AuditFileMaxSize:       config.GetInt("secret_audit_file_max_size"),
 	})
 
-	if config.GetString("secret_backend_command") != "" {
+	if config.GetString("secret_backend_command") != "" || config.GetString("secret_backend_type") != "" {
```
hush-hush, line 2345: We should add a method `IsEnabled` or similar on the `secretResolver` to keep the logic inside the comp. This could be done in another PR.

releasenotes/notes/secret_backend_type-and-secret_backend_config-77479d1dc04353f5.yaml
```diff
@@ -0,0 +1,11 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+features:
+  - |
+    Adds `secret_backend_type` and `secret_backend_config` options to simplify secret retrieval. Instead of providing a custom executable script, you can now configure the Agent to use the built-in `secret-generic-connector` to fetch secrets from a specified backend.
```
hush-hush:
```suggestion
    [Preview] The agent can now connect to AWS SSM, AWS Secrets, Hashicorp Vault and Azure Keyless secret management solution to resolve secrets without requiring a user provided binary. For this, two new setting are introduce: `secret_backend_type` and `secret_backend_config`.
    For more information see: https://docs.datadoghq.com/agent/configuration/secrets-management
```

hush-hush (changes requested):
I added a few comments.

pkg/config/setup/config.go
rahulkaukuntla: Sounds good, I can do it in a later PR.

releasenotes/notes/secret_backend_type-and-secret_backend_config-77479d1dc04353f5.yaml
rahulkaukuntla: Changed.

pkg/config/config_template.yaml
rahulkaukuntla: Changed.

comp/core/secrets/secretsimpl/secrets.go
rahulkaukuntla: fixed.

comp/core/secrets/secretsimpl/secrets.go
rahulkaukuntla: Ah, I was not aware of this. I will just remove this code block.

comp/core/secrets/secretsimpl/secrets.go
rahulkaukuntla: We talked about this offline, and agreed to use warning statements instead of error statements to account for these cases.

comp/core/secrets/secretsimpl/fetch_secret.go
rahulkaukuntla: I don't remember writing this--anyways I've removed it.

comp/core/secrets/secretsimpl/secrets.go
rahulkaukuntla: [Added](https://github.com/DataDog/datadog-agent/pull/36804/commits/68d372c896102995bb8e8f6b623e74d38df04262#diff-d70bd6205b69b01a142fd6978341b5f12ddf01af899aaeda4948abbea79e4538R293)

comp/core/secrets/secretsimpl/secrets.go
rahulkaukuntla: [Done.](https://github.com/DataDog/datadog-agent/pull/36804/commits/68d372c896102995bb8e8f6b623e74d38df04262#diff-d70bd6205b69b01a142fd6978341b5f12ddf01af899aaeda4948abbea79e4538R792)

comp/core/secrets/secretsimpl/secrets.go
```diff
@@ -246,6 +330,35 @@ func (r *secretResolver) Configure(params secrets.ConfigParams) {
...
+func validateTypeAndConfig(backendType string, backendConfig map[string]interface{}) error {
+	allowed, ok := allowedKeysByBackend[backendType]
+	if !ok {
+		return fmt.Errorf("unsupported backend type: %s", backendType)
+	}
+
+	for key, val := range backendConfig {
+		allowedVal, ok := allowed[key]
+		if !ok {
+			return log.Warnf("unsupported key '%s' for backend '%s'", key, backendType)
```
hush-hush: We should return `nil` for a warning. Warnings are not errors here and they're going to be logged twice with line 288.

also, how sure are we that we listed all the possible key in `allowedKeysByBackend` ? Like for example I see that Vault has `aws_region`, would it be possible for a user to use `gcp_region` ? What if the libraries down the line like `vault` or `AWS` evolve to support new settings ?

Unless we're sure that we've listed any possible key we should not warn. Thinking more about this I not sure it's a good idea to validate `backendConfig` in the end.

pkg/config/config_template.yaml
```diff
@@ -865,10 +865,38 @@ api_key:
...
+## The section contains information required to resolved secrets.
+## For more information see: https://docs.datadoghq.com/agent/configuration/secrets-management
+#
+# secret_backend_config:
+#   <KEY_1>: <VALUE_1>
+#   <KEY_2>: <VALUE_2>
+#   <KEY_3>:
+#     <KEY_3_1>: <VALUE_3_1>
+#     <KEY_3_2>: <VALUE_3_2>
+
```
hush-hush: No need to list that many keys, let's keep the template as concise if possible.

pkg/config/config_template.yaml
```diff
@@ -865,10 +865,38 @@ api_key:
...
 ## @param secret_backend_command - string - optional
 ## @env DD_SECRET_BACKEND_COMMAND - string - optional
-## `secret_backend_command` is the path to the script to execute to fetch secrets.
+## `secret_backend_command` is the path to your custom script to execute to fetch secrets.
 ## The executable must have specific rights that differ on Windows and Linux.
+## DO NOT SET this option if you are using the default secret backend.
```
hush-hush: Let's be as precise as possible in the public facing documentation. It's a good rule to assume users don't know everything about the agent behavior (ex `default secret backend` is a blurry concept).
Also, we should try to explain why users should not use something. So instead of saying don't use something, we should explain what the consequences are.

```suggestion
##
## This option take precedence over `secret_backend_type`.
```

pkg/config/config_template.yaml
```diff
@@ -877,6 +905,7 @@ api_key:
 ## @param secret_backend_arguments - list of strings - optional
 ## @env DD_SECRET_BACKEND_ARGUMENTS - space separated list of strings - optional
 ## If secret_backend_command is set, specify here a list of arguments to give to the command at each run.
+## DO NOT SET this option if you are using the default secret backend.
```
hush-hush:
```suggestion
##
## This option take precedence over `secret_backend_type`.
```

hush-hush (changes requested):

jeremy-hanna (approved):
👍 for agent-runtime files

pkg/config/config_template.yaml
rahulkaukuntla: fixed.

pkg/config/config_template.yaml
rahulkaukuntla: looks good, I committed the suggestion.

pkg/config/config_template.yaml
rahulkaukuntla: also committed this suggestion.

comp/core/secrets/secretsimpl/secrets.go
rahulkaukuntla: You raise a valid point about there being other options that could be missing. I also think that there are two other reasons why we shouldn't be validating here: 1) we don't validate other config options to this level (at least until a schema is added to the config_template), and 2) we provide customers with good documentation, so we can trust them to provide the correct config values.

As such, I agree with you that this validation isn't needed. So I have removed the validation for backendConfig in [this commit](https://github.com/DataDog/datadog-agent/pull/36804/commits/ad85615268a99d3b2c0feb23f1592856c2b3a19e).

hush-hush (approved):

comp/core/secrets/secretsimpl/fetch_secret.go
```diff
@@ -109,6 +109,12 @@ func (r *secretResolver) fetchSecret(secretsHandle []string) (map[string]string,
 		"version": secrets.PayloadVersion,
```
hush-hush, line 109: We need to bump the version since we changed the protocol

hush-hush (changes requested):

comp/core/secrets/secretsimpl/fetch_secret.go
rahulkaukuntla: done.

hush-hush (approved):

PR #36805, opened 2025-05-08
APM: Keep Unix Socket File (APMS-15670)

ajgajg1134:
### What does this PR do?
Fix issue where Trace-Agent socket could be deleted during an Agent upgrade by the previous Trace-Agent as it was shutting down.

This also resolves a semi-related issue where the trace-agent wouldn't shutdown the http server if it was configured to _only_ use unix sockets or windows named pipes.

### Motivation
In Go by default a UnixListener will delete the socket file that was created by it when it is shutdown. This creates an issue where if multiple trace agent's are running simultaneously (like during an agent upgrade). It is possible that this deletes the socket file of the NEW trace agent, thereby preventing any new connections from being formed.

This results in any new connection attempts by any of the APM tracing libraries failing forever until the trace-agent is restarted.
Unfortunately this change is a bit tough to test in an automated way. I performed a manual test here where I configured the trace-agent to only use a unix socket. I then started instance A of the trace-agent, emitted some test trace data verifying it was received. Then started an additional trace-agent B, verifying again that trace data continues to flow. I then stopped Trace-Agent A and observed that the socket file continued to exist (Before this change the file would be deleted). I also added a unit test to verify the socket is not deleted by shutting down the receiver.

Leaving this file around shouldn't cause any issues as we check for it on startup and this file is located in /var/run by default.

### DIFF
--- a/releasenotes/notes/APMNoDeleteSocket-559ce568afc088bd.yaml
+++ b/releasenotes/notes/APMNoDeleteSocket-559ce568afc088bd.yaml
@@ -8,4 +8,4 @@
 ---
 fixes:
   - |
-    APM: Fix issue where Trace-Agent socket could be deleted during an Agent upgrade by the previous Trace-Agent as it was shutting down.
+    APM: Fix an issue where the Trace-Agent socket could be deleted during an Agent upgrade by the previous Trace-Agent during shutdown.
--- a/releasenotes/notes/APMNoDeleteSocket-559ce568afc088bd.yaml
+++ b/releasenotes/notes/APMNoDeleteSocket-559ce568afc088bd.yaml
@@ -1,0 +1,11 @@
+# Each section from every release note are combined when the
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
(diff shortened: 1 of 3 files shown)


releasenotes/notes/APMNoDeleteSocket-559ce568afc088bd.yaml
```diff
@@ -0,0 +1,11 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+fixes:
+  - |
+    APM: Fix issue where Trace-Agent socket could be deleted during an Agent upgrade by the previous Trace-Agent as it was shutting down.
```
drichards-87, line 11:
```suggestion
    APM: Fix an issue where the Trace-Agent socket could be deleted during an Agent upgrade by the previous Trace-Agent during shutdown.
```

drichards-87 (approved):

ichinaski (approved):

PR #36810, opened 2025-05-09
Fix notarization step for agent dmg

CelianR:
### What does this PR do?

The error handling was not always working properly due to the `-e` option of the script.
Better handled errors with `EXIT_CODE` that is reset and also updated on log error.

Example [job](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/929085996#L2174) where the retry wasn't done.

### DIFF
--- a/.gitlab/package_build/build_agent_dmg.sh
+++ b/.gitlab/package_build/build_agent_dmg.sh
@@ -128,7 +128,7 @@ if [ "$SIGN" = true ]; then
         EXIT_CODE=
         RESULT=$(xcrun notarytool submit --timeout "$NOTARIZATION_TIMEOUT" --apple-id "$APPLE_ACCOUNT" --team-id "$TEAM_ID" --password "$NOTARIZATION_PWD" "$LATEST_DMG" --wait || EXIT_CODE=$?)
         echo "Results: $RESULT"
-        SUBMISSION_ID=$(echo "$RESULT" | awk "$1 == \"id:\"{print $2; exit}")
+        SUBMISSION_ID=$(echo "$RESULT" | awk "\$1 == \"id:\"{print \$2; exit}")
         echo "Submission ID: $SUBMISSION_ID"
         echo "Submission logs:"
         xcrun notarytool log --apple-id "$APPLE_ACCOUNT" --team-id "$TEAM_ID" --password "$NOTARIZATION_PWD" "$SUBMISSION_ID" || true
--- a/.gitlab/package_build/build_agent_dmg.sh
+++ b/.gitlab/package_build/build_agent_dmg.sh
@@ -121,29 +121,19 @@ if [ "$SIGN" = true ]; then
 
     # Send package for notarization; retrieve REQUEST_UUID
     echo "Sending notarization request."
-
-    # Apply timeout / retry
-    for attempt in $(seq 1 $NOTARIZATION_ATTEMPTS); do
+    export NOTARIZATION_TIMEOUT
+    export LATEST_DMG
+    # shellcheck disable=SC2016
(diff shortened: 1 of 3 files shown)


.gitlab/package_build/build_agent_dmg.sh
```diff
@@ -124,19 +124,25 @@ if [ "$SIGN" = true ]; then
...
-    xcrun notarytool log --apple-id "$APPLE_ACCOUNT" --team-id "$TEAM_ID" --password "$NOTARIZATION_PWD" "$SUBMISSION_ID"
-    if [ -n "$EXIT_CODE" ]; then
-        echo "Notarization attempt #$attempt/$NOTARIZATION_ATTEMPTS failed, retrying in $NOTARIZATION_WAIT_TIME"
-        sleep "$NOTARIZATION_WAIT_TIME"
-    else
-        echo "Successfully notarized the package"
-        break
-    fi
+        EXIT_CODE=
+        RESULT=$(timeout "$NOTARIZATION_TIMEOUT" xcrun notarytool submit --apple-id "$APPLE_ACCOUNT" --team-id "$TEAM_ID" --password "$NOTARIZATION_PWD" "$LATEST_DMG" --wait) || EXIT_CODE=$?
```
alopezz, line 128: If I understand correctly the EXIT_CODE here will be overwritten by the exit code of `xcrun notarytool log` below.

- Is that fine because we expect the `log` command to fail if `submit` failed? In that case I guess we could just ignore the error rather than setting EXIT_CODE.
- Should we instead consider the attempt failed at this stage and wrap it in error handling code of its own? In that case it might make sense to split both steps (each having its own set of retries, perhaps using the code in https://github.com/DataDog/datadog-agent/blob/main/tools/ci/retry.sh).

I think the current behavior might be wrong and that the second option might actually be the only correct way to do this. The reason being that if the submission succeeds and `log` fails, it might not make sense to repeatedly `submit` again, but to just retry the `log` call.

Looking at the docs, I also wonder why we use the `timeout` command instead of passing the `--timeout` flag that this command [seems to have](https://keith.github.io/xcode-man-pages/notarytool.1.html#timeout).

.gitlab/package_build/build_agent_dmg.sh
```diff
@@ -121,23 +121,19 @@ if [ "$SIGN" = true ]; then
...
+    export LATEST_DMG
+    # shellcheck disable=SC2016
+    ./tools/ci/retry.sh -n "$NOTARIZATION_ATTEMPTS" bash -c '
+        EXIT_CODE=
+        RESULT=$(xcrun notarytool submit --timeout "$NOTARIZATION_TIMEOUT" --apple-id "$APPLE_ACCOUNT" --team-id "$TEAM_ID" --password "$NOTARIZATION_PWD" "$LATEST_DMG" --wait || EXIT_CODE=$?)
+        echo "Results: $RESULT"
+        SUBMISSION_ID=$(echo "$RESULT" | awk "\$1 == \"id:\"{print \$2; exit}")
+        echo "Submission ID: $SUBMISSION_ID"
+        echo "Submission logs:"
+        xcrun notarytool log --apple-id "$APPLE_ACCOUNT" --team-id "$TEAM_ID" --password "$NOTARIZATION_PWD" "$SUBMISSION_ID" || true
```
alopezz, line 134: Per our discussion on slack, I think this might be fine as is here except for this part where we do want to fail (i.e. the original purpose of this PR):

```suggestion
        xcrun notarytool log --apple-id "$APPLE_ACCOUNT" --team-id "$TEAM_ID" --password "$NOTARIZATION_PWD" "$SUBMISSION_ID"
```

alopezz (approved):

chouetz (approved):

PR #36811, opened 2025-05-09
Make sure all the tests are run on .gitlab-ci.yml changes

KevinFairise2:
### What does this PR do?

Make sure all the tests are run on buildimages update


tasks/gotest.py
```diff
@@ -44,7 +44,7 @@
 
 WINDOWS_MAX_PACKAGES_NUMBER = 150
 WINDOWS_MAX_CLI_LENGTH = 8000  # Windows has a max command line length of 8192 characters
-TRIGGER_ALL_TESTS_PATHS = ["tasks/gotest.py", "tasks/build_tags.py", ".gitlab/source_test/*"]
+TRIGGER_ALL_TESTS_PATHS = ["tasks/gotest.py", "tasks/build_tags.py", ".gitlab/source_test/*", ".gitlab-ci.yml"]
```
chouetz, line 47: I'm wondering if we should run all tests when we modify anything in the configuration, or on the contrary if we should run only on specific stuff, and then we should extract the buildimage versions (like we have for test-infra), to prevent tests when modification of a non-related rule, wdyt?

chouetz (approved):

tasks/gotest.py
KevinFairise2: I think it would make sense to extract the buildimages environement variables and trigger the all tests only on changes to that file.
If that's okay for you I will merge this one as is, so we make sure we run everything on buildimages bumps. And create a card for us to extract the variables to a separate file

PR #36817, opened 2025-05-09
[EBPF] kmt: allow retrying test jobs

gjulianm:
### What does this PR do?

This PR adds a task, used as `dda inv kmt.retry-failed-pipeline --pipeline-id <pipeline-id>`, that detects failed KMT test jobs in a given pipeline (also detects skipped ones due to failed dependencies) and retries all the setup/dependency jobs and the tests in the correct order.

It also adds a suggestion to use the command when the KMT test jobs fail.

### Motivation

Allow users to re-run KMT pipelines in case of infra failures instead of running an entire pipeline.

Manually ran the command against different failed pipelines.

### Additional Notes

In order to model properly the requirements of the jobs, some extra data structures have been added in `tasks/kernel_matrix_testing/ci.py`, including "sub-pipelines" which model all the upload/dependency jobs for a given arch-component tuple (e.g, all jobs for amd64 and system-probe).


paulcacheux (approved):

mbertrone (approved):

CelianR (approved):
Nice!

tasks/kmt.py
```diff
@@ -2371,3 +2375,70 @@ def flare(ctx: Context, dest_folder: Path | str | None = None, keep_uncompressed
...
+    info(f"[+] Retrieving jobs for pipeline {pipeline_id}")
+    kmt_pipeline = KMTPipeline(pipeline_id)
+    kmt_pipeline.retrieve_jobs()
+
+    jobs_to_retry: list[KMTTestRunJob] = []
+    failed_jobs = [j for j in kmt_pipeline.test_jobs if j.status == GitlabJobStatus.FAILED and not j.is_retried]
+    if len(failed_jobs) == 0:
+        info(f"[+] No failed tests found in pipeline {pipeline_id}. Checking dependency failures")
+
+        # Skipped tests can be skipped because of failed setup/dependency upload jobs, so check
```
chouetz, line 2392: Is it possible they are skipped because of a rule? In this case we cannot restart them, could it cause an issue?

tasks/kmt.py
```diff
@@ -2371,3 +2375,70 @@ def flare(ctx: Context, dest_folder: Path | str | None = None, keep_uncompressed
...
+        # those too
+        skipped_jobs = [j for j in kmt_pipeline.test_jobs if j.status == GitlabJobStatus.SKIPPED]
+        for test in skipped_jobs:
+            if test.has_failed_dependencies:
+                jobs_to_retry.append(test)
+    else:
+        jobs_to_retry = failed_jobs
```
chouetz, line 2399: You don't check for skipped jobs if you have failures? I don't know if it's possible to have both failed and skipped jobs.

tasks/kmt.py
```diff
@@ -2371,3 +2375,70 @@ def flare(ctx: Context, dest_folder: Path | str | None = None, keep_uncompressed
...
+
+    if component is not None:
+        jobs_to_retry = [j for j in jobs_to_retry if j.component == component]
+
+    if len(jobs_to_retry) == 0:
+        info(f"[+] No test jobs to retry in pipeline {pipeline_id}")
+        return
+
+    info(f"[+] Found the following candidates to retry in pipeline {pipeline_id}:")
```
chouetz, line 2408: 🥜 
```suggestion
    info(f"[+] Found {len(jobs_to_retry)} candidates to retry in pipeline {pipeline_id}(https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/{pipeline_id}):")
```
And delete the similar comment below

tasks/kmt.py
```diff
@@ -2371,3 +2375,70 @@ def flare(ctx: Context, dest_folder: Path | str | None = None, keep_uncompressed
...
+    for job in jobs_to_retry:
+        info(f"[+] {job.name} (status: {job.status})")
+
+        if job.kmt_subpipeline is None:
+            raise Exit(f"[-] Job {job.name} has no associated subpipeline")
+
+    info(
+        f"[+] We will retry {len(jobs_to_retry)} jobs in pipeline {pipeline_id} (https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/{pipeline_id})"
+    )
```
chouetz, line 2417:
```
```
See above

tasks/kernel_matrix_testing/ci.py
```diff
@@ -225,33 +249,242 @@ def get_test_results(self) -> dict[str, bool | None]:
...
+        for setup_job in self.setup_jobs:
+            setup_job.cleanup_job = self.last_cleanup_job
+
+        for dependency_upload_job in self.dependency_upload_jobs:
+            dependency_upload_job.setup_job = self.last_setup_job
+
+            for setup_job in self.setup_jobs:
+                setup_job.dependency_upload_jobs.append(dependency_upload_job)
+
+    def get_job(self, job_id: int | str) -> KMTJob | None:
```
chouetz, line 445: Could/should it be a `@property` as well?

tasks/libs/pipeline/tools.py
```diff
@@ -19,6 +20,55 @@
 PIPELINE_FINISH_TIMEOUT_SEC = 3600 * 5
 
 
+# Reference: https://docs.gitlab.com/api/graphql/reference/#cijobstatus
+class GitlabJobStatus(Enum):
+    CANCELED = "canceled"  # A job that is canceled.
```
chouetz, line 25:
```suggestion
    canceled = auto()
```
I think the comment is useless and you can get the string with `GitlabJobStatus.canceled.name`. But that might not simplify the code, you can probably ignore this one

chouetz (commented):
Few minor comments and 2 questions about skipped jobs

.gitlab/kernel_matrix_testing/security_agent.yml
```diff
@@ -159,6 +159,7 @@ kmt_run_secagent_tests_x64:
   after_script:
     - !reference [.collect_outcomes_kmt]
     - !reference [.upload_junit_kmt]
+    - !reference [.suggest_kmt_retry_command]
```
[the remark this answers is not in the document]
gjulianm: I wanted to later refactor all of these, because now all the after_script is the same for KMT jobs so we'll probably move this to the common definition.

tasks/kmt.py
gjulianm: We're relying here on the users to not retry pipelines that didn't run. But I think in that case if it doesn't work, they'll either get an error from Gitlab about the jobs not being retriable, or the initial jobs will fail due to missing dependencies.

tasks/kmt.py
gjulianm: The idea here was to avoid complicating the logic too much. The idea about checking skipped jobs is because of dependency failures, the assumption is that it's rare to have jobs where some dependencies failed in a subset of jobs (e.g., system-probe) and then tests failed in another. Doing it this way makes it easier to avoid retrying KMT jobs that didn't run because there were no related changes.

tasks/kernel_matrix_testing/ci.py
gjulianm: It requires an argument so it needs to be a function

chouetz: Ha yes my bad, thanks

PR #36818, opened 2025-05-09
service discovery: Fix race in TestPythonFromBashScript

vitkyrka:
### What does this PR do?

The logic for detecting dash vs bash has a race condition since the
forked process initially has the sh command line even in the bash case
so it's mistakenly identified as the dash case and waits for a
subprocess which never appears. Combine the two Eventually loops to fix
it.

### Motivation

https://datadoghq.atlassian.net/browse/DSCVR-132

Change in test.


smeso-ddg (approved):
Thanks!

PR #36826, opened 2025-05-09
Optimize `batch_strategy` to stream compress logs

gh123man:
### What does this PR do?

Prior to this PR, `batch_strategy` would buffer all uncompressed log messages until the buffer became full (or timed out) and then flush them to the compressor. This meant that for a brief period of time, we are holding all of the message and the compressed payload in memory at the same time. 

This PR changes `batch_strategy` to stream compress messages as they flow in and discard the uncompressed messages as soon as they are compressed. 

This change also removes the `LineSerializer` because it's not used in production code. 

### Motivation

Memory optimization. 

UTs cover most cases. 

Sanity check: Ensure log collection still works as expected.


pkg/logs/sender/stream_strategy.go
```diff
@@ -46,7 +46,7 @@ func (s *streamStrategy) Start() {
 			}
 
 			unencodedSize := len(msg.GetContent())
-			s.outputChan <- message.NewPayload([]*message.Message{msg}, encodedPayload, s.compression.ContentEncoding(), unencodedSize)
+			s.outputChan <- message.NewPayload([]*message.MessageMetadata{&msg.MessageMetadata}, encodedPayload, s.compression.ContentEncoding(), unencodedSize)
```
ddrthall, line 49: I'm reasonably confident this is a regression, the direct reference to msg's subcomponent should keep all of msg alive as long as the payload exists.

pkg/logs/sender/serializer.go
```diff
@@ -11,67 +11,47 @@ import (
...
-
-	for i, message := range messages {
-		if i > 0 {
-			if _, err := writer.Write([]byte{','}); err != nil {
-				return err
-			}
+func (s *arraySerializer) Serialize(message *message.Message, writer io.Writer) error {
+	if s.isFirstMessage {
+		if _, err := writer.Write([]byte{'['}); err != nil {
+			return err
```
ddrthall, line 40: I'm still downloading all of the relevant context into my brain, but at second pass I'm a bit apprehensive about how errors in this function are now handled by the callers - is it possible for us to send malformed payloads now when it wasn't before?

gh123man: I put some thought into this, as I had similar concerns. 
[Ive introduced `addMessage` ](https://github.com/DataDog/datadog-agent/blob/39d198a3e822ecb4c9823ac89a425ee13c704571/pkg/logs/sender/batch_strategy.go#L152)which should deal with all (except for `Finish`) error handling. 

The general flow is:
- do we have room to add a message?
- if yes, try to add it, 
  - if this function returns an error, don't add it. 
  - if this function does not return an error, add it to the batch buffer.

[`Finish` has it's own error handling. ](https://github.com/DataDog/datadog-agent/blob/39d198a3e822ecb4c9823ac89a425ee13c704571/pkg/logs/sender/batch_strategy.go#L189)

pkg/logs/sender/message_buffer.go
```diff
@@ -9,27 +9,34 @@ import (
...
 func NewMessageBuffer(batchSizeLimit int, contentSizeLimit int) *MessageBuffer {
 	return &MessageBuffer{
-		messageBuffer:    make([]*message.Message, 0, batchSizeLimit),
+		messageBuffer:    make([]*message.MessageMetadata, 0, batchSizeLimit),
 		contentSizeLimit: contentSizeLimit,
 	}
 }
 
+// CanAddMessage returns true if the message can be added to the buffer.
+func (p *MessageBuffer) CanAddMessage(message *message.Message) bool {
```
soberpeach, line 28: What is the purpose of moving the previous logic of `AddMessage` to `CanAddMessage`? Is it so `addMessage` can determine if it should serialize the message?

I'm just curious because it seems like `addMessage` will inadvertently call `CanAddMesage` twice.

pkg/logs/sender/serializer.go
ddrthall: As long as a trailing comma like ```[VALID_MESSAGE,]``` isn't explosive to ingest, I am sold.

pkg/logs/sender/message_buffer.go
gh123man: It so we can ensure a message is only considered added if it was successfully serialized. 
It's possible that:
- there is room in the buffer
- serialize fails

In this case, we would need to NOT add it to the buffer. This wasn't supported in the old API. Adding `CanAddMessage` allows us to check if it would fit before adding it to ensure this failure mode is handled.

pkg/logs/sender/serializer.go
gh123man: trailing `,` are not supported and the payload won't be parsed by the intake (I tested this manually). 
However I can't think of a possible case that this would happen (short of programmer error) in practice. Curious if there is a case I am missing.

soberpeach (approved):

soberpeach (approved):

pkg/logs/sender/serializer.go
ddrthall: It is possible for us to produce a payload structured like ```[,,,{...partial_json..,,,,{...valid_json...},]```, given the constraints we are guaranteed from the Serialize function - we will persist a comma even if we fail the content write, and flush won't reset the state unless a successful content write is detected. Similarly, the content write is not guaranteed atomic and can partially succeed before being marked an error. 

But despite the code supporting a number of ways that the payload could be corrupted, I don't necessarily expect it to actually be corrupted. If you want to make the call that writing to memory is inherently more stable than writing to disk and we wouldn't realistically see any kind of writer error to the underlying compressor instance then I don't think I would argue with that.

gh123man: After chatting offline, we decided to retain the existing error handling behavior and just drop on any encoding error.

ddrthall (approved):

PR #36834, opened 2025-05-09
[DJM] Fix DD_TRACE_DEBUG config for Databricks installer

aboitreaud:
### What does this PR do?

Fix the var name for the DD_TRACE_DEBUG config. Current type means we never add the required env var in the tracer env vars file.


rkaneriya (approved):
🙊

PR #36836, opened 2025-05-09
Rename ContainerStatus.AllocatedResources field to fix conflict

sblumenthal:
### What does this PR do?

This PR renames the `AllocatedResources` field in the ContainerStatus struct to `ResolvedAllocatedResources`

### Motivation

Within the agent, the `AllocatedResources` field represents data which we ourselves are parsing and attaching to the data after the fact, the name of the field does not matter for our purposes. Unfortunately, there is already an `AllocatedResources` field [exposed on the ContainerStatus](https://github.com/kubernetes/kubernetes/blob/1b4c5b3f8605dc1c86050ae941a24556d4fbd678/pkg/apis/core/types.go#L2886-L2891) which is reported by the kubelet, which is gated behind the `InPlacePodVerticalScaling` feature flag (and it looks like this flag is [enabled by default on k8s 1.33+](https://github.com/kubernetes/kubernetes/blob/660df229bf3929741cf31659187060d0c651dcf9/pkg/features/kube_features.go#L1385-L1388))

The end result is that unmarshalling data from the /pods endpoint fails with:
```
Expected nil, but got: &errors.AgentError{message:"couldn't fetch \"podlist\": unable to unmarshal podlist, invalid or null: kubelet.Pod.Status: kubelet.Status.Containers: []kubelet.ContainerStatus: kubelet.ContainerStatus.AllocatedResources: []kubelet.ContainerAllocatedResource: decode slice: expect [ or n, but found {, error found in #10 byte of ...|sources\":{\n         |..., bigger context ...|4\",\n                        \"allocatedResources\":{\n                            \"cpu\":\"42m\",\n        |...", errorReason:1}
```

### DIFF
--- a/pkg/proto/pbgo/core/workloadmeta.pb.go
+++ b/pkg/proto/pbgo/core/workloadmeta.pb.go
@@ -898,22 +898,22 @@ func (x *ContainerAllocatedResource) GetID() string {
 }
 
 type Container struct {
-	state              protoimpl.MessageState        `protogen:"open.v1"`
-	EntityId           *WorkloadmetaEntityId         `protobuf:"bytes,1,opt,name=entityId,proto3" json:"entityId,omitempty"`
-	EntityMeta         *EntityMeta                   `protobuf:"bytes,2,opt,name=entityMeta,proto3" json:"entityMeta,omitempty"`
-	EnvVars            map[string]string             `protobuf:"bytes,3,rep,name=envVars,proto3" json:"envVars,omitempty" protobuf_key:"bytes,1,opt,name=key" protobuf_val:"bytes,2,opt,name=value"`
-	Hostname           string                        `protobuf:"bytes,4,opt,name=hostname,proto3" json:"hostname,omitempty"`
-	Image              *ContainerImage               `protobuf:"bytes,5,opt,name=image,proto3" json:"image,omitempty"`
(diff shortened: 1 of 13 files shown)


pkg/proto/pbgo/core/workloadmeta.pb.go
pgimalac, line 1: Could you update the codeowners for this file to Container Platform ? It is currently owned by Agent Runtimes but we don't own the proto file, I don't think that really makes sense 😄

pgimalac (approved):

rtrieu (approved):

pkg/proto/pbgo/core/workloadmeta.pb.go
sblumenthal: cc @wdhif thoughts here?

gjulianm (approved):
LGTM, thanks for fixing it!

justin-lesko (approved):

fisherevans (approved):

L3n41c (approved):

jose-manuel-almaza (approved):

gabedos (approved):
Rename looks good to me.

pkg/proto/pbgo/core/workloadmeta.pb.go
wdhif: Hey, yes I agree 👍 @gabedos will do a PR for that, thanks for the heads up!

PR #36841, opened 2025-05-09
[NDINT-290] Add Versa device hardware metrics

ken-schneider:
### What does this PR do?
For Versa specifically, this adds device hardware metrics around CPU, memory, and disk usage.

Additionally, this PR creates a generic sender interface for NDM integrations that is intended to be used by multiple NDM integrations like Versa, Cisco SD-WAN, and future NDM corechecks. Most of the code for the sender has been copied from the existing Cisco SD-WAN sender and modified.

### Motivation
Increase support for Versa in NDM, increase code re-use

Tested locally against our own Versa cluster pointed at staging.

In attempting to create a generic NDM sender interface, I've made some assumptions about what functionality will be used in future integrations.


pkg/networkdevice/sender/sender.go
```diff
@@ -0,0 +1,130 @@
...
+import (
+	"maps"
+	"time"
+
+	"github.com/DataDog/datadog-agent/pkg/aggregator/sender"
+	"github.com/DataDog/datadog-agent/pkg/networkdevice/utils"
+	"github.com/DataDog/datadog-agent/pkg/util/log"
+)
+
+const timestampExpiration = 6 * time.Hour
```
zoedt, line 22: just curious if there's docs / reason for this timestamp expiration for stuff we send? just to make sure i follow!

pkg/networkdevice/sender/sender.go
```diff
@@ -0,0 +1,130 @@
...
+type Sender interface {
+	sender.Sender
+	// GaugeWithTimestampWrapper wraps sender GaugeWithTimestamp with error handling
+	GaugeWithTimestampWrapper(name string, value float64, tags []string, ts float64)
+	// CountWithTimestamp wraps sender CountWithTimestamp with error handling
+	CountWithTimestampWrapper(name string, value float64, tags []string, ts float64)
+	// UpdateTimestamps updates the last sent timestamps
+	UpdateTimestamps(newTimestamps map[string]float64)
+	// SetDeviceTags sets the device tags map
+	SetDeviceTags(deviceTags map[string][]string)
```
zoedt, line 39: [nit / small discussion topic]
would it be useful to rename to `SetDeviceTagsMap`? the `GetDeviceTags` only gets one entry vs. i'm not sure how confusing it is ultimately but i originally expected this to only set the device tags for one device but this is a subjective thought from my end!! the docstrings help ultimately when i went to investigate my assumptions

pkg/networkdevice/sender/sender.go
ken-schneider: I don't think I'm actually using that right now, but the intent is to avoid holding on to data about metrics we send 6 hours ago. Basically if we have a metrics that was updated 6 hours ago (for metrics where the endpoint has a timestamp for last updated at) we should send it again even if there's no update. 6 hours is consistent with Cisco SD-WAN

pkg/networkdevice/sender/sender.go
ken-schneider: Agreed, updated the function signature

zoedt (approved):

pkg/networkdevice/sender/sender.go
TCheruy: I was afraid the `lastTimeSent` map would grow indefinitely if we don't expire at some point

TCheruy (approved):

PR #36848, opened 2025-05-10
fix `go-deps.graph` support for iot-agent

paulcacheux:
### What does this PR do?

Currently when you run
```
inv -e go-deps.graph -o linux -b iot-agent
```
you get
```
Warning: unrecognized build type, no build tags included.
```

This PR fixes this by ensuring the `build` field is correctly propagated, and the entrypoint is correct for the iot-agent.

### DIFF
--- a/tasks/diff.py
+++ b/tasks/diff.py
@@ -27,7 +27,7 @@ BINARIES: dict[str, dict] = {
     },
     "iot-agent": {
         "build": "agent",
-        "entrypoint": "cmd/agent",
+        "entrypoint": "cmd/iot-agent",
         "flavor": AgentFlavor.iot,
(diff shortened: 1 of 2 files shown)


pgimalac:
@paulcacheux iot is a flavor, so it should be `-b agent -f iot` if I remember correctly the arguments
I can check this in 10min

paulcacheux:
I tried that, but it won't pick up the correct entrypoint doing that (even fixing the `diff.py` file)

pgimalac (approved):
Ideally I'd rather use it as `-b agent -f iot` but this seems legit, thanks for fixing !

GustavoCaso (approved):
LGTM

PR #36851, opened 2025-05-11
Remove limit for max_conns_per_message

guyarb:
### What does this PR do?

Remove hard-coded limit for max-messages-per-payload.

### Motivation

We have a default value already, if anyone decides to override it, we should not prevent it.


guyarb:
Rejected by CNM

PR #36852, opened 2025-05-11
[USMON-1440] usm: Introduce generic encoders list

guyarb:
### What does this PR do?

The PR replaces hard-coded usm encoders in the marshaling logic of CNM.

The PR allows us to have a different set of encoders per os (HTTP for windows, HTTP, HTTP/2, Kafka, Postgres and Redis for Linux).

### Motivation

* Removing code duplication
* Preventing from non-accessible code to be loaded in unsupported OS
* Simplifying protocol management


pkg/network/encoding/marshal/usm_encoders_linux.go
```diff
@@ -0,0 +1,34 @@
...
+package marshal
+
+import (
+	"github.com/DataDog/datadog-agent/pkg/network"
+)
+
+func initializeUSMEncoders(conns *network.Connections) []usmEncoder {
+	encoders := make([]usmEncoder, 0)
+
+	if encoder := newHTTPEncoder(conns.HTTP); encoder != nil {
```
DanielLavie, line 17: Should it be configuration-aware? In other words, should we only initialize an encoder if the corresponding protocol is enabled in the configuration?

DanielLavie: One possible optimization is to filter the list upfront to include only the enabled protocols. However, the performance gain seems minimal, so we can defer it for now

guyarb: I agree with the later, as all encoders check the payload they are getting are not empty. So the check is minimal.
Since we're talking on a very limited number of encoders, the preconfigured list seems like a very small gain.

pkg/network/encoding/marshal/usm_encoders_windows.go
```diff
@@ -0,0 +1,22 @@
...
+package marshal
+
+import (
+	"github.com/DataDog/datadog-agent/pkg/network"
+)
+
+func initializeUSMEncoders(conns *network.Connections) []usmEncoder {
+	encoders := make([]usmEncoder, 0, 1)
+
+	if encoder := newHTTPEncoder(conns.HTTP); encoder != nil {
```
clarkb7, line 17: why is only HTTP added for Windows? I don't see any platform tags/cases in the original code.

guyarb: The windows version of USM supports only the http protocol
all other protocols are linux only

guyarb: > I don't see any platform tags/cases in the original code.

That's something I'm planning to change, this is one of the preliminary PRs to allow that :)

pkg/network/encoding/encoding_linux_test.go
```diff
@@ -0,0 +1,304 @@
...
+
+	model "github.com/DataDog/agent-payload/v5/process"
+
+	"github.com/DataDog/datadog-agent/pkg/network"
+	"github.com/DataDog/datadog-agent/pkg/network/encoding/marshal"
+	"github.com/DataDog/datadog-agent/pkg/network/encoding/unmarshal"
+	"github.com/DataDog/datadog-agent/pkg/network/protocols"
+	"github.com/DataDog/datadog-agent/pkg/network/protocols/http"
+	"github.com/DataDog/datadog-agent/pkg/network/protocols/kafka"
+	"github.com/DataDog/datadog-agent/pkg/process/util"
```
hmahmood, line 26: nit: these don't follow the guidelines here: https://github.com/DataDog/datadog-agent/blob/main/docs/dev/imports.md

hmahmood (approved):

pkg/network/encoding/encoding_linux_test.go
guyarb: you mean the extra new line on line 17?

hmahmood: Yes, should be 
```go
	"fmt"
	"testing"

	model "github.com/DataDog/agent-payload/v5/process"
	"github.com/golang/protobuf/proto"
	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/require"

	"github.com/DataDog/datadog-agent/pkg/network"
	"github.com/DataDog/datadog-agent/pkg/network/encoding/marshal"
	"github.com/DataDog/datadog-agent/pkg/network/encoding/unmarshal"
	"github.com/DataDog/datadog-agent/pkg/network/protocols"
	"github.com/DataDog/datadog-agent/pkg/network/protocols/http"
	"github.com/DataDog/datadog-agent/pkg/network/protocols/kafka"
	"github.com/DataDog/datadog-agent/pkg/process/util"

```

clarkb7 (approved):

pkg/network/encoding/encoding_linux_test.go
```diff
@@ -0,0 +1,304 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
```
amitslavin, line 4: fix the date

amitslavin (approved):

PR #36862, opened 2025-05-12
omnibus: cache python integrations build

chouquette:
### What does this PR do?

Include python integrations into the omnibus build cache

### Motivation

Shorten the build duration (around 1:30min) now that integrations-core version is pinned


dkirov-dd (approved):
lgtm 👍

alopezz (approved):

PR #36868, opened 2025-05-12
Move api version telemetery AFTER version check

guymor4:
### What does this PR do?
This PR moves the tracking of `datadog.system_probe.usm.kafka.classified_hits` to only report api versions that are supported in classification, to avoid spamming false positives.

### Motivation
A lot of false positives
image


pkg/network/ebpf/c/protocols/kafka/kafka-parsing.h
guyarb, line 1: The change is creating a regression, as `is_supported_kafka_api_version` checks the version is supported for DECODING, you'll report only versions supported by the decoder, and not by the classifier

PR #36875, opened 2025-05-12
chore(fleet): Add APM SSI status to inventoryagent payload

BaptisteFoy:
### What does this PR do?
Reports Linux host SSI in the agent's metadata payload. Windows will be done in another PR

### Motivation
[INPLAT-568]

Manual QA on a VM -- checked the result using the metadata payload in debug logs

### Additional Notes

[INPLAT-568]: https://datadoghq.atlassian.net/browse/INPLAT-568?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


coignetp (approved):

Hitsuji-M (approved):

comp/metadata/inventoryagent/inventoryagentimpl/inventoryagent.go
```diff
@@ -374,6 +375,13 @@ func (ia *inventoryagent) fetchECSFargateAgentMetadata() {
 
 func (ia *inventoryagent) fetchFleetMetadata() {
 	ia.data["config_id"] = ia.conf.GetString("config_id")
+
+	// APM host-based auto injection (SSI)
+	autoInstrumentationEnabled, err := ssi.IsAutoInstrumentationEnabled()
+	if err != nil {
+		ia.log.Warnf("could not check if APM auto-instrumentation is enabled: %s", err)
+	}
+	ia.data["feature_auto_instrumentation_enabled"] = autoInstrumentationEnabled
```
hush-hush, line 384: This make every binary/project that use the inventoryagent payload depends on the SSI package.
Could you use the `Set` method of the `inventoryagent` comp instead.

Also, please update the README for `inventoryagent` to document the new field so we have a trace/doc somewhere.

hush-hush (changes requested):

betterengineering (approved):
Thanks for doing this, and ahead of schedule @BaptisteFoy 🙌 . This looks good, but it may be worth removing some of the additional Docker states if we don't need them.

pkg/fleet/installer/packages/ssi/status.go
```diff
@@ -0,0 +1,14 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+// Package ssi adds methods to check the status of the APM Single Step Instrumentation
+package ssi
+
+// APMInstrumentationStatus contains the instrumentation status of APM Single Step Instrumentation.
+type APMInstrumentationStatus struct {
```
bmermet, line 10: For windows will it be possible to extend this with other fields we might need ? Like IISInstrumented for instance.

pkg/fleet/installer/packages/ssi/status_unix.go
```diff
@@ -0,0 +1,88 @@
...
+func IsAutoInstrumentationEnabled() (bool, error) {
+	injectorInstalled, err := isInjectorPkgInstalled()
+	if err != nil {
+		return false, fmt.Errorf("could not check if injector package is installed: %w", err)
+	}
+	injectionStatus, err := GetInstrumentationStatus()
+	if err != nil {
+		return false, fmt.Errorf("could not get APM injection status: %w", err)
+	}
+	return injectorInstalled && (injectionStatus.HostInstrumented || (injectionStatus.DockerInstrumented && injectionStatus.DockerInstalled)), nil
```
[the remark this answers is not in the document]
BaptisteFoy: For the command `datadog-installer status` we need both, to not display "docker not instrumented" when docker is not installed

pkg/fleet/installer/packages/ssi/status.go
BaptisteFoy: I added a new field that reports the instrumentation method(s). Today it's "host" and/or "docker", we'll be able to add IIS in the future

comp/updater/ssistatus/impl/ssistatus.go
```diff
@@ -0,0 +1,96 @@
...
+			case <-ticker.C:
+				ctx := context.Background()
+				ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
+				defer cancel()
+				// APM host-based auto injection (SSI)
+				autoInstrumentationEnabled, err := ssi.IsAutoInstrumentationEnabled(ctx)
+				if err != nil {
+					c.log.Warnf("could not check if APM auto-instrumentation is enabled: %s", err)
+				}
+				c.inventoryAgent.Set("feature_auto_instrumentation_enabled", autoInstrumentationEnabled)
```
hush-hush, line 69: Please update https://github.com/DataDog/datadog-agent/blob/main/comp/metadata/inventoryagent/README.md. Even if we set it in a different package we're trying to have 1 place documenting the entire payload.

You will also need to update the REDAPL schema: https://github.com/DataDog/infrastructure-resources/blob/prod/generated/yaml/datadog_agent.yaml

comp/updater/ssistatus/impl/ssistatus.go
```diff
@@ -0,0 +1,96 @@
...
+					c.log.Warnf("could not get APM auto-instrumentation status: %s", err)
+				}
+				instrumentationModes := []string{}
+				if instrumentationStatus.HostInstrumented {
+					instrumentationModes = append(instrumentationModes, "host")
+				}
+				if instrumentationStatus.DockerInstalled && instrumentationStatus.DockerInstrumented {
+					instrumentationModes = append(instrumentationModes, "docker")
+				}
+				c.inventoryAgent.Set("auto_instrumentation_modes", instrumentationModes)
```
hush-hush, line 84: Same for this one.

comp/updater/ssistatus/mock/mock.go
```diff
@@ -0,0 +1,20 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
```
hush-hush, line 1: Since your comp has no public method you might not need a mock.

comp/updater/ssistatus/fx/fx.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/DataDog/datadog-agent/pkg/util/fxutil"
+)
+
+// Module defines the fx options for this component
+func Module() fxutil.Module {
+	return fxutil.Component(
+		fxutil.ProvideComponentConstructor(
+			ssistatusimpl.NewComponent,
+		),
+		fxutil.ProvideOptional[ssistatus.Component](),
```
hush-hush, line 21: Since your comp has no public method nobody will depend on it and force its inclusion at init time.

You can force it's inclusion by adding a similar line: https://github.com/DataDog/datadog-agent/blob/main/comp/core/configsync/configsyncimpl/module.go#L50

hush-hush (approved):
Added one comment about the README, but looks great otherwise. Feel free to merge after.

hush-hush (approved):

comp/updater/installerexec/impl/installerexec.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// NewComponent creates a new installerexec component
+func NewComponent(_ Requires) (Provides, error) {
+	agentPath, err := os.Executable()
+	if err != nil {
+		return Provides{}, fmt.Errorf("could not retrieve datadog-agent path: %w", err)
+	}
+
+	installerPath, err := filepath.Rel(agentPath, "/../../embedded/bin/installer")
```
GustavoCaso, line 34: Should we move this to `defaultpaths` package? It seems that we have similar path and logic already present on that package

comp/updater/ssistatus/def/component.go
```diff
@@ -0,0 +1,12 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-present Datadog, Inc.
+
+// Package ssistatus is the ssistatus component.
```
GustavoCaso, line 6: We could probably provide a bit more information on what this package does. What is the ssistatus? 😄

comp/updater/installerexec/def/component.go
```diff
@@ -0,0 +1,16 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-present Datadog, Inc.
+
+// Package installerexec provides a component to execute installer commands
+package installerexec
+
+import "github.com/DataDog/datadog-agent/pkg/fleet/installer"
```
GustavoCaso, line 9: This import is causing to import +60 packages on the agent. 

image

Since this is just an interface, and by the looks of it, quite clean of dependencies 

```
// Installer is a package manager that installs and uninstalls packages.
type Installer interface {
	IsInstalled(ctx context.Context, pkg string) (bool, error)

	AvailableDiskSpace() (uint64, error)
	State(ctx context.Context, pkg string) (repository.State, error)
	States(ctx context.Context) (map[string]repository.State, error)
	ConfigState(ctx context.Context, pkg string) (repository.State, error)
	ConfigStates(ctx context.Context) (map[string]repository.State, error)

	Install(ctx context.Context, url string, args []string) error
	ForceInstall(ctx context.Context, url string, args []string) error
	SetupInstaller(ctx context.Context, path string) error
	Remove(ctx context.Context, pkg string) error
	Purge(ctx context.Context)

	InstallExperiment(ctx context.Context, url string) error
	RemoveExperiment(ctx context.Context, pkg string) error
	PromoteExperiment(ctx context.Context, pkg string) error

	InstallConfigExperiment(ctx context.Context, pkg string, version string, rawConfig []byte) error
	RemoveConfigExperiment(ctx context.Context, pkg string) error
	PromoteConfigExperiment(ctx context.Context, pkg string) error

	GarbageCollect(ctx context.Context) error

	InstrumentAPMInjector(ctx context.Context, method string) error
	UninstrumentAPMInjector(ctx context.Context, method string) error

	Close() error
}
```

I would suggest reorganizing the `github.com/DataDog/datadog-agent/pkg/fleet/installer` package so that when importing the interface, we minimize the number of imports. We can follow the pattern we use on components, splitting the interface from the implementation. 

Let me know if you need any help, I can provide.

GustavoCaso (changes requested):
The code looks great, we just need to ensure we do not import all of those packages through the `installer` interface.

comp/updater/installerexec/impl/installerexec.go
BaptisteFoy: Ah good catch I didn't know this package existed. Thanks!

comp/updater/installerexec/def/component.go
BaptisteFoy: Moving the interface reduces it to "only" 13 at most 👍.
Though it's a bit confusing to me why this PR would impact the serverless performance (see [job](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/947849356))

GustavoCaso: That is great. 
7 of those imports are the new components
+github.com/DataDog/datadog-agent/comp/updater/installerexec/def
+github.com/DataDog/datadog-agent/comp/updater/installerexec/fx
+github.com/DataDog/datadog-agent/comp/updater/installerexec/impl
+github.com/DataDog/datadog-agent/comp/updater/ssistatus/def
+github.com/DataDog/datadog-agent/comp/updater/ssistatus/fx
+github.com/DataDog/datadog-agent/comp/updater/ssistatus/impl
+github.com/DataDog/datadog-agent/pkg/fleet/installer/packages/ssi

The new type package you created 
+github.com/DataDog/datadog-agent/pkg/fleet/installer/types

I think these ones can be removed 
+github.com/DataDog/datadog-agent/pkg/fleet/installer/env
+github.com/DataDog/datadog-agent/pkg/fleet/installer/errors
+github.com/DataDog/datadog-agent/pkg/fleet/installer/exec
+github.com/DataDog/datadog-agent/pkg/fleet/installer/paths
+github.com/DataDog/datadog-agent/pkg/fleet/installer/repository

Looks like we are importing the `github.com/DataDog/datadog-agent/pkg/fleet/installer/repository` package in the interface type to use the `State` struct 

```
type State struct {
	Stable     string `json:"stable"`
	Experiment string `json:"experiment"`
}
```

Seeing that it is a simple structure, we can try to validate what it would look like if we move into the `types` package. 

What do you think?

GustavoCaso: Thanks for giving it a try 🤜 🤛

GustavoCaso (approved):
Great work. Thanks for taking into account my feedback

PR #36879, opened 2025-05-12
[AGTMETRICS-234]Update sanitizer when API keys change

StephenWakely:
### What does this PR do?

Ensures the api key replacer replaces the API key after an API key has been refreshed.

### Motivation

After the API is updated, the serializer needs to update it's replacers so that it can replace the new API keys with a placeholder.

Previously the replacers were sorting the API keys and restoring based on position. I have removed the sort, since the order would change if the API key changed, thus restoring to the wrong API key. We now rely only on the position the key is in the configuration.

## Caveats

- An additional complexity is that API keys can be deduped. Theoretically you can have a configuration like this:

```
api_key: ENC[mainkee]
additional_endpoints:
  "http://localhost:8080":
  - ENC[kee1]
```

`mainkee` and `kee1` could return the same API key. The forwarders dedup this to ensure only one transaction is sent to any given endpoint/api key pair. At runtime, the keys could be changed so they are no longer the same. If this happens, the number of API keys will change and thus be in different positions. So we need to make sure we store the positions prior to deduping.

- It is possible that both `mainkee` and `kee1` start as different keys, transactions are created and stored to disk. Then these keys could be updated to be the same key. Transactions intended for `mainkee` and `kee1` would contain the same API key now and thus duplicate keys could be sent to the endpoint.

- Whilst the agent is stopped, the config could be changed such that the order of the API keys are altered. On starting the agent, the agent would restore these transactions and send them to the wrong API key. I don't think there is any way to fix this... 

I have added some unit tests. 

Have tested manually with this config:

```
dd_url: http://localhost:8080
api_key: ENC[mainkee]

forwarder_storage_max_size_in_bytes: 50000000
forwarder_retry_queue_payloads_max_size: 10

additional_endpoints:
  "http://localhost:8080":
  - ENC[kee1]
``` 

Updating the secrets, then check the files created in `/opt/datadog-agent/run/transactions_to_retry/core/**/*` to ensure they do not contain an API key in them.


FlorentClarret:
Thanks for the fix! Can we add a changelog for this one?

hush-hush (approved):

comp/forwarder/defaultforwarder/internal/retry/http_transactions_serializer.go
```diff
@@ -248,19 +291,38 @@ func toTransactionDestinationProto(destination transaction.Destination) (Transac
 	}
 }
 
-func createReplacers(apiKeys []string) (*strings.Replacer, *strings.Replacer) {
+// createReplacersV1 creates replacers for transactions created with V1 of the serializer.
+// This version sorted the keys prior to serializing. The location in this list is used to
+// unscrub the key.
+func createReplacerV1(apiKeys []string) *strings.Replacer {
```
vickenty, line 297: Would it make sense to add telemetry to know when we can drop this method?

comp/forwarder/defaultforwarder/internal/retry/http_transactions_serializer_test.go
```diff
@@ -112,6 +115,65 @@ func TestHTTPTransactionSerializerMissingAPIKey(t *testing.T) {
 	r.Equal(1, errorCount)
 }
 
+func TestHTTPTransactionSerializerUpdateAPIKey(t *testing.T) {
```
vickenty, line 118: Would it make sense to also test this scenario?

- Transaction is created with the old key. 
- Keys are updated.
- Transaction is being serialized.

comp/forwarder/defaultforwarder/internal/retry/http_transactions_serializer.go
```diff
@@ -101,6 +116,21 @@ func (s *HTTPTransactionsSerializer) Add(transaction *transaction.HTTPTransactio
...
+	if s.resolver.GetAPIKeyVersion() != s.currentKeyVersion {
+		// API keys have been updated so we need to rebuild.
+		s.placeholderMutex.Lock()
+		defer s.placeholderMutex.Unlock()
+
+		keys, version := s.resolver.GetAPIKeysInfo()
+		s.apiKeyToPlaceholder, s.placeholderToAPIKey = createReplacers(keys)
+		s.currentKeyVersion = version
+	}
+}
```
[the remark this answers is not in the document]
StephenWakely: My thinking is that because the initial `GetAPIKeyVersion()` call is outside the `s.placeholderMutex.Lock()`, between getting the keys and version another thread could come in, find a newer `APIKeyVersion` and `Keys` and update the replacers. We then acquire `s.placeholderMutex` and then update the replacers to an older version.

To fix, I would have to wrap the whole function in `s.placeholderMutex`, but I don't think that would be desirable since most of the time the version will not be updated.

The way it currently is, I believe, we could potentially have multiple threads call `createReplacers` unnecessarily, but at least each time they would get the latest version. 

Does that sound right to you?

comp/forwarder/defaultforwarder/internal/retry/http_transactions_serializer_test.go
StephenWakely: Urg... good thinking.

StephenWakely: I actually need to maintain a history of replacers, so we catch any outdated API keys. Performance is a concern, but hopefully API keys are updated rarely enough that it shouldn't be an issue..

vickenty: Keeping a history of API keys works, and I could be missing something, but I think it would be better to not rely on string replacement at all: keep the api key identity (index) in HTTPTransaction as it is created, serialize it in a dedicated `HTTPTransactionProto` field and re-set the correct key when reading. Support for API in the query string seems to have been removed from the code, so all that remains for the serializer to do is remove the api key header header.

vickenty (approved):

releasenotes/notes/refreshed-api-key-sanitizer-0bb1588c65bd9b18.yaml
```diff
@@ -0,0 +1,15 @@
...
+#
+# Each section note must be formatted as reStructuredText.
+---
+fixes:
+  - |
+    If a metric transaction can't be sent to the endpoint for whatever
+    reason, this transaction can be serialized to disk. When this happens
+    the API key needs to be sanitised. This ensures that when an API key
+    that comes from a secret is refreshed the replacer will continue
+    to sanitise the new API key.
```
drichards-87, line 15:
```suggestion
    If a metric transaction can't be sent to the endpoint, the transaction can be serialized to disk. When this occurs,
    the API key must be sanitized. This ensures that when an API key sourced from a secret is refreshed, the replacer continues to sanitize the new key.
```

drichards-87 (approved):

PR #36882, opened 2025-05-12
Fix race condition when running tests in disk check due to global variables

jose-manuel-almaza:
### What does this PR do?
Eliminate several global variables and use the .WithX() pattern for non-deterministic methods to be mocked in our unit tests.
Dependency injection through the constructor is not possible in this case due to the impossibility of modifying the constructor, which is called in different places in our code, and would entail more changes.
We opted to apply this pattern instead of making the members to be mocked public to maintain the encapsulation of our code.
By default, the values ​​for these methods will be those used in production.
We also adjusted the clock increment to the default timeout (5s), which will trigger the timeout.

### Motivation
Fix a race condition in some tests, due to the use of shared global variables.

### DIFF
--- a/pkg/collector/corechecks/system/disk/diskv2/disk.go
+++ b/pkg/collector/corechecks/system/disk/diskv2/disk.go
@@ -633,30 +633,6 @@ func (c *Check) fetchAllDeviceLabels() error {
 	return c.fetchAllDeviceLabelsFromBlkid()
 }
 
-// WithClock sets a custom clock on the Check and returns the updated Check.
-func (c *Check) WithClock(clock clock.Clock) *Check {
-	c.clock = clock
-	return c
-}
-
-// WithDiskPartitions sets a diskPartitions call on the Check and returns the updated Check.
-func (c *Check) WithDiskPartitions(f func(bool) ([]gopsutil_disk.PartitionStat, error)) *Check {
-	c.diskPartitions = f
-	return c
-}
-
-// WithDiskUsage sets a diskUsage call on the Check and returns the updated Check.
-func (c *Check) WithDiskUsage(f func(string) (*gopsutil_disk.UsageStat, error)) *Check {
-	c.diskUsage = f
-	return c
-}
-
-// WithDiskIOCounters sets a diskIOCounters call on the Check and returns the updated Check.
-func (c *Check) WithDiskIOCounters(f func(...string) (map[string]gopsutil_disk.IOCountersStat, error)) *Check {
-	c.diskIOCounters = f
-	return c
-}
-
 // Factory creates a new check factory
 func Factory() option.Option[func() check.Check] {
 	return option.New(newCheck)
--- a/pkg/collector/corechecks/system/disk/diskv2/disk.go
+++ b/pkg/collector/corechecks/system/disk/diskv2/disk.go
@@ -35,10 +35,6 @@ const (
 )
 
 var (
-	// DiskPartitions returns a list of mounted disk partitions using gopsutil.
-	DiskPartitions = gopsutil_disk.Partitions
-	// DiskUsage returns disk usage statistics for the given path using gopsutil.
-	DiskUsage = gopsutil_disk.Usage
 	// DiskIOCounters returns disk I/O statistics for the specified devices using gopsutil.
 	DiskIOCounters = gopsutil_disk.IOCounters
 )
@@ -119,7 +115,10 @@ func compileRegExp(expr string, ignoreCase bool) (*regexp.Regexp, error) {
 // Check represents the Disk check that will be periodically executed via the Run() function
 type Check struct {
 	core.CheckBase
-	clock               clock.Clock
+	clock          clock.Clock
+	diskPartitions func(bool) ([]gopsutil_disk.PartitionStat, error)
+	diskUsage      func(string) (*gopsutil_disk.UsageStat, error)
+
 	initConfig          diskInitConfig
 	instanceConfig      diskInstanceConfig
 	includedDevices     []regexp.Regexp
@@ -134,6 +133,7 @@ type Check struct {
 
 // Run executes the check
 func (c *Check) Run() error {
+	log.Debug("weeeeeee")
 	sender, err := c.GetSender()
 	if err != nil {
 		return err
@@ -384,7 +384,7 @@ func (c *Check) configureIncludeMountPoint() error {
 }
 
 func (c *Check) collectPartitionMetrics(sender sender.Sender) error {
-	partitions, err := DiskPartitions(c.instanceConfig.IncludeAllDevices)
+	partitions, err := c.diskPartitions(c.instanceConfig.IncludeAllDevices)
 	if err != nil {
 		log.Warnf("Unable to get disk partitions: %s", err)
 		return err
@@ -466,13 +466,11 @@ func (c *Check) getDiskUsageWithTimeout(mountpoint string) (*gopsutil_disk.Usage
 	// Start the disk usage call in a separate goroutine.
 	go func() {
 		// UsageWithContext in gopsutil ignores the context for now (PR opened: https://github.com/shirou/gopsutil/pull/1837)
-		usage, err := DiskUsage(mountpoint)
+		usage, err := c.diskUsage(mountpoint)
 		// Use select to avoid writing to resultCh if timeout already occurred.
 		select {
 		case resultCh <- usageResult{usage, err}:
 		case <-timeoutCh:
-			// timeout occurred — avoid writing to resultCh
-			return
 		}
 	}()
 	// Use select to wait for either the disk usage result or a timeout.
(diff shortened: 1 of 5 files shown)


pkg/collector/corechecks/system/disk/diskv2/disk_nix_test.go
```diff
@@ -80,16 +80,15 @@ func TestGivenADiskCheckWithDefaultConfig_WhenCheckRuns_ThenAllUsageMetricsAreRe
 
 func TestGivenADiskCheckWithLowercaseDeviceTagConfigured_WhenCheckRuns_ThenLowercaseDevicesAreReported(t *testing.T) {
 	setupDefaultMocks()
-	diskv2.DiskPartitions = func(_ bool) ([]gopsutil_disk.PartitionStat, error) {
+	diskCheck := createCheck().(*diskv2.Check).WithDiskPartitions(func(_ bool) ([]gopsutil_disk.PartitionStat, error) {
```
GustavoCaso, line 83: I think, the fact that we need to convert value from `createCheck()` to `*diskv2.Check` to call their public API `WithDiskPartitions` in a way defeats the purpose of having a separate test package for the diskv2 package. 

I'm sure if we should continue having a separate test package, as it forces us to write this type of code. Nevertheless, I'm not going to die on this hill 😄

GustavoCaso (approved):
I'm ok with the changes on removing globals from the diskv2 package. Still, I left a comment regarding the code we are force to write because we have a separate test package, I would suggest to move everything on the same package and allow setting private fields directly on the test 😄

pkg/collector/corechecks/system/disk/diskv2/disk_test.go
```diff
@@ -1274,18 +1253,24 @@ service_check_rw: false
...
 				Mountpoint: "/dev/shm",
 				Fstype:     "tmpfs",
 				Opts:       []string{"rw", "nosuid", "nodev"},
 			}}, nil
-	}
-	diskv2.DiskUsage = func(_ string) (*gopsutil_disk.UsageStat, error) {
+	}).WithDiskUsage(func(_ string) (*gopsutil_disk.UsageStat, error) {
 		// Sleep 10s (longer than default timeout)
-		time.Sleep(10 * time.Second)
+		time.Sleep(60 * time.Second)
```
julien-lebot, line 1273: Should we update the comment here?

pkg/collector/corechecks/system/disk/diskv2/disk_test.go
```diff
@@ -1318,7 +1295,7 @@ func TestGivenADiskCheckWithDefaultConfig_WhenUsagePartitionTimeout_ThenUsageMet
 	// Explicitly wait until diskCheck.Run calls mockClock.After()
 	<-afterCalled
 	// Move the clock forward longer than default timeout (5s)
-	mockClock.Add(10 * time.Second)
+	mockClock.Add(5 * time.Second)
```
julien-lebot, line 1298: Same here

pkg/collector/corechecks/system/disk/diskv2/disk.go
```diff
@@ -640,26 +633,42 @@ func (c *Check) fetchAllDeviceLabels() error {
 	return c.fetchAllDeviceLabelsFromBlkid()
 }
 
-// Factory creates a new check factory
-func Factory() option.Option[func() check.Check] {
-	return option.New(newCheck)
+// WithClock sets a custom clock on the Check and returns the updated Check.
+func (c *Check) WithClock(clock clock.Clock) *Check {
```
pgimalac, line 637: We're adding more functions to the API just so that we can run tests from another package, that seems less and less convenient...

pgimalac (approved):

pkg/collector/corechecks/system/disk/diskv2/disk_windows_test.go
```diff
@@ -428,22 +273,18 @@ func TestGivenADiskCheckWithFileSystemGlobalExcludeNotConfigured_WhenCheckRuns_T
...
-	m.AssertMetric(t, "MonotonicCount", "system.disk.write_time", float64(450), "", []string{"device:/dev/sda1", "device_name:/dev/sda1"})
-	m.AssertMetric(t, "Rate", "system.disk.read_time_pct", float64(30), "", []string{"device:/dev/sda1", "device_name:/dev/sda1"})
-	m.AssertMetric(t, "Rate", "system.disk.write_time_pct", float64(45), "", []string{"device:/dev/sda1", "device_name:/dev/sda1"})
-	m.AssertMetric(t, "MonotonicCount", "system.disk.read_time", float64(500), "", []string{"device:/dev/sda2", "device_name:/dev/sda2"})
-	m.AssertMetric(t, "MonotonicCount", "system.disk.write_time", float64(150), "", []string{"device:/dev/sda2", "device_name:/dev/sda2"})
-	m.AssertMetric(t, "Rate", "system.disk.read_time_pct", float64(50), "", []string{"device:/dev/sda2", "device_name:/dev/sda2"})
-	m.AssertMetric(t, "Rate", "system.disk.write_time_pct", float64(15), "", []string{"device:/dev/sda2", "device_name:/dev/sda2"})
+	m.AssertMetric(t, "MonotonicCount", "system.disk.read_time", float64(300), "", []string{`device:\\?\Volume{a1b2c3d4-e5f6-7890-abcd-ef1234567890}\`, `device_name:?\volume{a1b2c3d4-e5f6-7890-abcd-ef1234567890}`})
```
julien-lebot, line 284: Can you explain why did the tags assertion change?
Previously we were setting up the test function like so:
```go
diskv2.DiskIOCounters = func(...string) (map[string]gopsutil_disk.IOCountersStat, error) {
		return map[string]gopsutil_disk.IOCountersStat{
			"\\\\?\\Volume{a1b2c3d4-e5f6-7890-abcd-ef1234567890}\\": {
				Name:       "sda1",
				ReadCount:  100,
				WriteCount: 200,
				ReadBytes:  1048576,
				WriteBytes: 2097152,
				ReadTime:   300,
				WriteTime:  450,
			},
		}, nil
	}
```
and now `CreateWindowsCheck` does:
```go
WithDiskIOCounters(func(...string) (map[string]gopsutil_disk.IOCountersStat, error) {
		return map[string]gopsutil_disk.IOCountersStat{
			"\\\\?\\Volume{a1b2c3d4-e5f6-7890-abcd-ef1234567890}\\": {
				Name:       "sda1",
				ReadCount:  100,
				WriteCount: 200,
				ReadBytes:  1048576,
				WriteBytes: 2097152,
				ReadTime:   300,
				WriteTime:  450,
			},
		}, nil
	})
	return diskCheck
}
```
but that doesn't explain the change in tags. What am I missing?

julien-lebot (commented):
Left some comments. Could you also update the description to explain why the timings in the code change?

pkg/collector/corechecks/system/disk/diskv2/disk_nix_test.go
jose-manuel-almaza: Using the .WithX() pattern instead of exposing struct fields directly allows you to control and validate state changes, maintain fluent configuration, and enable dependency injection when it can't be done via the constructor — improving testability, encapsulation, and long-term maintainability.
The reason is not running tests from another package.

pkg/collector/corechecks/system/disk/diskv2/disk_windows_test.go
jose-manuel-almaza: These tags come from `DiskPartitions` and now they're different by default for all Windows checks

pkg/collector/corechecks/system/disk/diskv2/disk_test.go
jose-manuel-almaza: Back to sleep of 10s

pkg/collector/corechecks/system/disk/diskv2/disk_test.go
jose-manuel-almaza: Same as what?

pkg/collector/corechecks/system/disk/diskv2/disk.go
pgimalac: Yeah but we're adding public APIs that are not used in the "real" code just so that we can test that, I don't see how that's better than just having the tests in the same package and setting the unexported fields in the tests

jose-manuel-almaza:
> Left some comments. Could you also update the description to explain why the timings in the code change?

Done

pkg/collector/corechecks/system/disk/diskv2/disk.go
GustavoCaso: I'm with @pgimalac on this one. The encapsulation is broken the moment we add an API that is only used, because we are not using the package in the test the way is going to be used in production (outside of the test package)

For this particular case, the fields are private, so they could only be modified inside the package, so being able to set them in the test seems less convoluted that the `.WithX()` pattern in this case

julien-lebot (approved):
👍 for Windows Product changes

pkg/collector/corechecks/system/disk/diskv2/disk.go
jose-manuel-almaza: Moved WithX methods from public API to some helpers into the `diskv2` package only available in tests.
This is a good solution after some research and combines the best of both worlds.

PR #36885, opened 2025-05-12
[CXP-2145] Moves the process checks enabled tests into the process component tests.

wiyu:
### What does this PR do?
This moves the tests out to the component and creates the component with mock component deps  to verify if the checks are enabled.

### Motivation
https://datadoghq.atlassian.net/browse/CXP-2145

### DIFF
--- a/comp/process/processcheck/processcheckimpl/check_linux_test.go
+++ b/comp/process/processcheck/processcheckimpl/check_linux_test.go
@@ -23,7 +23,7 @@ import (
 	"go.uber.org/fx"
 )
 
-// Make sure process checks run on the core agent only
+// TestProcessCheckEnablementOnCoreAgent Tests the process checks run on the core agent only
 func TestProcessCheckEnablementOnCoreAgent(t *testing.T) {
 	originalFlavor := flavor.GetFlavor()
 	defer flavor.SetFlavor(originalFlavor)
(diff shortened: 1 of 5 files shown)


comp/process/processcheck/processcheckimpl/check_test.go
```diff
@@ -0,0 +1,60 @@
...
+			},
+			enabled: false,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			c := fxutil.Test[processcheck.Component](t, fx.Options(
+				core.MockBundle(),
+				fx.Replace(config.MockParams{Overrides: tc.configs}),
```
Stephanie0829, line 49: Curious why this test uses an entire config, instead of a boolean  for `process_collection` configuration and building the config when running each test case (which is more similar to the linux test). Since it seems here that for the following code, the difference between the test cases configuration is only true or false for `process_collection`?
`configs: map[string]interface{}{
				"process_config.process_collection.enabled": true,
			},`

comp/process/processcheck/processcheckimpl/check_linux_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/DataDog/datadog-agent/comp/process/processcheck"
+	"github.com/DataDog/datadog-agent/pkg/util/flavor"
+	"github.com/DataDog/datadog-agent/pkg/util/fxutil"
+	"github.com/DataDog/datadog-go/v5/statsd"
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/fx"
+)
+
+// Make sure process checks run on the core agent only
+func TestProcessCheckEnablementOnCoreAgent(t *testing.T) {
```
Stephanie0829, line 27: (nit) Update comment to begin with function name `TestProcessCheckEnablementOnCoreAgent` to more align with go standards

Stephanie0829 (approved):

PR #36918, opened 2025-05-13
[CWS] Disable fentry on kernels >= 6.11

lebauce:
### What does this PR do?

Disable fentry on kernel >= 6.11

### Motivation

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=88c2a10e6c176c2860cd0659f4c0e9d20b3f64d1
broke our fentry support

### DIFF
--- a/pkg/security/probe/probe_ebpf.go
+++ b/pkg/security/probe/probe_ebpf.go
@@ -241,6 +241,10 @@ func (p *EBPFProbe) selectFentryMode() {
 		p.useFentry = false
 		seclog.Errorf("fentry enabled but not supported, falling back to kprobe mode")
 		return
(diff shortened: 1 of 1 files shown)


pkg/security/probe/probe_ebpf.go
```diff
@@ -237,7 +237,7 @@ func (p *EBPFProbe) selectFentryMode() {
 		return
 	}
 
-	if !p.kernelVersion.HaveFentrySupport() {
+	if !p.kernelVersion.HaveFentrySupport() || (p.kernelVersion.Code != 0 && p.kernelVersion.Code >= kernel.Kernel6_11) {
```
paulcacheux, line 240: I would be better to add a separate if section for this kernel check

paulcacheux (approved):

PR #36919, opened 2025-05-13
[CONTINT-4588] Treat AWS account ID as strings to avoid loosing leading zeroes

L3n41c:
### What does this PR do?

Change the type used to manipulate AWS account ID from `int` to `string`.

### Motivation

There are currently some bugs when an AWS account ID starts with leading zeroes because some components drop the leading zeroes (the agent when it manipulates it as `int`) and some don’t (crawlers?).

Alternatively, we could have only added the leading zeroes with a `Sprintf("%012d", …)` every time the AWS account ID is converted to a string, but that would have been more fragile. Indeed, if we use `Sprintf("%012d, …)` when building the tag, the behavior would have been correct, but if the logs were still using `%d`, we might have thought that it was not correct.
By using `string` everywhere, we are more sure to consistently not forget leading zeroes anywhere.

### Additional Notes

In an ideal world, we would have changed the type of AWS account ID in `agent-payload` as well.
https://github.com/DataDog/agent-payload/blob/94867fa6043c7ab5b2b10fccd98d04275c44819c/proto/process/agent.proto#L113-L114.
But that would have been a non-backward compatible change.

And, anyhow, the backend already implemented a hack to add the leading zeroes back: https://github.com/DataDog/dd-go/pull/168043/files


pgimalac (approved):

kangyili (approved):
thanks for the fix

sblumenthal (approved):

wdhif (approved):

rtrieu (approved):

PR #36926, opened 2025-05-13
[CWS-4500] Expose rule tags in secl

YoannGh:
### What does this PR do?

Expose rule tags in secl.

### Motivation

Allow writing rules such as the following and chaining rules based on specific tags.
```
exec.file.name == "foo" && rule.tags in ["tag:bar"]
```


drichards-87 (approved):

safchain (approved):

pkg/security/secl/model/model.go
```diff
@@ -131,6 +131,9 @@ type BaseEvent struct {
 	Service       string         `field:"event.service,handler:ResolveService,opts:skip_ad|gen_getters"` // SECLDoc[event.service] Definition:`Service associated with the event`
 	Hostname      string         `field:"event.hostname,handler:ResolveHostname"`                        // SECLDoc[event.hostname] Definition:`Hostname associated with the event`
 
+	// Fields populated upon rule evaluation
+	RuleTags []string `field:"rule.tags"` // SECLDoc[rule.tags] Definition:`Tags associated with the rule that's used to evaluate the event`
```
safchain, line 135: should we keep the same syntax has the other event level field ? `event.rule.tags`

PR #36927, opened 2025-05-13
[EBPF] kmt: fix timeouts in complexity generation job

gjulianm:
### What does this PR do?

This PR adds filter to the `download_complexity_data` KMT task so that it only downloads data from jobs with complexity collection enabled. This avoids an issue we've seen where downloading artifacts for files that don't exist take a long time (100 seconds) and cause job timeouts.

In order to avoid duplicating in the Python code the definition of which jobs have complexity collection enabled, the code parses the `.gitlab-ci.yml` configuration and grabs the requirements for the `notify_ebpf_complexity_changes` job.

This PR also adds timing data for each download, so that we can debug the issue if it happens again.

### Motivation

Avoid timeouts in complexity collection.

CI job passing and running below the timeout (8 minutes): https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/941537682


.gitlab/kernel_matrix_testing/common.yml
```diff
@@ -338,6 +338,6 @@ notify_ebpf_complexity_changes:
     - python3 -m dda self dep sync -f legacy-tasks
     - !reference [.setup_agent_github_app]
     - DD_API_KEY=$($CI_PROJECT_DIR/tools/ci/fetch_secret.sh $AGENT_API_KEY_ORG2 token) || exit $?; export DD_API_KEY
-    - GITLAB_TOKEN=$($CI_PROJECT_DIR/tools/ci/fetch_secret.sh $GITLAB_TOKEN read_api) || exit $?; export GITLAB_TOKEN
+    - GITLAB_TOKEN=$($CI_PROJECT_DIR/tools/ci/fetch_secret.sh $GITLAB_TOKEN write_api) || exit $?; export GITLAB_TOKEN  # Require write_api to be able to interpret the .gitlab-ci configuration, we need access to the CI linter
```
brycekahle, line 341: We need API write access to parse the Gitlab config?

gjulianm: Fixed, we don't need that token anymore :D

brycekahle (approved):

tasks/libs/types/types.py
```diff
@@ -105,3 +107,45 @@ class PermissionCheck(Enum):
...
+
+        return False
+
+    def _match_tagset(self, tagset: list[tuple[str, set[str]]], job_tags: list[str]) -> bool:
+        for wanted, value in zip(tagset, job_tags, strict=False):
+            wanted_values = wanted[1]
+            if value not in wanted_values:
+                return False
+
+        return True
```
CelianR, line 151: NIT: It would be nice to add some unit tests for that

CelianR (approved):

tasks/libs/types/types.py
gjulianm: Done!

PR #36936, opened 2025-05-13
[NPM-4440] Add TracerouteSerial variant

pimlu:
Stacked on [\[NPM-4440\] Change SACK traceroute to use AF_PACKET](https://github.com/DataDog/datadog-agent/pull/36893)

### What does this PR do?
This PR adds TracerouteSerial which uses a TracerouteDriver to make a sequential traceroute.

### Motivation
Unfortunately this is required for TCP SYN if we want to start randomizing the packet ID like `tcptraceroute` does, because when we get back the SYNACK, the packet ID is _not_ reflected in the response. Because of this, `tcptraceroute` is fundamentally not parallelizable unless we try additional tricks like the TCP timestamp option.

New tests + existing tests should pass:
```
go test -timeout 30s -tags linux,linux_bpf,npm,process,test github.com/DataDog/datadog-agent/pkg/networkpath/traceroute/common
```

### DIFF
--- a/pkg/networkpath/traceroute/common/traceroute_serial.go
+++ b/pkg/networkpath/traceroute/common/traceroute_serial.go
@@ -66,8 +66,6 @@ func TracerouteSerial(ctx context.Context, t TracerouteDriver, p TracerouteSeria
 
 		// wait for at least SendDelay to pass
 		<-sendDelay
-
-		cancel()
 	}
 
 	// if we got externally cancelled, report that
--- a/pkg/networkpath/traceroute/common/traceroute_serial.go
+++ b/pkg/networkpath/traceroute/common/traceroute_serial.go
@@ -16,15 +16,9 @@ type TracerouteSerialParams struct {
 	TracerouteParams
 }
 
-// MaxTimeout combines the timeout+probe delays into a total timeout for the traceroute
-func (p TracerouteSerialParams) MaxTimeout() time.Duration {
-	delaySum := p.TracerouteTimeout * time.Duration(p.ProbeCount())
-	return delaySum
-}
-
 // TracerouteSerial runs a traceroute in serial. Sometimes this is necessary over TracerouteParallel
 // because the driver doesn't support parallel.
-func TracerouteSerial(ctx context.Context, t TracerouteDriver, p TracerouteParallelParams) ([]*ProbeResponse, error) {
+func TracerouteSerial(ctx context.Context, t TracerouteDriver, p TracerouteSerialParams) ([]*ProbeResponse, error) {
 	if err := p.validate(); err != nil {
 		return nil, err
 	}
--- a/pkg/networkpath/traceroute/common/traceroute_serial.go
+++ b/pkg/networkpath/traceroute/common/traceroute_serial.go
@@ -1,0 +1,85 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-present Datadog, Inc.
+
+package common
+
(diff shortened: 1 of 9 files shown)


pkg/networkpath/traceroute/common/traceroute_serial.go
```diff
@@ -0,0 +1,85 @@
...
+package common
+
+import (
+	"context"
+	"fmt"
+	"time"
+)
+
+// TracerouteSerialParams are the parameters for TracerouteSerial
+type TracerouteSerialParams struct {
```
akarpz, line 15: is this used?

pkg/networkpath/traceroute/common/traceroute_serial.go
```diff
@@ -0,0 +1,85 @@
...
+	TracerouteParams
+}
+
+// MaxTimeout combines the timeout+probe delays into a total timeout for the traceroute
+func (p TracerouteSerialParams) MaxTimeout() time.Duration {
```
akarpz, line 20: same here

pkg/networkpath/traceroute/common/traceroute_serial.go
```diff
@@ -0,0 +1,85 @@
...
+	delaySum := p.TracerouteTimeout * time.Duration(p.ProbeCount())
+	return delaySum
+}
+
+// TracerouteSerial runs a traceroute in serial. Sometimes this is necessary over TracerouteParallel
+// because the driver doesn't support parallel.
+func TracerouteSerial(ctx context.Context, t TracerouteDriver, p TracerouteParallelParams) ([]*ProbeResponse, error) {
```
akarpz, line 27: maybe this is intended to take `TracerouteSerialParams`?

pkg/networkpath/traceroute/common/traceroute_serial.go
```diff
@@ -0,0 +1,85 @@
...
+	for i := int(p.MinTTL); i <= int(p.MaxTTL); i++ {
+		if ctx.Err() != nil {
+			break
+		}
+		sendDelay := time.After(p.SendDelay)
+
+		start := time.Now()
+		deadline := start.Add(p.TracerouteTimeout)
+		timeoutCtx, cancel := context.WithDeadline(ctx, deadline)
+		defer cancel()
```
akarpz, line 42: don't you already call cancel at the end of the loop?

pkg/networkpath/traceroute/common/traceroute_serial.go
pimlu: Whoops, will fix the references, thanks

pkg/networkpath/traceroute/common/traceroute_serial.go
pimlu: Good catch, thanks

pkg/networkpath/traceroute/common/traceroute_serial.go
pimlu: Yeah I think I actually don't need this for serial, thanks

pkg/networkpath/traceroute/common/traceroute_serial.go
pimlu: I do, but in theory if SendProbe() returns an error, timeoutCtx would leak without calling cancel(). To be honest it probably doesn't matter since I heard the memory leak solves itself once the deadline passes, but I just covered all bases.

Maybe it's not even worth calling cancel() a second time at the end of the loop and I should just leave it to `defer`.

akarpz: yeah I would be supportive of removing the latter cancel

akarpz (approved):

PR #36854, opened 2025-05-12
usm: marshaling: Unify store stats

guyarb:
### What does this PR do?

Unifies USM `store<protocol>Stats` methods using a generic function

### Motivation

Our current implementations for HTTP, HTTP/2, Kafka, Postgres and Redis, are just a copy-paste with slight modifications The change unifies the implementation to spare redundancy

PR #36937, opened 2025-05-13
[USMON-1441] usm: Convert USM protocol data to OS specific.

guyarb:
### What does this PR do?

Convert USM protocol data into OS specific, such that only the supported protocols will be accessible in each OS.

Windows - http
Linux - http, http2, kafka, postgres, redis
Anything else - no protocol.

### Motivation

* Preventing irrelevant code from being accessed in unsupported OS.
* Allowing linux only protocols to use auto-generate structs from our kernel.
* Reducing friction with adding USM protocols.

Followup on #36852 and #36854

### Additional Notes

Please review it commit by commit


akarpz (approved):
LGTM for CNM files

amitslavin (approved):
🚀

clarkb7 (approved):

PR #36943, opened 2025-05-14
[SINT-3271] Sign Testing Linux Repositories with a Test Key

cedricvanrompay-datadog:
### What does this PR do?

This PR changes the way test repositories (`apttesting.datad0g.com` and `yumtesting.datad0g.com`) are set up in Gitlab CI:
- Instead of reusing the same production GPG keys to sign these repositories' metadata, a set of dedicated testing signing keys is provided.
  - The public keys are stored in `apttesting.datad0g.com/test-keys-vault`.
  - These public keys are passed to the relevant tests via the `TESTING_KEYS_URL` variable (for install script-based tests), or other similar methods. Some E2E tests had to be manually updated because they were hardcoding the keys URL.
- The testing keys are stored in an Hashcorp Vault instance instead of AWS SSM, and are not directly retrieved. We use a homemade tool called `gpg-vault` that mimics a subset of the GPG operations, that can use Vault as its GPG key storage for signing requests.
- A manual job to export the testing GPG key has been added, because the `datadog-agent` Gitlab CI is the only place that can perform this operation.

### Motivation

Stop using the production GPG keys for non-production artifacts and repositories.

This PR was tested against all E2E tests that use the testing repositories, including the ones in agent-linux-install-script (see [results](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1050849829))

- `gpg-vault` is currently installed at runtime, which bears a small reliability risk (what if the download fails). We are considering adding it to the base CI images.

### Additional Notes

- The packages are still signed with the production GPG keys. We also aim at signing them with testing keys, but that will come later. The install tests still work because the public parts of both the testing and production keys are available in `apttesting.datad0g.com/test-keys-vault`. Once packages are also signed with the testing GPG keys, we will not need to store the public production GPG keys there anymore.
- Down the line, the public part of the testing GPG keys will be moved to a dedicated place. Also, the key ID components (user name, email, etc.) will also be stored in Vault, so that we do not have to specify it whenever we use `gpg-vault`.


.gitlab/e2e_testing_deploy/e2e_deploy.yml
```diff
@@ -2,18 +2,25 @@
...
-  - printf -- "$($CI_PROJECT_DIR/tools/ci/fetch_secret.sh $DEB_GPG_KEY)" | gpg --import --batch
-  - EXIT="${PIPESTATUS[0]}"; if [ $EXIT -ne 0 ]; then echo "Unable to locate credentials needs gitlab runner restart"; exit $EXIT; fi
-  - APT_SIGNING_KEY_PASSPHRASE=$($CI_PROJECT_DIR/tools/ci/fetch_secret.sh $DEB_SIGNING_PASSPHRASE) || exit $?
-
-.setup_signing_keys_package:
-  &setup_signing_keys_package # Set up prod apt repo to get the datadog-signing-keys package
+.setup_signing: &setup_signing
+  - aws s3 cp s3://binaries-ddbuild-io-prod/dd-source/domains/seceng/sit/apps/artifact-security/vault-gpg-client/gpg-vault/release/gpg-vault-v1.5.0-linux-amd64.tar.gz .
+  - tar -xzf gpg-vault-v1.5.0-linux-amd64.tar.gz
+  - mv gpg-vault-linux-amd64 /usr/bin/gpg-vault-bin
```
chouetz: Should we eventually install this directly on the buildimages?

.gitlab/e2e_testing_deploy/e2e_deploy.yml
```diff
@@ -60,15 +67,15 @@ deploy_deb_testing-a7_x64:
...
-    - echo "$APT_SIGNING_KEY_PASSPHRASE" | deb-s3 upload -c "stable-x86_64" -m 7 -b $DEB_TESTING_S3_BUCKET -a amd64 --sign=$DEB_GPG_KEY_ID --gpg_options="--passphrase-fd 0 --batch --digest-algo SHA512" --preserve_versions --visibility public --prefix datadog-agent/pipeline-${DD_PIPELINE_ID} $OMNIBUS_PACKAGE_DIR/datadog-signing-keys_${DD_PIPELINE_ID}.deb
-    - echo "$APT_SIGNING_KEY_PASSPHRASE" | deb-s3 upload -c "stable-x86_64" -m 7 -b $DEB_TESTING_S3_BUCKET -a x86_64 --sign=$DEB_GPG_KEY_ID --gpg_options="--passphrase-fd 0 --batch --digest-algo SHA512" --preserve_versions --visibility public --prefix datadog-agent/pipeline-${DD_PIPELINE_ID} $OMNIBUS_PACKAGE_DIR/datadog-signing-keys_${DD_PIPELINE_ID}.deb
+    - *setup_signing
+    - deb-s3 upload -c "stable-x86_64" -m 7 -b $DEB_TESTING_S3_BUCKET -a amd64 --sign=$GPG_TEST_KEY_ID --gpg_options="--batch --digest-algo sha512" --preserve_versions --visibility public --prefix datadog-agent/pipeline-${DD_PIPELINE_ID} $OMNIBUS_PACKAGE_DIR/datadog-*_7*amd64.deb
```
chouetz, line 75: I suppose this is fine to not have a passphrase? Should we do `echo ""` like in `rpm` case?

.gitlab/e2e_testing_deploy/e2e_deploy.yml
```diff
@@ -154,9 +159,8 @@ deploy_suse_rpm_testing_x64-a7:
   before_script:
     - ls $OMNIBUS_PACKAGE_DIR_SUSE
   script:
-    - *setup_rpm_signing_key
-    - set +x
-    - echo "$RPM_SIGNING_PASSPHRASE" | python2 /opt/rpm-s3/bin/rpm-s3 --verbose --visibility public-read -c "https://s3.amazonaws.com" -b $RPM_TESTING_S3_BUCKET -p "suse/testing/pipeline-$DD_PIPELINE_ID/7/x86_64/" -a "x86_64" --sign --metadata-signing-key $RPM_GPG_KEY_ID --repodata-store-public-key $OMNIBUS_PACKAGE_DIR_SUSE/datadog-*-7.*x86_64.rpm
+    - *setup_signing
+    - python2 /opt/rpm-s3/bin/rpm-s3 --verbose --visibility public-read -c "https://s3.amazonaws.com" -b $RPM_TESTING_S3_BUCKET -p "suse/testing/pipeline-$DD_PIPELINE_ID/7/x86_64/" -a "x86_64" --sign --metadata-signing-key $GPG_TEST_KEY_ID --repodata-store-public-key $OMNIBUS_PACKAGE_DIR_SUSE/datadog-*-7.*x86_64.rpm
```
chouetz, line 163:
```suggestion
    - echo "" | python2 /opt/rpm-s3/bin/rpm-s3 --verbose --visibility public-read -c "https://s3.amazonaws.com" -b $RPM_TESTING_S3_BUCKET -p "suse/testing/pipeline-$DD_PIPELINE_ID/7/x86_64/" -a "x86_64" --sign --metadata-signing-key $GPG_TEST_KEY_ID --repodata-store-public-key $OMNIBUS_PACKAGE_DIR_SUSE/datadog-*-7.*x86_64.rpm
```
I suggest to be consistent with the above (and I don't know if we need to pass the empty passphrase)

.gitlab/e2e_testing_deploy/e2e_deploy.yml
```diff
@@ -172,9 +176,8 @@ deploy_suse_rpm_testing_arm64-a7:
   before_script:
     - ls $OMNIBUS_PACKAGE_DIR_SUSE
   script:
-    - *setup_rpm_signing_key
-    - set +x
-    - echo "$RPM_SIGNING_PASSPHRASE" | python2 /opt/rpm-s3/bin/rpm-s3 --verbose --visibility public-read -c "https://s3.amazonaws.com" -b $RPM_TESTING_S3_BUCKET -p "suse/testing/pipeline-$DD_PIPELINE_ID/7/aarch64/" -a "aarch64" --sign --metadata-signing-key $RPM_GPG_KEY_ID --repodata-store-public-key $OMNIBUS_PACKAGE_DIR_SUSE/datadog-*-7.*aarch64.rpm
+    - *setup_signing
+    - python2 /opt/rpm-s3/bin/rpm-s3 --verbose --visibility public-read -c "https://s3.amazonaws.com" -b $RPM_TESTING_S3_BUCKET -p "suse/testing/pipeline-$DD_PIPELINE_ID/7/aarch64/" -a "aarch64" --sign --metadata-signing-key $GPG_TEST_KEY_ID --repodata-store-public-key $OMNIBUS_PACKAGE_DIR_SUSE/datadog-*-7.*aarch64.rpm
```
chouetz, line 180:
```suggestion
    - echo "" | python2 /opt/rpm-s3/bin/rpm-s3 --verbose --visibility public-read -c "https://s3.amazonaws.com" -b $RPM_TESTING_S3_BUCKET -p "suse/testing/pipeline-$DD_PIPELINE_ID/7/aarch64/" -a "aarch64" --sign --metadata-signing-key $GPG_TEST_KEY_ID --repodata-store-public-key $OMNIBUS_PACKAGE_DIR_SUSE/datadog-*-7.*aarch64.rpm
```
Same as previously

chouetz (commented):
Minor comments. No need to approve as the PR is still in draft mode (this review will be ignored)
Thanks for this contribution!

.gitlab/e2e_testing_deploy/e2e_deploy.yml
KSerrania: Eventually yes, but we are not ready for this yet:
- the `gpg-vault` project is not open-source yet (and its releases are not public), so hard to integrate into public buildimages
- once `gpg-vault` is installed this way, the original `gpg` cannot be used anymore. In our current state, we still have to use the original `gpg` binary in other jobs in the pipeline, so we cannot replace it yet.

.gitlab/e2e_testing_deploy/e2e_deploy.yml
KSerrania: I'm pretty sure this is not needed, I removed them all.

.gitlab/e2e_testing_deploy/e2e_deploy.yml
```diff
@@ -2,18 +2,25 @@
...
-
-.setup_apt_signing_key: &setup_apt_signing_key
-  - printf -- "$($CI_PROJECT_DIR/tools/ci/fetch_secret.sh $DEB_GPG_KEY)" | gpg --import --batch
-  - EXIT="${PIPESTATUS[0]}"; if [ $EXIT -ne 0 ]; then echo "Unable to locate credentials needs gitlab runner restart"; exit $EXIT; fi
-  - APT_SIGNING_KEY_PASSPHRASE=$($CI_PROJECT_DIR/tools/ci/fetch_secret.sh $DEB_SIGNING_PASSPHRASE) || exit $?
-
-.setup_signing_keys_package:
-  &setup_signing_keys_package # Set up prod apt repo to get the datadog-signing-keys package
+.setup_signing: &setup_signing
+  - aws s3 cp s3://binaries-ddbuild-io-prod/dd-source/domains/seceng/sit/apps/artifact-security/vault-gpg-client/gpg-vault/dev/gpg-vault-v1.5.1-linux-amd64.tar.gz .
```
spencergilbert: **💬 suggestion:** ‏Since we're already aware of the risk of flakey downloads here, can we wrap it in a retry? We should already have one available in the agent CI environment.

.gitlab/e2e_testing_deploy/e2e_deploy.yml
```diff
@@ -222,3 +225,18 @@ deploy_windows_testing-a7-fips:
...
+  stage: e2e_deploy
+  image: registry.ddbuild.io/ci/datadog-agent-buildimages/gitlab_agent_deploy$CI_IMAGE_GITLAB_AGENT_DEPLOY_SUFFIX:$CI_IMAGE_GITLAB_AGENT_DEPLOY
+  tags: ["arch:amd64"]
+  script:
+    - *setup_signing
+    - /usr/bin/gpg-vault export-public-key $GPG_TEST_KEY_ID > gpg-testing-public-key.asc
+  artifacts:
+    paths:
+      - gpg-testing-public-key.asc
+    expire_in: 1 day
```
spencergilbert: **❓ question:** ‏I assume this is just for manual inspection of the key, do we need it moving forward?

spencergilbert (approved):

.gitlab/e2e_testing_deploy/e2e_deploy.yml
KSerrania: I can comment it, but I'd rather leave its definition somewhere because it will be needed each time we rotate that testing key (since the datadog-agent CI is the only place that can export the public key for the testing key).

.gitlab/e2e_testing_deploy/e2e_deploy.yml
KSerrania: Yes, we can. I'll let DevX chime in on what's the preferred way. I could:
- do nothing because I believe all jobs are now retried by default, except test jobs,
- use the `${CI_PROJECT_DIR}/tools/ci/retry.sh` tool,
- use AWS params like [what is done in the macOS jobs](https://github.com/DataDog/datadog-agent/blob/4c44df5a5e04c61d2febb7520e98af75895c408d/.gitlab/common/macos.yml#L91-L93).

KevinFairise2: I think nothing should be required there, the retry should be handled directly with the configuration that is passed inside the image: https://github.com/DataDog/datadog-agent-buildimages/blob/main/agent-deploy/Dockerfile#L110

test/new-e2e/tests/installer/unix/all_packages_test.go
```diff
@@ -202,6 +202,10 @@ func (s *packageBaseSuite) updateCurlOnUbuntu() {
 func (s *packageBaseSuite) RunInstallScriptProdOci(params ...string) error {
 	env := map[string]string{}
 	installScriptPackageManagerEnv(env, s.arch)
+
+	// TODO: change keys URL env variable
+	// depending on s.os?
```
CelianR: Is this for a future PR or should this be updated in this PR?

.gitlab/e2e_testing_deploy/e2e_deploy.yml
```diff
@@ -2,18 +2,25 @@
...
-.setup_signing_keys_package:
-  &setup_signing_keys_package # Set up prod apt repo to get the datadog-signing-keys package
+.setup_signing: &setup_signing
+  - aws s3 cp s3://binaries-ddbuild-io-prod/dd-source/domains/seceng/sit/apps/artifact-security/vault-gpg-client/gpg-vault/dev/gpg-vault-v1.5.1-linux-amd64.tar.gz .
+  - tar -xzf gpg-vault-v1.5.1-linux-amd64.tar.gz
+  - mv gpg-vault-linux-amd64 /usr/bin/gpg-vault-bin
+
+  # Set up gpg shim
+  - echo '#!/bin/bash' > /usr/bin/gpg
+  - echo 'GPG_VAULT_COMPAT_MODE=true GPG_USER_ID="Datadog Packages" GPG_USER_EMAIL="package@datadoghq.com" /usr/bin/gpg-vault $@' >> /usr/bin/gpg
```
CelianR:
```suggestion
  - echo 'GPG_VAULT_COMPAT_MODE=true GPG_USER_ID="Datadog Packages" GPG_USER_EMAIL="package@datadoghq.com" /usr/bin/gpg-vault "$@"' >> /usr/bin/gpg
```

.gitlab/e2e_testing_deploy/e2e_deploy.yml
```diff
@@ -2,18 +2,25 @@
...
+  - chmod +x /usr/bin/gpg
+
+  # Set up gpg-vault shim
+  - echo '#!/bin/bash' > /usr/bin/gpg-vault
+  - echo 'GPG_USER_ID="Datadog Packages" GPG_USER_EMAIL="package@datadoghq.com" /usr/bin/gpg-vault-bin $@' >> /usr/bin/gpg-vault
```
CelianR:
```suggestion
  - echo 'GPG_USER_ID="Datadog Packages" GPG_USER_EMAIL="package@datadoghq.com" /usr/bin/gpg-vault-bin "$@"' >> /usr/bin/gpg-vault
```

CelianR (approved):

test/new-e2e/tests/installer/unix/all_packages_test.go
KSerrania: I'll remove it, in the end it's not necessary.

KSerrania:

KSerrania:

KSerrania:

BaptisteFoy (approved):

.gitlab/e2e_testing_deploy/e2e_deploy.yml
```diff
@@ -2,18 +2,25 @@
...
-  - printf -- "$($CI_PROJECT_DIR/tools/ci/fetch_secret.sh $DEB_GPG_KEY)" | gpg --import --batch
-  - EXIT="${PIPESTATUS[0]}"; if [ $EXIT -ne 0 ]; then echo "Unable to locate credentials needs gitlab runner restart"; exit $EXIT; fi
-  - APT_SIGNING_KEY_PASSPHRASE=$($CI_PROJECT_DIR/tools/ci/fetch_secret.sh $DEB_SIGNING_PASSPHRASE) || exit $?
-
-.setup_signing_keys_package:
-  &setup_signing_keys_package # Set up prod apt repo to get the datadog-signing-keys package
+.setup_signing: &setup_signing
+  - ${CI_PROJECT_DIR}/tools/ci/retry.sh aws s3 cp s3://binaries-ddbuild-io-prod/dd-source/domains/seceng/sit/apps/artifact-security/vault-gpg-client/gpg-vault/dev/gpg-vault-v1.5.1-linux-amd64.tar.gz .
+  - tar -xzf gpg-vault-v1.5.1-linux-amd64.tar.gz
+  - mv gpg-vault-linux-amd64 /usr/bin/gpg-vault-bin
```
rdesgroppes: Okay given:
> - `gpg-vault` is currently installed at runtime, which bears a small reliability risk (what if the download fails). We are considering adding it to the base CI images.

.gitlab/e2e_testing_deploy/e2e_deploy.yml
```diff
@@ -2,18 +2,25 @@
...
+
+  # Set up gpg shim
+  - echo '#!/bin/bash' > /usr/bin/gpg
```
rdesgroppes: nit: the canonical shebang for `bash` is `#!/usr/bin/env bash` (for [portability](https://en.wikipedia.org/wiki/Shebang_(Unix)#Portability)):
```suggestion
  - echo '#!/usr/bin/env bash' > /usr/bin/gpg
```
There are several occurrences in the file, but please **feel free to ignore**.

.gitlab/e2e_testing_deploy/e2e_deploy.yml
```diff
@@ -2,18 +2,25 @@
...
+  - echo 'GPG_VAULT_COMPAT_MODE=true GPG_USER_ID="Datadog Packages" GPG_USER_EMAIL="package@datadoghq.com" /usr/bin/gpg-vault "$@"' >> /usr/bin/gpg
+  - chmod +x /usr/bin/gpg
+
+  # Set up gpg-vault shim
+  - echo '#!/bin/bash' > /usr/bin/gpg-vault
+  - echo 'GPG_USER_ID="Datadog Packages" GPG_USER_EMAIL="package@datadoghq.com" /usr/bin/gpg-vault-bin "$@"' >> /usr/bin/gpg-vault
+  - chmod +x /usr/bin/gpg-vault
+
+# TODO maybe: add an empty "datadog-signing-keys" package to the pool,
```
rdesgroppes: I think `TODO`s and `FIXME`s are fine when they come with an `(owner)`, i.e. without having to `git blame`.

.gitlab/e2e_testing_deploy/e2e_deploy.yml
```diff
@@ -114,9 +121,8 @@ deploy_rpm_testing-a7_x64:
       "ddot_rpm-x64",
     ]
   script:
-    - *setup_rpm_signing_key
-    - set +x
-    - echo "$RPM_SIGNING_PASSPHRASE" | python2 /opt/rpm-s3/bin/rpm-s3 --verbose --visibility public-read -c "https://s3.amazonaws.com" -b $RPM_TESTING_S3_BUCKET -p "testing/pipeline-$DD_PIPELINE_ID/7/x86_64/" -a "x86_64" --sign --metadata-signing-key $RPM_GPG_KEY_ID $OMNIBUS_PACKAGE_DIR/datadog-*-7.*x86_64.rpm
+    - *setup_signing
+    - python2 /opt/rpm-s3/bin/rpm-s3 --verbose --visibility public-read -c "https://s3.amazonaws.com" -b $RPM_TESTING_S3_BUCKET -p "testing/pipeline-$DD_PIPELINE_ID/7/x86_64/" -a "x86_64" --sign --metadata-signing-key $GPG_TEST_KEY_ID $OMNIBUS_PACKAGE_DIR/datadog-*-7.*x86_64.rpm
```
rdesgroppes, line 125: note to self: omg, [python2](https://www.python.org/doc/sunset-python-2/)?!

.gitlab/e2e_testing_deploy/e2e_deploy.yml
```diff
@@ -222,3 +225,20 @@ deploy_windows_testing-a7-fips:
...
+#   stage: e2e_deploy
+#   image: registry.ddbuild.io/ci/datadog-agent-buildimages/gitlab_agent_deploy$CI_IMAGE_GITLAB_AGENT_DEPLOY_SUFFIX:$CI_IMAGE_GITLAB_AGENT_DEPLOY
+#   tags: ["arch:amd64"]
+#   script:
+#     - *setup_signing
+#     - /usr/bin/gpg-vault export-public-key $GPG_TEST_KEY_ID > gpg-testing-public-key.asc
+#   artifacts:
+#     paths:
+#       - gpg-testing-public-key.asc
+#     expire_in: 1 day
```
rdesgroppes: Do we want to push a commented out block to the mainline? Couldn't we mark the job as `manual` instead?

KSerrania: That was discussed in https://github.com/DataDog/datadog-agent/pull/36943#discussion_r2239827440 - I want the definition to be somewhere in the tree for the day we need it, but not having it active means one less job in the very long list of jobs in this pipeline.

.gitlab/e2e_testing_deploy/e2e_deploy.yml
KSerrania: Yes, until we either rewrite or drop `rpm-s3`.

.gitlab/e2e_testing_deploy/e2e_deploy.yml
rdesgroppes: I see. May I then suggest:
```yaml
rules:
  - when: never
  - !reference [.manual]
```
The job should not show up in the pipeline view and the configuration would still be validated by GitLab, or is it too much?

KSerrania: Sounds good to me

KSerrania: Done in https://github.com/DataDog/datadog-agent/pull/36943/commits/7facf5010ff1b403ec2c4a9c0bfd235bab11f8f4

rdesgroppes: I guess https://github.com/DataDog/datadog-agent/commit/7facf5010ff1b403ec2c4a9c0bfd235bab11f8f4 rather addresses https://github.com/DataDog/datadog-agent/pull/36943#discussion_r2245264615

KSerrania: Indeed, wrong comment thread. https://github.com/DataDog/datadog-agent/pull/36943/commits/8dfe3ade893852431c415742520fa01ab5dabc19 addresses this one.

KevinFairise2:
We have some code on test-infra-definitions that can fetch the artifacts from apt/yumtesting bucket: [here](https://github.com/DataDog/test-infra-definitions/blob/main/components/datadog/agent/host_linuxos.go#L33-L40). Could you handle the update here also once this PR is merged?

rdesgroppes (approved):

PR #36960, opened 2025-05-14
Get check loader priorities lazily

pgimalac:
### What does this PR do?
When registering a check loader, do not provide the priority directly but instead make it lazily returned by the factory.

### Motivation
We will soon need to pick loader priority depending on the configuration, so it can't be statically defined in an `init` function.
Having the priority returned by the factory function allows delaying the choice until the configuration is available.

CI is enough as there should be no functional change in this PR.

### DIFF
--- a/pkg/collector/corechecks/loader.go
+++ b/pkg/collector/corechecks/loader.go
@@ -86,9 +86,10 @@ func (gl *GoCheckLoader) String() string {
 }
 
 func init() {
-	factory := func(sender.SenderManager, option.Option[integrations.Component], tagger.Component) (check.Loader, error) {
-		return NewGoCheckLoader()
+	factory := func(sender.SenderManager, option.Option[integrations.Component], tagger.Component) (check.Loader, int, error) {
+		loader, err := NewGoCheckLoader()
+		return loader, 30, err
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/collector/corechecks/loader.go
```diff
@@ -86,9 +86,10 @@ func (gl *GoCheckLoader) String() string {
 }
 
 func init() {
-	factory := func(sender.SenderManager, option.Option[integrations.Component], tagger.Component) (check.Loader, error) {
-		return NewGoCheckLoader()
+	factory := func(sender.SenderManager, option.Option[integrations.Component], tagger.Component) (check.Loader, int, error) {
+		loader, err := NewGoCheckLoader()
+		return loader, 30, err
```
louis-cqrl, line 91: I know here we just mimic the value written before nonetheless can we add a comment that explain why this value (same for the 20 bellow) ?

louis-cqrl (approved):

pkg/collector/corechecks/loader.go
pgimalac: Discussed IRL, values are quite arbitrary, the point is just to sort the loader relatively to the others

PR #36961, opened 2025-05-14
Replace remaining `invoke` with `dda` in datadog-agent build definition

alopezz:
### What does this PR do?

Replaces remaining occurrences of `invoke` with their `dda` equivalents to make sure we're using the same version of the tooling.

### Motivation

https://github.com/DataDog/datadog-agent/pull/36423#discussion_r2088754134

Passing build.

### DIFF
--- a/omnibus/config/software/datadog-agent.rb
+++ b/omnibus/config/software/datadog-agent.rb
@@ -136,7 +136,7 @@ build do
 
   if not bundled_agents.include? "trace-agent"
     platform = windows_arch_i386? ? "x86" : "x64"
-    command "dda inv -- trace-agent.build --install-path=#{install_dir} --major-version #{major_version_arg} --flavor #{flavor_arg}", :env => env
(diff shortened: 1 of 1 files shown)


omnibus/config/software/datadog-agent.rb
```diff
@@ -136,7 +136,7 @@
 
   if not bundled_agents.include? "trace-agent"
     platform = windows_arch_i386? ? "x86" : "x64"
-    command "invoke trace-agent.build --install-path=#{install_dir} --major-version #{major_version_arg} --flavor #{flavor_arg}", :env => env
+    command "dda inv -- trace-agent.build --install-path=#{install_dir} --major-version #{major_version_arg} --flavor #{flavor_arg}", :env => env
```
FlorentClarret, line 139: can we also add `-e` to be consistent and get more data?

FlorentClarret (approved):

chouquette (approved):
Thanks!

omnibus/config/software/datadog-agent.rb
alopezz: Added!

PR #36968, opened 2025-05-14
[EBPF] kmt: improve fuzzy matching for VM names

gjulianm:
### What does this PR do?

This PR updates the way we generate possible fuzzy matching candidates:

- Stop generating candidates for architectures that don't exist. This could cause confusing errors if the user entered a name that matched an existing OS but with an unsupported architecture (e.g., Ubuntu 16 is supported only for x64, so `distro-ubuntu_16.04-arm64` would show a `KeyError` with no further explanation).
- Allow selecting OSs with only the minor version and without underscore (e.g., `ubuntu22-arm64-distro` would not correctly select `ubuntu22`).
- Also adds unit tests to ensure we don't regress in our behavior.

As we previously didn't have unit tests for KMT python code, this PR also enables that. This PR depends on dda 0.20.0 being available in CI, as it includes this PR https://github.com/DataDog/datadog-agent-dev/pull/136 that removes a conflictive python dependency (`libvirt-python`).

### Motivation

Make it easier to select distros.

Unit tests added.


mbertrone (approved):

.dda/version
```diff
@@ -1 +1 @@
-0.10.1
+0.20.0
```
chouetz: Do you really need to do this? 
I think bump of `dda` should be done by us as it could create unexpected behaviour on every `datadog-agent` contributor workspace.

tasks/unit_tests/kmt_tests.py
```diff
@@ -0,0 +1,46 @@
...
+                version,
+                plats[cast("KMTArchName", arch)],
+                f"{name} maps to {version} which is not a valid version for architecture {arch}",
+            )
+
+    def test_normalize_vm_def__returns_expected_values(self):
+        possible = vmconfig.list_possible()
+
+        self.assertEqual(
+            vmconfig.normalize_vm_def(possible, "ubuntu22-arm64-distro"), ("distro", "ubuntu_22.04", "arm64")
```
chouetz: Should we have some tests with `{os_id}{version}` or `{os_id}-{version}`? For the new lines you added in `get_distribution_mapping`

chouetz (approved):

.dda/version
gjulianm: Not 100% as it's for the tests only. Version 0.20 removes the need to install `libvirt-python` for KMT dependencies. Without this bump, we can't add the tests. However, we can wait for you to bump `dda` first and then we can merge this.

tasks/unit_tests/kmt_tests.py
gjulianm: Added two extra cases

PR #36969, opened 2025-05-14
Use single include for child pipeline trigger

kacper-murzyn:
### What does this PR do?

It adds a intermediate yaml with jobs configuration for containers deployment. It seems that in gitlab the `trigger:` keyword can have only a single value..

### Motivation

Fix container deployment job.

To be verified as part of the gitlab pipeline.


Pythyu (approved):
LGTM, hopefully the pipeline as well 🤞

PR #36995, opened 2025-05-14
fix(installer): Create /var/log/datadog if it doesn't exist when installing the injector

BaptisteFoy:
### What does this PR do?
Creates /var/log/datadog if it doesn't exist when creating /var/log/datadog/dotnet in the injector setup

### DIFF
--- a/pkg/fleet/installer/packages/apminject/apm_inject.go
+++ b/pkg/fleet/installer/packages/apminject/apm_inject.go
@@ -19,13 +19,14 @@ import (
 	"regexp"
 	"strings"
 
+	"go.uber.org/multierr"
+	"gopkg.in/yaml.v3"
+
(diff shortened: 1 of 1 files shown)


arbll (approved):

pkg/fleet/installer/packages/apminject/apm_inject.go
```diff
@@ -94,7 +95,7 @@ func (a *InjectorInstaller) Setup(ctx context.Context) error {
 	}
 
 	// Create mandatory dirs
-	err = os.Mkdir("/var/log/datadog/dotnet", 0777)
+	err = os.MkdirAll("/var/log/datadog/dotnet", 0755)
```
knusbaum, line 98: I think in the agent this is created with user `dd-agent`. Can it cause problems if we don't re-own this?

PR #36996, opened 2025-05-14
Guard against displaying fips-proxy configuration in datadog-fips-agent status output

jeremy-hanna:
### What does this PR do?

The `datadog-fips-agent` isn't compatible for use with the `fips-proxy` and [we guard](https://github.com/DataDog/datadog-agent/blob/main/pkg/config/setup/config.go#L2186) against configuring it.

However we did not guard against displaying the configuration on the `agent status` output. This PR prevents the configuration that is not used from being displayed.

Manually with a build and toggling the configuration.


jeremy-hanna:
closing in favor of https://github.com/DataDog/datadog-agent/pull/36988

PR #37002, opened 2025-05-14
add unsubscription of a remote config listener

brycekahle:
### What does this PR do?

Returns a function from `Subscribe` that allows the caller to "unsubscribe" from updates. Since functions are not comparable, there was not a great way to add a separate `Unsubscribe` function, so this seemed like the ideal API.

### Motivation

system-probe will soon be obtaining BTF via remote config. Once we've received the BTF, we no longer need updates for our RC product.

### Additional Notes

There were a number of custom interfaces which only implemented part of the client API, so that is why this affects so many teams.

https://datadoghq.atlassian.net/browse/EBPF-746


ajgajg1134:
This looks OK from APM perspective, I do notice that you have the QA/Done label but no description in the PR as for what QA / testing was done here?

L3n41c (approved):

NouhaManai96 (approved):

brycekahle:
Closing because unsubscription has too many drawbacks at the moment.

PR #37005, opened 2025-05-14
[DEBUG-3885] dyninst/ir: straw-man pass at a di IR

ajwerner:
This commit introduces key concepts that'll be produced by combining config with a binary and then will be used to generate eBPF programs and to decode their output.

### What does this PR do?

This PR introduces the first stab at an IR to be used for the next stab at Go DI. 

### Motivation

As we begin to build out the next generation of Go DI, we need some shared concepts to couple the work. The `ir` layer intends to combine information from a specific binary with the set of probes from the configuration that apply to that binary. With an `ir.Program`, the system should be able to generate an eBPF program and decode data that that program produces into its ultimate output format without needing any further access to the binary or the config.

See [DEBUG-3885](https://datadoghq.atlassian.net/browse/DEBUG-3885) for internal details.

[DEBUG-3885]: https://datadoghq.atlassian.net/browse/DEBUG-3885?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -465,6 +465,7 @@
 /pkg/diagnose/ports/*windows*.go        @DataDog/windows-agent
 /pkg/eventmonitor/                      @DataDog/ebpf-platform @DataDog/agent-security
 /pkg/dynamicinstrumentation/            @DataDog/debugger-go
+/pkg/dyninst/                           @DataDog/debugger-go
 /pkg/flare/                             @DataDog/agent-configuration
 /pkg/flare/clusteragent/manifests.go    @DataDog/container-ecosystems @DataDog/agent-configuration
 /pkg/flare/*_win.go                     @Datadog/windows-agent @DataDog/agent-configuration
--- a/pkg/dyninst/doc.go
+++ b/pkg/dyninst/doc.go
@@ -1,0 +1,11 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+//go:build linux_bpf
+
+// Package dyninst contains the implementation of the dynamic instrumentation as
+// provided by the agent that supports such products as Live Debugger and
+// Dynamic Instrumentation for Go.
+package dyninst
--- a/pkg/dyninst/ir/doc.go
+++ b/pkg/dyninst/ir/doc.go
@@ -1,0 +1,14 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+//go:build linux_bpf
+
+// Package ir contains a representation of probes as applied
+// to a specific binary.
+//
+// The key data structure is ir.Program, which represents all the
+// information needed to describe the set of probes that will be
+// connected to a single binary.
+package ir
--- a/pkg/dyninst/ir/expression.go
+++ b/pkg/dyninst/ir/expression.go
@@ -1,0 +1,43 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+//go:build linux_bpf
+
+package ir
+
+// Expression is a typed and validated set of operations for compilation
+// and evaluation.
+type Expression struct {
+	// The type of the expression.
+	Type Type
+	// The operations that make up the expression, in reverse-polish notation.
+	Operations []Op
+}
+
+var (
+	_ Op = (*LocationOp)(nil)
+)
+
+// Op is an operation that can be performed on an expression.
+type Op interface {
+	irOp() // marker
+}
+
+// LocationOp references data of Size bytes
+// at Offset in Variable.
+type LocationOp struct {
+	// The variable that is referenced.
+	Variable *Variable
+
+	// The offset in bytes from the start of the variable to extract.
+	Offset uint32
+
+	// The size of the data to extract in bytes.
+	Size uint32
+}
+
+func (LocationOp) irOp() {}
+
+// TODO: Define expressions as needed for probe generation.
--- a/pkg/dyninst/ir/program.go
+++ b/pkg/dyninst/ir/program.go
@@ -1,0 +1,144 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+//go:build linux_bpf
+
+package ir
+
+import "github.com/DataDog/datadog-agent/pkg/network/go/bininspect"
+
+// ProgramID is a ID corresponding to an instance of a Program.  It is used to
+// identify messages from this program as they are communicated over the ring
+// buffer.
+type ProgramID uint32
+
+// TypeID is a ID corresponding to a type in a program.
+type TypeID uint32
+
+// EventID is a ID corresponding to an event output by the program.  It is used
+// to identify events as they are communicated over the ring buffer.
+type EventID uint32
+
+// Program defines the information needed to generate an eBPF program for a set
+// of probes as they apply to a specific binary. This structure is used to drive
+// both eBPF program generation and the interpretation of data that comes out of
+// the generated program.
+type Program struct {
+	// ID is the ID of the program. This is used to identify messages from this
+	// program as they are communicated over the ring buffer.
+	ID ProgramID
+	// Probes are the set of probes that comprise the program.
+	Probes []*Probe
+	// Subprograms are a list of subprograms that will be probed.
+	Subprograms []*Subprogram
+	// Types are the types that are used in the program.
+	Types []Type
+	// MaxTypeID is the maximum type ID that has been assigned.
+	MaxTypeID TypeID
+}
+
+// Subprogram represents a function or method in the program.
+type Subprogram struct {
+	// Name is the name of the subprogram.
+	Name string
+	// OutOfLinePCRanges are the ranges of PC values that will be probed for the
+	// out-of-line-instances of the subprogram. These are sorted by start PC.
+	//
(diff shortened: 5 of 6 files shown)


pkg/di/ir/program.go
```diff
@@ -0,0 +1,162 @@
...
+
+	// ID is the ID of the program. This is used to identify messages
+	// from this program as they are communicated over the ring buffer.
+	ID ProgramID
+
+	// Probes are the set of probes that comprise the program.
+	Probes []*Probe
+
+	// Subprograms are a list of subprograms that will be probed.
+	Subprograms []*Subprogram
```
grantseltzer, line 36: Do Probes always map 1:1 to subprograms? I see you have a reference to Probe in Subprogram but not the other way around.

pkg/di/ir/program.go
```diff
@@ -0,0 +1,162 @@
...
+	Tags []string
+
+	// The subprogram to which the probe is attached.
+	Subprogram *Subprogram
+
+	// The events that trigger the probe.
+	Events []Event
+
+	// Whether the probe should capture a snapshot of the state of the program.
+	Snapshot bool
```
grantseltzer, line 136: Even though snapshots use the same log intake, in my mind I consider them to be different from regular DI log probes and it might make more sense to have a `ProbeKindSnapshot`.

pkg/di/ir/program.go
```diff
@@ -0,0 +1,162 @@
...
+	ID string
+
+	// The kind of the probe.
+	Type ProbeKind
+
+	// The version of the probe.
+	Version int
+
+	// Tags that are passed through the probe.
+	Tags []string
```
grantseltzer, line 127: What's an example tag?

grantseltzer (commented):
I have other thoughts but will share them tomorrow

pkg/di/ir/program.go
ajwerner: There can be many probes configured that target the same Subprogram (read: function).  A `Probe` corresponds to exactly one `Subprogram`, unless I misunderstand how `Where` works in the API. A `Probe` here corresponds to the `Probe` in the API like [here](https://github.com/DataDog/dd-go/blob/9bbe4002a415fd613c13ac8c986dc04c86211b22/remote-config/pkg/products/livedebugging/domain.go#L73-L88). There's a reference `Probe` to `Subprogram` so it is exactly "the other way around".

pkg/di/ir/program.go
ajwerner: You tell me. It's in the API [here](https://github.com/DataDog/dd-go/blob/9bbe4002a415fd613c13ac8c986dc04c86211b22/remote-config/pkg/products/livedebugging/domain.go#L81). I have to imagine it's a way that the user or backend can ensure that the output log or metric or span carries extra tags.

pkg/di/ir/program.go
ajwerner: I'm open to that. There's definitely other choices for how to do the modeling. For now I kept it close to the config API.

ajwerner (commented):
Thanks for the review!

pkg/di/ir/expression.go
```diff
@@ -0,0 +1,22 @@
...
+package ir
+
+// Expression is a typed and validated set of operations for compilation
+// and evaluation.
+type Expression interface {
+	expression() // marker interface
+}
+
+// VariableExpr is an expression that references a variable directly.
+type VariableExpr struct {
```
piob-io, line 17: After we introduce path expression support, this will become obsolete. To reduce future changes, how about we immediately use following type instead:

type LocationExpr struct {
  Variable *Variable
  Offset uint32
  Size uint32
}

but then initially only support direct variables, where offset will be 0 and size will be whole?

pkg/di/ir/types.go
```diff
@@ -0,0 +1,322 @@
...
+// Type represents a an in-memory representation of a type in the target
+// program, or a synthetic type used by the probe to communicate information
+// from eBPF to the userspace.
+type Type interface {
+	// The ID of the type.
+	GetID() TypeID
+	// The name of the type.
+	GetName() string
+	// The size of the type in bytes.
+	GetSize() int
```
piob-io, line 23: I kinda liked the more verbatim GetByteSize(), as to me it more clearly points to this being about memory layout. Not a strong opinion though.

piob-io (approved):
Nice, I think the general structure is right, just some nits. Approving preemptively, relying on Grant and other experts on making sure we interface right with other datadog components.

ajwerner:
One thing going on in this PR is that it models a *set* of probes that apply to a *binary*. This model worked pretty well for side-eye because:
* you’d generally be collecting a bunch of correlated probes for a short period of time,
* it was clear which processes we were selecting for which config, and 
* they wanted to share common definitions regarding how to handle data types.

I think there’s still benefit to sharing data type logic, but with the live debugger model it has different considerations: 
1) Whenever any new probe is added we’ll recompile the entire program for that binary.
    - In practice given ~all the logic is identical it doesn’t affect the verification time, and my guess is that the cost of generation for the full set of probes is only marginally more expensive than the cost of generating any individual probe.
2) The probes that map to processes are applied asynchronously and not over a single channel: unless we do some explicit coalescing, which will be hard and lead to higher latency, we'll likely see new probe definitions pop in for processes on a node at different time.
3) We will need to swap over to a new program as they come in: what does this mean from a data continuity perspective?

An alternative but still simple model would be to eliminate the Program->Probe concept and independently compile an eBPF object per probe. This eliminates questions about updates (at least outside of the fact that probes themselves are versioned and mutable, but I think it's okay to have a discontinuity then). 

There's a hybrid model that's probably the "best", but is complicated. In an idealized world there'd be a single, static eBPF program that is driven through map data and then probes somehow dispatch to the appropriate vm code and lookup tables. Is it worth going right there?

ajwerner:
Another downside of having one abstraction per probe is that the decoding structures for each probe which are likely to be highly overlapping will be duplicated in memory.

I chatted offline with @piob-io, and the thing he recognized is that this model as proposed can fully represent a 1:1 mapping between probe and "Program" if the user wants it to. We also recognized some other things: this model also can represent the incremental changes from a set of probes to another so long as the mapping from probe to process is handled at another layer (and that seems like the right model).

It seems like if we do move to a static probe model that's map driven, then being able to atomically flip over from one program to another for a "universe" of probes seems good. The details on how dispatch from a pc to a stack machine entrypoint works is out of scope for the IR. Living in a world where we transition from one immutable IR state to the next seems much better than trying to model something mutable.

Now there is another important and open concept that isn't directly handled by this model is the desire to discover and handle more interface types. For now, that is representable -- we just need to consider this set of types as part of the config (not directly modeled now, but we can add it later), and then so long as we can atomically switch over to a new program, then it all just falls out.

piob-io:
Agreed

The way I would phrase it, the management of lifetime of each single function ebpf probe should be pushed down into this DI library, rather than be managed above. In particular, it's the only way to allow updates without losing continuity. But also it unlocks optimizations, like caching and sharing of many expensive, internal things. And the current multi probe model represents that.

I don't see it causing much more complexity even to the most basic, initial implementation. We want dwarf processor to be able to handle many functions, same for codegen and decoding. And it's really just about iterating over a list and having a lookup map.

pkg/di/ir/types.go
ajwerner: Done.

ajwerner (commented):
Thanks for the review!

pkg/di/ir/expression.go
```diff
@@ -0,0 +1,43 @@
...
+	irOp() // marker
+}
+
+// LocationOp references data of Size bytes
+// at Offset in Variable.
+type LocationOp struct {
+	// The variable that is referenced.
+	Variable *Variable
+
+	// The offset of the variable to extract.
```
piob-io, line 34: The offset within the variable? The offset in the variable location list?

pkg/di/ir/program.go
grantseltzer: Hm ok, Tags as far as I understand are used to specify instances of services to instrument. Let's sync with someone from another runtime to ensure we're using it correctly.

pkg/di/ir/program.go
```diff
@@ -0,0 +1,150 @@
...
+	IsParameter bool
+}
+
+// Location is the location of a parameter or variable in the subprogram.
+type Location struct {
+	// PCRange is the range of PC values that will be probed.
+	Range PCRange
+	// The locations of the pieces of the parameter or variable.
+	Location []bininspect.ParameterPiece
+}
```
grantseltzer, line 95: What do you think about making the `Locations` field of `Variable` a map of `PCRange` to `[]bininspect.ParameterPiece`?

I also think we should take a good look at what's provided by the `bininspect` package, think what else we need from DWARF, and probably get rid of the dependency.

ajwerner: > What do you think about making the `Locations` field of `Variable` a map of `PCRange` to `[]bininspect.ParameterPiece`?

When you say "a map of", do you mean `map[PCRange][]bininspect.ParameterPiece`? That's not a good choice here because you're never going to have your hands on the range itself. Instead you're going to be searching for a PC that falls in that range or to intersect that range with another range. Tools like `sort.Search` fit nicely here, and a hash map doesn't. The other alternative would be a BTree or sorted structure, but I don't think it is warranted given the data isn't changing, so having a sorted slice is a good 

> I also think we should take a good look at what's provided by the `bininspect` package, think what else we need from DWARF, and probably get rid of the dependency.

Yeah, agreed. I was reticent to put this here. It's sort of a placeholder in my mind because I'm not yet sure what the DWARF analysis will look like.

pkg/di/ir/program.go
ajwerner: My sense is that it's these queries in Enablement that choose the services: https://github.com/DataDog/dd-go/blob/beeaadef389190e4e6bbabf82059aa0c91eba770/remote-config/apps/rc-api/products/livedebugging/jsonapiconf/domain.go#L63-L73

I think these are tags that should come out the other side. Here's an example I found in the real config:

```
  "tags": [
    "ui_session_id:<redacted>",
    "session_name:<redacted>'s session May 12, 2:07 pm"
  ],
  ```
  
  Note that I did the redacting.

pgimalac (approved):

pkg/di/ir/program.go
grantseltzer: Got it, thanks for digging into it

PR #37012, opened 2025-05-15
chore(e2e): Support using relative path in local packages

KevinFairise2:
### What does this PR do?

`ddagent:localPackage` expects an absolute path, if the user provide a relative path, convert it to an absolute path

### Motivation

Make the local-package feature more usable

### DIFF
--- a/tasks/new_e2e_tests.py
+++ b/tasks/new_e2e_tests.py
@@ -124,7 +124,7 @@ def run(
         parsed_params[parts[0]] = parts[1]
 
     if local_package:
-        parsed_params["ddagent:localPackage"] = local_package
+        parsed_params["ddagent:localPackage"] = Path(local_package).absolute().as_posix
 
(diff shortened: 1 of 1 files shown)


tasks/new_e2e_tests.py
```diff
@@ -124,7 +124,7 @@ def run(
         parsed_params[parts[0]] = parts[1]
 
     if local_package:
-        parsed_params["ddagent:localPackage"] = local_package
+        parsed_params["ddagent:localPackage"] = Path(local_package).absolute().as_posix
```
CelianR, line 127:
```suggestion
        parsed_params["ddagent:localPackage"] = Path(local_package).absolute().as_posix()
```
Isn't it a function?

CelianR (approved):

amenasria (approved):
LGTM

PR #37033, opened 2025-05-15
[NPM-4421] Add config to randomize SYN packet ID

pimlu:
### What does this PR do?
Adds a (not yet supported) configuration option to enable TCP SYN compatibility mode, where we will make packets as similar to `tcptraceroute` as possible.
### Motivation
There have been a bunch of cases where our SYN packets are seemingly getting dropped.

New config test should pass

The process of adding a config to traceroutes is currently labor-intensive. It involves changing code in a bunch of places. If we add more of these, we should coalesce it into a single struct that gets copied around.


hmahmood (approved):

dustmop (approved):
LGTM for agent-configuration

PR #37046, opened 2025-05-15
[USMON-1432]  fetch v13 classification

guymor4:
### What does this PR do?
Support kafka fetch versions in classification v13+ (was 12, now 17)
- Use topic id instead of topic name and verifies UUID is gen 4


pkg/network/ebpf/c/protocols/kafka/kafka-classification.h
```diff
@@ -268,6 +268,44 @@ static __always_inline bool validate_first_topic_name(pktbuf_t pkt, bool flexibl
...
+        }
+    } else {
+        offset += sizeof(s32);
+    }
+
+    if (offset + sizeof(topic_id) > pktbuf_data_end(pkt)) {
+        return false;
+    }
+
+    pktbuf_load_bytes(pkt, offset, topic_id, sizeof(topic_id));
```
guyarb, line 291: we should be using the `_with_telemetry` if that's possible

pkg/network/ebpf/c/protocols/kafka/kafka-classification.h
```diff
@@ -268,6 +268,44 @@ static __always_inline bool validate_first_topic_name(pktbuf_t pkt, bool flexibl
...
+    offset += sizeof(topic_id);
+
+    // The UUID version (13th digit)
+    if (topic_id[6] >> 4 != 0x4) {
```
guyarb, line 295: perhaps it will be clearer to do `topic_id[6] && 0x40 == 0x40`

pkg/network/ebpf/c/protocols/kafka/kafka-classification.h
guymor4: Should we always call the `_with_telemetry` variation? because we have other places in this file without it

guyarb: yes we should
In the past, we had places in which we had to avoid using `_with_telemetry` as we hit the instruction limit (or complexity limit). As long as we can use `_with_telemetry` we should

pkg/network/ebpf/c/protocols/kafka/kafka-classification.h
guyarb: same for the other condition

pkg/network/ebpf/c/protocols/kafka/kafka-classification.h
```diff
@@ -268,6 +268,44 @@ static __always_inline bool validate_first_topic_name(pktbuf_t pkt, bool flexibl
...
+    offset += sizeof(topic_id);
+
+    // The UUID version (13th digit 4 MSB) must be 4
+    if ((topic_id[6] & 0x40) != 0x40) {
+        // The UUID version is not 4
+        return false;
+    }
+
+    // The UUID variant (17th digit) may be 0x8, 0x9, 0xA or 0xB
+    if ((topic_id[8] >> 4) != 0x8 && (topic_id[8] >> 4) != 0x9 && (topic_id[8] >> 4) != 0xA && (topic_id[8] >> 4) != 0xB) {
```
guyarb, line 301:
```suggestion
    _u8 marker = topic_id[8] >> 4;
    return 0x8 <= marker && marker <= 0xb;
```

pkg/network/ebpf/c/protocols/kafka/kafka-classification.h
guymor4: I actually rolled it back, I like the `>> 4` syntax better as I can just do `__u8 uuid_version = topic_id[6] >> 4;` and use it like a normal number so `uuid_version == 4`
It feels safer because you don't have to remember anything for checks later

guyarb (approved):
LGTM
Can you re-deploy it on staging and verify it is working and detecting the protocols?

guymor4:
Posting validation checks:
Deployed agent version to resource-writer host, started getting updated fetch traffic:
image

Also classified as "kafka" in NPM:
image

Also deployed agent version to `oddish-c`
Kafka fetch v16 reports match deployment of agent:
image

PR #37049, opened 2025-05-15
APM: Add initial support for new v1.0 traces spec

ajgajg1134:
### What does this PR do?
Introduce support for the new "Efficient Trace Payload", this accepts v1.0 traces, keeping them in this compressed format where strings are deduplicated and serializing them out as a new field to intake.

(For DD internal: https://docs.google.com/document/d/1hNS6anKYutOYW-nmR759UlKXUdT6H0mRwVt7_L70ESc)

### Motivation
* Reduce serialization and deserialization costs via string dictionary compression
* Reduce memory pressure on Trace-Agent via same reduced string duplication
* Add initial support for "rich attributes" beyond the old "meta" / "metrics" model
* Promote certain very common fields to top level fields as appropriate
* Move certain fields to appropriate structure level, e.g. TraceID is now a chunk level field as a chunk must only contain spans from the same trace

See the linked RFC for a much more thorough review of the changes / motivations as well as various POC benchmarks to motivate these improvements.

This change is major, easily the largest change I've made in the trace-agent ever and thus difficult to validate. The review process here will be fairly unique as it was decided to avoid splitting functionality across many PRs. There will be a significant amount of in person reviews / Q&A to thoroughly vet these changes. 
**However**, this change also attempts to be as transparent as possible for payloads that are not sent via the v1.0 endpoint. This significantly de-risks this change at the cost of increased implementation complexity and duplicated code. Where de-duplication was deemed reasonable (e.g. obfuscation) I made modifications to avoid code duplication, I am open to further opportunities to do that here however I also want to avoid any indirection costs that could be introduced on the existing processing pipelines.

There are a significant number of unit tests that were ported over alongside the implementation, There are also some new integration tests. Further testing will be done internally in the staging environment using an implementation of v1.0 from the Go tracer. 

* This is a big change, There are risks that we miss a breaking change here. 
* This introduces some code duplication in the processing of traces, this was the choice made to reduce risk for non-v1.0 clients (read: all of them at the moment), however this creates tech debt. The long term goal is to push adoption for v1.0 across the tracer libraries to the point where we can remove this duplication and instead use an adapter to convert incoming payloads to v1.0.
* The performance in "real world" conditions is as-of-yet unknown, we will need to verify performance and use profiles to optimize these new code paths to ensure we get the benefits expected from v1.0


ajgajg1134:
@Cursor review

releasenotes/notes/efficient-trace-payload-c22e2e03e8689a03.yaml
```diff
@@ -0,0 +1,12 @@
...
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+
+features:
+  - |
+    APM: A more efficient trace payload encoding via the /v1.0/traces endpoint has been added.
```
buraizu:
```suggestion
    APM: A more efficient trace payload encoding through the /v1.0/traces endpoint has been added.
```

buraizu (approved):
Approving with a minor update requested

pkg/trace/event/extractor_metric.go
```diff
@@ -41,3 +42,21 @@ func (e *metricBasedExtractor) Extract(s *pb.Span, priority sampler.SamplingPrio
 	}
 	return extractionRate, true
 }
+
+// Extract decides whether to extract APM events from a span based on the value of the event extraction rate metric set
```
ichinaski:
```suggestion
// ExtractV1 decides whether to extract APM events from a span based on the value of the event extraction rate metric set
```

pkg/trace/event/extractor_legacy.go
```diff
@@ -44,3 +45,18 @@ func (e *legacyExtractor) Extract(s *pb.Span, _ sampler.SamplingPriority) (float
 	}
 	return extractionRate, true
 }
+
+// Extract decides to extract an apm event from the provided span if there's an extraction rate configured for that
```
ichinaski:
```suggestion
// ExtractV1 decides to extract an apm event from the provided span if there's an extraction rate configured for that
```

pkg/trace/event/extractor_fixed_rate.go
```diff
@@ -52,3 +53,23 @@ func (e *fixedRateExtractor) Extract(s *pb.Span, priority sampler.SamplingPriori
 	}
 	return extractionRate, true
 }
+
+// Extract decides to extract an apm event from a span if its service and name have a corresponding extraction rate
```
ichinaski:
```suggestion
// ExtractV1 decides to extract an apm event from a span if its service and name have a corresponding extraction rate
```

ichinaski (approved):
Minor comment edits, but overall LGTM. I think this is ready to be tested on a staging cluster 🚀

hush-hush (approved):

KevinFairise2 (approved):

duncanista (approved):

PR #37052, opened 2025-05-15
[NPM-4421] Add packet ID to matching logic, use it for compatibility mode

pimlu:
Stacked on [\[NPM-4421\] Add config to randomize SYN packet ID](https://github.com/DataDog/datadog-agent/pull/37033)

### What does this PR do?
This PR implements `tcp_syn_compatibility_mode`. In this mode, TCP SYN traceroutes attempt to mimic the `tcptraceroute` tool as much as possible.

### Motivation
Should address the issues people have been seeing with dropped TCP SYN packets.

I tested with compatibility mode enabled/disabled with static paths on Linux, and on Windows. I also tested UDP. I checked in wireshark, that it marks packets as retransmissions in compatibility mode, not reused ports.

Existing test suite should pass with updated tests:
```
sudo go test -timeout 30s -tags linux,linux_bpf,npm,process,test github.com/DataDog/datadog-agent/pkg/networkpath/...
```
I also ran the suite on Windows, and all tests passed (except for one that depends on `TCP_FAIL_CONNECT_ON_ICMP_ERROR` which my Windows version doesn't support):
```
go test -timeout 30s -tags windows,npm,process,test github.com/DataDog/datadog-agent/pkg/networkpath/...
```

The packet matching interface is also used by UDP meaning this also changes UDP code.

### DIFF
--- a/pkg/networkpath/traceroute/tcp/parser.go
+++ b/pkg/networkpath/traceroute/tcp/parser.go
@@ -79,7 +79,7 @@ func (tp *parser) parseTCP(header *ipv4.Header, payload []byte) (*tcpResponse, e
 // MatchTCP parses a TCP packet from a header and packet bytes and compares the information
 // contained in the packet to what's expected and returns the source IP of the incoming packet
 // if it's successful or a MismatchError if the packet can be read but doesn't match
-func (tp *parser) MatchTCP(header *ipv4.Header, packet []byte, localIP net.IP, localPort uint16, remoteIP net.IP, remotePort uint16, seqNum uint32, _packetID uint16) (net.IP, error) {
+func (tp *parser) MatchTCP(header *ipv4.Header, packet []byte, localIP net.IP, localPort uint16, remoteIP net.IP, remotePort uint16, seqNum uint32, _ uint16) (net.IP, error) {
 	if header.Protocol != 6 { // TCP
 		return net.IP{}, errors.New("expected a TCP packet")
 	}
--- a/pkg/networkpath/traceroute/tcp/parser.go
+++ b/pkg/networkpath/traceroute/tcp/parser.go
@@ -79,7 +79,7 @@ func (tp *parser) parseTCP(header *ipv4.Header, payload []byte) (*tcpResponse, e
 // MatchTCP parses a TCP packet from a header and packet bytes and compares the information
 // contained in the packet to what's expected and returns the source IP of the incoming packet
 // if it's successful or a MismatchError if the packet can be read but doesn't match
-func (tp *parser) MatchTCP(header *ipv4.Header, packet []byte, localIP net.IP, localPort uint16, remoteIP net.IP, remotePort uint16, seqNum uint32) (net.IP, error) {
+func (tp *parser) MatchTCP(header *ipv4.Header, packet []byte, localIP net.IP, localPort uint16, remoteIP net.IP, remotePort uint16, seqNum uint32, _packetID uint16) (net.IP, error) {
 	if header.Protocol != 6 { // TCP
 		return net.IP{}, errors.New("expected a TCP packet")
 	}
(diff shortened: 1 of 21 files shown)


pkg/networkpath/traceroute/tcp/parser.go
```diff
@@ -79,7 +79,7 @@ func (tp *parser) parseTCP(header *ipv4.Header, payload []byte) (*tcpResponse, e
 // MatchTCP parses a TCP packet from a header and packet bytes and compares the information
 // contained in the packet to what's expected and returns the source IP of the incoming packet
 // if it's successful or a MismatchError if the packet can be read but doesn't match
-func (tp *parser) MatchTCP(header *ipv4.Header, packet []byte, localIP net.IP, localPort uint16, remoteIP net.IP, remotePort uint16, seqNum uint32) (net.IP, error) {
+func (tp *parser) MatchTCP(header *ipv4.Header, packet []byte, localIP net.IP, localPort uint16, remoteIP net.IP, remotePort uint16, seqNum uint32, _packetID uint16) (net.IP, error) {
```
hmahmood, line 82: You can just use `_` here instead of `_packetID`.

pkg/networkpath/traceroute/tcp/tcpv4_unix.go
```diff
@@ -100,9 +97,20 @@ func (t *TCPv4) TracerouteSequential() (*common.Results, error) {
 	// hops should be of length # of hops
 	hops := make([]*common.Hop, 0, t.MaxTTL-t.MinTTL)
 
+	var packetID uint16
+	var seqNumber uint32
+	if t.CompatibilityMode {
+		seqNumber = rand.Uint32()
+	} else {
+		packetID = uint16(41821)
```
hmahmood, line 105: Why not a random id?

pkg/networkpath/traceroute/tcp/tcpv4_unix.go
```diff
@@ -100,9 +97,20 @@ func (t *TCPv4) TracerouteSequential() (*common.Results, error) {
...
+	}
 	for i := int(t.MinTTL); i <= int(t.MaxTTL); i++ {
-		seqNumber := rand.Uint32()
-		hop, err := t.sendAndReceive(rawIcmpConn, rawTCPConn, i, seqNumber, t.Timeout)
+		if t.CompatibilityMode {
+			packetID = uint16(rand.Uint32())
+		} else {
+			seqNumber = rand.Uint32()
+		}
```
hmahmood, line 112: You could make a small function like
```go
func (t *TCPv4) randSeqOrPacketID() (packetID uint16, seqNum uint32) {
    	if t.CompatibilityMode {
		return 0, rand.Uint32()
	}

        return rand.Uint16(), 0
}
```

pkg/networkpath/traceroute/tcp/tcpv4_unix.go
pimlu: I'm in favor of always randomizing the ID personally, which is what `tcptraceroute` does -- in [this thread](https://dd.slack.com/archives/C07CDMHLSUX/p1746484495032929) I proposed to do that. In that case, we don't need compatibility mode since it's "always" in compatibility mode. Alex thinks randomizing the ID could have load balancing implications though, which is why I'm making it a configurable switch.

pimlu: Also as for why it's always been `41821`, there's no reason I know of other than picking an arbitrary value that's nonzero (just in case middleboxes incorrectly drop traffic with 0 IDs)

hmahmood: But its not random in the loop below?

pimlu: Yeah, I think it would make sense to set the packet ID to a random per-traceroute value instead of `41821`. The reason I don't randomize it though is, I wanted this PR to not change anything unless you turn on compatibility mode. The existing traceroute logic uses an ID of `41821`

pkg/networkpath/traceroute/tcp/tcpv4_unix.go
pimlu: Updated

pkg/networkpath/traceroute/tcp/parser.go
pimlu: Updated

hmahmood (approved):

PR #37055, opened 2025-05-15
Properly handle slices in yaml config data

dustmop:
### What does this PR do?

Properly handle slices in yaml config data

### Motivation

A bug in the nodetreemodel config builder didn't handle slices properly, causing multiple unit test failures when it was enabled.

When run with `DD_CONF_NODETREEMODEL=enable`, this fixes the test `TestOptionalModule` in comp/core/configsync/configsyncimpl and `TestUnmarshalKeyTrapsConfig` in pkg/config/structure.

New unit tests in nodetreemodel that validate this behavior.

### DIFF
--- a/pkg/config/structure/unmarshal_test.go
+++ b/pkg/config/structure/unmarshal_test.go
@@ -108,7 +108,7 @@ user:
 `
 	mockConfig = newConfigFromYaml(t, confYaml)
 	err = unmarshalKeyReflection(mockConfig, "user", &person)
-	assert.ErrorContains(t, err, `at [jobs]: []T required, but input is not an array: &{30 file} of *nodetreemodel.leafNodeImpl`)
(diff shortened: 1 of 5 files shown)


pkg/config/structure/unmarshal_test.go
```diff
@@ -108,7 +108,7 @@ user:
 `
 	mockConfig = newConfigFromYaml(t, confYaml)
 	err = unmarshalKeyReflection(mockConfig, "user", &person)
-	assert.ErrorContains(t, err, `at [jobs]: []T required, but input is not an array: &{30 file} of *nodetreemodel.leafNodeImpl`)
+	assert.ErrorContains(t, err, `at [jobs]: []T required, but input is not an array: &{30`)
```
dustmop, line 111: These tests aren't great, since they expose inner details of the implementation. They should be fixed in a follow-up PR.

rahulkaukuntla (approved):
lgtm!

PR #37056, opened 2025-05-15
Create flare from autoscaler output

jennchenn:
### What does this PR do?

Add autoscaling store to agent flare output

### Motivation

Help with debugging issues with autoscaling

1. Deploy changes
2. Exec into cluster agent pod and run `agent flare`
3. Verify that output not has an `autoscaler-list.json` file that contains the contents of the autoscaling store

Currently flare is outputted in json format - because of the way errors are handled some of the error fields are duplicated. If we componentize / create a flare provider in autoscaling, we can use the same formatting used for the CLI to format the output


vboulineau (approved):

davidor (approved):

pkg/flare/archive_kubeapiserver.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"encoding/json"
+	"io"
+
+	apiutil "github.com/DataDog/datadog-agent/pkg/api/util"
+	autoscalingWorkload "github.com/DataDog/datadog-agent/pkg/clusteragent/autoscaling/workload"
+)
+
+// GetAutoscalerList fetches the autoscaler list from the given URL.
+func GetAutoscalerList(remoteURL string) ([]byte, error) {
```
hush-hush, line 19: Could we move this as a flare provider in a component ? Maybe `comp/autoscaling` ?

The goal long term is to migrate all the content from `pkg/flare` in the comp where they belong so we can include the flare mechanism in any binary without pulling a tone of dependencies.

pkg/flare/archive_kubeapiserver.go
```diff
@@ -0,0 +1,39 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	fct := func(w io.Writer) error {
+		autoscalerDump.Print(w)
+		return nil
+	}
+	return functionOutputToBytes(fct), nil
```
hush-hush, line 38: Why do we unmarshal to print as text right after ? Could we not simply save the HTTP answer into a file ?

That would also remove the dependency on `autoscalingWorkload`.

pkg/flare/clusteragent/archive_dca.go
```diff
@@ -225,6 +226,31 @@ func getLocalClusterAgentDiagnose(fb flaretypes.FlareBuilder, diagnose diagnose.
 	return fb.AddFile("diagnose.log", bytes)
 }
 
+func getDCAAutoscalerList() ([]byte, error) {
```
hush-hush, line 229: Could we add a unit test for this ?

hush-hush (commented):
Added one last comment, but looks good otherwise.

hush-hush (approved):

PR #37078, opened 2025-05-16
(fleet) migrate the prerm to the installer 1/n

arbll:
This PR starts the migration of the `prerm` script to the installer.

- In this first PR we are only migrating the service management to the installer
- To align the execution order of RPM and DEB we have to use the same strategy as the integration python scripts:
    - The "upgrade" `prerm` is run from the new package %pre
    - The "remove" `prerm` is run from the `prerm` as expected


arbll:

BaptisteFoy (approved):

PR #37087, opened 2025-05-16
[dyninst/ebpf] Port sideeye ebpf program implementation and compatible go framing

piob-io:
### What does this PR do?
Introduce ebpf program implementation ported from side-eye.
Strip parts that won't be relevant for foreseeable future (thread profile support, collecting data from multiple frames, etc).
Minor code cleanup, definition reorganiation.
Introduce framing.go for unframing ebpf program output data, with compatibility test.

### Motivation


pkg/dyninst/ebpf/framing.h
```diff
@@ -0,0 +1,60 @@
...
+
+// The header of a data item.
+typedef struct data_item_header {
+  // The type of the data item.
+  uint32_t type;
+  // The length of the data item.
+  uint32_t length;
+  // The address of the data item in the user process's address space.
+  uint64_t address;
+} __attribute__((aligned(8))) data_item_header_t;
```
grantseltzer, line 58: Is the purpose of type to reference the types full definition in the saved probe IR?

For non-basic types, will the length be the whole type (i.e. structs will be the sum of all fields)? 

Also what's the intention of address?

piob-io: Yes, exactly, type references types in ir, length is the total size of this data item in the output buffer, which must be equal to the byte_size() property of the type in ir (structs will be the sum of their field lenghts),

When some structure contains a pointer, data item for that structure will contain just the value of that pointer. We will emit another data item for the object that pointer was pointing too. Address can then be used to figure out that this another data item is what the pointer in the first data item pointed to.

pkg/dyninst/ebpf/framing.h
```diff
@@ -0,0 +1,60 @@
...
+
+  // The timestamp of the event according to CLOCK_MONOTONIC.
+  uint64_t ktime_ns;
+}
+// Use aligned attribute to ensure that the size of the structure is a multiple
+// of 8 bytes; the attribute leads to the compiler adding padding.
+__attribute((aligned(8))) event_header_t;
+
+// The maximum number of pcs in a captured stack trace.
+#define STACK_DEPTH 511
```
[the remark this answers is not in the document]
piob-io: So that whole structure, including the length field, is 4KiB.

pkg/dyninst/ebpf/framing.h
```diff
@@ -0,0 +1,60 @@
...
+
+  // The number of bytes for a stack trace that follows this header.
+  uint16_t stack_bytes;
+ 
+  // Bytes of the frame pointer from the stack root.
+  uint32_t frame_byte_depth;
+ 
+  // If the stack_bytes field is 0, then this field is the program counter of
+  // at the moment of the event.
+  uint64_t stack_hash_or_pc;
```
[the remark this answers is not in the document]
piob-io: Simplified and improved comment (it's a hash of the stack pcs that follow this entry).

pkg/dyninst/ebpf/framing.h
```diff
@@ -0,0 +1,60 @@
...
+  uint32_t data_byte_len;
+
+  // The ID of the program that produced this event.
+  uint32_t prog_id;
+
+  // The number of bytes for a stack trace that follows this header.
+  uint16_t stack_bytes;
+ 
+  // Bytes of the frame pointer from the stack root.
+  uint32_t frame_byte_depth;
```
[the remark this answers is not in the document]
piob-io: Removed, this is obsolete

pkg/dyninst/ebpf/framing.h
```diff
@@ -0,0 +1,60 @@
...
+  // own size, but for the snapshot header, the size of the header is included.
+  // Must be the first field in the structure, both ebpf and decoding components
+  // assume it is.
+  uint32_t data_byte_len;
+
+  // The ID of the program that produced this event.
+  uint32_t prog_id;
+
+  // The number of bytes for a stack trace that follows this header.
+  uint16_t stack_bytes;
```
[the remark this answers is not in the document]
piob-io: Done

pkg/dyninst/ebpf/debug.h
```diff
@@ -0,0 +1,32 @@
...
+
+#include "bpf_helpers.h"
+
+// DEBUG enables debug logging and controls the level. It is defined during
+// compilation.
+#ifdef DEBUG
+// LOG is a macro that prints a message if the level is less than or equal to
+// the DEBUG level.
+#define LOG(level, fmt, ...)                                                   \
+  if (level <= DEBUG) {                                                        \
```
brycekahle, line 12: I don't think `DEBUG` can both be a define and an integer level?

pkg/dyninst/ebpf/debug.h
```diff
@@ -0,0 +1,32 @@
...
+    bpf_printk(fmt, ##__VA_ARGS__);                                            \
```
brycekahle, line 13: If you use `log_debug` we will automatically handle the newline change that was made in 5.9, as long as the program patcher is included on the userspace side.

pkg/dyninst/ebpf/event.c
```diff
@@ -0,0 +1,108 @@
...
+    return 0;
+  }
+  *header = (event_header_t){
+      .data_byte_len = sizeof(event_header_t),
+      .stack_byte_len = 0, // set this if we collect stacks
+      .ktime_ns = start,
+  };
+
+  __attribute__((unused)) int process_steps = 0;
+  __attribute__((unused)) int chase_steps = 0;
```
brycekahle, line 53:
```suggestion
  __maybe_unused int process_steps = 0;
  __maybe_unused int chase_steps = 0;
```

pkg/dyninst/ebpf/queue.h
```diff
@@ -0,0 +1,80 @@
+#ifndef __QUEUE_H__
+#define __QUEUE_H__
+
+#include "bpf_helpers.h"
+#include "vmlinux.h"
+
+#define DEFINE_QUEUE(prefix, elem_t, max_length)                               \
```
brycekahle, line 7: How does this compare to `BPF_MAP_TYPE_QUEUE`?

pkg/dyninst/output/framing.go
```diff
@@ -0,0 +1,28 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+//go:build linux_bpf
+
+package output
+
+// This file must be kept in sync with the ../ebpf/framing.h file.
```
brycekahle, line 10: We have a way to handle this by auto-generating the corresponding Go type definitions. See https://github.com/DataDog/datadog-agent/blob/main/pkg/collector/corechecks/ebpf/probe/ebpfcheck/c_types.go (source) and https://github.com/DataDog/datadog-agent/blob/main/pkg/collector/corechecks/ebpf/probe/ebpfcheck/c_types_linux.go (generated) for an example. You will need to add the file to https://github.com/DataDog/datadog-agent/blob/175931d13d26fbde0c42cfb70e2269a14fc447f4/tasks/system_probe.py#L487

pkg/dyninst/ebpf/debug.h
piob-io: Maybe I misunderstand - preprocessor tracks defines for `#ifdef` evaluation independent of whether they map to a value or nothing, both `#define X` and `#define X 1` satisfying `#ifdef X`.

pkg/dyninst/ebpf/debug.h
piob-io: Is there an interest in supporting version of log_debug with verbosity levels? Maybe our debugging experience will be different here, but it was quite useful to us before. We actually want to take it a step further, having pairs of (tag, level) so that we can control verbosity independently for different components.

pkg/dyninst/ebpf/event.c
piob-io: Done

pkg/dyninst/ebpf/queue.h
piob-io: The reason we have this is that we need it to be per-cpu, for correctness. It is used by each probe to independently run BFS to "chase" pointers, to capture data behind them, following as many chains of pointers as needed.

As for performance, depending on usage, this will do better, as it is using per-cpu array and ops just do a direct memory access, while built-in ebpf queue synchronizes with a spinlock. I have not measured queues specifically, but we do have some data on other structures, I just shared it here: https://datadoghq.atlassian.net/wiki/x/XgCsMAE

pkg/dyninst/output/framing.go
piob-io: This is awesome, thank you!

pkg/dyninst/ebpf/debug.h
brycekahle: Right now, the object files we load by default have zero `bpf_printk` logging included (excluded during compilation) for performance reasons. We do also ship debug versions, but you have to explicitly configure them to be used. I don't think it is used, except for in a small number of support cases.

I know @gjulianm has expressed interest in a tag+level control of debug logging. Any change here would need to be carefully considered, because even internally the performance of the ebpf programs can have significant effects on the host.

pkg/dyninst/ebpf/debug.h
brycekahle: Currently `DEBUG` is either defined or defined as `1`. It seems this code is expecting `DEBUG` to be a verbosity level. That seems a bit unintuitive to me.

piob-io: Ah, DEBUG already has a specific meaning in dd ebpf land :) I will follow up on debug logging separately, for now just disabled all the logging we have.

pkg/dyninst/ebpf/debug.h
piob-io: Oh, I see, gtk

pkg/dyninst/ebpf/queue.h
piob-io: Likely it would be fine. The logical correctness is critical. Performance less so. At side-eye we supported capturing thread profiles, potentially with lots of interesting pointers across all frames across all threads' stacks (and pointers referenced by pointers). For now go di will support directly probing function call/lines, with maybe a few pointers to follow.

That being said, is there something concerning about this queue implementation?

piob-io:
@brycekahle @grantseltzer Thank you for review, any more comments?

ajwerner (approved):
LGTM even though we know things are going to move around a bit.

brycekahle (approved):

grantseltzer (approved):

PR #37088, opened 2025-05-16
[ AGNTLOG-210 ] Extend the logs agent to support a cardinality level config for the tagger

ddrthall:


ddrthall:
Feature is not currently useful enough for the logs agent to implement, as high cardinality tags are much more valuable for logs than for metrics.

PR #37089, opened 2025-05-16
[DJM] Fetch Databricks custom tags from Databricks API and inject them to agent config

RickyMarou:
### What does this PR do?
- [Jira](https://datadoghq.atlassian.net/browse/DJM-832)

This PR adds a new functionality during the agent setup for [Databricks](https://www.databricks.com/) hosts.
We check for the presence of two environment variables that allow us to authenticate with the Databricks API (see https://github.com/DataDog/dogweb/pull/142202), we use these to make HTTP request to the API, fetch tags, and add it to the agent config.

### Motivation

A key technology we support in the Data Jobs Monitoring product is Databricks.
When customers organize their workload on the Databricks platform, they heavily rely on custom tags that they can add through the Databricks UI or API. This is for an example how they can split the workload by teams, cost center, etc...

It's capital that we add these custom tags to the metrics we ingest in DD, as all cost estimates for their usage of Databricks and cloud compute is based of those (mainly CPU metrics). They are also what allow us to make recommendations.

We've explored many ways in which we can attach those custom tags to the metrics, and in the interest of time and KISS, we are moving forward with this method. You can read more about all of this in [this doc](https://docs.google.com/document/d/1JBg1fMHN7glxiaompmU3uDLq5D9P7wk42d9jT5EiRqk/edit?tab=t.0#heading=h.60wy60sbn9fn).

I used the agent installer generated by the pipeline on this repo in a global init script, and checked that that:
- If I don't provide the environment variables, the agent is starting correctly with the same config than the one we would have using the currently released installer 
- If I provide environment variables, cluster custom tags + job custom tags are correctly added to the config

I used this [test dashboard](https://ddstaging.datadoghq.com/dashboard/hya-nmv-5zz/marwan---debug-databricks-custom-tags?fromUser=true&refresh_mode=sliding&from_ts=1747733215026&to_ts=1747819615026&live=true) to monitor that the metrics are correctly ingested and also ssh'd into clusters to look at the agent config to check that the config was working correctly. 

The tags are not normalised before we set them in the config. I tried _REALLY_ hard to do this. You can see the last iteration of my attempt on [this commit](https://github.com/DataDog/datadog-agent/commit/d8f5b48ab1f572db127829304223cfa9c79c317d). After few hours of fighting CI checks I just give up, let's just rely on the intake to do the normalisation 😞 

### Checklist

- [x] Write tests for the code you wrote.
- [x] Preferably make sure that all tests pass locally.
  ```sh
   ~/go/src/github.com/DataDog/datadog-agent marwan/djm-databricks-metrics-tags
   ❯ go1.23.9 test github.com/DataDog/datadog-agent/pkg/fleet/installer/setup/djm

   ok      github.com/DataDog/datadog-agent/pkg/fleet/installer/setup/djm  0.338s
  ```
- [X] Summarize your PR with an explanatory title and a message describing your changes, cross-referencing any related bugs/PRs.
- [x] Use [Reno](#reno) to create a release note.
  - Initially provided a changelog but after looking into this more, it appears we usually don't put change logs for install script updates 
- [X] Open your PR against the `main` branch.
- [X] Sign the Contributor Licence Agreement.
- [X] Sign your commits.
- [X] Provide adequate QA/testing plan information.

### DIFF
--- a/pkg/fleet/installer/setup/djm/databricks.go
+++ b/pkg/fleet/installer/setup/djm/databricks.go
@@ -309,7 +309,8 @@ func loadLogProcessingRules(s *common.Setup) {
 
 // ClusterTags are custom tags from Databricks API response
 type ClusterTags struct {
-	CustomTags map[string]string `json:"custom_tags"`
+	CustomTags   map[string]string `json:"custom_tags"`
+	SparkVersion string            `json:"spark_version"`
 }
 
 // JobTags custom tags from Databricks API response
@@ -405,7 +406,18 @@ func fetchClusterTags(client *http.Client, host, token, clusterID string, s *com
 		return nil, fmt.Errorf("failed to parse response: %w", err)
 	}
 
-	return clusterResponse.CustomTags, nil
+	// Create a copy of the custom tags
+	tags := make(map[string]string)
+	for k, v := range clusterResponse.CustomTags {
+		tags[k] = v
+	}
+
+	// Add spark_version as runtime tag if available
+	if clusterResponse.SparkVersion != "" {
+		tags["runtime"] = clusterResponse.SparkVersion
+	}
+
+	return tags, nil
 }
 
 func fetchJobTags(client *http.Client, host, token, jobID string, s *common.Setup) (map[string]string, error) {
--- a/pkg/fleet/installer/setup/djm/databricks.go
+++ b/pkg/fleet/installer/setup/djm/databricks.go
@@ -9,12 +9,16 @@ package djm
 import (
 	"encoding/json"
 	"fmt"
+	"io"
+	"net/http"
 	"os"
 	"regexp"
 	"strings"
+	"time"
 
 	"github.com/DataDog/datadog-agent/pkg/fleet/installer/setup/common"
 	"github.com/DataDog/datadog-agent/pkg/fleet/installer/setup/config"
+	"github.com/DataDog/datadog-agent/pkg/fleet/installer/telemetry"
 	"github.com/DataDog/datadog-agent/pkg/util/log"
 )
 
@@ -22,6 +26,7 @@ const (
 	databricksInjectorVersion   = "0.39.1-1"
 	databricksJavaTracerVersion = "1.49.0-1"
 	databricksAgentVersion      = "7.63.3-1"
+	fetchTimeoutDuration        = 5 * time.Second
 )
 
 var (
@@ -105,6 +110,7 @@ func SetupDatabricks(s *common.Setup) error {
 	}
 
 	setupCommonHostTags(s)
+	fetchDatabricksCustomTags(s)
 	installMethod := "manual"
 	if os.Getenv("DD_DJM_INIT_IS_MANAGED_INSTALL") == "true" {
 		installMethod = "managed"
@@ -247,7 +253,6 @@ func setupDatabricksWorker(s *common.Setup) {
 		s.Span.SetTag("host_tag_set.worker_logs_enabled", "true")
 		s.Config.IntegrationConfigs["spark.d/databricks.yaml"] = sparkIntegration
 	}
-
 }
 
 func addCustomHostTags(s *common.Setup) {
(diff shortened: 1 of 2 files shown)


pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -301,3 +304,140 @@ func loadLogProcessingRules(s *common.Setup) {
...
+		if err != nil {
+			log.Warnf("Failed to fetch job tags: %v", err)
+			s.Out.WriteString(fmt.Sprintf("Failed to fetch job tags: %v\n", err))
+		} else {
+			addTagsToConfig(s, jobTags)
+		}
+	}
+}
+
+func fetchClusterTags(client *http.Client, host, token, clusterID string) (map[string]string, error) {
```
aboitreaud, line 357: I think we need some telemetry (spans rather than logs) for this function, can you please take inspiration from here https://github.com/DataDog/datadog-agent/blob/78ba70d05b94fd9d9790f52316941951078bc109/pkg/fleet/installer/setup/djm/emr.go#L189 ?

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -301,3 +304,140 @@ func loadLogProcessingRules(s *common.Setup) {
...
+	Settings struct {
+		Tags map[string]string `json:"tags"`
+	} `json:"settings"`
+}
+
+func fetchDatabricksCustomTags(s *common.Setup) {
+	token := os.Getenv("DATABRICKS_TOKEN")
+	host := os.Getenv("DATABRICKS_HOST")
+	if token == "" || host == "" {
+		log.Debug("DATABRICKS_TOKEN or DATABRICKS_HOST not set, skipping custom tags fetch")
```
aboitreaud, line 324: For telemetry, either use s.Out.WriteString to log in the console of the user of add create spans and add tags on them.
The logs are hard to find and we don't really use them, especially a log.Debug I have no clue how to enable these

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -301,3 +304,145 @@ func loadLogProcessingRules(s *common.Setup) {
...
+}
+
+func addTagsToConfig(s *common.Setup, tags map[string]string) {
+	if len(tags) == 0 {
+		return
+	}
+
+	for key, value := range tags {
+		sanitizedKey := strings.ReplaceAll(key, ":", "_")
+		sanitizedValue := strings.ReplaceAll(value, ":", "_")
```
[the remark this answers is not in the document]
aboitreaud: Not that I know of, I'll let you check

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -301,3 +304,142 @@ func loadLogProcessingRules(s *common.Setup) {
...
+func fetchJobTags(client *http.Client, host, token, jobID string) (map[string]string, error) {
+	url := fmt.Sprintf("%s/api/2.1/jobs/get?job_id=%s", host, jobID)
+
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create request: %w", err)
+	}
+
+	req.Header.Set("Authorization", "Bearer "+token)
+	req.Header.Set("Content-Type", "application/json")
```
aboitreaud, line 404: Apparently we don't need this one for a GET: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Type

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -301,3 +304,142 @@ func loadLogProcessingRules(s *common.Setup) {
...
+		}
+	}
+}
+
+func fetchClusterTags(client *http.Client, host, token, clusterID string) (map[string]string, error) {
+	url := fmt.Sprintf("%s/api/2.0/clusters/get", host)
+
+	reqBody := fmt.Sprintf(`{"cluster_id": "%s"}`, clusterID)
+
+	req, err := http.NewRequest("POST", url, strings.NewReader(reqBody))
```
aboitreaud, line 362: Are you sure the request should be a POST ? I found this for v2.1 https://docs.databricks.com/api/workspace/clusters/get

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -301,3 +304,142 @@ func loadLogProcessingRules(s *common.Setup) {
...
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		body, _ := io.ReadAll(resp.Body)
+		return nil, fmt.Errorf("API request failed with status %d: %s", resp.StatusCode, string(body))
+	}
+
+	var clusterResponse ClusterTags
+	body, err := io.ReadAll(resp.Body)
```
aboitreaud, line 382: body is read twice from `resp.Body`, at line 382 and 377, can we avoid that ?

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -301,3 +304,142 @@ func loadLogProcessingRules(s *common.Setup) {
...
+	req.Header.Set("Content-Type", "application/json")
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("failed to send request: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		body, _ := io.ReadAll(resp.Body)
```
aboitreaud, line 413: same for body here

pkg/fleet/installer/setup/djm/databricks.go
RickyMarou: yes the documentation is wrong. If you try the request like specified in the docs it just fails 🤦

RickyMarou: For posterity: `databricks api get /api/2.1/clusters/get?cluster_id=$CLUSTER_ID` does not work in the CLI, but the same request with cURL works. Other API requests with the databricks API are working

🤡

aboitreaud: Great, thanks for checking that !

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -106,6 +109,7 @@ func SetupDatabricks(s *common.Setup) error {
 	}
 
 	setupCommonHostTags(s)
+	fetchDatabricksCustomTags(s)
```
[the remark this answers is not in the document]
RickyMarou: Let's keep this check in the function:
- I don't think it makes any difference in terms of readability
- It's much easier to unit test that we don't drop the ball if the environment variables are not present

Or is that something that you feel strongly about? We could still do it, refactor unit test and move them at the level of the `SetupDatabricks` function 🤔

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -301,3 +304,140 @@ func loadLogProcessingRules(s *common.Setup) {
...
+	host := os.Getenv("DATABRICKS_HOST")
+	if token == "" || host == "" {
+		log.Debug("DATABRICKS_TOKEN or DATABRICKS_HOST not set, skipping custom tags fetch")
+		return
+	}
+
+	s.Out.WriteString("Fetching custom tags from Databricks API\n")
+
+	client := &http.Client{
+		Timeout: 30 * time.Second,
```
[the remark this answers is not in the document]
RickyMarou: On my machine the request are taking half a second to come back. I assume the hosts should have a decent connection workspace too. Let's scale this back to 5s that's more than enough time to get an error.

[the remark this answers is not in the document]
aboitreaud: Sounds good, please const the timeout duration at the top of the file too, like here: https://github.com/DataDog/datadog-agent/blob/d27ed7d77815c760df1f7fe8fed97607251fa6ec/pkg/fleet/installer/setup/common/setup.go#L29-L31

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -302,3 +310,152 @@ func loadLogProcessingRules(s *common.Setup) {
...
+	}
+
+	s.Out.WriteString("Fetching custom tags from Databricks API\n")
+
+	client := &http.Client{
+		Timeout: fetchTimeoutDuration,
+	}
+
+	clusterID := os.Getenv("DB_CLUSTER_ID")
+	if clusterID != "" {
```
aboitreaud, line 346: same double indent, can return early execpt if we still want to fetch job tags when cluster_id couldn't be extracted

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -302,3 +310,152 @@ func loadLogProcessingRules(s *common.Setup) {
...
+		clusterTags, err := fetchClusterTagsFunc(client, host, token, clusterID, s)
+		if err != nil {
+			s.Out.WriteString(fmt.Sprintf("Failed to fetch cluster tags: %v\n", err))
+		} else {
+			addTagsToConfig(s, clusterTags)
+		}
+	}
+
+	jobID, _, ok := getJobAndRunIDs()
+	if ok && jobID != "" {
```
aboitreaud, line 356: double indent makes it hard to read, can you return early with the opposite if condition?

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -302,3 +310,152 @@ func loadLogProcessingRules(s *common.Setup) {
...
+	fetchClusterTagsFunc = fetchClusterTags
+	fetchJobTagsFunc     = fetchJobTags
+)
+
+func fetchClusterTags(client *http.Client, host, token, clusterID string, s *common.Setup) (map[string]string, error) {
+	var err error
+	span, _ := telemetry.StartSpanFromContext(s.Ctx, "fetch.cluster.custom_tags")
+	defer func() { span.Finish(err) }()
+
+	url := fmt.Sprintf("%s/api/2.1/jobs/get?cluster_id=%s", host, clusterID)
```
aboitreaud, line 377: Wrong URL for cluster

pkg/fleet/installer/setup/djm/databricks.go
RickyMarou: yes we do still want to try job tags if cluster_id couldn't be extracted 👍

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -80,6 +85,10 @@ var (
 	}
 )
 
+const (
+	fetchTimeoutDuration = 5 * time.Second
+)
```
aboitreaud, line 91: Can you move this const with the other ones at the top of the file ?

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -301,3 +311,159 @@ func loadLogProcessingRules(s *common.Setup) {
...
+
+	for key, value := range tags {
+		// First replace colons with underscores, then apply additional normalization
+		normalizedKey := normalizeutil.NormalizeTagValue(strings.ReplaceAll(key, ":", "_"))
+		normalizedValue := normalizeutil.NormalizeTagValue(strings.ReplaceAll(value, ":", "_"))
+
+		tagString := fmt.Sprintf("%s:%s", normalizedKey, normalizedValue)
+		s.Config.DatadogYAML.Tags = append(s.Config.DatadogYAML.Tags, tagString)
+
+		s.Span.SetTag("host_tag_set."+normalizedKey, normalizedValue)
```
aboitreaud, line 467: Are we okay in terms of privacy to set these tag in clear on our spans ? Wondering if any tag can contain PII, if so we may want to tag our install spans with the number of custom tag fetched. That still gives some debug info without revealing the values

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -301,3 +311,159 @@ func loadLogProcessingRules(s *common.Setup) {
...
+// Fetch Cluster custom tags and Job custom tags from the databricks API
+// Will only do requests if the API key and Hostname are present
+// It should not do requests without those, it should not panic if the HTTP requests are failing.
+// It should add the custom tags to the datadog-agent tags config.
+func fetchDatabricksCustomTags(s *common.Setup) {
+	token := os.Getenv("DATABRICKS_TOKEN")
+	host := os.Getenv("DATABRICKS_HOST")
+	if token == "" || host == "" {
+		s.Span.SetTag("databricks_api_auth.provided", "false")
+		s.Out.WriteString("DATABRICKS_TOKEN or DATABRICKS_HOST not set, skipping custom tags fetch")
```
aboitreaud, line 336: Missing `\n` at end of string here

pkg/fleet/installer/setup/djm/databricks.go
RickyMarou: We already put custom tags on the Databricks spans, what this does (or should do) is add the same tags on the Spark spans. I think it's totally fine.

Also It would be pretty weird for customers to put PII in their databricks custom tags 🤔 Like you mean one of the collaborator email address or name? We already have `owner` which contains the email address of the DBX account that created the job, we don't seem to have a problem with that.

pkg/fleet/installer/setup/djm/databricks.go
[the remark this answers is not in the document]
aboitreaud: What's the issue with `:` that forces us to replace colons with underscores ? Can't we accept tags like `env:staging:east`?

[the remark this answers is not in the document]
RickyMarou: We can accept these tags, datadog will split on the first colon `:` so `env:staging:test` will have `env` as key.
But then something like `databricks:env:staging` will have `databricks` as a key.

Thinking about it, the regular tag normalization does not touch the colons, so let's leave them alone 👍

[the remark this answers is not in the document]
aboitreaud: Agree. To follow the key - value pair that databricks gives us and keep the value consistent with what the customer expects, I think we should replace the `:` only in the key, and not in the value

aboitreaud (approved):

BaptisteFoy (approved):
lgtm on the fleet side

PR #37091, opened 2025-05-16
[CONS-7278] Ensure missing volume mounts are added in case of webhook reinvocation

adel121:
### What does this PR do?

Fixes an issue that causes missing injecting volume mounts by the admission controller in case external webhooks are involved in the mutation process.

This occurs because the order by which a pod is mutated by multiple mutating admission webhooks is not well defined and can be random. Kubernetes keeps reinvoking webhooks on the pod until the pod template is stable and not changing anymore (with a recursion depth limit in case an infinite loop happens in some edge cases).

In this specific case, we have a utility function that injects a volume and adds volume mounts to the existing containers and init containers. The problem with the current implementation is that if we find out that the volume already exists (we match volumes name), we directly return without adding any volume mounts. Consequently, if a third party webhook injects an init container after we had already mutated the pod, the injected init container will not get the volume mount.

This PR makes sure we still update the volume mounts even if the volume already exists.  

### Motivation

Fix customer issue.

Related to https://github.com/DataDog/datadog-agent/pull/36475

- Updated unit tests
- We already have e2e tests.

If the user manually adds a volume colliding with volumes injected by datadog's admission controller, it might result in inadvertently adding volume mounts to containers.

However, this is an extreme edge case that is okay to not cover (specifically, it is not straightforward to handle it).


pkg/clusteragent/admission/mutate/common/common.go
```diff
@@ -157,14 +153,24 @@ func InjectVolume(pod *corev1.Pod, volume corev1.Volume, volumeMount corev1.Volu
...
-	if shouldInject {
+	volumeAlreadyExists := false
+	for _, vol := range pod.Spec.Volumes {
+		if vol.Name == volume.Name {
+			log.Debugf("Pod %q: Volume %q already exists. Not adding again", podStr, vol.Name)
+			volumeAlreadyExists = true
+		}
+	}
+
+	shouldInjectVolume := !volumeAlreadyExists && injectedVolumeMount
```
davidor, line 167: Why is `injectedVolumeMount` needed here?
If it’s set to false, that means the expected volume mounts were already there.
But shouldn’t the volume be injected anyway if it doesn’t already exist?

davidor (approved):

releasenotes-dca/notes/fix_bug_in_admission_controller_inject_vol-6e9ab84f4dcc4db0.yaml
```diff
@@ -0,0 +1,13 @@
...
+# CHANGELOG-DCA.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+fixes:
+  - |
+    Fix bug in the admission controller config webhook causing missing injection of
```
aliciascott, line 11:
```suggestion
    Fixes a bug in the admission controller webhook that caused volume mounts to be skipped when other webhooks injected init containers after our own volume mounts had been added.
```

releasenotes-dca/notes/fix_bug_in_admission_controller_inject_vol-6e9ab84f4dcc4db0.yaml
```diff
@@ -0,0 +1,13 @@
...
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+fixes:
+  - |
+    Fix bug in the admission controller config webhook causing missing injection of 
+    volume mounts when running alongside other webhooks that might inject init containers
```
aliciascott, line 12:
```suggestion
 
```

releasenotes-dca/notes/fix_bug_in_admission_controller_inject_vol-6e9ab84f4dcc4db0.yaml
```diff
@@ -0,0 +1,13 @@
...
+    after having injected volume mounts in the first place on our side.
```
aliciascott, line 13:
```suggestion

```

git-thuerk-done (changes requested):
Hi @adel121 just some suggestions to re-word this for clarity, let me know once fixed and I'll re-review!

git-thuerk-done (approved):
looks good!

PR #37117, opened 2025-05-16
fix buffered manifest tag collection

fisherevans:
A recent change to the agent resulted in all manifest payloads losing all configured `extraTags`:
[https://github.com/DataDog/datadog-agent/commit/453fc36fa12026e4738cc64d6fb804fed1cf22df#diff-62e0377b239aa3f3652dc40\[…\]aadd5f879c35076ca7843L57-R57](https://github.com/DataDog/datadog-agent/commit/453fc36fa12026e4738cc64d6fb804fed1cf22df#diff-62e0377b239aa3f3652dc400fb04f299043763f7fefaadd5f879c35076ca7843L57-R57)
Specifically this change in `ExtractModelManifests` in `base.go`:
`Tags: append(pctx.Cfg.ExtraTags, pctx.ApiGroupVersionTag)` -> `Tags: pctx.ExtraTags`

`pctx` is a `K8sProcessorContext`, which inherits from `BaseProcessorContext`, which had `ExtraTags` added to it:
[https://github.com/DataDog/datadog-agent/commit/453fc36fa12026e4738cc64d6fb804fed1cf22df#diff-efdb3926187ba5c5be426a8\[…\]50739c834581c919016e39c5aR47](https://github.com/DataDog/datadog-agent/commit/453fc36fa12026e4738cc64d6fb804fed1cf22df#diff-efdb3926187ba5c5be426a80623be06d631dce250739c834581c919016e39c5aR47)

but, when we buffer manifests, we never populate the newly created `ExtraTags` field (which `ExtractModelManifests` now pulls from), just the original `ExtraTags` in the `OrchestratorConfig`:
[https://github.com/DataDog/datadog-agent/blob/453fc36fa12026e4738cc64d6fb804fed1cf\[…\]kg/collector/corechecks/cluster/orchestrator/manifest\_buffer.go](https://github.com/DataDog/datadog-agent/blob/453fc36fa12026e4738cc64d6fb804fed1cf22df/pkg/collector/corechecks/cluster/orchestrator/manifest_buffer.go#L93-L105)

the `ExtraTags` added in `flushManifest` are never actually referenced anymore, thus don't end up in the `CollectorManifest` model.

to get back to where we were, it's as simple as setting the new `ExtraTags` field on the `pctx` instead of the cfg -- **but! it's never actually been right**.

The `pctx.ApiGroupVersionTag` was never being set right for manifests and was being lost during the buffering, so that tag was always missing on `orchmanif` - and that's because the buffer is shared across processor contexts, multiple resource types feed into the buffer.

> Note: We [DO NOT buffer CRD or CR manifest payloads](https://github.com/DataDog/datadog-agent/blob/main/pkg/collector/corechecks/cluster/orchestrator/collectors/k8s/crd.go#L48), which is why they are not suffering from the same tag issues (and always had the proper kube\_api\_version tags)

- When the per-resource collector runs, the original `model.Manifest` is created with the `ResourceTaggerTags` (which is only implemented for pods) - but once all the resource manifests are collected for a single resource run, `ChunkManifest` is called within `processor.Process` with a "real" `pctx` with the properly generated `ExtraTags`.
- At this point, all the tags are correct.
- The problem comes when we buffer manifests. Inside `collector.Run`, it sends the proper `CollectorManifest` list into a buffer channel. When this happens, it just buffers the internal `Manifest` models, dropping the outer `CollectorManifest` and it's tags.
- When it flushes, it creates new wrapper `CollectorManifest`s around the chunks with a "fake" `pctx` with the missing `ExtraTags`. The problem is that at this point, the collection of child `Manifest` messages are a mix of resource types, the newly created `CollectorManifest` is not being set with the proper `ExtraTags`, instead the unset `nil` `ExtraTags`.

-----------------------------------

The PR fixes this by replacing the newer `ExtraTags` on `pctx` with a dedicated `CollectorTags`. We now merge `ExtraTags` and `CollectorTags` when constructing the messages:
- For metadata, the two are merged and added to the `Collector*` model `Tags`
- For manifests, `ExtraTags` are used for the `CollectorManifest` model, and the `CollectorTags` are added to the inner `Manifest` model.

Also, merged into this PR was: https://github.com/DataDog/datadog-agent/pull/37120 - this adds:
- once the metadata model is fully created, and all tags are added to it, extract it and include it in the manifest model.
  - for pods, instead of re-querying the tagger, this relies on the tags previously pulled from the tagger

---------------------------------------

To validate:
- Deploy DCA and node agent (i.e. kind)
- ensure orch and orchmanif payloads have the same tags, sourced from all three locations:
  - global tags on the collector payloads
  - collector tags (i.e. kube api version)
    - for orch, on the parent collector* payloads
    - for orchmanif, on the manifest child payloads
  - resource + tagger tags in the child payloads
 
note: pod payloads source global tags differently than other resources because it is not in the DCA. extra tags will be added to pods via HMS - but orchmanif does not have that processor setup in EVP yet


releasenotes/notes/orchestrator-fix-buffered-manifest-collect-tags-80c90fdaf37bf70a.yaml
```diff
@@ -0,0 +1,3 @@
+fixes:
+  - |
+    Fix a bug that resulted in some Orhecreator K8S Manifests losing the configured "extraTags".
```
aliciascott, line 3:
```suggestion
    Fix a bug that resulted in some Orchestrator K8S Manifests losing the configured "extraTags".
```

git-thuerk-done (dismissed):
Hi @fisherevans Quick typo, let me know once fixed and I'll re-review thanks!

maycmlee (commented):
Some small suggestions and maybe a typo,

aureleoules (approved):

xlucas:
Do you think we could update manifest buffer test in that PR or a subsequent one to cover this case?

xlucas (approved):

maycmlee (approved):

git-thuerk-done (approved):

PR #37120, opened 2025-05-16
include all tags with manifests

fisherevans:
- once the metadata model is fully created, and all tags are added to it, extract it and include it in the manifest model.
  - for pods, instead of re-querying the tagger, this relies on the tags previously pulled from the tagger
---------------------------------------

To validate:
- Deploy DCA and node agent (i.e. kind)
- ensure orch and orchmanif payloads have the same pods, sourced from all three locations:
  - global tags on the collector payloads
  - collector tags (i.e. kube api version)
    - for orch, on the collector* payloads
    - for orchmanif, on the manifest child payloads
  - resource + tagger tags in the child payloads (for all resources)

To validate:
- Deploy DCA and node agent (i.e. kind)
- ensure orch and orchmanif payloads have the same tags, sourced from all three locations:
  - global tags on the collector payloads
  - collector tags (i.e. kube api version)
    - for orch, on the parent collector* payloads
    - for orchmanif, on the child resource payloads
  - resource + tagger tags in the child payloads (now, for all resources)

note: pod payloads source global tags differently than other resources because it is not in the DCA. extra tags will be added to pods via HMS - but orchmanif does not have that processor setup in EVP yet


pkg/collector/corechecks/cluster/orchestrator/processors/common/base.go
```diff
@@ -35,7 +35,7 @@ func (BaseHandlers) BuildManifestMessageBody(ctx processors.ProcessorContext, re
 }
 
 //nolint:revive // TODO(CAPP) Fix revive linter
-func (BaseHandlers) ResourceTaggerTags(ctx processors.ProcessorContext, resource interface{}) []string {
+func (BaseHandlers) GetMetadataTags(ctx processors.ProcessorContext, resource interface{}) []string {
```
fisherevans, line 38: instead of fetching tagger tags, just pull whatever tags exist in the metadata

fisherevans:
confirmed to be working - deployed locally, and queried for my agent deployment in ddsql
image
- from_fisher_laptop:true is from extra tags
- kube_api_version is from collector tags
- everything else is from resource tagger

git-thuerk-done (approved):
good for docs

xlucas (approved):

PR #37125, opened 2025-05-18
usm: kafka: Handle edge cases with first latency

guyarb:
### What does this PR do?

Our kafka kernel decoder can capture requests with multiple records. We had an assumption that the first request must have only 1 record, hence setting the FirstLantecySample. However, that's isn't true and can lead to inaccuracy in our backend.

The change ensures we set the FirstLatnecySample only if the first request has only 1 record.

### Motivation

Handling a potential inaccuracy by kafka.

### DIFF
--- a/pkg/network/protocols/kafka/stats.go
+++ b/pkg/network/protocols/kafka/stats.go
@@ -148,7 +148,7 @@ func (r *RequestStats) AddRequest(errorCode int32, count int, staticTags uint64,
 		}
 
 		// The kafka kernel decoder can capture multiple requests in a single packet, so
-		// in case of a case of a new event with multiple requests, we might not have a FirstLatencySample
+		// in case of a new event with multiple requests, we might not have a FirstLatencySample
 		// In such a case, we need to skip adding the latency sample to the sketch
 		if originalCount == 1 {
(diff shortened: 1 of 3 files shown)


pkg/network/protocols/kafka/stats.go
```diff
@@ -147,9 +147,14 @@ func (r *RequestStats) AddRequest(errorCode int32, count int, staticTags uint64,
 			return
 		}
 
-		// Add the deferred latency sample
-		if err := stats.Latencies.AddWithCount(stats.FirstLatencySample, float64(originalCount)); err != nil {
-			log.Debugf("could not add request latency to ddsketch: %v", err)
+		// The kafka kernel decoder can capture multiple requests in a single packet, so
+		// in case of a case of a new event with multiple requests, we might not have a FirstLatencySample
```
guymor4, line 151:
```suggestion
		// in case of a new event with multiple requests, we might not have a FirstLatencySample
```

guymor4 (approved):
LGTM my only reservation is that we are still missing latency points on requests with multiple records maybe we should think of a different approach but for now it's obviously better :)

PR #37142, opened 2025-05-19
[CWS] fix some file.path field not being in the invalidDiscarders list

safchain:
### What does this PR do?

Fix some file.path field not being in the invalidDiscarders list. Without this fix "empty discarder" could be pushed.

### DIFF
--- a/pkg/security/probe/discarders_linux.go
+++ b/pkg/security/probe/discarders_linux.go
@@ -61,9 +61,8 @@ var (
 type onDiscarderHandler func(rs *rules.RuleSet, event *model.Event, probe *EBPFProbe, discarder Discarder) (bool, error)
 
 var (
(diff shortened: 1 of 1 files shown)


pkg/security/probe/discarders_linux.go
```diff
@@ -477,26 +478,7 @@ func filenameDiscarderWrapper(eventType model.EventType, getter inodeEventGetter
...
-		for _, value := range values {
-			invalidDiscarders[invalidDiscarderEntry{
-				field: field,
-				value: value,
-			}] = true
-		}
-	}
-
-	return invalidDiscarders
+	return strings.HasSuffix(field, ".file.path") && value == ""
```
paulcacheux, line 481: what about `file.destination.path` ?

paulcacheux (changes requested):

paulcacheux (approved):

PR #37147, opened 2025-05-19
Add byoc field to redapl

mackjmr:
### What does this PR do?

Add BYOC Field to otel agent metadata, see: https://github.com/DataDog/infrastructure-resources/pull/2055.


tasks/otel_agent.py
```diff
@@ -40,7 +40,7 @@ def byoc_release(ctx, image=DDOT_DEV_AGENT_TAG, branch=DDOT_DEV_AGENT_BRANCH, re
 
 
 @task
-def build(ctx):
+def build(ctx, byoc="false"):
```
truthbk, line 43: IMHO opinion this should be a boolean
```suggestion
def build(ctx, byoc="false"):
```

tasks/otel_agent.py
```diff
@@ -51,6 +51,7 @@ def build(ctx):
     env = {"GO111MODULE": "on"}
     build_tags = get_default_build_tags(build="otel-agent")
     ldflags = get_version_ldflags(ctx, major_version='7')
+    ldflags += ' -X main.BYOC=' + byoc
```
truthbk, line 54:
```suggestion
    ldflags += f' -X main.BYOC={byoc}'
```

cmd/otel-agent/main.go
```diff
@@ -17,7 +17,13 @@ import (
 	_ "github.com/DataDog/datadog-agent/pkg/version"
 )
 
+var BYOC string
```
truthbk, line 20: I would maybe move this to `cmd.otel-agent.command.BYOC` and just have it next to `cmd.otel-agent.command.loggerName`, and keep the `MakeRootCommand` prototype unchanged. You can then access the variable directly in the package. 

Also while this **has** to be a string, because it's a build-time variable, all downstream calls should use a bool.

cmd/otel-agent/main.go
```diff
@@ -17,7 +17,13 @@ import (
...
+
 func main() {
+	var BYOCBool bool
```
truthbk, line 23: nit: I'd maybe put this logic inside `MakeRootCommand()` instead of here after moving the variable as suggested above. 

We have to be resilient to case, so I would make a case a insensitive comparison using. `strings.EqualFold(BYOC, "true")`

truthbk (commented):
Just some nits, looks mostly ready to go! 👍

tasks/otel_agent.py
mackjmr: I can't see a change in the suggestion, but if you are suggesting:
```
def build(ctx, byoc=False):
```

I don't think invoke can convert to bool:

```
➜  datadog-agent git:(mackjmr/byoc-redapl) ✗ inv otel-agent.build --byoc=false && ./bin/otel-agent/otel-agent run --config collector.yaml --core-config acfg.yaml
No idea what 'false' is!

➜  datadog-agent git:(mackjmr/byoc-redapl) ✗ inv otel-agent.build --byoc=False && ./bin/otel-agent/otel-agent run --config collector.yaml --core-config acfg.yaml
No idea what 'False' is!
```

truthbk: I meant exactly that:
```
def build(ctx, byoc=False):
```

I think invoke does support it, but it becomes something like `--byoc` or `--no-byoc`, or something like that. I'm almost sure boolean flags are supported.

truthbk (approved):

CelianR (approved):

PR #37148, opened 2025-05-19
Update developer docs dependencies

ofek:
### What does this PR do?

- Moves the `dda inv docs.*` commands to the [local command](https://datadoghq.dev/datadog-agent-dev/how-to/extend/local/) group `dda run docs`
- Dependencies are now defined here rather than in dda since it is repo-specific
- Update the syntax for blocks & admonitions to the new style (removed indentation accounts for the majority of the diff)

### Additional Notes

Dependencies are managed manually in a virtual environment since there is no need to import directly


.dda/extend/commands/run/docs/build/__init__.py
```diff
@@ -0,0 +1,39 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+import click
+
+from dda.cli.base import dynamic_command, pass_app
+
+if TYPE_CHECKING:
+    from dda.cli.application import Application
```
CelianR, line 10: If there is no `TYPE_CHECKING` won't this fail at line 16?

CelianR (approved):

.dda/extend/commands/run/docs/build/__init__.py
ofek: No because `from __future__ import annotations` sets type hints as lazy (more or less) and that will be the default/only way in 3.13 when we upgrade

tobz (approved):

remeh (approved):
Checked agent-metric-pipelines owned files, LGTM

misteriaud (approved):

PR #37157, opened 2025-05-19
[NDINT-290] Director Hardware Metrics

ken-schneider:
### What does this PR do?
_**I know it's long, but reviewers please read.**_

Adds parsing for hardware metrics for the Versa director including:
`versa.cpu.usage`, `versa.memory.usage`, `versa.disk.usage`, `versa.device.uptime`.

Unlike appliances, the director has increased granularity for some of its metrics.
- `versa.disk.usage` is split up by `partition` and is tagged with `partition`

This is because we receive disk usage in the following format:
```
partition=root,size=50GB,free=10GB,usedRatio=20.0;partition=opt,size=30GB,free=15GB,usedRatio=50.0
```
The metric is parsed from `usedRatio`

- `versa.device.uptime` is split by `type` which will be either `application` of `system`

This is because we receive uptime in the following structure:
```json
"systemUpTime": {
        "currentTime": "Mon Jan 1 00:00:00 UTC 1970",
        "applicationUpTime": "24 Days, 9 Hours, 55 Minutes, 3 Seconds.",
        "sysProcUptime": "56 Days, 10 Hours, 13 Minutes, 18 Seconds.",
        "sysUpTimeDetail": "00:00:00 up 256 days, 17:53,  1 user,  load average: 0.02, 0.06, 0.08"
    }
```
We parse `applicationUpTime` and tag `versa.device.uptime` with the additional type tag of `application` thus the `type` tag would be `type:director,application`. It's a similar story for `sysProc` but it's `director,sys_proc`.

I'm open to feedback on a separate tag key for the uptimes, but this was what made the most sense to me.

Ran the agent locally against a real Versa cluster

### DIFF
--- a/pkg/collector/corechecks/network-devices/versa/report/sender.go
+++ b/pkg/collector/corechecks/network-devices/versa/report/sender.go
@@ -192,7 +192,7 @@ func (s *Sender) SendDirectorStatus(director *client.DirectorStatus) {
 func (s *Sender) SendDirectorDeviceMetrics(director *client.DirectorStatus) {
 	ipAddress, err := director.IPAddress()
 	if err != nil {
-		log.Errorf("Error getting director IP address: %s", err)
+		log.Errorf("Error getting director IP address, skipping device metrics: %s", err)
 		return
 	}
 	tags := s.GetDeviceTags(defaultIPTag, ipAddress)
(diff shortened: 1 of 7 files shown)


pkg/collector/corechecks/network-devices/versa/report/sender.go
```diff
@@ -112,6 +128,11 @@ func (s *Sender) SendDeviceStatusMetrics(deviceStatus map[string]float64) {
 	for device, status := range deviceStatus {
 		tags := s.GetDeviceTags(defaultIPTag, device)
 		s.Gauge(versaMetricPrefix+"device.reachable", status, "", tags)
+		if status > 0 {
```
zoedt, line 131: **nit / more discussion topic:**
i recall you bringing up the metrics for reachability being a bit odd; i was curious if we could chat more about it later of when the graph wasn't working (i think i remember you tried following the way cisco sd-wan does it) but wasn't properly pulling the expected number of statuses for devices down/up

more of a discussion topic bc i suppose we may want some metric standardization across integrations over time (whether just one `reachable` metrics vs. having two?) not sure it really matters but worthwhile later, i think, when the team tries to decide how we want to structure when we introduce new integrations!

zoedt (approved):
overall, lgtm! ty for the detailed pr description + doing all that fun parsing 😭

pkg/collector/corechecks/network-devices/versa/report/sender.go
ken-schneider: Yeah I agree. In general I'm partial to having a single metric here `versa.device.reachable`. I think it's a cleaner approach, but even copying over the same exact formulas from another dashboard I wasn't able to get accurate counts. It might be an easy fix, it might be down to the amount of time we've been reporting these metrics.

I created a story to capture the need to discuss this further.

For now I'd rather have something that's usable which we can change before this goes GA.

PR #37170, opened 2025-05-19
(fleet) merge service management between deb/rpm/oci

arbll:
This PR finishes the merge between OCI/DEB/RPM systemd units and service management in package scripts.

Changes:
- DEB/RPM units are generated from the same source as the OCI units
- DEB/RPM now have experiment units. They won't be immediately useable but we'll add configuration change support as a fast follow
- The installer is now started by default on deb/rpm and will fast exit if it's not enabled. This is in line with the windows behavior.
- `postinst`, `postrm`, etc... logic is now pretty much the same across all package types 

Customer-facing changes:
- Systemd units are no longer shipped inside of the deb/rpm archive but instead written during the `postinst` by the installer. This is a subtle change but I'm adding a release note entry in case someone depends on it.

Notes:
- I technically removed support for overriding the `install_dir` in the units for deb/rpm but I don't see how this could have worked anyway since the package script all assume `/opt/datadog-agent` as the install dir...


hestonhoffman (approved):

arbll:

BaptisteFoy (approved):

CelianR (approved):

arbll:

chouquette (approved):

raphaelgavache (approved):

PR #37179, opened 2025-05-20
Don't install libssl-dev on docker images

alopezz:
### What does this PR do?

Removes installation of `libssl-dev`.

### Motivation

I saw the comment about this being required for Python integrations while working on other changes, and it sounded wrong. The Agent generally should not be depending on system libs, as the regular packages don't make this assumption either; so I don't think this applies anymore.

I expect any failure to be reflected in E2E tests (either here or integrations-core).


L3n41c (commented):
> I expect any failure to be reflected in E2E tests (either here or integrations-core).

Let’s keep in mind that:
* on `datadog-agent`’s pull requests, e2e tests are triggered depending on the files touched in the PR.
  This PR isn’t touching any of the files that would trigger the run of the containers e2e tests: https://github.com/DataDog/datadog-agent/blob/3f6d2bb4e93541967d6d1335733e0e4a2dc89459/.gitlab-ci.yml#L799-L822
  So, e2e tests will need to be triggered manually.
* The comment mentions that this was needed by some Python checks and:
  * `datadog-agent` e2e tests are not extensively validating the Python checks. We only use as few of them as possible to validate features that makes sense only in the context of Python checks.
  * `integrations-core` tests are not using the docker images produced by `datadog-agent` that are modified in this PR.
  * There’s always the risk of breaking customer’s custom integrations.

alopezz:

alopezz:
> * on `datadog-agent`’s pull requests, e2e tests are triggered depending on the files touched in the PR.
>   This PR isn’t touching any of the files that would trigger the run of the containers e2e tests: 
>   
>   So, e2e tests will need to be triggered manually.

Thanks for pointing that out! I'll trigger those to be extra sure before attempting to merge.

>   * `integrations-core` tests are not using the docker images produced by `datadog-agent` that are modified in this PR.

They will pick them up once this gets merged, so if something is truly broken we should figure it out in time before the change hits a release. I admit it's not ideal, but we also think that the risk is extremely low -- otherwise non-containerized versions of those integrations would have been broken for a long time.

>   * There’s always the risk of breaking customer’s custom integrations.

I think the risk of this is as good as zero based on a similar argument. But beyond that, when running Python integrations with the Agent, the rpath we set on our binaries pretty much already guarantees that the libssl/libcrypto that sits first in the search path is the one we embed in the Agent package.

alopezz:

chouquette (approved):

PR #37180, opened 2025-05-20
omnibus: python: drop dependency to ncurses

chouquette:
we don't need a fancy interactive shell for our purposes


chouquette:

alopezz (approved):

iliakur (approved):

PR #37190, opened 2025-05-20
[CWS] Handle cgroups in threaded mode

lebauce:
### What does this PR do?

Handle cgroup in threaded mode.

### Motivation

Process in a cgroup in threaded mode had an empty cgroup ID because
we failed to read the cgroup.procs file.

### DIFF
--- a/pkg/security/utils/cgroup.go
+++ b/pkg/security/utils/cgroup.go
@@ -289,7 +289,13 @@ func (cfs *CGroupFS) FindCGroupContext(tgid, pid uint32) (containerutils.Contain
 func checkPidExists(sysFScGroupPath string, expectedPid uint32) (bool, error) {
 	data, err := os.ReadFile(filepath.Join(sysFScGroupPath, "cgroup.procs"))
 	if err != nil {
-		return false, err
+		// the cgroup is in threaded mode, and in that case, reading cgroup.procs returns ENOTSUP.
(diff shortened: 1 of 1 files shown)


spikat (approved):

pkg/security/utils/cgroup.go
```diff
@@ -289,7 +289,13 @@ func (cfs *CGroupFS) FindCGroupContext(tgid, pid uint32) (containerutils.Contain
...
 	data, err := os.ReadFile(filepath.Join(sysFScGroupPath, "cgroup.procs"))
 	if err != nil {
-		return false, err
+		// the cgroup is in threaded mode, and in that case, reading cgroup.procs returns ENOTSUP.
+		// see https://github.com/opencontainers/runc/issues/3821
+		if errors.Is(err, unix.ENOTSUP) {
+			if data, err = os.ReadFile(filepath.Join(sysFScGroupPath, "cgroup.threads")); err != nil {
+				return false, err
+			}
+		}
```
YoannGh, line 298:
```suggestion
		} else {
		    return false, err
		}
```

YoannGh: or use `if !errors.Is(err, unix.ENOTSUP)` instead to make the code more readable

lebauce: Good catch, thanks. Fixed

YoannGh (approved):

PR #37195, opened 2025-05-20
Replace rand.Seed behavior to use new function calls

louis-cqrl:
### What does this PR do?

Replace rand.Seed behavior that has been deprecated in order to use `rand.New(rand.NewSource())`

### Motivation

```
[math/rand](https://tip.golang.org/pkg/math/rand/)[¶](https://tip.golang.org/doc/go1.24#mathrandpkgmathrand)
Calls to the deprecated top-level [Seed](https://tip.golang.org/pkg/math/rand#Seed) function no longer have any effect. To restore the old behavior use [GODEBUG setting](https://tip.golang.org/doc/godebug) randseednop=0. For more background see [proposal #67273](https://tip.golang.org/issue/67273).
```


paulcacheux (approved):

test/new-e2e/tests/gpu/capabilities.go
```diff
@@ -88,7 +88,7 @@ func (c *hostCapabilities) QuerySysprobe(path string) (string, error) {
 func (c *hostCapabilities) RunContainerWorkloadWithGPUs(image string, arguments ...string) (string, error) {
 	// Yes, it's deprecated, but apparently without this we get the same random string every time
 	// so seed it just in case, it's a test so we don't care too much about RNG safety
```
gjulianm, line 90: Could you remove this comment as it's now irrelevant? Thanks

test/new-e2e/tests/gpu/capabilities.go
```diff
@@ -171,7 +171,7 @@ func (c *kubernetesCapabilities) QuerySysprobe(path string) (string, error) {
 func (c *kubernetesCapabilities) RunContainerWorkloadWithGPUs(image string, arguments ...string) (string, error) {
 	// Yes, it's deprecated, but apparently without this we get the same random string every time
 	// so seed it just in case, it's a test so we don't care too much about RNG safety
```
gjulianm, line 173: Same with this comment, it's irrelevant now.

gjulianm (approved):
LGTM, thanks a lot for doing this!

test/new-e2e/tests/gpu/capabilities.go
louis-cqrl: yes done

dustmop (approved):
LGTM for agent-configuration

guyarb (approved):

jose-manuel-almaza (approved):

comp/snmptraps/oidresolver/oidresolverimpl/oid_resolver_test.go
```diff
@@ -293,7 +292,6 @@ func updateResolverWithIntermediateYAMLReader(t *testing.T, oidResolver *multiFi
 }
 
 func TestIsValidOID_PropertyBasedTesting(t *testing.T) {
-	rand.Seed(time.Now().Unix())
```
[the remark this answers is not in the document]
jedupau: ok great !

jedupau (approved):

pkg/network/usm/testutil/grpc/client.go
```diff
@@ -46,8 +46,8 @@ func (c *Client) HandleStream(ctx context.Context, numberOfMessages int32) error
 		input[i] = i + 1
 	}
 
-	rand.Seed(time.Now().UnixNano())
-	rand.Shuffle(len(input), func(i, j int) { input[i], input[j] = input[j], input[i] })
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
+	r.Shuffle(len(input), func(i, j int) { input[i], input[j] = input[j], input[i] })
```
L3n41c, line 50: The default generator of the `rand` package is now automatically randomly seeded at program startup.
This should be equivalent to:

```suggestion
	rand.Shuffle(len(input), func(i, j int) { input[i], input[j] = input[j], input[i] })
```

comp/core/secrets/secretsimpl/secrets_test.go
```diff
@@ -882,7 +883,7 @@ func TestStartRefreshRoutineWithScatter(t *testing.T) {
 
 			if tc.scatter {
 				// The set random seed has a the scatterDuration is 5.477027098s
-				mockClock.Add(6 * time.Second)
+				mockClock.Add(time.Duration(tc.r.Float64() * float64(10*time.Second)))
```
L3n41c, line 886: This change confuses me a bit.

The previous code:
```go
				// The set random seed has a the scatterDuration is 5.477027098s
				mockClock.Add(6 * time.Second)
```
looked like:
1. Given the seed used to initialize the random number generator, we know that `scatterDuration` is `5.477027098s`.
2. We make the `mockClock` progress of slightly more than that.

First point is incorrect (the comment).
While the random generator effectively returns `5.477027098s`, the use of `… / time.Second * time.Second` inside `startRefreshRoutine()` rounds that value to the closest integer number of seconds below than that value: `5s`.

This piece of code has been replaced by:
```go
				// The set random seed has a the scatterDuration is 5.477027098s
				mockClock.Add(time.Duration(tc.r.Float64() * float64(10*time.Second)))
```
which draws a new random number.

From a semantic point of view, we draw a new random number and
* sometimes we’ll make the `mockClock` progress more than `scatterDuration` like with the previous version of the code.
* sometimes we’ll make the `mockClock` progress less than `scatterDuration`, violating the assumption made in the previous version of the code.

Luckily or unluckily, because `rand` isn’t random at all, it happens that, with the chosen seed, the value returned by `tc.r.Float64()` happens to be bigger than the random number returned inside `startRefreshRoutine(…)` so that the test passes.

But this change looks wrong and test passes only by accident.
With other values of the seed, `mockClock` might not progress enough.

L3n41c (commented):
One change looks suspicious to me.
I’d need the code owner of that part to double check it.

comp/core/secrets/secretsimpl/secrets_test.go
L3n41c: @dustmop as you own this file and has already approved this PR, could help me understanding this change?
Why do we need to draw a new random number here?
Isn’t it an issue that we don’t guarantee whether the `mockClock` progresses more or less than `scatterDuration`?

louis-cqrl: from my POV: rand.Seed was Seed the rand inside `startRefreshRoutine()` as well, so we had to give the new rand.Rand to this function as well.
So this function should be okay.

for `mockClock.Add(time.Duration(tc.r.Float64() * float64(10*time.Second)))` I may have not well understood the code originally, I thought rand.Seed was impacting `mockClock.Add` but not really, if I get it right we can put back the old value of: `mockClock.Add(6 * time.Second)` right ?

pkg/network/usm/testutil/grpc/client.go
louis-cqrl: LGTM

comp/core/secrets/secretsimpl/secrets_test.go
louis-cqrl: fixed it let me know if this seems better

L3n41c (commented):

But I would have a remark concerning `comp/core/secrets/secretsimpl/secrets.go` and `pkg/util/backoff/backoff.go`.
`startRefreshRoutine(…)` and `randomBetween(…)` had both to be complexified to make them take an optional `rand.Rand` object as an extra parameter.
This `rand.Rand` object is only passed in the context of tests where the random generator is initialized with a hard-coded seed with the justification of making the test deterministic.

IMO, when we draw a random number, the tested function must pass the tests whatever that random number.
If a function works only for some values, them we must hard-code that value and use a random one.
Forcing a hard-coded seed is the best way to have tests that are consistently passing (or failing) by accident.

Because of this will to keep hard-coded seed value, functions like `randomBetween(…)` now have the following pattern:
```go
func randomBetween(r *rand.Rand, …) … {
    if r == nil {
        return // code that is never tested
    }
    return // code that is used only inside tests
}
```

I feel that such a pattern defeats the purpose of unit tests.

louis-cqrl:
> LGTM for the file owned by @DataDog/container-integrations.
> 
> But I would have a remark concerning `comp/core/secrets/secretsimpl/secrets.go` and `pkg/util/backoff/backoff.go`. `startRefreshRoutine(…)` and `randomBetween(…)` had both to be complexified to make them take an optional `rand.Rand` object as an extra parameter. This `rand.Rand` object is only passed in the context of tests where the random generator is initialized with a hard-coded seed with the justification of making the test deterministic.
> 
> IMO, when we draw a random number, the tested function must pass the tests whatever that random number. If a function works only for some values, them we must hard-code that value and use a random one. Forcing a hard-coded seed is the best way to have tests that are consistently passing (or failing) by accident.
> 
> Because of this will to keep hard-coded seed value, functions like `randomBetween(…)` now have the following pattern:
> 
> ```go
> func randomBetween(r *rand.Rand, …) … {
>     if r == nil {
>         return // code that is never tested
>     }
>     return // code that is used only inside tests
> }
> ```
> 
> I feel that such a pattern defeats the purpose of unit tests.

I do agree however for now I just tried to copy the same behavior as before in order to stop test from failing on main. However we should at some point refactor these tests to serve a better purpose IMO

comp/forwarder/defaultforwarder/blocked_endpoints_test.go
```diff
@@ -19,10 +18,6 @@ import (
 	"github.com/DataDog/datadog-agent/pkg/util/backoff"
 )
 
-func init() {
-	rand.Seed(10)
```
vickenty, line 23: Doesn't seem correct replacement considering this bit from the docs:

> Programs that call Seed with a known value to get a specific sequence of results should use New(NewSource(seed)) to obtain a local random generator.

L3n41c (approved):

comp/forwarder/defaultforwarder/blocked_endpoints_test.go
L3n41c: The question is more to know why seeding the random generator with a hard-coded constant was needed in the first place.

This change indeed changes the behavior of the test because it will make the random generator return random numbers. But is that an issue?

louis-cqrl: from what I saw it will only affect this function:
```
func randomBetween(min, max float64) float64 {
	return rand.Float64()*(max-min) + min
}
```

@vickenty any input that would say that would lead to incorrect testing behavior ?

PR #37204, opened 2025-05-20
[APM] RC callback for MRF (trace-agent)

ichinaski:
### What does this PR do?
Adds Multi-Region Failover (MRF) configuration support to the trace agent, allowing it to receive and process MRF updates from Remote Config. The trace agent now subscribes to `ProductAgentFailover` updates and properly handles failover configuration changes.

Changes:
- Added MRF client initialization in trace agent setup
- Implemented MRF configuration callback to process failover updates
- Added support for enabling/disabling MRF based on configuration
- Integrated with core agent's IPC system for MRF updates

### Motivation

The trace-agent has its own RC implementation, therefore we need to attach the MRF updates callback on its handler.

Added test coverage for MRF configuration handling and state management.


comp/trace/config/remote.go
```diff
@@ -17,6 +19,14 @@ import (
 	"github.com/DataDog/datadog-agent/pkg/version"
 )
 
+const (
+	// rcClientName is the default name for remote configuration clients in the trace agent
+	rcClientName = "trace-agent"
+	// rcClientPollInterval is the default poll interval for remote configuration clients. 5 seconds ensures that
+	// clients remain up to date without paying too much of a performance cost (polls that contain no updates are cheap)
+	rcClientPollInterval = time.Second * 5
```
ichinaski, line 27: 5 seconds matches the poll interval set on the core agent RC implementation.

pkg/trace/remoteconfighandler/remote_config_handler_test.go
```diff
@@ -229,3 +229,126 @@ func TestLogLevel(t *testing.T) {
...
+		RemoteConfigClient:    remoteClient,
+		MRFRemoteConfigClient: mrfClient,
+		DebugServerPort:       1,
+	}
+	h := New(&agentConfig, prioritySampler, rareSampler, errorsSampler)
+
+	// Test enabling MRF
+	enableAPM := true
+	mrfConfig := map[string]interface{}{
+		"failover_apm": &enableAPM,
```
ajgajg1134, line 279: looks like there's no need for a variable here? (or other similar tests)
```suggestion
		"failover_apm": true,
```

pkg/trace/remoteconfighandler/remote_config_handler_test.go
```diff
@@ -229,3 +229,126 @@ func TestLogLevel(t *testing.T) {
...
+	assert.False(t, h.agentConfig.IsMRFEnabled())
+
+	// Test invalid config
+	invalidConfig := state.RawConfig{
+		Config: []byte(`invalid json`),
+	}
+	h.mrfUpdateCallback(map[string]state.RawConfig{"datadog/2/AGENT_FAILOVER/config": invalidConfig}, applyEmpty)
+	assert.False(t, h.agentConfig.IsMRFEnabled())
+
+	ctrl.Finish()
```
ajgajg1134, line 314: supposedly we don't need to call .Finish anymore for gomock since we passed the testing.T into the controller creation

pkg/trace/remoteconfighandler/remote_config_handler.go
```diff
@@ -89,13 +91,64 @@ func (h *RemoteConfigHandler) Start() {
...
+// Note: we only care about APM (failover_apm) configuration, but Logs (failover_logs) and Metrics (failover_metrics)
+// may also be present on the MRF configuration. These are handled on the core agent RC.
+func (h *RemoteConfigHandler) mrfUpdateCallback(updates map[string]state.RawConfig, applyStateCallback func(string, state.ApplyStatus)) {
+	// If the updates map is empty, we should unset the failover settings if they were set via RC previously
+	if len(updates) == 0 {
+		log.Debugf("no MRF configuration in remote config update payload")
+		return
+	}
+
+	var enableAPM *bool
```
ajgajg1134, line 114: Maybe I'm missing something here, why is this a pointer to a bool? It looks like we accept a pointer in on the config file, is it expected that the value there could dynamically change so we need the pointer?

ajgajg1134 (changes requested):
Overall looks pretty good, a few small comments. 

Before merging is there anyway for us to test this once locally e2e to ensure we've got it working all the way?

pkg/trace/remoteconfighandler/remote_config_handler.go
```diff
@@ -89,13 +91,64 @@ func (h *RemoteConfigHandler) Start() {
...
+// mrfUpdateCallback is the callback function for the AGENT_FAILOVER configs.
+// It fetches all the configs targeting the agent and applies the failover_apm settings.
+// Note: we only care about APM (failover_apm) configuration, but Logs (failover_logs) and Metrics (failover_metrics)
+// may also be present on the MRF configuration. These are handled on the core agent RC.
+func (h *RemoteConfigHandler) mrfUpdateCallback(updates map[string]state.RawConfig, applyStateCallback func(string, state.ApplyStatus)) {
+	// If the updates map is empty, we should unset the failover settings if they were set via RC previously
+	if len(updates) == 0 {
+		log.Debugf("no MRF configuration in remote config update payload")
+		return
+	}
```
Hitsuji-M, line 112: Leaving a comment here as I had to make a fix for this on the core-agent side. This implementation will ignore any disable if there's at least one policy enabled. 
See the fix [here](https://github.com/DataDog/datadog-agent/pull/36629). I recommend to check the `enableAPM` var and fallback if its nil at the end of the loop

Hitsuji-M (approved):

pkg/trace/remoteconfighandler/remote_config_handler.go
ichinaski: It is a pointer to differentiate `nil` values from `false` ones.

heyronhay (approved):

pkg/trace/remoteconfighandler/remote_config_handler.go
```diff
@@ -89,13 +91,56 @@ func (h *RemoteConfigHandler) Start() {
...
+				Error: err.Error(),
+			})
+			continue
+		}
+
+		if mrfUpdate == nil || mrfUpdate.FailoverAPM == nil {
+			continue
+		}
+
+		if !(failoverAPM != nil && *failoverAPM) && mrfUpdate.FailoverAPM != nil {
```
GustavoCaso, line 129: This condition confusing. Can we just loop until we get a non-nil value for `mrfUpdate.FailoverAPM` and update the agent with that, if is true break out of the loop

```go
for cfgPath, update := range updates {
	mrfUpdate, err := parseMultiRegionFailoverConfig(update.Config)
	if err != nil {
		pkglog.Errorf("Multi-Region Failover update unmarshal failed: %s", err)
		applyStateCallback(cfgPath, state.ApplyStatus{
			State: state.ApplyStateError,
			Error: err.Error(),
		})
		continue
	}

	if mrfUpdate == nil || mrfUpdate.FailoverAPM == nil {
		continue
	}

	if mrfUpdate.FailoverAPM != nil {
		failoverAPMCfgPth = cfgPath
		h.agentConfig.MRFFailoverAPMRC = mrfUpdate.FailoverAPM
		pkglog.Infof("Setting `multi_region_failover.failover_apm: %t` through remote config", *failoverAPM)
		applyStateCallback(failoverAPMCfgPth, state.ApplyStatus{State: state.ApplyStateAcknowledged})
	}
	
	if *mrfUpdate.FailoverAPM {
		break
	}
}
```

pkg/trace/remoteconfighandler/agent_failover.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+// multiRegionFailoverConfig is a deserialized Multi-Region Failover configuration file
+type multiRegionFailoverConfig struct {
+	FailoverMetrics *bool `json:"failover_metrics"`
+	FailoverLogs    *bool `json:"failover_logs"`
```
GustavoCaso, line 16: Do we care about those in the context of APM?

ichinaski: We do not. I mostly kept them to keep full parity with the core agent implementation. Having 2 different implementations of the remote config client is already confusing. I will comment them out and add a comment there  to avoid deserializing those values 👍

GustavoCaso: Why have two separate implementations?. Can the trace agent not use the RC client used in Core?

pkg/trace/remoteconfighandler/remote_config_handler.go
```diff
@@ -89,13 +91,56 @@ func (h *RemoteConfigHandler) Start() {
...
+				Error: err.Error(),
+			})
+			continue
+		}
+
+		if mrfUpdate == nil || mrfUpdate.FailoverAPM == nil {
+			continue
+		}
+
+		if failoverAPM == nil || *mrfUpdate.FailoverAPM {
```
GustavoCaso, line 125: Are we only updating the value if `*mrfUpdate.FailoverAPM` is true? Could it be the case that we want to update the agentconfig for MRF from true to false?

ichinaski: We would also update the value if `mrfUpdate.Failover` is `false`, in case no other value has been assigned yet (first part of the conditional check). But if a value with `true` is received, that takes precedence. This is also the case with Metrics and Logs implementation of MRF.

GustavoCaso: We rubber 🦆 over Slack and the code account for all possible updates 😄

GustavoCaso (approved):
LGTM

ajgajg1134 (approved):

Hitsuji-M (approved):

PR #37208, opened 2025-05-20
Use docker remote cache on docker image builds

alopezz:
### What does this PR do?

Leverages the Docker remote cache to make our builds more resilient.

The current implementation is expected to have the following behavior:
- Each image variant has their separate cache.
- Pipelines on branches will read from the cache.
- Pipelines on pushes to `main` will read from and write to the cache.
- Nightly pipelines will **not** read from any cache and will write to the cache, ensuring we still get updates via `apt` regularly.

This involved some restructuring of the Dockerfiles to maximize what gets cached.
- Adding a named build context for `artifacts`, so that they don't impact the hash of the "local" context which rarely changes.
- Moving steps depending on the artifacts built on the same pipeline to after steps that involve network access, to ensure these don't invalidate the cache continuously.

### Motivation

[BARX-878](https://datadoghq.atlassian.net/browse/BARX-878)

This has been mostly motivated by recent incidents concerning being throttled when pulling dependencies with Maven. Yet we wanted to avoid creating yet another ad-hoc, manually solution for caching those dependencies, and that's why we're attempting a more generic solution based on using the [registry cache backend](https://docs.docker.com/build/cache/backends/registry/) supported by buildkit.

This solution allows these builds to also be resilient to existing commands that retrieve dependencies from 3rd parties (apt & apk), addressing to some extent some well-documented [long standing concerns](https://github.com/DataDog/architecture/blob/master/rfcs/agent-pipeline-caching/rfc.md#problem). It also accounts for any other 3rd party artifact that we may want to retrieve from docker builds in the future, reducing the need to manually put things in a bucket as an ad-hoc cache.

- The cache is used pretty much everywhere, for example the mvn commands are cached [here](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/951607365/viewer#L173).
- Running all builds and tests.

Once merged, I would manually monitor this for a few days to ensure it's working as expected.

- A big part of the cache size being saved is wasteful, because the Agent artifacts change on every pipeline and are thus not reusable from build to build. I'm not sure whether this can be avoided without more significant changes to the structure of the dockerfiles combined with the use of `--no-cache-filter` (assuming this does indeed affect what gets stored in the cache and not just retrieval). Could be a followup unless we consider the current state unacceptable.
  - Edit: I have another branch with those changes if we want them: https://github.com/DataDog/datadog-agent/compare/main...alopez-caching-optimize-spac
- The use of a single cache tag for storing the cache means that all pipelines share the same cache, which can probably lead to cache misses on branches where the dockerfile differs from main. It's probably not a big deal in practice, because we don't modify Dockerfiles that often; the only concern could be release branches but even there at this point the only risk is a few cache misses (i.e. reverting more or less to the previous behavior). There are ways that this could be worked around but the extra complexity is probably not worth it at this point.
- The cache can't be used for images that depend on images generated during the same pipeline (and tagged as so).
- The caching doesn't provide any significant speedups in builds, as the cached steps are normally fairly fast anyway and the bottlenecks are elsewhere.

### Additional Notes

If the cache is not behaving as expected, it would be fairly easy to disable that without reverting the whole PR (the restructuring is still probably good and has benefits if we ever build these locally).

[BARX-878]: https://datadoghq.atlassian.net/browse/BARX-878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


alopezz:

.gitlab/container_build/docker_linux.yml
```diff
@@ -4,17 +4,52 @@
...
     - |
       if [[ "$BUCKET_BRANCH" == "nightly" && ( "$IMAGE" =~ "ci/datadog-agent/agent" || "$IMAGE" =~ "ci/datadog-agent/cluster-agent" || "$IMAGE" =~ "ci/datadog-agent/cws-instrumentation"  || "$IMAGE" =~ "ci/datadog-agent/otel-agent" ) ]]; then
         export ECR_RELEASE_SUFFIX="-nightly"
       else
         export ECR_RELEASE_SUFFIX=${CI_COMMIT_TAG+-release}
       fi
+    # Caching setup
+    - DOCKER_CACHE_TARGET="${IMAGE}${TAG_SUFFIX}-${ARCH}:cache"
+    # Don't use caching on nightlies, to allow for regular cache invalidation,
+    # and update the cache on both main and nightly builds
```
alopezz, line 17: Question for reviewers: would it also make sense for deploy pipelines to run fully uncached?

.gitlab/container_build/docker_linux.yml
```diff
@@ -4,17 +4,52 @@
...
+        --file $BUILD_CONTEXT/Dockerfile \
+        --tag ${TARGET_TAG} \
+        --label "org.opencontainers.image.created=$(date --rfc-3339=seconds)" \
+        --label "org.opencontainers.image.authors=Datadog <package@datadoghq.com>" \
+        --label "org.opencontainers.image.source=https://github.com/DataDog/datadog-agent" \
+        --label "org.opencontainers.image.version=$(dda inv agent.version)" \
+        --label "org.opencontainers.image.revision=${CI_COMMIT_SHA}" \
+        --label "org.opencontainers.image.vendor=Datadog, Inc." \
+        --label "target=none" \
+        $BUILD_CONTEXT
```
chouquette, line 52: much more readable now, thanks!

.gitlab/container_build/docker_linux.yml
```diff
@@ -4,17 +4,52 @@
   script:
     - TAG_SUFFIX=${TAG_SUFFIX:-}
     - BUILD_ARG=${BUILD_ARG:-}
+    - EXTRA_BUILD_CONTEXT=${ARTIFACTS_BUILD_CONTEXT:+"--build-context artifacts=$ARTIFACTS_BUILD_CONTEXT"}
```
chouquette, line 7: I didn't realize you could have multiple named contexts, and I have to agree that this makes the flow more readable

.gitlab/container_build/docker_linux.yml
chouquette: I guess it does, but maybe with a possibility to override it in case maven is down while we're trying to do a release or similar scearios

.gitlab/container_build/docker_linux.yml
```diff
@@ -4,17 +4,52 @@
...
       if [[ "$BUCKET_BRANCH" == "nightly" && ( "$IMAGE" =~ "ci/datadog-agent/agent" || "$IMAGE" =~ "ci/datadog-agent/cluster-agent" || "$IMAGE" =~ "ci/datadog-agent/cws-instrumentation"  || "$IMAGE" =~ "ci/datadog-agent/otel-agent" ) ]]; then
         export ECR_RELEASE_SUFFIX="-nightly"
       else
         export ECR_RELEASE_SUFFIX=${CI_COMMIT_TAG+-release}
       fi
+    # Caching setup
+    - DOCKER_CACHE_TARGET="${IMAGE}${TAG_SUFFIX}-${ARCH}:cache"
+    # Don't use caching on nightlies, to allow for regular cache invalidation,
+    # and update the cache on both main and nightly builds
+    - CACHE_SOURCE="--cache-from type=registry,ref=${DOCKER_CACHE_TARGET}"
```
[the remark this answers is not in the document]
alopezz: Fixed, thanks!

.gitlab/container_build/docker_linux.yml
alopezz: I just added `--no-cache` for deploy pipelines for now. I think in case of an emergency we could probably just edit the pipeline on a release branch to use the cache, I think failures during releases probably don't happen often enough to add any additional mechanism for now.

chouquette (commented):
Overall LGTM with a few open question, but this looks like a very nice improvement!

chouquette (approved):

PR #37209, opened 2025-05-20
Additional scrubbing for the install_info tool

dustmop:
### What does this PR do?

Adds additional scrubbing to the install_info.tool field.

### Motivation

Normally there shouldn't be sensitive info in this field, but just in case there is, this adds anther layer of defense.

This field is also used in `comp/metadata/host/hostimpl/utils/host.go`, but that already [scrubs its entire payload](https://github.com/DataDog/datadog-agent/blob/1fcd8fd891c313f12ea6e7af8b34bb7aafcfa564/comp/metadata/host/hostimpl/host.go#L150).

Unit tests.


davidor (approved):

comp/metadata/clusteragent/impl/cluster_agent.go
```diff
@@ -127,19 +132,26 @@ func (dca *datadogclusteragent) getPayload() marshaler.JSONMarshaler {
...
+	return scrubString
+}
+
 func (dca *datadogclusteragent) initMetadata() {
 	tool := "undefined"
 	toolVersion := ""
 	installerVersion := ""
 
-	install, err := installinfo.Get(dca.conf)
+	install, err := installinfoGet(dca.conf)
```
hush-hush, line 147: I think `installinfo.Get` is only used for metadata, we could update that method to return scrubbed data instead of all the metadata provider. WDYT ?

Also, should we scrub everything returned, not just the tool ? Just to be on the safe side.

dustmop: Done, scrubbing is now being done in `pkg/util/installinfo` instead. Keeping the tests modified for each call-site in order to validate that they do the right thing.

hush-hush (approved):

PR #37217, opened 2025-05-21
Pin Bazel version

rdesgroppes:
Adding a `.bazelversion` file to the repository tells Bazelisk to use the exact version specified in the file when running in the workspace.

The fact that it's versioned inside the repository will then allow for atomic upgrades of Bazel including all necessary changes.

When `bazelisk` drives `bazel` on CI machines too, one will even test Bazel upgrades via a normal presubmit / pull request.

It will also ensure that users will not try to build the project with a different version of Bazel, which might cause failing builds.

Further hint:
https://github.com/bazelbuild/bazelisk?tab=readme-ov-file#ensuring-that-your-developers-use-bazelisk-rather-than-bazel


rdesgroppes:
Superseded by 73588c7f90d8b526bb925c274159203a1bae3293

PR #37220, opened 2025-05-21
[CWS] support parsing all 6 arguments in on demand events

paulcacheux:
### What does this PR do?

This PR adds support for parsing arguments 5 and 6 in the on-demand code.
In order to do that, this PR refactors a bit the code to try to make sure we use common code as much as possible (shared constants, shared unmarshalling functions).
This increase in the event size also requires the event to be stored in a per-cpu map instead of the stack.

### DIFF
--- a/pkg/security/secl/model/unmarshallers_linux.go
+++ b/pkg/security/secl/model/unmarshallers_linux.go
@@ -1359,13 +1359,14 @@ func (e *SyscallsEvent) UnmarshalBinary(data []byte) (int, error) {
 
 // UnmarshalBinary unmarshalls a binary representation of itself
 func (e *OnDemandEvent) UnmarshalBinary(data []byte) (int, error) {
-	if len(data) < 260 {
+	const eventSize = 4 + 6*OnDemandPerArgSize
+	if len(data) < eventSize {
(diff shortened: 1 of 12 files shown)


pkg/security/secl/model/unmarshallers_linux.go
```diff
@@ -1359,13 +1359,14 @@ func (e *SyscallsEvent) UnmarshalBinary(data []byte) (int, error) {
 
 // UnmarshalBinary unmarshalls a binary representation of itself
 func (e *OnDemandEvent) UnmarshalBinary(data []byte) (int, error) {
-	if len(data) < 260 {
+	const eventSize = 4 + 6*OnDemandPerArgSize
```
spikat, line 1362: Could we make the `6` and/or the `6*OnDemandPerArgSize` a constant?

spikat (approved):
A nit comment. Also, maybe we could add more tests related to args 6 and unitary arg length?

pkg/security/secl/model/unmarshallers_linux.go
spikat: thanks!

PR #37229, opened 2025-05-21
omnibus: globally disable doc/manpages

chouquette:
### What does this PR do?

Stop generating our 3rd party dependencies manpages & docs.

### Motivation

We don't need those, nor do our customers.
No significant package size reduction is expected, but this should still slightly reduce the size of our omnibus git cache

### DIFF
--- a/release.json
+++ b/release.json
@@ -2,7 +2,7 @@
     "base_branch": "main",
     "current_milestone": "7.68.0",
     "dependencies": {
-        "OMNIBUS_RUBY_VERSION": "95630818722a63cb9e6f3163984f363d799633a1",
+        "OMNIBUS_RUBY_VERSION": "34bf3ab7bc6ab0f1a08dee724590dcf0f76a443a",
         "JMXFETCH_VERSION": "0.49.7",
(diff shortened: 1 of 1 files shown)


omnibus/config/software/openssl3.rb
```diff
@@ -97,7 +97,9 @@
   # the crazy platform specific compiler flags at the end.
   configure_args << env["CFLAGS"] << env["LDFLAGS"]
 
-  configure(*configure_args, bin: configure_cmd, env: env, no_build_triplet: true)
+  # We need to specify a bunch of extra options here since openssl's configure is not
+  # the usual autoconf configure but something handmade written in perl
+  configure(*configure_args, bin: configure_cmd, env: env, no_build_triplet: true, install_man: true)
```
alopezz, line 102: I wonder if it might even worth replacing this with a regular `command` to highlight how this is not a "normal `configure` script".

chouquette: done in https://github.com/DataDog/datadog-agent/pull/37321

alopezz (approved):

PR #33190, opened 2025-01-21
[AGENTCFG-146] Feature parity between Python network check and Go network check

rahulkaukuntla:
### What does this PR do?
Adding features to the go (linux) network check that already exist in the [python (linux) network check](https://github.com/DataDog/integrations-core/tree/master/network):

- allowing for the configuration of the procfs_path to potentially be a path besides "/proc" (mimicked the logic from [here](https://github.com/DataDog/datadog-agent/blob/main/pkg/collector/embed_python.go#L41))
- adding general testing coverage to the netstatCounters method, and also collecting from the `net/snmp` file as well.
- adding more TCP, IP, IPExt, UDP, and TCPExt metrics from the list of Python check metrics.
- collecting various ethtool metrics and associated tags (driver name and version)
- collecting conntrack metrics, from a configured conntrack executable, as well as from "nf_conntrack_" files in the `${procfs_path}/sys/net/netfiler` directory.
- requesting send and receive queue metrics for TCP from either ss or netstat
- adding speed and mtu tags to interface metrics

A somewhat non-ideal side effect of this PR is that the network check now needs to make shell commands to get certain metrics:
- conntrack metrics are found by calling `sudo ${conntrackPath} -S`
- send and receive queue metrics for TCP is found by calling `sh -c ss --numeric --tcp --all --ipv${ip version}` or `sh -c netstat -n -u -t -a` if the ss executable couldn't be found.

### Describe how you validated your changes

I added unit tests, and manually ran the check with the new configuration options enabled to ensure that the expected metrics are being reported (using the output of the python check for cross-validation).

PR #5412, opened 2020-04-27
Add missing device name tags

kbogtob:
### What does this PR do?

It adds the new `device_name` tag on places that produced `device` tags.

### Motivation

Homogeneization of devices tags.

### Additional Notes

I tried to be exhaustive but I'm not 100% sure. It should so help me if it is not.

### Describe your test plan

`device_name` tags are produced on docker/containerd/network checks. I'm not sure how to test on series.

PR #37240, opened 2025-05-21
[AGENTRUN-158] Add windows networkv2 check from python

jeremy-hanna:
### What does this PR do?

This PR migrates the functionality in the [windows python network check](https://github.com/DataDog/integrations-core/blob/master/network/datadog_checks/network/check_windows.py) to go behind the `use_networkv2_check` configuration flag with full feature parity

* unit tests
* will add e2e test in follow up PR


pkg/collector/corechecks/net/networkv2/network_windows.go
```diff
@@ -0,0 +1,352 @@
...
+		return err
+	}
+
+	tcp6Stats, err := c.net.TcpStats("tcp6")
+	if err != nil {
+		return err
+	}
+
+	// Create tcp metrics that are a sum of tcp4 and tcp6 metrics
+	if !reflect.ValueOf(tcp4Stats).IsZero() && !reflect.ValueOf(tcp6Stats).IsZero() {
```
pgimalac: I'm not 100% sure but I believe `tcp4Stats` and `tcp6Stats` are pointers to struct, so `IsZero` just means `nil` pointer, I'm guessing that's not what you want to test here ?

Also is it possible that the syscall succeeds but doesn't fill the struct ? I don't think we need any check here

pkg/collector/corechecks/net/networkv2/network_windows.go
```diff
@@ -0,0 +1,352 @@
...
+		}
+		submitMetricsFromStruct(sender, "system.net.tcp", tcpAllStats, tcpStatsMapping)
+	}
+
+	if !reflect.ValueOf(tcp4Stats).IsZero() {
+		submitMetricsFromStruct(sender, "system.net.tcp4", tcp4Stats, tcpStatsMapping)
+	}
+	if !reflect.ValueOf(tcp6Stats).IsZero() {
+		submitMetricsFromStruct(sender, "system.net.tcp6", tcp6Stats, tcpStatsMapping)
+	}
```
pgimalac: Same comment as above

pkg/collector/corechecks/net/networkv2/network_windows.go
```diff
@@ -0,0 +1,352 @@
...
+}
+
+func submitMetricsFromStruct(sender sender.Sender, metricPrefix string, tcpStats *mibTcpStats, tcpStatsMapping map[string]string) {
+	s := reflect.ValueOf(tcpStats).Elem()
+	sType := s.Type()
+	for i := 0; i < s.NumField(); i++ {
+		fieldName := sType.Field(i).Name
+		metricName := metricPrefix + tcpStatsMapping[fieldName]
+		metricValue := s.Field(i).Uint()
+		if strings.HasSuffix(metricName, ".connections") || strings.HasSuffix(metricName, ".current_established") {
```
pgimalac, line 278: Can we use a set similarly to what is done in the Python version ?

pkg/collector/corechecks/net/networkv2/network_windows.go
```diff
@@ -0,0 +1,352 @@
...
+			sender.Gauge(metricName, float64(metricValue), "", nil)
+		} else {
+			sender.Rate(metricName, float64(metricValue), "", nil)
+			sender.MonotonicCount(fmt.Sprintf("%s.count", metricName), float64(metricValue), "", nil)
```
pgimalac: Sending a rate and/or a count is controlled by the configs `collect_rate_metrics` and `collect_count_metrics` in the Python version

pkg/collector/corechecks/net/networkv2/network_windows.go
```diff
@@ -0,0 +1,352 @@
...
+		}
+	}
+}
+
+func (c *NetworkCheck) isDeviceExcluded(deviceName string) bool {
+	for _, excludedDevice := range c.config.instance.ExcludedInterfaces {
+		if deviceName == excludedDevice {
+			return true
+		}
+	}
```
pgimalac: We can use `slices.Contains` instead ?

pkg/collector/corechecks/net/networkv2/network_windows.go
```diff
@@ -0,0 +1,352 @@
...
+	if c.config.instance.ExcludedInterfacePattern != nil {
+		return c.config.instance.ExcludedInterfacePattern.MatchString(deviceName)
+	}
+	return false
+}
+
+func submitInterfaceMetrics(sender sender.Sender, interfaceIO net.IOCountersStat) {
+	tags := []string{fmt.Sprintf("device:%s", interfaceIO.Name), fmt.Sprintf("device_name:%s", interfaceIO.Name)}
```
pgimalac: `device_name` seems to have been deprecated in the Python version ?

releasenotes/notes/add-experimental-windows-networkv2-check-b19ce067314b3c1d.yaml
```diff
@@ -0,0 +1,13 @@
...
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+features:
+  - |
+    Add a networkv2 check that is a port of the Windows Python network check to
+    Go. This version is disabled by default but can be enabled with ``use_networkv2_check``
+    in your configuration.
```
pgimalac: The check is still named `network`, not `networkv2`, and it's not a new check, the PR just adds the Windows version
Not super sure how to best explain that...

We might need to patch the [previous release note](https://github.com/DataDog/datadog-agent/pull/33190/files#diff-c05dd63dc559c750c4230a183109a9af6d74fb97962b8fd17665e3e9f42f6022) to make it clearer...

pkg/collector/corechecks/net/networkv2/network_windows.go
jeremy-hanna: hmm I followed the [behavior from the linux](https://github.com/DataDog/datadog-agent/blob/main/pkg/collector/corechecks/net/networkv2/network.go#L367-L368), I guess this is a bug there too

pgimalac: I didn't review the Linux one so I can't say, but I guess so ?

pkg/collector/corechecks/net/networkv2/network_windows.go
jeremy-hanna: looks to be a bug with the linux v2 check as well

jeremy-hanna: hmm this is a break in behavior compatibility with the current iot network check then?

pgimalac: I guess so, we don't plan to maintain both versions anyway, so I guess iot users will have to align with the new version...
We can discuss it with the team

pkg/collector/corechecks/net/networkv2/network_windows.go
```diff
@@ -0,0 +1,352 @@
...
+func getTcpStats(protocolName string) (*mibTcpStats, error) {
+	var inet uint
+	switch protocolName {
+	case "tcp4":
+		inet = AF_INET
+	case "tcp6":
+		inet = AF_INET6
+	}
+	tcpStats := &mibTcpStats{}
+	r0, _, _ := syscall.SyscallN(procGetTcpStatisticsEx.Addr(), uintptr(unsafe.Pointer(tcpStats)), uintptr(inet), 0)
```
clarkb7:
```suggestion
	r0, _, _ := procGetTcpStatisticsEx.Call(uintptr(unsafe.Pointer(tcpStats)), uintptr(inet))
```

git-thuerk-done (approved):
approved pending the `networkv2` update that is noted

pkg/collector/corechecks/net/networkv2/network_windows.go
jeremy-hanna: it looks like this is for compatibility with some container checks: https://github.com/DataDog/datadog-agent/pull/5412

I discussed with Srdjan in our 1:1 and I think we need to keep it for the linux check since it is only an additional tag but for the windows one I will remove it 👍

pkg/collector/corechecks/net/networkv2/network_windows.go
jeremy-hanna: fixed in https://github.com/DataDog/datadog-agent/pull/37240/commits/a8f06404638e4898f0962b431c7e75a65bd21ed9#diff-e88657054ac54654934fe01fd1722599b882419fb419c62b9d4a8c861ffe9dabR181

pkg/collector/corechecks/net/networkv2/network_windows.go
jeremy-hanna: fixed in https://github.com/DataDog/datadog-agent/pull/37240/commits/a8f06404638e4898f0962b431c7e75a65bd21ed9#diff-f8bfa24f1e3f0888eca542771acc5c5a2fa394dacf5e279c0c0955c8d97fdb41R212

releasenotes/notes/add-experimental-windows-networkv2-check-b19ce067314b3c1d.yaml
jeremy-hanna: fixed in https://github.com/DataDog/datadog-agent/pull/37240/commits/a8f06404638e4898f0962b431c7e75a65bd21ed9#diff-8c5d3577c2bade1bdf78db032b2962213c184297ddb80606278ab85dd3264fbdR11

pkg/collector/corechecks/net/networkv2/network_windows.go
jeremy-hanna: fixed in https://github.com/DataDog/datadog-agent/pull/37240/commits/a8f06404638e4898f0962b431c7e75a65bd21ed9#diff-f8bfa24f1e3f0888eca542771acc5c5a2fa394dacf5e279c0c0955c8d97fdb41L302

pkg/collector/corechecks/net/networkv2/network_windows.go
jeremy-hanna: fixed in https://github.com/DataDog/datadog-agent/pull/37240/commits/a8f06404638e4898f0962b431c7e75a65bd21ed9#diff-f8bfa24f1e3f0888eca542771acc5c5a2fa394dacf5e279c0c0955c8d97fdb41R281-R286

I'm unclear if this sets the correct default values but will confirm

pkg/collector/corechecks/net/networkv2/network_windows.go
jeremy-hanna: confirmed that this struct will always be populated on the syscall

fixed in https://github.com/DataDog/datadog-agent/pull/37240/commits/a8f06404638e4898f0962b431c7e75a65bd21ed9#diff-f8bfa24f1e3f0888eca542771acc5c5a2fa394dacf5e279c0c0955c8d97fdb41L242

pkg/collector/corechecks/net/networkv2/network_windows.go
jeremy-hanna: fixed in https://github.com/DataDog/datadog-agent/pull/37240/commits/a8f06404638e4898f0962b431c7e75a65bd21ed9#diff-f8bfa24f1e3f0888eca542771acc5c5a2fa394dacf5e279c0c0955c8d97fdb41L242

pkg/collector/corechecks/net/networkv2/network_windows.go
```diff
@@ -0,0 +1,357 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+//go:build windows
+
+//nolint:revive // TODO(PLINT) Fix revive linter
```
pgimalac, line 8: Can you fix this lint ?

pkg/collector/corechecks/net/networkv2/network_windows.go
```diff
@@ -0,0 +1,357 @@
...
+		return true
+	}
+	if c.config.instance.ExcludedInterfacePattern != nil {
+		return c.config.instance.ExcludedInterfacePattern.MatchString(deviceName)
+	}
+	return false
+}
+
+func submitInterfaceMetrics(sender sender.Sender, interfaceIO net.IOCountersStat) {
+	tags := []string{fmt.Sprintf("device:%s", interfaceIO.Name)}
```
pgimalac, line 302: The Python version also uses tags from the `tags` field of the instance config, is that done under the hood in sender ?

pkg/collector/corechecks/net/networkv2/network_windows.go
```diff
@@ -0,0 +1,357 @@
...
+
+	tcp4Stats, err := c.net.TcpStats("tcp4")
+	if err != nil {
+		return err
+	}
+
+	tcp6Stats, err := c.net.TcpStats("tcp6")
+	if err != nil {
+		return err
+	}
```
pgimalac: The Python version doesn't quite do that, if I'm not mistaken if one fails then it is skipped, and the `tcp` metrics are only sent if both succeed, but we don't fail the check

pkg/collector/corechecks/net/networkv2/network_windows.go
```diff
@@ -0,0 +1,357 @@
...
+		if !c.isDeviceExcluded(interfaceIO.Name) {
+			submitInterfaceMetrics(sender, interfaceIO)
+		}
+	}
+
+	// _tcp_stats
+	err = c.submitTcpStats(sender)
+	if err != nil {
+		return err
+	}
```
pgimalac: Similarly if we fail to send some metrics we probably want to log the error but not fail the whole run of the check ?

releasenotes/notes/add-experimental-windows-networkv2-check-b19ce067314b3c1d.yaml
```diff
@@ -0,0 +1,13 @@
...
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+features:
+  - |
+    Add a port of the Windows integrations-core Python network check to Go. This
+    version is disabled by default but can be enabled with ``use_networkv2_check``
```
pgimalac, line 12: Technically it also requires making the Go loader have priority, or setting `loader: core`
That's going to be complex to explain 😄

pgimalac (approved):
Some minor comments but overall LGTM

pkg/collector/corechecks/net/networkv2/network_windows.go
jeremy-hanna: yes these are set on the sender in the [CheckBase](https://github.com/DataDog/datadog-agent/blob/main/pkg/collector/corechecks/checkbase.go#L131) then included whenever [we send a metric](https://github.com/DataDog/datadog-agent/blob/main/pkg/aggregator/sender.go#L186)

releasenotes/notes/add-experimental-windows-networkv2-check-b19ce067314b3c1d.yaml
jeremy-hanna: This is how we phrase the [diskv2 check release notes](https://github.com/DataDog/datadog-agent/blob/main/releasenotes/notes/disk_feature_parity-f26158df4ead7b63.yaml#L13), since it is still experimental I think this is OK

pkg/collector/corechecks/net/networkv2/network_windows.go
jeremy-hanna: This is the current behavior of the existing network check in go, the linux networkv2 check and what happens when encountering errors with the other checks like disk and diskv2 - I think the error handling from the python side is relying on raised exceptions to fail the check. We can see if this is ever a significant cause of failures during testing.

pkg/collector/corechecks/net/networkv2/network_windows.go
jeremy-hanna: The linux networkv2 check also needs this fixed, I'll clean this up in a separate PR so we can get this in the release for testing

pkg/collector/corechecks/net/networkv2/network_windows.go
jeremy-hanna: I guess if we are just logging the errors from [the syscall](https://github.com/DataDog/datadog-agent/pull/37240#discussion_r2138287208), we will never return an error here

fixed in https://github.com/DataDog/datadog-agent/pull/37240/commits/e296a656ee9a448b4034dc1b444b59096c91ba55

pkg/collector/corechecks/net/networkv2/network_windows.go
jeremy-hanna: fixed in https://github.com/DataDog/datadog-agent/pull/37240/commits/1bd8c0d71d2d93734ba98e0b8eb3427a44babc10

PR #37247, opened 2025-05-21
[AGENTRUN-163] Transition to ipc.HTTPClient for secure-by-default IPC

misteriaud:
### What does this PR do?

This PR transitions from the legacy `GetClient()` function (from `pkg/api/util`) to the `ipc.HTTPClient` (introduced in #35923).
The new `ipc.HTTPClient` streamlines authentication by not requiring prior calls to `SetAuthToken()` or `CreateOrSetAuthToken()`, thus ensuring the necessary authentication artifacts are initialized.

This migration includes the following architectural adjustments:

* The underlying `http.Client` for `ipc.HTTPClient` is now shared across all IPC call locations. Consequently, client timeouts are configured per IPC request (e.g., as an option to `Get()`) instead of at the client instance level.
* The `pkg/config/settings` package now provides two client implementations:
    * `NewHTTPClient`: For SystemProbe, utilizing UDS/NamedPipe (which doesn't require authentication artifacts).
    * `NewHTTPSClient`: For all other TCP-based communications.

### Motivation

The primary goal is to eliminate the use of the `InsecureSkipVerify: true` option in HTTPS clients.

The legacy `pkg/api/util` solution included fallbacks for scenarios where authentication artifacts weren't properly initialized. These were implemented to ensure a smooth transition away from a previously less secure configuration.

By utilizing Fx dependency injection, we can now remove these fallback mechanisms, resulting in an IPC client that is secure by default.

In most cases, this PR should not alter the output of affected commands.
However, due to stricter authentication enforcement, commands that were partially functional (e.g., working without full authentication) might now fail to send IPC requests if authentication artifacts are missing or improperly initialized.

Please perform **before/after tests** with relevant commands under the following conditions:

1.  Agent is running.
2.  Agent is stopped.
3.  Agent has never started (simulate by stopping the Agent service and removing auth artifacts: `/etc/datadog-agent/auth_token` and `/etc/datadog-agent/ipc_cert.pem`).

Authentication artifact validation now occurs during `IPC.Component` initialization as part of the Fx dependency graph construction.

As a result, if authentication artifacts are absent when a command needing `ipc.HTTPClient` is executed (e.g., if a command is run before the Agent has started for the first time), the Fx application will terminate before the command's function is called. This is generally the intended behavior for enhanced security.

For commands that must run irrespective of the Agent's state (e.g., to allow fallback to local processing), we recommend using the `ipcfx.ModuleInsecure()` implementation. This module's construction will not fail due to missing auth artifacts. This approach is particularly useful for debugging commands like `flare` and `diagnose`, enabling them to function even if the Agent is unreachable.


julien-lebot (approved):
✅ for Windows Product files (systray changes)

L3n41c (approved):

adel121 (approved):
OK for #container-platform owned files

paulcacheux (approved):

rahulkaukuntla (approved):

vickenty (approved):

truthbk (approved):
Looking good for DDOT and configsync 👍

jennchenn (approved):
lgtm for @DataDog/container-autoscaling owned file

brycekahle (approved):

soberpeach (approved):

ameske (approved):

comp/trace/config/remote.go
```diff
@@ -21,8 +21,8 @@ func remote(c corecompcfg.Component, ipcAddress string, ipc ipc.Component) (conf
 	return rc.NewGRPCClient(
 		ipcAddress,
 		pkgconfigsetup.GetIPCPort(),
-		func() (string, error) { return ipc.GetAuthToken(), nil }, // TODO IPC: GRPC client will be provided by the IPC component
-		ipc.GetTLSClientConfig,
+		ipc.GetAuthToken(), // TODO IPC: GRPC client will be provided by the IPC component
+		ipc.GetTLSClientConfig(),
```
ichinaski, line 25: Could you also add it [here](https://github.com/DataDog/datadog-agent/blob/main/comp/trace/config/remote.go#L47) ?

(This was just recently merged, and it has the very same code as this function)

TCheruy (approved):

daniel-taf (approved):

NouhaManai96 (approved):

comp/trace/config/remote.go
misteriaud: Sure ! Done in eca27e30255b1cc77aa9e45b2f141614b769ce1d.
Thanks for notifying it.

jeremy-hanna (approved):

ichinaski (approved):

comp/core/tagger/impl-remote/remote_test.go
```diff
@@ -41,13 +40,6 @@ func TestNewComponent(t *testing.T) {
 		t.Skip("Skipping test on macOS runners with an existing Agent.")
 	}
 
-	ipcComp := ipcmock.New(t)
-
-	// Start a mock gRPC server.
-	grpcServer, err := grpc.NewMockGrpcSecureServer("5001", ipcComp.GetAuthToken(), ipcComp.GetTLSServerConfig())
-	require.NoError(t, err)
-	defer grpcServer.Stop()
-
```
adel121: Removing this makes this test identical to the test just below. 

The comment above this test says it is testing the tagger can initialize and start. But as you mentioned, it looks like we are not testing that it can start, which makes it identical to the test below (i.e. `TestNewComponentNonBlocking` ) when you remove the grpc server.

wiyu (approved):

PR #37251, opened 2025-05-21
Update url to apttesting and yumtesting to use s3 url

KevinFairise2:
### What does this PR do?

Stop relying on Cloudfront when downloading the agent. It should reduce NAT Gateway costs.
This is a follow-up PRs of https://github.com/DataDog/datadog-agent/pull/37062
These URLs are not used for OCI so using the S3 url directly works fine


chouetz (approved):

chouquette (approved):

BaptisteFoy (approved):
LGTM if CI is green

tbavelier (approved):
LGTM, but any reason we use `http` instead of `https` ?

aboitreaud (approved):

julien-lebot (approved):

KevinFairise2:
> LGTM, but any reason we use `http` instead of `https` ?

I have no idea, we were using http with the cloudfront so I kept http. It should work the same with https I think

PR #37257, opened 2025-05-21
(fleet) support upgrades on deb/rpm

arbll:
This PR adds support for upgrades on deb/rpm. Now that we have converged on package scripts this is fairly straightforward. All we need is to:
- Create the right symlinks for the installer to track the deb/rpm packages
- Ensure the repository cleanup supports symlinked installation directories

On disk this looks like the following:
```bash
ls -al /opt/datadog-packages/datadog-agent/
total 12
drwxr-xr-x 3 root     root     4096 May 23 09:12 .
drwxr-xr-x 5 root     root     4096 May 23 09:12 ..
drwxr-xr-x 9 dd-agent dd-agent 4096 May 23 09:12 7.65.2-1 -> /opt/datadog-agent
lrwxrwxrwx 1 root     root       42 May 23 09:12 experiment -> /opt/datadog-packages/datadog-agent/stable
lrwxrwxrwx 1 root     root       44 May 23 09:12 stable -> /opt/datadog-packages/datadog-agent/7.65.2-1
```


BaptisteFoy (approved):
LGTM. Some tests are failing because the stable symlink now always exist (we use it to check whether a package is installed by the installer -- we should probably resolve the path & check the prefix)

omnibus/config/software/datadog-agent-installer-symlinks.rb
```diff
@@ -0,0 +1,21 @@
...
+
+description "Create symlinks for the datadog-agent installer to track deb/rpm packages as installed"
+
+always_build true
+
+build do
+  license :project_license
+
+  block do
+    if linux_target? and install_dir == '/opt/datadog-agent'
```
chouquette, line 11: Isn't this logic inverted? I'd have expected the links to be created when the install dir targets the installer's location (ie `/opt/datadog-packages`) instead of the "normal" install_dir

chouquette: ah no OK got it, nevermind the question

chouquette (approved):

CelianR (approved):

arbll:

arbll:

arbll:

PR #37258, opened 2025-05-21
[AGNTLOG-166] Auto multiline detection is customizable on a per-source basis

soberpeach:
### What does this PR do?

Allows integrations to individually configure options for the auto multline feature. The settings override the options set in the universal `datadog.yaml` file while the samples add to the ones in `datadog.yaml`.

### Motivation

Allow more flexibility for using auto multi line detection with integrations.

Run the following one liner to write multi line logs:

```bash
while true; do echo -e "ERROR [DatabaseService] Failed to execute query: invalid column 'user_status'\n  Caused by: org.postgresql.util.PSQLException: ERROR: column \"user_status\" does not exist\n    at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2572)\n    at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2304)\n    at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:348)\n    at org.postgresql.jdbc.PgStatement.executeInternal(PgStatement.java:448)\n    at org.postgresql.jdbc.PgStatement.execute(PgStatement.java:374)\n    at com.example.myapp.repository.UserRepository.getUserStatus(UserRepository.java:75)\n    at com.example.myapp.service.UserService.updateUser(UserService.java:40)\n    at com.example.myapp.controller.UserController.handleUpdateRequest(UserController.java:120)\n    at com.example.myapp.server.RequestDispatcher.dispatch(RequestDispatcher.java:55)"; sleep 3; done >> /tmp/custom_app.log
```
Have the agent tail this file by creating the following integration:

```yaml
logs:
  - type: file
    path: /tmp/custom_app.log
    service: custom_app
    source: custom_app
    auto_multi_line_detection: true
    auto_multi_line_detection_custom_samples:
      - sample: "ERROR [DatabaseService]"
    auto_multi_line:
        enable_json_detection: true
        enable_datetime_detection: true
        tokenizer_max_input_bytes: 50
```

observe that the agent tails these logs and aggregates them as one log.

#### Container Labels

Testing whether container label auto multi line settings are being passed to the auto multi line detector is a little more involved.

1. Create a logging container with container labels specifying `auto_multi_line` settings.

I used the following `app.py` to log and `Dockerfile` to build the container.
`app.py`
```python3
import time
import sys

def print_java_stack_trace_repeatedly():
    """
    Prints a fake Java-like stack trace block to standard error
    every 3 seconds indefinitely.
    """
    stack_trace_block = "SEVERE Main main Exception occurred\njava.lang.Exception: Something bad happened!\nat Main.funcd(Main.java:50)\nat Main.funcc(Main.java:49)\nat Main.funcb(Main.java:48)\nat Main.funca(Main.java:47)\nat Main.main(Main.java:29)"

    print("Starting Java-like stack trace log generation...", file=sys.stdout)
    print("(Press Ctrl+C to stop)", file=sys.stdout)

    try:
        while True:
            # Print the entire block to standard error
            print(stack_trace_block)
            # # Add an extra newline for better separation in the output stream
            # print("", file=sys.stderr)

            # Wait for 3 seconds
            time.sleep(3)

    except KeyboardInterrupt:
        # Handle Ctrl+C gracefully
        print("\nStopping log generation.", file=sys.stdout)
        sys.exit(0)

# Example of how to run the function (when the script is executed directly)
if __name__ == "__main__":
    print_java_stack_trace_repeatedly()
```

`Dockerfile`
```Dockerfile
# Dockerfile
FROM python:3.9-alpine

WORKDIR /app

COPY app.py /app/

# Install any necessary packages (none for this simple script)
# RUN pip install ...

# Command to run the application
CMD ["python", "app.py"]
```

Use the following `docker-compose.yml` to run and add labels to the container.
```yaml
# docker-compose.yml
version: '3.8'

services:
  # Service for our log-generating application
  log-generator:
    build: . # Build the image from the Dockerfile in the current directory
    container_name: my-log-container-compose # Give it a specific container name
    labels:
      com.datadoghq.ad.logs: '[{"source": "docker", "service": "logs-qa","auto_multi_line_detection": true, "auto_multi_line": {"enable_json_detection": false}, "auto_multi_line_detection_custom_samples": [{"sample": "SEVERE Main main Exception occurred"}]}]'
    restart: unless-stopped # Keep the container running unless manually stopped
```

Run *this branch's* agent container using this `docker-compose.yml`
`docker-compose.yml`
```yaml
services:
  datadog-agent:
    container_name: datadog-agent
    image: datadog/agent-dev:lucas-liseth-auto-multi-line-settings-per-integration-py3-jmx
    cgroup: host
    pid: host

    environment:
      # --- Required ---
      # Replace with your actual Datadog API Key, or use a .env file (recommended)
      - DD_API_KEY=${DD_API_KEY}
      # Replace with your Datadog site (e.g., datadoghq.com, datadoghq.eu)
      - DD_SITE=${DD_SITE:-datadoghq.com} # Uses datadoghq.com as default if DD_SITE not set in environment or .env

      # --- Enable Log Collection ---
      - DD_LOGS_ENABLED=true
      - DD_LOGS_CONFIG_CONTAINER_COLLECT_ALL=true
      - DD_CONTAINER_EXCLUDE="name:datadog-agent"
      - DD_LOG_LEVEL=debug
      - DD_CONTAINER_LISTEN_MODE=docker # Agent listens for Docker events for auto-discovery

    volumes:
      - /var/run/docker.sock:/var/run/docker.sock:ro
      - /var/lib/docker/containers:/var/lib/docker/containers:ro
      - /proc/:/host/proc/:ro
      - /sys/fs/cgroup/:/host/sys/fs/cgroup:ro

    restart: unless-stopped
```

Now search the logs produced by the agent container for the log with message matching `Adding source user sample...`. This indicates the user sample in the container label was passed through to the `auto_multi_line` settings.

### Additional Notes

Need to update public docs.


pkg/logs/internal/decoder/auto_multiline_detection/user_samples.go
```diff
@@ -45,24 +46,32 @@ type UserSamples struct {
...
 		}
 
 		if err != nil {
-			log.Error("Failed to unmarshal custom samples: ", err)
+			log.Error("Failed to unmarshal main config custom samples: ", err)
 			s = make([]*UserSample, 0)
 		}
 	}
 
+	for _, sample := range sourceSamples {
```
gh123man, line 67: I think samples are applied in order. Do you think it makes sense to prepend the source samples such that they take precedence over the global samples? That way if a user has a global sample/regex, and then supplies a more specific sample/regex on the source, it will override the global one.

pkg/logs/internal/decoder/auto_multiline_handler.go
```diff
@@ -25,28 +26,56 @@ type AutoMultilineHandler struct {
...
-func NewAutoMultilineHandler(outputFn func(m *message.Message), maxContentSize int, flushTimeout time.Duration, tailerInfo *status.InfoRegistry) *AutoMultilineHandler {
+func NewAutoMultilineHandler(outputFn func(m *message.Message), maxContentSize int, flushTimeout time.Duration, tailerInfo *status.InfoRegistry, sourceSettings *config.SourceAutoMultiLineOptions, sourceSamples []*config.AutoMultilineSample) *AutoMultilineHandler {
 
 	// Order is important
 	heuristics := []automultilinedetection.Heuristic{}
 
-	heuristics = append(heuristics, automultilinedetection.NewTokenizer(pkgconfigsetup.Datadog().GetInt("logs_config.auto_multi_line.tokenizer_max_input_bytes")))
-	heuristics = append(heuristics, automultilinedetection.NewUserSamples(pkgconfigsetup.Datadog()))
+	tokenizerMaxInputBytes := pkgconfigsetup.Datadog().GetInt("logs_config.auto_multi_line.tokenizer_max_input_bytes")
+	if sourceSettings != nil && sourceSettings.TokenizerMaxInputBytes != nil {
```
gh123man, line 35: Can we move `sourceHasSettings := sourceSettings != nil` above this, and also use it here? Alternatively, we could expect this function to require a non-nil `sourceSettings` to avoid the `nil` checks.

comp/logs/agent/config/auto_multiline_samples.go
```diff
@@ -0,0 +1,14 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+package config
+
+// AutoMultilineSample defines a sample used to create auto multiline detection
+// rules
+type AutoMultilineSample struct {
```
gh123man, line 10: Are we able to use the existing [`UserSamples` struct](https://github.com/DataDog/datadog-agent/blob/main/pkg/logs/internal/decoder/auto_multiline_detection/user_samples.go#L22) instead of creating a new one? 

also - note that `matchThreshold` is not exposed in this new struct. Probably worth also exposing here.

gh123man (commented):
Nice work, this is very clean and looks good!
Just some very minor comments/improvements - don't consider my review blocking. 

One other small suggestion - could we add QA steps for a pod annotation/container label? (ex JSON config)

jhgilbert (approved):

pkg/logs/internal/decoder/auto_multiline_detection/user_samples.go
soberpeach: Yeah that makes sense, changing it now.

pkg/logs/internal/decoder/auto_multiline_handler.go
gh123man: > how could we avoid the nil checks?
We would still ned to do the `sourceSettings.EnableJSONAggregation != nil` checks. 
However you could avoid the `sourceHasSettings := sourceSettings != nil` and `if sourceHasSettings &&` if we set the expectation that this struct is required. 

For example - `tailerInfo` is required. If you pass a `nil` tailer info, we will panic. 
One could argue that the `nil` check is safer, and in this case it won't actually save us that much code, so it may not be worth doing here (especially since I think `source.Config().AutoMultilineSamples` can in fact be nil). But I just wanted to call it out as an alternative.

soberpeach: Ah okay makes sense

pkg/logs/internal/decoder/auto_multiline_detection/user_samples.go
```diff
@@ -45,20 +46,31 @@ type UserSamples struct {
...
+		log.Debugf("Adding source user sample: %+v", sample)
+		s = append(s, &UserSample{
+			Sample:         sample.Sample,
+			Label:          sample.Label,
+			Regex:          sample.Regex,
+			MatchThreshold: sample.MatchThreshold,
+		})
+	}
+
+	rawMainSamples := config.Get("logs_config.auto_multi_line_detection_custom_samples")
```
ddrthall: Checking to make sure this is desired - my gut is that a specific config would override the main config, but I may well be missing some use cases.

comp/logs/agent/config/auto_multiline_samples.go
```diff
@@ -0,0 +1,15 @@
...
```
ddrthall: Would it be possible to move this to integration_config.go, alongside the SourceAutoMultilineOptions struct?

ddrthall: Additionally, would merging this with UserSample make sense? E.g. making it an embedded struct.

soberpeach: Made it an embedded struct 👍

ddrthall (approved):

PR #37278, opened 2025-05-22
[DJM] Bump all djm-pinned injector versions to 0.40.0

aboitreaud:
### What does this PR do?

Use the newly released injector version `v0.40.0` that contains a fix for Databricks arm64 instances.


labbati (approved):
🚀

PR #37287, opened 2025-05-22
Build the agent binary with Bazel, and make it work on CI

alopezz:
**NOTE: This PR is not targeting actual main, but our Bazel PoC branch.**

### What does this PR do?

Primarily, it replaces the `agent.build` invocation in Omnibus with equivalent Bazel build steps, only for Linux and the "base" flavor of the Agent, and sets things up so that it does so correctly in CI. All other changes are directed towards that goal and having as much of CI green as possible. It also enables the bazel remote cache.

### Motivation

[BARX-842](https://datadoghq.atlassian.net/browse/BARX-842)

Passing builds and a good number of green e2e tests.

Failing jobs:
- Cluster Agent builds: the extra `ignore` tag that we added to break a circular dependency with Bazel breaks these builds (because this needed file gets ignored). That's something we can try to fix separately.
- `new-e2e-agent-platform-install-script-centos-a7-x86_64` (and FIPS variant): unclear, but since it also happens on the FIPS build, this may be something that might get fixed by updating with main.
- `new-e2e-agent-platform-install-script-debian-a7-x86_64: It fails to detect the Python version. Earlier I was getting this error in all platforms, now it's only present here; postponing this into the future.
- A number of other jobs have unrelated failures that might get fixed once we catch up to main.
- Windows related failures are due to something unrelated that was fixed in https://github.com/DataDog/test-infra-definitions/pull/1525 which we still haven't picked up in this branch.

### Additional Notes

- The caching still shows some unexpected performance behavior that we'll look into later.
- The immediate continuation to this will be to try to bring python and all of its dependencies out of the Bazel build into the package so that we don't build them twice like we currently are.

[BARX-842]: https://datadoghq.atlassian.net/browse/BARX-842?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.bazelrc
```diff
@@ -14,6 +14,16 @@ build:windows --action_env=PATH --define=MINGW_PATH=c:/tools/msys64/mingw64 --de
 # Don't fallback to local toolchains, only use the hermetic ones
 #build --action_env BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN=1
 
+# Caching
+build:ci --config=cache
+common:ci --config=cache
```
rdesgroppes, line 19: Theoretically, `build` settings "inherit" `common` settings. Isn't it the case for configs, i.e. doesn't `build:ci` inherit `common:ci`?

.bazelrc
```diff
@@ -14,6 +14,16 @@ build:windows --action_env=PATH --define=MINGW_PATH=c:/tools/msys64/mingw64 --de
...
+common:cache --remote_instance_name=remote-caching
+common:cache --remote_cache=grpc://bazel-buildbarn-frontend.ddbuild.io:443
+common:cache --remote_timeout=60
+common:cache --remote_retries=1
+
+build:cache --remote_instance_name="remote-caching"
```
rdesgroppes, line 25: Same question: doesn't `build:cache` inherit `common:cache`?

.bazelrc
alopezz: I didn't even plan to commit these files yet 😅 but I was planning to take @chouquette 's work on https://github.com/DataDog/datadog-agent/tree/chouquette/bazel/remote-caching as a starting point to enable remote caching in a second commit.

As for the comment specifically, I think you're probably right here.

chouquette: To be honest this is copied from https://datadoghq.atlassian.net/wiki/spaces/EEEUSG/pages/2319384594/Bazel+Caching with not much thought put into it besides checking that it works for our purposes :sweat_smile:

alopezz: I've added the lines now "officially" in a separate commit, my intent is that on CI we write to the cache while locally we'll only read.

deps/repos.MODULE.bazel
```diff
@@ -82,14 +82,14 @@ http_archive(
     url = "https://github.com/besser82/libxcrypt/releases/download/v{0}/libxcrypt-{0}.tar.xz".format(LIBXCRYPT_VERSION),
 )
 
-LIBFFI_VERSION = "3.4.6"
+LIBFFI_VERSION = "3.2.1"
```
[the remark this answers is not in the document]
alopezz: This was necessary because datadog-agent-three links against it, and the health check would complain that it wouldn't find the right library file. Granted, this will stop being a problem when we start actually adding the bazel-built libraries to the package, but I prefer to go step by step.

As far as I can tell I did get it to work, though, it only required adjusting the libdir in a couple of places. But I'm fine with updating this on Omnibus, but let's do that in the actual main before updating it here.

rtloader/BUILD.bazel
```diff
@@ -47,6 +46,20 @@ cc_binary(
...
+    srcs = ["three/constants.h.in"],
+    outs = ["constants.h"],
+    cmd_bash = """
+    sed -E 's/@Python3_STDLIB@/../g' $< > $@
+    """,
+)
+
+cc_library(
+    name = "constants_header",
+    hdrs = [":constants_h"],
```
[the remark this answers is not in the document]
chouquette: alright, thanks for the details!

rdesgroppes:
:+1: for merging this increment.

PR #37288, opened 2025-05-22
[CXP-2184] [agent] add new disabled process collector with basic logic

gengnamstyle:
### What does this PR do?
- adds a new **DISABLED** process collector with basic collection logic as part of the service discovery/process collection consolidation 
- main implementation functions are `collect` and `stream`
    - `collect`
        - calls probe to get process information (duplicate of process check, which will be changed in the future)
        - parses information to send to channel for async processing 
            - I believe as little work should be done in the `stream` function which handles all the events (bottleneck)
    - `stream`
        - receives events and transforms into workload meta events to then store + notify subscribers
- updates workloadmeta process entity by adding new fields
    - should not impact existing behaviour

### Motivation

- added unit tests

- workload meta entity will expose new fields but nothing will be populated, potentially misleading for devs who try to work with the new struct

### Additional Notes

- this collector is default **DISABLED** right now as it is not production ready


comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -0,0 +1,267 @@
...
+				proc, exists := procs[pid]
+
+				// old process was deleted
+				if !exists {
+					deletedProcesses = append(deletedProcesses, processToWorkloadMetaProcess(oldProc))
+				} else if exists && oldProc.Stats.CreateTime != proc.Stats.CreateTime {
+					// old process but same PID
+					deletedProcesses = append(deletedProcesses, processToWorkloadMetaProcess(oldProc))
+				}
+			}
```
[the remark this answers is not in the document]
daniel-taf: Also, the logic seems to be very similar between determining deleted and created processes. Could we leverage that to have less code? Maybe also worth using the hashing mechanism from [here](https://github.com/DataDog/datadog-agent/blob/a939f80358798f45f90c66e7be78a8dc067ef936/pkg/process/metadata/workloadmeta/extractor.go#L111-L127).

comp/core/workloadmeta/collectors/internal/process/process_collector_test.go
```diff
@@ -0,0 +1,557 @@
...
+						CPUPercent:  nil,
+						CPUTime:     nil,
+						MemInfo:     nil,
+						MemInfoEx:   nil,
+						IOStat:      nil,
+						IORateStat:  nil,
+						CtxSwitches: nil,
+					},
+				},
+			},
```
daniel-taf, line 103: I think we could have a function that creates these test processes instead so we don't have to paste them everywhere like [here](https://github.com/DataDog/datadog-agent/blob/main/pkg/process/metadata/workloadmeta/extractor_test.go#L30-L37).

comp/core/workloadmeta/collectors/internal/process/process_collector_test.go
```diff
@@ -0,0 +1,557 @@
...
+	processCollector := newProcessCollector(collectorID, mockStore, workloadmeta.NodeAgent, mockClock, mockProbe, processEventCh, make(map[int32]*procutil.Process))
+
+	return collectorTest{&processCollector, mockProbe, mockClock, mockStore}
+}
+
+// TestCreatedProcessesCollection tests the collector capturing new processes
+func TestCreatedProcessesCollection(t *testing.T) {
+	originalFlavor := flavor.GetFlavor()
+	defer flavor.SetFlavor(originalFlavor)
+	flavor.SetFlavor(flavor.DefaultAgent)
```
daniel-taf, line 61: We don't need to check the flavor in this collector.

comp/core/workloadmeta/collectors/internal/process/process_collector_test.go
```diff
@@ -0,0 +1,557 @@
...
+					Name:    "some name",
+					Cwd:     "some_directory/path",
+					Exe:     "test",
+					Comm:    "",
+					Cmdline: []string{"c1", "c2", "c3"},
+					Uids:    []int32{1, 2, 3, 4},
+					Gids:    []int32{1, 2, 3, 4, 5},
+					// time is created like this so it has the same precision as the
+					//CreationTime: time.UnixMilli(creationTime1.UnixMilli()).UTC(),
+					CreationTime: time.UnixMilli(creationTime1).UTC(),
```
daniel-taf, line 123: Could you explain more about this?

comp/core/workloadmeta/collectors/internal/process/process_collector_test.go
```diff
@@ -0,0 +1,557 @@
...
+			c.probe.On("ProcessesByPID", mock.Anything, mock.Anything).Return(tc.processesToCollectA, nil).Times(1)
+			// update clock to trigger processing
+			c.mockClock.Add(collectionInterval)
+
+			assert.EventuallyWithT(t, func(cT *assert.CollectT) {
+				for pid, expectedProc := range tc.expectedProcessesA {
+					actualProc, err := c.mockStore.GetProcess(pid)
+					assert.NoError(cT, err)
+					assert.Equal(cT, expectedProc, actualProc)
+				}
```
daniel-taf, line 537: I think using [ListProcesses](https://github.com/DataDog/datadog-agent/blob/main/comp/core/workloadmeta/def/component.go#L97) is more appropriate here.

comp/core/workloadmeta/collectors/internal/process/process_collector_test.go
```diff
@@ -0,0 +1,557 @@
...
+		// start
+		processesToCollectA map[int32]*procutil.Process
+		expectedProcessesA  map[int32]*workloadmeta.Process
+
+		// end
+		processesToCollectB map[int32]*procutil.Process
+		expectedProcessesB  map[int32]*workloadmeta.Process
+	}{
+		{
+			description:     "2 new processes and 1 finishes",
```
daniel-taf, line 272: Lets add a case where a new process has the same pid as a deleted one.

comp/core/workloadmeta/collectors/internal/process/process_collector_test.go
```diff
@@ -0,0 +1,557 @@
...
+					Uids:         []int32{},
+					Gids:         []int32{},
+					CreationTime: time.UnixMilli(creationTime2).UTC(),
+				},
+			},
+			processesToCollectB: map[int32]*procutil.Process{},
+			expectedProcessesB: map[int32]*workloadmeta.Process{
+				1234: nil,
+				9999: nil,
+			},
```
daniel-taf, line 516: If you just need the pids and liveness, lets use a slice of structs with that information or two different slices (live vs dead). I think that makes the test more clear

comp/core/workloadmeta/def/types.go
```diff
@@ -1199,10 +1200,20 @@ var _ Entity = &ContainerImageMetadata{}
...
+	Cwd          string
+	Exe          string
+	Comm         string
+	Cmdline      []string
+	Uids         []int32
+	Gids         []int32
 	ContainerID  string
 	CreationTime time.Time
 	Language     *languagemodels.Language
+	// TODO: add new fields
```
daniel-taf, line 1216: Lets add some further expansion on this comment about whats left.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -0,0 +1,267 @@
...
+			// categorize the processes into storable events
+			var createdProcesses []*workloadmeta.Process
+			var deletedProcesses []*workloadmeta.Process
+
+			// determine new processes
+			for pid, proc := range procs {
+				if proc == nil {
+					log.Warnf("collected process %d was nil", pid)
+					continue
+				}
```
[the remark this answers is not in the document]
gengnamstyle: I don't want to assume proc is non nil for safety. 

Looking at how we currently get these procs, it looks like this case would not happen, but I feel like it's not right to make these assumptions, but I'm okay with removing this check as well.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -0,0 +1,267 @@
...
+			return
+		}
+	}
+}
+
+// processToWorkloadMetaProcess maps a procutil process to a workloadmeta process
+func processToWorkloadMetaProcess(process *procutil.Process) *workloadmeta.Process {
+	if process == nil {
+		return nil
+	}
```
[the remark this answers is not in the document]
gengnamstyle: Answered for similar question above, but posting for ease:

I don't want to assume proc is non nil for safety.

Looking at how we currently get these procs, it looks like this case would not happen, but I feel like it's not right to make these assumptions, but I'm okay with removing this check as well.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -0,0 +1,267 @@
...
+	defer cancel()
+	for {
+		select {
+		case <-health.C:
+
+		case processEvent := <-c.processEventsCh:
+			if processEvent == nil {
+				log.Warn("sent process Event was nil")
+				continue
+			}
```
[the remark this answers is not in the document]
gengnamstyle: Answered for similar question above, but posting for ease:

I don't want to assume proc is non nil for safety.

Looking at how we currently get these procs, it looks like this case would not happen, but I feel like it's not right to make these assumptions, but I'm okay with removing this check as well.

comp/core/workloadmeta/collectors/internal/process/process_collector_test.go
gengnamstyle: added

comp/core/workloadmeta/def/types.go
gengnamstyle: updated as 
```
// TODO: add future fields for privileged/service discovery data
```

comp/core/workloadmeta/collectors/internal/process/process_collector_test.go
gengnamstyle: ah removed

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -0,0 +1,267 @@
...
+
+			// TODO: implement the logic to handle events
```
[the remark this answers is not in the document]
gengnamstyle: removed

comp/core/workloadmeta/collectors/internal/process/process_collector.go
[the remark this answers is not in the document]
gengnamstyle: Updated. I initially strayed away from the hashing method as it seemed less clear to me. I created a helper function to reduce the duplicated code.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -0,0 +1,267 @@
...
+			// Fetch process data and submit events to streaming channel for asynchronous processing
+			procs, err := c.processProbe.ProcessesByPID(time.Now(), false)
+			if err != nil {
+				log.Errorf("Error getting processes by pid: %v", err)
+				return
+			}
+
+			// categorize the processes into storable events
+			var createdProcesses []*workloadmeta.Process
+			var deletedProcesses []*workloadmeta.Process
```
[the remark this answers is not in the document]
gengnamstyle: updated

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -0,0 +1,267 @@
...
+
+// GetFxOptions returns the FX framework options for the collector
+func GetFxOptions() fx.Option {
+	return fx.Provide(NewProcessCollectorProvider)
+}
+
+// isEnabled returns a boolean indicating if the process collector is enabled and what collection interval to use if it is.
+func (c *collector) isEnabled() (bool, time.Duration) {
+	// TODO: implement the logic to check if the process collector is enabled based on dependent configs (process collection, language detection, service discovery)
+	return false, time.Second * 10
```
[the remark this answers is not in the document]
gengnamstyle: added

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -0,0 +1,267 @@
...
+	clock clock.Clock
+
+	// fetches process data
+	processProbe procutil.Probe
+
+	// channel for async processing of events
+	processEventsCh chan *Event
+
+	// cache of last collect processes for diff generation
+	lastCollectedProcesses map[int32]*procutil.Process
```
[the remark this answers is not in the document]
gengnamstyle: removed

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -0,0 +1,267 @@
...
+const (
+	collectorID       = "process-collector"
+	componentName     = "workloadmeta-process"
+	cacheValidityNoRT = 2 * time.Second
+)
+
+type collector struct {
+	id      string
+	store   workloadmeta.Component
+	catalog workloadmeta.AgentType ``
```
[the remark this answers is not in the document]
gengnamstyle: removed

comp/core/workloadmeta/collectors/internal/process/process_collector_test.go
gengnamstyle: removed

comp/core/workloadmeta/collectors/internal/process/process_collector.go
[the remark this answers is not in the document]
gengnamstyle: I'll remove these checks, but I do feel like code is always subject to change and I like to err on the side of caution.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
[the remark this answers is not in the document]
gengnamstyle: will remove

comp/core/workloadmeta/collectors/internal/process/process_collector.go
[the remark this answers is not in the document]
gengnamstyle: I think this check should stay as I believe the functionality of this function mapping from 1 struct to another should account for a nil input and output nil as well.

comp/core/workloadmeta/collectors/internal/process/process_collector_test.go
gengnamstyle: updated with a helper function

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -0,0 +1,237 @@
...
+// GetFxOptions returns the FX framework options for the collector
+func GetFxOptions() fx.Option {
+	return fx.Provide(NewProcessCollectorProvider)
+}
+
+// isEnabled returns a boolean indicating if the process collector is enabled and what collection interval to use if it is.
+func (c *collector) isEnabled() (bool, time.Duration) {
+	// TODO: implement the logic to check if the process collector is enabled based on dependent configs (process collection, language detection, service discovery)
+	// hardcoded to false until the new collector has all functionality/consolidation completed (service discovery, language collection, etc)
+	return false, time.Second * 10
```
wiyu, line 77: [minor] Instead of returning an interval as it introduces an additional responsiblity for the method, let's store it as an attribute on the collector.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -0,0 +1,237 @@
...
+		} else if procB.Stats.CreateTime != procA.Stats.CreateTime {
+			// same process PID exists, but different process due to creation time
+			newProcs = append(newProcs, procA)
+		}
+	}
+	return newProcs
+}
+
+// collect captures all the required process data for the process check
+func (c *collector) collect(ctx context.Context, collectionTicker *clock.Ticker) {
```
wiyu, line 115: Instead of taking in the ticker, let's just create it in this method.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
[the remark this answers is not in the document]
wiyu: I would agree with @daniel-taf on this one. If you look across the codebase for the usage of `*procutil.Process`, you'll notice we don't perform nil checks as the [probe](https://github.com/DataDog/datadog-agent/blob/deb796fa8fd53a5673c852ee233666afdac1b4cb/pkg/process/procutil/process_linux.go#L281-L314) does not return nil processes. The pointer usage here is to signify pass by reference semantics, not to indicate the value is optional. I think adding the nil check is a good defensive measure, but it also introduces ambiguity of the semantics when we look across the code base due to the inconsistency. 

Example: https://github.com/DataDog/datadog-agent/blob/27502336a34eb2602ecb53cab73a460adba984d2/pkg/process/checks/process.go#L651

comp/core/workloadmeta/collectors/internal/process/process_collector.go
daniel-taf: Something to be aware about, creating tickers within a go-routine can lead to non-determinism when trying to use mocked time in a test. Since there is a race condition between the creation of the ticker and time advancement. I used this pattern to avoid this when I made the language collector tests. I think we should keep this so we don't have to deal with potential flakes.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
gengnamstyle: Separated interval into a different function, but decided to not store as an attribute because I don't think it's necessary at the moment, as it is only used once and should not be mutated at runtime.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -0,0 +1,237 @@
...
+		c.store = store
+		go c.collect(ctx, c.clock.Ticker(collectionInterval))
+		go c.stream(ctx)
+	}
+
+	return nil
+}
+
+// getProcessCacheDifference returns new processes that exist in procCacheA and not in procCacheB
+func getProcessCacheDifference(procCacheA map[int32]*procutil.Process, procCacheB map[int32]*procutil.Process) []*procutil.Process {
```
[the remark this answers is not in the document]
gengnamstyle: updated

comp/core/workloadmeta/collectors/internal/process/process_collector_test.go
```diff
@@ -0,0 +1,349 @@
...
+
+		})
+	}
+}
+
+// TestCreatedProcessesCollection tests the collector capturing lifecycle of a process (creation, deletion)
+func TestProcessLifecycleCollection(t *testing.T) {
+	originalFlavor := flavor.GetFlavor()
+	defer flavor.SetFlavor(originalFlavor)
+	flavor.SetFlavor(flavor.DefaultAgent)
```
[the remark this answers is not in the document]
gengnamstyle: removed

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -0,0 +1,237 @@
...
+	// hardcoded to false until the new collector has all functionality/consolidation completed (service discovery, language collection, etc)
+	return false, time.Second * 10
+}
+
+// Start starts the collector. The collector should run until the context
+// is done. It also gets a reference to the store that started it so it
+// can use Notify, or get access to other entities in the store.
+func (c *collector) Start(ctx context.Context, store workloadmeta.Component) error {
+	// TODO: implement the start-up logic for the process collector
+	// Once setup logic is complete, start collection and streaming goroutines
```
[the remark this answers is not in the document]
gengnamstyle: removed

comp/core/workloadmeta/collectors/internal/process/process_collector_test.go
```diff
@@ -0,0 +1,349 @@
...
+	return collectorTest{&processCollector, mockProbe, mockClock, mockStore}
+}
+
+// start used for testing purposes while we wait for configuration logic to be sorted out
+func (c *collector) start(ctx context.Context, store workloadmeta.Component, collectionInterval time.Duration) error {
+	c.store = store
+	go c.collect(ctx, c.clock.Ticker(collectionInterval))
+	go c.stream(ctx)
+	return nil
+}
```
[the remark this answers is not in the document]
gengnamstyle: Yeah i understand. I thought it was a clean way to encapsulate starting the collector when testing as it would be simpler to swap out when the configuration is sorted, but I removed it.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
gengnamstyle: left as is unless there are other objections.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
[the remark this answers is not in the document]
gengnamstyle: all good with removing it. However, I'm curious as to why we aren't as weary of future code changes because I think it may be difficult to refactor all these indirectly coupled assumptions? 

If changes do occur do we take the stance of lets deal with it when it happens?

daniel-taf (approved):

comp/core/workloadmeta/collectors/internal/process/process_collector_nop.go
```diff
@@ -0,0 +1,18 @@
...
+//go:build !linux
+
+// Package process implements the process collector for Workloadmeta.
+package process
+
+import (
+	"go.uber.org/fx"
+)
+
+// GetFxOptions returns the FX framework options for the collector
```
wiyu, line 15: Do we have tests for this?

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -0,0 +1,237 @@
...
+
+	workloadmeta "github.com/DataDog/datadog-agent/comp/core/workloadmeta/def"
+	"github.com/DataDog/datadog-agent/pkg/status/health"
+	"github.com/DataDog/datadog-agent/pkg/util/log"
+)
+
+const (
+	collectorID       = "process-collector"
+	componentName     = "workloadmeta-process"
+	cacheValidityNoRT = 2 * time.Second
```
robertjli, line 28: Nit, optional: this seems unused, should we remove?

I see it's typically passed in to the [ContainerProvider](https://github.com/DataDog/datadog-agent/blob/06e08cf387fb430c0acb1508a8e0b55778d2f6c7/pkg/process/util/containers/containers.go#L59-L63) functions which are not included yet, also I would recommend a more descriptive name (understanding that this is the name used in many other ContainerProvider calls)

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -0,0 +1,237 @@
...
+	lastCollectedProcesses map[int32]*procutil.Process
+}
+
+// Event is a message type used to communicate with the stream function asynchronously
+type Event struct {
+	Created []*workloadmeta.Process
+	Deleted []*workloadmeta.Process
+}
+
+func newProcessCollector(id string, store workloadmeta.Component, catalog workloadmeta.AgentType, clock clock.Clock, processProbe procutil.Probe, processEventsCh chan *Event, lastCollectedProcesses map[int32]*procutil.Process) collector {
```
robertjli, line 47: Suggestion, optional: `processEventsCh` and `lastCollectedProcesses` could be allocated inside this function instead of passed in. In my opinion it makes the function cleaner and you're allocating these objects within or immediately before the call anyway.

comp/core/workloadmeta/collectors/internal/process/process_collector_test.go
```diff
@@ -0,0 +1,341 @@
...
+		},
+	} {
+		t.Run(tc.description, func(t *testing.T) {
+			c := setUpCollectorTest(t, tc.configOverrides)
+			ctx, cancel := context.WithCancel(context.TODO())
+			defer cancel()
+
+			// TODO: we should use Start() instead of 3 lines below when configuration is sorted as Start() is currently
+			// by default disabled
+			c.collector.store = c.mockStore
```
robertjli, line 201: Nit, optional: you already set `collector.store` in the constructor on line 52. I would recommend actually removing `store` from the constructor parameters, as when we are using `Start()` everywhere we won't ever be passing it into the constructor.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -0,0 +1,237 @@
...
+func (c *collector) collectionIntervalConfig() time.Duration {
+	// TODO: read configured collection interval once implemented
+	return time.Second * 10
+}
+
+// Start starts the collector. The collector should run until the context
+// is done. It also gets a reference to the store that started it so it
+// can use Notify, or get access to other entities in the store.
+func (c *collector) Start(ctx context.Context, store workloadmeta.Component) error {
+	enabled := c.isEnabled()
```
robertjli, line 90: Question: why not inline into the if statement?

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -0,0 +1,237 @@
...
+		c.store = store
+		go c.collect(ctx, c.clock.Ticker(c.collectionIntervalConfig()))
+		go c.stream(ctx)
+	}
+
+	return nil
+}
+
+// processCacheDifference returns new processes that exist in procCacheA and not in procCacheB
+func processCacheDifference(procCacheA map[int32]*procutil.Process, procCacheB map[int32]*procutil.Process) []*procutil.Process {
```
robertjli, line 102: Suggestion: since this function relies on "time", i.e. A is "newer" than B, I would rename these params to something like `prev` and `curr`

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -0,0 +1,237 @@
...
+			// same process PID exists, but different process due to creation time
+			newProcs = append(newProcs, procA)
+		}
+	}
+	return newProcs
+}
+
+// collect captures all the required process data for the process check
+func (c *collector) collect(ctx context.Context, collectionTicker *clock.Ticker) {
+	// TODO: implement the full collection logic for the process collector. Once collection is done, submit events.
```
robertjli, line 120: Question: Is this TODO still relevant?

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -0,0 +1,237 @@
...
+	ctx, cancel := context.WithCancel(ctx)
+	defer collectionTicker.Stop()
+	defer cancel()
+	for {
+		select {
+		case <-collectionTicker.C:
+			// Fetch process data and submit events to streaming channel for asynchronous processing
```
robertjli, line 127:
```suggestion
			// fetch process data and submit events to streaming channel for asynchronous processing
```
Nit, optional: for consistency

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -0,0 +1,237 @@
...
+			procs, err := c.processProbe.ProcessesByPID(time.Now(), false)
+			if err != nil {
+				log.Errorf("Error getting processes by pid: %v", err)
+				return
+			}
+
+			// categorize the processes into events for workloadmeta
+			createdProcs := processCacheDifference(procs, c.lastCollectedProcesses)
+			var wlmCreatedProcs []*workloadmeta.Process
```
robertjli, line 136: Suggestion: since you know the length of this will be exactly the length of `createdProcs` you can preallocate the entire buffer. Same with `wlmDeletedProcs` below.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -0,0 +1,237 @@
...
+		go c.collect(ctx, c.clock.Ticker(c.collectionIntervalConfig()))
+		go c.stream(ctx)
+	}
+
+	return nil
+}
+
+// processCacheDifference returns new processes that exist in procCacheA and not in procCacheB
+func processCacheDifference(procCacheA map[int32]*procutil.Process, procCacheB map[int32]*procutil.Process) []*procutil.Process {
+	var newProcs []*procutil.Process
```
robertjli, line 103: Suggestion: you can pre-allocate `abs(len(a) - len(b))` for this slice

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -0,0 +1,237 @@
...
+				})
+			}
+
+			c.store.Notify(events)
+
+		case <-ctx.Done():
+			err := healthCheck.Deregister()
+			if err != nil {
+				log.Warnf("error de-registering health check: %s", err)
+			}
```
robertjli, line 195: Nit: should we also have an INFO log that the stream goroutine has stopped, similar to the collector?

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -0,0 +1,237 @@
...
+func (c *collector) stream(ctx context.Context) {
+	healthCheck := health.RegisterLiveness(componentName)
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	for {
+		select {
+		case <-healthCheck.C:
+
+		case processEvent := <-c.processEventsCh:
+			var events []workloadmeta.CollectorEvent
```
robertjli, line 172: Suggestion: similar to above, we can pre-allocate this slice

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -0,0 +1,237 @@
...
+	return time.Second * 10
+}
+
+// Start starts the collector. The collector should run until the context
+// is done. It also gets a reference to the store that started it so it
+// can use Notify, or get access to other entities in the store.
+func (c *collector) Start(ctx context.Context, store workloadmeta.Component) error {
+	enabled := c.isEnabled()
+
+	if enabled {
```
gabedos, line 92: Nit optional: could we throw an error here if the collector is started when it's not meant to be (similar to the docker, kubelet, etc. collectors).

`if ( ... ) { return errors.NewDisabled(componentName, "Agent is not running Linux Process collector") }`

comp/core/workloadmeta/collectors/internal/process/process_collector.go
gengnamstyle: removed `cacheValidityNoRT` since it's unused as of now. The other variables are actually currently used. It is following a similar pattern to the existing process language collector which I'd like to uphold until we consolidate the language collector.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
gengnamstyle: remnant of this function returning the collection interval as well. Good callout. Inlined.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
gengnamstyle: Added a basic check + log if process collection is disabled. 

As for next steps, there are floating questions about if this collection should enabled by default because of the service discovery consolidation + language detection, so we will likely leave this as is until we are all synced.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
gengnamstyle: Added abs(len(a) - len(b)) as a starting point.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
gengnamstyle: yes still relevant until we finish the container data + language collection

comp/core/workloadmeta/collectors/internal/process/process_collector.go
gengnamstyle: fixed

comp/core/workloadmeta/collectors/internal/process/process_collector.go
gengnamstyle: updated. good optimization thanks.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
gengnamstyle: done.

comp/core/workloadmeta/collectors/internal/process/process_collector_nop.go
gengnamstyle: I don't believe so. It does not appear other components that utilize uber fx have tests after searching for usages. There does appear to be 2 windows tests that call it though.

gabedos (commented):
For debugging, I would recommend adding the new process fields in the wmeta.Process String [method](https://github.com/DataDog/datadog-agent/blob/2a7fea5489c1956adf720c62f7a8b44f44ae2015/comp/core/workloadmeta/def/types.go#L1251). This is both useful for us developers and customer debugging too in order to verify that wmeta has the proper view of an entity. I think most of these attributes could be masked under `verbose`

gabedos (approved):

gengnamstyle:
> For debugging, I would recommend adding the new process fields in the wmeta.Process String [method](https://github.com/DataDog/datadog-agent/blob/2a7fea5489c1956adf720c62f7a8b44f44ae2015/comp/core/workloadmeta/def/types.go#L1251). This is both useful for us developers and customer debugging too in order to verify that wmeta has the proper view of an entity. I think most of these attributes could be masked under `verbose`

Yup! I'm planning to add to the `String` function as linked **later** since I'm unsure of exposing any new fields that aren't certain to stay or mislead anybody as there's a lot more work to be done before enabling this collector.

PR #37294, opened 2025-05-22
[draft] Add grpc endpoint to stream config updates for `RemoteAgent` service

rayz:
### What does this PR do?
[draft] 

This pr adds a grpc endpoint for streaming configuration updates to RemoteAgents.

We establish a stream when a new `RemoteAgent` is registered. When deregistering, we call the cancel func associated with the stream.

### Additional Notes

ADP side pr: https://github.com/DataDog/saluki/pull/713

### DIFF
--- a/comp/core/remoteagentregistry/impl/remoteagentregistry.go
+++ b/comp/core/remoteagentregistry/impl/remoteagentregistry.go
@@ -9,6 +9,7 @@ package remoteagentregistryimpl
 import (
 	"context"
 	"crypto/tls"
+	"encoding/json"
 	"fmt"
 	"strings"
 	"sync"
@@ -16,6 +17,7 @@ import (
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/protobuf/types/known/structpb"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
 	"github.com/prometheus/client_golang/prometheus"
@@ -70,10 +72,11 @@ func NewComponent(reqs Requires) Provides {
 func newRemoteAgent(reqs Requires) *remoteAgentRegistry {
 	shutdownChan := make(chan struct{})
 	comp := &remoteAgentRegistry{
-		conf:         reqs.Config,
-		agentMap:     make(map[string]*remoteAgentDetails),
-		shutdownChan: shutdownChan,
-		telemetry:    reqs.Telemetry,
+		conf:             reqs.Config,
+		agentMap:         make(map[string]*remoteAgentDetails),
+		shutdownChan:     shutdownChan,
+		telemetry:        reqs.Telemetry,
+		configEventsChan: make(chan *settingsUpdate),
 	}
 
 	reqs.Lifecycle.Append(compdef.Hook{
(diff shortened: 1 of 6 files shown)


comp/core/remoteagentregistry/impl/remoteagentregistry.go
```diff
@@ -183,14 +183,49 @@ func (ra *remoteAgentRegistry) start() {
...
 	}()
+	go ra.sendConfigUpdatesLoop()
+}
+
+func (ra *remoteAgentRegistry) sendConfigUpdatesLoop() {
+	log.Info("Remote Agent config update loop started.")
+
+	for {
+
+		// TODO: Implement config update loop
```
rayz, line 208: TODO: Actual implementation details regarding how/where we get the configuration updates is missing. `configsync`(?)

comp/core/remoteagentregistry/impl/remoteagentregistry.go
```diff
@@ -183,14 +206,62 @@ func (ra *remoteAgentRegistry) start() {
...
 	}()
+
+	ra.conf.OnUpdate(func(setting string, oldValue, newValue any) {
+		ra.configUpdatesChan <- &settingsUpdate{
+			setting:  setting,
+			oldValue: oldValue,
+			newValue: newValue,
+		}
+	})
+	go ra.sendConfigUpdatesLoop()
```
GustavoCaso, line 231: Any reason why adding it to a new go routine instead of adding as part of the main loop on line 193?

pkg/proto/datadog/remoteagent/remoteagent.proto
```diff
@@ -84,3 +86,8 @@ message GetTelemetryResponse {
     string prom_text = 1;
   }
 }
+
+message PushedConfig {
```
GustavoCaso, line 90: PushedConfig does not feel good (at least for me :)). What about ConfigUpdate?

comp/core/remoteagentregistry/impl/remoteagentregistry.go
```diff
@@ -183,14 +206,62 @@ func (ra *remoteAgentRegistry) start() {
...
+						delete(ra.agentMap, id)
+						log.Infof("Remote agent '%s' deregistered after being idle for %s.", id, remoteAgentIdleTimeout)
+					}
 				}
 
 				ra.agentMapMu.Unlock()
 			}
 		}
 	}()
+
```
GustavoCaso, line 223: Also, I noticed that the call to the start function is creating a go routine, and then in the start function, we create another goroutine. I think we can have a single goroutine

comp/core/remoteagentregistry/impl/remoteagentregistry.go
```diff
@@ -183,14 +206,62 @@ func (ra *remoteAgentRegistry) start() {
...
+	ra.conf.OnUpdate(func(setting string, oldValue, newValue any) {
+		ra.configUpdatesChan <- &settingsUpdate{
+			setting:  setting,
+			oldValue: oldValue,
+			newValue: newValue,
+		}
+	})
```
GustavoCaso, line 230: Should we move this to line 186 below `ra.registerCollector()` I think it makes more sense and is clearer where are we getting the updates from 😄

comp/core/remoteagentregistry/impl/remoteagentregistry.go
```diff
@@ -497,10 +568,16 @@ func newRemoteAgentClient(registration *remoteagentregistry.RegistrationData) (p
...
 type remoteAgentDetails struct {
-	lastSeen    time.Time
-	displayName string
-	apiEndpoint string
-	client      pb.RemoteAgentClient
+	lastSeen     time.Time
+	displayName  string
+	apiEndpoint  string
+	client       pb.RemoteAgentClient
+	configStream *configStream
```
GustavoCaso, line 575: Does it need to be a pointer? If not we can remove some nil checks we are doing in the code. As all remoteAgentdetails have a dedicated configStream 😄

comp/core/remoteagentregistry/impl/remoteagentregistry.go
misteriaud: +1

pkg/proto/datadog/remoteagent/remoteagent.proto
rayz: Updated it to `ConfigUpdate`

comp/core/remoteagentregistry/impl/remoteagentregistry.go
rayz: Good point, consolidated it to be in the same select loop in the go routine in `start`

comp/core/remoteagentregistry/impl/remoteagentregistry.go
rayz: updated and removed the now unnecessary nil checks 🙂

comp/core/remoteagentregistry/impl/remoteagentregistry.go
```diff
@@ -160,6 +213,14 @@ func (ra *remoteAgentRegistry) registerCollector() {
 func (ra *remoteAgentRegistry) start() {
 	remoteAgentIdleTimeout := ra.conf.GetDuration("remote_agent_registry.idle_timeout")
 	ra.registerCollector()
+	ra.conf.OnUpdate(func(setting string, oldValue, newValue any) {
+		ra.configEventsChan <- &settingsUpdate{
+			setting:   setting,
+			timestamp: time.Now(),
```
GustavoCaso, line 219: I think maybe to avoid race conditions between updates, that the timestamp should be included by the source of the updates (aka the config component) 
https://github.com/DataDog/datadog-agent/blob/fb32c2a66b48a57d0c60d00088d262fec3b70c08/pkg/config/viperconfig/viper.go#L115

That way, the remote agent registry only propagates the timestamp.

What do you think?

rayz:
Closing this outdated pr (pre RFC) in favor of https://github.com/DataDog/datadog-agent/pull/38578

PR #37300, opened 2025-05-22
[NDINT-290] Versa SLA Metrics w/ Session Auth

ken-schneider:
### What does this PR do?
Adds back support for session and CSRF token based authentication to allow calls to Versa analytics APIs which do not support HTTP Basic Auth.

With this change, we can enable collection of the SLA metrics from a previous PR.

Session/Token auth is only used when SLA metrics are being called for, basic auth is still the default for now.

Ran the agent locally against our own Versa cluster.


pkg/collector/corechecks/network-devices/versa/versa.go
```diff
@@ -266,15 +266,16 @@ func filterOrganizations(orgs []client.Organization, includedOrgs []string, excl
 	return filteredOrgs
 }
 
-// generateDeviceNameToIDMap generates a map of device ID to device name to enrich the results from Analytics responses
-func generateDeviceNameToIDMap(deviceMetadata []devicemetadata.DeviceMetadata) map[string]string {
-	deviceNameToIDMap := make(map[string]string)
+// TODO: should we convert the tags map to use ID instead of IP?
```
zoedt, line 269: woops totally valid, i think device_id probably is better here? 🤦 id/ip really mixed up easily

zoedt: not blocker for the pr, but agree! thanks for adding this todo

pkg/collector/corechecks/network-devices/versa/client/auth.go
```diff
@@ -0,0 +1,185 @@
...
+
+	// now we can actually login and get a session cookie
+	err = client.runJSpringSecurityCheck(&authPayload)
+	if err != nil {
+		return fmt.Errorf("failed to run j_spring_security_check to get session token: %w", err)
+	}
+
+	// Request to /versa/analytics/login to obtain Analytics CSRF prevention token
+	analyticsPayload := url.Values{}
+	analyticsPayload.Set("endpoint", client.analyticsEndpoint) // TODO: WHY? Where can we get this for others?
```
zoedt, line 43: **non blocking question**
i forget if during our 1:1 you made a successful call to `/versa/login` that doesn't require this `endpoint` parameter but if i recall correctly, returns the token we're looking for?

tbh it's very close to this `/versa/analytics/login` so maybe i mixed it up 🤔

pkg/collector/corechecks/network-devices/versa/versa.go
```diff
@@ -162,6 +152,16 @@ func (v *VersaCheck) Run() error {
 		v.metricsSender.SendDirectorStatus(directorStatus)
 	}
 
+	// Temporarily commenting out the SLA metrics collection until Versa Analytics auth is sorted out
```
zoedt, line 155: we can remove this comment if the metrics are flowing in ok!

pkg/collector/corechecks/network-devices/versa/client/auth.go
ken-schneider: This flow requires the analytics endpoint parameter to be set. I haven't been able to get that version working, it's something I'm going to look at when I get back.

zoedt (approved):
lgtmmm ty for re-adding !!

ken-schneider:
At this point, the logic in `auth.go` has been updated to avoid making multiple calls to any endpoint. We've added a single GET call to `/versa/analytics/auth/user` which allows us to make the other two pre-existing calls to log in to the director and analytics once.

PR #37301, opened 2025-05-22
[clusteragent/autoscaling] Clear autoscaling telemetry gauges when autoscaler deleted

jennchenn:
### What does this PR do?

Clear autoscaling telemetry when autoscaler is deleted. [CASCL-330](https://datadoghq.atlassian.net/browse/CASCL-330)

### Motivation

When an autoscaler is removed, we do not currently clear out the telemetry - this causes metric counts to be inflated as the agent continues to report the last known telemetry value for the deleted autoscaler.

1. Check all gauge telemetry metrics reported by the cluster agent for autoscaling
2. Delete a `DatadogPodAutoscaler`
3. Verify that metrics continue to be reported for it
4. Pull in changes; delete another `DatadogPodAutoscaler`
5. Verify that metrics stop being reported for that autoscaler

### Additional Notes

[CASCL-330]: https://datadoghq.atlassian.net/browse/CASCL-330?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/clusteragent/autoscaling/workload/telemetry.go
```diff
@@ -137,15 +151,45 @@ func setHorizontalScaleAppliedRecommendations(toReplicas float64, ns, targetName
...
-	for _, source := range validRecommendationSources {
-		telemetryHorizontalScaleAppliedRecommendations.Delete(
-			ns,
-			targetName,
-			autoscalerName,
-			string(source),
-			le.JoinLeaderValue,
-		)
-	}
+	unsetHorizontalTelemetry(ns, targetName, autoscalerName, telemetryHorizontalScaleAppliedRecommendations)
```
vboulineau, line 154: nit: seems like these functions are called once, would be a bit less verbose to have everything in the `deletePodAutoscalerTelemetry` with a single definition of each label set + delete on all related metrics.

pkg/clusteragent/autoscaling/workload/controller.go
```diff
@@ -212,6 +212,7 @@ func (c *Controller) syncPodAutoscaler(ctx context.Context, key, ns, name string
 		if podAutoscalerInternal.Deleted() || podAutoscalerInternal.Spec().Owner != datadoghqcommon.DatadogPodAutoscalerRemoteOwner {
 			log.Infof("Object %s not present in Kubernetes and flagged for deletion (remote) or owner == local, clearing internal store", key)
 			c.store.UnlockDelete(key, c.ID)
+			deletePodAutoscalerTelemetry(podAutoscalerInternal.Namespace(), podAutoscalerInternal.Spec().TargetRef.Name, podAutoscalerInternal.Name())
```
vboulineau, line 215: To avoid trying to track that, you can setup a `TelemetryObserver` on the `Store` and that will make sure you don't miss any source of deletion (we align store content with metrics).

With this I think the telemetry will not be accurate on follower -> leader transitions. Thinking about it, it's probably the same with the `update` BTW.

jennchenn: Would it be inaccurate as we'd be incorrectly clearing the metrics on transition? Would that be fixed by the next reconcile/recommendation update

jeremy-hanna (approved):

pkg/clusteragent/autoscaling/workload/telemetry.go
```diff
@@ -112,6 +112,27 @@ func trackPodAutoscalerStatus(podAutoscaler *datadoghq.DatadogPodAutoscaler) {
...
+		le.JoinLeaderLabel: le.JoinLeaderValue,
+	}
+
+	metrics := []telemetry.Gauge{
+		telemetryHorizontalScaleAppliedRecommendations,
+		telemetryHorizontalScaleReceivedRecommendations,
+		telemetryVerticalScaleReceivedRecommendationsLimits,
+		telemetryVerticalScaleReceivedRecommendationsRequests,
+		autoscalingStatusConditions,
+	}
```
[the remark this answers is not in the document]
jennchenn: thank you for pointing this out, just updated!

GustavoCaso (approved):
LGTM

davidor (approved):

pkg/clusteragent/autoscaling/workload/controller.go
```diff
@@ -495,6 +499,15 @@ func (c *Controller) updateLocalFallbackEnabled(podAutoscalerInternal *model.Pod
 	trackLocalFallbackEnabled(*activeHorizontalSource, *podAutoscalerInternal)
 }
 
+func (c *Controller) unsetTelemetry(key, _ string) {
```
vboulineau: nit: It does not need to be a member function

vboulineau (approved):

pkg/clusteragent/autoscaling/workload/controller.go
vboulineau: The leader/follower stores should be aligned (almost, depends on `Informer` speed), but what we want is these metrics to be representative of the in-memory state of the `Store`, which will be the case with the observer.

PR #37302, opened 2025-05-22
[serverless-init] Fix multiline log splitting

nhulston:
### What does this PR do?

Note: This is only for in-process serverless-init. Does not affect sidecar serverless-init.

This PR fixes multi-line logs such as stacktraces or prettified JSONs from being split into multiple log messages.

Screenshot 2025-05-28 at 11 57 43 AM

This is very annoying for customers.

In the original code, we split by new line because when we pipe stdout/stderr into `channel_writer.go`, we receive messages in a stream which is not guaranteed to be a full message -- there could be multiple logs there are two edge cases:
1. For very long payloads, `Write()` may be called multiple times, each with partial payloads. Therefore, we might send incomplete payloads. So, the original code waits until we receive a `\n` char before flushing.
2. When writes to stdout/stderr occur back-to-back with no delay, they can be combined into one payload. This is mostly fine since this is rare we still send the data (just combined). However, if both logs are JSONs and we send two JSONs at once, the Datadog backend drops the second message because it only searches for the first valid JSON.

For example, if we have an app like this:
```python
def someMethod:
    print('{"msg": "A"}')
    print('{"msg": "B"}')
```

then `Write()` is occasionally called with a payload `{"msg": "A"}\n{"msg": "B"}\n`, leading to "B" getting dropped by the backend. More info [here](https://datadoghq.atlassian.net/wiki/spaces/AL/pages/2727873101/Interesting+Logs+Intake+Behaviors#Handling-of-malformed-payloads).

The original code splits by `\n`, solving this issue.

However, the original code was the wrong approach because it creates the issue of stacktraces or other multiline logs being split up (leading to many tickets being opened).

This PR solves this problem by _not_ splitting by new line. And we still solve problems from the two edge cases by:
- wait until we've seen at least one `\n` char before flushing
- parse incoming payloads and split JSONs into separate logs
- As a side effect, this also fixes the case where we receive a log like `{"msg": "A"}{"msg": "B"}\n` (with no new line between the JSONs) which in the original code would lead to "B" being dropped, but now both "A" and "B" will arrive in Datadog

Note: Initially, I tried using `SOCK_DGRAM` instead of the current pipe-based approach, which is message-based rather than stream-based. I hoped this would preserve full messages without combining them, but this did not work and we still got merged payloads, leading to dropped JSON logs. So I implemented this JSON splitting logic.

Thorough unit tests.

Manually - I have this Python App
```python
@app.route('/')
def hello_world():
    print('{"msg": "A"}')
    print('{"msg": "B"}')
    print('C')

    raise Exception("Test error")
```

and see the logs in Datadog like:
Screenshot 2025-05-28 at 1 06 46 PM
Screenshot 2025-05-28 at 1 06 58 PM
with the full stacktrace intact.

I also tested with Node:
Screenshot 2025-05-28 at 1 53 33 PM
Screenshot 2025-05-28 at 1 53 39 PM

This fix should work everywhere regardless of runtime or if we're in GCP or Azure.

Multiline JSONs are also correctly parsed by Datadog, instead of being split up. Originally, this (which contains four `\n` characters):
```javascript
console.log(JSON.stringify({ foo: 'bar', baz: 'qux' }, undefined, 2));
```
would be sent as 4 different logs. Now, it is sent as a single JSON to Datadog and is correctly parsed by the backend:
Screenshot 2025-05-28 at 2 38 53 PM

Performance -- our JSON parsing takes `O(n)`

If multiple logs are sent to stdout with no delay and in plaintext (not JSON), then they are occasionally combined into one log in Datadog.

If customers are concerned about this, they can do any of the following options:
- Use sidecar instead
- Use a logging library that converts logs to JSON format before printing to stdout/stderr
- Add a small delay between logs

Full doc of possible solutions and tradeoffs: https://docs.google.com/document/d/1lFxDyiaGqneUUb1aYupUjYWFL4EbMS1s1isqtgsF_7c/edit?tab=t.0#heading=h.dnewoeovur60


cmd/serverless-init/log/channel_writer.go
```diff
@@ -34,41 +31,21 @@ func NewChannelWriter(ch chan *logConfig.ChannelMessage, isError bool) *ChannelW
...
-		line, err := cw.Buffer.ReadString('\n')
-		if err == io.EOF {
-			// If EOF, push the line back to buffer and wait for more data
-			cw.Buffer.WriteString(line)
-			break
-		}
-		if err != nil {
-			return n, err
-		}
+	fmt.Printf("[agent] Write called with: %s\n", p)
```
nhulston, line 34: will delete, just testing for now

drichards-87 (approved):

nhulston:
Closing in favor of https://github.com/DataDog/datadog-agent/pull/37551

PR #37309, opened 2025-05-23
[CWS] Add support for inherited variables

lebauce:
### What does this PR do?

Add support for inherited scoped variables.

```
id: rule1
actions:
  - set:
      name: inherited_variable
      field: process.file.name
      inherited: true
```

If a process `p1` triggers `rule1`, its value and the values for its children of `inherited_variable` will be `p1`.
If a child `p2` of `p1` triggers `rule1` again, its value and the values for its children of `inherited_variable`  will be `p1` but `inherited_variable` for `p1` will not change.

### DIFF
--- a/pkg/security/secl/rules/model.go
+++ b/pkg/security/secl/rules/model.go
@@ -149,6 +149,7 @@ type SetDefinition struct {
 	Size         int                    `yaml:"size" json:"size,omitempty"`
 	TTL          *HumanReadableDuration `yaml:"ttl" json:"ttl,omitempty"`
 	Private      bool                   `yaml:"private" json:"private,omitempty"`
+	Inherited    bool                   `yaml:"inherited" json:"inherited,omitempty"`
 }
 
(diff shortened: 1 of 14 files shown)


pkg/security/secl/rules/model.go
```diff
@@ -149,6 +149,7 @@ type SetDefinition struct {
 	Size         int                    `yaml:"size" json:"size,omitempty"`
 	TTL          *HumanReadableDuration `yaml:"ttl" json:"ttl,omitempty"`
 	Private      bool                   `yaml:"private" json:"private,omitempty"`
+	Inherited    bool                   `yaml:"inherited" json:"inherited,omitempty"`
```
YoannGh, line 152: Can we make it so this new field is reported in ruleset_loaded events?

lebauce: Done

YoannGh (approved):

PR #37317, opened 2025-05-23
[autodiscovery] Delete Start/Stop from interface

davidor:
### What does this PR do?

Deletes the `Start` and `Stop` functions from the autodiscovery interface.

The `Stop` function is never called, and `Start` is only invoked from `LoadComponents()` in `cmd/agent/common/loader.go`. However, the function that instantiates autodiscovery already defines fx lifecycle hooks that call `Start`, making the call in `LoadComponents()` unnecessary.

CI.

I also manually ran the commands that call `LoadComponents()` to confirm they still work.

### DIFF
--- a/comp/core/autodiscovery/autodiscoveryimpl/autoconfig.go
+++ b/comp/core/autodiscovery/autodiscoveryimpl/autoconfig.go
@@ -17,6 +17,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/cenkalti/backoff"
 	"go.uber.org/atomic"
 	"go.uber.org/fx"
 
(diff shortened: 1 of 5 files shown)


comp/core/autodiscovery/autodiscoveryimpl/autoconfig.go
```diff
@@ -391,23 +390,19 @@ func (ac *AutoConfig) fillFlare(fb flaretypes.FlareBuilder) error {
...
-func (ac *AutoConfig) Start() {
-	ac.startOnce.Do(func() {
-		listeners.RegisterListeners(ac.serviceListenerFactories)
-		providers.RegisterProviders(ac.providerCatalog)
-		setupAcErrors()
-		// Start the service listener
-		go ac.serviceListening()
-	})
+// start will listen to the service channels before anything is sent to them
+func (ac *AutoConfig) start() {
```
wdhif, line 394: nit: Would it make sense to have not have the `start()` and `stop()` function at all and just have all the code in the Fx `OnStart` and `OnStop` functions?

wdhif (approved):

hush-hush (approved):
:ok_hand:

PR #37318, opened 2025-05-23
[ACIX-470] Refactor linting tasks

Ishirui:
### What does this PR do?
No functional changes ; this PR is just moving code around and moving any functions not defining tasks out of `tasks/linter.py` into `libs/` to improve readability.
### Motivation
When starting work on ACIX-470 I had trouble figuring out what was relevant to the ticket in the 1000-line `linter.py` file. By breaking this out, we improve readability and maintainability.

The output of `dda inv -l` is identical before and after the PR ; I tried running every affected task and got the same output as before the PR.

- Might be confusing for people familiar with the current structure of this module, as code has moved around quite a lot.
- The file/folder structure I chose might not be ideal/intuitive - please do give feedback !
- Renders git history/blame pretty much useless for these files


tasks/linter/__init__.py
```diff
@@ -0,0 +1,42 @@
...
+    gitlab_ci,
+    gitlab_ci_jobs_codeowners,
+    gitlab_ci_jobs_needs_rules,
+    gitlab_ci_jobs_owners,
+    gitlab_ci_shellcheck,
+    job_change_path,
+    list_get_parameter_calls,
+    list_parameters,
+    ssm_parameters,
+)
```
Ishirui, line 17: For reviewers: is there a standard / best practice for "bubbling up" methods to top-level like this ?

Explicitely importing and setting `__all__` for every module makes it very clear what is being made public, but it might be harder to maintain since multiple levels of modules have to be modified when, for example, adding a new function. If anyone has an opinion I'd love to hear it 😊

tasks/linter/__init__.py
```diff
@@ -0,0 +1,42 @@
...
+    gitlab_ci_jobs_codeowners,
+    gitlab_ci_jobs_needs_rules,
+    gitlab_ci_jobs_owners,
+    gitlab_ci_shellcheck,
+    helpers,
+    job_change_path,
+    list_parameters,
+    ssm_parameters,
+    tasks,
+)
```
Ishirui, line 18: For reviewers: is there a standard / best practice for "bubbling up" methods to top-level like this ?

Explicitely importing and setting __all__ for every module makes it very clear what is being made public, but it might be harder to maintain since multiple levels of modules have to be modified when, for example, adding a new function. If anyone has an opinion I'd love to hear it 😊

CelianR (approved):

PR #37340, opened 2025-05-23
AGNTLOG-217 - fix panic during write failures in logs tcp destination

ddrthall:
### What does this PR do?
A panic has been detected within the TCP destination entity of the logs agent. This panic is strictly scoped:
1. the destination must be established on an unreliable endpoint (aka has retry disabled) and 
2. a write failure must occur on a successfully established connection that is old enough to be reset

In order to resolve this case moving forward, we will preemptively exit the send loop on write failure without attempting a connection reset.

An additional unit test has been drafted would catch the panic in the original code


jszwedko (approved):
👍 to the added unit test here.

iadjivon (approved):
:shipit: 

From Docs side (yaml file)

PR #37345, opened 2025-05-23
[AGENTCFG-259] Adding comparison tests for GetEnvVars, GetKnownKeysLowercased, and AllKeysLowercased

rahulkaukuntla:
### What does this PR do?
Adding upon [this PR](https://github.com/DataDog/datadog-agent/pull/37289), I test for the listed methods are being added to ensure that the implementations of viper and ntm are consistent. After writing the tests, I noticed that the `GetEnvVars` implementation for ntm was returning the keys that the env vars were binded to (`c.configEnvVars` has the keys as keys in a hashmap, with the list of env vars that are binded to said key as values in a []string), as opposed to the env vars themselves. We also noticed that `AllKeysLowercased` had an issue with returning unknown keys from the yaml config. I have included the fixes for both issues.


pkg/config/nodetreemodel/compatibility_test.go
```diff
@@ -130,3 +131,176 @@ func TestCompareAllSettingsWithoutDefault(t *testing.T) {
...
+		viperConf, ntmConf := constructBothConfigs(dataYaml, false, func(cfg model.Setup) {
+			cfg.SetKnown("port")
+			cfg.SetDefault("HOST", "localhost")
+			cfg.BindEnv("api_key", "TEST_API_KEY")
+		})
+
+		wantKeys := []string{"port", "host", "api_key"}
+		assert.ElementsMatch(t, wantKeys, viperConf.AllKeysLowercased())
+		assert.ElementsMatch(t, wantKeys, ntmConf.AllKeysLowercased())
+	})
```
dustmop, line 296: Could you also add a test here that uses an unknown key? I believe `AllKeysLowercased` behaves incorrectly with regards to unknown keys and it will need a fix in this PR.

pkg/config/nodetreemodel/compatibility_test.go
```diff
@@ -130,3 +131,176 @@ func TestCompareAllSettingsWithoutDefault(t *testing.T) {
...
+}
+
+// Utility function to extract keys from a map
+func keys(m map[string]interface{}) []string {
```
dustmop, line 300: https://pkg.go.dev/maps#Keys

pkg/config/nodetreemodel/getter.go
```diff
@@ -49,8 +49,8 @@ func (c *ntmConfig) GetEnvVars() []string {
 	c.RLock()
 	defer c.RUnlock()
 	vars := make([]string, 0, len(c.configEnvVars))
-	for v := range c.configEnvVars {
-		vars = append(vars, v)
+	for _, v := range c.configEnvVars {
+		vars = append(vars, v...)
```
dustmop, line 53: Nice catch! Glad this is a simple fix.

dustmop (changes requested):

pkg/config/nodetreemodel/compatibility_test.go
rahulkaukuntla: good idea, I've removed the helper function and just used this.

pkg/config/nodetreemodel/getter.go
rahulkaukuntla: AllKeysLowercased required more complicated fixes :-(

pkg/config/nodetreemodel/compatibility_test.go
rahulkaukuntla: It does--I've made the fixes and added a test case for unknown keys.

pkg/config/nodetreemodel/compatibility_test.go
```diff
@@ -130,3 +131,176 @@ func TestCompareAllSettingsWithoutDefault(t *testing.T) {
...
+		viperConf, ntmConf := constructBothConfigs("", false, func(cfg model.Setup) {
+			cfg.SetKnown("PORT")
+			cfg.SetDefault("TIMEOUT", 30)
+			cfg.BindEnv("log.level", "TEST_LOG_LEVEL")
+		})
+
+		wantKeys := []string{"port", "timeout", "log", "log.level"}
+		assert.ElementsMatch(t, wantKeys, keys(viperConf.GetKnownKeysLowercased()))
+		assert.ElementsMatch(t, wantKeys, keys(ntmConf.GetKnownKeysLowercased()))
+	})
```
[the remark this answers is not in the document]
rahulkaukuntla: done.

pkg/config/nodetreemodel/compatibility_test.go
```diff
@@ -130,3 +133,205 @@ func TestCompareAllSettingsWithoutDefault(t *testing.T) {
...
+
+		viperEnvVars := viperConf.GetEnvVars()
+		ntmEnvVars := ntmConf.GetEnvVars()
+
+		sort.Strings(viperEnvVars)
+		sort.Strings(ntmEnvVars)
+
+		assert.Equal(t, viperEnvVars, ntmEnvVars, "viper and ntm should return the same environment variables")
+
+		expected := []string{"TEST_PORT", "TEST_HOST", "TEST_LOG_LEVEL"}
```
dustmop: If you change the order to `"TEST_HOST", "TEST_LOG_LEVEL", "TEST_PORT"` then you can simplify this by using `assert.Equals(...)` instead of `assert.Contains(...)`. It also makes a stronger guarantee: that the return value of `GetEnvVars` is *exactly* this list instead of a superset of it.

Using `assert.Contains` may have been necessary if you didn't have the call to `sort.Strings` above. However, by sorting you make the return value deterministic.

pkg/config/nodetreemodel/compatibility_test.go
```diff
@@ -130,3 +133,205 @@ func TestCompareAllSettingsWithoutDefault(t *testing.T) {
...
+			cfg.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
+			cfg.BindEnv("db.connection.url")
+		})
+
+		expected := "MYAPP_DB_CONNECTION_URL"
+
+		assert.Contains(t, viperConf.GetEnvVars(), expected, "viper should apply prefix and replacer")
+		assert.Contains(t, ntmConf.GetEnvVars(), expected, "ntm should apply prefix and replacer")
+	})
+}
```
dustmop, line 223: Can you add one more test here: define a unknown setting using the dataYaml variable, define a known setting using `SetKnown`, define a default setting using `SetDefault`, then call `GetEnvVars` and show that none of those settings are included in the return value. In order words, test that `GetEnvVar` *only* returns env vars and not settings defined other ways.

pkg/config/nodetreemodel/config.go
```diff
@@ -435,32 +435,52 @@ func (c *ntmConfig) mergeAllLayers() error {
...
+	// 1. Collect all known keys from schema
+	c.collectKeysFromNode(c.schema, path, keySet, true)
+
+	// 2. Collect all keys from merged tree (only ones with values)
+	c.collectKeysFromNode(c.root, "", keySet, false)
+
+	// 3. Collect all unknown keys, even if they have no value set
+	c.collectKeysFromNode(c.unknown, "", keySet, true)
+
+	allKeys := make([]string, 0, len(keySet))
```
dustmop:
```suggestion
	allKeys := maps.Keys(keySet)
```

pkg/config/nodetreemodel/config.go
```diff
@@ -435,32 +435,52 @@ func (c *ntmConfig) mergeAllLayers() error {
...
+	slices.Sort(allKeys)
+	return allKeys
+}
+
+func (c *ntmConfig) collectKeysFromNode(node InnerNode, path string, keySet map[string]struct{}, includeAllSchemaKeys bool) {
 	for _, name := range node.ChildrenKeys() {
 		newPath := joinKey(path, name)
 
-		child, _ := node.GetChild(name)
+		child, err := node.GetChild(name)
```
dustmop: It should be safe to ignore the error here, as the `name` comes from the node's own `ChildrenKeys` method. A node won't be returning children keys for which it doesn't have a child.

pkg/config/nodetreemodel/config.go
```diff
@@ -435,32 +435,52 @@ func (c *ntmConfig) mergeAllLayers() error {
...
+	allKeys := make([]string, 0, len(keySet))
+	for key := range keySet {
+		allKeys = append(allKeys, key)
+	}
+
+	slices.Sort(allKeys)
+	return allKeys
+}
+
+func (c *ntmConfig) collectKeysFromNode(node InnerNode, path string, keySet map[string]struct{}, includeAllSchemaKeys bool) {
```
dustmop: For clarity's sake, I would rename `includeAllSchemaKeys` to just `includeAllKeys`. This method is being called for (1) the schema, (2) the merged tree, and (3) the unknown set. Firstly, the implementation of this method should not reference details about what its callers are doing (in other words, a function doesn't know who is calling it). Secondly, when this function is called upon the unknown set (within `computeAllSettings`), this bool is true but we're not talking about the schema, so the name is inaccurate.

pkg/config/nodetreemodel/config.go
```diff
@@ -435,32 +435,52 @@ func (c *ntmConfig) mergeAllLayers() error {
...
+		if err != nil {
+			continue
+		}
+
 		if leaf, ok := child.(LeafNode); ok {
-			if leaf.Source() != model.SourceSchema {
-				knownKeys = append(knownKeys, newPath)
-			} else if c.leafAtPathFromNode(newPath, c.root) != missingLeaf {
-				knownKeys = append(knownKeys, newPath)
+			// For schema nodes, include all keys if requested
```
dustmop:
```suggestion
			// Include all keys if requested
```

(as noted above, we also end up here for the unknown set, which is not the schema)

pkg/config/nodetreemodel/read_config_file.go
```diff
@@ -134,11 +134,16 @@ func loadYamlInto(dest InnerNode, source model.Source, inData map[string]interfa
...
 			if !allowDynamicSchema {
 				continue
-			} else if isScalar(value) || isSlice(value) {
-				schemaChild = newLeafNode(value, model.SourceSchema)
-			} else {
-				schemaChild = newInnerNode(make(map[string]Node))
 			}
+
+			// if the key is not defined in the schema, we can still add it to the destination
+			if value == nil || isScalar(value) || isSlice(value) {
```
dustmop, line 140: Do you know of a situation where we end up with `value == nil` here? Is there a unit test covering that case?

dustmop (commented):
Awesome! Really cool to see how you found a solution to `AllKeysLowercased` for nodetreemodel so that it matches what viper does. We're almost there, but I have a few more requests before this is ready to merge

pkg/config/nodetreemodel/read_config_file.go
rahulkaukuntla: Yup, if an unknown key in the yaml isn't set ([customkey1](https://github.com/DataDog/datadog-agent/blob/rahul/test-compare-viper-ntm/pkg/config/nodetreemodel/compatibility_test.go#L342) in this test), then it hits the nil condition here:

> === Failed
=== FAIL: pkg/config/nodetreemodel TestCompareAllKeysLowercased/Includes_unknown_YAML_keys (0.00s)
    compatibility_test.go:353: 
                Error Trace:    /Users/rahul.kaukuntla/dd/datadog-agent/pkg/config/nodetreemodel/compatibility_test.go:353
                Error:          elements differ
                            
                                extra elements in list A:
                                ([]interface {}) (len=1) {
                                 (string) (len=10) "customkey1"
                                }
                            
                            
                                listA:
                                ([]string) (len=4) {
                                 (string) (len=4) "port",
                                 (string) (len=4) "host",
                                 (string) (len=10) "customkey1",
                                 (string) (len=10) "customkey2"
                                }
                            
                            
                                listB:
                                ([]string) (len=3) {
                                 (string) (len=10) "customkey2",
                                 (string) (len=4) "host",
                                 (string) (len=4) "port"
                                }
                Test:           TestCompareAllKeysLowercased/Includes_unknown_YAML_keys

=== FAIL: pkg/config/nodetreemodel TestCompareAllKeysLowercased (0.00s)

pkg/config/nodetreemodel/config.go
rahulkaukuntla: Ah yes, that name should have been updated. fixed.

pkg/config/nodetreemodel/config.go
rahulkaukuntla: done.

pkg/config/nodetreemodel/config.go
rahulkaukuntla: sounds good. removed the error check.

pkg/config/nodetreemodel/config.go
rahulkaukuntla: TIL. It seems that I needed to switch to an [experimental library](https://pkg.go.dev/golang.org/x/exp/maps) to do this--hopefully that is okay (I implemented your suggestion for now).

pkg/config/nodetreemodel/compatibility_test.go
rahulkaukuntla: Good idea and done.

pkg/config/nodetreemodel/compatibility_test.go
rahulkaukuntla: Sure, I made the change.

pkg/config/nodetreemodel/compatibility_test.go
dustmop: You should be creating an unknown setting by having a setting in the `dataYaml` variable that isn't declared by any of `SetKnown`, `SetDefault`, etc and then pass `dataYaml` as the first argument to `constructBothConfigs`. Basically the same thing as what you're doing in the case `"Handles unknown YAML keys"`: that test contains `customKey1` and `customKey2` in the dataYaml but they not declared otherwise, hence they are unknown. We want the same kind of setup here, but as a way to exercise `GetEnvVars` instead of `GetKnownKeysLowercased`.

pkg/config/nodetreemodel/config.go
dustmop: That library used to be experimental, but it's standard now. You're just linking to the old docs, the new ones [are here](https://pkg.go.dev/maps).

pkg/config/nodetreemodel/read_config_file.go
dustmop: Awesome, thanks for the explanation!

pkg/config/nodetreemodel/config.go
```diff
@@ -13,13 +13,14 @@ import (
...
 	"reflect"
-	"slices"
 	"strings"
 	"sync"
 
 	"go.uber.org/atomic"
 
 	"github.com/mitchellh/mapstructure"
+	"golang.org/x/exp/maps"
+	"golang.org/x/exp/slices"
```
dustmop:
```suggestion
	"maps"
	"slices"
```

pkg/config/nodetreemodel/compatibility_test.go
```diff
@@ -130,3 +133,223 @@ func TestCompareAllSettingsWithoutDefault(t *testing.T) {
...
+		})
+
+		expected := "MYAPP_DB_CONNECTION_URL"
+
+		assert.Contains(t, viperConf.GetEnvVars(), expected, "viper should apply prefix and replacer")
+		assert.Contains(t, ntmConf.GetEnvVars(), expected, "ntm should apply prefix and replacer")
+	})
+
+	t.Run("Adding an unknown setting in the yaml", func(t *testing.T) {
+		viperConf, ntmConf := constructBothConfigs("", false, func(cfg model.Setup) {
```
dustmop: This isn't passing in `dataYaml` so it won't see the unknown setting.

pkg/config/nodetreemodel/compatibility_test.go
dustmop: Ah, I see you've now defined `dataYaml` at the top-level of this function method which works! But you then also need to pass it as the first argument to `constructBothConfigs`.

dustmop (approved):
LGTM, aside from the last two very small changes: (1) the imports for maps & slices, (2) passing `dataYaml` for the unknown setting case of `GetEnvVars`. Approving now so you can merge once those are fixed.

pkg/config/nodetreemodel/compatibility_test.go
rahulkaukuntla: oops, thanks for catching!

pkg/config/nodetreemodel/config.go
rahulkaukuntla: sounds good, I've gone back to the standard library.

PR #37356, opened 2025-05-26
Use the same `randomBetween(…)` function in tests and outside of tests

L3n41c:
### What does this PR do?

Ensure unit tests are testing the exact the same code as the one used in the “regular” agent.

### Motivation

The current structure of `randomBetween` is:
```go
func randomBetween(r *rand.Rand, …) {
    if r == nil {
        return // code that is never tested
    }
    return // code that is used only inside tests
```

Run the `TestRandomBetween` unit test.

### Additional Notes

* Followup of #37195.

* I didn’t touch the user variables named `min` and `max`.
  They’re shadowing builtin functions ([`min`](https://pkg.go.dev/builtin#min) and [`max`](https://pkg.go.dev/builtin#max))

### DIFF
--- a/pkg/util/backoff/backoff_test.go
+++ b/pkg/util/backoff/backoff_test.go
@@ -6,7 +6,6 @@
 package backoff
 
 import (
-	"math"
 	"math/rand"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


pkg/util/backoff/backoff_test.go
```diff
@@ -15,21 +14,21 @@ import (
...
 func TestRandomBetween(t *testing.T) {
-	r := rand.New(rand.NewSource(10))
 	getRandomMinMax := func() (float64, float64) {
-		a := float64(r.Intn(10))
-		b := float64(r.Intn(10))
-		min := math.Min(a, b)
-		max := math.Max(a, b)
-		return min, max
+		a := float64(rand.Intn(10))
+		b := float64(rand.Intn(10))
```
louis-cqrl, line 19: this will not have the same behavior as before is it intended ?

louis-cqrl (approved):
small comment but lgtm if we can avoid the using of seeding logic

pkg/util/backoff/backoff_test.go
louis-cqrl: the random value used to be deterministic with a rand.Seed(10) before, now it will use a random seed each time

louis-cqrl: Sorry if I wasn't clear, my point is more to understand why it was seeded before, and would it impact the test if it's not seeded anymore, it's an open question, if we can remove the seed safely without introducing any flakyness then I think we should indeed remove it. (more generally there is the question across all tests in the agent that use a specific seed, to see if we really need those seeds)

PR #37357, opened 2025-05-26
[ACIX-613] Use stackcleaner to delete E2E stacks

CelianR:
### What does this PR do?

> [!WARNING]
> This will migrate **only** `new-e2e-cws` (5 stacks) and the cleanup job to the stack cleaner service with `REMOTE_STACK_CLEANING` env var.

Now we have the [stack cleaner service](https://github.com/DataDog/dd-source/pull/210306) which will destroy stacks upon job completion and also the [agent ci api](https://github.com/DataDog/dd-source/pull/232706) to trigger the stack cleaner service from our ci. This:
- Adds tasks to use the agent ci API.
- Removes the manual destroy at the end of each test and uses the API instead.
- Updates the cleanup job to do a full pipeline cleanup instead (in case a job is cancelled). This is done fully in parallel.


.gitlab/e2e/e2e.yml
```diff
@@ -1041,7 +1052,7 @@ new-e2e-eks-cleanup-on-failure:
     # Use S3 backend
     - PULUMI_CONFIG_PASSPHRASE=$($CI_PROJECT_DIR/tools/ci/fetch_secret.sh $AGENT_QA_E2E pulumi_passphrase) || exit $?; export PULUMI_CONFIG_PASSPHRASE
     - pulumi login "s3://dd-pulumi-state?region=us-east-1&awssdk=v2&profile=$AWS_PROFILE"
-    - dda inv -- -e new-e2e-tests.cleanup-remote-stacks --stack-regex "ci-init-$CI_PIPELINE_ID.*eks.*" --pulumi-backend=dd-pulumi-state
+    - dda inv -- -e new-e2e-tests.cleanup-remote-stacks --stack-regex "ci-init-$CI_PIPELINE_ID-.*" --pulumi-backend=dd-pulumi-state
```
CelianR, line 1055: Clean every stack instead, in case a job gets cancelled

ofek (approved):

.gitlab/e2e/e2e.yml
```diff
@@ -49,6 +49,16 @@
     - export GOOGLE_APPLICATION_CREDENTIALS=~/gcp-credentials.json
     # Generate external links to CI VISIBILITY, used by artifacts:reports:annotations
     - dda inv -- -e gitlab.generate-ci-visibility-links --output=$EXTERNAL_LINKS_PATH
+    # Setup authanywhere to interact with agent ci api and trigger stack cleaner
```
chouetz, line 52: Should we install this in the images later?

tasks/api.py
chouetz, line 1: `api` is maybe a too generic name. It's probably only atlas related for now, correct?

tasks/api.py
```diff
@@ -0,0 +1,133 @@
...
+        endpoint: The endpoint name (see `prefix` for full endpoint).
+        method: The HTTP method to use (default is 'get' if payload is empty, otherwise 'post').
+        jq: One of 'no', 'auto' or 'yes'. Will pipe the json result to jq for pretty printing if jq present.
+        ty: The RAPID type.
+        attrs: The RAPID attributes.
+        payload: Raw payload to send. Prefer using `ty` and `attrs` if possible.
+        env: One of 'prod', 'staging' or 'local'.
+        localport: The port to use for the local endpoint.
+        prefix: The prefix to use for the endpoint.
+        jq: Will use jq to pretty print the JSON response if set to 'yes' or 'auto'.
```
chouetz:
```suggestion
        ty: The RAPID type.
        attrs: The RAPID attributes.
        payload: Raw payload to send. Prefer using `ty` and `attrs` if possible.
        env: One of 'prod', 'staging' or 'local'.
        localport: The port to use for the local endpoint.
        prefix: The prefix to use for the endpoint.
        jq: One of 'no', 'auto' or 'yes'. Will pipe the json result to jq for pretty printing if jq present.
```

chouetz: your `jq` comment was duplicated

tasks/api.py
```diff
@@ -0,0 +1,133 @@
...
+        query: JQ query for the output.
```
chouetz:
```suggestion
        query: jq query for the output.
```
Nit, to be consistent with the rest

tasks/api.py
```diff
@@ -0,0 +1,133 @@
...
+        silent_curl: If True, will silence curl verbose output.
+
+    Examples:
+        $ dda inv -- api hello --env staging
+        $ dda inv -- api hello --env local
+        # Note that we can cast types using a type prefix, e.g. `int:`, `float:` or `bool:`
```
chouetz: What's the point of doing this?

tasks/api.py
```diff
@@ -0,0 +1,133 @@
...
+    ), f'Unknown environment: {env}. Supported environments are: prod, staging, local.'
+    if payload or ty or attrs:
+        assert not (payload and (ty or attrs)), 'Cannot specify payload and type/attributes at the same time.'
+
+    # Construct the payload
+    if ty or attrs:
+        assert ty and attrs, 'Type and attributes must be specified together.'
+
+        # From the comma separated attributes, we build a payload
+        all_attrs = {kv.split('=')[0]: cast_type('='.join(kv.split('=')[1:])) for kv in attrs.split(',')}
```
chouetz: Maybe you could do a little bit more check on the attributes. I suppose there is a risk of badly formatted string that could lead you send incorrect data (`value`, `key=not_expected_type:value`), wdyt?

tasks/api.py
```diff
@@ -0,0 +1,133 @@
...
+        body = {
+            'data': {
+                'type': ty,
+                'attributes': all_attrs,
+            }
+        }
+        payload = json.dumps(body)
+
+    method = method or ('get' if payload == '' else 'post')
+    has_jq = ctx.run('which jq', hide=True, warn=True).ok
```
chouetz: I would recommend to use the "native" [implementation](https://stackoverflow.com/a/34177358) (it is valid for all OS)

tasks/api.py
```diff
@@ -0,0 +1,133 @@
...
+        f"http://localhost:{localport}/{prefix}{endpoint}"
+        if is_local
+        else f"https://agent-ci-api.{dc}/{prefix}{endpoint}"
+    )
+    silent = '-s' if silent_curl else ''
+    use_jq = jq == 'yes' or (jq == 'auto' and has_jq and not from_ci)
+    if payload:
+        payload = f'-d \'{payload}\''
+
+    cmd = f'curl {silent} -X {method.upper()} {url} {token} -H {extra_header} {payload}'
```
chouetz: Could we use `requests` instead of calling curl?

tasks/tools/e2e_stacks.py
```diff
@@ -1,8 +1,14 @@
-import subprocess
+import os
+
+from tasks.api import run
 
 
 # This function cannot be defined in a file that imports invoke.tasks. Otherwise it fails when called with multiprocessing.
```
chouetz:
```suggestion
```
I think this comment is not valid anymore

test/new-e2e/pkg/e2e/suite.go
```diff
@@ -653,6 +654,11 @@ func (bs *BaseSuite[Env]) AfterTest(suiteName, testName string) {
 	}
 }
 
+// IsWithinCI returns true if the test suite is running in a CI environment.
+func (bs *BaseSuite[Env]) IsWithinCI() bool {
+	return os.Getenv("CI_JOB_ID") != ""
```
chouetz: Should we cover [all kind of CI](https://github.com/DataDog/datadog-agent/blob/main/tasks/libs/common/utils.py#L80-L89) (even if you only need Gitlab right now?)

test/new-e2e/pkg/e2e/suite.go
```diff
@@ -681,25 +687,39 @@ func (bs *BaseSuite[Env]) TearDownSuite() {
...
 
-		bs.T().Logf("Destroying stack %s with provisioner %s", bs.params.stackName, id)
-		if err := provisioner.Destroy(ctx, bs.params.stackName, newTestLogger(bs.T())); err != nil {
-			bs.T().Errorf("unable to delete stack: %s, provisioner %s, err: %v", bs.params.stackName, id, err)
+		if bs.IsWithinCI() && os.Getenv("REMOTE_STACK_CLEANING") == "true" {
+			fullStackName := fmt.Sprintf("organization/e2eci/%s", stackName)
+			bs.T().Logf("Remote stack cleaning enabled for stack %s", fullStackName)
+
+			// If we are within CI, we let the stack be destroyed by the stackcleaner-worker service
+			cmd := exec.Command("dda", "inv", "api", "stackcleaner/stack", "--env", "prod", "--ty", "stackcleaner_workflow_request", "--attrs", fmt.Sprintf("stack_name=%s,job_name=%s,job_id=%s,pipeline_id=%s,ref=%s,ignore_lock=bool:true,ignore_not_found=bool:false", fullStackName, os.Getenv("CI_JOB_NAME"), os.Getenv("CI_JOB_ID"), os.Getenv("CI_PIPELINE_ID"), os.Getenv("CI_COMMIT_REF_NAME")))
```
chouetz: Could we call the api directly, rather than use a wrapper in python? I agree this is convenient and might reduce duplication, it's just weird to call a python wrapper through `exec.Command` from a go code..

chouetz (approved):
Minor comments, feel free to take them into account.
Looking forward to see the impact on CI duration!
Thanks

.gitlab/e2e/e2e.yml
CelianR: Yes, I did a card because it will be required for gitlab PATs deprecation

tasks/api.py
CelianR: `agent-ci-api` ? Or `service-api` ? I prefer to avoid `Atlas` but `Service` lgtm

tasks/api.py
CelianR: We have to send the correct types in the json payload otherwise it fails, this is used to cast strings to any other useful type

tasks/api.py
CelianR: I'm not sure to understand, do you want an extra check to avoid an error that won't be covered by casting types (e.g. `int()` will raise an error if the string is not an integer). If you are talking about verifying the signature, it would duplicate the Atlas + Agent CI API code and fails anyway when we send the invalid payload to the API

test/new-e2e/pkg/e2e/suite.go
CelianR: I prefer to use a wrapper since the curl command would be quite long / hard to read and it also wraps the authentication.
It's weird but it's more maintainable IMO, what do you think?
The full curl command would be:
```sh
curl -s -X POST https://agent-ci-api.us1.ddbuild.io/internal/agent-ci-api/stackcleaner/stack -H "$(authanywhere --audience rapid-agent-devx)" -H "X-DdOrigin: 16478618" -d '{"data": {"type": "stackcleaner_workflow_request", "attributes": {"stack_name": "organization/e2eci/ci-1036045561-4670-cws-agentsuite-gcp", "job_name": "new-e2e-cws: [--run TestAgentSuite(EC2|GCP)]", "job_id": "1036045561", "pipeline_id": "70986903", "ref": "celian/mybranch", "ignore_not_found": true}}}'
```

sblumenthal (approved):

tasks/api.py
CelianR: (you can have a look at this [payload](https://github.com/DataDog/datadog-agent/pull/37357/files/117d5dd8568580f71bee882f3e7ede2d1c660d61#diff-972ffab55269f2d81aa7d9b62d837b392b92095cb2be5589936dab584abc2ac6) using it)

PR #37362, opened 2025-05-26
chore(fleet): Create APM SSI script

BaptisteFoy:
### What does this PR do?
Creates an APM SSI install script flavour that will install the packages & copy the installer to `/usr/bin/datadog-installer` if the agent isn't installed.
Updates the APM scripts to reference that installer instead of the packaged one.

Manual QA, E2E tests


pkg/fleet/installer/setup/common/packages.go
```diff
@@ -72,7 +75,8 @@ func resolvePackages(env *env.Env, packages Packages) []packageWithVersion {
 
 // Packages is a list of packages to install
 type Packages struct {
-	install map[string]packageWithVersion
+	install            map[string]packageWithVersion
+	CopyInstallerToBin bool // CopyInstallerToBin indicates whether the installer should be copied to /usr/bin/datadog-installer
```
arbll: Nit: can you move setting this to true to a method instead of a public var to keep consistent with the rest ?

test/new-e2e/tests/installer/script/all_scripts_test.go
```diff
@@ -48,9 +48,8 @@ var (
 	}
 	scriptTestsWithSkippedFlavors = []installerScriptTestsWithSkippedFlavors{
 		{t: testDatabricksScript},
-		{t: testDefaultScript, skippedFlavors: []e2eos.Descriptor{
-			e2eos.CentOS7, // CentOS 7 is not supported by the default script because of SELinux
-		}},
+		{t: testDefaultScript, skippedFlavors: []e2eos.Descriptor{e2eos.CentOS7}},
+		{t: testSSIScript, skippedFlavors: []e2eos.Descriptor{e2eos.CentOS7, e2eos.RedHat9, e2eos.FedoraDefault, e2eos.AmazonLinux2}},
```
arbll: Why do we skip those platforms ?

BaptisteFoy: I copied over from the SSI suite; but indeed this is not required anymore. I have removed them.

arbll (approved):

PR #37375, opened 2025-05-26
[AGENTRUN-297] Detail comments in Python executable configuration

Enzu83:
### What does this PR do?

It was at first to remove the path of the Python executable in RTLoader Python configuration.
After investigation, the Windows issue seems to be a bug that should have been fixed but is not.

Only the workaround comment for `program_name` variable setup has been removed.

### Motivation

Removing this line breaks a unit test, only on Windows. It shouldn't break any tests.

### Additional Notes

The test that fails is a mock test that mocks the Python executable path:

https://github.com/DataDog/datadog-agent/blob/f07a0ac61c1078e0ded3a06660671a2131417600/rtloader/test/rtloader/rtloader_test.go#L77

On Windows, the path may have been overwritten.

This test passes even if the `program_name` config is not set, but fails if `executable` config is removed.

Found out that, on Windows only, the Python variable `sys.executable` doesn't always return the path to the Python interpreter executable. Here's some links related to this issue in other contexts:
 - https://github.com/qgis/QGIS/issues/45646
 - https://github.com/python/cpython/issues/124241

`sys.executable` may no be always trustable on Windows.

The issue might come from the mock test, but the reason is still unknown.


rtloader/three/three.cpp
```diff
@@ -85,15 +85,14 @@ bool Three::init()
 
     // Configure Python executable if provided
     if (!_pythonExe.empty()) {
-        // Note: While PyConfig provides the necessary functionality, we maintain this additional configuration
-        // to ensure consistent behavior across all platforms, including Windows.
+        // Set the program name
```
louis-cqrl, line 88: We should add a comment here saying that we may be able to remove this variable nonetheless we decided to keep it to ensure keeping the same behavior

rtloader/three/three.cpp
```diff
@@ -85,15 +85,14 @@ bool Three::init()
...
         status = PyConfig_SetBytesString(&_config, &_config.program_name, _pythonExe.c_str());
         if (PyStatus_Exception(status)) {
             setError("Failed to set program name" + (status.err_msg ? ": " + std::string(status.err_msg) : ""));
             PyConfig_Clear(&_config);
             return false;
         }
-
+        
         // Set the executable path
```
louis-cqrl, line 98: we should add a comment here saying why we cannot remove this variable, maybe add some link to public documentation showing behavior discrepancies with windows. It should also show that according to python doc this variable should be automatically populated with program_name and python_home but this goes not well in our case

louis-cqrl (changes requested):

rtloader/three/three.cpp
Enzu83: Done, thanks!

rtloader/three/three.cpp
Enzu83: I mentioned which test fails and added a link to a GitHub issue that is about the same problem, but couldn't find the public documentation related to the config of `program_name` and `python_home`.

rtloader/three/three.cpp
```diff
@@ -85,16 +85,21 @@ bool Three::init()
 
     // Configure Python executable if provided
     if (!_pythonExe.empty()) {
-        // Note: While PyConfig provides the necessary functionality, we maintain this additional configuration
-        // to ensure consistent behavior across all platforms, including Windows.
+        // Set the program name
+        // Removing the setup of `program_name` doesn't break any test. 
+        // Thus, it might be not required but we keep it to ensure that nothing else breaks.
```
louis-cqrl: maybe something in this tone ?
```
// While removing `program_name` does not currently cause any test failures,
// we retain it as a precaution to avoid unforeseen issues in environments or cases not covered by tests.
```

rtloader/three/three.cpp
```diff
@@ -85,16 +85,21 @@ bool Three::init()
...
         if (PyStatus_Exception(status)) {
             setError("Failed to set program name" + (status.err_msg ? ": " + std::string(status.err_msg) : ""));
             PyConfig_Clear(&_config);
             return false;
         }
-
+        
         // Set the executable path
+        // If we remove the setup of `executable`, the test `TestSysExecutableValue` 
+        // located at `./rtloader/test/rtloader/rtloader_test.go:77` fails but only on Windows.
```
louis-cqrl:
```suggestion
        // (located at ./rtloader/test/rtloader/rtloader_test.go:77) to fail on Windows.
```

rtloader/three/three.cpp
```diff
@@ -85,16 +85,21 @@ bool Three::init()
...
+        // The path of the Python executable isn't mocked or is overwritten, the reason is unclear.
+        // Here's a GitHub issue that tackles the same problem in another context: https://github.com/python/cpython/issues/124241
```
louis-cqrl:
```
// The root cause is unclear; it may be due to the executable path not being properly mocked or being overwritten elsewhere.
// For related discussion, see: https://github.com/python/cpython/issues/124241
```

maybe something more in this tone ?

louis-cqrl:
Can you update the name of the PR to be closer to what the PR is doing ?

rtloader/three/three.cpp
Enzu83: The phrasing is better with your suggestion, thanks!

rtloader/three/three.cpp
Enzu83: It's clearer with the parenthesis, thanks!

rtloader/three/three.cpp
Enzu83: I prefer your phrasing here too, thanks!

louis-cqrl (approved):

PR #37381, opened 2025-05-26
[CWS] Enable to track SetSockOpt syscall

theop-dd:
### What does this PR do?
- Add all the needed structures and function to provide an end-to-end tracking of the set sock opt syscall from kernel to user space.
-> Return to the user space 2 parameters of the syscall: socket level and socket option name.
- Add the possibility to create rules based on that parameters: setsockopt.level ; setsockopt.optname.
- Add new variables related to setsockopt syscall that can be used in the rules: SOL_SOCKET and SO_ATTACH_FILTER.
### Motivation
This PR was created to replace the current ondemand tracking of bpfdoor by a proper ruleset. We can detect if the syscall setsockopt is called with the parameters SOL_SOCKET and SO_ATTACH_FILTER.
The rule to detect bpfdoor would be : setsockopt.level == SOL_SOCKET && setsockopt.optname == SO_ATTACH_FILTER
One functional test was added TestSetSockOpt.go. It tests if the syscalls is well tracked and if it returns the goods values for the tracked parameters (level and optname). The test mimics a bpfdoor execution by calling setsockopt with the parameters SOL_SOCKET and SO_ATTACH_FILTER.


pkg/security/secl/model/unmarshallers_linux.go
```diff
@@ -1504,3 +1504,23 @@ func (e *SysCtlEvent) UnmarshalBinary(data []byte) (int, error) {
...
+	if len(data) < 12 {
+		return 0, ErrNotEnoughData
+	}
+
+	e.Socket = binary.NativeEndian.Uint32(data[0:4])
+	fmt.Printf("UnmarshalBinary SetSockOptEvent: Socket %d\n", e.Socket)
+
+	e.Level = binary.NativeEndian.Uint32(data[4:8])
+	e.OptName = binary.NativeEndian.Uint32(data[8:12])
+	return 12, nil
```
paulcacheux, line 1525: 12 is only the size excluding syscall event and syscall context, you should probably return `12 + read`

theop-dd: Yes you're right, I'll do it

drichards-87 (approved):

pkg/security/ebpf/c/include/hooks/setsockopt.h
```diff
@@ -0,0 +1,66 @@
...
+    pop_syscall(EVENT_SETSOCKOPT);
+    return 0;
+}
+
+
+HOOK_SYSCALL_ENTRY3(setsockopt, int, socket, int, level, int , optname) {
+    return trace__sys_setsock_opt(SYNC_SYSCALL, socket, level, optname);
+}
+
+HOOK_SYSCALL_EXIT(setsockopt) {
```
paulcacheux, line 62: In addition to this hook, and in order to support centos 7, you will need to also do something like https://github.com/DataDog/datadog-agent/blob/d5b5714e6aedfe058546204a8645317c6ab4aa04/pkg/security/ebpf/c/include/hooks/chmod.h#L94

pkg/security/ebpf/c/include/events_definition.h
```diff
@@ -518,4 +518,16 @@ struct sysctl_event_t {
...
 };
 
+struct setsockopt_event_t {
+    struct kevent_t event;
+    struct process_context_t process;
+    struct span_context_t span;
+    struct container_context_t container;
+    struct syscall_t syscall;
+
+    int socket;
```
spikat, line 528: no really needed (see other comment)

pkg/security/ebpf/c/include/hooks/setsockopt.h
```diff
@@ -0,0 +1,70 @@
...
+int __attribute__((always_inline)) sys_set_sock_opt_ret(void *ctx, int retval) {
+    struct syscall_cache_t *syscall = pop_syscall(EVENT_SETSOCKOPT);
+
+    if (!syscall) {
+        return 0;
+    }
+
+    struct setsockopt_event_t event = {
+        .syscall.retval = retval,
+        .event.flags = syscall->async ? EVENT_FLAGS_ASYNC : 0,
```
spikat, line 38: the async stuff is related to io_uring. IMHO you could remove this line.

pkg/security/secl/model/consts_linux.go
```diff
@@ -845,6 +845,12 @@ var (
 		"SYSCTL_READ":  SysCtlReadAction,
 		"SYSCTL_WRITE": SysCtlWriteAction,
 	}
+	// SetSockoptConstants is the list of available actions for setsockopt events
+	// generate_constants:SetSockopt Actions,SetSockopt Actions are the supported actions for the setsockopt event.
+	SetSockoptConstants = map[string]int{
+		"SOL_SOCKET":       syscall.SOL_SOCKET,
+		"SO_ATTACH_FILTER": syscall.SO_ATTACH_FILTER,
+	}
```
spikat, line 853: you should have two different lists here, a first for all possible levels, and a second for all possible optnames

pkg/security/tests/setsockopt_test.go
```diff
@@ -0,0 +1,101 @@
...
+			Len    uint16
+			_      [6]byte
+			Filter *SockFilter
+		}
+
+		defer func() {}()
+
+		test.WaitSignal(t, func() error {
+			var err error
+			fd, err = syscall.Socket(syscall.AF_INET, syscall.SOCK_RAW, syscall.IPPROTO_TCP)
```
spikat, line 58: nit: you could do a `fd, err := ` instead

pkg/security/tests/setsockopt_test.go
```diff
@@ -0,0 +1,101 @@
...
+			assertTriggeredRule(t, rule, "test_rule_setsockopt")
+
+			val, _ := event.GetFieldValue("setsockopt.socket")
+			assert.Equal(t, int(fd), val, "setsockopt.socket mismatch")
+
+			level, _ := event.GetFieldValue("setsockopt.level")
+			assert.Equal(t, int(syscall.SOL_SOCKET), level)
+
+			optname, _ := event.GetFieldValue("setsockopt.optname")
+			assert.Equal(t, int(syscall.SO_ATTACH_FILTER), optname)
```
spikat, line 97: You can remove this part as it's already validated by the rule itself, but maybe you could check the retval?

pkg/security/tests/setsockopt_test.go
paulcacheux: I don't think so because he wants to set the value to the fd from the outer scope. Using `:=` would create a new variable in this tight scope and make the test fail

pkg/security/ebpf/c/include/hooks/setsockopt.h
theop-dd: Do you mean the .event.flags line?

pkg/security/secl/model/consts_linux.go
theop-dd: Yes I agree, I will add it.

pkg/security/ebpf/c/include/hooks/setsockopt.h
theop-dd: I checked the CI and it seems that some CentOS test continue to fail. I will try again

pkg/security/ebpf/c/include/hooks/setsockopt.h
spikat: yep!

pkg/security/ebpf/c/include/hooks/setsockopt.h
paulcacheux: I believe you are missing the pkg/security/ebpf/probes/raw_sys_exit.go part

pkg/security/secl/model/consts_linux.go
theop-dd: Do I add every possible value even if our only check is level == SOL_SOCKET && optname == SO_ATTACH_FILTER ?

paulcacheux: Yes please add every values that make sense, to make this thing re-usable

pkg/security/ebpf/c/include/hooks/setsockopt.h
theop-dd: Ah yes, I didn't know this file yet, thanks

pkg/security/ebpf/c/include/events_definition.h
```diff
@@ -518,4 +518,17 @@ struct sysctl_event_t {
     char sysctl_buffer[MAX_SYSCTL_BUFFER_LEN];
 };
 
+struct setsockopt_event_t {
+    struct kevent_t event;
+    struct process_context_t process;
+    struct span_context_t span;
+    struct container_context_t container;
+    struct syscall_t syscall;
+    struct syscall_context_t syscall_ctx;
```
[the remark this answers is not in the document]
theop-dd: I agree. It's done

pkg/security/secl/model/consts_linux.go
theop-dd: I've added all the constants defined in the kernel code that were defined in the syscall GO lib.

paulcacheux: that's great, thanks !

pkg/security/secl/model/consts_linux.go
```diff
@@ -845,6 +845,245 @@ var (
...
+		"IPV6_TCLASS":               syscall.IPV6_TCLASS,       // 67
+		"IPV6_AUTOFLOWLABEL":        70,                        // 70
+		"IPV6_ADDR_PREFERENCES":     72,                        // 72
+		"IPV6_MINHOPCOUNT":          73,
+		"IPV6_ORIGDSTADDR":          74,
+		"IPV6_TRANSPARENT":          75,
+		"IPV6_UNICAST_IF":           76,
+		"IPV6_RECVFRAGSIZE":         77,
+		"IPV6_FREEBIND":             78,
+		// IPPROTO_ICMPV6
```
spikat: we should remove this line

pkg/security/secl/model/consts_linux.go
```diff
@@ -845,6 +845,245 @@ var (
...
+		"IP_ROUTER_ALERT":           syscall.IP_ROUTER_ALERT,   // 5
+		"IP_RECVOPTS":               syscall.IP_RECVOPTS,       // 6
+		"IP_RETOPTS":                syscall.IP_RETOPTS,        // 7
+		"IP_PKTINFO":                syscall.IP_PKTINFO,        // 8
+		"IP_PKTOPTIONS":             9,                         // 9
+		"IP_MTU_DISCOVER":           syscall.IP_MTU_DISCOVER,   // 10
+		"IP_RECVERR":                syscall.IP_RECVERR,        // 11
+		"IP_RECVTTL":                syscall.IP_RECVTTL,        // 12
+		"IP_RECVTOS":                syscall.IP_RECVTOS,        // 13
+		"IP_MTU":                    14,                        // 14
```
spikat: please remove all comments if they don't bring any value

pkg/security/secl/model/consts_linux.go
```diff
@@ -845,6 +845,245 @@ var (
...
+		"IP_PASSSEC":                syscall.IP_PASSSEC,        // 18
+		"IP_TRANSPARENT":            syscall.IP_TRANSPARENT,    // 19
+		"IP_ORIGDSTADDR":            syscall.IP_ORIGDSTADDR,    // 20
+		"IP_MINTTL":                 syscall.IP_MINTTL,         // 21
+		"IP_NODEFRAG":               22,                        // 22
+		"IP_CHECKSUM":               23,                        // 23
+		"IP_BIND_ADDRESS_NO_PORT":   24,                        // 24
+		"IP_RECVFRAGSIZE":           25,                        // 25
+		"IP_RECVERR_RFC4884":        26,                        // 26
+		"IP_MULTICAST_IF":           32,                        // 32
```
spikat: Please for all optnames that are not define in go `syscall` package:
* did you find any in the `unix` package?
* we tend to avoid using direct values if possible, but here I'm not sure we should create a define for all of them. But at least, could you please add a comment with the link(s) you used to find those values?

pkg/security/secl/model/consts_linux.go
theop-dd: Yep, I just let it for you to see. The idea is just to show that I was not able to find values for this proto yet

pkg/security/secl/model/consts_linux.go
theop-dd: Yes, I'll do it in the next commit, was just here to help for the review

pkg/security/secl/model/consts_linux.go
theop-dd: I've not checked unix package yet, I will do it. Anyway, I will provide the link of the kernel source code I used. Where do you want me to put the links ?

pkg/security/secl/model/consts_linux.go
spikat: Hm ok! I think it's totally fine to not having it in this PR. If you want you can add a more explicit comment

pkg/security/secl/model/consts_linux.go
spikat: thanks!

pkg/security/secl/model/consts_linux.go
spikat: You can put links before all the values, like line 860

theop-dd: I will add path to kernel files

pkg/security/secl/model/consts_linux.go
theop-dd: I've removed the comment. Anyway the headers that define ICMPv6 constant is very small with only some const. It's better not to put it imo because I don't have a perfect understanding of its

pkg/security/secl/model/consts_linux.go
theop-dd: Edit: I've added the link to the elixir files ^^

spikat (approved):

paulcacheux (approved):

PR #37382, opened 2025-05-26
(fleet) collect journalctl logs on error

arbll:
This PR collects `journalctl` error logs when we fail to restart the agent service in the install script.


raphaelgavache (approved):
nice

PR #37390, opened 2025-05-26
[CWS] allow changing the gRPC connection direction

safchain:
### What does this PR do?

Allow to change the direction of the gRPC connection between the security-agent and system-probe for the events and the activity dumps. Two gRPC sockets are now used to handle both API & Event connections. Depending on a feature flags the event gRPC server can be either the security-agent or system-probe.


pkg/security/module/server.go
```diff
@@ -187,29 +171,6 @@ func (a *APIServer) SendActivityDump(dump *api.ActivityDumpStreamMessage) {
 	}
 }
 
-// GetEvents waits for security events
-func (a *APIServer) GetEvents(_ *api.GetEventParams, stream api.SecurityModule_GetEventsServer) error {
-	if prev := a.connEstablished.Swap(true); !prev {
```
safchain, line 192: Need to handle this with the new approach

pkg/security/probe/config/config.go
```diff
@@ -219,6 +222,7 @@ func NewConfig() (*Config, error) {
 
 		// event server
 		SocketPath:       pkgconfigsetup.SystemProbe().GetString(join(evNS, "socket")),
+		CmdSocketPath:    pkgconfigsetup.SystemProbe().GetString(join(evNS, "cmd_socket")),
```
YoannGh, line 225: Shouldn't this be under the runtime-security namespace?

dustmop (approved):

spikat (approved):

pkg/security/rules/engine.go
```diff
@@ -247,6 +247,9 @@ func (e *RuleEngine) startSendHeartbeatEvents(ctx context.Context) {
 
 // StartRunningMetrics starts sending the running metrics
 func (e *RuleEngine) StartRunningMetrics(ctx context.Context) {
+	e.Lock()
```
paulcacheux: What is this lock protecting ? the wait group add operation ?

pkg/security/rules/engine.go
```diff
@@ -532,6 +535,9 @@ func (e *RuleEngine) RuleMatch(ctx *eval.Context, rule *rules.Rule, event eval.E
 
 // Stop stops the rule engine
 func (e *RuleEngine) Stop() {
+	e.Lock()
```
paulcacheux: Can't this create a deadlock ?
let's say you call `Stop()`, you take the lock of the rule engine.
Now the goroutine launched in `StartRunningMetrics` tries to take the read lock and never finishes no ? And this causes `e.wg.Wait()` to block forever

pkg/security/rules/engine.go
safchain: No, I think this is something remaining from some previous iteration. Going to remove it

pkg/security/rules/engine.go
safchain: Removed the lock in the StartRunningMetrics

paulcacheux: I think the problematic line is https://github.com/DataDog/datadog-agent/blob/7614f1b3f9724aa47805b761f4c2af2d63a8ff57/pkg/security/rules/engine.go#L291 this read lock can still create the race I think

paulcacheux (approved):

PR #37395, opened 2025-05-27
Static quality gates entrypoint to allow on-CI debugging

Pythyu:
### What does this PR do?

This PR create a `debug_static_quality_gates` manual job that allows to run comparaison of files within packages and images of a specific static quality gate.

### Motivation

Give more insight of where a PR cause an increase in a static quality gate.

### Additional Notes

Package size debug [link](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/969101442)
Docker image debug [link](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/980641555)


.gitlab-ci.yml
```diff
@@ -167,32 +167,32 @@ variables:
...
-  CI_IMAGE_RPM_X64: v66293343-2eef00c4
-  CI_IMAGE_RPM_X64_SUFFIX: ""
-  CI_IMAGE_RPM_ARM64: v66293343-2eef00c4
-  CI_IMAGE_RPM_ARM64_SUFFIX: ""
-  CI_IMAGE_RPM_ARMHF: v66293343-2eef00c4
-  CI_IMAGE_RPM_ARMHF_SUFFIX: ""
-  CI_IMAGE_WIN_LTSC2022_X64: v66293343-2eef00c4
-  CI_IMAGE_WIN_LTSC2022_X64_SUFFIX: ""
+  CI_IMAGE_BTF_GEN: v66796378-17ead302
+  CI_IMAGE_BTF_GEN_SUFFIX: "_test_only"
```
ofek: LGTM after the suffixes are removed!

ofek (approved):

.gitlab-ci.yml
chouetz: This is [failing the pipeline](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/972022779), so you can approve if this is the only issue with the PR 😄

.gitlab/JOBOWNERS
```diff
@@ -146,6 +146,7 @@ trace_agent_integration_tests @DataDog/agent-apm
 serverless_cold_start_performance-deb_x64      @DataDog/serverless
 serverless_cold_start_performance-deb_x64_dind @DataDog/agent-devx
 static_quality_gates                           @DataDog/agent-delivery
+debug_static_quality_gates                           @DataDog/agent-delivery
```
alopezz: Nit, let's keep these lined up.

```suggestion
debug_static_quality_gates                     @DataDog/agent-delivery
```

tasks/quality_gates.py
```diff
@@ -36,6 +36,35 @@
...
+            raise Exit(code=1, message="The current commit has no pipeline attached.")
+        current_pipeline = pipeline_list[0]
+        job_generator = filter(
+            lambda job: job.name == "debug_static_quality_gates", current_pipeline.jobs.list(iterator=True)
+        )
+    except Exception as e:
+        print(f"Failed to fetch debug_static_quality_gates url !\n{traceback.format_exc()}\n{str(e)}")
+        return ""
+    if not any(job_generator):
+        return ""
```
alopezz: Nit: This behaviour should probably be delegated to the function that generates the message, returning an empty string here sort of works the way the message template is worded but is not very obvious behavior.

tasks/quality_gates.py
```diff
@@ -36,6 +36,35 @@
...
+    debug_job = next(job_generator)
```
alopezz: This looks to me like it _should_ not work.`any` will consume an element from the iterator, and if we only have one job satisfying the name-based filter, it's very likely going to cause the `next` to raise an exception (because of the iterator being exhausted). Also, this `next` call could also still raise an exception from the GitLab API (because the gitlab iterator underneath may still make network calls an dfail).

You should be able to do this without saving the generator to a variable in order to reduce the chances of their behavior tripping us up, something like:

```python
try:
    debug_job = next(job for job in current_pipeline.jobs.list(iterator=True) if job.name == "debug_static_quality_gates")
except StopIteration:
    # Error handling code
```

tasks/quality_gates.py
```diff
@@ -36,6 +36,35 @@
...
+    return f"https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/{debug_job.get_id()}"
```
alopezz: Can we avoid hardcoding the url pattern here? It should be possible to at least use `repo.url` ([doc](https://python-gitlab.readthedocs.io/en/stable/api/gitlab.html#gitlab.Gitlab.url)), or maybe even something like `debug_job.attrs['web_url']` (much less sure about this one).

alopezz (changes requested):
I've made a partial pass, I find the code on `get_debug_url` a bit suspicious as per my comment, let's clarify that.

tasks/quality_gates.py
Pythyu: 🤔 `repo.url` doesn't exist, it's in the `Gitlab` class which has the `url` attribute and it's only the gitlab domain. So I'll go with `debug_job._attrs["web_url"]` which gives the complete web url with the job id 👍

tasks/quality_gates.py
Pythyu: I agree but I want to avoid overloading the `display_pr_comment` function with error cases for everything it calls

tasks/quality_gates.py
Pythyu: Completely missed it indeed, since I had some leftover of some other function mixed in didn't notice, thanks !

tasks/quality_gates.py
alopezz: I wonder if that should be `attributes` instead (based on [this](https://python-gitlab.readthedocs.io/en/stable/api/gitlab.html#gitlab.base.RESTObject.attributes)). Just in case, since I don't see the `_attrs` thing documented and the underscore suffix would normally indicate something that's not part of the public API.

tasks/quality_gates.py
```diff
@@ -88,8 +118,8 @@ def getMetric(*metric_names, gate_name=gate['name']):
             error_message = gate['message'].replace('\n', '<br>')
             body_error_footer += f"|{gate_name}|{gate['error_type']}|{error_message}|\n"
             with_error = True
-
-    body_error_footer += "\n</details>\n\nStatic quality gates prevent the PR to merge! You can check the static quality gates [confluence page](https://datadoghq.atlassian.net/wiki/spaces/agent/pages/4805854687/Static+Quality+Gates) for guidance. We also have a [toolbox page](https://datadoghq.atlassian.net/wiki/spaces/agent/pages/4887448722/Static+Quality+Gates+Toolbox) available to list tools useful to debug the size increase.\n"
+    debug_info_footer = footer_error_debug_advice.format(get_debug_job_url())
```
alopezz: As you can tell by [this error](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/990293810#L801), this results in a request to the GitLab API even when we're not intending to display it (that is, when there's no error). While not the end of the world, it's still unnecessary.

tasks/quality_gates.py
```diff
@@ -36,6 +36,36 @@
...
+            raise Exit(code=1, message="The current commit has no pipeline attached.")
+        current_pipeline = pipeline_list[0]
+        debug_job = next(
+            job for job in current_pipeline.jobs.list(iterator=True) if job.name == "debug_static_quality_gates"
+        )
+    except StopIteration:
+        print("Job debug_static_quality_gates wasn't found in the current pipeline !")
+        return ""
+    except Exception as e:
+        print(f"Failed to fetch debug_static_quality_gates url !\n{traceback.format_exc()}\n{str(e)}")
```
alopezz: Nits around punctuation in these messages:
```suggestion
        print("Job debug_static_quality_gates wasn't found in the current pipeline!")
        return ""
    except Exception as e:
        print(f"Failed to fetch debug_static_quality_gates url!\n{traceback.format_exc()}\n{str(e)}")
```

tasks/static_quality_gates/lib/docker_agent_lib.py
```diff
@@ -116,3 +121,67 @@ def generic_docker_agent_quality_gate(gate_name, arch, jmx=False, flavor="agent"
...
+def get_images_content_diff(ctx, url_1, url_2):
+    ctx.run("mkdir image1 image2 image1/out/ image2/out/")
+    # Pull images locally to get on disk size
+    ctx.run(f"crane pull {url_1} image1/output.tar")
+    ctx.run(f"crane pull {url_2} image2/output.tar")
+    # The downloaded image contains some metadata files and another tar.gz file.
+    for src_folder in ["image1", "image2"]:
+        ctx.run(f"cd {src_folder} && tar -xf output.tar")
+        image_content = ctx.run(
+            "tar -tvf " + f"{src_folder}/output.tar" + " | awk -F' ' '{print $3; print $6}'", hide=True
```
alopezz: I see no reason for the concatenation here:
```suggestion
          f "tar -tvf {src_folder}/output.tar | awk -F' ' '{print $3; print $6}'", hide=True
```

But also, with no context is impossible to know what the `awk` command is extracting here.

tasks/static_quality_gates/lib/docker_agent_lib.py
```diff
@@ -116,3 +121,67 @@ def generic_docker_agent_quality_gate(gate_name, arch, jmx=False, flavor="agent"
...
+        ).stdout.splitlines()
+        image_tar_gz = []
+        for k, _ in enumerate(image_content):
+            if k % 2 == 0:
+                if "tar.gz" in image_content[k + 1]:
+                    image_tar_gz.append(image_content[k + 1])
```
alopezz: - Why do we only need one out of every two elements (the `k % 2`?). I suppose it's somewhat related to the `awk` above but then again why are we extracting two fields with it?
- You should be able to avoid all the index manipulation by doing

```suggestion
        for line in image_content[::2]:
            if "tar.gz" in line:
                image_tar_gz.append(line)
```

tasks/static_quality_gates/lib/docker_agent_lib.py
```diff
@@ -116,3 +121,67 @@ def generic_docker_agent_quality_gate(gate_name, arch, jmx=False, flavor="agent"
...
+        if image_tar_gz:
+            for image in image_tar_gz:
+                ctx.run(f"cd {src_folder} && tar -xf {image} -C ./out", hide=True)
+        else:
+            print(color_message("[WARN] No tar.gz file found inside of the image", "orange"), file=sys.stderr)
+    # Compare both image content
+    _diff("image1/out", "image2/out")
```
alopezz: The function is called `get_` but this is not returning anything, it's printing it.

tasks/static_quality_gates/lib/docker_agent_lib.py
```diff
@@ -1,7 +1,12 @@
 import os
 import sys
 
+from invoke import Exit
+
+from tasks.libs.ciproviders.gitlab_api import get_gitlab_repo
 from tasks.libs.common.color import color_message
+from tasks.libs.common.diff import diff as _diff
```
alopezz: I would advise against this kind of renaming, the initial underscore in Python conveys that it's a private function which makes it a bit awkward when called.

tasks/static_quality_gates/lib/docker_agent_lib.py
Pythyu: this causes issues 
Capture d’écran 2025-06-20 à 11 56 51

tasks/static_quality_gates/lib/docker_agent_lib.py
Pythyu: Indeed it's confusing because it has some part removed, we don't need two values for this function. (An earlier does with awk doing print size; print layer name). Fixing it

tasks/static_quality_gates/lib/docker_agent_lib.py
alopezz: Ah right, you need to double the `{` and `}` to escape them. Though I still wonder whether using `awk` here is a needed _at all_ (which would remove this problem).

Pythyu: I use it to extract the tar.gz layer name from the image artifact. I could maybe figure something out with sed but awk seemed the most natural to me

Pythyu: I didn't know about `{{` :o suggestions applied

alopezz: Can't we just use Python?
```python
with tarfile.open(os.path.join(src_folder, "output.tar"), "r") as output_tar:
    image_content = output_tar.getnames()
```

tasks/quality_gates.py
```diff
@@ -295,6 +325,48 @@ def manual_threshold_update(self, filename="static_gate_report.json"):
...
+    gate_inputs = {"ctx": ctx, "nightly": nightly_run}
+    try:
+        gate_module = getattr(quality_gates_module, gate_name)
+    except AttributeError as e:
+        raise Exit(
+            code=0,
+            message=f"The provided quality gate to debug ({gate_name}) is invalid and wasn't found as part of tasks.static_quality_gates.",
+        ) from e
+
+    # As it is a debug job we do not want the job to actually fail on failures.
```
alopezz, line 363: Given that this seems to be mainly intended to be run when gates have already failed, and that it's set to allow to fail, why not let it fail?

tasks/static_quality_gates/lib/docker_agent_lib.py
```diff
@@ -116,3 +121,66 @@ def generic_docker_agent_quality_gate(gate_name, arch, jmx=False, flavor="agent"
...
+        image_content = ctx.run(
+            f"tar -tvf {src_folder}/output.tar | awk -F' ' '{{print $6;}}'", hide=True
+        ).stdout.splitlines()
+        image_tar_gz = []
+        for content in image_content:
+            if "tar.gz" in content:
+                image_tar_gz.append(content)
+        if image_tar_gz:
+            for image in image_tar_gz:
+                ctx.run(f"cd {src_folder} && tar -xf {image} -C ./out", hide=True)
```
alopezz, line 154: Could be a later improvement but I wonder if it might not be significantly faster to do this without actually performing the extraction. My understanding is that computing the diff only requires the path/name and the size of each file; `tarfile.TarInfo` objects [give you both](https://docs.python.org/3/library/tarfile.html#tarfile.TarInfo.name).

Incidentally I think that would push us closer to a more ideal implementation where we store a summary of a package's contents as a dictionary, which could lead to a simpler diff implementation (for example by using set operations to determine which files appear in both or either packages) as well as to the possibility of just storing the ancestor data directly in s3, avoiding extra artifact downloads.

tasks/static_quality_gates/lib/package_agent_lib.py
```diff
@@ -53,26 +61,102 @@ def generic_package_agent_quality_gate(gate_name, arch, os, flavor, **kwargs):
...
+            package_arch = "x86_64"
+        elif arch == "armhf":
+            package_arch = "armv7hl"
+
+    package_os = sys_os
+    if sys_os == "heroku":
+        package_os = "debian"
+
+    ancestor_download_dir = tempfile.TemporaryDirectory()
+    current_download_dir = tempfile.TemporaryDirectory()
```
alopezz: Is there anything preventing us from using a context manager here?

tasks/static_quality_gates/lib/package_agent_lib.py
```diff
@@ -53,26 +61,102 @@ def generic_package_agent_quality_gate(gate_name, arch, os, flavor, **kwargs):
...
+        filter(lambda job: job.name == build_job_name, ancestor_pipeline.jobs.list(iterator=True))
+    )
+    # Download & extract the artifact from the build_job
+    download_job_artifacts(repo, ancestor_job.get_id(), download_dir)
+
+
+def debug_package_size(ctx, package_os, package_path, ancestor_package_path):
+    # Manual control over temporary folders
+    current_pipeline_extract_dir = tempfile.TemporaryDirectory()
+    ancestor_pipeline_extract_dir = tempfile.TemporaryDirectory()
```
alopezz: Same here, why not use a context manager?

tasks/quality_gates.py
Pythyu: That's true but I'm not sure if there is a point to make it fail either if the job is allowed to fail since the job is supposed to be checked manually 🤔

tasks/static_quality_gates/lib/docker_agent_lib.py
Pythyu: Indeed I think this is would be good future improvement. I'm looking into improving the ancestor s3 data we're sending by adding also the delta computed by PRs to have an easier exception process. I think this change would fit within that PR.

tasks/static_quality_gates/lib/package_agent_lib.py
Pythyu: Not really, this code used to have other temporary file and stacking context manager made it hard to read. Now that it's splited we can use context manager. I'll swap them

tasks/static_quality_gates/lib/package_agent_lib.py
```diff
@@ -53,26 +61,92 @@ def generic_package_agent_quality_gate(gate_name, arch, os, flavor, **kwargs):
...
+        if arch == "arm64":
+            package_arch = "aarch64"
+        elif arch == "amd64":
+            package_arch = "x86_64"
+        elif arch == "armhf":
+            package_arch = "armv7hl"
+
+    package_os = sys_os
+    if sys_os == "heroku":
+        package_os = "debian"
```
alopezz: It would've been nice to separate this into its own function to avoid spreading this into multiple sources.

Pythyu:
New debug run: 
- [package](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1011660367)
- [docker](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1012013279)

alopezz (approved):

tasks/static_quality_gates/lib/package_agent_lib.py
Pythyu: Agreed adding this as a last change before merge

PR #37397, opened 2025-05-27
[autodiscovery] Refactor to reuse code when generating config checks

davidor:
### What does this PR do?

This PR includes a small refactor of the autodiscovery component.

The autodiscovery implementation had two functions that were almost identical: `GetConfigCheck` and `getRawConfigCheck`. The only difference was that the first one scrubbed the configs. This PR extracts the shared code into a reusable function.

CI.


gabedos (approved):
GetConfigCheck refactor looks good

PR #37399, opened 2025-05-27
ebpf: generate full compilation database instead of target specific one

paulcacheux:
### What does this PR do?

This PR simplifies the generation of the compilation database to dump all the rules instead of only the ones specific to the target being generated. This allows the CWS tests to not overwrite the compdb when running ninja.


mbertrone (commented):
The change LGTM from my perspective. I'll defer the approve to someone in the team with more context.

val06 (approved):

PR #37407, opened 2025-05-27
[EUDM] Add System Probe module for software inventory collection

julien-lebot:
### What does this PR do?
This PR introduces a module that will collect Software Inventory in System Probe and expose it via a named pipe to the Agent.
The module will collect the inventory from multiple sources but won't merge them together. This can be done at the Agent level or in the backend.

A follow-up PR will add the core Agent module.

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1463

This PR introduces unit tests for all functionality, along with some integration tests that will compare the collected inventory against known sources.

Currently collects from both the registry and the msi database, resulting in some duplicated entries. Entries can be merged based on a common criteria (Display name, version etc...) but this was not done at the collector level for simplicity.
This can result in extra data being exchanged on the named pipe, although the volume of data should be trivial.
It's currently not possible to chose from which source to collect the inventory - this could be a future improvement.


joepeeples (approved):

brycekahle (approved):

pkg/config/setup/system_probe.go
```diff
@@ -439,6 +440,9 @@ func InitSystemProbeConfig(cfg pkgconfigmodel.Config) {
 	cfg.BindEnvAndSetDefault(join(gpuNS, "max_streams"), 100)
 	cfg.BindEnvAndSetDefault(join(gpuNS, "max_stream_inactivity_seconds"), 30) // 30 seconds by default, includes two checks at the default interval of 15 seconds
 
+	// Software Inventory
+	cfg.BindEnvAndSetDefault(join(swNS, "enabled"), false, "DD_SOFTWARE_INVENTORY_ENABLED")
```
hush-hush, line 444: No need to define a env alias since it's the same as the setting name.
```suggestion
	cfg.BindEnvAndSetDefault(join(swNS, "enabled"), false)
```

hush-hush (changes requested):

pkg/config/setup/system_probe.go
julien-lebot: Fixed, thanks!

hush-hush (approved):

pkg/collector/corechecks/system/winsoftware/integration_test.go
```diff
@@ -0,0 +1,225 @@
...
+						}}
+					} else {
+						psInventory[name] = append(psInventory[name], SoftwareEntry{
+							DisplayName: name,
+							Version:     version,
+							Properties:  map[string]string{"ProductCode": productCode},
+						})
+					}
+				}
+			}
```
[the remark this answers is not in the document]
julien-lebot: Yeah it's a remnant of my testing - I used CSV to quick compare visually the output from the various commands with my own. JSON would add some extra symbols that would clutter the output.

cmd/system-probe/modules/software_inventory_windows.go
```diff
@@ -0,0 +1,65 @@
...
+// SoftwareInventory Factory
+var SoftwareInventory = &module.Factory{
+	Name:             config.InventorySoftwareModule,
+	ConfigNamespaces: []string{"software_inventory"},
+	Fn: func(_ *sysconfigtypes.Config, _ module.FactoryDependencies) (module.Module, error) {
+		log.Infof("Starting the Windows Software inventory module")
+		var err error = nil
+		if err != nil {
+			return nil, fmt.Errorf("unable to start the Windows Software Inventory module: %w", err)
+		}
```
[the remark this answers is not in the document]
julien-lebot: Leftover from debugging - I'll remove thanks!

pkg/collector/corechecks/system/winsoftware/msi_collector.go
```diff
@@ -0,0 +1,166 @@
...
+		entries = append(entries, entry)
+		index++
+	}
+	return entries, warnings, nil
+}
+
+func getMsiProductInfo(productCode []uint16, propertiesToFetch []string) (*SoftwareEntry, error) {
+	// Helper to fetch a property
+	getProp := func(propName string) string {
+		const bufSize = windows.MAX_PATH
```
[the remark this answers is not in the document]
julien-lebot: I based my code on the MSDN sample which uses `MAX_PATH` for the initial buffer size: https://learn.microsoft.com/en-us/windows/win32/msi/determining-installation-context

but yeah, I can't assume that all the properties will be less than 260 chars :D

jeremy-hanna (approved):

pkg/collector/corechecks/system/winsoftware/msi_collector.go
```diff
@@ -0,0 +1,170 @@
...
+	getProp := func(propName string) (string, error) {
+		buf := make([]uint16, windows.MAX_PATH)
+		bufLen := uint32(len(buf))
+		ret := windows.ERROR_MORE_DATA
+		for errors.Is(ret, windows.ERROR_MORE_DATA) {
+			ret = msiGetProductInfo(propName, &productCode[0], &buf[0], &bufLen)
+			if errors.Is(ret, windows.ERROR_SUCCESS) {
+				return windows.UTF16ToString(buf[:bufLen]), nil
+			}
+			buf = make([]uint16, bufLen)
```
clarkb7: UTF16ToString may panic if the string isn't null terminated, and MsiGetProductInfo has annoying size mechanics.

> On input, this is the full size of the buffer, including a space for a terminating null character. If the buffer passed in is too small, the count returned does not include the terminating null character.

Add one to the returned `bufLen` value to ensure there is space for the null term in the allocation.

clarkb7 (approved):
LGTM with small fix to bufLen

cmd/system-probe/modules/software_inventory_windows.go
```diff
@@ -0,0 +1,60 @@
...
+	},
+}
+
+var _ module.Module = &softwareInventoryModule{}
+
+type softwareInventoryModule struct {
+}
+
+func (sim *softwareInventoryModule) Register(httpMux *module.Router) error {
+	httpMux.HandleFunc("/check", utils.WithConcurrencyLimit(1, func(w http.ResponseWriter, _ *http.Request) {
```
iglendd, line 37: Why `check` name and not `inventory` or something like this?

cmd/system-probe/modules/software_inventory_windows.go
```diff
@@ -0,0 +1,60 @@
...
+	return nil
+}
+
+func (sim *softwareInventoryModule) GetStats() map[string]interface{} {
+	return map[string]interface{}{}
+}
+
+func (sim *softwareInventoryModule) Close() {
+
+}
```
iglendd, line 60: What is the purpose of these functions?

.github/CODEOWNERS
```diff
@@ -442,6 +442,7 @@
 /pkg/collector/corechecks/system/windowscertificate/ @DataDog/windows-agent
 /pkg/collector/corechecks/system/winkmem/         @DataDog/windows-agent
 /pkg/collector/corechecks/system/winproc/         @DataDog/windows-agent
+/pkg/collector/corechecks/system/winsoftware/   @DataDog/windows-agent
```
iglendd, line 445: Curious why it is part of corecheck or can/should be referred as a check since if we report in inventory-compatible way it is not a check, is it?

Also, I feel that we need platform-neutral name since it will be running on Mac, Linux etc as well.

pkg/collector/corechecks/system/winsoftware/collector.go
```diff
@@ -0,0 +1,77 @@
...
+type SoftwareCollector interface {
+	// Collect returns a list of software entries and any warnings encountered
+	Collect() ([]*SoftwareEntry, []*Warning, error)
+}
+
+// defaultCollectors returns the default collectors for production use
+func defaultCollectors() []SoftwareCollector {
+	return []SoftwareCollector{
+		&MSICollector{},
+		&RegistryCollector{},
```
iglendd, line 25: Probably will see it later but do we need to capitalize (make them exportable) MSICollector and RegistryCollecter?

pkg/collector/corechecks/system/winsoftware/collector.go
```diff
@@ -0,0 +1,77 @@
...
+func GetSoftwareInventoryWithCollectors(collectors []SoftwareCollector) ([]*SoftwareEntry, []*Warning, error) {
+	var warn []*Warning
+	var allEntries []*SoftwareEntry
+
+	// Collect from all sources
+	for _, collector := range collectors {
+		entries, warnings, err := collector.Collect()
+		if err != nil {
+			// Log error but continue with other collectors
+			warn = append(warn, warnf("error collecting software: %v", err))
```
iglendd, line 44: Does it makes sense here to add not only error and also warnings? Can they be non-null if there is an error present?

pkg/collector/corechecks/system/winsoftware/collector.go
```diff
@@ -0,0 +1,77 @@
...
+			continue
+		}
+
+		// Add any warnings from the collector
+		warn = append(warn, warnings...)
+
+		// Add entries to result list
+		for _, entry := range entries {
+			if entry == nil {
```
iglendd, line 53: Why would our implementation return a nil entry?

pkg/collector/corechecks/system/winsoftware/collector.go
```diff
@@ -0,0 +1,77 @@
...
+			allEntries = append(allEntries, entry)
+		}
+	}
+
+	return allEntries, warn, nil
+}
+
+// trimVersion trims leading zeros from each part of a version string.
+// For example, "4.08.09032" becomes "4.8.9032".
+// That makes it easier to compare versions as strings, as Windows will sometimes trim leading zeros.
```
iglendd, line 66: That is interesting, Do you have any specifics in mind (regarding "as Windows will sometimes trim leading zeros")

pkg/collector/corechecks/system/winsoftware/collector.go
```diff
@@ -0,0 +1,77 @@
...
+	parts := strings.Split(version, ".")
+	for i, part := range parts {
+		trimmed := strings.TrimLeft(part, "0")
+		if trimmed == "" {
+			trimmed = "0"
+		}
+		parts[i] = trimmed
+	}
+	return strings.Join(parts, ".")
+}
```
iglendd, line 77: I wonder if it can be replaced by server module https://pkg.go.dev/golang.org/x/mod/semver

pkg/collector/corechecks/system/winsoftware/collector.go
```diff
@@ -0,0 +1,77 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+//go:build windows
+
+// Package winsoftware implements code to collect installed software from a Windows system.
+package winsoftware
```
iglendd, line 9: Again the same comment about winsoftware packaging naming which is not cross-platform naming here and elsewhere

pkg/collector/corechecks/system/winsoftware/config_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"testing"
+
+	configmock "github.com/DataDog/datadog-agent/pkg/config/mock"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSoftwareInventoryModuleConfig(t *testing.T) {
+	t.Run("via YAML", func(t *testing.T) {
+		cfg := configmock.NewSystemProbe(t)
+		cfg.SetWithoutSource("software_inventory.enabled", true)
```
iglendd, line 20: Were else software_inventory.enabled is used in the code, I have not find it? Also should we add top level section, for this case it udm.software_inventory.enabled, since we will add related configurations. Perhaps even udm.enabled would be sufficient to turn orr subsection on by  default.

pkg/collector/corechecks/system/winsoftware/integration_test.go
```diff
@@ -0,0 +1,225 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
```
iglendd, line 1: why the file is called `integration`, are you testing `collectors` here?

iglendd: Hmm, after reviewing whole module I am not sure that it actually tests it seems to be testing all potentially possible implementation mechanisms? If so do we need to do that all the time?

pkg/collector/corechecks/system/winsoftware/msi_collector.go
```diff
@@ -0,0 +1,172 @@
...
+// Collect enumerates all products in the Windows Installer database.
+// Note: Some entries may only exist in MSI and not in the registry. This can happen with:
+//   - A non-ARP MSI install, e.g. (ARPNOREMOVE=1, ARPNOREPAIR=1, ARPSYSTEMCOMPONENT=1).
+//     This suppresses Add/Remove Programs listing without uninstalling the software.
+//     The product is still fully installed and active.
+//   - A broken/unregistered Uninstall key. If an MSI install was interrupted or rolled back,
+//     the product might remain in "UserData\Products" but not under "Uninstall".
+//   - An intentional exclusion (e.g., via ARPSYSTEMCOMPONENT). Some enterprise MSIs never register under ARP because:
+//     They're intended for scripting or automation and / or they don't want to be visible to end users.
+//     These apps still have valid MSI registration but no Uninstall entry.
```
iglendd, line 66: Should we detect/report that state/discrepancy between registry MSI DB?

pkg/collector/corechecks/system/winsoftware/msi_collector.go
```diff
@@ -0,0 +1,172 @@
...
+func (mc *MSICollector) Collect() ([]*SoftwareEntry, []*Warning, error) {
+	// When making multiple calls to MsiEnumProducts to enumerate all the products, each call should be made from the same thread.
+	runtime.LockOSThread()
```
iglendd, line 69: Can you comment (and remind me) why LockOSThread is required here?

pkg/collector/corechecks/system/winsoftware/win32.go
```diff
@@ -0,0 +1,83 @@
...
+	msi                = syscall.NewLazyDLL("msi.dll")
+	msiGetProductInfoW = msi.NewProc("MsiGetProductInfoW")
+	// For detecting per-user installations
+	msiEnumProductsExW = msi.NewProc("MsiEnumProductsExW")
+	advapi32           = syscall.NewLazyDLL("advapi32.dll")
+	procRegLoadKey     = advapi32.NewProc("RegLoadKeyW")
+	procRegUnLoadKey   = advapi32.NewProc("RegUnLoadKeyW")
+)
+
+func msiEnumProducts(index uint32, productCodeBuf *uint16, context *uint32, sidBuf *uint16, sidLen *uint32) windows.Errno {
```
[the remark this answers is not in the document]
iglendd: agree, hard to find in the MS docs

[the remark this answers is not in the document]
iglendd: Also, do you think it is a value to have added that thin wrapper instead of call the proc directly?

pkg/collector/corechecks/system/winsoftware/msi_collector.go
```diff
@@ -0,0 +1,172 @@
...
+			ret = msiGetProductInfo(propName, &productCode[0], &buf[0], &bufLen)
+			if errors.Is(ret, windows.ERROR_SUCCESS) {
+				return windows.UTF16ToString(buf[:bufLen]), nil
+			}
+			// If the buffer passed in is too small, the count returned does not include the terminating null character.
+			// If the error was not ERROR_MORE_DATA we'll just exit the loop.
+			bufLen++
+		}
+		return "", fmt.Errorf("unexpected return from msiGetProductInfo for %s: %w", propName, ret)
+	}
```
iglendd, line 125: What is is the point to have function pointer here instead of setting up a standalone function?

pkg/collector/corechecks/system/winsoftware/msi_collector.go
```diff
@@ -0,0 +1,172 @@
...
+		entries = append(entries, entry)
+		index++
+	}
+	return entries, warnings, nil
+}
+
+func getMsiProductInfo(productCode []uint16, propertiesToFetch []string) (*SoftwareEntry, error) {
+	// Helper to fetch a property
+	getProp := func(propName string) (string, error) {
+		bufLen := uint32(windows.MAX_PATH)
```
iglendd, line 112: Interesting inconsistence. In the function above you are using hardcoded 256 here you are using MAX_PATH

pkg/collector/corechecks/system/winsoftware/msi_collector.go
```diff
@@ -0,0 +1,172 @@
...
+		ret := windows.ERROR_MORE_DATA
+		for errors.Is(ret, windows.ERROR_MORE_DATA) {
+			buf := make([]uint16, bufLen)
+			ret = msiGetProductInfo(propName, &productCode[0], &buf[0], &bufLen)
```
iglendd, line 116: How slow these functions are in contrast with registry read. Are they reading registry or not?

pkg/collector/corechecks/system/winsoftware/msi_collector.go
```diff
@@ -0,0 +1,172 @@
...
+		if mappedName, exists := mapMsiPropertiesToRegistryProperties[msiPropName]; exists {
+			mappedProperties[mappedName] = msiPropValue
+		} else {
+			mappedProperties[msiPropName] = msiPropValue
+		}
+	}
+
+	name := mappedProperties[displayName]
+	if name == "" {
+		name = properties[msiProductName]
```
iglendd, line 154: Is there a chance that msiProductName does not exist?

pkg/collector/corechecks/system/winsoftware/msi_collector.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	if name == "" {
+		return nil, fmt.Errorf("no valid name found for product")
+	}
+
+	version := mappedProperties[displayVersion]
```
iglendd, line 160: same here

pkg/collector/corechecks/system/winsoftware/msi_collector.go
```diff
@@ -0,0 +1,172 @@
...
+	if version == "" {
+		version = properties[msiVersionString]
+	}
+
+	return &SoftwareEntry{
+		DisplayName: name,
+		Version:     version,
+		InstallDate: mappedProperties[installDate],
```
iglendd, line 168: dmsr here for installDate (if it is missing)

pkg/collector/corechecks/system/winsoftware/registry_collector.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+
+	// 2. All loaded user hives (HKU)
+	hku, err := registry.OpenKey(registry.USERS, "", registry.READ)
+	var loadedUserSids []string
+	if err == nil {
+		// We intentionally ignore the Close error here as it's unlikely to fail
+		// and there's not much we can do about it in a defer
+		defer func() { _ = hku.Close() }()
+		loadedUserSids, _ = hku.ReadSubKeyNames(wantAll)
```
iglendd, line 84: should we track error here as well?

pkg/collector/corechecks/system/winsoftware/registry_collector.go
```diff
@@ -0,0 +1,262 @@
...
+		for _, sid := range loadedUserSids {
+			// Only collect user hives for regular users, not system accounts
```
iglendd, line 86: That is interesting "requirement" in terms of reason. Assertion above does not specifies it. There is a good clues are reported in https://learn.microsoft.com/en-us/windows/win32/api/msi/nf-msi-msienumproductsexw
> Note  The special SID string s-1-5-18 (System) cannot be used to enumerate products or patches installed as per-machine. When dwContext is set to MSIINSTALLCONTEXT_MACHINE only, szUserSid must be NULL.

pkg/collector/corechecks/system/winsoftware/registry_collector.go
```diff
@@ -0,0 +1,262 @@
...
+
+	for _, profile := range userProfiles {
+		// Skip if this profile is already loaded in HKU
+		if slices.Contains(loadedUserSids, profile.SID) {
+			continue
+		}
+
+		// Try to mount and collect from the user's NTUSER.DAT
+		ntuser := filepath.Join(profile.ProfilePath, "NTUSER.DAT")
+		if _, err = os.Stat(ntuser); err == nil {
```
iglendd, line 119: should we report the error?

pkg/collector/corechecks/system/winsoftware/registry_collector.go
```diff
@@ -0,0 +1,262 @@
...
+				for _, p := range paths {
+					entries, warns := collectFromKey(registry.USERS, `temp\`+p.subkey, p.view)
+					warnings = append(warnings, warns...)
+					for _, entry := range entries {
+						entry.UserSID = profile.SID
+						results = append(results, entry)
+					}
+				}
+				err = unmountHive()
+				if err != nil {
```
iglendd, line 130: It is interesting since unmountHive is last step and it will not affect technically content or outcome.

iglendd: And another potential inconsistency error on mount return warning but unmount return error

pkg/collector/corechecks/system/winsoftware/registry_collector.go
```diff
@@ -0,0 +1,262 @@
...
+			// Use the subkey name as the product code
+			properties[msiProductCode] = skey
+			entry := &SoftwareEntry{
+				DisplayName: name,
+				Version:     trimVersion(properties[displayVersion]),
+				InstallDate: properties[installDate],
+				Source:      fmt.Sprintf("%s[%s]", softwareTypeDesktop, sourceRegistry),
+				Properties:  properties,
+				Is64Bit:     view == registry.WOW64_64KEY,
+			}
```
iglendd, line 234: Similar here as with MSI property access which may not contain that key, I think.

pkg/collector/corechecks/system/winsoftware/win32.go
```diff
@@ -0,0 +1,83 @@
...
+		uintptr(unsafe.Pointer(buf)),
+		uintptr(unsafe.Pointer(bufLen)),
+	)
+	return windows.Errno(ret)
+}
+
+// regLoadKey loads a registry hive file into HKU\temp.
+// This is a low-level function that calls RegLoadKey via syscall.
+// Loading a hive requires special privileges (SE_RESTORE_NAME and SE_BACKUP_NAME)
+func regLoadKey(hivePath string) windows.Errno {
```
iglendd, line 58: What is the effect of mount hive on RSS I am curious?

cmd/system-probe/modules/software_inventory_windows.go
julien-lebot: Because the `GetCheck` function expects it: https://github.com/DataDog/datadog-agent/blob/d54f38cc285caedea4b1269a379182800333140f/pkg/system-probe/api/client/client.go#L172

cmd/system-probe/modules/software_inventory_windows.go
julien-lebot: It's part of the Module interface: https://github.com/DataDog/datadog-agent/blob/d84840abe8e635729662e77919935c312084fbcb/pkg/system-probe/api/module/common.go#L27-L31

pkg/collector/corechecks/system/winsoftware/collector.go
julien-lebot: We can move this file to a different package once we get to the Mac implementation

pkg/collector/corechecks/system/winsoftware/collector.go
julien-lebot: I think this is for goland modules, which must begin with `v` which is not our case

pkg/collector/corechecks/system/winsoftware/collector.go
julien-lebot: `Microsoft .NET Framework 4.8.1 SDK` in registry is written as "4.08.09032" I'm normalizing it so that it possible to compare.

pkg/collector/corechecks/system/winsoftware/config_test.go
julien-lebot: It's defined here: https://github.com/DataDog/datadog-agent/blob/e6ac6287e6f93932d8478f332418e41cc07b928b/pkg/config/setup/system_probe.go#L445

We can revisit the flags name to change it to EUDM enabled

pkg/collector/corechecks/system/winsoftware/msi_collector.go
julien-lebot: The detection of invalid software should be done at the Agent module level or at the backend but not directly in System Probe.

pkg/collector/corechecks/system/winsoftware/msi_collector.go
julien-lebot: See comment on line 68.

pkg/collector/corechecks/system/winsoftware/msi_collector.go
julien-lebot: I used MAX_PATH for `msiGetProductInfo` like the sample code for the MSDN: https://learn.microsoft.com/en-us/windows/win32/msi/determining-installation-context

The 256 above is for the SID size.

pkg/collector/corechecks/system/winsoftware/win32.go
julien-lebot: We should benchmark or check in Datadog

pkg/collector/corechecks/system/winsoftware/win32.go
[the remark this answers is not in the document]
julien-lebot: Sorry for missing this comment, I'll address it in: https://github.com/DataDog/datadog-agent/pull/38072

PR #37417, opened 2025-05-27
[ACIX-894] Refactor gitlab linter tasks logic

Ishirui:
### What does this PR do?

Cleans up and unifies the way gitlabci linting tasks run, especially in regards to how they generate the "full gitlab config". 
Also fixes some wrong behavior with `gitlab_change_paths` and `job_change_path`, these two tasks were linting only the 'root' pipeline (defined in `.gitlab-ci.yml`), but not any "called pipelines", like `.gitlab/deploy_containers`

**Call graph before the PR:**
Gitlab CI Linters Call Graph 0 Explained
> Key:
> - **Grey arrows:** Calls
> - **Purple arrows:** Need file as input
> - **Blue:** Functions handling the "full gitlab config", i.e. with multiple entrypoints, i.e. including "called pipelines"
> - **Red:** Functions handling only a single config object, i.e. only one entrypoint (usually the main `.gitlab-ci.yml`)

**Call graph after the PR:**
Gitlab CI Linters Call Graph 3 Raw
Notable changes:
- Fixed `job_change_path` and `gitlab_change_paths` to lint all entrypoints instead of only the root `.gitlab-ci.yml`
- Removed `get_gitlab_ci_configuration` and `generate_gitlab_full_configuration`, instead making the tasks call `get_all_gitlab_ci_configurations` directly. These functions had too little useful functionality and made the code much harder to read.
- Renamed the recursive helper function `get_ci_configurations` to `_traverse_config_search_triggers`
- `test_gitlab_configuration` no longer calls methods leading to `resolve_gitlab_ci_configurations`, instead taking the config object as input. This avoids calling the `/lint` endpoint one extra time, improving efficiency. It also does not depend on other methods in the call graph anymore, so is not represented above.
- Made `get_gitlab_ci_lintable_jobs` take config files as input only optionally, falling back to generating the config objects if necessary

### Motivation

Before trying to implement a new top-level task for running all gitlabci-related linters [ACIX-470], I needed to make sure that different linting tasks would not try to regenerate the unified gitlab config multiple times, to ensure these tasks would run as quickly as possible.

But due to the organic nature of how these tasks were built over time, there were many different paths to generating this config, and many different functions all with the same goal.
With a more unified method of generating this config, we make the code more readable and efficient, while assuring no task tries to regenerate the config if it does not need to.

- All affected tasks run without errors on the current gitlab config
- Inputting a broken gitlab config (simple change for each of the linter tasks) does result in an error
- Some tasks are already covered by unit tests

- Performance impact: By fixing the two tasks mentioned above, they become a bit slower as they have to handle multiple entrypoints instead of only the root one (should be very minimal)
- Behavioral change: To the best of my knowledge, there should be none, but I might have missed some transformation in one of the helper functions that is not being executed anymore
- Loss of habit: As with any significant refactor, anyone familiar with the old code might not recognize it. In my opinion it is more easily understandable and readable in its current state, however.

### Additional Notes

[ACIX-470]: https://datadoghq.atlassian.net/browse/ACIX-470?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tasks/gitlab_helpers.py
+++ b/tasks/gitlab_helpers.py
@@ -289,7 +289,7 @@ def print_entry_points(ctx):
     """Prints gitlab ci configuration entry points."""
 
     print(color_message('info:', Color.BLUE), 'Fetching entry points...')
-    entry_points = get_all_gitlab_ci_configurations(ctx, filter_configs=True, clean_configs=True)
+    entry_points = get_all_gitlab_ci_configurations(ctx, postprocess_options={"do_filtering": True})
 
     print(len(entry_points), 'entry points:')
     for entry_point, config in entry_points.items():
--- a/tasks/gitlab_helpers.py
+++ b/tasks/gitlab_helpers.py
@@ -14,8 +14,8 @@ from tasks.kernel_matrix_testing.ci import get_kmt_dashboard_links
 from tasks.libs.ciproviders.gitlab_api import (
     compute_gitlab_ci_config_diff,
     get_all_gitlab_ci_configurations,
-    get_gitlab_ci_configuration,
     get_gitlab_repo,
+    post_process_gitlab_ci_configuration,
     print_gitlab_ci_configuration,
     resolve_gitlab_ci_configuration,
 )
@@ -270,14 +270,13 @@ def print_ci(
         This requires a full api token access level to the repository
     """
 
-    yml = get_gitlab_ci_configuration(
-        ctx,
-        input_file,
-        job=job,
+    yml = resolve_gitlab_ci_configuration(ctx, input_file, partial_resolve=partial_resolve, git_ref=git_ref)
+
+    yml = post_process_gitlab_ci_configuration(
+        yml,
+        jobs={job},
         clean=clean,
         expand_matrix=expand_matrix,
-        git_ref=git_ref,
-        partial_resolve=partial_resolve,
         keep_special_objects=keep_special_objects,
     )
 
--- a/tasks/gitlab_helpers.py
+++ b/tasks/gitlab_helpers.py
@@ -252,7 +252,7 @@ def print_ci(
     keep_special_objects: bool = False,
     expand_matrix: bool = False,
     git_ref: str | None = None,
-    with_lint: bool = True,
+    partial_resolve: bool = True,
 ):
     """Prints the full gitlab ci configuration.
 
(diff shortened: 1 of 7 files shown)


brycekahle (approved):

tasks/gitlab_helpers.py
```diff
@@ -252,7 +252,7 @@ def print_ci(
     keep_special_objects: bool = False,
     expand_matrix: bool = False,
     git_ref: str | None = None,
-    with_lint: bool = True,
+    partial_resolve: bool = True,
```
CelianR: NIT: Maybe `no_lint` or `resolve_only_includes` would be more explicit?

tasks/libs/linter/gitlab.py
```diff
@@ -64,34 +65,45 @@ def list_get_parameter_calls(file):
...
     Args:
         diff_file: Path to the diff file used to build MultiGitlabCIDiff obtained by compute-gitlab-ci-config.
         config_file: Path to the full gitlab ci configuration file obtained by compute-gitlab-ci-config.
+        > If none of these are passed, the full config will be generated automatically, but this will be slower.
+
+    Returns:
+        A (jobs, full_config) tuple.
+        `jobs` is itself a tuple of (job_name: str, job_contents: dict). If `only_names` is True, it will be a simple list of all the job names.
+        `full_config` is a gitlabci config object, of the same structure as returned by `get_all_gitlab_ci_configurations`
+
```
CelianR:
```suggestion
```

tasks/libs/linter/gitlab.py
```diff
@@ -64,34 +65,45 @@ def list_get_parameter_calls(file):
...
-                if is_leaf_job(job, job_contents)
-            ]
-    else:
+    if diff_file:
+        # Special handling of diff files
         with open(diff_file) as f:
             diff = MultiGitlabCIDiff.from_dict(yaml.safe_load(f))
 
-        full_config = diff.after
+        configs = diff.after  # type: ignore
```
CelianR, line 90: NIT: Do we really want to add `type: ignore`? Maybe it's better to not type `configs`

CelianR (approved):
LGTM if the linters are passing 👍

tasks/libs/linter/gitlab.py
Ishirui: IIRC here the `diff.after` is typed as `dict[str, dict] | None` - so basically the `# type: ignore` is just to ignore the possibility of it being None. I guess I could add an explicit `is not None` check to accomplish the same thing ?

In general I prefer doing it this way (typing the var and using an ignore on assignment) because then we get more IDE suggestions. It might lead to more runtime errors though, is there a convention for this sort of thing ?

tasks/gitlab_helpers.py
Ishirui: I like `resolve_only_includes` - a bit longer but definitely clearer :+1:

tasks/libs/linter/gitlab.py
CelianR: Most of the time we don't use this but as you wish, I think adding an explicit check is too much for this

PR #37428, opened 2025-05-27
[AGTMETRICS-245]Domain resolver updates health check when API keys are changed.

StephenWakely:
### What does this PR do?

Ensures the forwarder healthchecker is updated when API keys are refreshed.

### Motivation

The healthchecker maintains it's own list of API keys to check and needs these to be updated when the API keys need to be refreshed. Rather than duplicating a lot of code from the domain resolver to ensure the API keys are updated properly, handling duplicates etc... this changes it so the resolver holds a reference to the healthchecker and it will notify it when it has updated the API keys.

Added some unit tests and have tested manually.

Adding a reference to the health checker in the resolvers does add more dependencies and make things a bit more complex overall...


comp/forwarder/defaultforwarder/resolver/domain_resolver.go
```diff
@@ -136,15 +150,20 @@ func updateAdditionalEndpoints(resolver DomainResolver, setting string, config c
 	removed := missing(oldKeys, newKeys)
 	added := missing(newKeys, oldKeys)
 
+	health := resolver.GetForwarderHealth()
+	if health != nil {
```
remeh, line 154: nit
```suggestion
	if health := resolver.GetForwarderHealth(); health != nil {
```

comp/forwarder/defaultforwarder/resolver/domain_resolver.go
```diff
@@ -97,6 +107,10 @@ func OnUpdateConfig(resolver DomainResolver, log log.Component, config config.Co
 		newAPIKey, ok2 := newValue.(string)
 		if ok1 && ok2 {
 			resolver.UpdateAPIKey(setting, oldAPIKey, newAPIKey)
+			health := resolver.GetForwarderHealth()
+			if health != nil {
```
remeh, line 111: nit
```suggestion
			if health := resolver.GetForwarderHealth(); health != nil {
```

comp/forwarder/defaultforwarder/resolver/domain_resolver.go
```diff
@@ -210,19 +229,27 @@ func (r *SingleDomainResolver) GetAPIKeyVersion() int {
...
-			missing = append(missing, scrubber.HideKeyExceptLastFiveChars(key))
+			missing = append(missing, key)
 		}
 	}
 
 	return missing
 }
 
+// scrubKeys scrubs the API key to avoid leaking the key when logging.
+func scrubKeys(keys []string) []string {
```
remeh, line 245: nit: no real need to create a new slice in this function, you could do it in-place. 

something along these lines:
```go
// scrubKeys scrubs the API key to avoid leaking the key when logging.
func scrubKeys(keys []string) []string {
	for i, key := range keys {
		keys[i] = scrubber.HideKeyExceptLastFiveChars(key)
	}
	return keys
}
```

I'm returning the `keys` slice only for good measure it's not even necessary since it can't grow. (I didn't check it compiles, just a suggestion).

Feel free to discard if you prefer working with a new slice for some reasons.

comp/forwarder/defaultforwarder/forwarder_health_test.go
remeh, line 1: Since you extracted `scrubKeys` to its own function, what do you think about adding an unit test to it, making sure it's properly modifying all entries in the slice.

comp/forwarder/defaultforwarder/resolver/domain_resolver.go
```diff
@@ -136,15 +150,20 @@ func updateAdditionalEndpoints(resolver DomainResolver, setting string, config c
 	removed := missing(oldKeys, newKeys)
 	added := missing(newKeys, oldKeys)
 
+	health := resolver.GetForwarderHealth()
+	if health != nil {
+		health.UpdateAPIKeys(resolver.GetBaseDomain(), removed, added)
+	}
+
```
remeh, line 157: While this is correct, manually calling `scrubKeys` within following log lines feels a bit brittle, if someone were to add or a change log line without paying too much attention it isn't ideal.

What do you think of doing the scrubbing before the logging logic, here?

```suggestion

removed = scrubKeys(removed)
added = scrubKeys(added)
```

remeh (approved):
Made some comments, but LGTM overall!

Make sure to look at the failing jobs in the CI.

hush-hush (approved):

PR #37344, opened 2025-05-23
[NPM-4440] Move packet capture to separate package, add Sink interface

pimlu:
This is a recreation of #36991. I did a stacked PR on top of [\[NPM-4440\] Add TracerouteSerial variant](https://github.com/DataDog/datadog-agent/pull/36936), but after I merged it, for some reason #36991 wasn't automatically retargeted to the `main` branch like normal. So then I rebased, and disaster struck... 😅 

### What does this PR do?
1. Consolidates packet-capture related stuff by moving to the `packets` package
2. Adds the `packets.Sink` interface which is a platform-agnostic interface to send packets
3. Changes `packets.Source` to start at the IP layer, and strip the ethernet layer -- this is because it turns out Windows won't give us the ethernet layer

### Motivation
This is necessary for my next PR which will create the TCP SYN implementation of TracerouteDriver (the main goal of this series)

### Describe how you validated your changes

Updated test suite should pass:
```
sudo go test -timeout 30s -tags linux,linux_bpf,npm,process,test github.com/DataDog/datadog-agent/pkg/networkpath/traceroute/packets
```
SACK traceroute should still work properly:
```
go build github.com/DataDog/datadog-agent/pkg/networkpath/traceroute/sack/portable_sack
```

PR #37433, opened 2025-05-27
[CNM-4440] Add TCP SYN traceroute driver

pimlu:
Stacked on [\[NPM-4440\] Move packet capture to separate package, add Sink interface](https://github.com/DataDog/datadog-agent/pull/37344)

### What does this PR do?
* Adds `tcpDriver`, an implementation of `TracerouteDriver` for TCP SYN.
* Adds generated mocks for `Source` and `Sink`, the platform-agnostic packet interfaces `tcpDriver` uses.
* Adds a test suite for correctness of `tcpDriver` which uses the mocks
* Small changes to the `common` package to support this
* Added a CLI tool `portable_syn` which runs a traceroute with `tcpDriver`.

### Motivation
Working towards replacing the traceroute implementations with platform-agnostic and (mostly) IPv4/IPv6 agnostic versions.

Remaining work:
1. Add UDP driver which implements `TracerouteDriver`
2. Make our Windows `RawConn` implement `Source` and `Sink` so all our traceroutes work on Windows too

I added a big test suite for `tcpDriver` in `tcp_driver_test.go` which you can run via:
```
go test -timeout 30s -tags linux,linux_bpf,npm,process,test github.com/DataDog/datadog-agent/pkg/networkpath/traceroute/tcp
```

Tests don't cover paris traceroute mode right now, but hopefully we are removing that mode.


pkg/networkpath/traceroute/packets/packet_sink_unix.go
```diff
@@ -24,14 +24,30 @@ type sinkUnix struct {
 
 var _ Sink = &sinkUnix{}
 
-// NewSinkUnix returns a new SinkUnix implementing packet sink
+// NewSinkUnix returns a new sinkUnix implementing packet sink
 func NewSinkUnix(addr netip.Addr) (Sink, error) {
```
hmahmood, line 28: Can this just be `NewSink`?

pkg/networkpath/traceroute/tcp/tcp_driver.go
```diff
@@ -0,0 +1,279 @@
...
+// SendProbe sends a traceroute packet with a specific TTL
+func (t *tcpDriver) SendProbe(ttl uint8) error {
+	packetID, seqNum := t.getNextPacketIDAndSeqNum(ttl)
+	_, buffer, _, err := t.config.createRawTCPSynBuffer(packetID, seqNum, int(ttl))
+	if err != nil {
+		return fmt.Errorf("tcpDriver SendProbe failed to createRawTCPSynBuffer: %w", err)
+	}
+
+	log.Tracef("sending probe with ttl=%d, packetID=%d, seqNum=%d", ttl, packetID, seqNum)
+	t.storeProbe(probeData{
```
hmahmood, line 134: Is `SendProbe` called in parallel? If it is, `storeProbe` will not store the probes in order of increasing (or decreasing) ttl.

hmahmood (commented):
One comment about file naming: e.g., for the files under `tcp` folder, the prefix `tcp_` seems redundant. Same for the files under the `packets` folder.

pkg/networkpath/traceroute/packets/packet_sink_unix.go
pimlu: I think this is a good idea, but if it's alright with you, I plan to rename this after this and Fabrice's chain of PRs get merged to reduce conflicts

pkg/networkpath/traceroute/tcp/tcp_driver.go
pimlu: There's one goroutine calling `SendProbe` and one calling `ReceiveProbe`. So `SendProbe` can overlap in time with `ReceiveProbe` but never with itself.

hmahmood: Is parallel traceroute going to change this assumption?

pimlu: That's with parallel traceroute -- it's mainly parallel in the sense that multiple probes are on the wire at the same time, but the actual sending and receiving are serialized (within two goroutines).

* [Goroutine 1](https://github.com/DataDog/datadog-agent/blob/e7b7917af9600acb5ca9fc5b217a114d48a75744/pkg/networkpath/traceroute/common/traceroute_parallel.go#L77): Calls `SendProbe()` in a loop
* [Goroutine 2](https://github.com/DataDog/datadog-agent/blob/e7b7917af9600acb5ca9fc5b217a114d48a75744/pkg/networkpath/traceroute/common/traceroute_parallel.go#L95): Calls `ReceiveProbe()` in a loop

hmahmood (approved):

PR #37453, opened 2025-05-28
[autodiscovery] Include all collected configs in flare and configcheck verbose

davidor:
### What does this PR do?

This PR fixes a bug in the autodiscovery component.

The `agent configcheck --verbose` command and flares were supposed to display a section with all unresolved configurations (both matched and unmatched), as shown here: https://github.com/DataDog/datadog-agent/blob/3dfd9d29a724757a73c8bde25472039e9c101194/pkg/flare/config_check.go#L45

However, this broke in a previous version, and the list of unresolved configs is now always empty. This is because autodiscovery relies on its template cache store to retrieve them, but that store is never updated: https://github.com/DataDog/datadog-agent/blob/3dfd9d29a724757a73c8bde25472039e9c101194/comp/core/autodiscovery/autodiscoveryimpl/autoconfig.go#L642

I believe that having all collected configurations (both matched and unmatched) is helpful for debugging. It’s sometimes important to see exactly what configurations autodiscovery collected. That’s why I think it’s worth fixing this, even though it has been broken for several versions.

Thanks to this change we can also delete some unused code in `comp/core/autodiscovery/autodiscoveryimpl/templatecache.go` and `comp/core/autodiscovery/autodiscoveryimpl/store.go`, but I'll do that in a separate PR.

This is an example of what `agent configcheck --verbose` shows in the “Collected configs” section after this change (only the first few lines are shown, as there are many configurations). The first section, which lists the configured checks, remains unchanged.

```

... Configured checks here (this part remains unchanged) ...

=== Collected configs (matched and unmatched) ===

Auto-discovery IDs: cluster-agent
check_name: datadog_cluster_agent
init_config: null
instances:
- prometheus_url: http://%%host%%:5000/metrics
logs_config: null


Auto-discovery IDs: proxyv2,proxyv2-rhel8
check_name: istio
init_config: null
instances:
- istio_mesh_endpoint: http://%%host%%:15020/stats/prometheus
  send_histograms_buckets: true
  tag_by_endpoint: false
logs_config: null


Auto-discovery IDs: kyototycoon
check_name: kyototycoon
init_config: null
instances:
- report_url: http://%%host%%:1978/rpc/report
logs_config: null


Auto-discovery IDs: kubedns-amd64,k8s-dns-kube-dns-amd64,k8s-dns-kube-dns
check_name: kube_dns
init_config: null
instances:
- prometheus_endpoint: http://%%host%%:10055/metrics
  tags:
  - dns-pod:%%host%%
logs_config: null

...

```

I added unit tests and also manually confirmed that  `agent configcheck --verbose` and the flares now include all collected configurations.

### DIFF
--- a/pkg/flare/config_check.go
+++ b/pkg/flare/config_check.go
@@ -8,6 +8,7 @@ package flare
 import (
 	"fmt"
 	"io"
+	"strings"
 
 	"github.com/fatih/color"
 
@@ -43,13 +44,11 @@ func PrintConfigCheck(w io.Writer, cr integration.ConfigCheckResponse, withDebug
 			}
 		}
 		if len(cr.Unresolved) > 0 {
-			fmt.Fprintf(w, "\n=== %s Configs ===\n", color.YellowString("Unresolved"))
-			for ids, configs := range cr.Unresolved {
-				fmt.Fprintf(w, "\n%s: %s\n", color.BlueString("Auto-discovery IDs"), color.YellowString(ids))
-				fmt.Fprintf(w, "%s:\n", color.BlueString("Templates"))
-				for _, config := range configs {
-					fmt.Fprintln(w, config.String())
-				}
+			fmt.Fprintf(w, "\n=== %s configs (matched and unmatched) ===\n", color.MagentaString("Collected"))
+			for _, config := range cr.Unresolved {
+				adIdentifiers := strings.Join(config.ADIdentifiers, ",") // Will be empty for non-template configs
+				fmt.Fprintf(w, "\n%s: %s\n", color.BlueString("Auto-discovery IDs"), color.YellowString(adIdentifiers))
+				fmt.Fprintln(w, config.String())
 			}
 		}
 	}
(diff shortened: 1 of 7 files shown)


pkg/flare/config_check.go
```diff
@@ -43,13 +44,11 @@ func PrintConfigCheck(w io.Writer, cr integration.ConfigCheckResponse, withDebug
...
 		}
 		if len(cr.Unresolved) > 0 {
-			fmt.Fprintf(w, "\n=== %s Configs ===\n", color.YellowString("Unresolved"))
-			for ids, configs := range cr.Unresolved {
-				fmt.Fprintf(w, "\n%s: %s\n", color.BlueString("Auto-discovery IDs"), color.YellowString(ids))
-				fmt.Fprintf(w, "%s:\n", color.BlueString("Templates"))
-				for _, config := range configs {
-					fmt.Fprintln(w, config.String())
-				}
+			fmt.Fprintf(w, "\n=== %s configs (matched and unmatched) ===\n", color.MagentaString("Collected"))
```
davidor, line 47: I changed the color because yellow can give the impression that it’s a warning, but this is just informational.

rahulkaukuntla (approved):

drichards-87 (approved):

gabedos (approved):
Looks good to me to expose all of the configs AD collected with the verbose command. Thanks for catching this!

davidor:
Rebased on top of main to fix conflicts

PR #37459, opened 2025-05-28
Fix DWARF register macros, get dyninst further to working on ARM64

grantseltzer:
### What does this PR do?

In currently committed headers of the new dynamic instrumentation implementation the macros which reference specific registers have been set up incorrectly, leading to index out of bounds errors on ARM64. Also added a missing header file.

### Motivation

Building on ARM

Manually built as we're still early in active development


piob-io (approved):
Thank you!

brycekahle (approved):

PR #37462, opened 2025-05-28
APM: Resolve shutdown panic from send to closed channel

ajgajg1134:
### What does this PR do?
Resolves an issue where during shutdown we could panic trying to send on a closed channel. This would result in the payload that was being processed being lost.

### Motivation
While this is rare and only occurs during shutdown it's non-ideal to panic here when we already have the payload and could instead finish processing it before shutting down.

This issue seems incredibly rare, I was unable to directly reproduce it locally but my hypothesis is that the general mechanism is as follows:
1. A trace payload begins being decoded by `handleTraces`
2. The agent is stopped via some signal (e.g. SIGTERM)
3. The receiver's Stop() function is called and the output channel is closed.
4. The handler go routine from step 1 attempts to send to the closed output channel triggering a panic.

I was unable to reproduce this panic locally, however there was already a wait group in the code here that was lost at some point - indicating that this may have been a known issue. Given all this and the relatively small impact here I elected to implement this straightforward fix and see if this improves things internally.

I ran this trace-agent locally emitting traces to it, I then sent a SIGTERM and ensured the server properly shutdown as expected.


drichards-87 (approved):
Left a suggestion and comment from Docs and approved the PR.

ichinaski (approved):

PR #37473, opened 2025-05-28
Document auto_multi_line_detection in config template

jszwedko:
### What does this PR do?

Documents the `logs_config.auto_multi_line_detection` setting in the config template.

### Motivation

This is an option we'd like to encourage use of and is commonly enabled by customers so we'd like to add it to the template.

Will rely on CI to flag failures.

More users will use this feature. If there are bugs, more users will be exposed to them. We don't anticipate this, though, and the value of the feature outweighs the risk.


pkg/config/config_template.yaml
```diff
@@ -987,6 +987,14 @@ api_key:
   #     name: <RULE_NAME>
   #     pattern: <RULE_PATTERN>
 
+  ## @param auto_multi_line_detection - boolean - optional - default: false
+  ## @env DD_LOGS_CONFIG_AUTO_MULTI_LINE_DETECTION - boolean - optional - default: false
+  ## Enable automatic aggregation of multiline logs for common log patterns.
```
drichards-87, line 992:
```suggestion
  ## Enable automatic aggregation of multi-line logs for common log patterns.
```

drichards-87 (approved):
Left a small suggestion from Docs and approved the PR.

dustmop (approved):

PR #37474, opened 2025-05-29
[CNM] fix size of tcp_recvmsg eBPF map

akarpz:
### What does this PR do?

the size of this map shouldn't scale with the # of connections. it scales with maxactive which is constant.

### DIFF
--- a/pkg/network/ebpf/c/tracer/maps.h
+++ b/pkg/network/ebpf/c/tracer/maps.h
@@ -60,7 +60,7 @@ BPF_HASH_MAP(udp_sendpage_args, __u64, struct sock *, 1024)
  * Map to hold struct sock parameter for tcp_recvmsg/tcp_read_sock calls
  * to be used in kretprobe/tcp_recvmsg/tcp_read_sock
  */
-BPF_HASH_MAP(tcp_recvmsg_args, __u64, struct sock *, 0)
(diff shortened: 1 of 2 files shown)


pkg/network/ebpf/c/tracer/maps.h
```diff
@@ -60,7 +60,7 @@ BPF_HASH_MAP(udp_sendpage_args, __u64, struct sock *, 1024)
  * Map to hold struct sock parameter for tcp_recvmsg/tcp_read_sock calls
  * to be used in kretprobe/tcp_recvmsg/tcp_read_sock
  */
-BPF_HASH_MAP(tcp_recvmsg_args, __u64, struct sock *, 0)
+BPF_HASH_MAP(tcp_recvmsg_args, __u64, struct sock *, 2048)
```
hmahmood, line 63: Isn't this a constant in userspace? We should set it there instead imo.

akarpz: I was under the impression our paradigm was to set constants here, and then anything dynamic we would set in userspace (the other changed file in this PR)

akarpz: that makes sense. I can rewrite this to set the size in userspace as a function of maxactive

PR #37482, opened 2025-05-29
[AGTMETRICS-247] Add orchestrator_additional_endpoints to the list of allowed secret paths for API key refresh

StephenWakely:
### What does this PR do?

- Adds orchestrator_additional_endpoints to the list of allowed paths to receive API key secret updates.
- Allows for secrets to be specified when `additional_endpoints` and `orchestrator_additional_endpoints` are configured with JSON objects, typically used when configuring via envirenment variables.

### Motivation

Most additional_endpoints are configured with an `additional_endpoints` config key. Orchestrator is configured slightly differently with an `orchestrator_additional_endpoints` key, so wasn't receiving API key updates.

To allow secrets via JSON objects an number of changes were needed around the config loading and secret resolution to deserialise and then reserialise the field during lookup.


rahulkaukuntla (approved):

StephenWakely:
Closing in favour of #37548.

PR #37488, opened 2025-05-29
Hardcode DD_CC and related vars in repackaging task

alopezz:
### What does this PR do?

Removes the dependance in the `*_PATH` vars introduced in https://github.com/DataDog/datadog-agent-buildimages/pull/849 in favor of just hardcoding the values like it's done in CI jobs.

### Motivation

We don't have a satisfying solution for setting those variables in the build images while avoiding the issues they currently cause. Even if we don't like hardcoding these here, we prefer to not proliferate extra environment variables for now.

I ran the task and ran e2e tests on the produced artifacts with success.


chouquette (approved):
LGTM, but I wonder if it wouldn't be simpler to normalize the arch and have a common update block that adds `{arch}-unknown-linux-gnu-*` to the env instead.
Definitely not blocking though

PR #37495, opened 2025-05-29
add preaggregation pipeline dual shipping support

lukesteensen:
### What does this PR do?

Add a configurable ability to dual ship series payloads to a new preaggregation endpoint, with those payloads always enriched with the full set of host tags. We also refactor some existing logic for MRF and local autoscaling failover to unify some of how we approach these three similar features.

### Motivation

The current implementation is largely aimed to enable further testing of the new preaggregation service in staging.

### Additional Notes

TPOT-18


pkg/config/utils/endpoints.go
```diff
@@ -165,6 +165,26 @@ func GetMultipleEndpoints(c pkgconfigmodel.Reader) (map[string][]APIKeys, error)
 			Keys:              []string{c.GetString("multi_region_failover.api_key")},
 		}}
 	}
+
+	// Populate preaggregation endpoint (only if unique)
+	//
+	// TODO(?): This assumes we wouldn't have a unique preaggregation API key
+	// without a unique preaggregation URL, but we should validate that.
+	if c.GetBool("enable_preaggr_pipeline") {
+		preaggURL := c.GetString("preaggr_dd_url")
```
vickenty: Would it make sense to follow the pattern of MRF just above and make a namespace for config keys? 

```suggestion
	if c.GetBool("pregaggregation_pipeline.enabled") {
		preaggURL := c.GetString("pregaggregation_pipeline.dd_url")
```

pkg/serializer/serializer.go
```diff
@@ -357,7 +357,7 @@ func (s *Serializer) SendIterableSeries(serieSource metrics.SerieSource) error {
 	failoverActiveForMRF, allowlistForMRF := s.getFailoverAllowlist()
 	failoverActiveForAutoscaling, allowlistForAutoscaling := s.getAutoscalingFailoverMetrics()
 	failoverActive := (failoverActiveForMRF && len(allowlistForMRF) > 0) || (failoverActiveForAutoscaling && len(allowlistForAutoscaling) > 0)
-	pipelines := make([]metricsserializer.Pipeline, 0, 3)
+	pipelines := make([]metricsserializer.Pipeline, 0, 4)
```
vickenty: We can probably leave counting to Go here:

```suggestion
	pipelines := []metricsserializer.Pipeline{}
```

pkg/config/setup/preaggregation.go
```diff
@@ -0,0 +1,16 @@
...
+
+package setup
+
+import (
+	pkgconfigmodel "github.com/DataDog/datadog-agent/pkg/config/model"
+)
+
+func setupPreaggregation(config pkgconfigmodel.Setup) {
+	config.BindEnvAndSetDefault("enable_preaggr_pipeline", false)
+	config.BindEnvAndSetDefault("preaggr_dd_url", "https://api.datad0g.com")
```
dustmop: Is this default url value correct? Is it supposed to be using staging?

comp/forwarder/defaultforwarder/default_forwarder.go
```diff
@@ -502,6 +502,23 @@ func (f *DefaultForwarder) createAdvancedHTTPTransactions(endpoint transaction.E
 	for _, payload := range payloads {
 		for domain, dr := range f.domainResolvers {
 			drDomain, destinationType := dr.Resolve(endpoint) // drDomain is the domain with agent version if not local
+
+			if payload.Destination == transaction.PreaggrOnly {
+				preaggURL := f.config.GetString("preaggregation.dd_url")
+				primaryURL := f.config.GetString("dd_url")
+				
+				// If preaggr uses same URL as primary, allow all primary resolvers
```
vickenty: > all primary resolvers

I may not be reading correctly, but there is only one resolver per domain, so one resolver for primary. Did you mean API keys?

comp/forwarder/defaultforwarder/default_forwarder.go
```diff
@@ -502,6 +502,23 @@ func (f *DefaultForwarder) createAdvancedHTTPTransactions(endpoint transaction.E
...
+				// If preaggr uses different URL, only allow that specific domain
+				if preaggURL != primaryURL && domain != preaggURL {
+					continue
+				}
+				endpoint = endpoints.PreaggrSeriesEndpoint
+			}
+			// TODO(?): If the preaggregation.dd_url is the same as the primary dd_url,
+			// we will also inherit any additional API keys from the
+			// configuration of that site, meaning we'll send preaggr payloads
+			// for each of those orgs. Not sure if this is a problem or not.
```
vickenty: Doesn't it mean that metrics are written to the org from the agent as well as from pre-aggregator, negating any wins from pre-aggregation?

Do you plan to resolve the todo items before merging?

comp/forwarder/defaultforwarder/default_forwarder.go
```diff
@@ -502,6 +502,23 @@ func (f *DefaultForwarder) createAdvancedHTTPTransactions(endpoint transaction.E
 	for _, payload := range payloads {
 		for domain, dr := range f.domainResolvers {
 			drDomain, destinationType := dr.Resolve(endpoint) // drDomain is the domain with agent version if not local
+
+			if payload.Destination == transaction.PreaggrOnly {
+				preaggURL := f.config.GetString("preaggregation.dd_url")
+				primaryURL := f.config.GetString("dd_url")
```
vickenty: Primary url can be derived from `site` as well as specified explicitly, so the check below is probably going to miss common cases (see how GetMultipleEndpoints, that feeds domain resolver does it https://github.com/datadog/datadog-agent/blob/00a177679056c4d1e1edf913570bfbe8f63aa708/pkg/config/utils/endpoints.go#L142)

comp/forwarder/defaultforwarder/default_forwarder.go
```diff
@@ -502,6 +502,23 @@ func (f *DefaultForwarder) createAdvancedHTTPTransactions(endpoint transaction.E
...
+				
+				// If preaggr uses same URL as primary, allow all primary resolvers
+				// If preaggr uses different URL, only allow that specific domain
+				if preaggURL != primaryURL && domain != preaggURL {
+					continue
+				}
+				endpoint = endpoints.PreaggrSeriesEndpoint
```
vickenty: We know the correct endpoint when we build the payload, would it make sense to submit it to the forwarder with the correct endpoint from the beginning?

pkg/config/setup/preaggregation.go
```diff
@@ -0,0 +1,16 @@
...
+
+package setup
+
+import (
+	pkgconfigmodel "github.com/DataDog/datadog-agent/pkg/config/model"
+)
+
+func setupPreaggregation(config pkgconfigmodel.Setup) {
+	config.BindEnvAndSetDefault("preaggregation.enabled", false)
+	config.BindEnvAndSetDefault("preaggregation.dd_url", "https://api.datad0g.com")
```
vickenty: Even for testing, we shouldn't ship staging urls with the agent. If we don't know what the default should be for now, let's leave it empty.

comp/forwarder/defaultforwarder/default_forwarder.go
lukesteensen: Yes good catch, though reading this again I'm not sure that we actually need to do this comparison at all.

comp/forwarder/defaultforwarder/default_forwarder.go
lukesteensen: Yes, I think this is a leftover claude comment I should have cleaned up. It is indeed API keys that we'll reuse from the primary if they are the same.

comp/forwarder/defaultforwarder/default_forwarder.go
lukesteensen: That's a very good question. The current pattern is somewhat awkward for a few different reasons, but this might be one we can resolve. I'll take a look.

comp/forwarder/defaultforwarder/default_forwarder.go
lukesteensen: Yes, the current implementation is purely additive for testing purposes. To see actual efficiency gains we will need to stop sending the original series data to normal intake, which we won't do until we're ready to rely purely on the preaggregated data.

I think the resolution of the TODO is that this behavior is fine for the time being. We have a larger task to figure out how these preaggregation payloads should interact with normal configuration (e.g. multiple API keys, additional endpoints, etc), which will likely result in matching the behavior anyway.

pkg/config/setup/preaggregation.go
lukesteensen: Agreed, I will remove the default.

pkg/config/setup/preaggregation.go
lukesteensen: Sorry I missed this comment! I've removed the default.

dustmop (approved):
LGTM for files owned by agent-configuration

tobz (approved):

vickenty (approved):

PR #37498, opened 2025-05-30
Fix dev docs publishing

ofek:
### Motivation

On PRs we want to only run when something has changed but we want to publish anew on every commit to the `main` branch because parts of the documentation are fetched externally e.g. the tooling installation instructions https://github.com/DataDog/datadog-agent/blob/69e3afee8f4c95d25fbde4be99d0fe8fa694d021/docs/public/.hooks/inject_variables.py#L24-L35

### DIFF
--- a/.github/workflows/docs-dev.yml
+++ b/.github/workflows/docs-dev.yml
@@ -4,9 +4,6 @@ on:
   push:
     branches:
     - main
-    paths:
-    - docs/**
-    - .github/workflows/docs-dev.yml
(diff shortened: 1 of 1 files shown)


Ishirui (approved):
Good catch ! Do we also want to change the PR trigger ? I don't see why they should be different, unless it's just to prevent extra noisiness on PR checks

ofek:
Yeah it's really just about extra noise, trying to reduce friction.

PR #37499, opened 2025-05-30
Prevent simultaneous use of --no-cache and --no-cache-filter

alopezz:
### What does this PR do?

Make `--no-cache` and `--no-cache-filter` exclusive when building Docker images.

### Motivation

Nightly pipelines triggered this untested code path which set both `--no-cache` and `--no-cache-filter`, which are not compatible, resulting in failures such as [this](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/960508277#L131), with this error message:

```
error: --no-cache and --no-cache-filter cannot currently be used together
```

I'll set the variable to trigger the "nightly" code path to confirm that it works before merging.
UPDATE: Done: https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/960747514

### DIFF
--- a/.gitlab/container_build/docker_linux.yml
+++ b/.gitlab/container_build/docker_linux.yml
@@ -11,6 +11,8 @@
       else
         export ECR_RELEASE_SUFFIX=${CI_COMMIT_TAG+-release}
       fi
+    # XXX: Test 'nightly' behaviour
+    - BUCKET_BRANCH="nightly"
     # Caching setup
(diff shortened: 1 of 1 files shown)


spencergilbert (approved):

alopezz:
Note: I'm skipping the merge queue and merging directly because this does not impact the results of the performance regression detector but is needed to fix our nightly pipelines.

PR #37504, opened 2025-05-30
discovery: module: add /services endpoint alongside existing /check endpoint

Yumasi:
- Add new /services endpoint that accepts pids parameter and bypasses port retry logic
- Restore original /check endpoint functionality with port retry logic
- Separate test files: impl_linux_test.go for /check, impl_services_test.go for /services
- Update params.go to handle both heartbeat and pids parameters
- Maintain existing CheckResponse format for /check endpoint
- Use new ServicesResponse format for /services endpoint

### What does this PR do?

Adds a new `/services` endpoint to the discovery module that accept explicit PID lists and bypasses the port retry logic used by the existing `/check` endpoint.

### Motivation

DSCVR-134

Change covered by unit tests


pkg/collector/corechecks/servicediscovery/module/impl_linux.go
```diff
@@ -1188,3 +1209,71 @@ func (s *discovery) getServices(params params) (*model.ServicesResponse, error)
...
+// getServiceWithoutRetry extracts service information for a PID without port retry logic.
+// Unlike getService(), this function immediately returns nil if no ports are found,
+// rather than tracking retry attempts. This is used by the /services endpoint.
+func (s *discovery) getServiceWithoutRetry(context parsingContext, pid int32) *model.Service {
+	if s.shouldIgnorePid(pid) {
+		return nil
+	}
+	if s.shouldIgnoreComm(pid) {
+		return nil
+	}
```
vitkyrka, line 1245: Why can't these two checks be done in the caller in the core agent?

pkg/collector/corechecks/servicediscovery/module/impl_linux.go
```diff
@@ -1188,3 +1209,71 @@ func (s *discovery) getServices(params params) (*model.ServicesResponse, error)
...
+
+	ports, err := s.getPorts(context, pid)
+	if err != nil {
+		return nil
+	}
+	if len(ports) == 0 {
+		return nil
+	}
+
+	rss, err := getRSS(pid)
```
vitkyrka, line 1255: Why is this endpoint calculating RSS?

pkg/collector/corechecks/servicediscovery/module/impl_linux.go
```diff
@@ -1188,3 +1209,71 @@ func (s *discovery) getServices(params params) (*model.ServicesResponse, error)
...
+	if err != nil {
+		return nil
+	}
+
+	var info *serviceInfo
+	cached, ok := s.cache[pid]
```
vitkyrka, line 1261: Why is this putting stuff into a cache here in SP?

pkg/collector/corechecks/servicediscovery/impl_linux.go
```diff
@@ -21,7 +21,7 @@ func init() {
 }
 
 type linuxImpl struct {
-	getDiscoveryServices func(client *sysprobeclient.CheckClient) (*model.ServicesResponse, error)
+	getDiscoveryServices func(client *sysprobeclient.CheckClient) (*model.CheckResponse, error)
```
vitkyrka, line 24: It would have been nice to have kept this rename outside of this PR since it's not strictly necessary.

pkg/collector/corechecks/servicediscovery/module/impl_linux.go
```diff
@@ -1188,3 +1209,71 @@ func (s *discovery) getServices(params params) (*model.ServicesResponse, error)
 
 	return response, nil
 }
+
+// getServices processes a list of PIDs and returns service information for each.
+// This is used by the /services endpoint which accepts explicit PID lists and bypasses
+// the port retry logic used by the /check endpoint.
```
vitkyrka, line 1215: _Why_ is it bypassing the port retry logic? I guess it's because it will be done in the core agent? Maybe that should be mentioned here.

pkg/collector/corechecks/servicediscovery/module/impl_linux.go
Yumasi: Good catch. I also forgot to remove the CPU stats check from the tests, so I fixed that too.

pkg/collector/corechecks/servicediscovery/module/impl_linux.go
Yumasi: The ignored PID has indeed its place in the CA. I left shouldIgnoreComm in SP on purpose though, since it requires moving the config option from SP to the CA, and I think that keep be done in a later PR.

vitkyrka (approved):

PR #37506, opened 2025-05-30
Disallow local VMs for KMT remote setup

usamasaqib:
### What does this PR do?
This PR makes KMT remote setup mode explicit, and disallows setup of local VMs when KMT is initialized with `--remote-setup-only` parameter.

### Motivation
Match KMT behavior to user intent.


mbertrone (commented):
Overall the PR improves the clarity of kmt init command. Thanks for that!
A few nits can be found below

mbertrone (approved):

PR #37527, opened 2025-05-31
Change sub-types to TypeID instead of full Type redefinition

grantseltzer:
### What does this PR do?

Flattens the type structure of the IR to have TypeID for referencing of other types within things like struct fields, or pointed-to-types.

### Motivation

While writing decoding code I realized that i'd have to redefine the Type that already exists in the Type field of Program.

Forthcoming discussion with team.

### DIFF
--- a/pkg/dyninst/ir/types.go
+++ b/pkg/dyninst/ir/types.go
@@ -114,7 +114,7 @@ type PointerType struct {
 	GoTypeAttributes
 
 	// Pointee is the type that the pointer points to.
-	Pointee Type
+	Pointee TypeID
 }
 
 func (t *PointerType) irType() {}
@@ -139,7 +139,7 @@ type Field struct {
 	// Offset in the parent structure.
 	Offset uint32
 	// Type is the type of the field.
-	Type Type
+	Type TypeID
 }
 
 // ArrayType is an array type in the target program.
@@ -152,7 +152,7 @@ type ArrayType struct {
 	// HasCount is true if the array has a count.
 	HasCount bool
 	// Element is the type of the element in the array.
-	Element Type
+	Element TypeID
 }
 
 func (t *ArrayType) irType() {}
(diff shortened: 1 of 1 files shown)


grantseltzer:
I realize this has a lot of implications for code needing to be updated in irgen, but starting with this for the sake of discussion.

piob-io:
Note that these type structures do not have to be repeated. These type fields are pointers behind the interface. The instances that are pointed too may be, and will be, reused. You can see how it is done here (directly, not via irgen): https://github.com/DataDog/datadog-agent/blob/241a2339b88ce4439dcb3e3d36c5b649cbdc209f/pkg/dyninst/compiler/compile_test.go (file from a pending commit). In particular there will only be single instance of string type implementation.

In a more meta sense, the pointers are like ids, but with extra benefit that they point directly to memory, and don't need an extra id map. I don't view these pointers as redundancy with the id map, because both reference same instances of type implementations. However they are found, they are the same.

I am not sure I follow the third point. The structure will be predictable, but I don't see how it could be broken if the IR gen decides not to necessarily reuse the instances of the type implementation objects (as long as it is still correct data within them).

And the reason for using the pointers, not IDs, is that you don't have to pass around a "type resolver" map around across all the functions that need to process these type instances. Not a huge deal to do that, but also a nice convenience not to have to.

tylfin (approved):

ajwerner:
To clarify, are you saying that the primary motivation for this change is to make it easier to construct ir by hand for testing? Is there any motivation that's not related to testing ergonomics?

ajwerner:
I will say that there are ways in which using an ID-oriented approach can be more efficient from an object-size perspective and a GC overhead perspective, but we're not in a place of optimizing at that level just yet.

ajwerner:
I think another thing that maybe wasn't said but should have been said is there should be an explicit assumption that for a given instance of a program there's exactly one instance of a Type with a given ID and that you can check that with pointer equality.

PR #37528, opened 2025-05-31
WINA-1171: CPU info rework

guohdd:
### What does this PR do?
This PR addresses the issue of inaccurate cpu_cores reported by "gohai". Specifically, The system.cpu.num_cores metric could be wrong on certain Windows platforms. This PR intends to retrieve the correct CPU data from Windows systems across Windows platforms and independent of CPU architecture, providing a generic solution to CPU metric data.
Relevant JIRA cases are:
* WINA-1171: Inaccurate cpu_cores reported by "gohai"
* AGENT-13192: wrong number of cpu cores in dashboard
* AGENT-12417: Inaccurate cpu_cores reported by "gohai"

### Motivation
Both external customers and internal users have reported similar issues, the Windows Agent team also reproduced the issue. Inaccurate data could lead to user confusion.

The issue can be reproduced relatively easily, which helped develop and validate the solution. The changes were verified by running the test command on the systems which reproduced the problem. Use the following command for local test: go test -tags=test -v ./pkg/gohai/cpu
### Additional Notes

Based on the investigation, the previous implementation retrieves data from native system calls on Windows, and there were assumptions of certain data sizes and hardcoded data offset values, which was potentially risky and resulted in incorrect data retrieved. This PR removes the above assumptions and made data collection independent of CPU architecture., which means we let the OS do the heavy lifting and use the data provided by the system to figure out the location of each CPU record.


releasenotes/notes/windows-agent-cpu-info-18e54faf1ffbd15a.yaml
```diff
@@ -0,0 +1,12 @@
...
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+fixes:
+  - |
+    * Fixed inaccurate cpu_cores reported by gohai
+    * JIRA: https://datadoghq.atlassian.net/browse/WINA-1171
```
brett0000FF:
```suggestion
```
I would recommend removing this and/or adding whatever context is relevant from that Jira card for Datadog Agent users outside of the Datadog org.

brett0000FF (approved):

releasenotes/notes/windows-agent-cpu-info-18e54faf1ffbd15a.yaml
guohdd: Thanks Brett, I have updated PR summary based on your feedback.

pkg/gohai/cpu/cpu_windows.c
```diff
@@ -0,0 +1,93 @@
+// This file is licensed under the MIT License.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-present Datadog, Inc.
+
+#include "cpu_windows.h"
+#include <windows.h>
+#include <stdio.h>
+
+// computeCoresAndProcessors gets CPU information using Windows API
+int computeCoresAndProcessors(CPU_INFO* cpuInfo) {
```
alexn-dd, line 25: I believe the naming convention in DD CGO is Snake Case with underscores.

pkg/gohai/cpu/cpu_windows.c
```diff
@@ -0,0 +1,93 @@
...
+    SYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX* buffer = NULL;
+    BOOL done = FALSE;
+    DWORD returnLength = 0;
+    DWORD byteOffset = 0;
+
+    // Initialize the struct
+    memset(cpuInfo, 0, sizeof(CPU_INFO));
+
+    // First call to get required buffer size
+    GetLogicalProcessorInformationEx(RelationAll, NULL, &buflen);
```
alexn-dd: Please check the return value first before GLE to verify the expected failure.

pkg/gohai/cpu/cpu_windows.c
```diff
@@ -0,0 +1,93 @@
...
+                    }
+                }
+                break;
+
+            case RelationNumaNode:
+                cpuInfo->numaNodeCount++;
+                break;
+
+            case RelationCache:
+                switch (ptr->Cache.Level) {
```
alexn-dd: Should explicitly handle the default case and not fallthrough.

pkg/gohai/cpu/cpu_windows.go
```diff
@@ -59,96 +46,110 @@ type SYSTEM_INFO struct {
...
 
-func countBits(num uint64) (count int) {
-	count = 0
-	for num > 0 {
-		if (num & 0x1) == 1 {
-			count++
-		}
-		num >>= 1
+func extract(caption, field string) string {
+	re := regexp.MustCompile(fmt.Sprintf("%s [0-9]* ", field))
```
dustmop, line 48: From what I can tell, this function is only called with `field == "Family"`. How about:
1) Rename this to `extractFamily`
2) Get rid of the `field` variable and inline "Family" into the regex's text
2) Move the regexp to the top-level of this file and get rid of the need to compile it here

That let's us avoid the possibility of this panic'ing at runtime, in case someone in the future attempt to use this function in a less straight-forward context.

pkg/gohai/cpu/cpu_windows.go
```diff
@@ -59,96 +46,110 @@ type SYSTEM_INFO struct {
...
-	cpuInfo.CPUPkgs = utils.NewValueFrom(uint64(cpus.pkgcount), cpuerr)
-	cpuInfo.CPUNumaNodes = utils.NewValueFrom(uint64(cpus.numaNodeCount), cpuerr)
-	cpuInfo.CPUCores = utils.NewValueFrom(uint64(cpus.corecount), cpuerr)
-	cpuInfo.CPULogicalProcessors = utils.NewValueFrom(uint64(cpus.logicalcount), cpuerr)
-	cpuInfo.CacheSizeL1Bytes = utils.NewValueFrom(uint64(cpus.l1CacheSize), cpuerr)
-	cpuInfo.CacheSizeL2Bytes = utils.NewValueFrom(uint64(cpus.l2CacheSize), cpuerr)
-	cpuInfo.CacheSizeL3Bytes = utils.NewValueFrom(uint64(cpus.l3CacheSize), cpuerr)
+	if mhz, _, err := k.GetIntegerValue("~MHz"); err == nil {
+		cInfo.Mhz = float64(mhz)
+	}
```
dustmop, line 92: How about log.Error if the err is not nil? Same for the next few lines. If there are known cases where this fails, and we don't want to be too noisy, maybe explicitly check for them.

pkg/gohai/cpu/cpu_windows.go
```diff
@@ -59,96 +46,110 @@ type SYSTEM_INFO struct {
...
-	l3CacheSize   uint32 // layer 3 cache size
-	//nolint:unused
-	relationGroups int // number of cpu relation groups
-	//nolint:unused
-	maxProcsInGroups int // max number of processors
-	//nolint:unused
-	activeProcsInGroups int // active processors
+// cpuInfo holds CPU information
+type cpuInfo struct {
+	corecount           int
```
dustmop: I'd suggest leaving the per-field comments, I don't see a need to drop them.

pkg/gohai/cpu/cpu_windows.go
```diff
@@ -59,96 +46,110 @@ type SYSTEM_INFO struct {
...
+	logicalcount        int
+	pkgcount            int
+	numaNodeCount       int
+	relationGroups      int
+	maxProcsInGroups    int
+	activeProcsInGroups int
+	l1CacheSize         uint64
+	l2CacheSize         uint64
+	l3CacheSize         uint64
+	VendorID            string
```
dustmop: Since these fields get copied over to the exported `Info` struct, how about lower-casing these next few fields to make them unexported? Switching visibility mid-declaration seems strange to me.

pkg/gohai/cpu/cpu_windows.c
```diff
@@ -0,0 +1,93 @@
...
+                cpuInfo->pkgcount++;
+                break;
+
+            case RelationGroup:
+                cpuInfo->relationGroups = ptr->Group.MaximumGroupCount;
+                for (WORD i = 0; i < ptr->Group.ActiveGroupCount; i++) {
+                    cpuInfo->maxProcsInGroups += ptr->Group.GroupInfo[i].MaximumProcessorCount;
+                    cpuInfo->activeProcsInGroups += ptr->Group.GroupInfo[i].ActiveProcessorCount;
+                }
+                break;
```
iglendd, line 94: Is this information will be used anywhere now or in the future? Should we remove this and simplify code a bit

pkg/gohai/cpu/cpu_windows.c
guohdd: Great catch. Since this call always fails because we provide NULL for the buffer, the actual error code is what we need to check. I will add better clarity.

pkg/gohai/cpu/cpu_windows.c
guohdd: sure, adding the default case will make the implementation clearer.

pkg/gohai/cpu/cpu_windows.c
guohdd: Good question. Though this is not used or reported, it can help understand or debug a large system,. I think it's beneficial to have the information in place. Please let me know if there are further concerns.

pkg/gohai/cpu/cpu_windows.go
guohdd: Looking at the code history, it appears the intention of function `extract `was implemented to retrieve value from a certain string pattern. While current implementation only uses it to extract CPU's "Family" number, it might be useful to keep it "as is" in case we need to get some other values in a similar fashion in the future.  However, I am open to make the changes if there's a big concern of possible runtime panic.

pkg/gohai/cpu/cpu_windows.go
guohdd: Yes, it'll be beneficial to log errors.

pkg/gohai/cpu/cpu_windows.go
guohdd: Good catch, I missed it, will add back.

pkg/gohai/cpu/cpu_windows.go
guohdd: Ah, I should have paid more attention.  Thanks for catching this.

pkg/gohai/cpu/cpu_windows.c
iglendd: Perfect logic and totally correct but odd for the eyes since the reasoning implicit. For experience Windows developer GetLastError check nearly always predicated by success/failure check since in former case normally GetLastError makes no sense or even could lead to a subtle errors. And so visually for a reviewer it would force to mentally run this calculation and figure out that this code is factually correct. This effort and the comment can be avoided by following common pattern. This is my 2 cents since the code is factually correct.

pkg/gohai/cpu/cpu_windows.c
julien-lebot: We're not super consistent with this in the repo, see https://github.com/DataDog/datadog-agent/blob/cb0da4e182c7b1ed857924c184da74e29d5becb7/comp/etw/impl/session.c#L24-L37 for example.
I think we should consider this a follow-up where we clean-up the entire cgo codebase.

pkg/gohai/cpu/cpu_windows.c
```diff
@@ -0,0 +1,101 @@
...
+                // Count logical processors in this core
+                for (WORD i = 0; i < ptr->Processor.GroupCount; i++) {
+                    GROUP_AFFINITY* groupAffinity = &ptr->Processor.GroupMask[i];
+                    ULONG64 mask = groupAffinity->Mask;
+                    while (mask) {
+                        if (mask & 1) {
+                            cpuInfo->logicalcount++;
+                        }
+                        mask >>= 1;
+                    }
```
julien-lebot: **❓ question:** ‏ I wonder if we could use GCC's __builtin_popcountll here

pkg/gohai/cpu/cpu_windows.go
```diff
@@ -1,51 +1,38 @@
 // This file is licensed under the MIT License.
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
-// Copyright © 2015 Kentaro Kuribayashi <kentarok@gmail.com>
```
julien-lebot, line 3: **❓ question:** ‏ Why was this removed?

pkg/gohai/cpu/cpu_windows.go
```diff
@@ -1,51 +1,38 @@
...
-
-// SYSTEM_INFO contains information about the current computer system.
+// systemInfo contains information about the current computer system.
 // This includes the architecture and type of the processor, the number
 // of processors in the system, the page size, and other such information.
 // see https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/ns-sysinfoapi-system_info
 //
 //nolint:revive
-type SYSTEM_INFO struct {
+type systemInfo struct {
```
julien-lebot: **✏️ suggestion:** ‏ I believe the `SYSTEM_INFO` name was to keep in sync with the Windows API https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/ns-sysinfoapi-system_info - I think we should keep it as is.

pkg/gohai/cpu/cpu_windows.go
```diff
@@ -59,96 +46,118 @@ type SYSTEM_INFO struct {
...
+	} else {
+		log.Errorf("failed to get Identifier from registry: %v", err)
+	}
 
-	si := getSystemInfo()
-	cpuInfo.Model = utils.NewValue(strconv.Itoa(int((si.wProcessorRevision >> 8) & 0xFF)))
-	cpuInfo.Stepping = utils.NewValue(strconv.Itoa(int(si.wProcessorRevision & 0xFF)))
+	// Get system info for model and stepping
+	var si systemInfo
+	var mod = windows.NewLazyDLL("kernel32.dll")
```
julien-lebot: **✏️ suggestion:** ‏ 
```suggestion
	var mod = windows.NewLazySystemDLL("kernel32.dll")
```

releasenotes/notes/windows-agent-cpu-info-18e54faf1ffbd15a.yaml
julien-lebot: @guohdd this was marked as resolved but I still see the JIRA link in the release note

julien-lebot (commented):
Left a few comments and suggestions

pkg/gohai/cpu/cpu_windows.go
iglendd: > **✏️ suggestion:** ‏

good catch

pkg/gohai/cpu/cpu_windows.c
iglendd: Agree, manual bit count is overrated. Was tempted to comment on that too. There is a Windows API https://learn.microsoft.com/en-us/windows/win32/api/winnt/nf-winnt-populationcount64 or intrinsics https://learn.microsoft.com/en-us/cpp/intrinsics/popcnt16-popcnt-popcnt64?view=msvc-170

julien-lebot: Can't use the `__popcnt64` intrinsic because we are compiling using GCC but I like the `PopulationCount64` method!

pkg/gohai/cpu/cpu_windows.c
```diff
@@ -0,0 +1,101 @@
...
+
+    if (!GetLogicalProcessorInformationEx(RelationAll, buffer, &buflen)) {
+        free(buffer);
+        return GetLastError();
+    }
+
+    // Walk through the buffer
+    byteOffset = 0;
+    while (byteOffset < buflen) {
+        SYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX* ptr = (SYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX*)((char*)buffer + byteOffset);
```
alexn-dd: Let's keep the maximum length to about 80-100 chars.
It is very easy to miss subtle code behavior while reading if a line is clipped or word-wrapped.

pkg/gohai/cpu/cpu_windows.c
```diff
@@ -0,0 +1,101 @@
...
+                    }
+                }
+                break;
+
+            case RelationNumaNode:
+                cpuInfo->numaNodeCount++;
+                break;
+
+            case RelationCache:
+                switch (ptr->Cache.Level) {
```
alexn-dd: It is difficult to maintain embedded switch blocks.
I suggest refactoring this inner switch block into an inline function.

pkg/gohai/cpu/cpu_windows.c
iglendd: But this is a list of binary structures. Also I doubt 100 and maybe even 1k bytes would be enough 64 processers alone

pkg/gohai/cpu/cpu_windows.go
guohdd: Is that expected copyright? The file has following;
`// Copyright 2014-present Datadog, Inc.`

pkg/gohai/cpu/cpu_windows.go
guohdd: Great catch!  It's not exactly the same struct as the Windows SYSTEM_INFO, and I think there's some risk of using a Windows struct in a go source. I'd like to move this also to cgo. Will have another revision for review.

pkg/gohai/cpu/cpu_windows.go
guohdd: Again, plan to move this to cgo.

pkg/gohai/cpu/cpu_windows.c
guohdd: The buffer size needed could be large on a large system. On my 8-core system (11th Gen Intel(R) Core(TM) i9-11950H @ 2.60GHz), the buffer size needed is 2344 bytes.

releasenotes/notes/windows-agent-cpu-info-18e54faf1ffbd15a.yaml
lrbhatia: @guohdd We should remove this internal jira link from this release notes.
It could simply be: "Fixed a bug where the system.cpu.num_cores metric could be incorrect on certain Windows platforms."

pkg/gohai/cpu/cpu_windows.c
```diff
@@ -0,0 +1,108 @@
+// This file is licensed under the MIT License.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-present Datadog, Inc.
+
+#include "cpu_windows.h"
+
+inline void getCacheSize(SYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX* ptr, CPU_INFO* cpuInfo) {
```
alexn-dd: Prefix with `static` to reduce visibility of the function.

pkg/gohai/cpu/cpu_windows.c
alexn-dd: Sorry, I meant the length of the line of code, and not the buffer size.

```
 SYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX* ptr =
     (SYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX*)((char*)buffer + byteOffset);
 ```

Comment (guohdd) [reply pkg/gohai/cpu/cpu_windows.c]: got it, thanks!

File: pkg/gohai/cpu/cpu_windows.c [hunk 17]
```diff
@@ -0,0 +1,113 @@
...
+            default:
+                // NOTE: ignore other relationships
+                break;
+        }
+
+        byteOffset += ptr->Size;
+    }
+
+    free(buffer);
+    return 0;
```
alexn-dd, line 105: Should return `ERROR_SUCCESS`, but fine for now.

alexn-dd (approved):

pkg/gohai/cpu/cpu_windows.c
guohdd: Good point, will keep in mind.

dustmop (approved):

PR #37544, opened 2025-06-02
[AGTMETRICS-245]Only set endpoint valid if it is true to avoid unsetting it.

StephenWakely:
@
### What does this PR do?

#37428 introduced a bug whereby we would unset the api valid status if an endpoint was valid, so the health of the endpoint was determined by the last domain check. Before it was valid as long as one of the domains were valid. 

This reverts to that behaviour.

### Motivation

Before that #37428, the health checker essentially did

```
valid := false

for each endpoint {
  if endpoint has valid api key {
    valid = true
  }
}

return valid
```

That PR changes it to

```
valid := false

for each endpoint {
   valid = isEndpointValid(endpoint)
}

return valid
}

func isEndpointValid(endpoint) (valid bool) {
  if endpoint is valid {
    valid = true
  }

  return
}
```

In other words it changes it from setting valid = true to valid being whatever the validity of the last endpoint is...

- Have added unit tests to test this. 
- Deployed a custom image to a staging cluster to verify it started up successfully.


comp/forwarder/defaultforwarder/forwarder_health.go
```diff
@@ -289,7 +289,17 @@ func (fh *forwarderHealth) checkValidAPIKey() bool {
 	fh.keyMapMutex.Unlock()
 
 	for domain, apiKeys := range keysPerDomain {
-		apiError, validKey = fh.checkValidAPIKeys(domain, apiKeys)
+		endpointApiError, endpointValidKey := fh.checkValidAPIKeys(domain, apiKeys)
+
+		// Only set the valid if the endpoint valid is true to ensure
+		// we don't unset the flag if one endpoint is not valid.
+		if endpointValidKey {
+			validKey = true
```
remeh: An alternative would be:

```
validKey = validKey || endpointValidKey
```

same for the error below.

Discard if you prefer your implementation.

remeh (approved):

PR #37546, opened 2025-06-02
Use raw tracepoints for net_dev_queue

usamasaqib:
### What does this PR do?

This PR adds support to use raw_tracepoints for net_dev_queue probes.

### Motivation

The recursion protection for kprobes and tracepoints is fairly heavy handed. It works by disallowing eBPF programs to nest on the same CPU, using a per-cpu variable [bpf_prog_active](https://elixir.bootlin.com/linux/latest/source/kernel/trace/bpf_trace.c#L117).

For the probe in question this is problematic because it runs from within IRQ context, which means it can nest on to another eBPF program running from within a user context. This causes the probe to be skipped in such cases. Depending on the number of eBPF programs attached and the load on a system these misses can be very high.

Raw tracepoints on the other hand have more precise nesting control. An eBPF program attached to a raw tracepoint is skipped only if it is nesting a running instance of the same program. Converting this probe to use a raw tracepoint instead can therefore completely eliminate misses due to nesting, since these only run from a single context.

Covered by existing UTs.

### Additional Notes

Statistics on misses of raw tracepoints are exposed via recursion counts. These stats were added from [kernel versions 6.7+](https://github.com/torvalds/linux/commit/dd8657894c11b03c6eb0fd53fe9d7fec2072d18b).

To observe these misses happening, we can use https://github.com/DataDog/ami-builder/pull/191 (running kernel 6.8), run system-probe, generate load, and then use the eBPF check to get recursion counts for all attached programs.

### DIFF
--- a/pkg/network/tracer/offsetguess/tracer.go
+++ b/pkg/network/tracer/offsetguess/tracer.go
@@ -95,7 +95,7 @@ func NewTracerOffsetGuesser() (OffsetGuesser, error) {
 				{ProbeIdentificationPair: idPair(probes.IP6MakeSkb)},
 				{ProbeIdentificationPair: idPair(probes.IP6MakeSkbPre470)},
 				{ProbeIdentificationPair: idPair(probes.TCPv6ConnectReturn), KProbeMaxActive: 128},
-				{ProbeIdentificationPair: idPair(probes.NetDevQueue)},
+				{ProbeIdentificationPair: idPair(probes.NetDevQueueTracepoint)},
 			},
 		},
 	}, nil
@@ -218,7 +218,7 @@ func (*tracerOffsetGuesser) Probes(c *config.Config) (map[probes.ProbeFuncName]s
 		return nil, fmt.Errorf("could not kernel version: %w", err)
 	}
 	if kv >= kernel.VersionCode(4, 7, 0) {
-		enableProbe(p, probes.NetDevQueue)
+		enableProbe(p, probes.NetDevQueueTracepoint)
 	}
 
 	if c.CollectTCPv6Conns || c.CollectUDPv6Conns {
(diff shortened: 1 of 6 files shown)


pkg/network/tracer/offsetguess/tracer.go
```diff
@@ -95,7 +95,7 @@ func NewTracerOffsetGuesser() (OffsetGuesser, error) {
 				{ProbeIdentificationPair: idPair(probes.IP6MakeSkb)},
 				{ProbeIdentificationPair: idPair(probes.IP6MakeSkbPre470)},
 				{ProbeIdentificationPair: idPair(probes.TCPv6ConnectReturn), KProbeMaxActive: 128},
-				{ProbeIdentificationPair: idPair(probes.NetDevQueue)},
+				{ProbeIdentificationPair: idPair(probes.NetDevQueueTracepoint)},
```
hmahmood, line 98: Don't you have to do a similar feature check here as for the tracer?

hmahmood:
I think you also need to update this: https://github.com/DataDog/datadog-agent/blob/main/pkg/network/ebpf/c/prebuilt/offset-guess.c#L340

pkg/network/tracer/offsetguess/tracer.go
usamasaqib: The feature check is only required when using raw_tracepoints. For the offset guesser we always default to tracepoints which was always the default behavior.

usamasaqib:
> I think you also need to update this: https://github.com/DataDog/datadog-agent/blob/main/pkg/network/ebpf/c/prebuilt/offset-guess.c#L340

I figured we can leave offset guess as-is. I do not see a strong need to use raw_tracepoints here.

hmahmood (approved):

PR #37549, opened 2025-06-02
[CWS-3907] Enable security profiles for systemd cgroups

mftoure:
### What does this PR do?
This PR enabled security profiles for systemd cgroups, thus, enabling anomaly detection events for these cgroups.


pkg/security/ebpf/c/include/hooks/raw_syscalls.h
```diff
@@ -20,7 +20,7 @@ int sys_enter(struct _tracepoint_raw_syscalls_sys_enter *args) {
     fill_container_context(proc_cache_entry, &event.container);
 
     // check if this event should trigger a syscall drift event
-    if (is_anomaly_syscalls_enabled() && event.container.container_id[0] != 0) {
+    if (is_anomaly_syscalls_enabled() && (event.container.container_id[0] != 0 || event.container.cgroup_context.cgroup_flags != 0)) {
```
spikat: could we only rely on cgroup_flags here? (and remove the check on container_id

pkg/security/resolvers/tags/resolver_linux.go
```diff
@@ -119,15 +120,49 @@ func (t *LinuxResolver) checkTags(pendingWorkload *Workload) {
...
+		// Extract service name from the cgroup ID path
+		serviceName := cgroupID
+		if parts := strings.Split(cgroupID, "/"); len(parts) > 0 {
+			serviceName = parts[len(parts)-1]
+		}
+		// Create basic tags for the cgroup
+		newTags = []string{
+			"service:" + serviceName,
+			"version:latest",
+		}
```
spikat: discussed offline but: we should already have the ~same logic in activity dumps right? should we reuse it instead (and get the service from sbom if avail?)

pkg/security/security_profile/profile/profile.go
```diff
@@ -646,6 +646,12 @@ func (p *Profile) MatchesSelector(entry *model.ProcessCacheEntry) bool {
 		if entry.ContainerID == workload.ContainerID {
 			return true
 		}
+		// Check if the cgroup IDs match when the workload has a cgroup context and no container ID
+		if entry.ContainerID == "" && workload.ContainerID == "" &&
+			entry.CGroup.CGroupID != "" && workload.CGroupContext.CGroupID != "" &&
+			entry.CGroup.CGroupID == workload.CGroupContext.CGroupID {
+			return true
+		}
```
spikat: we should move this bloc before the first if `entry.ContainerID == workload.ContainerID`

pkg/security/security_profile/secprofs.go
```diff
@@ -330,10 +389,17 @@ func (m *Manager) unlinkProfileMap(profile *profile.Profile, workload *tags.Work
...
+	if workload.ContainerID != "" {
+		if err := m.securityProfileMap.Delete([]byte(workload.ContainerID)); err != nil {
+			seclog.Errorf("couldn't unlink workload %s (selector: %s) with profile %s: %v", workload.ContainerID, workload.Selector.String(), profile.Metadata.Name, err)
+		}
+		seclog.Infof("workload %s (selector: %s) successfully unlinked from profile %s", workload.ContainerID, workload.Selector.String(), profile.Metadata.Name)
+	} else if workload.CGroupContext.CGroupID != "" {
+		if err := m.securityProfileMap.Delete([]byte(workload.CGroupContext.CGroupID)); err != nil {
+			seclog.Errorf("couldn't unlink cgroup %s (selector: %s) with profile %s: %v", workload.CGroupContext.CGroupID, workload.Selector.String(), profile.Metadata.Name, err)
+		}
+		seclog.Infof("cgroup %s (selector: %s) successfully unlinked from profile %s", workload.CGroupContext.CGroupID, workload.Selector.String(), profile.Metadata.Name)
```
spikat: There is multiple part where we are manually checking in which case we are. Would we instead for code clarity add some helpers? Here for example we could replace
`m.securityProfileMap.Delete([]byte(workload.ContainerID));
`by
`m.securityProfileMap.Delete([]byte(workload.GetSelectorKey()));` (something like that)
Maybe this way we could simplify other parts of the code, WDYT?

spikat:
In addition to my first comments, this should unlock profile based feature like anomalies and auto-suppression right?
It would be great if you could add functional tests for all the "basic" profile features here (yeah I know .. just thinking about it's already too tedious xD)

spikat:
Another question: I didn't yet took a deep dive into it but: how are we handling "warm-up period" for cgroup profiles? The same way as we are for container's ones?

pkg/security/resolvers/tags/resolver_linux.go
```diff
@@ -119,15 +120,49 @@ func (t *LinuxResolver) checkTags(pendingWorkload *Workload) {
...
+	var newTags []string
+	var err error
+
+	if workload.ContainerID != "" {
+		// Container-based workload
+		newTags, err = t.ResolveWithErr(workload.ContainerID)
+		if err != nil {
+			return fmt.Errorf("failed to resolve container %s: %w", workload.ContainerID, err)
+		}
+	} else if workload.CGroupID != "" {
```
lebauce: Should we also check that CGroupFlags is systemd ?

pkg/security/security_profile/secprofs.go
```diff
@@ -55,37 +56,60 @@ func (m *Manager) LookupEventInProfiles(event *model.Event) {
...
+			}
+		}
 	}
-	selector, err := cgroupModel.NewWorkloadSelector(utils.GetTagValue("image_name", event.ContainerContext.Tags), "*")
-	if err != nil {
-		return
+
+	// If no profile found and there's a cgroup ID, try cgroup-based lookup
+	event.FieldHandlers.ResolveCGroupID(event, event.CGroupContext)
+	event.FieldHandlers.ResolveCGroupManager(event, event.CGroupContext)
```
lebauce: Is this required ?

pkg/security/security_profile/secprofs.go
```diff
@@ -55,37 +56,60 @@ func (m *Manager) LookupEventInProfiles(event *model.Event) {
...
+	event.FieldHandlers.ResolveCGroupVersion(event, event.CGroupContext)
+	if profile == nil && event.CGroupContext.CGroupID != "" {
+		// For cgroups, we look for profiles using the service name derived from the cgroup ID
+		// This is a simplified approach - in production you may need more sophisticated mapping
+		serviceName := string(event.CGroupContext.CGroupID)
+		// Extract service name from cgroup ID
+		if parts := strings.Split(serviceName, "/"); len(parts) > 0 {
```
lebauce: This could be a function `Service` on a CGroupContext

pkg/security/probe/probe_ebpf.go
```diff
@@ -792,8 +792,25 @@ func (p *EBPFProbe) DispatchEvent(event *model.Event, notifyConsumers bool) {
 
 	// handle anomaly detections
 	if event.IsAnomalyDetectionEvent() {
-		imageTag := utils.GetTagValue("image_tag", event.ContainerContext.Tags)
-		p.profileManager.FillProfileContextFromContainerID(event.FieldHandlers.ResolveContainerID(event, event.ContainerContext), &event.SecurityProfileContext, imageTag)
+		var workloadID interface{}
```
spikat: Can we use a string instead of an interface here? Or maybe even create a dedicated type/alias for "workloadID" ?

pkg/security/resolvers/tags/resolver_linux.go
```diff
@@ -29,6 +29,26 @@ type Workload struct {
 	retries  int
 }
 
+// GetWorkloadID returns the workload ID for a workload
+func (w *Workload) GetWorkloadID() interface{} {
```
spikat: same here for the interface

pkg/security/secl/model/model_unix.go
```diff
@@ -1017,3 +1017,14 @@ type SetSockOptEvent struct {
...
+// GetWorkloadID returns an ID that represents the workload
+func (pc *ProcessCacheEntry) GetWorkloadID() interface{} {
+	if pc.ContainerID != "" {
+		return pc.ContainerID
+	}
+	if pc.CGroup.CGroupID != "" {
+		return pc.CGroup.CGroupID
+	}
+	return nil
+}
```
spikat: this is almost the same func as in the tags resolvers. could we have a common type here ?

pkg/security/security_profile/profile/profile.go
```diff
@@ -646,7 +646,8 @@ func (p *Profile) MatchesSelector(entry *model.ProcessCacheEntry) bool {
 	defer p.InstancesLock.Unlock()
 
 	for _, workload := range p.Instances {
-		if entry.ContainerID == workload.ContainerID {
+		// Check if the workload IDs match
+		if entry.GetWorkloadID() != nil && workload.GetWorkloadID() != nil && entry.GetWorkloadID() == workload.GetWorkloadID() {
```
spikat: ~same here, if we add a workloadID type we could add a "match" func to it :)

pkg/security/tests/cmdwrapper.go
```diff
@@ -11,22 +11,25 @@ package tests
...
-	noWrapperType     wrapperType = "" //nolint:deadcode,unused
-	stdWrapperType    wrapperType = "std"
-	dockerWrapperType wrapperType = "docker"
-	podmanWrapperType wrapperType = "podman"
-	multiWrapperType  wrapperType = "multi"
+	noWrapperType      wrapperType = "" //nolint:deadcode,unused
+	stdWrapperType     wrapperType = "std"
+	dockerWrapperType  wrapperType = "docker"
+	podmanWrapperType  wrapperType = "podman"
+	systemdWrapperType wrapperType = "systemd"
```
spikat, line 31: that's awesome :) thanks for adding it

pkg/security/tests/security_profile_test.go
```diff
@@ -2654,3 +2654,403 @@ func TestSecurityProfileSyscallDriftNoNewSyscall(t *testing.T) {
...
+			t.Fatal(err)
+		}
+		time.Sleep(6 * time.Second) // a quick sleep to let the profile be loaded (5sec debounce + 1sec spare)
+
+		// Wait for the stable period to pass
+		time.Sleep(11 * time.Second)
+
+		err = test.GetCustomEventSent(t, func() error {
+			// Execute the new reload command, it should trigger an anomaly
+			err := serviceInstance.reload()
```
spikat, line 3009: Isn't it the same cmd (sleep 1) which should not generate an anomaly here?

mftoure: in this test, the command is not known to the profile and the profile is stable, so it should generate an anomaly

pkg/security/security_profile/secprofs.go
```diff
@@ -55,42 +55,64 @@ func (m *Manager) LookupEventInProfiles(event *model.Event) {
...
-	if imageTag == "" {
-		imageTag = "latest" // not sure about this one
-	}
-
 	ctx, found := profile.GetVersionContext(imageTag)
 	if found {
 		ctx.LastSeenNano = uint64(m.resolvers.TimeResolver.ComputeMonotonicTimestamp(time.Now()))
 	} else {
-		evictedVersions := profile.PrepareNewVersion(imageTag, event.ContainerContext.Tags, m.config.RuntimeSecurity.SecurityProfileMaxImageTags, uint64(m.resolvers.TimeResolver.ComputeMonotonicTimestamp(time.Now())))
+		fmt.Printf(" %s %v %v\n", imageTag, tags, m.config.RuntimeSecurity.SecurityProfileMaxImageTags)
```
spikat: debug printf to be removed :)

spikat (approved):
Only a debug printf to remove before merging, otherwise LGTM!

PR #37558, opened 2025-06-02
Add support for secrets when loading additional_endpoints from env vars

hush-hush:
### What does this PR do?

Add support for secrets when loading `additional_endpoints` from env vars as JSON payload.

Overall we're moving the env vars support within the config from each module. The code should not have custom cases depending on the source of a setting, once loaded they any source should produce the same types in the configuration. This also allow us to apply post-processing logic in all cases (like secrets resolution).

Set additional endpoints as through the env vars as JSON payload.

Env vars to test per team:

agent-metrics-logs:
* `logs_config.additional_endpoints`

agent-apm:
* `apm_config.profiling_additional_endpoints`
* `apm_config.additional_endpoints`
* `apm_config.debugger_additional_endpoints`
* `debugger_diagnostics_additional_endpoints`
* `apm_config.symdb_additional_endpoints`
* `apm_config.telemetry.additional_endpoints`
* `ol_proxy_config.additional_endpoints`
* `evp_proxy_config.additional_endpoints`

container-app:
* `orchestrator_explorer.orchestrator_additional_endpoints`
* `process_config.orchestrator_additional_endpoints`

process-agent:
* `process_config.additional_endpoints`
* `process_config.events_additional_endpoints`

OTEL Agent team should validate that their special cases for `logs_config.additional_endpoints` still works as expected. The change from `interface{}` to `string` as the key for the map should not break anything since we're serializing a struct that can be loaded with `mapstructure` (ie: map[string]interface{}).


pkg/util/common/common.go
```diff
@@ -119,13 +119,13 @@ func GetSliceOfStringMap(slice []interface{}) ([]map[string]string, error) {
...
 
 	for _, e := range slice {
-		value, ok := e.(map[interface{}]interface{})
+		value, ok := e.(map[string]interface{})
 		if !ok {
 			return nil, fmt.Errorf("unexpected type for slice value")
 		}
 		entry := map[string]string{}
 		for k, v := range value {
-			entry[fmt.Sprintf("%v", k)] = fmt.Sprintf("%v", v)
```
pgimalac, line 128: Are we sure this change won't break other users of this function ?

hush-hush: Yes, we're serializing a struct that can be loaded with mapstructure by the logs-agent. While we should remove this bit of logic as fast as possible the change should not break anything.

pgimalac: I checked and the function is used in a single place 👍

pgimalac (approved):

StephenWakely (approved):

releasenotes/notes/support-for-secrets-additional-endpoints-from-env-var-85c903770a231df4.yaml
```diff
@@ -0,0 +1,8 @@
+---
+features:
+  - |
+    Secrets are now supported in nested JSON string from environment variables when configuring additional_endpoints.
```
git-thuerk-done, line 4:
```suggestion
    Secrets are now supported in nested JSON string from environment variables when configuring additional endpoints.
```

git-thuerk-done (commented):
hi @hush-hush just a minor typo but approved

rahulkaukuntla (approved):

pkg/config/setup/config.go
```diff
@@ -2521,6 +2543,14 @@ func bindEnvAndSetLogsConfigKeys(config pkgconfigmodel.Setup, prefix string) {
 	config.BindEnv(prefix + "logs_dd_url") // Send the logs to a proxy. Must respect format '<HOST>:<PORT>' and '<PORT>' to be an integer
 	config.BindEnv(prefix + "dd_url")
 	config.BindEnv(prefix + "additional_endpoints")
+	config.ParseEnvAsSlice(prefix+"additional_endpoints", func(in string) []interface{} {
+		var mappings []interface{}
+		if err := json.Unmarshal([]byte(in), &mappings); err != nil {
+			log.Errorf(`"%s" can not be parsed: %v`, prefix+"additional_endpoints", err)
+		}
+		return mappings
+	})
```
GustavoCaso, line 2585: nit + personal opinion + no-blocker: should we make it into a helper function, same as `enableJSONParsingFromEnv`? I know. It is used only in one place, but it would make the file more homogeneous

pkg/config/setup/apm.go
```diff
@@ -148,7 +148,9 @@ func setupAPM(config pkgconfigmodel.Setup) {
 	config.BindEnv("apm_config.max_sender_retries", "DD_APM_MAX_SENDER_RETRIES")
 	config.BindEnv("apm_config.profiling_dd_url", "DD_APM_PROFILING_DD_URL")
 	config.BindEnv("apm_config.profiling_additional_endpoints", "DD_APM_PROFILING_ADDITIONAL_ENDPOINTS")
+	enableJSONParsingFromEnv(config, "apm_config.profiling_additional_endpoints")
 	config.BindEnv("apm_config.additional_endpoints", "DD_APM_ADDITIONAL_ENDPOINTS")
+	enableJSONParsingFromEnv(config, "apm_config.additional_endpoints")
 	config.BindEnv("apm_config.replace_tags", "DD_APM_REPLACE_TAGS")
```
GustavoCaso, line 154: question: Wondering if it would make more sense to have a wrapper around `BindEnv` to have an options argument to indicate if we should enable JSON parsing per key? 

```
config.BindEnv("apm_config.additional_endpoints", "DD_APM_ADDITIONAL_ENDPOINTS", nested_json_map)
```

jeremy-hanna (approved):

robertjli (approved):

pkg/config/setup/config.go
hush-hush: This is very specific to the way logs product configure their additional_endpoint. They parse is as a slice to then be able to load it with `mapstrucutre`. I'm not sure we want to encourage this double parsing in more place in the Agent.

pkg/config/setup/apm.go
hush-hush: Yes, that what where we want to take the API, but that would be for another PR entirely (the end goal is to have a single method to declare a settings with options).

PR #37568, opened 2025-06-02
CXP-2266 [agent] enrich process collector with container data

gengnamstyle:
### What does this PR do?
- adds a `owner` field to the workloadmeta process entity to reference the container this process is in
    - allows future queries that ask wlm for process data to also include the container data at the same time (optimization) and prevent duplicate data storage by storing container data on the process entity
    - similar pattern exists for workloadmeta [Container struct ](https://github.com/DataDog/datadog-agent/blob/4270f621c687c35ff6fba6413bd0e8928f265527/comp/core/workloadmeta/def/types.go#L559)

### Motivation
- processes are already enriched with container data on the check and language collector. We can reduce operations on the check and aid with the future language collector consolidation if we enrich the new process collector with container data

- unit tests

### DIFF
--- a/comp/core/workloadmeta/collectors/internal/process/process_collector.go
+++ b/comp/core/workloadmeta/collectors/internal/process/process_collector.go
@@ -126,7 +126,7 @@ func createdProcessesToWorkloadmetaProcesses(createdProcs []*procutil.Process, p
 	return wlmProcs
 }
 
-// deletedProcessesToWorkloadMetaProcesses helper function to convert deletedProcs into wlm entities
+// deletedProcessesToWorkloadMetaProcesses helper function to convert deletedProcs into wlm entities. wlm only uses the EventType, Source, ID, and Kind for deletion events
 func deletedProcessesToWorkloadmetaProcesses(deletedProcs []*procutil.Process) []*workloadmeta.Process {
 	wlmProcs := make([]*workloadmeta.Process, len(deletedProcs))
 	for i, proc := range deletedProcs {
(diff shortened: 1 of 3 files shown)


wdhif (approved):

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -97,6 +108,24 @@ func (c *collector) Start(ctx context.Context, store workloadmeta.Component) err
...
+		cid, exists := pidToCid[int(proc.Pid)]
+		if exists {
+			wlmProcs[i].ContainerID = cid // existing behaviour which we will maintain until the collector is enabled
+			// storing container id as an entity field
+			wlmProcs[i].Owner = &workloadmeta.EntityID{
+				Kind: workloadmeta.KindContainer,
+				ID:   cid,
+			}
+		}
+	}
```
daniel-taf, line 125: Do we need to be doing this for dead processes? What happens in that situation?

gengnamstyle: I looked at the implementation of how the wlm events are handled, and for deleted entities, we never need more data other than the `Source`, `ID`, and `Kind`. So I removed this to reflect that for deleted processes. 👍

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -137,18 +180,18 @@ func (c *collector) collect(ctx context.Context, collectionTicker *clock.Ticker)
...
-			}
+
+			wlmCreatedProcs := createdProcessesToWorkloadmetaProcesses(createdProcs, pidToCid)
 
 			deletedProcs := processCacheDifference(c.lastCollectedProcesses, procs)
-			wlmDeletedProcs := make([]*workloadmeta.Process, len(deletedProcs))
-			for i, proc := range deletedProcs {
-				wlmDeletedProcs[i] = processToWorkloadMetaProcess(proc)
-			}
+			// wlm only uses the EventType, Source, ID, and Kind for deletion events
```
daniel-taf: nit: I think this comment should be in the `deletedProcessesToWorkloadmetaProcesses` function, not here.

daniel-taf (approved):

comp/core/workloadmeta/collectors/internal/process/process_collector.go
gengnamstyle: updated thanks.

PR #37577, opened 2025-06-02
[DEBUG-3985] dyninst: run system-probe tests

ajwerner:
This tree is now part of the system-probe development process and will eventually be a part of the system probe. We want kmt and the likes to run against it.

### What does this PR do?

This PR makes `pkg/dyninst/...` changes trigger the system-probe tests.

### Motivation

This is where the new implementation of dynamicinstrumentation is going. It will eventually replace that package.

Relates to https://datadoghq.atlassian.net/browse/DEBUG-3883


chouetz (approved):

ajwerner:
Thanks for the review!

PR #37581, opened 2025-06-02
[clusteragent/autoscaling] Allow configuring local fallback activation threshold

jennchenn:
### What does this PR do?

Allow configuring local fallback activation threshold by setting the following in the DPA:
```
spec:
  owner: Local
  ...
  fallback:
    horizontal:
      enabled: true
      triggers:
        staleRecommendationThresholdSeconds: 600
```

Also change the default threshold to activate local fallback to `30 minutes`

### Motivation

Previously, `staleRecommendationThresholdSeconds` controlled when a metric would be considered stale when calculating the local fallback recommendations, even though the documentation had it intended to control when local fallback would be activated. 

1. Modify `staleRecommendationThresholdSeconds`
2. Verify that local fallback activation occurs following the newly configured threshold value

We no longer give users the ability to control which metrics are considered stale for the purposes of local fallback recommendations (by default we use a timeframe of `60s`).

### DIFF
--- a/pkg/clusteragent/autoscaling/workload/controller.go
+++ b/pkg/clusteragent/autoscaling/workload/controller.go
@@ -42,7 +42,7 @@ const (
 
 	controllerID = "dpa-c"
 
-	staleTimestampThreshold = time.Minute * 10 // time to wait before considering a recommendation stale
+	defaultStaleTimestampThreshold = time.Minute * 10 // time to wait before considering a recommendation stale
 )
 
 var (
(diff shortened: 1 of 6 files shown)


pkg/clusteragent/autoscaling/workload/controller.go
```diff
@@ -517,19 +517,25 @@ func getActiveScalingSources(currentTime time.Time, podAutoscalerInternal *model
 	fallbackHorizontalScalingValues := podAutoscalerInternal.FallbackScalingValues().Horizontal
 
 	// If main scaling values are not stale, use those
-	if mainHorizontalScalingValues != nil && !isTimestampStale(currentTime, mainHorizontalScalingValues.Timestamp) {
+	if mainHorizontalScalingValues != nil && !isTimestampStale(currentTime, mainHorizontalScalingValues.Timestamp, nil) {
```
vboulineau: Shouldn't that check the staleness compared with the setting? If I want a local fallback after 5mins, this will return true while it should not

jennchenn: updated!

vboulineau (approved):

PR #37594, opened 2025-06-03
[AGENTRUN-310] Fix FIPS Agent status output

misteriaud:
### What does this PR do?
This PR prevent from printing the FIPS Proxy status section when the `status` command is executed with the FIPS Agent.
The FIPS Proxy status should be printed only when:
- `fips.enabled` config is set to true
- the Agent is not FIPS Agent 

### Motivation
This PR fix a regression introduced in #36988


rahulkaukuntla (approved):

releasenotes/notes/fix_fips_agent_status-ad9cf20c53bdec9e.yaml
```diff
@@ -0,0 +1,11 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+fixes:
+  - |
+     Removed the FIPS Proxy status section from the agent status page when running the FIPS Agent.
```
buraizu:
```suggestion
     Remove the FIPS Proxy status section from the Agent status page when running the FIPS Agent. 
```

buraizu (commented):
Left some minor feedback for style and consistency with other release notes

releasenotes/notes/fix_fips_agent_status-ad9cf20c53bdec9e.yaml
sgnn7: @buraizu QQ: Since these are put into the release notes, would it make more sense for them to be past-tense?

buraizu (approved):

releasenotes/notes/fix_fips_agent_status-ad9cf20c53bdec9e.yaml
buraizu: @sgnn7 good question, thanks for following up. Release notes are generally kept in present tense to clearly indicate that the change is available in the current version. This is the pattern followed by most of our other release notes as well.

sgnn7: May be good to communicate this wider in the Agent group as we're a bit inconsistent on this https://github.com/DataDog/datadog-agent/releases 😅

PR #37599, opened 2025-06-03
chore(fleet): Report APM SSI status in the inventoryagent payload

BaptisteFoy:
### What does this PR do?
Reports Linux host SSI in the agent's metadata payload. Windows will be done in another PR.

This is a revert of a revert in #37496; the main modification from the original PR (https://github.com/DataDog/datadog-agent/pull/36875) is that we don't call the installer binary anymore.

### Motivation
[INPLAT-568]

Manual QA on a VM -- checked the result using the metadata payload in debug logs

### Additional Notes

[INPLAT-568]: https://datadoghq.atlassian.net/browse/INPLAT-568?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


comp/updater/ssistatus/impl/ssistatus_unix.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/DataDog/datadog-agent/pkg/fleet/installer/packages/ssi"
+)
+
+// autoInstrumentationStatus checks if the APM auto-instrumentation is enabled on the host. This will return false on Kubernetes
+func (c *ssiStatusComponent) autoInstrumentationStatus() (bool, []string, error) {
+	// Ideally we would call the installer package to check if the injector is installed (using the status), but doing so
+	// significantly increases the memory footprint of the agent, so we check the filesystem directly.
+
+	// TODO: fix this
```
coignetp, line 22: **🥜 nitpick:** ‏expected?

coignetp (approved):

comp/updater/ssistatus/impl/ssistatus_unix.go
BaptisteFoy: Yes, the plan is to find a solution to do this cleanly and without bumping memory too much in another PR

dustmop (approved):

comp/updater/ssistatus/fx/fx.go
```diff
@@ -0,0 +1,34 @@
...
+		),
+		fxutil.ProvideOptional[ssistatus.Component](),
+
+		// ssistatus is a component with no public method, therefore nobody depends on it and FX only instantiates
+		// components when they're needed. Adding a dummy function that takes our Component as a parameter force
+		// the instantiation of ssistatus. This means that simply using 'ssistatusfx.Module()' will run our
+		// component (which is the expected behavior).
+		//
+		// This prevent silent corner case where including 'ssistatus' in the main function would not actually
+		// instantiate it. This also remove the need for every main using ssistatus to add the line bellow.
```
pgimalac:
```suggestion
		// ssistatus is a component with no public method, therefore nobody depends on it and FX only instantiates
		// components when they're needed. Adding a dummy function that takes our Component as a parameter forces
		// the instantiation of ssistatus. This means that simply using 'ssistatusfx.Module()' will run our
		// component (which is the expected behavior).
		//
		// This prevents silent corner case where including 'ssistatus' in the main function would not actually
		// instantiate it. This also removes the need for every main using ssistatus to add the line below.
```

pgimalac (approved):
LGTM for Agent Runtimes

PR #37608, opened 2025-06-03
[CWS] Add execution contexts tests

Gui774ume:
### What does this PR do?

This PR adds tests on execution contexts.

### Motivation

This is powered by variables (and probably somehow covered by other unit tests), but this test puts all the necessary features together into one chain of rules. We want to guarantee this specific use case is covered in our test suit.

### DIFF
--- a/pkg/security/secl/rules/policy_test.go
+++ b/pkg/security/secl/rules/policy_test.go
@@ -14,6 +14,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 	"syscall"
 	"testing"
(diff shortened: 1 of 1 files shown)


pkg/security/secl/rules/policy_test.go
```diff
@@ -942,6 +943,230 @@ func TestActionSetVariableInherited(t *testing.T) {
...
+
+func TestActionSetVariableInheritedFilter(t *testing.T) {
+	testPolicy := &PolicyDef{
+		Rules: []*RuleDefinition{
+			{
+				ID:         "first_execution_context",
+				Expression: `open.file.path == "/tmp/first"`,
+				Actions: []*ActionDefinition{
+					{
+						Filter: stringPtr(`${process.correlation_key} == ""`),
```
spikat: I'm sure it's intended but, why not put the `${process.correlation_key} == ""` directly on the expression?

Gui774ume: Just to follow the same pattern as the other rules ... but you're right, for this rule we could have put it directly on the expression !

Gui774ume: We could probably do something similar for the other rules actually ...

Gui774ume: (as in put the larger case between the two as part of the rule, and only do a final `!= ""` for the `parent_correlation_keys` variable change) ... on it !

spikat: Ahah ok :smile:  :+1:

spikat (approved):

PR #37622, opened 2025-06-03
[AGENTRUN-349] remove unused JS functions and HTTP handler from the agent GUI

GustavoCaso:
### What does this PR do?

While looking for ways to help the Data stream handle checks reload, I stumbled upon some old code for the GUI that reloads checks that is no longer used, or at least seems that way. 

This PR removes the code for reading and running checks from the GUI, as it no longer seems to be used in the GUI interface 

1. Build the agent 
2. Start the Agent 
3. Launch the GUI `./bin/agent/agent launch-gui --cfgpath ./bin/agent/dist`
4. Validate the Check view works and there is not JS errors in the console panel


julien-lebot (approved):

comp/core/gui/guiimpl/checks.go
```diff
@@ -56,11 +49,8 @@ func getFleetPoliciesPath() string {
 }
 
 // Adds the specific handlers for /checks/ endpoints
-func checkHandler(r *mux.Router, collector collector.Component, ac autodiscovery.Component) {
+func checkHandler(r *mux.Router) {
 	r.HandleFunc("/running", http.HandlerFunc(sendRunningChecks)).Methods("POST")
-	r.HandleFunc("/run/{name}", http.HandlerFunc(runCheckHandler(collector, ac))).Methods("POST")
```
dustmop, line 61: Nothing else uses these endpoints? They were only being called by the GUI's js code?

dustmop (approved):
LGTM, just one small question I wanted to check on

comp/core/gui/guiimpl/checks.go
GustavoCaso: Correct. The GUI has its server, and these handlers were only used with the JS functions I removed 

https://github.com/DataDog/datadog-agent/pull/37622/files#diff-2691a73750ec49784b440e9a5ea54064d02f5e0718d68a6e3a3d66bde03b0073L642

PR #37624, opened 2025-06-03
[CXP-2282] Add stubs for new endpoints in system probe process module

daniel-taf:
### What does this PR do?

- Adds stubs for new endpoints that will be used by the process collector and process/discovery checks that just return an empty OK response. Actual functionality will be added later.
- Some minor refactors so the code is cleaner with multiple endpoints.
- Add unit tests for the stats requests

### Additional Notes

We don't really need this but might make some testing easier down the line.

### DIFF
--- a/cmd/system-probe/modules/process_test.go
+++ b/cmd/system-probe/modules/process_test.go
@@ -18,7 +18,6 @@ import (
 	model "github.com/DataDog/agent-payload/v5/process"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"google.golang.org/protobuf/proto"
 
 	"github.com/DataDog/datadog-agent/pkg/process/encoding"
 	reqEncoding "github.com/DataDog/datadog-agent/pkg/process/encoding/request"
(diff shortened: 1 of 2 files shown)


gengnamstyle (approved):

cmd/system-probe/modules/process_test.go
```diff
@@ -0,0 +1,85 @@
...
+	"google.golang.org/protobuf/proto"
+
+	"github.com/DataDog/datadog-agent/pkg/process/encoding"
+	reqEncoding "github.com/DataDog/datadog-agent/pkg/process/encoding/request"
+	"github.com/DataDog/datadog-agent/pkg/process/procutil"
+	"github.com/DataDog/datadog-agent/pkg/process/procutil/mocks"
+	processProto "github.com/DataDog/datadog-agent/pkg/proto/pbgo/process"
+)
+
+func TestStatsHandler(t *testing.T) {
```
gengnamstyle, line 29: nit: do you think this is worth formatting as a table driven test? The function we're testing seems pretty simple so up to you.

daniel-taf: Will keep as is since there aren't other cases

PR #37629, opened 2025-06-03
[usm] fix the script generating the gotls/lookup/luts.go file

andreimatei:
Before this patch, the generation of the `gotls/lookup/luts.go` file was
failing: Go versions below 1.16 cannot parse the go.mod file of the
x/net module at versions above 0.35 (released a few months ago). This
patch pins the module at v0.35. We were arbitrarily using the latest
x/net version (and thus using the debug info for the latest); now we're
arbitrarily using a fixed version. Luckily, the debug info we need for
x/net is stable -- we need the offset of an embedded field in a struct.

The regenerated `luts.go` doesn't have any changes. FWIW, I think this shows
that the debug info in that file is valid for Go 1.24; I'm not sure we
ever ran the script for 1.24, since I think the release of x/net that
broke it happened a little before 1.24.

### DIFF
--- a/pkg/network/go/lutgen/run.go
+++ b/pkg/network/go/lutgen/run.go
@@ -33,8 +33,11 @@ import (
 // Each lookup table is implemented as a function,
 // which are configured in LookupFunction types.
 type LookupTableGenerator struct {
-	Package                string
-	MinGoVersion           goversion.GoVersion
+	Package      string
+	MinGoVersion goversion.GoVersion
+	// The maximum Go version to include in the lookup table. If empty, the
+	// latest released version (including RC versions) will be used.
+	MaxGoVersion           goversion.GoVersion
 	Architectures          []string
 	CompilationParallelism int
 	LookupFunctions        []LookupFunction
@@ -158,7 +161,8 @@ func (g *LookupTableGenerator) getVersions(ctx context.Context) ([]goversion.GoV
 	for _, version := range allVersions {
 		if version.Rev != 0 ||
 			version.Proposal != "" ||
-			!version.AfterOrEqual(g.MinGoVersion) {
+			!version.AfterOrEqual(g.MinGoVersion) ||
+			(g.MaxGoVersion != (goversion.GoVersion{}) && !g.MaxGoVersion.AfterOrEqual(version)) {
 			continue
 		}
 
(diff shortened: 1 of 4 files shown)


guyarb:
> I'm not sure we
ever ran the script for 1.24, since I think the release of x/net that
broke it happened a little before 1.24.

We did run it. Go 1.24 was released on February 11th, while `x/net` v0.36 (the version that broke the support) was released only on March 4th

pkg/network/go/lutgen/run.go
```diff
@@ -324,12 +328,25 @@ func setupGoModule(ctx context.Context, cmd *exec.Cmd, programPath string) error
...
+	// format. Newer versions of the package have a go.mod file that can't be
+	// parsed by Go <= 1.16.
+	getCmd := exec.CommandContext(ctx, "go", "get", "golang.org/x/net@v0.35.0")
+	getCmd.Env = cmd.Env
+	getCmd.Dir = cmd.Dir
+	getCmd.Path = cmd.Path
+	output, err := getCmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("error executing 'go get': %s\n%s", err, output)
+	}
```
guyarb: That's a problematic change. You assume that `v0.35` will represent all future versions, while we can still have changes.

If the issue happens only with go1.16 and below, then the fix should take into account and pin the version only when we run the script for go1.16 and below

guyarb (changes requested):
reviewed

pkg/network/go/lutgen/run.go
andreimatei: I've considered doing that, but this all leads to a deeper question -- which version of the library do we care about? Why would would we particularly care about the latest (and only the latest)? Then, caring about different library versions for different compiler versions seems confusing -- although you can argue that it makes some sense since new versions of the lib don't work with old compilers... I'm in the live debugger team, and so the real answer is perhaps that we should actually look at the particular binary and dynamically analyze its debug info :).

But also you can argue that dealing with the debug info for the `x/net` library at all is kinda silly -- the only thing we look at is the offset of an embedded field into a struct -- which seems likely to stay 0 for as long as that field exists. That's why I thought that the complexity of dealing with multiple versions is not really worth it. I guess using the latest version of the library tells us that the embedded field continues to exist, so there's some value in it...

Having written all this, if you think the complexity of changing the library version based on the compiler version is worth it, I'm happy to do it.

andreimatei: I've made the script use different versions of the library based on the Go version, as you suggested.

andreimatei:
A side-note -- the script does not work for Go 1.25rc, which switched to generating dwarf v5 and our library that processes location lists doesn't like that. We need to fix this for Live Debugger too. I've added a commit to the PR that lets you set a max go version so we can still run the script in the meantime.

PR #37633, opened 2025-06-04
Report Fleet Status in Agent Command

san-jos:
### What does this PR do?
Create a Fleet status module displaying fleet enablement status via the `datadog-agent status` command

### Motivation
https://datadoghq.atlassian.net/browse/FA-990

Manually validated


louis-cqrl (approved):

rahulkaukuntla (approved):

comp/daemonchecker/def/component.go
```diff
@@ -0,0 +1,14 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-present Datadog, Inc.
+
+// Package daemonchecker retrieves the running status of the installer daemon
+package daemonchecker
```
BaptisteFoy: **🥜 Nitpick:** ‏it's a bit weird to me to have this component at the top-level. Can we either move it under `comp/updater` or rename it?

comp/fleetstatus/impl/status_templates/fleetstatus.tmpl
```diff
@@ -0,0 +1,16 @@
+{{- if .HTML }}
+<div class="stat">
+  <span class="stat_title">Fleet Status</span>
+  <span class="stat_data">
+{{- end }}
+{{ with .fleetAutomationStatus }}
+{{ if .fleetAutomationEnabled }}Fleet Management is enabled{{ else }}Fleet Management is disabled{{ end }}
+
+Remote Management Status:    {{ if .remoteManagementEnabled }}Enabled{{ else }}Disabled{{ end }}
+Remote Configuration Status: {{ if .remoteConfigEnabled }}Enabled{{ else }}Disabled{{ end }}
```
BaptisteFoy: **❓ Question:** Do we need this when we have the Remote Config section? IIRC if there's no RC enabled the daemon shuts down‏

BaptisteFoy (approved):

comp/fleetstatus/impl/status_templates/fleetstatus.tmpl
san-jos: I thought this might help troubleshoot the issue: RC is a prerequesite. It might be the reason why remote management is not active.

BaptisteFoy: It is indeed, but the Remote Config section isn't that far and the status command output is already big so I'd rather avoid duplicate data.
I'll let other agent folks chime in if needed.

san-jos: Removed

PR #37635, opened 2025-06-04
winbuildscript: don't install dda if it's already available

chouquette:
### What does this PR do?

Don't force `dda` install in our windows build scripts

### Motivation

This ensures we won't use an unexpected version of dda in our CI before we bump the build images


chouquette:
closing since the initial assumption that dda was already available in the build image isn't verified

PR #37639, opened 2025-06-04
chore: upgrade go-libddwaf to the latest v3

RomainMuller:
### What does this PR do?

Upgrades `github.com/DataDog/go-libddwaf/v3` to the latest, which supports `go1.24`, and ensure some test fails if building the serverless agent with an unsupported Go release.

It also changes the startup mechanism to no longer fail if AppSec fails to start, instead logs a warning and uses a no-op `httpsec.Monitorer`.

### Motivation

Customers rely on this feature.


RomainMuller:
* **#37639** <a href="https://app.graphite.dev/github/pr/DataDog/datadog-agent/37639?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/datadog-agent/37639?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

eliottness (approved):

pkg/serverless/appsec/appsec.go
```diff
@@ -40,9 +40,10 @@ func New(demux aggregator.Demultiplexer) (lp *httpsec.ProxyLifecycleProcessor, e
 func NewWithShutdown(demux aggregator.Demultiplexer) (lp *httpsec.ProxyLifecycleProcessor, shutdown func(context.Context) error, err error) {
 	appsecInstance, err := newAppSec() // note that the assigned variable is in the parent scope
 	if err != nil {
-		return nil, nil, err
-	}
-	if appsecInstance == nil {
+		log.Warnf("appsec: failed to start appsec: %v -- appsec features are not available!", err)
+		// Nil-out the error so we don't incorrectly report it later on...
+		err = nil
```
RamyElkest, line 45: why don't we return here? what's the value of the proxy without appsec

RomainMuller: If the proxy does not run, the Lambda cannot work (if AppSec is configured in such a way that we even attempt starting the proxy; the runtime _IS_ proxied).

RamyElkest: gotcha [configuration ref](https://github.com/DataDog/datadog-lambda-go/blob/a5836ece0a8aa3cc8a7c31759ff8ed82930cbc18/ddlambda.go#L368).

pkg/serverless/appsec/appsec_test.go
```diff
@@ -32,18 +34,8 @@ func TestNew(t *testing.T) {
 			if stop != nil {
 				defer stop(context.Background())
 			}
-			if err := wafHealth(); err != nil {
-				if ok, _ := waf.SupportsTarget(); ok {
-					// host should be supported by appsec, error is unexpected
-					require.NoError(t, err)
```
Julio-Guerra, line 38: @RomainMuller: this overall case to "skip" the test when the host platform is not supported sounds valid to me, and the problem rather is the use of `NoError` instead of `Error` in this line.

I think we need to restore this and just change the line to `require.Error(t, err)` when the current host is supported.

Julio-Guerra (changes requested):

pkg/serverless/appsec/appsec_test.go
RomainMuller: "Unsupported platform" includes "unsupported runtime" - we don't have a way to distinguish. And in this case the platform is guaranteed to always be supported because it's Lambda (so AmazonLinux variants)... the only "safe" way to guarantee app sec works is to not allow tests to pass if the waf fails to load at all.

Julio-Guerra: ok, so we need to skip unsupported GOOS/GOARCH because this is running in the datadog agent CI

Julio-Guerra: Updated in https://github.com/DataDog/datadog-agent/pull/37639/commits/deba3ad7e9d566b4d8d95ffed525bfde33e9cd00

Julio-Guerra (approved):

PR #37647, opened 2025-06-04
Build ddot-only packages instead of agent+ddot

alopezz:
### What does this PR do?

It adds the build of deb and rpm packages containing only the ddot. These packages will depend on the datadog-agent at the package-manager level.

### Motivation

[ABLD-56](https://datadoghq.atlassian.net/browse/ABLD-56) (includes a link to the RFC where this path was decided upon).

I performed manual validation of the deb and rpm packages, and implemented basic e2e testing that ensures the basic dependency behavior that we want between the new ddot package and the datadog-agent.

### Additional Notes

- This PR replaces existing jobs for deb packages that are no longer used. These were building a combination a "fat" package which combined the regular datadog-agent package with this additional binary. Since we've discarded that approach now, this should make sense.
- This PR also removes references to "ot" as an Agent flavor throughout as well as related logic, as treating it as such is no longer useful.
- The e2e tests I added are heavily based on the existing step-by-step tests, but focus exclusively on the most relevant part of what's implemented here: ensuring that the dependency relationship is being honored by package managers and both packages are properly installed.

[ABLD-56]: https://datadoghq.atlassian.net/browse/ABLD-56?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


omnibus/config/projects/otel-agent.rb
```diff
@@ -48,25 +48,32 @@
 
 build_iteration 1
 
-description ''
+description 'Datadog Distribution of OTel Collector'
```
liustanley:
```suggestion
description 'Datadog Distribution of OpenTelemetry Collector'
```

omnibus/config/projects/otel-agent.rb
```diff
@@ -76,5 +85,62 @@
...
-strip_build true
+strip_build do_build
 debug_path ".debug"  # the strip symbols will be in here
+
+# ------------------------------------
+# Packaging
+# ------------------------------------
+
+if debian_target?
+  runtime_dependency "datadog-agent (= 1:#{build_version}-1)"
```
truthbk: 🍰

omnibus/config/projects/otel-agent.rb
```diff
@@ -76,5 +85,62 @@
...
+  runtime_recommended_dependency 'datadog-signing-keys (>= 1:1.4.0)'
+elsif redhat_target?
+  safe_version = Omnibus::Packager::RPM::safe_version(build_version)
+  runtime_dependency "dummy-agent = 1:#{safe_version}-1"
```
truthbk: What's this `dummy-agent` dep about? Pretty sure this is just a draft PR thing, just want to make sure why this is being done this way as opposed to the `debian_target` above?

truthbk (commented):
Nice, just a couple of questions...

omnibus/config/projects/otel-agent.rb
alopezz: That's a copy-paste thing from an experimental setup. Will fix!

omnibus/config/software/datadog-otel-agent.rb
```diff
@@ -9,7 +9,12 @@
 
 name 'datadog-otel-agent'
 
-source path: '..'
+source path: '..',
```
truthbk, line 12: I think the software definition may currently be incomplete. I think we're missing something like:

```
dependency 'zstd'
```
Or something like that for cgo compression support? Crazy part is I was kind of expecting that for the `datadog-agent` as well, but I can't seem to find the dependency were I was expecting it (it's only referenced by the `rpm` software definition, but not sure how/if that gets pulled).

truthbk (commented):
A thought about `zstd`....

omnibus/config/projects/ddot.rb
```diff
@@ -0,0 +1,152 @@
+# Unless explicitly stated otherwise all files in this repository are licensed
```
alopezz, line 1: Notes for reviewers about this file:
- This is a rename from otel-agent.rb (for some reason GitHub doesn't pick that up even though my local git shows it as such).
- Most of the changes from that file are taken from similar purpose code out of agent.rb or other wise adapted to achieve close behavior to that where relevant.

omnibus/config/software/datadog-otel-agent.rb
truthbk: Resolving this conversation. `zstd` is statically linked into the binary, so no need for a explicit dependency, we're good. Verified on a build.

omnibus/config/projects/agent.rb
```diff
@@ -42,10 +42,6 @@
 end
 
 if windows_target?
-  if ot_target?
-    raise UnknownPlatform
-  end
```
liustanley, line 47: Is there a reason why this is being removed from `windows_target?` when we're only adding support for Linux in this PR?

alopezz: This is in the context of building the main Agent package. Since we're going with the idea of having a separate package, I remove the only job that built a joint package (a package which wasn't really used anywhre) and with it all the code that refers to including ddot in the same package with the Agent is unnecessary and I've just removed.

chouquette (approved):

swang392 (approved):
lgtm for files owned by agent onboarding

chouetz (approved):

louis-cqrl (approved):
LGTM for ARUN files

ekump (approved):

ajgajg1134 (approved):

wiyu (approved):

jeremy-hanna (approved):

PR #37653, opened 2025-06-04
[CWS] allow to set variable in cli eval

safchain:
### What does this PR do?

This will allow to write unit test for rules with variables.

### DIFF
--- a/pkg/security/clihelpers/eval.go
+++ b/pkg/security/clihelpers/eval.go
@@ -185,7 +185,7 @@ func variablesFromTestData(testData TestData) (map[string]eval.SECLVariable, err
 	for k, v := range testData.Variables {
 		switch v := v.(type) {
 		case string:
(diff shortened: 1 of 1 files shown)


pkg/security/clihelpers/eval.go
```diff
@@ -186,6 +174,68 @@ func eventDataFromJSON(file string) (eval.Event, error) {
...
+	variables := make(map[string]eval.SECLVariable)
+
+	// copy the embedded variables
+	for k, v := range model.SECLVariables {
+		variables[k] = v
+	}
+
+	for k, v := range testData.Variables {
+		switch v := v.(type) {
+		case string:
```
lebauce, line 187: Should we also support []string and []int (and use NewScopedStringArrayVariable and NewScopedIntArrayVariable) ?

safchain: @lebauce indeed. I just pushed a commit to handle them.

lebauce (approved):

PR #37656, opened 2025-06-04
Add support for TLS settings in serializerexporter

mackjmr:
### What does this PR do?
This PR adds support for tls settings in serializer exporter. Namely we want to be able to set this in dd exporter [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/0cdf0be2200ea5def029a02293d9b32b71ded932/exporter/datadogexporter/factory.go#L366) based on whether the dd exporter was configured with `tls::insecure_skip_verify`.

### DIFF
--- a/comp/otelcol/otlp/components/exporter/serializerexporter/serializer.go
+++ b/comp/otelcol/otlp/components/exporter/serializerexporter/serializer.go
@@ -40,6 +40,8 @@ func setupForwarder(config pkgconfigmodel.Config) {
 
 	config.Set("forwarder_num_workers", 1, pkgconfigmodel.SourceDefault)
 	config.Set("forwarder_stop_timeout", 2, pkgconfigmodel.SourceDefault)
+	config.Set("forwarder_http_protocol", "auto", pkgconfigmodel.SourceDefault)
+
 	// Forwarder retry settings
 	config.Set("forwarder_backoff_factor", 2, pkgconfigmodel.SourceDefault)
 	config.Set("forwarder_backoff_base", 2, pkgconfigmodel.SourceDefault)
@@ -108,6 +110,7 @@ func initSerializer(logger *zap.Logger, cfg *ExporterConfig, sourceProvider sour
 			setupSerializer(pkgconfig)
 			setupForwarder(pkgconfig)
 			pkgconfig.Set("logging_frequency", int64(500), pkgconfigmodel.SourceDefault)
+			pkgconfig.Set("skip_ssl_validation", cfg.ClientConfig.InsecureSkipVerify, pkgconfigmodel.SourceFile)
 			return pkgconfig
 		}),
 		fx.Provide(func(log *zap.Logger) (logdef.Component, error) {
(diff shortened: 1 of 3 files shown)


songy23 (approved):

comp/otelcol/otlp/components/exporter/serializerexporter/serializer.go
```diff
@@ -108,6 +110,7 @@ func initSerializer(logger *zap.Logger, cfg *ExporterConfig, sourceProvider sour
 			setupSerializer(pkgconfig)
 			setupForwarder(pkgconfig)
 			pkgconfig.Set("logging_frequency", int64(500), pkgconfigmodel.SourceDefault)
+			pkgconfig.Set("skip_ssl_validation", cfg.ClientConfig.InsecureSkipVerify, pkgconfigmodel.SourceFile)
```
truthbk, line 113: I'm actually a little concerned by the way in which we're using `fx` here 🤔 

If I remember correctly `fx` only implements a singleton for each type, and since we also instantiate a `config.Component` in the [application](https://github.com/DataDog/datadog-agent/blob/main/cmd/otel-agent/subcommands/run/command.go#L115-L117), this is a bit unorthodox. I think this probably works right now because here we are creating a new `fx` application instance - but this is also a potential can of worms that can lead to unexpected behavior because we're leveraging different `config` objects in different parts of the Collector.

I know this `serializerexporter` component is also used in the OSS collector and so we need to provision the config somehow, but I'm not sure I'm a fan of the current process if I'm understanding all of this correctly.

Also, to add, if the `datadog.yaml` defined `skip_ssl_validation` as `true`, it doesn't seem like that would be applied to DDOT when it probably should, since it's a top-level configuration element in the `datadog.yaml`.

truthbk (commented):
Retroactive review with some thoughts @mackjmr.

comp/otelcol/otlp/components/exporter/serializerexporter/serializer.go
mackjmr: @truthbk The `initSerializer` logic only gets called in the OSS collector path, see the logic [here](https://github.com/DataDog/datadog-agent/blob/db059c516b292b9a7f8ff29863096b21de44df1c/comp/otelcol/otlp/components/exporter/serializerexporter/factory.go#L183-L184). In the DDOT path, the serializer.MetricSerializer will not be nil, so we won't go in this path. In the DDOT path the serializer is a [Required input](https://github.com/DataDog/datadog-agent/blob/db059c516b292b9a7f8ff29863096b21de44df1c/comp/otelcol/collector/impl/collector.go#L72) which gets passed [here](https://github.com/DataDog/datadog-agent/blob/main/cmd/otel-agent/subcommands/run/command.go#L181-L185) so I think it uses the config passed [here](https://github.com/DataDog/datadog-agent/blob/main/cmd/otel-agent/subcommands/run/command.go#L115-L117).

truthbk: Makes sense.

PR #37658, opened 2025-06-04
Fix nodetreemodel not handling all map interface

hush-hush:
### What does this PR do?

YAML will only return map[string]interface{} and map[interface{}]interface{}. But through manual `Set`, other types of map can make their way in the config.

Running the CI with nodetreemodel is enough

### DIFF
--- a/pkg/config/structure/unmarshal_test.go
+++ b/pkg/config/structure/unmarshal_test.go
@@ -299,6 +299,81 @@ endpoints:
 	}
 }
 
+func TestUnmarshalAllMapString(t *testing.T) {
+	mockConfig := newEmptyMockConf(t)
+	mockConfig.SetKnown("test")
+
+	type testString struct {
+		A string
+		B string
+	}
+	checkString := func() {
+		obj := testString{}
+		err := unmarshalKeyReflection(mockConfig, "test", &obj)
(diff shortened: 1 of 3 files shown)


pkg/config/structure/unmarshal_test.go
```diff
@@ -299,6 +299,81 @@ endpoints:
...
+		assert.Equal(t, testInt{A: 1, B: 2}, objInt)
+	}
+	mockConfig.Set("test", map[string]int{"a": 1, "b": 2}, model.SourceAgentRuntime)
+	checkInt()
+
+	mockConfig.Set("test", map[interface{}]int{"a": 1, "b": 2}, model.SourceAgentRuntime)
+	checkInt()
+
+	mockConfig.Set("test", map[interface{}]interface{}{"a": 1, "b": 2}, model.SourceAgentRuntime)
+	checkInt()
```
dustmop, line 350: Could we get a test case for `map[int]string` to validate that it works correctly?

dustmop: Nevermind, seems like this is a bigger fix than this PR is doing.

dustmop (approved):
LGTM

PR #37661, opened 2025-06-04
[dyninst] Expand ebpf stack machine capabilities

piob-io:
### What does this PR do?
Add support for capturing strings and slices.
Return ebpf attachpoints, for multi-probe support.

### Motivation

### DIFF
--- a/pkg/dyninst/compiler/compile.go
+++ b/pkg/dyninst/compiler/compile.go
@@ -22,29 +22,25 @@ import (
 //go:generate $GOPATH/bin/include_headers pkg/dyninst/ebpf/event.c pkg/ebpf/bytecode/build/runtime/dyninst_event.c pkg/ebpf/c
 //go:generate $GOPATH/bin/integrity pkg/ebpf/bytecode/build/runtime/dyninst_event.c pkg/ebpf/bytecode/runtime/dyninst_event.go runtime
 
-func getCFlags(config *ddebpf.Config) []string {
-	cflags := []string{
-		"-g",
-		"-Wno-unused-variable",
-		"-Wno-unused-function",
-		"-DDYNINST_GENERATED_CODE",
-	}
-	if config.BPFDebug {
-		cflags = append(cflags, "-DDEBUG=1")
-	}
-	return cflags
+// CompiledBPF holds compiled object of the eBPF program and its metadata.
+type CompiledBPF struct {
+	Obj ebpfruntime.CompiledOutput
+
+	// Program to attach and list of pcs to attach at, along with the cookie that should be provided at that attach point.
+	ProgramName  string
+	Attachpoints []codegen.BPFAttachPoint
 }
 
 // CompileBPFProgram compiles the eBPF program.
-func CompileBPFProgram(program ir.Program, extraCodeSink io.Writer) (ebpfruntime.CompiledOutput, error) {
+func CompileBPFProgram(program ir.Program, extraCodeSink io.Writer) (CompiledBPF, error) {
 	generatedCode := bytes.NewBuffer(nil)
 	smProgram, err := sm.GenerateProgram(program)
 	if err != nil {
-		return nil, err
+		return CompiledBPF{}, err
 	}
-	err = codegen.GenerateCCode(smProgram, generatedCode)
+	attachpoints, err := codegen.GenerateCCode(smProgram, generatedCode)
 	if err != nil {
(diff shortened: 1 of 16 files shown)


pkg/dyninst/integration_test.go
```diff
@@ -181,17 +181,21 @@ func testDyninst(t *testing.T, sampleServicePath string) {
...
+		typ, ok := irp.Types[ir.TypeID(di.Type)]
+		if !ok {
+			t.Fatalf("unknown type: %d", di.Type)
+		}
+		pos += uint32(unsafe.Sizeof(*di))
+		t.Logf("di: %s @0x%x: %#v", typ.GetName(), di.Address, record.RawSample[pos:pos+uint32(di.Length)])
+		pos += uint32(di.Length)
+		if pos%8 > 0 {
+			pos += 8 - pos%8
+		}
```
ajwerner: should we assert anything about the contents?

pkg/dyninst/compiler/compile.go
```diff
@@ -22,29 +22,25 @@ import (
...
-		"-Wno-unused-variable",
-		"-Wno-unused-function",
-		"-DDYNINST_GENERATED_CODE",
-	}
-	if config.BPFDebug {
-		cflags = append(cflags, "-DDEBUG=1")
-	}
-	return cflags
+// BPFObj holds compiled eBPF program and its metadata.
+type BPFObj struct {
```
ajwerner: nit: I'm not in love with this name. What about `compiler.Output` or `compiler.CompiledProgram`?

pkg/dyninst/integration_test.go
```diff
@@ -142,22 +154,31 @@ func testDyninst(t *testing.T, sampleServicePath string) {
...
-	require.NoError(t, err)
-	defer func() { require.NoError(t, bpfProbe.Close()) }()
-
-	attached, err := sampleLink.Uprobe("main.sliceArg", bpfProg, &link.UprobeOptions{
-		PID:    sampleProc.Process.Pid,
-		Cookie: 0,
-	})
-	require.NoError(t, err)
-	defer func() { require.NoError(t, attached.Close()) }()
+	textSection, err := obj.TextSectionHeader()
```
ajwerner, line 157: I've got it on my list to add this to the IR: https://datadoghq.atlassian.net/browse/DEBUG-3883

ajwerner (approved):
LGTM

pkg/dyninst/compiler/sm/instructions.go
```diff
@@ -55,18 +55,22 @@ func makeInstruction(op Op) codeFragment {
 		}
 
 	case ExprDereferenceCfaOp:
-		bytes := make([]byte, 0, 8)
+		bytes := make([]byte, 0, 12)
 		bytes = binary.LittleEndian.AppendUint32(bytes, op.Offset)
 		bytes = binary.LittleEndian.AppendUint32(bytes, op.Len)
+		bytes = binary.LittleEndian.AppendUint32(bytes, op.OutputOffset)
```
grantseltzer, line 61: Anywhere we use binary.LittleEndian we should use binary.NativeEndian.

Although I don't know if this is the only thing that'd make arm64 big endian work.

piob-io: If we were to read them as integers, yes. But program is parsed byte-by-byte, in little endian order: https://github.com/DataDog/datadog-agent/blob/1ddecc062117256242e17edcc8bda6a6588ecc2e/pkg/dyninst/ebpf/stack_machine.h#L85

pkg/dyninst/integration_test.go
piob-io: Sg

pkg/dyninst/compiler/compile.go
piob-io: CompiledBPF

pkg/dyninst/integration_test.go
piob-io: I think it's ok if we do it by attempting to decode in this test. If we see that too hard to debug, we can have a more targeted test that directly checks the bytes for simpler cases.

pkg/dyninst/compiler/sm/instructions.go
grantseltzer: so it'd fail either way. You can just leave it as is then.

PR #37494, opened 2025-05-29
add rollback to config experiments on Windows

clarkb7:
### What does this PR do?
add rollback to config experiments on Windows

Same as update experiments, watchdog process restores stable if services stop or timeout 

Adds `set-config-catalog` daemon API/subcommand set, similar to `set-catalog`, to allow for overriding the configs used by the daemon.

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1529

### Describe how you validated your changes

new e2e tests cover scenario where service starts but then fails somehow
- TestRevertsConfigExperimentWhenServiceDies
- TestRevertsConfigExperimentWhenTimeout

existing e2e test cover scenario where service fails to start at all
- TestConfigUpgradeFailure (invalid config)

### Additional Notes

local API tests work fine but requires follow-up PR for correctness with the backend
- https://github.com/DataDog/datadog-agent/pull/37667

PR #37667, opened 2025-06-04
background experiment hooks for Agent package on Windows

clarkb7:
### What does this PR do?
Fix issue that prevents fleet daemon from reporting task completed state to the backend during experiments, causing an increase in experiment times.

Moves work for Agent package hooks on Windows into a background process. This applies to both update and config experiments. And must apply to any future hooks that restart the daemon.
Adds new `package-command` subcommand to provide an entry point for the new background processes.
process chain is now: `daemon` -> `start-experiment` -> `hook postStartExperiment` -> (detached, no wait) `package-command datadog-agent postStartExperimentBackground`
After the `*Background` process is created, the other processes will exit, unblocking the daemon so it can update task state and also properly handle a stop signal from the `*Background` process.

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1542
Allow package subcommands to exit and return control to daemon, so it can mark task status as DONE (which triggers the backend to move to the next step) and then stop cleanly (without hitting fx timeout).

Updates existing E2E tests
- checks result of daemon experiment commands, which previously returned EOF when the service was killed. These should now return successfully.

Working on future PR to update task state during local API commands and not just remote commands. Then we can assert the task state is DONE in E2E tests, too.

Adding `qa/rc-required` label so we can test remotely updating to release candidates. Working on getting a dev build on staging, too.
- The update should succeed in less than 10 minutes and there shouldn't be a timeout error from the HostUpdateLoop in the backend 

### Additional Notes

Remote updates work in previous versions despite the backend waiting on the task because after the backend hits the 10 minute timeout it rechecks the experiment state, finds it correct, and moved to the next phase. After this change we should see a decrease in overall experiment times for Windows hosts.

After a subprocess terminates, Go's os/exec waits for standard IO handles to complete and these handles can also be inherited by child processes who will hold them open, thereby blocking the parent tree. So we must be careful to detach standard IO from our background processes.

Since update experiments are executed by the experiment's `datadog-installer.exe`, this change will take effect for any remote updates TO 7.68 or later, (e.g. 7.65 -> 7.68), and will not be available for downgrades to a version lower than 7.68.

Creates new `package-command` subcommand interface rather than using the existing `hooks` subcommand interface in order to `hooks` as a clean interface between the installer and the packages.


clarkb7:

clarkb7:

clarkb7:

pkg/fleet/installer/installer.go
```diff
@@ -524,6 +524,11 @@ func (i *installerImpl) InstallConfigExperiment(ctx context.Context, pkg string,
 		)
 	}
 
+	// HACK: close so package can be updated as watchdog runs
+	if pkg == packageDatadogAgent && runtime.GOOS == "windows" {
```
arbll: Not sure I follow this hack, I would have expected the hook to return quickly after launching the background command and close the DB ?

pkg/fleet/installer/packages/packages.go
```diff
@@ -269,3 +269,30 @@ func getHook(pkg string, name string) packageHook {
...
+
+// PackageCommandHandler is a function that handles the execution of a package-specific command.
+//
+// Implement this function and add it to the packagesCommands map to enable package-specific commands
+// for a given package. Package commands are currently intended to be used internally by package hooks
+// and not exposed to the user.
+// For example, the Agent Windows package hooks must start some background worker processes.
+//
+// The content of the command string is entirely defined by the individual package. Do NOT include
+// private information in the command string, use environment variables instead.
```
arbll, line 281: Nit: I think this will likely end up being specialised for the datadog windows agent scripts, so I would probably not make generic tooling for now and only do that if we end up having other usecase. I assume you did that for .net but unfortunately the needs are different

pkg/fleet/installer/installer.go
clarkb7: i copied it over from update's `InstallExperiment`, so I'm not sure about the origin of the `HACK` label, I left the comment as-is to ensure they'd be updated at the same time.

I'd rather close the db ahead of time to prevent any races. I think there's some general work to be done since the db is always opened by the constructor but many of the installer subcommands never actually use it.

clarkb7: oh this change ended up in the diff when github auto rebased it after another PR passed the merge queue. sorry for the confusion.
https://github.com/DataDog/datadog-agent/pull/37494/files#diff-5097e32b945d38543ad437011a6c5e9e11312ede815b9788d7788164c871627a

rtrieu (approved):

jack0x2 (approved):

PR #37668, opened 2025-06-04
CXP-2267 [agent] feat: add language data collection

gengnamstyle:
### What does this PR do?
- add's language collection via `languagedetector` to replicate functionality from [process language collector](https://github.com/DataDog/datadog-agent/blob/feaecfbfd70058b178a2f8d53d7084fbfa0f84ed/comp/core/workloadmeta/collectors/internal/processlanguage/process_collector.go#L52)
    - uses fx to get config and checks if `language_detection.enabled` is true
- uses fx to inject system probe config (process language collector utilizes the non-component version, but we want to move towards componentization as part of the [component framework effort](https://datadoghq.atlassian.net/wiki/spaces/ARUN/pages/2667481965/Component+Framework))

### Motivation
- we already have a process language collector that we want to merge into this new collector as part of the process + service discovery consolidation effort, by duplicating the functionality, we can move forward with eventually removing the process language collector

- unit tests
    - unit tests do not test privileged language collection (for `go`). if wanted I can include them in this pr

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -333,7 +333,6 @@
 /comp/core/sysprobeconfig @DataDog/ebpf-platform
 /comp/core/tagger @DataDog/container-platform
 /comp/core/workloadmeta @DataDog/container-platform
-/comp/core/workloadmeta/collectors/internal/process @DataDog/container-experiences
 /comp/forwarder/eventplatform @DataDog/agent-log-pipelines
 /comp/forwarder/eventplatformreceiver @DataDog/agent-log-pipelines
 /comp/forwarder/orchestrator @DataDog/container-app
@@ -599,6 +598,7 @@
 /pkg/snmp/                              @DataDog/ndm-core
 /pkg/tagger/                            @DataDog/container-platform
 /pkg/windowsdriver/                     @DataDog/windows-agent
+/comp/core/workloadmeta/collectors/internal/process  @DataDog/container-experiences
 /comp/core/workloadmeta/collectors/internal/cloudfoundry  @DataDog/agent-integrations
 /comp/core/workloadmeta/collectors/internal/containerd    @DataDog/container-platform @DataDog/container-integrations
 /comp/core/workloadmeta/collectors/internal/crio          @DataDog/container-platform @DataDog/container-integrations
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -598,7 +598,6 @@
 /pkg/snmp/                              @DataDog/ndm-core
 /pkg/tagger/                            @DataDog/container-platform
 /pkg/windowsdriver/                     @DataDog/windows-agent
-/comp/core/workloadmeta/collectors/internal/process  @DataDog/container-experiences
 /comp/core/workloadmeta/collectors/internal/cloudfoundry  @DataDog/agent-integrations
 /comp/core/workloadmeta/collectors/internal/containerd    @DataDog/container-platform @DataDog/container-integrations
 /comp/core/workloadmeta/collectors/internal/crio          @DataDog/container-platform @DataDog/container-integrations
@@ -610,6 +609,7 @@
 /comp/core/workloadmeta/collectors/internal/kubemetadata  @DataDog/container-platform @DataDog/container-integrations
 /comp/core/workloadmeta/collectors/internal/nvml          @DataDog/ebpf-platform
 /comp/core/workloadmeta/collectors/internal/podman        @DataDog/container-platform @DataDog/container-integrations
+/comp/core/workloadmeta/collectors/internal/process       @DataDog/container-experiences @DataDog/container-platform
 /comp/core/tagger/collectors            @DataDog/container-platform @DataDog/container-integrations
 /pkg/sbom/                              @DataDog/agent-security @DataDog/container-integrations
 /pkg/networkpath/                       @DataDog/Networks
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -333,6 +333,7 @@
 /comp/core/sysprobeconfig @DataDog/ebpf-platform
 /comp/core/tagger @DataDog/container-platform
 /comp/core/workloadmeta @DataDog/container-platform
+/comp/core/workloadmeta/collectors/internal/process @DataDog/container-experiences
 /comp/forwarder/eventplatform @DataDog/agent-log-pipelines
 /comp/forwarder/eventplatformreceiver @DataDog/agent-log-pipelines
 /comp/forwarder/orchestrator @DataDog/container-app
--- a/comp/core/workloadmeta/collectors/internal/process/process_collector.go
+++ b/comp/core/workloadmeta/collectors/internal/process/process_collector.go
@@ -179,7 +179,7 @@ func processCacheDifference(procCacheA map[int32]*procutil.Process, procCacheB m
 	return newProcs
 }
 
-func (c *collector) detectLanguages(processes []*procutil.Process, systemProbeConfig pkgconfigmodel.Reader) []*languagemodels.Language {
+func (c *collector) detectLanguages(processes []*procutil.Process) []*languagemodels.Language {
 	languageInterfaceProcs := make([]languagemodels.Process, len(processes))
 	for i, proc := range processes {
 		languageInterfaceProcs[i] = languagemodels.Process(proc)
--- a/comp/core/workloadmeta/collectors/internal/process/process_collector.go
+++ b/comp/core/workloadmeta/collectors/internal/process/process_collector.go
@@ -13,6 +13,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/DataDog/datadog-agent/comp/core/config"
 	"github.com/DataDog/datadog-agent/comp/core/sysprobeconfig"
 	pkgconfigmodel "github.com/DataDog/datadog-agent/pkg/config/model"
 	"github.com/DataDog/datadog-agent/pkg/languagedetection/languagemodels"
@@ -40,6 +41,7 @@ type collector struct {
 	catalog                workloadmeta.AgentType
 	clock                  clock.Clock
 	processProbe           procutil.Probe
+	config                 pkgconfigmodel.Reader
 	systemProbeConfig      pkgconfigmodel.Reader
 	processEventsCh        chan *Event
 	lastCollectedProcesses map[int32]*procutil.Process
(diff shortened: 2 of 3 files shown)


comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -186,8 +201,12 @@ func (c *collector) collect(ctx context.Context, collectionTicker *clock.Ticker)
 
 			// categorize the processes into events for workloadmeta
 			createdProcs := processCacheDifference(procs, c.lastCollectedProcesses)
-
-			wlmCreatedProcs := createdProcessesToWorkloadmetaProcesses(createdProcs, pidToCid)
+			languageInterfaceProcs := make([]languagemodels.Process, len(createdProcs))
+			for i, proc := range createdProcs {
+				languageInterfaceProcs[i] = languagemodels.Process(proc)
+			}
+			languages := languagedetection.DetectLanguage(languageInterfaceProcs, c.systemProbeConfig)
```
daniel-taf: Let's put this in a separate function.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -109,10 +121,13 @@ func (c *collector) Start(ctx context.Context, store workloadmeta.Component) err
 }
 
 // createdProcessesToWorkloadMetaProcesses helper function to convert createdProcs with container data into wlm entities
-func createdProcessesToWorkloadmetaProcesses(createdProcs []*procutil.Process, pidToCid map[int]string) []*workloadmeta.Process {
+func createdProcessesToWorkloadmetaProcesses(createdProcs []*procutil.Process, pidToCid map[int]string, languages []*languagemodels.Language) []*workloadmeta.Process {
 	wlmProcs := make([]*workloadmeta.Process, len(createdProcs))
 	for i, proc := range createdProcs {
 		wlmProcs[i] = processToWorkloadMetaProcess(proc)
+		wlmProcs[i].Language = languages[i] // TODO: implicit ordering is used to map proc <--> language which is not robust
+
```
daniel-taf: Why do you think it's not robust?

comp/core/workloadmeta/collectors/internal/process/process_collector_test.go
```diff
@@ -148,29 +180,51 @@ func TestCreatedProcessesCollection(t *testing.T) {
...
+			},
+			expectedProcesses: map[int32]*workloadmeta.Process{
+				proc3.Pid: workloadmetaProcess(proc3,
+					&languagemodels.Language{
+						Name: languagemodels.Unknown,
+					}, nil),
+			},
+		},
+		{
+			description:     "single new process with python",
```
daniel-taf: I'm thinking we should try to test the language detection in a separate test than existing ones. The tests are starting to get quite large for "unit" tests.

daniel-taf:
I think CODEOWNERS needs to be updated to include our team for this new collector. Our team isn't tagged in the review.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
gengnamstyle: In general I think assuming order is risky unless explicitly defined. Things could go wrong if either that function is changed or whatever as processes as input is mutated. But we do have tests so and are codeowners, so I guess its fine?

I think the simplest fix is to create/return a mapping from the languagedetection side but I don't want to complicate things more, when this approach was what was used previously. I can remove the comment if you think that's the right step.

daniel-taf: I think its good to acknowledge it as you have. But maybe remove the TODO if we don't have a solid plan in place to change the behavior.

comp/core/workloadmeta/collectors/internal/process/process_collector_test.go
```diff
@@ -56,12 +60,15 @@ func setUpCollectorTest(t *testing.T, configOverrides map[string]interface{}) co
...
-	return collectorTest{&processCollector, mockProbe, mockClock, mockStore, mockContainerProvider}
+	// mock language detection system probe config
+	mockSystemProbeConfig := configmock.New(t)
+	processCollector := newProcessCollector(collectorID, workloadmeta.NodeAgent, mockClock, mockProbe, mockSystemProbeConfig)
+
+	return collectorTest{&processCollector, mockProbe, mockSystemProbeConfig, mockClock, mockStore, mockContainerProvider}
 }
 
-func createTestProcess1(pid int32, createTime int64) *procutil.Process {
+func createTestPythonProcess1(pid int32, createTime int64) *procutil.Process {
```
daniel-taf: Nit: we don't need append the `1` here since the language is already a differentiator. 
```suggestion
func createTestPythonProcess(pid int32, createTime int64) *procutil.Process {
```

comp/core/workloadmeta/collectors/internal/process/process_collector_test.go
```diff
@@ -80,7 +87,7 @@ func createTestProcess1(pid int32, createTime int64) *procutil.Process {
 	return proc
 }
 
-func createTestProcess2(pid int32, createTime int64) *procutil.Process {
+func createTestJavaProcess2(pid int32, createTime int64) *procutil.Process {
```
daniel-taf:
```suggestion
func createTestJavaProcess(pid int32, createTime int64) *procutil.Process {
```

comp/core/workloadmeta/collectors/internal/process/process_collector_test.go
```diff
@@ -100,7 +107,27 @@ func createTestProcess2(pid int32, createTime int64) *procutil.Process {
 	return proc
 }
 
-func workloadmetaProcess(proc *procutil.Process, owner *workloadmeta.EntityID) *workloadmeta.Process {
+func createTestUnknownProcess3(pid int32, createTime int64) *procutil.Process {
```
daniel-taf:
```suggestion
func createTestUnknownProcess(pid int32, createTime int64) *procutil.Process {
```

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -164,6 +179,14 @@ func processCacheDifference(procCacheA map[int32]*procutil.Process, procCacheB m
 	return newProcs
 }
 
+func (c *collector) detectLanguages(processes []*procutil.Process) []*languagemodels.Language {
+	languageInterfaceProcs := make([]languagemodels.Process, len(processes))
+	for i, proc := range processes {
+		languageInterfaceProcs[i] = languagemodels.Process(proc)
+	}
+	return languagedetection.DetectLanguage(languageInterfaceProcs, c.systemProbeConfig)
+}
```
daniel-taf: Forgot to mention on first review, we should only collect the languages if the config `language_detection.enabled` is true.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
gengnamstyle: removed TODO and added a comment

comp/core/workloadmeta/collectors/internal/process/process_collector.go
gengnamstyle: added a check within this function

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -109,10 +130,16 @@ func (c *collector) Start(ctx context.Context, store workloadmeta.Component) err
 }
 
 // createdProcessesToWorkloadMetaProcesses helper function to convert createdProcs with container data into wlm entities
-func createdProcessesToWorkloadmetaProcesses(createdProcs []*procutil.Process, pidToCid map[int]string) []*workloadmeta.Process {
+func createdProcessesToWorkloadmetaProcesses(createdProcs []*procutil.Process, pidToCid map[int]string, languages []*languagemodels.Language) []*workloadmeta.Process {
 	wlmProcs := make([]*workloadmeta.Process, len(createdProcs))
+	isLanguageDataAvailable := len(languages) == len(createdProcs) // language data is not always available, so we have to check
```
daniel-taf, line 135: nit: I think it make more sense to check that the language slice is not nil. `languagedetection.DetectLanguage` always returns a slice that's the same size as `procs`. But, `detectLanguages` returns nil when the feature is not enabled.

daniel-taf (approved):

.github/CODEOWNERS
```diff
@@ -596,6 +596,7 @@
 /pkg/snmp/                              @DataDog/ndm-core
 /pkg/tagger/                            @DataDog/container-platform
 /pkg/windowsdriver/                     @DataDog/windows-agent
+/comp/core/workloadmeta/collectors/internal/process  @DataDog/container-experiences
```
gabedos: Can you make @container-platform also a co-owner of this collector for visibility. I think if multiple teams are co-owners, only a review from 1 team is actually required but we'll still get tagged/notifs from the PR. 

Also small nit: alphabetical ordering of the CODEOWNER. `comp/core/workloadmeta/collectors/internal/process` would be last.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -47,21 +54,30 @@ type Event struct {
...
 
-func newProcessCollector(id string, catalog workloadmeta.AgentType, clock clock.Clock, processProbe procutil.Probe) collector {
+func newProcessCollector(id string, catalog workloadmeta.AgentType, clock clock.Clock, processProbe procutil.Probe, config pkgconfigmodel.Reader, systemProbeConfig pkgconfigmodel.Reader) collector {
 	return collector{
 		id:                     id,
 		catalog:                catalog,
 		clock:                  clock,
 		processProbe:           processProbe,
+		config:                 config,
+		systemProbeConfig:      systemProbeConfig,
```
gabedos, line 64: Slightly unrelated and doesn't need to be added right now but,

All of the wmeta collectors send the wmeta entity events via `c.store.Notify(events)` instead of directly through a channel. It might make sense for uniformity to also store wmeta in the collector and feed the events in `func (c *collector) collect` like that.

gabedos (approved):
Approved with small nits

.github/CODEOWNERS
gengnamstyle: added @DataDog/container-platform as a co-owner!

comp/core/workloadmeta/collectors/internal/process/process_collector.go
gengnamstyle: Yeah I was considering that, but I was worried that this function could then have a fatal error if the length of languages is not the same length as the processes due to the index access in the loop below. I understand we "own" both functions but it seems scary to have a fatal error as a possibility.

PR #37670, opened 2025-06-04
[usm] fix handling of missing pieces in location lists

andreimatei:
### What does this PR do?

The utility we have for interpreting variable location lists, used by
the script that generates the lookup table for Go networking probes, was
buggy when presented with location lists containing missing pieces (i.e.
partially-unavailable variables): the library would silently ignore
missing pieces, and then it would erroneously map the next piece to the
wrong offset within the variable. For example, imagine a 3-word
structure (e.g. a slice header) that's described by a location list
like:
```
DW_OP_fbreg +8, DW_OP_piece 0x8, DW_OP_piece 0x8, DW_OP_fbreg +16, DW_OP_piece 0x8
```

This location list says that the first word is available at some stack
offset, the 2nd word is unavailable, and the 3rd word is again on the
stack. For such a case, `locexpr.Exec()` would return
`[]LocationPiece{{Size:8, StackOffset:16}, {Size:8, StackOffset:24}}`.
There's no information in there about the missing middle piece. USM code
would interpret such a result as the first two words being available and
the third unavailable, which is wrong. This is bad for the live
debugger, which is using this library.

This patch modifies `locexpr.Exec()` to not pretend to deal with location
lists with unavailable pieces and cleans up various TODOs about dealing
with availability gaps. Instead, Exec() now considers such variables to
be completely unavailable (returning an empty result).

I considered extending the results of Exec() to describe
available/unavailable pieces and teaching the callers to consult that,
but it wasn't trivial and it's not needed at the moment. It also doesn't
help that the underlying Delve library we call into does not explicitly
tell you that a piece is unavailable (it tells you that the piece has a
static value of 0). The live debugger will, in the future, want to deal
with partially-available variables, but that's for another day.

Regenerationg the luts.go file with locations for the variables we want
to probe, this patch affects only a single variable (all the others are
completely available). The variable in question has its last piece
missing (it's lucky that it's the last piece because otherwise the bug
would cause corruption). It turns out that, for that particular
variable, we already had special handling because Go 1.16 did not have
debug information for it at all. We were hardcoding locations for Go
1.16, identical to locations for Go 1.13-1.15. We now hardcode the
locations for Go 1.13->1.16.

This patch also fixes another problem in locexpr.Exec() -- it returned 0
as the size of LocationPieces corresponding to single-piece expressions.
Now, it will return the size of the respective variable.

No validation; the change is not intended to have a visible effect.

### DIFF
--- a/pkg/network/go/bininspect/types.go
+++ b/pkg/network/go/bininspect/types.go
@@ -139,10 +139,11 @@ type FunctionMetadata struct {
 	// This is unfortunate; I don't think there is a clean way to work around this
 	// (since the data is just missing).
 	// The best solution I have found is to:
-	// - manually handle cases where entire parameters are missing
-	//   if it's known that they only occur on a specific Go version
-	//   (Go 1.16.* is one that I have found to be troublesome)
-	//   and insert the appropriate parameter locations if known.
+	// - manually handle cases where entire parameters are missing if it's known
+	//   that they only occur on a specific Go version (Go 1.16.* is completely
+	//   missing locations for a param we're interested in, and go 1.13-1.15
+	//   only has partial location info for that param) and insert the
+	//   appropriate parameter locations if known.
 	// - if all of the parameters are missing,
 	//   then I have found that manually re-implementing
 	//   the stack/register allocation algorithm for function arguments/returns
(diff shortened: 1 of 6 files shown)


andreimatei:
Note for reviewers: only the last commit is to be reviewed here. The rest are https://github.com/DataDog/datadog-agent/pull/37663 and https://github.com/DataDog/datadog-agent/pull/37629.

andreimatei:
cc @grantseltzer @ajwerner @piob-io @GreenMatan

PR #37673, opened 2025-06-05
omnibus: disable cache pruning on CI

chouquette:
### What does this PR do?

Stop attempting to remove duplicated tags from our remote cache when running on the CI

### Motivation

The cache key is our only way of addressing a specific cache version & the associated content.
If a cache key can yield different results, then we might end up fetching a cache artifact but only to fail to restore from it.
After the build, we would attempt to upload the new cache version but with the same cache key, which is likely to cause the same situation in the next build and so on, effectively rendering the cache useless.

### DIFF
--- a/tasks/omnibus.py
+++ b/tasks/omnibus.py
@@ -318,13 +318,17 @@ def build(
     os.remove(pip_config_file)
 
     if use_omnibus_git_cache:
-        stale_tags = ctx.run(f'git -C {omnibus_cache_dir} tag --no-merged', warn=True).stdout
-        # Purge the cache manually as omnibus will stick to not restoring a tag when
-        # a mismatch is detected, but will keep the old cached tags.
-        # Do this before checking for tag differences, in order to remove staled tags
-        # in case they were included in the bundle in a previous build
-        for _, tag in enumerate(stale_tags.split(os.linesep)):
-            ctx.run(f'git -C {omnibus_cache_dir} tag -d {tag}')
+        # We only want to evict stalled tags when using the cache locally.
+        # If the cache was fetched through a cache key and downloaded from S3 we
+        # want to keep it immutable.
+        if not use_remote_cache:
+            stale_tags = ctx.run(f'git -C {omnibus_cache_dir} tag --no-merged', warn=True).stdout
+            # Purge the cache manually as omnibus will stick to not restoring a tag when
+            # a mismatch is detected, but will keep the old cached tags.
+            # Do this before checking for tag differences, in order to remove staled tags
+            # in case they were included in the bundle in a previous build
+            for _, tag in enumerate(stale_tags.split(os.linesep)):
+                ctx.run(f'git -C {omnibus_cache_dir} tag -d {tag}')
         with timed(quiet=True) as durations['Updating omnibus cache']:
             if use_remote_cache and ctx.run(f"git -C {omnibus_cache_dir} tag -l").stdout != cache_state:
                 ctx.run(f"git -C {omnibus_cache_dir} bundle create {bundle_path} --tags")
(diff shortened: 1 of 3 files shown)


tasks/omnibus.py
```diff
@@ -318,13 +318,17 @@ def build(
...
 
     if use_omnibus_git_cache:
-        stale_tags = ctx.run(f'git -C {omnibus_cache_dir} tag --no-merged', warn=True).stdout
-        # Purge the cache manually as omnibus will stick to not restoring a tag when
-        # a mismatch is detected, but will keep the old cached tags.
-        # Do this before checking for tag differences, in order to remove staled tags
-        # in case they were included in the bundle in a previous build
-        for _, tag in enumerate(stale_tags.split(os.linesep)):
-            ctx.run(f'git -C {omnibus_cache_dir} tag -d {tag}')
+        # We only want to evict stalled tags when using the cache locally.
```
alopezz:
```suggestion
        # We only want to evict stale tags when using the cache locally.
```

tasks/omnibus.py
```diff
@@ -318,13 +318,17 @@ def build(
...
+        # If the cache was fetched through a cache key and downloaded from S3 we
+        # want to keep it immutable.
```
alopezz: Isn't the cache still mutable even if we don't purge stale tags? In the sense that Omnibus might still write changes to it. Not sure whether this triggers or not the failure mode that we're trying to fix (I still haven't fully grasped it yet).

chouquette: Yes it's theoretically mutable.
Ideally I'd like to have a hard failure when restoring from the remote cache and a change is detected at build time, but that is spread throughout omnibus-ruby and `tasks/omnibus.py` so I'm not sure how to do it cleanly.

In theory, we shouldn't have a case where we fetch from the remote cache and still attempt to build something that isn't marked as `always_build`, but the current state appears to contradict that (https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/968923516/ for instance)

tasks/omnibus.py
```diff
@@ -318,22 +319,25 @@ def build(
...
+                # If we got a cache miss, we can upload the cache
+                if cache_state is None:
+                    with timed(quiet=True) as durations['Updating omnibus cache']:
+                        ctx.run(f"git -C {omnibus_cache_dir} bundle create {bundle_path} --tags")
+                        ctx.run(f"{aws_cmd} s3 cp --only-show-errors {bundle_path} {git_cache_url}")
+                        bundle_dir.cleanup()
+                else:
+                    send_cache_mutation_event(
+                        ctx, os.environ.get('CI_PIPELINE_ID'), remote_cache_name, os.environ.get('CI_JOB_ID')
+                    )
```
alopezz: I think the outcome is more or less the same, but wouldn't it make sense to switch the order of the conditionals? In particular what I mean is that if `cache_state` is `None` we probably don't need to compare to the current state.

```suggestion
            if cache_state is None:
                with timed(quiet=True) as durations['Updating omnibus cache']:
                    ctx.run(f"git -C {omnibus_cache_dir} bundle create {bundle_path} --tags")
                    ctx.run(f"{aws_cmd} s3 cp --only-show-errors {bundle_path} {git_cache_url}")
                    bundle_dir.cleanup()
            elif ctx.run(f"git -C {omnibus_cache_dir} tag -l").stdout != cache_state:
                send_cache_mutation_event(
                    ctx, os.environ.get('CI_PIPELINE_ID'), remote_cache_name, os.environ.get('CI_JOB_ID')
                )
```

It would also be nice if we could add a couple of unit tests to this suite to confirm and document these scenarios and behaviours.
https://github.com/DataDog/datadog-agent/blob/a9930fa7e474bd152da5adef7fc359c816ff8fdb/tasks/unit_tests/omnibus_tests.py#L48

chouquette: fair request, and done

alopezz (approved):

tasks/omnibus.py
alopezz: 👍 Thanks!

PR #37682, opened 2025-06-05
[DJM] Enable GPU integration for Databricks cluster if env. var is present

RickyMarou:
### What does this PR do?

- [Jira](https://datadoghq.atlassian.net/browse/DJM-847)

This PR adds the [GPU agent integration](https://github.com/DataDog/integrations-core/tree/master/gpu) configuration to the databricks install scripts if the environment variable `GPU_MONITORING_ENABLED` is present.

There is currently a problem when using the GPU integration within Databricks cluster: when the init script for the cluster runs, the NVML check fails because the library is not yet initialised.
While I am looking at changing this in the integration directly (by having it poll for a moment before failing and bailing out), we need a more immediate solution to solve for a particular customer. 

The immediate fix is to schedule an agent restart a few seconds in the future when we are in a Databricks cluster with the GPU integration enabled. I tried various methods for that, and it turns out that creating a bash script is the most reliable method.

### Motivation

At Data Jobs Monitoring we have a prospect customer that spins up 15K+ databricks clusters every day (!!), and most, if not all, are GPU-enabled instances. 

We made a demo of GPU Monitoring on Databricks clusters using a duct-taped init bash script on one of our own Databricks workspace, after seeing it, they expressed interest in trying this out themselves. 

To simplify the configuration for them we are doing in the installer. However we don't want all customers to always have the integration enabled. It's a similar case to https://github.com/DataDog/datadog-agent/pull/37089, while the work is direct result of this customer's interest, it is also useful for other customers.

In a similar fashion that we allow customers to opt-in to Databricks logs collection, we want to allow them to opt-in to GPU monitoring.

This is done through the addition of an environment variable. While customers using the "Manual" version of the databricks integration can modify init-script(s) themselves, we can also do follow-up work to make the GPU integration configurable via the Integration tile UI for customers on the "Managed" version of the databricks integration (see how we can dynamically do that [here](https://github.com/DataDog/dogweb/blob/marwan/djm-create-dbx-resources/integration2/databricks/domain/global_init_script.py#L25-L28))

I have tested this by taking the install scripts generated by this PR's pipeline and using it a GPU-enabled cluster, then validating that I was indeed receiving GPU metrics.

The only drawback is that this might end up unused, however considering trends around running ML workloads on databricks, that's very unlikely.

### DIFF
--- a/pkg/fleet/installer/setup/djm/databricks.go
+++ b/pkg/fleet/installer/setup/djm/databricks.go
@@ -15,7 +15,6 @@ import (
 	"os/exec"
 	"regexp"
 	"strings"
-	"syscall"
 	"time"
 
 	"github.com/DataDog/datadog-agent/pkg/fleet/installer/setup/common"
@@ -30,7 +29,7 @@ const (
 	databricksAgentVersion       = "7.66.0-1"
 	fetchTimeoutDuration         = 5 * time.Second
 	gpuIntegrationRestartTimeout = 30 * time.Second
-	restartScriptPath            = "/tmp/datadog-gpu-restart"
+	restartLogFile               = "/var/log/gpu-restart"
 )
 
 var (
@@ -226,6 +225,7 @@ func setClearHostTag(s *common.Setup, tagKey, value string) {
 	s.Span.SetTag("host_tag_value."+tagKey, value)
 }
 
+// setupGPUIntegration configures GPU monitoring integration
 func setupGPUIntegration(s *common.Setup) {
 	if os.Getenv("DD_GPU_MONITORING_ENABLED") != "true" {
 		return
@@ -245,9 +245,7 @@ func setupGPUIntegration(s *common.Setup) {
 
 	s.Span.SetTag("host_tag_set.gpu_monitoring_enabled", "true")
 
-	// NVML is not initialized when the databricks init script executes.
-	// This causes the GPU integration to fail initialisation, we must retry after
-	// a timeout so the NVML checks are passing and the GPU integration starts polling metrics.
+	// Agent must be restarted after NVML initialization, which occurs after init script execution
 	scheduleDelayedAgentRestart(s, gpuIntegrationRestartTimeout)
 }
 
@@ -255,45 +253,13 @@ func setupGPUIntegration(s *common.Setup) {
 func scheduleDelayedAgentRestart(s *common.Setup, delay time.Duration) {
 	s.Out.WriteString(fmt.Sprintf("Scheduling agent restart in %v for GPU monitoring\n", delay))
 
-	// Create log file for the restart process
-	logFile, err := os.OpenFile(restartScriptPath+".log", os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0644)
-	if err != nil {
-		s.Out.WriteString(fmt.Sprintf("Failed to create log file: %v\n", err))
-		return
-	}
-
-	fmt.Fprintf(logFile, "[%s] GPU restart process initiated, delay: %v\n", time.Now().UTC().Format(time.RFC3339), delay)
-	logFile.Close()
-
-	restartCmd := fmt.Sprintf(
-		`sleep %d && \
-echo "[$(date -u +%%Y-%%m-%%dT%%H:%%M:%%SZ)] Restarting Datadog agent for GPU monitoring..." >> %s.log && \
-systemctl restart datadog-agent.service >> %s.log 2>&1 && \
-echo "[$(date -u +%%Y-%%m-%%dT%%H:%%M:%%SZ)] Successfully restarted Datadog agent for GPU monitoring" >> %s.log || \
-echo "[$(date -u +%%Y-%%m-%%dT%%H:%%M:%%SZ)] Failed to restart agent for GPU monitoring: $?" >> %s.log`,
-		int(delay.Seconds()),
-		restartScriptPath,
-		restartScriptPath,
-		restartScriptPath,
-		restartScriptPath,
-	)
-
-	cmd := exec.Command("sh", "-c", restartCmd)
-	cmd.SysProcAttr = &syscall.SysProcAttr{
-		Setsid: true, // Create a new session, detaching from the parent
-	}
-
+	cmd := exec.Command("nohup", "bash", "-c", fmt.Sprintf("echo '[$(date -u +%%Y-%%m-%%dT%%H:%%M:%%SZ)] Waiting %v...' >> %[2]s && sleep %d && echo '[$(date -u +%%Y-%%m-%%dT%%H:%%M:%%SZ)] Restarting agent...' >> %[2]s && systemctl restart datadog-agent >> %[2]s 2>&1", delay, restartLogFile, int(delay.Seconds())))
 	if err := cmd.Start(); err != nil {
-		s.Out.WriteString(fmt.Sprintf("Failed to start restart process: %v\n", err))
+		s.Out.WriteString(fmt.Sprintf("Failed to schedule restart: %v\n", err))
 		return
 	}
 
-	if err := cmd.Process.Release(); err != nil {
-		s.Out.WriteString(fmt.Sprintf("Failed to release process: %v\n", err))
-		return
-	}
-
-	s.Out.WriteString(fmt.Sprintf("GPU restart process started in background (check %s.log)\n", restartScriptPath))
+	s.Out.WriteString("Restart scheduled\n")
 }
 
 func setupDatabricksDriver(s *common.Setup) {
--- a/pkg/fleet/installer/setup/djm/databricks.go
+++ b/pkg/fleet/installer/setup/djm/databricks.go
@@ -225,8 +225,19 @@ func setClearHostTag(s *common.Setup, tagKey, value string) {
 func setupGPUIntegration(s *common.Setup) {
 	if os.Getenv("GPU_MONITORING_ENABLED") != "" {
 		s.Out.WriteString("GPU monitoring enabled via GPU_MONITORING_ENABLED environment variable\n")
+		
+		// Configure datadog.yaml for GPU tags and NVML detection
 		s.Config.DatadogYAML.CollectGPUTags = true
 		s.Config.DatadogYAML.EnableNVMLDetection = true
+		
+		// Configure system-probe.yaml for GPU monitoring
+		if s.Config.SystemProbeYAML == nil {
+			s.Config.SystemProbeYAML = &config.SystemProbeConfig{}
+		}
+		s.Config.SystemProbeYAML.GPUMonitoringConfig = config.GPUMonitoringConfig{
+			Enabled: true,
+		}
+		
 		s.Span.SetTag("gpu_monitoring_enabled", "true")
 	}
 }
--- a/pkg/fleet/installer/setup/djm/databricks.go
+++ b/pkg/fleet/installer/setup/djm/databricks.go
@@ -24,12 +24,12 @@ import (
 )
 
 const (
-	databricksInjectorVersion    = "0.40.0-1"
-	databricksJavaTracerVersion  = "1.49.0-1"
-	databricksAgentVersion       = "7.66.0-1"
-	fetchTimeoutDuration         = 5 * time.Second
-	gpuIntegrationRestartTimeout = 30 * time.Second
-	restartLogFile               = "/var/log/datadog-gpu-restart"
+	databricksInjectorVersion   = "0.40.0-1"
+	databricksJavaTracerVersion = "1.49.0-1"
+	databricksAgentVersion      = "7.66.0-1"
+	fetchTimeoutDuration        = 5 * time.Second
+	gpuIntegrationRestartDelay  = 60 * time.Second
+	restartLogFile              = "/var/log/datadog-gpu-restart"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -220,6 +224,57 @@ func setClearHostTag(s *common.Setup, tagKey, value string) {
 	s.Span.SetTag("host_tag_value."+tagKey, value)
 }
 
+func setupGPUIntegration(s *common.Setup) {
+	if os.Getenv("GPU_MONITORING_ENABLED") != "" {
```
RickyMarou: TODO: check for actual value, "true" or 1. 

Also return early

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -220,6 +223,66 @@ func setClearHostTag(s *common.Setup, tagKey, value string) {
...
+	s.Config.SystemProbeYAML.GPUMonitoringConfig = config.GPUMonitoringConfig{
+		Enabled: true,
+	}
+
+	s.Span.SetTag("gpu_monitoring_enabled", "true")
+
+	// NVML is not initialized when the databricks init script executes.
+	// This causes the GPU integration to fail initialisation, we must retry after
+	// a timeout so the NVML checks are passing and the GPU integration starts polling metrics.
+	scheduleDelayedAgentRestart(s, 30*time.Second)
```
aboitreaud: Can you please make move the delay as const at the top of the file ?

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -220,6 +223,66 @@ func setClearHostTag(s *common.Setup, tagKey, value string) {
...
+	s.Config.DatadogYAML.EnableNVMLDetection = true
+
+	if s.Config.SystemProbeYAML == nil {
+		s.Config.SystemProbeYAML = &config.SystemProbeConfig{}
+	}
+	s.Config.SystemProbeYAML.GPUMonitoringConfig = config.GPUMonitoringConfig{
+		Enabled: true,
+	}
+
+	s.Span.SetTag("gpu_monitoring_enabled", "true")
```
aboitreaud: Can you please put this tag in the same dict as the others ? Logs one are like `host_tag_set.worker_logs_enabled` so `host_tag_set. gpu_monitoring_enabled` would make sense I think for keeping things consistent on install spans

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -220,6 +223,66 @@ func setClearHostTag(s *common.Setup, tagKey, value string) {
 	s.Span.SetTag("host_tag_value."+tagKey, value)
 }
 
+func setupGPUIntegration(s *common.Setup) {
+	gpuEnabled := os.Getenv("GPU_MONITORING_ENABLED")
+
+	if gpuEnabled != "1" && gpuEnabled != "true" {
```
aboitreaud: How common is it for people to enable config vars with `1` instead of `true` ? Asking because it's the first time I see the need and if common, we may have to update other variables too.

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -220,6 +223,66 @@ func setClearHostTag(s *common.Setup, tagKey, value string) {
...
+echo "[$(date -u +%%Y-%%m-%%dT%%H:%%M:%%SZ)] Restarting Datadog agent for GPU monitoring..." >> /tmp/datadog-gpu-restart.log
+service datadog-agent restart >> /tmp/datadog-gpu-restart.log 2>&1
+if [ $? -eq 0 ]; then
+    echo "[$(date -u +%%Y-%%m-%%dT%%H:%%M:%%SZ)] Successfully restarted Datadog agent for GPU monitoring" >> /tmp/datadog-gpu-restart.log
+else
+    echo "[$(date -u +%%Y-%%m-%%dT%%H:%%M:%%SZ)] Failed to restart agent for GPU monitoring" >> /tmp/datadog-gpu-restart.log
+fi
+`, delay, int(delay.Seconds()))
+
+	scriptFile := "/tmp/datadog-gpu-restart.sh"
```
aboitreaud: Move as const at the top of the file

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -220,6 +223,66 @@ func setClearHostTag(s *common.Setup, tagKey, value string) {
...
+	// This causes the GPU integration to fail initialisation, we must retry after
+	// a timeout so the NVML checks are passing and the GPU integration starts polling metrics.
+	scheduleDelayedAgentRestart(s, 30*time.Second)
+}
+
+// scheduleDelayedAgentRestart schedules an agent restart after the specified delay
+func scheduleDelayedAgentRestart(s *common.Setup, delay time.Duration) {
+	s.Out.WriteString(fmt.Sprintf("Scheduling agent restart in %v for GPU monitoring\n", delay))
+
+	script := fmt.Sprintf(`#!/bin/bash
```
aboitreaud: Let's wait for review from team fleet-automation but I think we could explore using this existing restart function https://github.com/DataDog/datadog-agent/blob/cbf55473bb455e8ff016e285f79614aa1a848c8d/pkg/fleet/installer/setup/common/services_nix.go#L17 please ? It has baked-in telemetry, and looks cleaner + more reliable than writing a bash file I think. We can still have the delay before running it.

If we cannot and must use this script, then we need more telemetry around it. It's good to have the logs on the customer infra but we'd prefer having info on our install traces. This function also does `os.exec` and has some telem https://github.com/DataDog/datadog-agent/blob/a8ea5d9587bfa6b146e1040ad52c59493b89f5f0/pkg/fleet/installer/setup/common/setup.go#L158. Can you take inspiration / reuse it ?

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -220,6 +223,66 @@ func setClearHostTag(s *common.Setup, tagKey, value string) {
 	s.Span.SetTag("host_tag_value."+tagKey, value)
 }
 
+func setupGPUIntegration(s *common.Setup) {
+	gpuEnabled := os.Getenv("GPU_MONITORING_ENABLED")
```
aboitreaud: In the [integration readme](https://github.com/DataDog/integrations-core/tree/master/gpu), the env var they seem to have adopted is `DD_ GPU_MONITORING_ENABLED` I think it would make sense to ask the same variable to our users

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -220,6 +223,66 @@ func setClearHostTag(s *common.Setup, tagKey, value string) {
...
+	s.Span.SetTag("gpu_monitoring_enabled", "true")
+
+	// NVML is not initialized when the databricks init script executes.
+	// This causes the GPU integration to fail initialisation, we must retry after
+	// a timeout so the NVML checks are passing and the GPU integration starts polling metrics.
+	scheduleDelayedAgentRestart(s, 30*time.Second)
+}
+
+// scheduleDelayedAgentRestart schedules an agent restart after the specified delay
+func scheduleDelayedAgentRestart(s *common.Setup, delay time.Duration) {
```
RickyMarou: Some more details about this: I initially tried with a goroutine but since the init script terminates, it also kills the goroutine and the delayed restart doesn't happen.
In the end the best solution i've found is this duct-taped bash script. We work in the `/tmp` folder to avoid permissions issues.

pkg/fleet/installer/setup/djm/databricks.go
RickyMarou: yes we should update other env vars. It's something @labbati mentioned, if we only check for env. var presence then we will enable the feature when customers do `FEATURE_ENABLED=false` 😬

pkg/fleet/installer/setup/djm/databricks.go
RickyMarou: I can't use the restart function, albeit I would really like to:
- I need to exit the init-script, so the rest of the databricks cluster init can run and NVML actually start
- If we schedule the restart with this go function in a goroutine, when the init-script exits, it will kill the goroutine, and it will never execute the restart

I could use the `ExecuteCommandWithTimeout` function and instead of a bash script do a command like:

```sh
nohup bash -c "sleep 30 && service restart datadog-agent"
```

But I don't think it will be better. The entire point is that we need to detach the process so we execute the restart after the init script finishes, so I think we can't get any telemetry from the install script for the restart 😬 I would be happy to be proved wrong here.

Really the good solution is to change the integration so it polls for the NVML initialisation (they are already doing polling for other signals, but that's after the NVML checks succeeds). I am in touch with the GPU integration team, they are looking into it as I am writing this.

pkg/fleet/installer/setup/djm/databricks.go
aboitreaud: I don't think we're only checking for the presence of env vars. We do check that the value is `true`, but not `1`. My question is whether or not it's worth checking it can be `1`, on top of checking for `true`, like who does that ? Is it specific to GPU monitoring ?

RickyMarou: Right, after looking closely we explicitly check for `true` for env. vars that are set by our users 👍 

I also put "1" here without thinking too much about it as it's a pretty common thing to be able to use 1/0 interchangeably with true/false for env. vars in my experience 🤷  Maybe I spent too many years in the JS ecosystem 😅😇 

I will change it to only be `true` so it's consistent with the rest and we don't have undocumented behaviours.

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -220,6 +225,64 @@ func setClearHostTag(s *common.Setup, tagKey, value string) {
...
+	s.Out.WriteString(fmt.Sprintf("Scheduling agent restart in %v for GPU monitoring\n", delay))
+
+	script := fmt.Sprintf(`#!/bin/bash
+echo "[$(date -u +%%Y-%%m-%%dT%%H:%%M:%%SZ)] GPU restart script started, waiting %v..." >> %s.log
+sleep %d
+echo "[$(date -u +%%Y-%%m-%%dT%%H:%%M:%%SZ)] Restarting Datadog agent for GPU monitoring..." >> %s.log
+service datadog-agent restart >> %s.log 2>&1
```
aboitreaud: Any reason why you're using `service` rather than the [recommended command](https://docs.datadoghq.com/agent/basic_agent_usage/amazonlinux/) to restart the Agent ?

RickyMarou: That's a good point. I think when I played with `service` from the first time i opened a databricks cluster shell an ran with that forever. But they are Amazon Linux machines so let's stick with the recommended tools from the docs.

RickyMarou: You know what, after reading more closely the docs:
> Note: If the service wrapper is not available on your system, use:

with `service` we don't have to know which version of amazon linux we are runnnig (Amazon Linux 2, Amazon Linux 2022/2023 --vs-- Amazon Linux). It just works on both. Let's keep `service` unless there is a good argument against it.

pkg/fleet/installer/setup/djm/databricks.go
aboitreaud: I was exploring alternatives and consulting with go experts 😅 and it seems possible to avoid the bash script with a child process that we detach with fork/exec/setsid. I can implement that if you want

BaptisteFoy: I'm really not a fan of writing a full bash script on disk to achieve this; it's much safer from a security perspective to have the full command executed from the Go code. Also this feels like we're just adding a Go layer to execute a bash install script again 🥲 

We can do what you suggested:
```bash
nohup bash -c "sleep 30 && service restart datadog-agent"
```

Or we can also detach the process from Go directly (there's a [SO question](https://stackoverflow.com/a/23365445) about this). This is probably cleaner but as `nohup` is POSIX-compliant it's probably fine to keep using it 😄

RickyMarou: yes please

pkg/fleet/installer/setup/djm/databricks.go
BaptisteFoy: [AL2 uses systemd](https://docs.aws.amazon.com/AL2/latest/relnotes/relnotes.html) and so does Amazon Linux 2022/2023, so we can use `systemctl`.

I'd prefer that given that the installer was mostly built under the assumption that the underlying OSes would be running `systemd`; that will save us from some headaches later on

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -220,6 +225,64 @@ func setClearHostTag(s *common.Setup, tagKey, value string) {
...
+if [ $? -eq 0 ]; then
+    echo "[$(date -u +%%Y-%%m-%%dT%%H:%%M:%%SZ)] Successfully restarted Datadog agent for GPU monitoring" >> %s.log
+else
+    echo "[$(date -u +%%Y-%%m-%%dT%%H:%%M:%%SZ)] Failed to restart agent for GPU monitoring" >> %s.log
+fi
+`, delay, restartScriptPath, int(delay.Seconds()), restartScriptPath, restartScriptPath, restartScriptPath, restartScriptPath)
+
+	scriptFile := restartScriptPath + ".sh"
+	err := os.WriteFile(scriptFile, []byte(script), 0755)
```
BaptisteFoy: Should every user be able to run this? I believe `700` would be better here

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -220,6 +225,64 @@ func setClearHostTag(s *common.Setup, tagKey, value string) {
...
+		s.Out.WriteString(fmt.Sprintf("Failed to write restart script: %v\n", err))
+		return
+	}
+
+	cmd := exec.Command("bash", "-c", fmt.Sprintf("nohup %s > /dev/null 2>&1 &", scriptFile))
+	err = cmd.Run()
+	if err != nil {
+		s.Out.WriteString(fmt.Sprintf("Failed to start background restart script: %v\n", err))
+		return
+	}
```
BaptisteFoy: This command exec & the file write above are breaking the current pattern for install scripts:
- The databricks.go file declares what we want to do (what config to write, which packages to install)
- `Run()` in [setup.go](https://github.com/DataDog/datadog-agent/blob/main/pkg/fleet/installer/setup/common/setup.go#L108) actually executes the commands

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -220,6 +225,64 @@ func setClearHostTag(s *common.Setup, tagKey, value string) {
...
+	s.Config.SystemProbeYAML.GPUMonitoringConfig = config.GPUMonitoringConfig{
+		Enabled: true,
+	}
+
+	s.Span.SetTag("host_tag_set.gpu_monitoring_enabled", "true")
+
+	// NVML is not initialized when the databricks init script executes.
+	// This causes the GPU integration to fail initialisation, we must retry after
+	// a timeout so the NVML checks are passing and the GPU integration starts polling metrics.
```
BaptisteFoy: **❓ Question:** ‏Can't we just run the install script after NVML is initialized?

RickyMarou: No we have no control over when the init script is called sadly. The right approach to avoid an agent restart is to have the GPU integration do some polling for the NVML checks rather than a single check. I'm in touch with them so we do this.

Even if we do it we'd have to wait an agent release train and we need to get a working solution for a customer out ASAP. I'll be happy to delete all these restart antics if/when we update the GPU integration.

BaptisteFoy (approved):

pkg/fleet/installer/setup/djm/databricks.go
BaptisteFoy: 😭

PR #37684, opened 2025-06-05
[DJM] Add resource tags to databricks hosts

RickyMarou:
### What does this PR do?

- [Jira](https://datadoghq.atlassian.net/browse/DJM-828)

This PR adds resource tagging for databricks jobs and cluster. It boils down to repeating existing cluster id and job id tags, but using a 💫magic🪄Datadog✨key the tag key this time ([docs](https://datadoghq.atlassian.net/wiki/spaces/METRICSIN/pages/2750514666/Resource+resolution+for+points))

### Motivation

Detailed informal RFC [here](https://docs.google.com/document/d/1JBg1fMHN7glxiaompmU3uDLq5D9P7wk42d9jT5EiRqk/edit?tab=t.0)

TL;DR:
- We want to tag databricks hosts metrics with the databricks custom tags
- These tags are not available in the hosts, but they are in the API
- So we are creating REDAPL resources for databricks jobs and clusters (see https://github.com/DataDog/infrastructure-resources/pull/2048 and https://github.com/DataDog/dogweb/pull/142672) when we crawl
- Then use HMS resolver to tag the metrics

Tested on databricks manuall using the install scripts generated by CI on this very PR.

Not much, we're duplicating existing tag but with a different key, and these are `dd.` tags, so invisible to users.


labbati (approved):
SGTM

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -161,8 +161,12 @@ func setupCommonHostTags(s *common.Setup) {
 	if ok {
 		setHostTag(s, "jobid", jobID)
 		setHostTag(s, "runid", runID)
+		setHostTag(s, "dd.internal.resource:databricks_job", jobID)
 	}
 	setHostTag(s, "data_workload_monitoring_trial", "true")
+
+	setIfExists(s, "DBX_CLUSTER_ID", "dd.internal.resource:databricks_cluster", nil)
```
aboitreaud: Are you sure the env var we need here is `DBX_CLUSTER_ID` ? Above we're using `DB_CLUSTER_ID` to set `databricks_cluster_id` so I just want to confirm

RickyMarou: ah yes nice catch 🙂

aboitreaud (approved):

pkg/fleet/installer/setup/djm/databricks.go
```diff
@@ -157,12 +157,20 @@ func setupCommonHostTags(s *common.Setup) {
 		return clusterNameRegex.ReplaceAllString(v, "_")
 	})
 
-	jobID, runID, ok := getJobAndRunIDs()
-	if ok {
-		setHostTag(s, "jobid", jobID)
-		setHostTag(s, "runid", runID)
+	if jobName := os.Getenv("DD_JOB_NAME"); jobName != "" {
+		setHostTag(s, "dd.internal.resource:databricks_job", jobName)
```
aboitreaud: Curious why we're no longer setting `jobid` and `runid` host tags in that case ? These are used to correlate infra metrics on the trace side panel so I think we still need them

RickyMarou: That's totally a mistake - sorry i forgot to mention this i squeezed that one before handing you over the PR and quickly made a change with Claude Code and had not reviewed the code yet 😞 

We should still set jobid and runid for them. Maybe it's even best to remove the custom logic for when DD_JOB_NAME is present (i.e. one time job runs) because I still need to flesh out how we're going to deal with them in REDAPL.

PR #37685, opened 2025-06-05
(fleet) fix integrations restore python script file permissions on linux

arbll:
This PR fixes a permission bug in the python script used to restore integrations.

The scripts were lacking any enforced mode when creating their `.txt` files. This defaulted to 0666 for files masked by the umask used to install the agent (system-specific). With the migration to the installer the python script are executed in a context where umask is 0, exposing this bug even more.

To fix this I enforced a chmod 0644 for all the files the script creates. I believe there's no race issue since all of this is sequential and you can't upgrade the agent twice in parallel.

### DIFF
--- a/omnibus/python-scripts/packages.py
+++ b/omnibus/python-scripts/packages.py
@@ -69,7 +69,7 @@ def check_file_owner_system_windows(filename):
     # check if file exists
     if not os.path.exists(filename):
         return True
-    
+
     # get NT System account SID
     system_sid = win32security.ConvertStringSidToSid("S-1-5-18")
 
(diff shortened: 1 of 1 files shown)


omnibus/python-scripts/packages.py
```diff
@@ -69,7 +69,7 @@ def check_file_owner_system_windows(filename):
     # check if file exists
     if not os.path.exists(filename):
         return True
-
```
iliakur, line 72: deep 😂

iliakur (approved):

NouemanKHAL (approved):

omnibus/python-scripts/packages.py
arbll: I have auto-format on save enabled everywhere, it's nice until you touch a file that isn't formatted 😬

iliakur: same problem

steveny91 (approved):

PR #37690, opened 2025-06-05
[ACIX-470] Create a top-level invoke task for running all gitlabci linters

Ishirui:
### What does this PR do?

Building upon the work of #37417, continue refactoring the set of invoke tasks used for linting gitlabci-related in files in the agent repo. Notable refactoring changes include:
- Moving a lot of the task logic to `tasks/libs/linter` from the main `linter.py` file, leaving only the absolutely necessary parts in `linter.py`
- Unify the way tasks run on config diffs/full config objects/lists of jobs - before, each task had very different interfaces which was quite confusing. Now they can all be called in the same way and use similarly-built main inputs.
- Create a more robust system for handling and "bubbling up" linting failures to the top-level task, which allows each task to handle displaying failures as it prefers.

This PR also adds a new top-level task, `full-gitlab-ci`, that runs all the gitlabci-related linters together, sharing a preconfigured configuration object which takes a while to compute.
This new task is now run both in CI and as a pre-push hook, which increases the safety of gitlabci modifications.

### Motivation

We had a few linters for gitlabci that were only run manually, which is wasteful - we should run them in CI and/or as a pre-push. The issue is that a couple of these require a "fully-resolved" gitlabci configuration file as input, and computing that takes a while. Therefore it is critical that we do not compute this configuration file multiple times, which is easier if we use one single task grouping all the "sublinters".

Running all linters, also with failing examples + unit tests and in CI.

Performance: we are now running more linters, especially as a pre-push, which makes it quite a bit slower. This might make changes to gitlabci files more frustrating to make as the pre-push hook that is run when modifying them is quite a bit slower than it was before.


.pre-commit-config.yaml
```diff
@@ -118,21 +118,13 @@ repos:
       pass_filenames: false
     - id: gitlab-configuration
       name: gitlab-configuration
-      description: test the gitlab configuration on main
-      entry: 'dda inv linter.gitlab-ci -t main'
+      description: Lint and test GitLab CI configuration files
+      entry: "dda inv linter.full-gitlab-ci -t main"
```
Ishirui: This is still quite slow, even for a pre-push. Running it on the diff instead also does not help too much, as generating the full config/diff is actually one of the more time-consuming parts. Maybe we should disable some checks from the "full" linter suite when running as pre-push and only run them in CI (I think shellcheck is quite slow for example ?)

I will do some proper profiling to see which step(s) take the most time.

CelianR: Shellcheck hasn't been enabled yet, it was meant to only show warnings so you can do a "pre-commit config" for that. Could you do a little benchmark to compare before / after your change? Without shellcheck, I think that it isn't really slower than what we have today

tasks/libs/linter/gitlab.py
```diff
@@ -1,18 +1,370 @@
...
+# === Task code bodies === #
+def gitlabci_lint_task_template(
+    task_body: Callable,
+    success_message: str,
+    ctx,
+    configs_or_diff_file: str | None = None,
+    use_diff: bool = False,
+):
+    """
+    Generic task template for gitlabci linting tasks.
```
CelianR:
```suggestion
    """Generic task template for gitlabci linting tasks.
```
NIT: We prefer to use Google style doc strings now

tasks/libs/linter/gitlab.py
```diff
@@ -1,18 +1,370 @@
...
+):
+    """Helper function used in `full_gitlab_ci` to run a 'sublinter' (i.e. a linting task) and handle any failures."""
+    print(f'[{color_message("INFO", Color.BLUE)}] {info_message}')
+    try:
+        sublinter(*args, **kwargs)
+    except (GitlabLintFailure, MultiGitlabLintFailure) as e:
+        if isinstance(e, MultiGitlabLintFailure):
+            failures.extend(e.failures)
+        else:
+            failures.append(e)
```
CelianR: Weird OOP usage, IMO:
- MultiGitlabLintFailure should extends from GitlabLintFailure -> one type to except
- These classes should have a function to get failures such that we don't have to use `isinstance` but only calling an unpure function / extending with a pure function

tasks/libs/linter/gitlab.py
```diff
@@ -1,18 +1,370 @@
...
+
+        if fail_fast and e.level == FailureLevel.ERROR:
+            print(e.pretty_print())
+            sys.exit(e.exit_code)
```
CelianR: Usually we `raise Exit(...)`, see [this](https://github.com/search?q=repo%3ADataDog%2Fdatadog-agent%20raise%20Exit&type=code)

tasks/linter.py
```diff
@@ -236,101 +238,208 @@ def github_actions_shellcheck(
...
+    test: str = "all",
+    custom_context: str = "",
+    shellcheck_exclude: str = "",
+    shellcheck_verbose: bool = False,
+    shellcheck_args: str = "",
+    shellcheck_only_errors: bool = False,
+    path_jobowners: str = '.gitlab/JOBOWNERS',
+):
+    """
+    Top-level task for running all gitlabci-related linters
```
CelianR:
```suggestion
    """Top-level task for running all gitlabci-related linters.
```

tasks/linter.py
```diff
@@ -236,101 +238,208 @@ def github_actions_shellcheck(
...
+        configs = load_or_generate_gitlab_ci_configs(ctx, configs_or_diff_file)
+        jobs = extract_gitlab_ci_jobs(configs=configs)
+        full_config = configs  # type: ignore
+
+    if not jobs:
+        print(f'[{color_message("INFO", Color.BLUE)}] No changed jobs to lint, skipping')
+        return
+
+    failures = []
+    gitlabci_run_sublinter_helper(
```
CelianR: I think it would be a way better to replace these by decorators such as we can configure a linter in each function, e.g.:
```py
@linter(info_message=..., disable_on_pre_commit=..., allowed_to_fail=...)
def my_lint():
```

tasks/linter.py
```diff
@@ -236,101 +238,208 @@ def github_actions_shellcheck(
...
+    # If we have to generate a gitlabci config object, use a minimally-resolved one
+    # We do not need the fully-resolved one here, and this is much faster
+    configs = load_or_generate_gitlab_ci_configs(
+        ctx, configs_file, input_file=input_file, resolve_only_includes=True, postprocess_options=False
     )
+    try:
+        lint_and_test_gitlab_ci_config(configs, test=test, custom_context=custom_context)
+    except (GitlabLintFailure, MultiGitlabLintFailure) as e:
+        print(e.pretty_print())
+        sys.exit(e.exit_code)
```
CelianR: Same as above

tasks/linter.py
CelianR: Note: the decorator will populate a global list containing all functions / configs

tasks/linter.py
```diff
@@ -236,101 +238,208 @@ def github_actions_shellcheck(
...
-
-        # Lint scripts
-        for keyword in ('before_script', 'script', 'after_script'):
-            if keyword in content:
-                scripts[f'{job}.{keyword}'] = f'#!/bin/bash\n{flatten_script(content[keyword]).strip()}\n'
-
-    shellcheck_linter(ctx, scripts, exclude, shellcheck_args, fail_fast, use_bat, only_errors)
+    except (GitlabLintFailure, MultiGitlabLintFailure) as e:
+        print(e.pretty_print())
+        sys.exit(e.exit_code)
```
CelianR: Same as above

tasks/linter.py
```diff
@@ -236,101 +238,208 @@ def github_actions_shellcheck(
...
 
-    shellcheck_linter(ctx, scripts, exclude, shellcheck_args, fail_fast, use_bat, only_errors)
+    shellcheck_linter(ctx, scripts, exclude, shellcheck_args, fail_fast, only_errors)
 
 
 # === GITLAB === #
 ## === Main linter tasks === ##
 @task
-def gitlab_ci(ctx, test="all", custom_context=None, input_file=".gitlab-ci.yml"):
+def full_gitlab_ci(
```
CelianR, line 256: NIT: Could we add an argument specifying which linter to execute, like `--only lint_and_test_gitlab_ci_config`, `--only lint_and_test_gitlab_ci_config,myotherlinter`?

CelianR: Note: maybe we can make the names more homogeneous / specify them in the `linter` decorator I propose

tasks/linter.py
```diff
@@ -236,101 +238,208 @@ def github_actions_shellcheck(
...
+        check_owners_gitlab_ci_jobs,
+        failures=failures,
+        fail_fast=fail_fast,
+        info_message='Checking gitlabci jobs have defined codeowners...',
+        success_message='All gitlabci jobs have defined codeowners.',
+        jobs=jobs,
+        ci_linters_config=ci_linters_config,
+        jobowners=jobowners,
+    )
+
```
CelianR, line 360: It would be nice to have a recap here of which linter failed / passed. I prefer to not use `exit` but rather throw exceptions, then print a recap and then exit

tasks/linter.py
Ishirui: I wasn't aware of the convention 😓 Thanks for noticing ! Fixed.

tasks/libs/linter/gitlab.py
Ishirui: I hadn't realized this was something from invoke, I thought we were raising the standardlib `SystemExit` which IIRC is *not* best practice.
But if invoke provides something it's definitely better to use it. Fixed, thanks !

tasks/linter.py
```diff
@@ -236,109 +238,230 @@ def github_actions_shellcheck(
...
+    if failures:
+        if len(failures) == 1:
+            grouped_failure = failures[0]
+        else:
+            grouped_failure = MultiGitlabLintFailure(failures)
+
+        with gitlab_section('GitlabCI Linter failures'):
```
Ishirui: Would it be better to create a section per each gitlab ci "sub" linter ? Currently the GitlabLintFailure objects do not track which linter they are from (that info is only contained in the freeform `details`), so I would need to implement this as well

tasks/libs/linter/gitlab_exceptions.py
```diff
@@ -0,0 +1,151 @@
...
+from enum import Enum
+
+from tasks.libs.common.color import Color, color_message
+
+
+class FailureLevel(int, Enum):
+    """Enum for different criticalities of gitlabci linting failures."""
+
+    CRITICAL = 3  # Something went wrong while linting
+    ERROR = 2  # The linter found something wrong with the file being linted
```
CelianR: Is there a real difference between both?

tasks/libs/linter/gitlab_exceptions.py
```diff
@@ -0,0 +1,151 @@
...
+
+    failures: list[SingleGitlabLintFailure]
+
+    def pretty_print(self, min_level: FailureLevel = FailureLevel.IGNORED) -> str:
+        """Outputs a nice string detailing the failure, meant for CLI output."""
+        level_out = self.level.pretty_print()
+        if len(self.entry_points) > 1:
+            entry_point = ", ".join(self.entry_points)
+        elif len(self.entry_points) == 1:
+            entry_point = next(iter(self.entry_points))
```
CelianR: I think it isn't necessary since `', '.join(['a']) == 'a'`

tasks/libs/linter/gitlab.py
```diff
@@ -1,22 +1,403 @@
...
+                    e.ignore()
+                raise e
+
+            # We never get here if the function raises an exception
+            print(f"[{color_message('OK', Color.GREEN)}] {success_message}")
+
+        ALL_GITLABCI_SUBLINTERS.append(wrapper)
+        if run_on_prepush:
+            PREPUSH_GITLABCI_SUBLINTERS.append(wrapper)
+        return wrapper
```
CelianR, line 63: Nice 👏

tasks/linter.py
```diff
@@ -236,109 +242,182 @@ def github_actions_shellcheck(
...
+            "jobs": jobs,
+            "exclude": shellcheck_exclude,
+            "verbose": verbosity > 0,
+            "shellcheck_args": shellcheck_args,
+            "only_errors": shellcheck_only_errors,
+            "ci_linters_config": ci_linters_config,
+            "jobowners": jobowners,
+        }
+
+        to_run = ALL_GITLABCI_SUBLINTERS if running_in_ci() else PREPUSH_GITLABCI_SUBLINTERS
```
CelianR: NIT: I think we should have an option to run everything locally but which isn't enabled by default

.pre-commit-config.yaml
CelianR: Another thing, IMO we should have `--fail-fast` here since it is a hook

CelianR (approved):
Gg! Minor suggestions

tasks/libs/linter/gitlab_exceptions.py
Ishirui: `CRITICAL` isn't used at this point, I just added it for completeness's sake. But yeah maybe better to remove it, it's just confusing.

tasks/libs/linter/gitlab_exceptions.py
Ishirui: Good catch ! Thanks.

tasks/linter.py
Ishirui: Or maybe even go further and add a seperate flag (`--pre-push-linters`) that would be set in the `.pre-commit-config.yaml`, this way the task runs everything by default when called manually ?

CelianR: Yep 👍

tasks/linter.py
CelianR: I think we have to do sections for each sub linter and also for the failure summary

CelianR: Indeed it's interesting to see from which linter the error comes from (even if we can see this via the stack trace)

CelianR (approved):

tasks/linter.py
Ishirui: Probably not going to spend time implementing this since we already have all the individual tasks for running the sublinters if that's OK with you ?

tasks/libs/linter/gitlab.py
```diff
@@ -152,6 +158,12 @@ def shellcheck_gitlab_ci_jobs(
     Note:
         Will raise an Exit if any errors are found.
     """
+    # TODO[@agent-devx]: Remove this once we have shellcheck in CI
```
CelianR:
```suggestion
    # TODO(@agent-devx): Remove this once we have shellcheck in CI
```

CelianR (approved):

PR #37692, opened 2025-06-05
[AMCC-11] dogstatsd: additional metric control mechanism between the sampler and the serializer

remeh:
### What does this PR do?

Adds an additional processing step when the time samplers is flushing data to the serializer for serialization.
This extra step is necessary since the aggregation from the time sampler might possibly generate new metrics, that would not have been correctly visible to the metric control implementation which is running in the listening part of DogStatsD.

This implementation creates a sublist in order to avoid having to do a complete second pass with the configured list. This sublist contain only the metrics names generated from histograms, a heuristic try to do this using a static list of histogram aggregates and percentile suffixes.

Manual tested E2E on an org with the feature enabled, also, will be extensively dogfooded.


pgimalac (approved):
LGTM for Agent Runtimes

comp/dogstatsd/server/server.go
```diff
@@ -64,6 +66,9 @@ var (
 
 	defaultChannelBuckets = []float64{100, 250, 500, 1000, 10000}
 	once                  sync.Once
+
+	histAggrs    = []string{"max", "min", "median", "avg", "sum", "count"}
+	percentileRx = regexp.MustCompile(`\.\d?\dpercentile$`)
```
olivielpeau: Rather than use a hardcoded list/regex of aggregates and percentiles, WDYT about using values based on the configured `histogram_aggregates` and `histogram_percentiles` on this agent?

Pros:
- This would not cause blocked histogram aggregates/percentiles to be missed, since this agent cannot produce histogram aggr/percentiles that do not end with suffixes based on `histogram_aggregates` and `histogram_percentiles`
- This would eliminate the risk of false positives for the histogram matching logic (blocking metrics that are not actually histogram aggregates/percentiles). This risk is quite low though.
- Slightly reduces the coupling of dogstatsd server logic with histogram aggr/percentile logic (if the logic is re-used in the code)
- (minor) this would make the matching logic slightly more efficient, since the list would be smaller (very likely negligible though)

Cons:
- may complexify the code?

remeh: I had this for aggregates, I'll revert the commit where I removed it and I'll implement it for percentile. (I removed it to simplify the code)

pkg/aggregator/time_sampler_worker.go
```diff
@@ -49,6 +57,11 @@ type dumpTrigger struct {
 	done chan error
 }
 
+type blocklistTrigger struct {
+	trigger
```
vickenty: Since the receiver does not use any of the `trigger` fields, would it make sense to send just `*utilstrings.Blocklist`, instead of creating a new struct?

comp/dogstatsd/server/server.go
```diff
@@ -511,11 +512,54 @@ func (s *server) SetExtraTags(tags []string) {
...
+	histoBlocklist := utilstrings.NewBlocklist(histoMetricNames, matchPrefix)
+	s.demultiplexer.ReconfigTimeSamplersBlocklist(&histoBlocklist)
+}
+
+// create a list based on all `metricNames` but only containing metric names
+// with histogram aggregates suffixes.
+func (s *server) createHistogramsBlocklist(metricNames []string) []string {
+	aggrs := s.config.GetStringSlice("histogram_aggregates")
+
+	percentiles := metrics.ParsePercentiles(s.config.GetStringSlice("histogram_percentiles"))
```
olivielpeau, line 540: Instead of `GetStringSlice("histogram_percentiles")`, `pkg/metrics/histogram` uses `structure.UnmarshalKey(config, "histogram_percentiles", &c)`. I'm not sure why but I think it's worth digging into it and making both config accesses consistent

pkg/aggregator/demultiplexer.go
```diff
@@ -54,6 +55,10 @@ type Demultiplexer interface {
 	// Main idea is to reduce the garbage generated by slices allocation.
 	GetMetricSamplePool() *metrics.MetricSamplePool
 
+	// ReconfigTimeSamplersBlocklist triggers a reconfiguration of the blocklist
+	// applied in the time samplers.
+	ReconfigTimeSamplersBlocklist(*utilstrings.Blocklist)
```
vickenty: Since it's `SetBlocklist` in dogstatsd, for consistency:

```suggestion
	SetTimeSamplersBlocklist(*utilstrings.Blocklist)
```

comp/dogstatsd/server/server.go
vickenty, line 540: Would it make sense to move this function to the `metrics` package? So that all handling of percentiles is concentrated in one, rather than multiple packages.

pkg/aggregator/time_sampler.go
```diff
@@ -195,6 +197,12 @@ func (s *TimeSampler) dedupSerieBySerieSignature(
 	}
 
 	for _, serie := range serieBySignature {
+		// it is the final stage before flushing the series to the serialisation
+		// part of the pipeline but also, here is a stage where all series have been
+		// generated & processed (even the ones generated from a histogram metric).
+		if blocklist.Test(serie.Name) {
```
vickenty: Does it need to check for nil blocklist here? You added a check in dogstatsd server code in the previous PR.

pkg/aggregator/time_sampler.go
```diff
@@ -124,7 +125,7 @@ func (s *TimeSampler) newSketchSeries(ck ckey.ContextKey, points []metrics.Sketc
 	return ss
 }
 
-func (s *TimeSampler) flushSeries(cutoffTime int64, series metrics.SerieSink) {
+func (s *TimeSampler) flushSeries(cutoffTime int64, series metrics.SerieSink, blocklist *utilstrings.Blocklist) {
```
vickenty, line 128: Rather than plumbing it through the entire flush call chain, would it make sense to store the blocklist directly in the time sampler, rather than the worker? That way we will need to refer to it only in two places.

comp/dogstatsd/server/server.go
```diff
@@ -511,11 +512,54 @@ func (s *server) SetExtraTags(tags []string) {
...
+	percentileAggrs := make([]string, len(percentiles))
+	for i, percentile := range percentiles {
+		percentileAggrs[i] = fmt.Sprintf("%dpercentile", percentile)
```
olivielpeau, line 543: nit: you might want to add the `.` to the strings here? (same comment for `aggrs`)

vickenty (approved):
Left some cosmetic comments, but otherwise LGTM.

pkg/aggregator/time_sampler_worker.go
```diff
@@ -35,6 +41,8 @@ type timeSamplerWorker struct {
 	samplesChan chan []metrics.MetricSample
 	// use this chan to trigger a flush of the time sampler
 	flushChan chan flushTrigger
+	// use this chan to trigger a blocklist reconfiguration
+	blocklistChan chan blocklistTrigger
```
olivielpeau: why use a `chan blocklistTrigger` rather than just a `chan *utilstrings.Blocklist`? (I don't understand what the trigger struct provides)

olivielpeau (approved):
Thanks! LGTM, I left a few comments inline but none of them are blockers (feel free to either address them or discard them with an explanation)

pkg/aggregator/time_sampler.go
remeh: The `TimeSampler` isn't waiting on a select or anything, each one are actually managed by the TimeSamplerWorker. I'm not sure that it would decrease the size of the chain since it would be on the worker to pass the list to the sampler: AFAIU (only checked quickly) the worker would still have the plumbing to pass the list to the sampler, it would just be stored somewhere else. Please let me know if you think I should look more into this, I can address it in a separate PR.

comp/dogstatsd/server/server.go
remeh: It's use-case is IMO closer to statsd than the percentiles/histograms themselves, don't you think?

pkg/aggregator/time_sampler.go
remeh: Good catch, thank you. (it would've worked because I test for nil in the Test function itself, but I prefer to have it visible in the caller place, showing it _can_ be nil).

comp/dogstatsd/server/server.go
remeh: Since I'm not doing any processing to the aggrs ones as I just read them through the config, there is no step where doing this: I preferred to keep this operation consistent on both the percentiles and aggrs.

pkg/aggregator/time_sampler_worker.go
remeh: Addressed.

pkg/aggregator/demultiplexer.go
remeh: Good idea, addressed.

pkg/aggregator/time_sampler_worker.go
remeh: Adressed.

PR #37693, opened 2025-06-05
[CNM-4440] Add UDP traceroute driver

pimlu:
### What does this PR do?
This PR adds a UDP traceroute driver (for IPv4)
### Motivation
This is the last variant needed to have feature parity with the existing traceroute methods

1. Build for an ec2 box:
```
GOARCH=amd64 go build github.com/DataDog/datadog-agent/pkg/networkpath/traceroute/udp/portable_udp
```
2. `scp` it to the ec2 box
3. Traceroute to example.com: `sudo -E ./portable_udp 23.192.228.84:33434`

Doesn't work for IPv6 yet. Also this isn't tested like the TCP driver yet

### DIFF
--- a/pkg/networkpath/traceroute/common/common.go
+++ b/pkg/networkpath/traceroute/common/common.go
@@ -73,7 +73,7 @@ func LocalAddrForHost(destIP net.IP, destPort uint16) (*net.UDPAddr, net.Conn, e
 	// this is a quick way to get the local address for connecting to the host
 	// using UDP as the network type to avoid actually creating a connection to
 	// the host, just get the OS to give us a local IP and local ephemeral port
-	conn, err := net.Dial("udp4", net.JoinHostPort(destIP.String(), strconv.Itoa(int(destPort))))
+	conn, err := net.Dial("udp", net.JoinHostPort(destIP.String(), strconv.Itoa(int(destPort))))
 	if err != nil {
 		return nil, nil, err
 	}
--- a/pkg/networkpath/traceroute/udp/udp_traceroute_linux.go
+++ b/pkg/networkpath/traceroute/udp/udp_traceroute_linux.go
@@ -14,7 +14,8 @@ import (
 	"github.com/DataDog/datadog-agent/pkg/networkpath/traceroute/packets"
 )
 
-func (u *UDPv4) getTracerouteDriver() (*udpDriver, error) {
+//nolint:unused // This is used, but not on all platforms yet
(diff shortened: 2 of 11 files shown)


pkg/networkpath/traceroute/common/common.go
```diff
@@ -73,7 +73,7 @@ func LocalAddrForHost(destIP net.IP, destPort uint16) (*net.UDPAddr, net.Conn, e
 	// this is a quick way to get the local address for connecting to the host
 	// using UDP as the network type to avoid actually creating a connection to
 	// the host, just get the OS to give us a local IP and local ephemeral port
-	conn, err := net.Dial("udp4", net.JoinHostPort(destIP.String(), strconv.Itoa(int(destPort))))
+	conn, err := net.Dial("udp", net.JoinHostPort(destIP.String(), strconv.Itoa(int(destPort))))
```
hmahmood, line 76: ipv6 is supported now?

pkg/networkpath/traceroute/udp/udp_traceroute_linux.go
```diff
@@ -0,0 +1,36 @@
...
+package udp
+
+import (
+	"fmt"
+
+	"github.com/DataDog/datadog-agent/pkg/networkpath/traceroute/common"
+	"github.com/DataDog/datadog-agent/pkg/networkpath/traceroute/packets"
+)
+
+func (u *UDPv4) getTracerouteDriver() (*udpDriver, error) {
```
hmahmood: Does this return a new instance every time?

pkg/networkpath/traceroute/common/common.go
Fabrice-Matrat: I have a branch for ipv6 that will follow

Fabrice-Matrat (approved):

pkg/networkpath/traceroute/udp/udp_traceroute_linux.go
pimlu: It does, yeah.

hmahmood:
Why? Could we create it once and reuse it?

pkg/networkpath/traceroute/udp/udp_traceroute_linux.go
hmahmood: I think renaming this to `newTracerouteDriver` would be more accurate. `get...` seems to suggest we are getting a singleton.

hmahmood (approved):

PR #37696, opened 2025-06-05
Decoding in dynamic instrumentation

grantseltzer:
### What does this PR do?

Adds the new `decode` package which provides functionality for decoding of raw events from the ringbuffer.

### Motivation

Decoding raw events.

Added the sample service from first iteration of DI, updated integration_test.go with test cases.

### Additional Notes

[DEBUG-4013](https://datadoghq.atlassian.net/browse/DEBUG-4013)

[DEBUG-4013]: https://datadoghq.atlassian.net/browse/DEBUG-4013?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/dyninst/decode/decoder.go
+++ b/pkg/dyninst/decode/decoder.go
@@ -1,0 +1,449 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+//go:build linux_bpf
+
+package decode
+
+import (
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"io"
+	"math"
+	"reflect"
+	"strconv"
+
+	"github.com/go-json-experiment/json/jsontext"
+
+	"github.com/DataDog/datadog-agent/pkg/dyninst/ir"
+	"github.com/DataDog/datadog-agent/pkg/dyninst/output"
+)
+
+type probeEvent struct {
+	event *ir.Event
+	probe *ir.Probe
+}
+
+// Decoder decodes the output of the BPF program into a JSON format.
+// It is not guaranteed to be thread-safe.
+type Decoder struct {
+	program     *ir.Program
+	probeEvents map[ir.TypeID]probeEvent
+}
+
+var errorUnimplemented = errors.New("errorUnimplemented type")
+
+// NewDecoder creates a new Decoder for the given program.
+func NewDecoder(program *ir.Program) (*Decoder, error) {
+	decoder := &Decoder{
+		program:     program,
+		probeEvents: make(map[ir.TypeID]probeEvent),
+	}
+	for _, probe := range program.Probes {
+		for _, event := range probe.Events {
+			decoder.probeEvents[event.Type.ID] = probeEvent{
+				event: event,
+				probe: probe,
+			}
+		}
+	}
+	return decoder, nil
+}
+
+// typeAndAddr is a type and address pair. It is used to uniquely identify a data item in the data items map.
+// Addresses may not be unique to a type, for example if an address is taken for the first field of a struct.
+type typeAndAddr struct {
+	irType uint32
+	addr   uint64
+}
+
+// Decode decodes the given event into the given writer.
+func (d *Decoder) Decode(event output.Event, out io.Writer) error {
+
+	frames, err := event.StackPCs()
+	if err != nil {
+		return err
+	}
+	enc := jsontext.NewEncoder(out)
+	err = enc.WriteToken(jsontext.BeginObject)
+	if err != nil {
+		return err
+	}
+
+	err = enc.WriteToken(jsontext.String("stack_frames"))
+	if err != nil {
+		return err
+	}
+	err = enc.WriteToken(jsontext.BeginArray)
+	if err != nil {
+		return err
+	}
+	for i := range frames {
+		err = enc.WriteToken(jsontext.Uint(frames[i]))
+		if err != nil {
+			return err
+		}
+	}
+	err = enc.WriteToken(jsontext.EndArray)
+	if err != nil {
+		return err
+	}
+
+	err = enc.WriteToken(jsontext.String("captures"))
+	if err != nil {
+		return err
+	}
+	err = enc.WriteToken(jsontext.BeginObject)
+	if err != nil {
+		return err
+	}
+	addressReferenceCount := map[typeAndAddr]output.DataItem{}
+
+	var (
+		rootData []byte
+		rootType *ir.EventRootType
+	)
+	for item, err := range event.DataItems() {
+		if err != nil {
+			return err
+		}
+		if rootType == nil {
+			rootData = item.Data()
+			var ok bool
+			rootType, ok = d.program.Types[ir.TypeID(item.Header().Type)].(*ir.EventRootType)
+			if !ok {
+				return errors.New("expected event of type root first")
+			}
+			continue
+		}
+
+		// We need to keep track of the address reference count for each data item.
+		// This is used to avoid infinite recursion when encoding pointers.
+		// We use a map to store the address reference count for each data item.
+		// The key is a type and address pair.
+		// The value is a data item with a counter of how many times it has been referenced.
+		// If the counter is greater than 1, we know that the data item is a pointer to another data item.
+		// We can then encode the pointer as a string and not as an object.
+		addressReferenceCount[typeAndAddr{
+			irType: uint32(item.Header().Type),
+			addr:   item.Header().Address,
+		}] = item
+	}
+
+	// This map is used to avoid infinite recursion when encoding pointer values.
+	// When an address has already been encountered it is added to this map, and
+	// removed when the recursive function call chain returns. This way unique
+	// branches of an object graph contain the value of the object, and not a
+	// pointer to the object.
+	currentlyEncoding := map[typeAndAddr]struct{}{}
+
+	// We iterate over the 'Expressions' of the EventRoot which contains
+	// metadata and raw bytes of the parameters of this function.
+	for _, expr := range rootType.Expressions {
+		parameterType := expr.Expression.Type
+		parameterData := rootData[expr.Offset : expr.Offset+parameterType.GetByteSize()]
+		err = enc.WriteToken(jsontext.String(expr.Name))
+		if err != nil {
+			return err
+		}
+		err = d.encodeValue(enc, addressReferenceCount, currentlyEncoding, parameterType, parameterData)
+		if err != nil {
+			return fmt.Errorf("error parsing data for field %s: %w", rootType.Name, err)
+		}
+	}
+
+	err = enc.WriteToken(jsontext.EndObject) // captures
+	if err != nil {
+		return err
+	}
+	err = enc.WriteToken(jsontext.EndObject) // full object
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (d *Decoder) encodeValue(
+	enc *jsontext.Encoder,
+	dataItems map[typeAndAddr]output.DataItem,
+	currentlyEncoding map[typeAndAddr]struct{},
+	irType ir.Type,
+	data []byte,
+) error {
+	switch v := irType.(type) {
+	case *ir.BaseType:
+		return d.encodeBaseTypeValue(enc, v, data)
+	case *ir.PointerType:
+		if len(data) < int(v.GetByteSize()) {
+			return errors.New("passed data not long enough for pointer")
+		}
+		addr := binary.NativeEndian.Uint64(data)
+		if addr == 0 { // nil pointers are encoded as 0.
+			err := enc.WriteToken(jsontext.String("nil"))
+			if err != nil {
+				return err
+			}
+			return nil
+		}
+		key := typeAndAddr{
+			irType: uint32(irType.GetID()),
+			addr:   addr,
+		}
+		pointedValue, ok := dataItems[key]
+		if !ok {
+			return errors.New("pointer not found in address pass map")
+		}
+		header := pointedValue.Header()
+		if pointedValue.Header().Address == 0 {
+			return enc.WriteToken(jsontext.String(fmt.Sprintf("0x%x", header.Address)))
+		}
+		err := enc.WriteToken(jsontext.BeginObject)
+		if err != nil {
+			return err
+		}
+		err = enc.WriteToken(jsontext.String("Address"))
+		if err != nil {
+			return err
+		}
+		err = enc.WriteToken(jsontext.String(fmt.Sprintf("0x%x", header.Address)))
+		if err != nil {
+			return err
+		}
+
+		if _, ok := currentlyEncoding[key]; !ok {
+			currentlyEncoding[key] = struct{}{}
+			defer delete(currentlyEncoding, key)
+
+			err = enc.WriteToken(jsontext.String("Value"))
+			if err != nil {
+				return err
+			}
+			err = d.encodeValue(
+				enc,
+				dataItems,
+				currentlyEncoding,
+				d.program.Types[ir.TypeID(header.Type)],
+				pointedValue.Data(),
+			)
+			if err != nil {
+				return fmt.Errorf("could not get pointed-at value: %s", err)
+			}
+		}
+		err = enc.WriteToken(jsontext.EndObject)
+		if err != nil {
+			return err
+		}
+	case *ir.StructureType:
+		err := enc.WriteToken(jsontext.BeginObject)
+		if err != nil {
+			return err
+		}
+		structure := irType.(*ir.StructureType)
+		for _, field := range structure.Fields {
+			err = enc.WriteToken(jsontext.String(field.Name))
+			if err != nil {
+				return err
+			}
+			err = d.encodeValue(
+				enc,
+				dataItems,
+				currentlyEncoding,
+				field.Type,
+				data[field.Offset:field.Offset+field.Type.GetByteSize()],
+			)
+			if err != nil {
+				return err
+			}
+		}
+		err = enc.WriteToken(jsontext.EndObject)
+		if err != nil {
+			return err
+		}
+	case *ir.ArrayType:
+		return errorUnimplemented
+	case *ir.GoEmptyInterfaceType:
+		return errorUnimplemented
+	case *ir.GoInterfaceType:
+		return errorUnimplemented
+	case *ir.GoSliceHeaderType:
+		return errorUnimplemented
+	case *ir.GoSliceDataType:
+		return errorUnimplemented
+	case *ir.GoChannelType:
+		return errorUnimplemented
+	case *ir.GoStringHeaderType:
+		stringHeader := irType.(*ir.GoStringHeaderType)
+		var address, length uint64
+		for _, field := range stringHeader.Fields {
+			if field.Name == "str" {
+				if uint32(field.Type.GetByteSize()) != 8 || len(data) < int(field.Offset+field.Type.GetByteSize()) {
+					return errors.New("malformed string field 'str'")
+				}
+				address = binary.NativeEndian.Uint64(data[field.Offset : field.Offset+uint32(field.Type.GetByteSize())])
+			} else if field.Name == "len" {
+				if field.Type.GetByteSize() != 8 {
+					return errors.New("malformed string field 'len'")
+				}
+				length = binary.NativeEndian.Uint64(data[field.Offset : field.Offset+uint32(field.Type.GetByteSize())])
+			}
+		}
+		stringValue, ok := dataItems[typeAndAddr{
+			irType: uint32(irType.GetID()),
+			addr:   address,
+		}]
+		if !ok {
+			return fmt.Errorf("string content not present in data items")
+		}
+		if len(stringValue.Data()) < int(length) {
+			return errors.New("string content not long enough for known length")
+		}
+		err := enc.WriteToken(jsontext.String(string(stringValue.Data()[:length])))
+		if err != nil {
+			return err
+		}
+		return nil
+	case *ir.GoStringDataType:
+		return errorUnimplemented
+	case *ir.GoMapType:
+		return errorUnimplemented
+	case *ir.GoHMapHeaderType:
+		return errorUnimplemented
+	case *ir.GoHMapBucketType:
+		return errorUnimplemented
+	case *ir.GoSwissMapHeaderType:
+		return errorUnimplemented
+	case *ir.GoSwissMapGroupsType:
+		return errorUnimplemented
+	case *ir.EventRootType:
+		return errorUnimplemented
+	default:
+		return errors.New("invalid type")
+	}
+	return nil
+}
+
+func (d *Decoder) encodeBaseTypeValue(enc *jsontext.Encoder, irType *ir.BaseType, data []byte) error {
+	kind, ok := irType.GetGoKind()
+	if !ok {
+		return errors.New("No go kind")
+	}
+	switch kind {
+	case reflect.Bool:
+		if len(data) < 1 {
+			return errors.New("passed data not long enough for bool")
+		}
+		return enc.WriteToken(jsontext.Bool(data[0] == 1))
+	case reflect.Int:
+		if len(data) < 8 {
+			return errors.New("passed data not long enough for int")
+		}
+		return enc.WriteToken(jsontext.Int(int64(binary.NativeEndian.Uint64(data))))
+	case reflect.Int8:
+		if len(data) < 1 {
+			return errors.New("passed data not long enough for int8")
+		}
+		return enc.WriteToken(jsontext.Int(int64(int8(data[0]))))
+	case reflect.Int16:
+		if len(data) < 2 {
+			return errors.New("passed data not long enough for int16")
+		}
+		return enc.WriteToken(jsontext.Int(int64(int16(binary.NativeEndian.Uint16(data)))))
+	case reflect.Int32:
+		if len(data) != 4 {
+			return errors.New("passed data not long enough for int32")
+		}
+		return enc.WriteToken(jsontext.Int(int64(int32(binary.NativeEndian.Uint32(data)))))
+	case reflect.Int64:
+		if len(data) != 8 {
+			return errors.New("passed data not long enough for int64")
+		}
+		return enc.WriteToken(jsontext.Int(int64(binary.NativeEndian.Uint64(data))))
+	case reflect.Uint:
+		if len(data) != 8 {
+			return errors.New("passed data not long enough for uint")
+		}
+		return enc.WriteToken(jsontext.Int(int64(binary.NativeEndian.Uint64(data))))
+	case reflect.Uint8:
+		if len(data) != 1 {
+			return errors.New("passed data not long enough for uint8")
+		}
+		return enc.WriteToken(jsontext.Int(int64(uint8(data[0]))))
+	case reflect.Uint16:
+		if len(data) != 2 {
+			return errors.New("passed data not long enough for uint16")
+		}
+		return enc.WriteToken(jsontext.Int(int64(uint16(binary.NativeEndian.Uint16(data)))))
+	case reflect.Uint32:
+		if len(data) != 4 {
+			return errors.New("passed data not long enough for uint32")
+		}
+		return enc.WriteToken(jsontext.Int(int64(uint32(binary.NativeEndian.Uint32(data)))))
+	case reflect.Uint64:
+		if len(data) != 8 {
+			return errors.New("passed data not long enough for uint64")
+		}
+		return enc.WriteToken(jsontext.Int(int64(binary.NativeEndian.Uint64(data))))
+	case reflect.Uintptr:
+		if len(data) != 8 {
+			return errors.New("passed data not long enough for uintptr")
+		}
+		return enc.WriteToken(jsontext.Int(int64(binary.NativeEndian.Uint64(data))))
+	case reflect.Float32:
+		if len(data) != 4 {
+			return errors.New("passed data not long enough for float32")
+		}
+		return enc.WriteToken(jsontext.Float(float64(math.Float32frombits(binary.NativeEndian.Uint32(data)))))
+	case reflect.Float64:
+		if len(data) != 8 {
+			return errors.New("passed data not long enough for float64")
+		}
+		return enc.WriteToken(jsontext.Float(math.Float64frombits(binary.NativeEndian.Uint64(data))))
+	case reflect.Complex64:
+		if len(data) != 8 {
+			return errors.New("passed data not long enough for complex64")
+		}
+		realBits := math.Float32frombits(binary.NativeEndian.Uint32(data[0:4]))
+		imagBits := math.Float32frombits(binary.NativeEndian.Uint32(data[4:8]))
+		return enc.WriteToken(jsontext.String(strconv.FormatComplex(complex(float64(realBits), float64(imagBits)), 'f', -1, 64)))
+	case reflect.Complex128:
+		if len(data) != 16 {
+			return errors.New("passed data not long enough for complex128")
+		}
+		realBits := math.Float64frombits(binary.NativeEndian.Uint64(data[0:8]))
+		imagBits := math.Float64frombits(binary.NativeEndian.Uint64(data[8:16]))
+		err := enc.WriteToken(jsontext.Float(float64(realBits)))
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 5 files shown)


pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,332 @@
...
+	"github.com/go-json-experiment/json/jsontext"
+)
+
+type probeEvent struct {
+	event *ir.Event
+	probe *ir.Probe
+}
+
+type Decoder struct {
+	uploader               *jsontext.Encoder
```
ajwerner: I'm not so confident that directly writing always into a stream is the right approach. Errors become unrecoverable. I do think that having a single jsontext.Encoder per Decoder for reuse makes sense, but I think the signature of Decode should change to output bytes. That's what anybody on the other side is going to want too I think. In that way you'd use a bytes.Buffer that you'd also reuse and then copy out at the end.

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,332 @@
...
+	eventRoot, ok := d.program.Types[ir.TypeID(firstDataItem.header.Type)].(*ir.EventRootType)
+	if !ok {
+		return errors.New("expected event of type root first")
+	}
+
+	err = d.uploader.WriteToken(jsontext.BeginObject)
+	if err != nil {
+		return err
+	}
+	defer d.uploader.WriteToken(jsontext.EndObject)
```
ajwerner: I probably wouldn't defer this -- in an error it's hard and I think not worthwhile to be trying to produce a legit json object.

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,332 @@
...
+func (d *Decoder) parseDataToGoValue(addressPassMap map[uint64]*dataItem, irType ir.Type, data []byte) (any, error) {
+	switch v := irType.(type) {
+	case *ir.BaseType:
+		return d.parseDataToGoBaseTypeValue(v, data)
+	case *ir.PointerType:
+		if len(data) < int(v.GetByteSize()) {
+			return nil, errors.New("passed data not long enough for pointer")
+		}
+		addr := binary.NativeEndian.Uint64(data)
+		pointedValue, ok := addressPassMap[addr]
```
ajwerner: for reasons, I think this map should include the type information of the pointee and should resolve given it.

pkg/ebpf/bytecode/runtime/dyninst_event.go
ajwerner, line 1: You can delete this

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,332 @@
...
+		}
+		err = json.MarshalEncode(d.uploader, out)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (d *Decoder) parseDataToGoValue(addressPassMap map[uint64]*dataItem, irType ir.Type, data []byte) (any, error) {
```
ajwerner: nit: `addressPassMap` as a name is not particularly evocative for me

pkg/dyninst/decode/json_schema.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+// StackFrame represents a single entry in a stack trace
+type StackFrame struct {
+	FileName string `json:"fileName,omitempty"`
+	Function string `json:"function,omitempty"`
+	Line     int    `json:"lineNumber,omitempty"`
+}
+
+func new() {
```
ajwerner: ?

pkg/dyninst/decode/json_schema.go
ajwerner, line 1: This doesn't seem like it's in use

pkg/dyninst/decode/json_schema.go
```diff
@@ -0,0 +1,128 @@
...
+// Copyright 2016-present Datadog, Inc.
+
+//go:build linux_bpf
+
+package decode
+
+import (
+	"github.com/google/uuid"
+)
+
```
grantseltzer: This file is going to go away, i'm just using it as reference.

pkg/dyninst/decode/doc.go
```diff
@@ -0,0 +1,11 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+//go:build linux_bpf
+
+// Package decode is responsible for decoding raw event bytes from dynamic
+// instrumentation's ebpf ringbuffer into JSON toe be uploaded to the datadog
```
ajwerner: s/toe/to/

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,332 @@
...
+		//TODO
+	case *ir.EventRootType:
+		//TODO
+	default:
+		return nil, errors.New("invalid type")
+	}
+	return nil, nil
+}
+
+func (d *Decoder) parseDataToGoBaseTypeValue(irType *ir.BaseType, data []byte) (any, error) {
```
ajwerner: in general, I'm not convinced we want to bounce through a go type at all. Also, for snapshots there's a whole envelope protocol you seem to have skipped, right? I think it'd be better to just thread the `jsontext.Encoder` down the call stack.

I think you'll find it is not more code

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,332 @@
...
+	case *ir.PointerType:
+		if len(data) < int(v.GetByteSize()) {
+			return nil, errors.New("passed data not long enough for pointer")
+		}
+		addr := binary.NativeEndian.Uint64(data)
+		pointedValue, ok := addressPassMap[addr]
+		if !ok {
+			return nil, errors.New("pointer not found in address pass map")
+		}
+		pointeeValue, err := d.parseDataToGoValue(addressPassMap, d.program.Types[ir.TypeID(pointedValue.header.Type)], pointedValue.data)
```
ajwerner: be careful, this can blow out the stack whenever there's a pointer cycle! I think we need to thread down the recursion some limit -- or some tracker of the pointers we've already denormalized.

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,332 @@
...
+					return nil, errors.New("malformed string field 'len'")
+				}
+				length = binary.NativeEndian.Uint64(data[field.Offset : field.Offset+uint32(field.Type.GetByteSize())])
+			} else if field.Name == "cap" {
+				if field.Type.GetByteSize() != 8 {
+					return nil, errors.New("malformed string field 'len'")
+				}
+				cap = binary.NativeEndian.Uint64(data[field.Offset : field.Offset+uint32(field.Type.GetByteSize())])
+			} else {
+				fmt.Println(field.Name)
```
ajwerner: ?

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,332 @@
...
+			}
+		}
+
+		//TODO
+		fmt.Println(array, length, cap)
+
+	case *ir.GoSliceDataType:
+		//TODO
+	case *ir.GoChannelType:
+		//TODO
```
ajwerner: let's return unimplemented errors and then test them explicitly -- or do the implementation

pkg/dyninst/decode/data_item_iterator.go
```diff
@@ -0,0 +1,178 @@
...
+	header *output.DataItemHeader
+	data   []byte
+}
+
+type dataEventIterator struct {
+	buf            []byte
+	eventHeader    *output.EventHeader
+	idx            int
+	addressPassMap map[uint64]*dataItem
+}
```
ajwerner: This thing is doing more than being an iterator. In particular, it seems odd to me for it to have anything to do with holding onto this map. The user of this thing should be interacting with the map. 

Here's what I think I would do for this:

```go
type dataItem struct {
    header *output.DataItemHeader
    data []byte
}

type Event []byte

func (e Event) eventHeader() (*output.EventHeader, error) { ... }

func (e Event) dataItems() iter.Seq2[dataItem, error] { ... }

// An optimization:
func (e Event) firstDataItem() (dataItem, error) { ... } 

func (e Event) stackFrames() ([]uint64, error) { ... }
```

In this model I think we avoid all allocations.

pkg/dyninst/integration_test.go
```diff
@@ -25,46 +29,958 @@ import (
...
+	expectedDecodingOutput string
+	expectedToPass         bool
 }
 
 func TestDyninst(t *testing.T) {
-	skipIfKernelNotSupported(t)
-	cfgs := testprogs.GetCommonConfigs(t)
-	for _, cfg := range cfgs {
-		t.Run(cfg.String(), func(t *testing.T) {
+	testCases := []testCase{
```
ajwerner: this is going to get out of hand quickly. I think we should replace it with something more snapshot oriented or datadriven sooner rather than later, but perhaps not in this commit or pr.

ajwerner (commented):
Here's my quick initial feedback

pkg/dyninst/decode/json_schema.go
ajwerner: Ack, left my comment below before seeing this.

pkg/dyninst/decode/decoder.go
grantseltzer: You make a good point about recovering from errors being difficult, and we don't want to send stream partial JSON.

Isn't that the whole point of passing io.Writer (in `NewDecoder`) though? The user of this API can just pass a bytes.Buffer as you suggest and if there's no error, upload the JSON. And clear the buffer regardless.

pkg/dyninst/decode/decoder.go
grantseltzer: Can you elaborate? Why is address not acceptable?

pkg/dyninst/decode/data_item_iterator.go
grantseltzer: I like this model and have implemented it. One thing i'll point out in line though for a chance for further optimization.

pkg/dyninst/integration_test.go
grantseltzer: Yea agreed, been mostly for sanity at the moment.

pkg/dyninst/decode/decoder.go
grantseltzer: Can you elaborate on "bounce through a go type"? 

And by envelope protocol do you mean the enclosing structure of the JSON that actually gets uploaded?

pkg/ebpf/bytecode/runtime/dyninst_event.go
brycekahle, line 1: This file should not be committed.

pkg/dyninst/decode/data_item_iterator.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+
+		idx := eventHeaderSize // Skip event header
+		if len(e) < idx+int(eventHeader.Stack_byte_len) {
+			yield(dataItem{}, errors.New("not enough bytes to read stack trace"))
+			return
+		}
+		idx = advanceBuffer(idx, int(eventHeader.Stack_byte_len)) // Skip stack trace data
+		for {
+			if idx+dataItemHeaderSize >= len(e) {
```
ajwerner: return an error if there's garbage at the end

pkg/dyninst/decode/data_item_iterator.go
```diff
@@ -0,0 +1,114 @@
...
+	return func(yield func(dataItem, error) bool) {
+		eventHeader, err := e.eventHeader()
+		if err != nil {
+			yield(dataItem{}, err)
+			return
+		}
+
+		idx := eventHeaderSize // Skip event header
+		if len(e) < idx+int(eventHeader.Stack_byte_len) {
+			yield(dataItem{}, errors.New("not enough bytes to read stack trace"))
```
ajwerner: It's generally helpful to report the relevant indexes in the event

pkg/dyninst/decode/data_item_iterator.go
```diff
@@ -0,0 +1,114 @@
...
+		idx = advanceBuffer(idx, int(eventHeader.Stack_byte_len)) // Skip stack trace data
+		for {
+			if idx+dataItemHeaderSize >= len(e) {
+				return
+			}
+			header := (*output.DataItemHeader)(unsafe.Pointer(&e[idx]))
+			idx = advanceBuffer(idx, dataItemHeaderSize)
+
+			if idx+int(header.Length) > len(e) {
+				yield(dataItem{}, errors.New("not enough bytes to read data item"))
```
ajwerner: It's generally helpful to report the relevant indexes in the event

pkg/dyninst/decode/decoder.go
ajwerner: > Isn't that the whole point of passing io.Writer (in NewDecoder) though? The user of this API can just pass a bytes.Buffer as you suggest and if there's no error, upload the JSON. And clear the buffer regardless.

To me the point is about not needing to allocate intermediate buffers during decoding. It'd be reasonable to make `Decode` take an `io.Writer` so the caller can control the lifetime of that buffer.

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,377 @@
...
+
+	err = d.uploader.WriteToken(jsontext.String("captures"))
+	if err != nil {
+		return err
+	}
+	err = d.uploader.WriteToken(jsontext.BeginObject)
+	if err != nil {
+		return err
+	}
+	itemsInMemory := map[uint64]*dataItem{}
```
ajwerner: Why the `*dataItem` given we did all that work to make `dataItem` not have to allocate?

ajwerner: What do you think of `itemsByAddr` as a variable name? Everything is "InMemory"

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,377 @@
...
+	for item, err := range event.dataItems() {
+		if err != nil && errors.Is(err, finishedIterating) {
+			return nil
+		} else if err != nil {
+			return err
+		}
+		if counter == 0 {
```
ajwerner: get rid of `counter` and just do `if eventRoot != nil { ... } else { ... }`

pkg/dyninst/decode/decoder.go
ajwerner: ??

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,377 @@
...
+			} else if field.Name == "cap" {
+				if field.Type.GetByteSize() != 8 {
+					return errors.New("malformed string field 'len'")
+				}
+				cap = binary.NativeEndian.Uint64(data[field.Offset : field.Offset+uint32(field.Type.GetByteSize())])
+			} else {
+				fmt.Println(field.Name)
+			}
+		}
+		fmt.Println(array, length, cap)
```
ajwerner: debugging code

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,377 @@
...
+	err = d.uploader.WriteToken(jsontext.EndObject) // full object
+	if err != nil {
+		return err
+	}
+
+	return nil
+
+}
+
+func (d *Decoder) parseDataToGoValue(dataItemsInMemory map[uint64]*dataItem, irType ir.Type, data []byte) error {
```
ajwerner: I don't think the name of this function makes sense anymore

ajwerner (commented):
I implore you to stop writing directly to the io.Writer. The jsontext.Encoder is doing a good bit to ensure that you don't create an invalid json object.

pkg/dyninst/decode/decoder.go
grantseltzer: Sounds reasonable to me too

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,377 @@
...
+		return err
+	}
+
+	err = d.uploader.WriteToken(jsontext.BeginObject)
+	if err != nil {
+		return err
+	}
+
+	d.uploader.WriteToken(jsontext.String("stack_frames"))
+	d.uploader.WriteToken(jsontext.BeginArray)
```
ajwerner: missing error checking

pkg/dyninst/decode/decoder.go
ajwerner: Another thing we can do is stick this map into the Decoder itself and then defer clearing it so that we don't reallocate the hash map on every decode call.

grantseltzer: It'd just mean we can't decode from multiple routines, unless you think having a lock is worth it to save the allocations.

pkg/dyninst/decode/decoder.go
grantseltzer: I made a patch to fix this, what do you think about the UX of something like:


```go

type node struct {
	val int
	b   *node
}

//nolint:all
//go:noinline
func test_pointer_loop(a *node) {}
```
```json
{
  "stack_frames": [
    8448128,
    8449724,
    331092,
    573076
  ],
  "captures": {
    "a": {
      "Address": "0x400026d7a0",
      "Value": {
        "val": 1,
        "b": "0x400026d7a0"
      }
    }
  }
}
```
(will also add type info)

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,383 @@
...
+)
+
+type probeEvent struct {
+	event *ir.Event
+	probe *ir.Probe
+}
+
+type Decoder struct {
+	program                *ir.Program
+	typesToProbesAndEvents map[ir.TypeID]probeEvent
```
ajwerner: nit: what if we call this `probeEvents` or if you insist `probeEventsByTypeID`? It's not every type that is in this map.

pkg/dyninst/decode/decoder.go
ajwerner: Because in go you can take the address of a field of a struct and then it's a different type of pointer but has the same address. Consider:

```go
type Outer struct {
    V Inner
    X int
}

type Inner struct {
    // ...
}
```

If in the go program somebody took the address of `Outer.V` then it'd be a pointer to `Inner` with the same address as `Outer`

pkg/dyninst/decode/decoder.go
ajwerner: It's a reasonable starting point, yeah. Later we'll want to see what happens in the UI.

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,408 @@
...
+			}
+		}
+	}
+	return decoder, nil
+}
+
+type dataItemWithAddrCounter struct {
+	dataItem    output.DataItem
+	addrCounter uint8
+}
```
ajwerner: this deserves commentary -- what is an `addrCounter`?

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,407 @@
...
+
+func (d *Decoder) encodeValue(enc *jsontext.Encoder, itemsByAddress map[uint64]*dataItemWithAddrCounter, irType ir.Type, data []byte) error {
+	switch v := irType.(type) {
+	case *ir.BaseType:
+		return d.encodeBaseTypeValue(enc, v, data)
+	case *ir.PointerType:
+		if len(data) < int(v.GetByteSize()) {
+			return errors.New("passed data not long enough for pointer")
+		}
+		// handle nil pointer
```
ajwerner: nit: does this comment add a lot of value? Also it feels slightly misplaced. I could see it next to the `if` as an inline comment.

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,407 @@
...
+	itemsByAddress := map[uint64]*dataItemWithAddrCounter{}
+	for item, err := range event.DataItems() {
+		if errors.Is(err, output.ErrFinishedIterating) {
+			break
+		}
+		if err != nil {
+			return err
+		}
+		if eventRoot == nil {
+			// item is EventRoot
```
ajwerner: nit: does this comment add much value? I could see more exposition or just eliminating it.

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,407 @@
...
+			}
+			return nil
+		}
+		pointedValue, ok := itemsByAddress[addr]
+		if !ok {
+			return errors.New("pointer not found in address pass map")
+		}
+		pointedValue.addrCounter++
+		if pointedValue.addrCounter > 1 {
+			return enc.WriteToken(jsontext.String(fmt.Sprintf("0x%x", pointedValue.dataItem.Header().Address)))
```
ajwerner: nit: pull header and data out into variables -- if only because it'll make these very long lines shorter.

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,407 @@
...
+		}
+		err = enc.WriteToken(jsontext.String(fmt.Sprintf("0x%x", pointedValue.dataItem.Header().Address)))
+		if err != nil {
+			return err
+		}
+		err = enc.WriteToken(jsontext.String("Value"))
+		if err != nil {
+			return err
+		}
+		err = d.encodeValue(enc, itemsByAddress, d.program.Types[ir.TypeID(pointedValue.dataItem.Header().Type)], pointedValue.dataItem.Data())
```
ajwerner: nit: can we do some wrapping -- I'm not militant but 136 is quite qide

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,407 @@
...
+	case reflect.Float32:
+		if len(data) != 4 {
+			return errors.New("passed data not long enough for float32")
+		}
+		return enc.WriteToken(jsontext.Float(float64(math.Float32frombits(binary.NativeEndian.Uint32(data)))))
+	case reflect.Float64:
+		if len(data) != 8 {
+			return errors.New("passed data not long enough for float64")
+		}
+		return enc.WriteToken(jsontext.Float(float64(math.Float64frombits(binary.NativeEndian.Uint64(data)))))
```
ajwerner: nit: isn't this cast meaningless (because Float64frombits returns a float64)?

pkg/dyninst/output/event_iterator.go
```diff
@@ -0,0 +1,134 @@
...
+	numFrames := stackTraceLen / 8
+	if stackTraceLen%8 != 0 {
+		return nil, errors.New("stack trace length is not a multiple of 8 bytes")
+	}
+	frames := make([]uint64, numFrames)
+	for i := range numFrames {
+		framePtr := (*uint64)(unsafe.Pointer(&stackData[i*8]))
+		frames[i] = *framePtr
+	}
+	return frames, nil
```
ajwerner: Elsewhere we go through some pain to reuse the data directly, why don't we do that here? In particular, can't we use https://pkg.go.dev/unsafe#Slice?

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,407 @@
...
+	"github.com/DataDog/datadog-agent/pkg/dyninst/ir"
+	"github.com/DataDog/datadog-agent/pkg/dyninst/output"
+)
+
+type probeEvent struct {
+	event *ir.Event
+	probe *ir.Probe
+}
+
+// Decoder decodes the output of the BPF program into a JSON format.
```
ajwerner, line 30: Comment on whether this thing is thread safe? It currently is -- but do we want to lock ourselves into that?

pkg/dyninst/output/event_iterator.go
```diff
@@ -0,0 +1,134 @@
...
+
+		idx := eventHeaderSize // Skip event header
+		if len(e) < idx+int(header.Stack_byte_len) {
+			yield(DataItem{}, fmt.Errorf("not enough bytes to read stack trace: %d < %d", len(e), idx+int(header.Stack_byte_len)))
+			return
+		}
+		idx = advanceBuffer(idx, int(header.Stack_byte_len)) // Skip stack trace data
+		for {
+			if idx == len(e) {
+				yield(DataItem{}, ErrFinishedIterating)
```
ajwerner: I don't understand the reason for `ErrFinishedIterating`. You just return

pkg/dyninst/output/event_iterator.go
```diff
@@ -0,0 +1,134 @@
...
+	return func(yield func(DataItem, error) bool) {
+		header, err := e.Header()
+		if err != nil {
+			yield(DataItem{}, err)
+			return
+		}
+
+		idx := eventHeaderSize // Skip event header
+		if len(e) < idx+int(header.Stack_byte_len) {
+			yield(DataItem{}, fmt.Errorf("not enough bytes to read stack trace: %d < %d", len(e), idx+int(header.Stack_byte_len)))
```
ajwerner: nit: wrap a bit here and elsewhere

pkg/dyninst/output/event_iterator.go
```diff
@@ -0,0 +1,134 @@
...
+				return
+			}
+			if idx+dataItemHeaderSize >= len(e) {
+				yield(DataItem{}, fmt.Errorf("not enough bytes to read data item header: %d > %d", idx+dataItemHeaderSize, len(e)))
+				return
+			}
+			header := (*DataItemHeader)(unsafe.Pointer(&e[idx]))
+			idx = advanceBuffer(idx, dataItemHeaderSize)
+
+			if idx+int(header.Length) > len(e) {
```
ajwerner: nit: assign `int(header.Length)` to a variable, perhaps `dataLen`?

pkg/dyninst/output/event_iterator.go
```diff
@@ -0,0 +1,134 @@
...
+				data:   data,
+			}
+			if !yield(item, nil) {
+				return
+			}
+		}
+	}
+}
+
+// StackPCs decodes the program counters of the stack trace from the event
```
ajwerner: nit: end sentences with a `.`.

pkg/dyninst/output/event_iterator.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		return nil, err
+	}
+	idx := eventHeaderSize
+	stackTraceLen := int(header.Stack_byte_len)
+	if len(e) < idx+stackTraceLen {
+		return nil, errors.New("not enough bytes to read stack trace")
+	}
+
+	// Each stack frame is 8 bytes (uint64)
```
ajwerner: nit: end sentences with a ..

pkg/dyninst/output/event_iterator.go
```diff
@@ -0,0 +1,134 @@
...
+// DataItems is an iterator over the data items in the event.
+func (e Event) DataItems() iter.Seq2[DataItem, error] {
+	return func(yield func(DataItem, error) bool) {
+		header, err := e.Header()
+		if err != nil {
+			yield(DataItem{}, err)
+			return
+		}
+
+		idx := eventHeaderSize // Skip event header
```
ajwerner: nit: inline comments aren't sentences so `// skip event header`

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,407 @@
...
+		if err != nil {
+			return err
+		}
+		structure := irType.(*ir.StructureType)
+		for _, field := range structure.Fields {
+			err = enc.WriteToken(jsontext.String(field.Name))
+			if err != nil {
+				return err
+			}
+			err = d.encodeValue(enc, itemsByAddress, field.Type, data[field.Offset:field.Offset+uint32(field.Type.GetByteSize())])
```
ajwerner: does this cast do anything? isn't `GetByteSize()` already returning a `uint32`?

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,407 @@
...
+type dataItemWithAddrCounter struct {
+	dataItem    output.DataItem
+	addrCounter uint8
+}
+
+// Decode decodes the given event into the given writer.
+func (d *Decoder) Decode(event output.Event, out io.Writer) error {
+	var (
+		rootData  []byte
+		eventRoot *ir.EventRootType
```
ajwerner: uber-nit: I might call this `rootType` for symmetry with `rootData`

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,407 @@
...
+		ok        bool
```
ajwerner: pre-declaring this `ok` (and `err`) feels odd to me. You don't need it in outer scopes. Honestly, I'd encourage you to move the variable declarations closer to where they are first assigned. Like the reason to have rootData and its type pulled out is because you're only going to process it only after all the other items. To me it makes most sense to go right above the loop where you process all the items.

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,407 @@
...
+	}
+	err = enc.WriteToken(jsontext.BeginObject)
+	if err != nil {
+		return err
+	}
+	itemsByAddress := map[uint64]*dataItemWithAddrCounter{}
+	for item, err := range event.DataItems() {
+		if errors.Is(err, output.ErrFinishedIterating) {
+			break
+		}
```
ajwerner: If you just don't have this error and return from the iterator it'll stop; I don't think this error is doing anybody any favors.

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,407 @@
...
+				return err
+			}
+			return nil
+		}
+		pointedValue, ok := itemsByAddress[addr]
+		if !ok {
+			return errors.New("pointer not found in address pass map")
+		}
+		pointedValue.addrCounter++
+		if pointedValue.addrCounter > 1 {
```
ajwerner: I wonder if what we should do is keep it uniform by providing the address field and just skip the value if it forms a cycle. Later we may want to provide more information about the cycle -- but making the shape of the data non-uniform seems like it may in the future hurt.

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,407 @@
...
+		return errorUnimplemented
+	case *ir.EventRootType:
+		return errorUnimplemented
+	default:
+		return errors.New("invalid type")
+	}
+	return nil
+}
+
+func (d *Decoder) encodeBaseTypeValue(enc *jsontext.Encoder, irType *ir.BaseType, data []byte) error {
```
ajwerner, line 328: something interesting is that it seems to me like all variables are supposed to get wrapped in some sort of envelope dictating their type and value, though I may not be reading the document correctly: https://docs.google.com/document/d/10u7Yor4nGgpXv22Hsj-IRFMYHpxYeSdzv67tOfcO_Ug/edit?tab=t.0

pkg/dyninst/decode/decoder.go
ajwerner: Definitely not a lock -- my general assumption and I think the right assumption in Go should be that data structures are not thread safe unless explicitly marked as being thread safe. We don't need this one to be thread safe.

pkg/dyninst/decode/decoder.go
grantseltzer: Yes i'll handle the proper schema for upload in a followup.

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,451 @@
...
+	}
+	return decoder, nil
+}
+
+// dataItemWithAddrCounter is a data item with a counter of how many times it has been referenced.
+// This is used to avoid infinite recursion when encoding pointers.
+type dataItemWithAddrCounter struct {
+	dataItem       output.DataItem
+	referenceCount uint8
+}
```
ajwerner: This needs to go away

pkg/dyninst/decode/decoder.go
```diff
@@ -0,0 +1,451 @@
...
+		// The value is a data item with a counter of how many times it has been referenced.
+		// If the counter is greater than 1, we know that the data item is a pointer to another data item.
+		// We can then encode the pointer as a string and not as an object.
+		addressReferenceCount[typeAndAddr{
+			irType: uint32(item.Header().Type),
+			addr:   item.Header().Address,
+		}] = item
+	}
+
+	currentlyEncoding := map[typeAndAddr]struct{}{}
```
ajwerner, line 141: For future readers, it would probably be good to explain the scheme we're using to avoid serializing cycles in commentary

ajwerner (approved):
:ship::it:

ajwerner (approved):

PR #37704, opened 2025-06-05
[CXP-2248] Update hostname setup for the process and container checks in fargate

wiyu:
### What does this PR do?
- Fix how the hostname is set for the process and containers check when running in the Agent in AWS Fargate
- Fix the existing ECS Fargate tests to ensure the process-agent starts when running in AWS Fargate by explicitly disabling 
`DD_PROCESS_CONFIG_RUN_IN_CORE_AGENT_ENABLED` 

### Motivation
https://datadoghq.atlassian.net/browse/CXP-2248

The PR is validated through e2e tests.


tasks/build_tags.py
```diff
@@ -109,6 +110,7 @@
         "crio",
         "docker",
         "ec2",
+        "fargateprocess",
```
wiyu, line 113: Excluded from the heroku build

test/new-e2e/tests/process/fargate_test.go
```diff
@@ -46,8 +46,13 @@ func getFargateProvisioner(configMap runner.ConfigMap) provisioners.TypedProvisi
 func TestECSFargateTestSuite(t *testing.T) {
 	t.Parallel()
 	s := ECSFargateSuite{}
+
+	extraConfig := runner.ConfigMap{
+		"ddagent:extraEnvVars": auto.ConfigValue{Value: "DD_PROCESS_CONFIG_RUN_IN_CORE_AGENT_ENABLED=false"},
+	}
```
wiyu, line 53: This test's intention is to validate the process checks running in the process-agent, but it broke after we defaulted the process checks to run in the core agent. This explicitly sets the check to run in the `process-agent` to ensure the data is coming from the expected agent.

test/new-e2e/tests/process/fargate_test.go
```diff
@@ -96,20 +95,19 @@ func TestECSFargateCoreAgentTestSuite(t *testing.T) {
 func (s *ECSFargateCoreAgentSuite) TestProcessCheckInCoreAgent() {
 	t := s.T()
 
-	// Flush fake intake to remove any payloads which may have
-	s.Env().FakeIntake.Client().FlushServerAndResetAggregators()
```
wiyu, line 100: This isn't necessary since we've split the two tests into their own suite so the envs are fresh

pkg/process/checks/host_info.go
```diff
@@ -58,6 +58,15 @@ func CollectHostInfo(config pkgconfigmodel.Reader, hostnameComp hostnameinterfac
...
 
 func resolveHostName(config pkgconfigmodel.Reader, hostnameComp hostnameinterface.Component, ipc ipc.Component) (string, error) {
+	// Fargate is handled as an exceptional case (there is no concept of a host, so we use the ARN in-place).
+	if fargate.IsFargateInstance() {
+		hostname, err := fargate.GetFargateHost(context.Background())
+		if err == nil {
+			return hostname, nil
+		}
+		log.Errorf("failed to get Fargate host: %v", err)
+	}
```
kkhor-datadog: This seems to change the behavior of the process check in the process agent. Previously, we would first use the hostname specified in the agent config: https://github.com/DataDog/datadog-agent/blob/095afc879d107ce2e6573df970c5ca1129d958db/pkg/process/checks/host_info.go#L70-L72

Now we end up always overriding it with the Fargate task name.

wiyu: Good catch. I'll fix this and add some test coverage.

releasenotes/notes/process-checks-fix-fargate-hostname-9c00191b888ad634.yaml
```diff
@@ -0,0 +1,3 @@
+fixes:
+  - |
+    Fix how the Live Process and Live Containers sets the hostname when running in the Agent when running in AWS Fargate
```
michaelcretzman:
```suggestion
    Fix how the Live Process and Live Containers sets the hostname when running in an Agent that is running in AWS Fargate
```

michaelcretzman (approved):
Approved with one suggestion.

kkhor-datadog (approved):

chouetz (approved):

PR #37708, opened 2025-06-05
Add remote certificate collection to the Windows Certificate Store integration

mrafi97:
### What does this PR do?
This PR adds Remote Certificate collection to the Windows Certificate Store integration. This will allow users to specify a server and a username and password to use to login to the server. If a username and password is not provided, the Agent will attempt to access the server with its own credentials.

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1446

Created an e2e test that creates two hosts, one with the agent installed and another that the agent host will connect to get certificates. The certificate host has a test certificate added to its `MY` certificate store. The Agent host has the following configuration set on the `windows_certificate.d\conf.yaml` file to collect that certificate:
```
init_config:
instances:
  - certificate_store: MY
    server: <SERVER>
    username: <USERNAME>
    password: <PASSWORD>
```
The e2e test succeeds if the Agent is able to connect to the remote host and collect metrics with the expected tags.


pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
```diff
@@ -397,3 +502,11 @@ func getAttributeTypeName(oid string) string {
...
 	}
 }
+
+var netAddConnection = func(remoteName, localName, password, username string) error {
+	return wNetAddConnection2(remoteName, localName, password, username)
+}
+
+var netCancelConnection = func(name string) error {
+	return wNetCancelConnection2(name)
+}
```
iglendd: Why cannot we call these functions directly?

pkg/collector/corechecks/system/windowscertificate/windows_certificate_wnet.go
```diff
@@ -0,0 +1,100 @@
...
+	Scope       uint32
+	Type        uint32
+	DisplayType uint32
+	Usage       uint32
+	localName   *uint16
+	remoteName  *uint16
+	comment     *uint16
+	provider    *uint16
+}
+
```
iglendd: It looks like we already have (partial) implementation of that https://github.com/DataDog/datadog-agent/blob/4ecfddab199edcf035d86ea1a71ab9b743d61124/pkg/collector/corechecks/system/disk/diskv2/disk_windows.go#L115. It would be good to unify both implementations in winutil module

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
```diff
@@ -147,9 +152,24 @@ func (w *WinCertChk) Run() error {
...
+		if err != nil {
+			return err
+		}
+		serverTag = "server:" + w.config.Server
+	} else {
+		certificates, err = getCertificates(w.config.CertificateStore, w.config.CertSubjects)
+		if err != nil {
+			return err
+		}
+		hostname, err := os.Hostname()
```
iglendd, line 169: With this arrangement one cannot tell (without comparing with host tag) was it local certificate store or not (from metrics alone). I wonder if an alternative(s) would be "better" (alternative drop server tag, have server tag name "local", have additional tag)

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
```diff
@@ -203,25 +223,89 @@ func getCertificates(store string, certFilters []string) ([]*x509.Certificate, e
...
+			log.Errorf("Error canceling connection: %v", err)
 		}
 	}()
 
+	log.Debugf("Opening remote registry on %s", server)
+
+	// After establishing a connection to the remote server, we open its Local Machine registry key
+	remoteRegKey, err = registry.OpenRemoteKey(server, registry.LOCAL_MACHINE)
+	if err != nil {
+		log.Errorf("Error opening remote registry key: %v", err)
```
iglendd: I would add a little bit more context like server name and the same link to the docs to simplify troubleshooting.

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
```diff
@@ -203,25 +223,89 @@ func getCertificates(store string, certFilters []string) ([]*x509.Certificate, e
...
+		return nil, err
+	}
+	log.Debugf("Remote registry opened successfully")
+	defer remoteRegKey.Close()
+
+	// Once the remote registry is opened, we use its handle to open the registry key of the certificate store
+	certStoreKey, err = registry.OpenKey(remoteRegKey, registryPath, registry.READ)
+	if err != nil {
+		log.Errorf("Error opening %s registry key: %v", registryPath, err)
```
iglendd: I would add a little bit more context like server name and the same link to the docs.

michaelcretzman (approved):

chouetz (approved):
ok for `agent-devx` files

pkg/collector/corechecks/system/windowscertificate/windows_certificate_wnet.go
mrafi97: Makes sense, I think in this case I can move these functions to the winutil module and then revisit the disk check in a separate PR for simplicity.

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
mrafi97: @iglendd something like the following? 
```
log.Errorf("Error opening remote registry key for server %s: %v For more information see, https://learn.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regconnectregistryw#remarks", server, err)
```

iglendd (approved):
Looks good

pkg/util/winutil/wnet.go
```diff
@@ -0,0 +1,124 @@
...
+			return fmt.Errorf("failed to convert username to UTF16: %w", err)
+		}
+	}
+	rc, _, err := procWNetAddConnection2W.Call(
+		uintptr(unsafe.Pointer(&netResource)),
+		uintptr(unsafe.Pointer(_password)),
+		uintptr(unsafe.Pointer(_username)),
+		uintptr(flags),
+	)
+	if rc != 0 {
```
julien-lebot: **✏️ suggestion:** ‏ use `windows.ERROR_SUCCESS` instead of 0

test/new-e2e/tests/windows/windows-certificate/windows_certificate_remote_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+	config := `
+init_config:
+instances:
+  - certificate_store: MY
+    server: %s
+    username: %s
+    password: %s`
+
+	checkConfig := fmt.Sprintf(config, certificateHost.HostOutput.Address, TestUser, TestPassword)
```
julien-lebot: **❓ question:** ‏ Does go validate the `Sprintf` string if it's in a variable and not as a string constant? Would it make sense to move the variable to a string constant since it's not used anywhere else?

julien-lebot (commented):
Left some minor comments

pkg/util/winutil/wnet.go
mrafi97: From [Microsoft's docs](https://learn.microsoft.com/en-us/windows/win32/api/winnetwk/nf-winnetwk-wnetaddconnection2w#return-value) the function returns NO_ERROR upon success so I think `windows.NO_ERROR` would be better here.

test/new-e2e/tests/windows/windows-certificate/windows_certificate_remote_test.go
mrafi97: Do you mean moving the `config` variable to const?
```
const (
	TestUser     = "TestUser"
	TestPassword = "Test1234#"
	config       = `
init_config:
instances:
  - certificate_store: MY
    server: %s
    username: %s
    password: %s`
)
```

julien-lebot: No I meant like this:
```
checkConfig := fmt.Sprintf(`
init_config:
instances:
  - certificate_store: MY
    server: %s
    username: %s
    password: %s`, certificateHost.HostOutput.Address, TestUser, TestPassword)
```

pkg/util/winutil/wnet.go
julien-lebot: Yes, the same goes for WNetCancelConnection2W

PR #37710, opened 2025-06-05
[serverless-init] Fix flushing error logs

nhulston:
### What does this PR do?

Fixes the `ChannelWriter.Write` method to return the full byte count after flushing `stderr`, preventing the writer from retrying the same payload indefinitely.

### Motivation

Bug introduced in https://github.com/DataDog/datadog-agent/pull/37551/

Tested manually


duncanista (approved):

cmd/serverless-init/log/channel_writer.go
```diff
@@ -37,7 +37,7 @@ func (cw *ChannelWriter) Write(p []byte) (n int, err error) {
 	// Flush full stacktrace without splitting by new line
 	if cw.IsError {
 		cw.sendPayload(p)
```
apiarian-datadog, line 39: can we confirm that whoever reads the data from the `cw.Channel` that ultimately gets the payload isn't going to run into any kind of bytes corruption issues? (or we can clone the buffer before wending it along)

apiarian-datadog (approved):

PR #37714, opened 2025-06-06
[ACIX-882][BARX-847] Added retry to codesign

CelianR:
### What does this PR do?

We have failures using codesign due to "timestamp service not available" ([example](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/939700971#L2159)). Added retries to reduce flakiness.

This error happen on only few jobs and only few files for each job, adding 5 retries will probably reduce heavily the flakiness.

[Example job](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/971860187#L2182).

### Additional Notes
Also covers BARX-847.

### DIFF
--- a/omnibus/config/software/datadog-agent-finalize.rb
+++ b/omnibus/config/software/datadog-agent-finalize.rb
@@ -203,11 +203,11 @@ build do
 
             if code_signing_identity
                 # Codesign everything
-                command "find #{install_dir} -type f | grep -E '(\\.so|\\.dylib)' | xargs -I{} codesign #{hardened_runtime}--force --timestamp --deep -s '#{code_signing_identity}' '{}'"
-                command "find #{install_dir}/embedded/bin -perm +111 -type f | xargs -I{} codesign #{hardened_runtime}--force --timestamp --deep -s '#{code_signing_identity}' '{}'"
-                command "find #{install_dir}/embedded/sbin -perm +111 -type f | xargs -I{} codesign #{hardened_runtime}--force --timestamp --deep -s '#{code_signing_identity}' '{}'"
-                command "find #{install_dir}/bin -perm +111 -type f | xargs -I{} codesign #{hardened_runtime}--force --timestamp --deep -s '#{code_signing_identity}' '{}'"
-                command "codesign #{hardened_runtime}--force --timestamp --deep -s '#{code_signing_identity}' '#{install_dir}/Datadog Agent.app'"
+                command "find #{install_dir} -type f | grep -E '(\\.so|\\.dylib)' | xargs -I{} tools/ci/retry.sh codesign #{hardened_runtime}--force --timestamp --deep -s '#{code_signing_identity}' '{}'"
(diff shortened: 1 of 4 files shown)


omnibus/config/software/datadog-agent-finalize.rb
```diff
@@ -202,12 +202,15 @@
             end
 
             if code_signing_identity
+                # Sometimes the timestamp service is not available, so we retry
+                codesign = "../tools/ci/retry.sh codesign"
```
Pythyu, line 206: ❓ Question :
Is there a way to access `retry.sh` without relative path ? / or do it without bash
Also anyway we would we still be able to run it locally with code_signing if we really want to ? (not a big deal if not)

alopezz, line 206: Can we try not to do this with bash and adding this somewhat hidden dependency on a far-away shell script?

omnibus/config/software/datadog-agent-finalize.rb
```diff
@@ -202,12 +202,15 @@
...
+
                 # Codesign everything
-                command "find #{install_dir} -type f | grep -E '(\\.so|\\.dylib)' | xargs -I{} codesign #{hardened_runtime}--force --timestamp --deep -s '#{code_signing_identity}' '{}'"
-                command "find #{install_dir}/embedded/bin -perm +111 -type f | xargs -I{} codesign #{hardened_runtime}--force --timestamp --deep -s '#{code_signing_identity}' '{}'"
-                command "find #{install_dir}/embedded/sbin -perm +111 -type f | xargs -I{} codesign #{hardened_runtime}--force --timestamp --deep -s '#{code_signing_identity}' '{}'"
-                command "find #{install_dir}/bin -perm +111 -type f | xargs -I{} codesign #{hardened_runtime}--force --timestamp --deep -s '#{code_signing_identity}' '{}'"
-                command "codesign #{hardened_runtime}--force --timestamp --deep -s '#{code_signing_identity}' '#{install_dir}/Datadog Agent.app'"
+                command "find #{install_dir} -type f | grep -E '(\\.so|\\.dylib)' | xargs -I{} #{codesign} #{hardened_runtime}--force --timestamp --deep -s '#{code_signing_identity}' '{}'", cwd: Dir.pwd
```
chouquette, line 209: out of curiosity, is `cwd: Dir.pwd` required? It feels like that would already be the default

chouquette (approved):

omnibus/config/software/datadog-agent-finalize.rb
CelianR: Otherwise we are not in the agent dir but the build/omnibus dir so it wasn't working properly

omnibus/config/software/datadog-agent-finalize.rb
CelianR: This script comes from [here](https://github.com/DataDog/datadog-agent/blob/main/tools/ci/retry.sh), it should work locally too then

alopezz (approved):

omnibus/config/software/datadog-agent-finalize.rb
alopezz: Discussed offline.

PR #37716, opened 2025-06-06
update omnibus-ruby

chouquette:
### What does this PR do?

Update omnibus-ruby hash

### Motivation

The current version has a bug which causes systematic cache misses even after successfully downloading a valid cache

### DIFF
--- a/release.json
+++ b/release.json
@@ -2,7 +2,7 @@
     "base_branch": "main",
     "current_milestone": "7.68.0",
     "dependencies": {
-        "OMNIBUS_RUBY_VERSION": "34bf3ab7bc6ab0f1a08dee724590dcf0f76a443a",
+        "OMNIBUS_RUBY_VERSION": "49bffd5cbf5f4fc8a357afca145482111c8880ee",
         "JMXFETCH_VERSION": "0.49.7",
         "JMXFETCH_HASH": "470b1a19933efca4a48157b54ce48ab08a6ec6fb51f0b10337003dc428f50676",
         "MACOS_BUILD_VERSION": "master",


alopezz (approved):

release.json
```diff
@@ -2,7 +2,7 @@
     "base_branch": "main",
     "current_milestone": "7.68.0",
     "dependencies": {
-        "OMNIBUS_RUBY_VERSION": "34bf3ab7bc6ab0f1a08dee724590dcf0f76a443a",
+        "OMNIBUS_RUBY_VERSION": "49bffd5cbf5f4fc8a357afca145482111c8880ee",
```
aiuto, line 5: I am curious about this change. The other dependencies seems to include both version and hash. This just uses an opaque value.  I'm guessing this is a git commit hash. Just using a hash does not give an easy human verifiable path back to release versions. And, if we are not pointing to explicit releases, then we have a different problem.

alopezz: Both this `OMNIBUS_RUBY_VERSION` as well as `INTEGRATIONS_CORE_VERSION` point at commits, from [DataDog/omnibus-ruby](https://github.com/DataDog/omnibus-ruby) and [DataDog/integrations-core](https://github.com/DataDog/integrations-core) respectively.

These repositories don't have "explicit releases", but they do get tags on Agent release branches and this file points to those on release branches ([example](https://github.com/DataDog/datadog-agent/blob/a0b58897f677f754700c84eca9cdbf1706f483af/release.json#L5-L6)). But it's currently either/or, this `VERSION` thing is essentially a git ref.

PR #37718, opened 2025-06-06
Add support for exporter setting tls::insecure_skip_verify in DDOT

mackjmr:
### What does this PR do?
This PR adds support for dd exporter setting `tls::insecure_skip_verify` in DDOT.


cmd/otel-agent/config/agent_config.go
```diff
@@ -153,6 +153,7 @@ func NewConfigComponent(ctx context.Context, ddCfg string, uris []string) (confi
 	pkgconfig.Set("site", ddc.API.Site, pkgconfigmodel.SourceFile)
 
 	pkgconfig.Set("dd_url", ddc.Metrics.Endpoint, pkgconfigmodel.SourceFile)
+	pkgconfig.Set("skip_ssl_validation", ddc.ClientConfig.TLSSetting.InsecureSkipVerify, pkgconfigmodel.SourceFile)
```
truthbk: I _think_ we should only override this setting if it's set to `true` in the OTel configuration, otherwise the top-level setting can take precedence. Though I see how that could be regarded as a breaking the established pattern.

So, my concern here is that by doing this, we will get whatever `ddc.ClientConfig.TLSSetting.InsecureSkipVerify` is set to, no matter if it's explicitly set in the OTel configuration or not, right? So if nothing is set, I presume this would always default to the zero value (ie. `false`). What is weird with that is that if the top-level option is set to `skip_ssl_validation: true`, then that might not work as expected, and it would force customers using proxies to actually modify their OTel configuration to make it fly as opposed to "just working"....

I think I double-down on the idea that it only really makes sense to override if the setting is `true`. Feel free to challenge my take! :bow:

truthbk (commented):
Small thought. Let me know what you think.

cmd/otel-agent/config/agent_config.go
mackjmr: Only caveat I see is if user sets top level `skip_ssl_validation` to true, but explicitly sets it to false in the exporter config, then their config won't be taken into account. But I agree for the vast majority of cases, users that set the top level `skip_ssl_validation to true, and don't set anything in exporter config wouldn't expect the "default" false to override agent config. I'm fine with this approach, and updated in last commit.

mackjmr: There was actually some work upstream to differentiate between a default value vs user set value which would fix the caveat above, but it hasn't been implemented yet. 
Update: turns out it was implemented last week, see comment below.

truthbk: Indeed. It's tough to figure out the most sensible behavior. It would be so useful if we could tell if a setting is defined explicitly. We should maybe try to contribute upstream to try to figure out a way to make that distinction. 🤔

mackjmr: Appears like support was actually added last week: https://github.com/open-telemetry/opentelemetry-collector/pull/13044. Once the release is made, we can start using this optional type for config elements where we need to know if it's the default or a user set val.

mackjmr:
@truthbk Am I good to merge this ?

truthbk (approved):

PR #37719, opened 2025-06-06
Instrument docker pull for lint and tests on Windows

KevinFairise2:
### What does this PR do?

This PR do the pull of the Windows build image in a separate command that is instrumented with datadog-ci. That way we can know how much time is spent pulling the windows image


Ishirui (approved):
LGTM !
This new instrumented docker pull is not used in every place where we run stuff in the `WINBUILDIMAGE` but I assume that is on purpose ? (we don't really care about getting the same data 100 different times right ?)

KevinFairise2:
> LGTM ! This new instrumented docker pull is not used in every place where we run stuff in the `WINBUILDIMAGE` but I assume that is on purpose ? (we don't really care about getting the same data 100 different times right ?)

Yep actually the lint and tests are the one that increased a lot. So I targetted them, but actually the pull time should be pretty similar between all the jobs since they are all pulling the same image

PR #37720, opened 2025-06-06
Update image publish jobs for mutable tags to new method

kacper-murzyn:
### What does this PR do?

This PR updates the jobs run during the Agent RC build to update mutable tags based on the image indexes published with the immutable tag, instead of pushing whole new index again.

We have a new method in public-images which can just create a tag if we provide `IMG_TAG_REFERENCE` and `IMG_NEW_TAGS`

### Motivation

Simplify and improve image publish jobs.

Changes validated locally, there will be still a need to check in the actual RC build which is not easy to simulate without poisoning the git repo and repositories.

Instruction - example for `7.68.0-rc.1`
1. Run (or let Release Coordinator) the Agent RC build (7.68.0 would be the first affected),
2. Check jobs from `deploy_mutable_image_tags` stage. They all should pass,
3. Check produced RC images, for example on [dockerhub](https://hub.docker.com/r/datadog/agent). Tags starting with `7-rc` should be present and pointing to the respective `7.68.0-rc.1` indexes, for example: `7-rc-servercore-ltsc2019` should point to `7.68.0-rc.1-servercore-ltsc2019`, etc.


.gitlab/deploy_containers/deploy_containers_a7.yml
```diff
@@ -118,49 +113,35 @@ deploy_containers-a7-fips:
...
+  extends: .docker_publish_job_definition
+  stage: deploy_containers
   rules:
-    !reference [.on_rc]
-  variables:
-    VERSION: 7-rc
+    !reference [.manual_on_deploy_auto_on_rc]
+  dependencies: []
+  before_script:
+    - VERSION="$(dda inv agent.version --url-safe --pipeline-id $PARENT_PIPELINE_ID)" || exit $?
```
FlorentClarret:
```suggestion
    - VERSION="$(dda inv -- agent.version --url-safe --pipeline-id $PARENT_PIPELINE_ID)" || exit $?
```

nit, to stay on the safe side even if not strictly required in this case for now

.gitlab/deploy_containers/deploy_mutable_image_tags.yml
```diff
@@ -9,12 +9,128 @@
     - VERSION="$(dda inv agent.version --url-safe --pipeline-id $PARENT_PIPELINE_ID)" || exit $?
     - export IMG_TAG_REFERENCE=${AGENT_REPOSITORY}:${VERSION}${IMG_TAG_REFERENCE_SUFFIX}
 
-# Stable tags
+.deploy_mutable_image_tags-base_rc:
+  extends: .docker_publish_job_definition
+  stage: deploy_mutable_image_tags
+  dependencies: []
+  before_script:
+    - VERSION="$(dda inv agent.version --url-safe --pipeline-id $PARENT_PIPELINE_ID)" || exit $?
```
FlorentClarret:
```suggestion
    - VERSION="$(dda inv -- agent.version --url-safe --pipeline-id $PARENT_PIPELINE_ID)" || exit $?
```

FlorentClarret (approved):

.gitlab/deploy_containers/deploy_mutable_image_tags.yml
```diff
@@ -9,7 +9,109 @@
...
+
+deploy_mutable_image_tags-a7-fips-rc:
+  extends: .deploy_mutable_image_tags_base
+  stage: deploy_mutable_image_tags
+  rules:
+    !reference [.on_rc]
+  parallel:
+    matrix:
+      - IMG_TAG_REFERENCE_SUFFIX: ["-fips"]
+        IMG_NEW_TAGS: ["7-fips-rc"]
```
AliDatadog, line 84: Can we go for `7-rc-fips` instead ?

.gitlab/deploy_containers/deploy_mutable_image_tags.yml
```diff
@@ -9,7 +9,109 @@
...
+      - IMG_TAG_REFERENCE_SUFFIX: ["-fips-jmx"]
+        IMG_NEW_TAGS: ["7-fips-rc-jmx"]
```
AliDatadog, line 86: Same here `7-fips-rc-jmx` feels weird

.gitlab/deploy_containers/deploy_mutable_image_tags.yml
```diff
@@ -9,7 +9,109 @@
...
+
+deploy_mutable_image_tags-a7-fips_internal-rc:
+  extends: .deploy_mutable_image_tags_base
+  stage: deploy_mutable_image_tags
+  rules:
+    !reference [.on_internal_rc]
+  parallel:
+    matrix:
+      - IMG_TAG_REFERENCE_SUFFIX: ["-fips"]
+        IMG_NEW_TAGS: ["7-fips-rc"]
```
AliDatadog, line 111: Same here

.gitlab/deploy_containers/deploy_mutable_image_tags.yml
```diff
@@ -9,12 +9,128 @@
...
     - export IMG_TAG_REFERENCE=${AGENT_REPOSITORY}:${VERSION}${IMG_TAG_REFERENCE_SUFFIX}
 
-# Stable tags
+.deploy_mutable_image_tags-base_rc:
+  extends: .docker_publish_job_definition
+  stage: deploy_mutable_image_tags
+  dependencies: []
+  before_script:
+    - VERSION="$(dda inv agent.version --url-safe --pipeline-id $PARENT_PIPELINE_ID)" || exit $?
+    - export IMG_TAG_REFERENCE=${AGENT_REPOSITORY}:${VERSION}${IMG_TAG_REFERENCE_FLAVOR}${IMG_TAG_REFERENCE_WIN_OS}${IMG_TAG_REFERENCE_JMX}
```
AliDatadog: nit: can we try to further simplify that with maybe something like `${AGENT_REPOSITORY}:${VERSION}${SUFFIX}` ?
I also think `IMG_TAG_*` variable names can be shortened

.gitlab/deploy_containers/deploy_mutable_image_tags.yml
kacper-murzyn: I agree, but this is what we push now and I was hesitant to change it before checking first if it is used somewhere. I will ask around first and then I can check if it's safe enough

.gitlab/deploy_containers/deploy_mutable_image_tags.yml
kacper-murzyn: `IMG_TAG_REFERENCE` has to stay, as it is passed to `public-images`. I wanted to make those names explicit, but I guess we can use the shortening here.

.gitlab/deploy_containers/deploy_mutable_image_tags.yml
FlorentClarret: If it is, please do not modify it in this PR and do it in another one

.gitlab/deploy_containers/deploy_mutable_image_tags.yml
AliDatadog: I just think that this might be difficult to read and change for example:
```
      - IMG_TAG_REFERENCE_FLAVOR:
          - ""
          - "-servercore"
          - "-linux"
        IMG_TAG_REFERENCE_JMX:
          - ""
          - "-jmx"
```

Where as this would be easier to maintain:
```
- SUFFIX: ""
- SUFFIX: "-jmx"
- SUFFIX: "-servercore"
- SUFFIX: "-servercore-jmx"
- SUFFIX: "-linux"
- SUFFIX: "-linux-jmx"

```

kacper-murzyn: Good idea, I tried to change it to look more like that and it should work and is much easier to read.

.gitlab/deploy_containers/deploy_mutable_image_tags.yml
kacper-murzyn: It is used only in https://github.com/DataDog/image-vuln-scans. With this change everything would be actually simpler but it makes sense to split it.

AliDatadog (approved):

PR #37721, opened 2025-06-06
Refactor KMT Readme

mbertrone:
### What does this PR do?

This PR provides a comprehensive rewrite of the README for the Kernel Matrix Testing (KMT). The primary goal is to improve the documentation's clarity, structure, and overall user-friendliness for both new and existing users.

Key changes include:
A new Structure:
- A Table of Contents for quick navigation.
- Separate Quick Start guides for local, remote, and mixed (local + remote) VM setups.
- A Core Concepts section to explain fundamental ideas like stacks and vmsets.
- Command Reference with dedicated sections for each kmt task.

### Motivation

Provide to developers a better User Experience when using this documentation.

Documentation changes only

### DIFF
--- a/tasks/kernel_matrix_testing/README.md
+++ b/tasks/kernel_matrix_testing/README.md
@@ -5,6 +5,7 @@
 *   [Quick Start](#quick-start)
     *   [Quick Start for Local VMs](#quick-start-for-local-vms)
     *   [Quick Start for Remote VMs](#quick-start-for-remote-vms)
+    *   [Quick Start for Local and Remote VMs](#quick-start-for-local-and-remote-vms)
 *   [Core Concepts](#core-concepts)
     *   [Stacks](#stacks)
     *   [VMsets File](#vmsets-file)
@@ -173,6 +174,49 @@ When you are finished, destroy the stack to terminate the EC2 instance.
 
 ```bash
 dda inv -e kmt.destroy-stack --stack=my-remote-stack
+```
+
+### Quick Start for Local and Remote VMs
+
+This guide shows how to run a mixed stack with both local and remote VMs.
+
+**1. Initialize KMT for a Full Setup** (only needs to be done once)
+
+To run both local and remote VMs, you need to download VM images and configure an SSH key. The `kmt.init` command with an `--images` flag will handle both.
+
+```bash
+# This will download the Ubuntu 22.04 image and start the SSH key setup wizard
+dda inv -e kmt.init --images=ubuntu_22.04
+```
+
+**2. Create and Configure a Mixed Stack**
+
+Provide a list of both local and remote VMs to the `gen-config` command.
+
+```bash
+# This configures one local Ubuntu 22.04 VM and one remote x86_64 Ubuntu 22.04 VM
+dda inv -e kmt.gen-config --vms=ubuntu22-local-distro,x86-jammy-distro --init-stack --stack=my-mixed-stack
+```
+
+**3. Launch the Stack**
+
+The same `launch-stack` command will start the local VM and launch the remote EC2 instance.
+
+```bash
+dda inv -e kmt.launch-stack --stack=my-mixed-stack
+```
+> You may be prompted for your `sudo` password for the local VM setup. The remote VM will use your pre-configured SSH key.
+
+**4. Connect and Clean Up**
+
+The `status`, `ssh-config`, and `destroy-stack` commands work just like in the other scenarios, managing all VMs in the stack together.
+
+```bash
+# Check status of all VMs
+dda inv -e kmt.status --stack=my-mixed-stack
+
+# When finished, destroy all resources
+dda inv -e kmt.destroy-stack --stack=my-mixed-stack
 ```
 
 ## Core Concepts
--- a/tasks/kernel_matrix_testing/README.md
+++ b/tasks/kernel_matrix_testing/README.md
@@ -336,8 +336,6 @@ Generates or updates the `vmsets` configuration file for a stack, specifying whi
 dda inv -e kmt.gen-config --vms=<list> [--stack=<name>]
 ```
 
-The file can be generated incrementally. You can run `kmt.gen-config` and `kmt.launch-stack` multiple times to add more VMs to the same stack.
-
 **Common Flags:**
 
 -   `--vms=<list>`: A comma-separated list of VMs to add to the stack. See [Specifying VMs](#specifying-vms) below for the format.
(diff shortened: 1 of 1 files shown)


tasks/kernel_matrix_testing/README.md
```diff
@@ -1,425 +1,564 @@
...
-#### Configuring stack from a failed CI pipeline
+**1. Initialize KMT & Configure SSH Key** (only needs to be done once)
 
-A common use case is to configure the stack to replicate the jobs that failed in a CI pipeline. This can be easily done by using the `--from-ci-pipeline` flag, using the ID of the pipeline that failed in Gitlab as the argument:
+To create EC2 instances for remote VMs, KMT needs an SSH key to provision them. The `kmt.init` command will guide you through an interactive wizard to set up a default SSH key, so you don't have to provide it for every command. For a detailed explanation of the supported methods, see the [`kmt.config-ssh-key`](#kmtconfig-ssh-key) command documentation.
 
 ```bash
-dda inv -e kmt.gen-config --from-ci-pipeline=<pipeline-id> --use-local-if-possible --stack=demo-stack
+# Using --remote-setup-only is sufficient if you don't plan to use local VMs
+dda inv -e kmt.init --remote-setup-only
```
usamasaqib, line 139: I would call out that using local VMs is highly preferred for day to day use, since remote VMs launch metal instances which can be costly.

Remote VMs are recommended if a developer is running tests against a large number of VMs concurrently.

tasks/kernel_matrix_testing/README.md
```diff
@@ -1,425 +1,564 @@
...
-The `--images` parameter is required unless `--remote-setup-only` or `--all-images` is specified. You can use `dda inv -e kmt.ls` to see available images.
+**2. Create and Configure a Stack**
 
-This command will also ask you for the default SSH key to use, so it does not need to be provided every time. You can configure the SSH key again at any time running `dda inv -e kmt.config-ssh-key`. See more details below
+A stack is a collection of VMs. This command will create a new stack and configure it to run one local Ubuntu 22.04 VM.
 
-#### SSH key configuration
+```bash
+# --init-stack creates the stack if it does not exist
+# If --stack is not provided, a name is generated from your git branch
```
usamasaqib, line 77: I think the **Quick Start** should not include the `--stack` parameter. The pattern we want to encourage is to rely on KMT to automatically generate stack names based on git branches.

Explicitly specifying the stack name is a "power user" feature.

tasks/kernel_matrix_testing/README.md
```diff
@@ -1,425 +1,564 @@
...
 
 ```bash
-dda inv -e kmt.create-stack --stack=demo-stack
+dda inv -e kmt.status --stack=my-first-stack
 ```
 
-### Configure stack
-
-We will configure the stack to launch
+**5. Connect to the VM**
```
usamasaqib: I think this should not be listed in the **Quick Start** section, since this is more of a "power user" feature.

The normal way to interact with the VMs is to use `inv kmt.test` for running tests against the VMs.

tasks/kernel_matrix_testing/README.md
```diff
@@ -1,425 +1,564 @@
...
-You can generate a SSH config file to connect to the VMs using the following command:
+You can view the running VMs and their IP addresses. It may take a few moments for the VM to be assigned an IP.
 
 ```bash
-dda inv -e kmt.ssh-config > ~/.ssh/kmt_ssh_config
+dda inv -e kmt.status --stack=my-remote-stack
 ```
 
-This will generate a file `kmt_ssh_config` in the `~/.ssh` directory, which you can include in your config by using the line `Include ~/.ssh/kmt_ssh_config` in your `~/.ssh/config` file. Then you can then connect to the VMs using the following command:
+**5. Connect to the VM**
```
usamasaqib: Same as above. This should not be part of the **Quick Start**

tasks/kernel_matrix_testing/README.md
```diff
@@ -1,425 +1,564 @@
...
 ## Dependencies
 
-All dependencies are installed by the `kmt.init` command, as defined [below](#initializing-the-environment).
+All dependencies are installed by the `kmt.init` command. See [kmt.init](#kmtinit) for details.
 
-> For macOS users: Internet Sharing might need to be enabled for the networking to work properly. Enable it in System Settings -> General -> Sharing if you find problems with the VM networks. It does not matter which interface you enable it on, as long as it is enabled and the connection being shared is the one you use for Internet connection. We'd also appreciate it if you reported it to the eBPF platform team, as it's not clear still whether Internet Sharing needs to be enabled at some point or not.
+For macOS users, see the [macOS Network Issues](#macos-network-issues) section in troubleshooting.
 
-## Getting started
+## Quick Start
```
usamasaqib, line 53: I think the quick start section is too long. This section should just contain 1 recommended approach for using KMT. This means
- Initializing with `inv kmt.init`
- using local VMs
- relying on KMT to generate the stack name. So no `--stack` parameter is passed.
- Using `inv kmt.test` to interact with the VMs.


All other features are power-user features, which should be explained below in the documentation in detail.

tasks/kernel_matrix_testing/README.md
```diff
@@ -1,425 +1,564 @@
...
+```bash
+dda inv -e kmt.ls
+```
+
+Arguments:
+
+-   `--distro`: Only list distribution images (default).
+-   `--custom`: Only list custom kernel images.
+
+### `kmt.create-stack`
```
usamasaqib, line 326: Maybe we can remove this task, and remove it in subsequent PRs. This is almost never used by anyone.

tasks/kernel_matrix_testing/README.md
```diff
@@ -1,425 +1,564 @@
...
-- The stack to target specified with [--stack=<name>]
-- The list of VMS to generate specified by --vms=<list>. See [VMs list](#vms-list) for details on how to specify the list.
-- Optional paramenter `--init-stack` to automatically initialize the stack. This can be specified to automatically run the creating stack step.
-- Optional parameter `--new` to generate a fresh configuration file.
+-   `jammy-local-distro`, `distro-local-jammy`, `local-ubuntu22-distro` all resolve to a **local Ubuntu 22.04 distribution VM**.
+-   `amazon4.14-x86-distro`, `distro-x86_64-amazon4.14` all resolve to a **remote x86_64 Amazon Linux 2 (kernel 4.14) distribution VM**.
+-   `custom-arm-5.4`, `5.4-arm64-custom` all resolve to a **remote arm64 custom kernel 5.4 VM**.
 
-The file can be incrementally generated. This means a user may generate a vmset file. Launch it. Add more VMs. Launch them in the same stack.
+#### Incremental Configuration Examples
```
usamasaqib: This feature does not work correctly as of today.

mbertrone: Removing it from the readme. should we also cleanup the code (in a future PR), or this feature is already removed from the code?

tasks/kernel_matrix_testing/README.md
mbertrone: Good Point. I agree that kmt.test is the standard way to interact with the VMs, which is why I've highlighted it as the "primary way" in that section.

I chose to include the SSH instructions as an optional, secondary step because debugging is a very common part of the development workflow. My goal for the Quick Start is to give developers the essential information to be productive right away, and that includes knowing how to inspect the environment if a test fails.

Since the section is clearly marked as optional, it shouldn't distract from the main workflow.

tasks/kernel_matrix_testing/README.md
mbertrone: That's a good point. I agree that the Quick Start section should be as streamlined as possible for new users.

My goal was to present a complete, end-to-end example of the most common development workflow.

Based on your feedback, I've simplified the examples in the guide to better reflect the "happy path." I've removed the explicit `--stack` parameter to encourage the default behavior of using the git branch name.

The Quick Start for Local VMs now presents the following essential steps for a typical session:

- kmt.init
- kmt.gen-config
- kmt.launch-stack
- kmt.status
- kmt.test (with SSH as an optional step for debugging)
- kmt.destroy-stack

I believe this structure will be a good compromise.

tasks/kernel_matrix_testing/README.md
mbertrone: Let me keep the documentation for now. I'll open a PR to remove it both from README and the code. I'll post the PR on the internal channel asking if anyone has concerns about this command being removed before merging it.

tasks/kernel_matrix_testing/README.md
usamasaqib: We would like to support this feature at some point, so it is not removed from the code.

tasks/kernel_matrix_testing/README.md
```diff
@@ -1,425 +1,546 @@
...
+```bash
+dda inv -e kmt.launch-stack
+```
+> Since this is a local VM, you may be prompted for your password to run some commands with `sudo`.
 
-In order to create EC2 instances, we need to specify which SSH key will be provisioned on them. As that key is usually always the same, we ask the user to provide it once and store it in the configuration file. This way, the user does not need to provide it every time they launch a stack (although it's possible to change it by using the argument `--ssh-key` with KMT tasks).
+**4. Check Status**
 
-The configuration wizards will ask for the SSH key to use. There are three methods supported:
+You can view the running VMs and their IP addresses.
```
usamasaqib: We can specify that this will only show VMs in the current stack.

tasks/kernel_matrix_testing/README.md
```diff
@@ -1,425 +1,546 @@
...
 
-In order to update the resources for launching VMs run:
+This command will also guide you through the [default SSH key configuration](#kmtconfig-ssh-key).
+
+### `kmt.update-resources`
+
+Updates the resources for launching VMs, such as VM images and tools.
 
 ```bash
 dda inv -e kmt.update-resources
```
usamasaqib, line 310: We should add an example usage of `--images` here as well to show that specific VM images can be downloaded with this task.

tasks/kernel_matrix_testing/README.md
```diff
@@ -1,425 +1,546 @@
...
 ```
 
-Updating will first destroy all running stacks, and then use checksums to decide which packages need to be updated from S3.
-If there is an error during update, original packages are restored from backup.
+This command will first destroy all running stacks, then use checksums to decide which packages need to be updated from S3. Original packages are restored from a backup if an error occurs.
```
usamasaqib: The backup restore feature has been removed since it took too much local disk space.
I think this sentence can be simplified as a Note saying

```
This command will first destroy all running stacks before downloading the images.
```

tasks/kernel_matrix_testing/README.md
```diff
@@ -1,425 +1,546 @@
...
+Lists available VM images.
 
 ```bash
-dda inv -e kmt.create-stack [--stack=<name>]
+dda inv -e kmt.ls
 ```
 
-The developer can provide a name to associate with the `stack`.
-If no name is provided one is automatically generated from the current branch name.
+Arguments:
```
usamasaqib: We can remove this from the documentation. Only `--distro` is supported by KMT now. `--custom` is deprecated.

usamasaqib (approved):
Looks good! Provided a few more suggestions.

PR #37724, opened 2025-06-06
[AGENTRUN-359] Remove gRPC Gateway from Agent CMD API 

misteriaud:
### What does this PR do?
This PR removes the gRPC Gateway used in the Agent CMD API. To do so it:
- removes the use of `grpc-gateway` in `pkg/proto`
- updates the `protobuf.py` invoke task to remove gateway code generation

### Motivation
The gRPC to REST gateway was not used anymore, so there is no reason to keep it.

This change was manually tested by running the Agent and checking that the gRPC endpoint on CMD API is still running fine.


pgimalac (approved):
🚀

tasks/protobuf.py
```diff
@@ -11,15 +11,15 @@
...
-    'autodiscovery': (False, False),
+    'model/v1': (False),
+    'remoteconfig': (False),
+    'api/v1': (False),
+    'trace': (True),
+    'process': (False),
+    'workloadmeta': (False),
+    'languagedetection': (False),
+    'remoteagent': (False),
+    'autodiscovery': (False),
```
ofek: Heads up that this changes each values to a `bool`, if you want to maintain a tuple you have to do the following:

```suggestion
    'model/v1': (False,),
    'remoteconfig': (False,),
    'api/v1': (False,),
    'trace': (True,),
    'process': (False,),
    'workloadmeta': (False,),
    'languagedetection': (False,),
    'remoteagent': (False,),
    'autodiscovery': (False,),
```

ofek (approved):
🥇

comp/api/api/apiimpl/api_test.go
```diff
@@ -164,154 +140,3 @@ func TestStartBothServersWithObservability(t *testing.T) {
...
-	if grpc.gateway {
-		return &s{
-			body: "GRPC GATEWAY OK",
-		}, nil
-	}
-
-	return nil, fmt.Errorf("error")
-}
-
-func TestStartServerWithGrpcServer(t *testing.T) {
```
GustavoCaso, line 201: This test still relevant as we want to test that the gRPC component is being use when building the API component. I would say that we could remove the `gateway` attribute of the `grpcServer` struct and keep the test.

GustavoCaso, line 201: This test still relevant as we want to test that the gRPC component is being use when building the API component. I would say that we could remove the `gateway` attribute of the `grpcServer` struct and keep the test.

comp/api/api/apiimpl/api_test.go
```diff
@@ -164,154 +140,3 @@ func TestStartBothServersWithObservability(t *testing.T) {
...
-			fx.Annotate(&grpcServer{
-				grpcServer: true,
-				gateway:    false,
-			}, fx.As(new(grpc.Component))),
-		)))
-
-	assert.Error(t, errApp)
-}
-
-func TestStartServerWithoutGrpcServer(t *testing.T) {
```
GustavoCaso, line 274: This test still relevant as we want to test that the gRPC component is being use when building the API component. I would say that we could remove the `gateway` attribute of the `grpcServer` struct and keep the test.

comp/api/api/apiimpl/api_test.go
```diff
@@ -164,154 +140,3 @@ func TestStartBothServersWithObservability(t *testing.T) {
...
-	content, err = io.ReadAll(resp.Body)
-	assert.NoError(t, err)
-	t.Log(string(content))
-
-	// test the api routes grpc request to the grpc server
-	assert.Equal(t, http.StatusOK, resp.StatusCode)
-	assert.Equal(t, "GRPC SERVER OK", string(content))
-}
-
-func TestStartServerWithGrpcServerFailGateway(t *testing.T) {
```
GustavoCaso, line 256: This test we can remove 😄

GustavoCaso (changes requested):
Everything looks great. Left a few comments regarding a few test that we should keep. 

Also, we probably have to update the license file and remove the entry for `github.com/grpc-ecosystem/grpc-gateway/v2/runtime` also to remove the import from `internal/tools/proto/tools.go` and `tasks/install_tasks.py`

comp/api/api/apiimpl/api_test.go
misteriaud: Indeed ! I updated it in 22159998daa11b658eada90249393622b34f3031.

comp/api/api/apiimpl/api_test.go
misteriaud: Indeed ! I updated it in 22159998daa11b658eada90249393622b34f3031.

GustavoCaso (approved):
Amazing work. How satisfying is to delete stuff 😄

PR #37736, opened 2025-06-06
[dyninst/irgen] Support inlined functions

piob-io:
### What does this PR do?

Handle inlined functions when parsing DWARF.

Inlined functions are split between 3 types of DWARF entries:
 - subprogram, with inlined tag - abstract definition with its name, and names of params/variables
 - inlined_subroutine - instance of function, referencing the abstract definition including pc and location information
 - subprogram, out-of-line instance - present whenever function is inlined, but not everywhere, one more instance of function, referencing the abstract definition including pc and location information

We combine all the information to produce ir.Subprograms and ir.Probes.

### Motivation

https://datadoghq.atlassian.net/browse/DEBUG-3998

### DIFF
--- a/pkg/dyninst/irgen/irgen.go
+++ b/pkg/dyninst/irgen/irgen.go
@@ -488,9 +488,8 @@ func populateEventsRootExpressions(probes []*ir.Probe, typeCatalog *typeCatalog)
 			}
 			event.Type = &ir.EventRootType{
 				TypeCommon: ir.TypeCommon{
-					ID: id,
-					// TODO: Give this a better name.
-					Name:     "ProbeEvent",
+					ID:       id,
+					Name:     fmt.Sprintf("Probe[%s]", probe.Subprogram.Name),
 					ByteSize: uint32(byteSize),
 				},
 				// TODO: Populate the presence bitset size and expressions.
--- a/pkg/dyninst/irgen/irgen.go
+++ b/pkg/dyninst/irgen/irgen.go
@@ -36,8 +36,6 @@ import (
 
 // TODO: Handle creating return events.
 
-// TODO: Handle inline subprograms.
-
 // TODO: Properly set up the presence bitset.
 
 // TODO: Support hmaps.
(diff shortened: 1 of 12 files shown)


pkg/dyninst/irgen/irgen.go
```diff
@@ -791,20 +847,99 @@ func processInlinedSubroutineEntry(root *rootVisitor, unit *dwarf.Entry, subrout
...
 		unit: unit,
-		sp: &inlinedSubprogram{
-			abstractOrigin: abstractOrigin,
-			ranges:         ranges,
-		},
+		sp:   sp,
 	}, nil
 }
 
+func (v *rootVisitor) processVariable(unit *dwarf.Entry, entry *dwarf.Entry,
```
ajwerner: uber-nit: I think it looks better when wrapping arguments to do it like:


```go
// If the arguments list post-wrapping fits on one line:
func (v *rootVisitor) processVariable(
    unit *dwarf.Entry, entry *dwarf.Entry, isParameter bool, parseLocations bool,
) (*ir.Variable, error) {

// If the argument list is long:
func (v *rootVisitor) processVariable(
    unit *dwarf.Entry,
    entry *dwarf.Entry,
    isParameter bool,
    parseLocations bool,
    subprogramPCRanges []ir.PCRange
) (*ir.Variable, error) {

// Though if there are repeated parameters of the same type I might combine them like:
func (v *rootVisitor) processVariable(
    unit, entry *dwarf.Entry,
    isParameter bool,
    parseLocations bool,
    subprogramPCRanges []ir.PCRange
) (*ir.Variable, error) {

```

pkg/dyninst/irgen/irgen.go
```diff
@@ -818,7 +953,7 @@ func (v *inlinedSubroutineChildVisitor) push(
 ) (childVisitor visitor, err error) {
 	switch entry.Tag {
 	case dwarf.TagInlinedSubroutine:
-		return processInlinedSubroutineEntry(v.root, v.unit, entry)
+		return processInlinedSubroutineEntry(v.root, v.unit, entry, false)
```
ajwerner: nit: in cockroach we had a style thing of inline commenting bare nils and booleans like:
```go
processInlinedSubroutineEntry(v.root, v.unit, entry, false /* outOfLineInstance */)
// or
const outOfLineInstance = false
processInlinedSubroutineEntry(v.root, v.unit, entry, outOfLineInstance)
```

ajwerner (approved):
neat

pkg/dyninst/irgen/irgen.go
piob-io: Thanks, yeah, this is much better (if only linter would do it for me...). Wrapped some long calls too.

pkg/dyninst/irgen/irgen.go
piob-io: Done

PR #37742, opened 2025-06-06
APM: wait to shutdown trace-agent components

ajgajg1134:
Wait to shutdown until we've processed everything - otherwise we can lose data during the shutdown process and even panic due to [trying to use the obfuscation cache concurrently while it's being closed](https://github.com/outcaste-io/ristretto/issues/13) 

### What does this PR do?
Add coordination so that we wait for active `Process` calls to complete before shutting down all the rest of the agent components.

### Motivation
Ideally we don't want to lose data, and we definitely don't want to panic (even if it only happens during shutdown)

New unit test here catches this failure quite well

This could potentially lengthen the shutdown process, however this is in an effort to reduce lost data / panics so it seems worthwhile.


releasenotes/notes/APMWaitShutdown-d35a8b0cedec0fe8.yaml
```diff
@@ -0,0 +1,11 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+fixes:
+  - |
+    APM: Fix an issue where the trace-agent could panic during shutdown trying to obfuscate a sql payload.
```
jhgilbert:
```suggestion
    APM: Fix an issue where the trace-agent could panic during shutdown trying to obfuscate a SQL payload.
```

jhgilbert (approved):
Approved with a minor suggestion for style, thanks!

ichinaski (approved):

PR #37744, opened 2025-06-06
[dyninst] Testing expansion

piob-io:
### What does this PR do?

Share probe configs across tests.
Add snapshot test for compiler, dropping old, temporary test.
Merge simple and events_simple testprogs, they end up being almost the same.
Fix arm cfa offset calculation, exposed by more test cases.

### Motivation


pkg/dyninst/compiler/testdata/snapshot/simple.arch=amd64,toolchain=go1.24.3.c
```diff
@@ -0,0 +1,186 @@
...
+		SM_OP_EXPR_PREPARE, 
+		SM_OP_RETURN, 
+		SM_OP_EXPR_SAVE, 0x01, 0x00, 0x00, 0x00, 0x08, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 
+		SM_OP_RETURN, 
+
+	// 0xab: ProcessEvent[Probe[main.inlined]@491c66]
+		SM_OP_PREPARE_EVENT_ROOT, 0x2a, 0x00, 0x00, 0x00, 0x09, 0x00, 0x00, 0x00, 
+		SM_OP_CALL, 0x9b, 0x00, 0x00, 0x00, // ProcessExpression[Probe[main.inlined]@0x491c66.expr[0]]
+		SM_OP_RETURN, 
+		SM_OP_ILLEGAL,
```
ajwerner, line 150: nit: inject some commentary around the padding?

ajwerner (approved):
:ship: :it: 

Seeing the generated code makes it so clear we should find a way to just stick this into const data maps and stop compiling on the host

pkg/dyninst/compiler/testdata/snapshot/simple.arch=amd64,toolchain=go1.24.3.c
piob-io: Done

pkg/dyninst/compiler/snapshot_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/DataDog/datadog-agent/pkg/dyninst/object"
+	"github.com/DataDog/datadog-agent/pkg/dyninst/testprogs"
+	"github.com/DataDog/datadog-agent/pkg/util/safeelf"
+)
+
+var rewriteFromEnv = func() bool {
+	rewrite, _ := strconv.ParseBool(os.Getenv("REWRITE"))
+	return rewrite
+}()
+var rewrite = flag.Bool("rewrite", rewriteFromEnv, "rewrite the test files")
```
tylfin, line 32: I'm surprised there isn't some existing machinery for snapshots, golden files, and rewrites already?

piob-io: There definitely are, not sure how popular within datadog. However, we are looking for something really simple, while knowing exactly how we want to manage these files. I think a bar to use a framework has not been reached.

PR #37748, opened 2025-06-09
Allow RDS autodiscovery to discover all instances by specifying an empty tag list

sethsamuel:
### What does this PR do?
Allows RDS Autodisovery to discover all instances if the user specifies an empty array for the AD tag list.

### Motivation
Users may want to monitor all RDS instances without having to apply tags to trigger Autodiscovery. This can be useful for existing fleets where adding tags is a point of frustration, or for small fleets that can be easily monitored by a single agent.


.python-version
```diff
@@ -1 +1 @@
-3.12
```
eric-weaver: This looks unintentional?

eric-weaver (approved):

cswatt (approved):
release notes approved for docs 👍

PR #37752, opened 2025-06-09
Add ^ and $ to 'excluded_filesystems' and 'excluded_disks' config keys

jose-manuel-almaza:
### What does this PR do?
This PR updates the regular expression handling in Go to ensure consistent behavior with the previous Python implementation for 'excluded_filesystems' and 'excluded_disks' config keys.

### Motivation
In the original Python code, exclude patterns were applied [using re.match()](https://github.com/DataDog/integrations-core/blob/1412bd38c5f1a0196ce5fd005d841dab7b486964/disk/datadog_checks/disk/disk.py#L245), which implicitly anchors the pattern at the start of the string. Additionally, we were [appending a $ to the pattern](https://github.com/DataDog/integrations-core/blob/1412bd38c5f1a0196ce5fd005d841dab7b486964/disk/datadog_checks/disk/disk.py#L365), effectively enforcing that the entire string must match the given word (i.e., full-string match).

In Go, [regexp.MatchString()](https://pkg.go.dev/regexp#Regexp.MatchString) behaves different, it matches anywhere in the string unless explicitly anchored. As a result, patterns like `tmpfs$` were incorrectly matching strings like `devtmpfs` as we detected during our deployment on staging.

To ensure equivalent semantics, we now explicitly wrap all exclude patterns with `^` and `$`, so that MatchString() behaves like a full-string match (^pattern$). This guarantees that only exact matches (as originally intended) are excluded.

Following a TDD methodology when fixing a bug, the first thing added were tests that failed, and then the code that fixed them was added.


pkg/collector/corechecks/system/disk/diskv2/disk.go
```diff
@@ -279,9 +279,29 @@ func processRegExpSlices(slices [][]string, ignoreCase bool) ([]regexp.Regexp, e
 	return regExpList, nil
 }
 
+func processRegExpSlicesWholeWord(slices [][]string, ignoreCase bool) ([]regexp.Regexp, error) {
+	regExpList := []regexp.Regexp{}
+	for _, slice := range slices {
+		for _, val := range slice {
+			expr := fmt.Sprintf("^%s$", val)
```
jeremy-hanna, line 286: https://github.com/DataDog/integrations-core/blob/master/disk/datadog_checks/disk/disk.py#L365

it does not look like the python version adds the beginning anchor (`^`)

jose-manuel-almaza: You're right. But not needed because it's implicit.
We use Pattern.match [here](https://github.com/DataDog/integrations-core/blob/1412bd38c5f1a0196ce5fd005d841dab7b486964/disk/datadog_checks/disk/disk.py#L245) and according to its [documentation](https://docs.python.org/3/library/re.html#re.Pattern.match):

> If zero or more characters at the **beginning of string** match this regular expression, return a corresponding [Match](https://docs.python.org/3/library/re.html#re.Match). Return None if the string does not match the pattern; note that this is different from a zero-length match.

However, in the `go` version we need to explicitly indicate this to reproduce the same behavior as in `python` since [MatchString](https://pkg.go.dev/regexp#Regexp.MatchString) used [here](https://github.com/DataDog/datadog-agent/blob/ad0e415c98bf366d5739b86c240fb2c376b1f00d/pkg/collector/corechecks/system/disk/diskv2/disk.go#L93) behaves differently, returning `true` if it finds the string, regardless of the position.

> MatchString reports whether the string s **contains any match** of the regular expression re.

Added some tests to confirm this point.

jose-manuel-almaza: I've updated the PR description to make this clearer.

jeremy-hanna (approved):

PR #37797, opened 2025-06-10
remove unused Makefile.trace

chouquette:
### What does this PR do?

Remove unused `Makefile.trace` file

### Motivation

This appears to be a leftover from agent 5 which we don't need anymore


Makefile.trace
```diff
@@ -1,33 +0,0 @@
-# This Makefile is used within the release process of the Datadog Agent 5:
-# https://github.com/DataDog/dd-agent-omnibus/blob/fad8aeb/config/software/datadog-trace-agent.rb#L105
```
ichinaski, line 2: This Makefile is referenced [here](https://github.com/DataDog/dd-agent-omnibus/blob/fad8aeb/config/software/datadog-trace-agent.rb#L105), I don't think this is safe to be removed?

ichinaski (commented):
I'm not familiar with Agent 5 build/release process, but this file seems to be needed according to both the Makefile comment and the dd-agent-omnibus file it points to?

https://github.com/DataDog/dd-agent-omnibus/blob/fad8aeb/config/software/datadog-trace-agent.rb#L105

Makefile.trace
chouquette: If I understand things correctly, `dd-agent-omnibus` is our Agent 5 specific fork of omnibus (unlike datadog/omnibus-ruby which is our agent 6/7 fork of omnibus)
The [build script](https://github.com/DataDog/dd-agent-omnibus/blob/master/omnibus_build.sh#L12) in that repo seems to clone the `dd-agent` repo, so if we ever need to build agent 5, this commit will not impact that build

ichinaski: I'm honestly not familiar with the Omnibus build process for Agent 5. I can leave this PR approved from the trace-agent perspective, but could you ask @DataDog/agent-delivery for an approval as well? 🙏

ichinaski (approved):

FlorentClarret (approved):

PR #37812, opened 2025-06-10
[ AGNTLOG-116 ] - Keep registry entries alive even if they are not actively ingesting data

ddrthall:
### What does this PR do?
The auditor component of the Logs Agent serves to track and store offsets for each log source being tailed by the agent. These offsets record how much of a given log source the agent has processed, and are used to resume from the correct location when the agent is restarted. 

Currently, the auditor times out offset entries when it hasn't recorded any bytes from the relevant source in a set amount of time. This means that sources expected to be quiet for long intervals (23 hours by default) will find themselves resending all of their recorded data on agent restart. In order to correct this behavior, this PR seeks to alter how the auditor performs retention.

The auditor (synonymous with the term "registry" in the code) is integrated primarily into each type of logs agent "launcher". Some launchers, such as the listener launcher, do not have a concept of log offsets and therefore do not actually make use of the auditor. 

For the container, journald, and windows event launchers, the auditor reference has been shared with the equivalent tailers and each has been adjusted to notify the auditor on tailer start and stop. The auditor will never attempt to clean up an offset entry belonging to a tailer denoted as active through this flow, and will only begin applying the TTL cleanup once the configured number of hours have passed since the associated tailer has been stopped. 

The file launcher requires alternative logic. There is a (configurable) upper limit to the number of files that can be tailed, but as long as a file remains present on the host it should not be vulnerable to losing its offset by an errant TTL cleanup. In addition to the auditor start/stop detection on the file tailers, the file launcher and its dependencies have been adjusted to also record new "last seen" times for all files detected as tail-able. This is true regardless of whether the file in question ends up actually being attached to a tailer, and this "last seen" modification effectively resets the TTL cleanup counter.

Another subtle modification made to the auditor is the removal of the automatic auditor cleanup on start up. This allows the agent 5 minutes (the time between cleanup runs) to determine what tailers/files should be protected. The auditor still initiates cleanup automatically on shutdown, so shortlived agent runs are still guaranteed the ability to maintain their registries.

Registry files can be manually created with commands like 

> echo -n '{"Version":2,"Registry":{"file:/Users/ryan.hall/trial_log.txt":{"LastUpdated":"2025-06-10T18:00:50.184171Z","Offset":"6","TailingMode":"","IngestionTimestamp":1749578255928630000},"file:/Users/ryan.hall/trial_log_2.txt":{"LastUpdated":"2025-06-10T18:00:50.184171Z","Offset":"6","TailingMode":"","IngestionTimestamp":1749578255928630000}}}' > /opt/datadog-agent/run/registry.json

Last updated timestamps can be set to any time as part of this process. Some code modification was required to see certain cases live (such as modifying the TTL configuration from being X hours to X seconds in the the code).

Testing consisted of confirming
1. All existing files that are expected to be tailed are regularly updated in the registry.json, and all files not expected to be tailed are not updated.
2. All container/journald/windows event json entries retained as long as the entity being tailed exists/is configured, and are pruned when not.

Automated testing was added, including testing for cases wherein the number of files able to be tailed is less than the number of files configured to be tailed and test cases to guarantee the new auditor functions behave as expected with regards to the cleanup operation.


estherk15 (approved):

clarkb7 (approved):
seems appropriate for `windowsevent` tailer.
Do I understand correctly that the offset saved in the registry will be removed ~23 hours after its respective config is removed from the Agent configuration?

comp/logs/auditor/impl/auditor.go
```diff
@@ -253,6 +282,10 @@ func (a *registryAuditor) cleanupRegistry() {
 	expireBefore := time.Now().UTC().Add(-a.entryTTL)
 	for path, entry := range a.registry {
 		if entry.LastUpdated.Before(expireBefore) {
+			if a.tailedSources[path] {
```
soberpeach, line 285: question: Would it be more clear to this reverse this conditional? Something like
```go
if !a.tailedSources[path] {
    a.log.Debug("TTL"...)
    delete()
} else {
    a.log.Debugf("TTL for...")
}
```

Both ways work, I'm just wondering if it would be more clear to have the conditional be something like "if (condition) then do work" as opposed to "if condition, skip". I can see either way though

soberpeach (approved):

sblumenthal (approved):

PR #37813, opened 2025-06-10
[AGENTRUN-466] Strip cluster-agent binaries

pgimalac:
### What does this PR do?
Strip cluster-agent binaries.

### Motivation
Reduce cluster-agent binary size by 45MiB and docker image size by 30MiB.

### Additional Notes

I discussed this with container-platform and they considered it was probably just forgotten.
Other binaries in other images are stripped.

The unstripped binary is needed for fips compliance checks, so I had to keep an unstripped version in the artifacts.

### DIFF
--- a/.gitlab/binary_build/cluster_agent.yml
+++ b/.gitlab/binary_build/cluster_agent.yml
@@ -9,10 +9,14 @@
     - !reference [.retrieve_linux_go_deps]
   script:
     - dda inv -- check-go-version
-    - dda inv -- -e cluster-agent.build --major-version "$AGENT_MAJOR_VERSION" --strip-binary
+    - dda inv -- -e cluster-agent.build --major-version "$AGENT_MAJOR_VERSION"
+    # keep the unstripped binary for fips compliance checks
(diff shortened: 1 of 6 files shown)


chouquette (approved):

NouemanKHAL (approved):
Nice!

L3n41c (approved):
The figures are so significant that I’m wondering if we could have had the same result with `-ldflags="-s -w"` ?

I just checked the output of a [build job](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/975391572#L83) and it seems we only have `-ldflags="… -w"`

`-ldflags="-s"` is for:
```
 -s    disable symbol table
```
So, I’m wondering if it would have the same effect as `strip`?

pgimalac:
👋 Hey @L3n41c sorry I didn't see the comment before the PR got merged !

I believe strip is pretty much equivalent to using `-ldflags="-s -w"`, I initially tried using that (this is what we do it for `system-probe` and `cws-instrumentation`), but we have a fips job which requires the symbol table to be part of the binary to check for compliance, so we can't "strip" at build time
That's why I ended up building without stripping, making a copy of the unstripped binary, and stripping the original binary

PR #37815, opened 2025-06-10
fix(uprobes): Reduce noisy logging in uprobe attacher

### LINKED ISSUE #38198: [BUG] SYS-PROBE | WARN | (pkg/ebpf/uprobes/attacher.go:632 in handleProcessStart) | could not attach to process
**Agent Environment**
```
$ kubectl exec -it datadog-agent-xq2hj -- agent version
Defaulted container "agent" out of: agent, trace-agent, process-agent, system-probe, init-volume (init), init-config (init), seccomp-setup (init)
Agent 7.67.0 - Commit: bdf863ccc9 - Serialization version: v5.0.150 - Go version: go1.23.9
```

**Describe what happened:**
We upgraded the Agent from `7.66.1` to `7.67.0`
These kind of warnings started appearing in our agent logs:
```
2025-06-24 14:49:16 UTC | SYS-PROBE | WARN | (pkg/ebpf/uprobes/attacher.go:632 in handleProcessStart) | could not attach to process 661472: unexpected error: stat /host/proc/661472/root/runc: no such file or directory
```

No new warnings after upgrading.

Upgrade the DD Agent from `7.66.1` to `7.67.0.`

- Amazon Linux 2023 - EKS optimised.
- EKS 1.32
- Datadog Helm Chart 3.119.0

mbertrone:
### What does this PR do?
This pull request introduces two changes to reduce log noise originating from the eBPF uprobe attacher:
- Adds a Log Limiter: A rate limiter is added to the handleLibraryOpen and handleProcessStart functions in the UprobeAttacher. This prevents floods of warnings when the agent tries to attach probes to very short-lived processes, which is a common and usually benign race condition.
- Disables Detailed Logging: The EnableDetailedLogging flag is explicitly set to false for the USM/TLS uprobe attacher configuration. This prevents verbose debug logs, which are not intended for production, from being enabled.

### Motivation
In environments with high process churn the system-probe can generate a large volume of warnings

- code compiles
- existing tests are passing correctly

### Additional Notes
https://datadoghq.atlassian.net/browse/EBPF-756
https://datadoghq.atlassian.net/browse/EBPF-757

Fixes https://github.com/DataDog/datadog-agent/issues/38198

### DIFF
--- a/pkg/ebpf/uprobes/attacher.go
+++ b/pkg/ebpf/uprobes/attacher.go
@@ -625,7 +625,10 @@ func (ua *UprobeAttacher) shouldLogRegistryError(err error) bool {
 	}
 
 	var unknownErr *utils.UnknownAttachmentError
-	return errors.As(err, &unknownErr)
+	if errors.As(err, &unknownErr) {
+		return ua.attachLimiter.ShouldLog()
+	}
+	return false
 }
 
 // handleProcessStart is called when a new process is started, wraps AttachPIDWithOptions but ignoring the error
@@ -633,9 +636,7 @@ func (ua *UprobeAttacher) shouldLogRegistryError(err error) bool {
 func (ua *UprobeAttacher) handleProcessStart(pid uint32) {
 	err := ua.AttachPIDWithOptions(pid, false) // Do not try to attach to libraries on process start, it hasn't loaded them yet
 	if ua.shouldLogRegistryError(err) {
-		if ua.attachLimiter.ShouldLog() {
-			log.Warnf("could not attach to process %d: %v", pid, err)
-		}
+		log.Warnf("could not attach to process %d: %v", pid, err)
 	}
 }
 
@@ -650,9 +651,7 @@ func (ua *UprobeAttacher) handleLibraryOpen(libpath sharedlibraries.LibPath) {
 
 	err := ua.AttachLibrary(string(path), libpath.Pid)
 	if ua.shouldLogRegistryError(err) {
-		if ua.attachLimiter.ShouldLog() {
-			log.Warnf("could not attach to library %s (PID %d): %v", path, libpath.Pid, err)
-		}
+		log.Warnf("could not attach to library %s (PID %d): %v", path, libpath.Pid, err)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


guyarb (approved):

pkg/ebpf/uprobes/attacher.go
```diff
@@ -629,7 +633,9 @@ func (ua *UprobeAttacher) shouldLogRegistryError(err error) bool {
 func (ua *UprobeAttacher) handleProcessStart(pid uint32) {
 	err := ua.AttachPIDWithOptions(pid, false) // Do not try to attach to libraries on process start, it hasn't loaded them yet
 	if ua.shouldLogRegistryError(err) {
-		log.Warnf("could not attach to process %d: %v", pid, err)
+		if ua.attachLimiter.ShouldLog() {
```
brycekahle: Should we be checking `attachLimiter` inside of `shouldLogRegistryError` instead?

pkg/ebpf/uprobes/attacher.go
```diff
@@ -629,7 +633,9 @@ func (ua *UprobeAttacher) shouldLogRegistryError(err error) bool {
...
+			log.Warnf("could not attach to process %d: %v", pid, err)
```
brycekahle: I'm wondering about the value of these logs at all? If you are debugging a specific process, and they are limited, then you might miss the log anyways.

pkg/ebpf/uprobes/attacher.go
mbertrone: Good point. Addressed.

pkg/ebpf/uprobes/attacher.go
mbertrone: Great point. If I understood correctly, your concern is about the usefulness of those logs when part of the data is missing. For that use case, the `EnableDetailedLogging` flag should address the issue.

By the way, the flag is currently explicitly set to false and is not exposed in the agent configuration. I can open a separate PR to expose it, if that would be helpful.

val06 (approved):

pkg/ebpf/uprobes/attacher.go
brycekahle: if `EnableDetailedLogging` is set, it will still rate limit the logs, correct?

mbertrone: if `EnableDetailedLogging` is set, it bypasses any filtering and rate limiting logic, and prints all the logs.

https://github.com/DataDog/datadog-agent/blob/mbertrone/fix/uprobe-libcrypto-loglimiter/pkg/ebpf/uprobes/attacher.go#L623

brycekahle (approved):

PR #37825, opened 2025-06-10
[chore] remove upstream zaplogger dependency serializerexporter

jackgopack4:
### What does this PR do?
creates `ZapLoggerOtel` type for serializerexporter

### Motivation
Remove external dependency for logger in serializerexporter

added unit tests for functionality

slight duplication of code from upstream, but worth it to remove external dependency

### DIFF
--- a/comp/otelcol/otlp/components/exporter/serializerexporter/go.mod
+++ b/comp/otelcol/otlp/components/exporter/serializerexporter/go.mod
@@ -194,7 +194,7 @@ require (
 )
 
 require (
-	github.com/DataDog/datadog-agent/pkg/trace v0.65.1 // indirect
+	github.com/DataDog/datadog-agent/pkg/trace v0.65.1
 	go.opentelemetry.io/collector/consumer/xconsumer v0.127.0 // indirect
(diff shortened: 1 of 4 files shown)


mackjmr (approved):
LGTM but why is this needed / will we import this one in contrib and remove the on from contrib ?

jackgopack4:
> LGTM but why is this needed / will we import this one in contrib and remove the on from contrib ?

there were some comments from @songy23 on my PR in contrib https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/40556 that made me think this was a good idea. If you wanted to move zaplogger into agent repo entirely I would probably propose this as a separate package as opposed to one under serializerexporter

jackgopack4:
the alternative approach would be wait until that PR drops in contrib then just update this reference to the new `agentcomponents/zaplogger` package

songy23 (approved):
This should be temporary, in the later releases serializer should use agentcomponents/zaplogger instead

mackjmr:
makes sense, thanks 👍

jackgopack4:
have to double check if I need to close this PR or modify it to use pkg/datadog/agentcomponents

PR #37826, opened 2025-06-10
Set 'HostProcMountinfo' for Disk integration

jose-manuel-almaza:
### What does this PR do?
This PR introduces a new configuration option `proc_mountinfo_path` (defaulting to "`/proc/self/mounts`") and updates our disk check to call `disk.PartitionsWithContext(ctx, includeAll)` instead of the previous call to `disk.Partitions(includeAll)` from `gopsutil`

### Motivation
When migrating our disk checks from Python’s `psutil` to Go’s `gopsutil`, we discovered that by default:

- `psutil` reads mount information from the current process’s namespace (`/proc/self/mounts`).
- `gopsutil` reads from `/proc/1/mountinfo`, which doesn’t reflect the container’s view.

These differences led to inconsistent results inside containers (see [here](https://ddstaging.datadoghq.com/notebook/12587989/diskv2-check-comparison)). This PR introduces a new `proc_mountinfo_path` config (defaulting to "`/proc/self/mounts`") and switches all disk-check calls to `PartitionsWithContext` (instead of `Partitions`). By carrying the override in the context, we now have per-call control of the mountinfo source, making `gopsutil` behave like `psutil` by default, while still retaining its full fallback logic for other advanced scenarios.

### DIFF
--- a/cmd/agent/dist/conf.d/disk.d/conf.yaml.default
+++ b/cmd/agent/dist/conf.d/disk.d/conf.yaml.default
@@ -244,8 +244,8 @@ instances:
 
     ## @param proc_mountinfo_path - string - optional
     ## Path to the file from which to read mount information when enumerating disk partitions.
-    ## By default this points at /proc/self/mounts (the same source used by Python’s psutil).
-    ## If this setting is empty or does not exist, gopsutil’s built-in fallback
+    ## By default this points at /proc/self/mounts (the same source used by Python’s disk check).
+    ## If this setting is empty or does not exist, Go’s disk check built-in fallback
     ## logic will be used in order: /proc/1/mountinfo → /proc/self/mountinfo → /proc/mounts.
     #
     # proc_mountinfo_path: /proc/self/mounts
--- a/cmd/agent/dist/conf.d/disk.d/conf.yaml.default
+++ b/cmd/agent/dist/conf.d/disk.d/conf.yaml.default
@@ -241,3 +241,12 @@ instances:
     #     host: nfsserver
     #     share: /mnt/nfs_share
     #     type: nfs
+
+    ## @param proc_mountinfo_path - string - optional
+    ## Path to the file from which to read mount information when enumerating disk partitions.
+    ## By default this points at /proc/self/mounts (the same source used by Python’s psutil).
+    ## If this setting is empty or does not exist, gopsutil’s built-in fallback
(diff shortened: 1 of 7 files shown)


pgimalac (approved):

cmd/agent/dist/conf.d/disk.d/conf.yaml.default
```diff
@@ -241,3 +241,12 @@ instances:
     #     host: nfsserver
     #     share: /mnt/nfs_share
     #     type: nfs
+
+    ## @param proc_mountinfo_path - string - optional
+    ## Path to the file from which to read mount information when enumerating disk partitions.
+    ## By default this points at /proc/self/mounts (the same source used by Python’s psutil).
```
GustavoCaso: Do we need to mention Python psutil here? It seems that it is the technical details that might not be relevant for people using the Go disk version

pkg/collector/corechecks/system/disk/diskv2/disk.go
```diff
@@ -410,7 +413,11 @@ func (c *Check) configureIncludeMountPoint() error {
 }
 
 func (c *Check) collectPartitionMetrics(sender sender.Sender) error {
-	partitions, err := c.diskPartitions(c.instanceConfig.IncludeAllDevices)
+	ctx := context.Background()
+	if c.instanceConfig.ProcMountInfoPath != "" {
+		ctx = context.WithValue(ctx, common.EnvKey, common.EnvMap{common.HostProcMountinfo: c.instanceConfig.ProcMountInfoPath})
+	}
```
GustavoCaso, line 419: Super cool usage of context 👏

cmd/agent/dist/conf.d/disk.d/conf.yaml.default
jose-manuel-almaza: It's possible. I've included it in case we (or someone else, like support engineers) ever wonder why that configuration key exists in the future.

GustavoCaso: Since the configuration is also exposed to customers, I would remove it. We left a comment on the implementation https://github.com/DataDog/datadog-agent/pull/37826/files#diff-a553cd581968cebebcde0f68ac7e3f4faea0bf7be8a78f3f398d835a7e96b2b8R724 so we can always refer to that one 😄

GustavoCaso (approved):

rahulkaukuntla (approved):

PR #37829, opened 2025-06-10
Disable CWS and HA Agent e2e tests on FIPS Agent pipelines

misteriaud:
### What does this PR do?
[INCIDENT-39082]
This PR disable on FIPS pipelines the Agent e2e tests that are relying on remote-configuration.

### Motivation
Since remote-configuration is not available on FIPS Agent, this feature have been disable on #36988.
Disabling it started to make fail the e2e test relying on it.

### Additional Notes

We wait either for the tests to split so the core feature is tested without needing remote-configuration, or for remote-configuration to be available with the FIPS Agent.


CelianR (approved):
NIT: Could you add [INCIDENT-39082] in your PR description to track it?

PR #37832, opened 2025-06-10
APM: Reduce likelihood of receiver shutdown panic

ajgajg1134:
### What does this PR do?
Move our usage of waitgroup to cover the entire `process` function to ensure that the Stop() function doesn't close the `out` channel until we've processed all payloads

### Motivation
This should reduce the incidence of panics during shutdown

I ran this locally with a test tracer to ensure it still functions correctly and saw it was successful. I also attempted to recreate the `send after close` panic (before and after) and could not do so locally.

This does not fully resolve this issue, there is still a narrow window in which this issue can still occur. I have been unable to identify the underlying issue that is causing server.Shutdown() to not block until all handlers return, so this is a stop-gap measure to reduce incidence while continuing investigation.

The use of a wait group here for ever incoming trace payload will incur a very small performance cost.


ekump (commented):
LGTM FWIW

ichinaski (approved):

PR #37833, opened 2025-06-10
ignore uprobes for perf stats in userspace

brycekahle:
### What does this PR do?

Ignores uprobe for kprobe perf stats collection in ebpf check, rather than relying on the ebpf side, which seems to be unreliable.

### Motivation

Lots of logs like 
```
yyyy-MM-dd HH:mm:ss UTC | SYS-PROBE | WARN | (pkg/collector/corechecks/ebpf/probe/ebpfcheck/probe.go:538 in getKprobeMisses) | error in eBPF program while recording kprobe statistics for *: fd is not a perf event fd
```
with uprobe/uretprobe programs
Screenshot 2025-06-10 at 12 16 49 PM

### Additional Notes
https://datadoghq.atlassian.net/browse/EBPF-758

### DIFF
--- a/pkg/collector/corechecks/ebpf/c/runtime/ebpf-kern-user.h
+++ b/pkg/collector/corechecks/ebpf/c/runtime/ebpf-kern-user.h
@@ -3,6 +3,8 @@
 
 #include "ktypes.h"
 
+#define EBPF_CHECK_KPROBE_MISSES_CMD 0x70C14
+
 typedef struct {
     __u32 map_id;
     __u32 cpu;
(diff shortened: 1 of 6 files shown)


usamasaqib:
The root cause is that there is a race between when we build the map containing the programs to inspect and when we query the stats. Uprobes (and other programs in fact) may be removed in the meanwhile.

Pairing a function to remove mappings with the detach hook will reduce these logs. I can make a PR for doing that.

mbertrone (approved):
LGTM!

PR #37835, opened 2025-06-10
[NDINT-91] update flush timing to use ticker

nathalieabello:
### What does this PR do?
The variable `now` reflects the "ideal logical time" the ticker represents instead of the "real wall time".

### Motivation
Using time.Now() to determine the current time introduces potential drift and inconsistency (delays due to scheduling, GC pause, etc.). We should instead use the time value implied by the ticker.


jmw51798 (approved):
LGTM 😄

PR #37841, opened 2025-06-11
[DEBUG-3986] dyninst/actuator: add a stateful control layer

ajwerner:
This layer orchestrates the lifecycle of programs and how they map to processes. Most of the implementation and testing goes into the state machine and its updates. I suspect it could have a better implementation -- right now it's just a bunch of imperative case crushing. I tried to make up for that with a good deal of testing and testing infrastructure.

The testing comes in two main forms:
 - TestSnapshot: uses golden files in a yaml DSL to list the set of events that should happen and then reports the state transitions and effects
 - TestStateMachineProperties: uses random event generation to try to shake out interesting behaviors

Both of these tests rely on invariant checking of the state itself.

In addition to the state, there's the implementation of the Actuator which hooks up the state machine and the effects it generates to real dependencies.

One thing this change makes abundantly clear is the need to have fine-grained error reporting out of ir generation, and then separately a clearer interface for how to report these errors out into the diagnostics flows.


pkg/dyninst/actuator/settings.go
```diff
@@ -0,0 +1,132 @@
...
+// Note that the dispatcher should be thread-safe.
+type Dispatcher interface {
+
+	// RegisterProgram registers a program with the dispatcher.
+	RegisterProgram(program *ir.Program)
+
+	// UnregisterProgram unregisters a program from the dispatcher.
+	UnregisterProgram(program ir.ProgramID)
+
+	// HandleMessage handles a message from the dispatcher.
```
piob-io: From a registered program?

pkg/dyninst/actuator/state.go
```diff
@@ -0,0 +1,880 @@
...
+	baseEvent
+	programID       ir.ProgramID
+	compiledProgram *compiledProgram
+}
+
+type eventProgramCompilationFailed struct {
+	baseEvent
+	programID ir.ProgramID
+	err       error
+}
```
piob-io, line 157: Could IR generation and compilation be combined into a single asynchronous action, that does both, in sequence? They both feel like steps of "probe initialization"? Maybe program loading too?

pkg/dyninst/actuator/update.go
piob-io, line 1: process.go maybe? it's not just update, and I think process might be clearer?

pkg/dyninst/actuator/settings.go
```diff
@@ -0,0 +1,132 @@
...
+	"io"
+
+	"github.com/DataDog/datadog-agent/pkg/dyninst/compiler"
+	"github.com/DataDog/datadog-agent/pkg/dyninst/config"
+	"github.com/DataDog/datadog-agent/pkg/dyninst/ir"
+)
+
+type settings struct {
+	ringBufSize      int
+	dispatcher       Dispatcher
```
piob-io: Dispatcher wasn't too clear to me, before diving into code (as there is lots of dispatching going around with state machine). "Sink" maybe to indicate it has to do with dealing with output/results of the whole actuation? Or output or ProbeOutput or ProgramOutput?

piob-io (commented):
Neat tests!

pkg/dyninst/actuator/settings.go
ajwerner: Fixed.

pkg/dyninst/actuator/state.go
ajwerner: Done.

pkg/dyninst/actuator/update.go
ajwerner: Done.

pkg/dyninst/actuator/settings.go
ajwerner: Called it `MessageSink`.

piob-io (approved):

grantseltzer (approved):

PR #37845, opened 2025-06-11
[CWS] Sort replayed events

Gui774ume:
### What does this PR do?

This PR sorts events that are replayed when a new ruleset is loaded (or on startup).

### Motivation

This ensures that we evaluate execution events against our rule set in a consistent manner, regardless of how the execution was is captured (snapshot / runtime).

### DIFF
--- a/pkg/security/probe/probe_ebpf.go
+++ b/pkg/security/probe/probe_ebpf.go
@@ -17,6 +17,7 @@ import (
 	"os"
 	"path/filepath"
 	"slices"
+	"sort"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


pkg/security/probe/probe_ebpf.go
```diff
@@ -656,6 +657,29 @@ func (p *EBPFProbe) playSnapshot(notifyConsumers bool) {
...
+		if tsA.IsZero() {
+			return eventA.PIDContext.Pid < eventB.PIDContext.Pid
+		}
+
+		tsB := eventB.ProcessContext.ExecTime
+		if tsB.IsZero() {
+			return eventA.PIDContext.Pid < eventB.PIDContext.Pid
+		}
+
+		if tsA.Equal(tsB) {
```
paulcacheux: it's definitely a nitpick, but I'm wondering if
```
if tsA.IsZero() || tsB.IsZero() || tsA.Equal(tsB) {
  return eventA.PIDContext.Pid < eventB.PIDContext.Pid
}
```
wouldn't be simpler

paulcacheux (approved):

pkg/security/probe/probe_ebpf.go
Gui774ume: haha sure 👍🏻 Updating ...

paulcacheux (approved):

safchain (approved):

spikat (approved):

PR #37852, opened 2025-06-11
[BARX-1043] Update dca images publish jobs to use tag method for mutable tags instead of full index push

kacper-murzyn:
### What does this PR do?
This PR updates the jobs which publish mutable image tags for cws-instrumentation images to use new approach which does not push full new index, but instead creates a tag that points to the existing reference.

### Motivation
Unify the images publish mechanism.

To be properly tested with the 7.68.0 RC build.

Instruction (`7.68.0-rc.1` example):

1. Trigger (or let release coordinator) build for `7.68.0-rc.1`,
2. Check jobs in `deploy_dca` stage - they should all pass,
3. Verify that the rc tag published points to the `7.68.0-rc.1` tag in public repos, for example in [dockerhub](https://hub.docker.com/r/datadog/cluster-agent/tags)

### DIFF
--- a/.gitlab/deploy_dca/deploy_dca.yml
+++ b/.gitlab/deploy_dca/deploy_dca.yml
@@ -6,6 +6,7 @@ include:
 # DCA image tagging & manifest publication
 #
 
+# Basic flavor
 .deploy_containers-dca-base:
   extends: .docker_publish_job_definition
   stage: deploy_dca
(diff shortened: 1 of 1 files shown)


.gitlab/deploy_dca/deploy_dca.yml
```diff
@@ -21,38 +22,59 @@ include:
...
+    - VERSION="$(dda inv -- agent.version --url-safe --pipeline-id $PARENT_PIPELINE_ID)" || exit $?
+    - export IMG_TAG_REFERENCE=${CLUSTER_AGENT_REPOSITORY}:${VERSION}
+
 deploy_containers-dca:
   extends: .deploy_containers-dca-base
   rules: !reference [.on_deploy_manual_auto_on_rc]
 
 deploy_containers-dca-rc:
-  extends: .deploy_containers-dca-base
+  extends: .deploy_mutable_dca_tags-base
```
FlorentClarret:
```suggestion
  extends: .deploy_mutable_dca_tags-base
```

nit, someone will drop them in a follow up PR, adding extra noise to their PR

FlorentClarret (approved):

PR #36422, opened 2025-04-24
feat(e2e): Add capacity to rerun a failed test within a job

chouetz:
### What does this PR do?
Add the `rerun_fails` option to the new-e2e tests execution. As a consequence every failed test will be re-run inside of the job before triggering a failure.

### Motivation
It shoud decrease [the number of job rerun](https://app.datadoghq.com/ci/pipeline-executions?query=ci_level%3Ajob%20%40ci.pipeline.name%3ADataDog%2Fdatadog-agent%20%40git.branch%3Amain%20%40ci.job.name%3Anew-e2e-installer&agg_m=count&agg_m_source=base&agg_q=%40ci.pipeline.id&agg_q_source=base&agg_t=count&cipipeline_explorer_sort=time%2Cdesc&cols=%40git.branch%2C%40ci.status%2Ctimestamp%2C%40ci.pipeline.name%2C%40ci.stage.name%2C%40ci.job.name%2C%40duration%2C%40ci.pipeline.id%2C%40git.repository.name&fromUser=true&index=cipipeline&mode=sliding&saved-view-id=2643993&top_n=1000&top_o=top&viz=toplist&x_missing=true&start=1742891342809&end=1745483342809&paused=false), which takes more time as we need to get another executor, etc.
https://datadoghq.atlassian.net/browse/ACIX-589

### Describe how you validated your changes
Executed locally the following command: `dda inv -- -e new-e2e-tests.run`:
- First without any option, the result is `DONE 16 tests in 12.751s`
- Changed a value in `test/new-e2e/system-probe/vm-metrics/vm-metrics_test.go:43`: `DONE 16 tests, 1 failure in 14.438s`
- Execute with rerun `dda inv -- -e new-e2e-tests.run --rerun-fails 2`: `DONE 3 runs, 18 tests, 3 failures in 10.576s`
- Test with another value of rerun `dda inv -- -e new-e2e-tests.run --rerun-fails 0`: `DONE 16 tests, 1 failure in 8.203s`

### Possible Drawbacks / Trade-offs
Currently the resources will be probably released after each test run. I will deal with this in a follow-up PR

PR #37862, opened 2025-06-11
[ACIX-883] Auto-retry e2e tests in CI

Ishirui:
Note: this PR depends on #38227

### What does this PR do?

Following-up on ACIX-859 and #36422, try a new method for automatically rerunning failing e2e tests. This new method does not re-run tests that are already known as flaky and whose failure is accepted, contrary to what the previous method would do.

### Motivation

Increase CI reliability and diminish the impact (lost time) of flaky (but not yet marked as flaky) tests

TBD

Increased pipeline runtime + lost visibility on which tests are failing/flaky

### DIFF
--- a/.gitlab/e2e_install_packages/common.yml
+++ b/.gitlab/e2e_install_packages/common.yml
@@ -40,7 +40,7 @@
         END_MAJOR_VERSION: [7]
   script:
     - DATADOG_AGENT_API_KEY=$($CI_PROJECT_DIR/tools/ci/fetch_secret.sh $INSTALL_SCRIPT_API_KEY_ORG2 token ) || exit $?; export DATADOG_AGENT_API_KEY
-    - dda inv -- -e new-e2e-tests.run --targets $TARGETS --max-retries 3 --junit-tar "junit-${CI_JOB_ID}.tgz" ${EXTRA_PARAMS} --src-agent-version $START_MAJOR_VERSION --dest-agent-version $END_MAJOR_VERSION --test-washer
+    - dda inv -- -e new-e2e-tests.run --targets $TARGETS --junit-tar "junit-${CI_JOB_ID}.tgz" ${EXTRA_PARAMS} --src-agent-version $START_MAJOR_VERSION --dest-agent-version $END_MAJOR_VERSION --test-washer
 
 .new-e2e_script_upgrade_persisting_integrations:
   stage: e2e_install_packages
@@ -51,7 +51,7 @@
     E2E_LOGS_PROCESSING_TEST_DEPTH: 2 # We use a single test suite and run all the platforms test as subtest
   script:
     - DATADOG_AGENT_API_KEY=$($CI_PROJECT_DIR/tools/ci/fetch_secret.sh $INSTALL_SCRIPT_API_KEY_ORG2 token) || exit $?; export DATADOG_AGENT_API_KEY
-    - dda inv -- -e new-e2e-tests.run --targets $TARGETS --max-retries 3 --junit-tar "junit-${CI_JOB_ID}.tgz" ${EXTRA_PARAMS} --src-agent-version 7 --test-washer
+    - dda inv -- -e new-e2e-tests.run --targets $TARGETS --junit-tar "junit-${CI_JOB_ID}.tgz" ${EXTRA_PARAMS} --src-agent-version 7 --test-washer
 
 .new-e2e_rpm:
   stage: e2e_install_packages
@@ -62,4 +62,4 @@
     E2E_LOGS_PROCESSING_TEST_DEPTH: 2 # We use a single test suite and run all the platforms test as subtest
   script:
     - DATADOG_AGENT_API_KEY=$($CI_PROJECT_DIR/tools/ci/fetch_secret.sh $INSTALL_SCRIPT_API_KEY_ORG2 token) || exit $?; export DATADOG_AGENT_API_KEY
-    - dda inv -- -e new-e2e-tests.run --targets $TARGETS --max-retries 3 --junit-tar "junit-${CI_JOB_ID}.tgz" ${EXTRA_PARAMS} --test-washer
+    - dda inv -- -e new-e2e-tests.run --targets $TARGETS --junit-tar "junit-${CI_JOB_ID}.tgz" ${EXTRA_PARAMS} --test-washer
--- a/.gitlab/e2e_install_packages/common.yml
+++ b/.gitlab/e2e_install_packages/common.yml
@@ -40,7 +40,7 @@
         END_MAJOR_VERSION: [7]
   script:
     - DATADOG_AGENT_API_KEY=$($CI_PROJECT_DIR/tools/ci/fetch_secret.sh $INSTALL_SCRIPT_API_KEY_ORG2 token ) || exit $?; export DATADOG_AGENT_API_KEY
-    - dda inv -- -e new-e2e-tests.run --targets $TARGETS --junit-tar "junit-${CI_JOB_ID}.tgz" ${EXTRA_PARAMS} --src-agent-version $START_MAJOR_VERSION --dest-agent-version $END_MAJOR_VERSION --test-washer
+    - dda inv -- -e new-e2e-tests.run --targets $TARGETS --max-retries 3 --junit-tar "junit-${CI_JOB_ID}.tgz" ${EXTRA_PARAMS} --src-agent-version $START_MAJOR_VERSION --dest-agent-version $END_MAJOR_VERSION --test-washer
 
 .new-e2e_script_upgrade_persisting_integrations:
   stage: e2e_install_packages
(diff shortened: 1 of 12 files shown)


tasks/new_e2e_tests.py
```diff
@@ -195,18 +197,48 @@ def run(
...
+        if to_retry:
+            print(
+                color_message(
+                    f"Retrying {len(to_retry)} failed tests:\n- {'\n- '.join(f'{package} {test_name}' for package, test_name in sorted(to_retry))}",
+                    "yellow",
+                )
+            )
+
+            # Retry the failed tests only
+            affected_packages = {package.partition("/test/new-e2e/")[2] for package, _ in to_retry}
```
KevinFairise2: Maybe try os.path.relpath() or something else to compute path relative to `test/new-e2e`

Ishirui (commented):
Todo:
- Default retries to 0
- If retry logic enabled, pass "nth retry" and "max retries" to [`TearDownSuite`](https://github.com/DataDog/datadog-agent/blob/161f8c67d580706dfcc2ecca4a84941304ed3750/test/new-e2e/pkg/e2e/suite.go#L649) method to prevent tearing down the infra on test failure if there are going to be further retries
> Try to make it so that if we don't pass any `--max-retries` flag, the behavior does not change at all 
- In order to do pass these env vars, add it [here](https://github.com/DataDog/datadog-agent/blob/91f63ee656e33684f62168757ff9c1fe14f037cb/test/new-e2e/pkg/runner/parameters/store_env.go#L16) so it gets loaded as a global go var

.gitlab/e2e_install_packages/common.yml
```diff
@@ -32,7 +32,7 @@
         END_MAJOR_VERSION: [7]
   script:
     - DATADOG_AGENT_API_KEY=$($CI_PROJECT_DIR/tools/ci/fetch_secret.sh $INSTALL_SCRIPT_API_KEY_ORG2 token ) || exit $?; export DATADOG_AGENT_API_KEY
-    - dda inv -- -e new-e2e-tests.run --targets $TARGETS --junit-tar "junit-${CI_JOB_ID}.tgz" ${EXTRA_PARAMS} --src-agent-version $START_MAJOR_VERSION --dest-agent-version $END_MAJOR_VERSION --test-washer
+    - dda inv -- -e new-e2e-tests.run --targets $TARGETS --max-retries 3 --junit-tar "junit-${CI_JOB_ID}.tgz" ${EXTRA_PARAMS} --src-agent-version $START_MAJOR_VERSION --dest-agent-version $END_MAJOR_VERSION --test-washer
```
Ishirui: I added retries to all usages of the `new-e2e-tests.run` task, but maybe I should only add it to the one in the main `e2e.yml` file ? Not sure if we should enable retries these other instances of e2e tests.

tasks/new_e2e_tests.py
```diff
@@ -429,6 +465,28 @@ def write_result_to_log_files(logs_per_test, log_folder, test_depth=1):
             f.write("".join(logs))
 
 
+def _create_test_selection_regex(test_names: list[str]) -> str:
```
Ishirui: Not sure if this was the best place to put this, maybe I should move it to a directory under `libs/` or something ?

test/new-e2e/pkg/e2e/suite.go
```diff
@@ -664,6 +665,26 @@ func (bs *BaseSuite[Env]) TearDownSuite() {
...
+
+	remainingRetries, err := strconv.Atoi(remainingRetriesStr)
+	if err != nil {
+		bs.T().Logf("unable to parse remaining retries, err: %v, defaulting to 0", err)
+		remainingRetries = 0
+	}
+
+	// Note: this check might sometimes skip stack deletion even if it would be needed
+	// For example if the only tests that failed were known flaky tests and thus not retried.
+	// This is not too big a deal as the stack is cleaned up manually in the invoke task handling retries.
```
Ishirui: Is this too dirty ? I wasn't sure if it was better to recompute all the logic for determining whether we should retry in here or not, it seems very cumbersome for not that much protection

.gitlab/source_test/linux.yml
```diff
@@ -196,7 +196,7 @@ new-e2e-unit-tests:
     - pulumi login "s3://dd-pulumi-state?region=us-east-1&awssdk=v2&profile=$AWS_PROFILE"
   script:
     - dda inv -- -e gitlab.generate-ci-visibility-links --output=$EXTERNAL_LINKS_PATH
-    - dda inv -- -e new-e2e-tests.run --targets ./pkg/utils --junit-tar junit-${CI_JOB_ID}.tgz ${EXTRA_PARAMS}
+    - dda inv -- -e new-e2e-tests.run --targets ./pkg/utils --max-retries 3 --junit-tar junit-${CI_JOB_ID}.tgz ${EXTRA_PARAMS}
```
KevinFairise2: Let's not put it on the unit tests, it should be only unit testing the function we have to create and destroy stacks, if this is flaky there is something strange

tasks/new_e2e_tests.py
```diff
@@ -195,18 +197,54 @@ def run(
...
+                os.path.relpath(package, "github.com/DataDog/datadog-agent/test/new-e2e/") for package, _ in to_retry
+            }
+            e2e_module.test_targets = list(affected_packages)
+            args["run"] = '-test.run ' + _create_test_selection_regex([test for _, test in to_retry])
+        else:
+            break
+
+    # In some cases the retry logic might leave resources undestroyed - make sure they are cleaned up
+    if max_retries > 0:
+        _clean_stacks(ctx, skip_destroy=not keep_stacks)
```
KevinFairise2: If we do that it will break the feature we have that allows team to keep the stack when there are some failures, with: https://github.com/DataDog/datadog-agent/blob/main/test/new-e2e/tests/otel/otel-agent/loadbalancing_test.go#L37

KevinFairise2: Do you think we can find a way to make sure that this call ignore the stack we explicitly marked to be kept?

tasks/new_e2e_tests.py
```diff
@@ -195,18 +197,54 @@ def run(
...
+            affected_packages = {
+                os.path.relpath(package, "github.com/DataDog/datadog-agent/test/new-e2e/") for package, _ in to_retry
+            }
+            e2e_module.test_targets = list(affected_packages)
+            args["run"] = '-test.run ' + _create_test_selection_regex([test for _, test in to_retry])
+        else:
+            break
+
+    # In some cases the retry logic might leave resources undestroyed - make sure they are cleaned up
+    if max_retries > 0:
```
KevinFairise2: Let's use `max_retries > 0 and not keep_stacks`.
Calling `_clean_stacks` with `skip_destroy` will not destroy the stack but still remove it from the remote state, which will make us unable to destroy it properly later on

Ishirui: Right - so we should always set `skip_destroy = False` right ? (In this case I mean)

tasks/new_e2e_tests.py
Ishirui: Ack, I feared something like this would be the case. I think this means I need to have all the logic for determining whether to cleanup or not in go in `TearDownSuite`. That'll be a bit more work, especially since I am not sure how to access all the data I need :sweat_smile:. I'll have to take a further look.

Ishirui: Note after meeting with @KevinFairise2: there are actually a million different edge cases and complex logic to handle here, implementing this will require a lot more work. In particular, we should not do a blind clean of all pulumi-managed stacks, but instead implement a mechanism to run only the `TearDownSuite` method.

tasks/new_e2e_tests.py
Ishirui: Not going to matter anymore

test/new-e2e/pkg/e2e/suite.go
Ishirui: Not relevant anymore

tasks/new_e2e_tests.py
```diff
@@ -195,18 +197,90 @@ def run(
...
+                "yellow",
+            )
+        )
+        affected_packages = {
+            os.path.relpath(package, "github.com/DataDog/datadog-agent/test/new-e2e/") for package, _ in to_teardown
+        }
+        e2e_module.test_targets = list(affected_packages)
+        args["run"] = '-test.run ' + _create_test_selection_regex([test for _, test in to_teardown])
+        env_vars["E2E_TEARDOWN_ONLY"] = "true"
+        test_flavor(
```
Ishirui, line 457: This is not amazing, the command exits with a nonzero exit code so we get some extra log messages about failures which should not be there (we are only trying to tear stuff down, we don't care about any failures). Not sure what's worse here: extra noise in the CI logs, or further making a mess of the codebase by adding a `ignore_failures` param or something.

.gitlab/source_test/linux.yml
Ishirui: Done :+1:

tasks/new_e2e_tests.py
Ishirui: Same comment now for the `filter_only_leaf_tests` function

tasks/new_e2e_tests.py
```diff
@@ -362,18 +365,98 @@ def run(
...
+            modules=[e2e_module],
+            args=args,
+            cmd=cmd,
+            env=env_vars,
+            junit_tar=junit_tar,
+            result_json=result_json,
+            test_profiler=None,
+        )
 
     success = process_test_result(test_res, junit_tar, AgentFlavor.base, test_washer)
```
Ishirui, line 476: TODO: Change how this works. Due to our retries, the json file gets constantly overwritten, so this does not hold any meaning anymore - this especially bad when there is a `teardown_only` run, as that will _also_ overwrite the json file.
The best solution is probably to make sure that the json file gets _updated_ on each retry, instead of overwritten - except for the teardown, in which case we just write to `/dev/null`.

.gitlab/e2e_install_packages/common.yml
KevinFairise2: Let's enable it on `new-e2e-installer*` jobs instead: https://github.com/DataDog/datadog-agent/blob/main/.gitlab/e2e/e2e.yml#L540
So that we can remove the `retry: 1` that we have in the job definition.

Ishirui: Ah, thanks for the clarification - I must have misunderstood. Should be fixed now :+1:

tasks/new_e2e_tests.py
KevinFairise2: Yes I think we can move them in `libs/`. Maybe `libs/testing` would be a goo d place

KevinFairise2 (approved):

sblumenthal (approved):

PR #37868, opened 2025-06-11
Move in sample service, add decoding to integration test, and add probe definitions

grantseltzer:
### What does this PR do?

- Adds the sample service, removes old ones
- Adds decoding to the integration test
    - For now, we just print the output of decoding. Im about to make changes to the output format (to include the expected enveloping structure), so I'll add golden output files then.
- Adds probe definition yaml files

### Motivation

Improving testing

Ran the integration test


pkg/dyninst/integration_test.go
```diff
@@ -19,13 +20,14 @@ import (
 	"time"
 	"unsafe"
 
+	"github.com/alecthomas/assert/v2"
```
ajwerner: I think you want `"github.com/stretchr/testify/assert"` to the extent you want to assert and not require.

pkg/dyninst/testprogs/progs/sample/other.go
```diff
@@ -0,0 +1,44 @@
...
+
+//nolint:all
+//go:noinline
+func test_trigger_verifier_error(t triggerVerifierErrorForTesting) {}
+
+// return_goroutine_id gets the goroutine ID and returns it
+//
+//nolint:all
+//go:noinline
+func Return_goroutine_id() uint64 {
```
ajwerner: nit: why the strange function name with underscores?

ajwerner (commented):
Let's kill the underscore function names while you're here.

pkg/dyninst/integration_test.go
```diff
@@ -5,9 +5,19 @@
 
 //go:build linux_bpf
 
+// This test is used to run the dynamic instrumentation product on a sample binary from end to end.
+// Update test data by setting SAVE_OUTPUT=true and running the test.
```
piob-io: Let's use same environment variable as in other tests. So we should either use REWRITE= here, or SAVE_OUTPUT= in the other tests.

pkg/dyninst/integration_test.go
```diff
@@ -53,13 +68,190 @@ func TestDyninst(t *testing.T) {
...
+			defer func() {
+				preserve, _ := strconv.ParseBool(os.Getenv("KEEP_TEMP"))
+				if preserve || t.Failed() {
+					t.Logf("leaving temp dir %s for inspection", tempDir)
+				} else {
+					require.NoError(t, os.RemoveAll(tempDir))
+				}
+			}()
+
+			// Load the binary and generate the IR.
```
piob-io: The code being commented here does more then described - beyond generating IR we compile and run ebpf, and decode. Let's adjust the comment to specify actions that cover whole test, or maybe just drop this comment.

pkg/dyninst/integration_test.go
```diff
@@ -53,13 +68,190 @@ func TestDyninst(t *testing.T) {
...
+
+	// Save actual output to YAML file if SAVE_OUTPUT is set
+	if saveOutput, _ := strconv.ParseBool(os.Getenv("SAVE_OUTPUT")); saveOutput {
+		expectedOutput["main."+probe.GetID()] = out.String()
+		testprogs.SaveActualOutput(t, "sample", expectedOutput)
+	}
+
+}
+
+func testAllProbesAtOnce(t *testing.T, sampleServicePath string, probes []config.Probe, expectedOutput map[string]string) {
```
piob-io: Could code between testAllProbesAtOnce and testSingleProbe be shared (I am thinking about having a single function that is sufficiently paramterized)? I do prefer verbose test code, but here it feels that we are repeating significant amount of code. FYI - I'm moving parts of the test code into utility functions (https://github.com/DataDog/datadog-agent/pull/37911/files#diff-5d2f3702833d9ba56fb056e46c4bff52042032606cbf3093c15eb07b9ba30233), which would reduce size of each test functions here, still big portions across testAllProbesAtOnce and testSingleProbe will stay the same, hence the suggestion to share a single paramterized function, while my refactor will then reduce size of that single function further.

pkg/dyninst/integration_test.go
```diff
@@ -53,13 +68,190 @@ func TestDyninst(t *testing.T) {
...
+	// Validate the output. For now we just check the total length.
+	t.Logf("processing output")
+	rd, err := ringbuf.NewReader(bpfCollection.Maps["out_ringbuf"])
+	require.NoError(t, err)
+
+	bpfOutDump, err := os.Create(filepath.Join(tempDir, "probe.bpf.out"))
+	require.NoError(t, err)
+	defer func() { require.NoError(t, bpfOutDump.Close()) }()
+
+	// Inlined function is called twice, hence extra event.
```
piob-io: Stale comment? Although I think we need to somehow read the buffer twice for the inline function case... Maybe we can read the ringbuf until it's empty? The snapshot output comparison will then assure we have seen enough of it.

pkg/dyninst/testprogs/testdata/expected_output/sample.yaml
piob-io, line 1: This testdata is specific to the integration_test.go, not something sharable across all tests that use testprogs. I think it therefore belongs undersomething like pkg/dyninst/testdata/integration_test/*
I'm not an expert on go directory structure tho (but want to keep packages reasonably-narrow-purpose). Wdyt?

pkg/dyninst/testprogs/probes.go
```diff
@@ -52,3 +52,47 @@ func GetProbeCfgs(t *testing.T, name string) []config.Probe {
 	}
 	return probesCfgs
 }
+
+// GetExpectedOutput returns the expected output for a given service.
```
piob-io: See comment for testdata/expected_output/sample.yaml. I don't think this code would be shared across other tests, it's specific to how integration test is written. Should these helpers just be in integration_test.go file?

pkg/dyninst/integration_test.go
```diff
@@ -53,13 +68,190 @@ func TestDyninst(t *testing.T) {
...
+	record, err := rd.Read()
+	require.NoError(t, err)
+	bpfOutDump.Write(record.RawSample)
+
+	header := (*output.EventHeader)(unsafe.Pointer(&record.RawSample[0]))
+	require.Equal(t, uint32(len(record.RawSample)), header.Data_byte_len)
+	t.Logf("header: %#v", *header)
+
+	pos := uint32(unsafe.Sizeof(*header)) + uint32(header.Stack_byte_len)
+	for pos < header.Data_byte_len {
```
piob-io: This code here was just to do some sanity-checks on the output, in lieu of decoding. Now that we have decoding, it better validates that the output is correct, and we do not need the checks and logs here.

pkg/dyninst/integration_test.go
```diff
@@ -177,6 +367,13 @@ func testDyninst(t *testing.T, sampleServicePath string) {
...
 
+	// Have a 'reverse map' of probe output, to ensure all events are received
+	// and no unexpected events are received.
+	probeOutput := map[string]struct{}{}
+	for _, o := range expectedOutput {
+		probeOutput[o] = struct{}{}
+	}
+
 	// Inlined function is called twice, hence extra event.
 	for range len(probes) + 1 {
```
piob-io: Ditto

pkg/dyninst/testprogs/progs/sample/arrays.go
```diff
@@ -0,0 +1,117 @@
...
+func testArrayOfArraysOfArrays(b [2][2][2]int) {}
+
+//nolint:all
+//go:noinline
+func testArrayOfStructs(a [2]nestedStruct) {
+}
+
+//nolint:all
+//go:noinline
+func testOverLimitArrayParameters(a, b, c, d, e, f, g,
```
piob-io: I think if we break lines on the function call, first argument also goes to the next line.

pkg/dyninst/decode/decoder.go
```diff
@@ -60,6 +60,16 @@ type typeAndAddr struct {
...
+// TypeIDToProbeEvent returns the probe event for the given type ID.
+func (d *Decoder) TypeIDToProbeEvent(typeID ir.TypeID) *ir.Event {
+	return d.probeEvents[typeID].event
+}
+
+// TypeIDToProbe returns the probe for the given type ID.
+func (d *Decoder) TypeIDToProbe(typeID ir.TypeID) *ir.Probe {
+	return d.probeEvents[typeID].probe
+}
+
```
ajwerner: These don't seem to be used.

piob-io (commented):
Thank you!

pkg/dyninst/integration_test.go
```diff
@@ -53,13 +68,190 @@ func TestDyninst(t *testing.T) {
 			if cfg.GOARCH != runtime.GOARCH {
 				t.Skipf("cross-execution is not supported, running on %s", runtime.GOARCH)
 			}
-			bin := testprogs.GetBinary(t, "simple", cfg)
-			testDyninst(t, bin)
+			bin := testprogs.GetBinary(t, "sample", cfg)
+			t.Logf("loading binary")
```
ajwerner: This log line is a bit odd given nothing that follows it involves loading a binary.

pkg/dyninst/testprogs/probes.go
```diff
@@ -52,3 +52,47 @@ func GetProbeCfgs(t *testing.T, name string) []config.Probe {
...
+func GetExpectedOutput(t *testing.T, name string) map[string]string {
```
ajwerner: I do not find it very reasonable to have this function or the output of a test in another package to be underneath `testprogs`. This should be *much* closer to the test that uses it and the testdata for that test should be underneath its own package.

pkg/dyninst/testprogs/testdata/expected_output/sample.yaml
ajwerner: +1 -- the testdata directory for a test belongs underneath the directory of the package in which that test lives.

pkg/dyninst/testprogs/testdata/probes/sample.yaml
```diff
@@ -0,0 +1,254 @@
+binary: sample
+probes:
+- id: test_single_byte
+  type: LOG_PROBE
+  where:
+    method_name: main.test_single_byte
```
ajwerner: Didn't all these functions get renamed to camelCase? How does this pass?

grantseltzer: Needed to rebuild the binaries locally, I totally missed that. Fixing.

brycekahle (approved):

grantseltzer:
Closing in favor of #37952

PR #37872, opened 2025-06-11
extend service command timeout by number of dependent services

clarkb7:
### What does this PR do?
extend service command timeout by number of dependent services

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1548
We have up to 6 services to stop but it only takes 2 to hang to hit the 30 second timeout. This can block remote updates, and cause issues with other upgrade paths (files in use, etc.)

This is difficult to test, you can't simply suspend a process b/c then you'll trigger the SCM pipe timeout instead. easiest method seems to be to add a `time.Sleep(30*time.Second)` to a subprocess and rebuild it.
I added it to `cmd\trace-agent\subcommands\run\command.go` at the end of `runTraceAgentProcess`.
Then set the `DD_WINDOWS_SERVICE_STOP_TIMEOUT_SECONDS=60` environment variable at the machine level to prevent the hard stop timeout from triggering, too.
then run
```PowerShell
& 'C:\Program Files\Datadog\Datadog Agent\bin\agent.exe' stop-service
```
This command should fail in previous versions of the Agent, with this PR it should succeed

### DIFF
--- a/pkg/util/winutil/service.go
+++ b/pkg/util/winutil/service.go
@@ -237,9 +237,6 @@ func StopService(serviceName string) error {
 func doStopServiceWithDependencies(manager *mgr.Mgr, service *mgr.Service,
 	depenStatus svc.ActivityStatus, callback stopServiceCallback) error {
 
-	ctx, cancel := context.WithTimeout(context.Background(), defaultServiceCommandTimeout*time.Second)
-	defer cancel()
-
 	// open dependent services
 	depServiceNames, err := service.ListDependentServices(depenStatus)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


rtrieu (approved):

pkg/util/winutil/service.go
```diff
@@ -258,6 +255,11 @@ func doStopServiceWithDependencies(manager *mgr.Mgr, service *mgr.Service,
 		defer depService.Close()
 	}
 
+	// extend deadline to account for all services we are trying to stop
+	totalTimeout := time.Duration(len(depServices)+1) * defaultServiceCommandTimeout * time.Second
```
alexn-dd, line 259: These types of timeouts are easier to replicate under low resources.
Unfortunately, there is no specific timeout value that serves well for various customer conditions.  At some point, one needs to prepare to handle a dirty state.

alexn-dd (approved):

pkg/util/winutil/service.go
clarkb7: agreed. In the majority of cases I think we'd prefer to force kill the processes rather than error out. This is the behavior from other platform's service managers. We'll have to follow up on this.

iglendd (approved):
Looks good

pkg/util/winutil/service.go
clarkb7: https://datadoghq.atlassian.net/browse/WINA-1559

PR #37874, opened 2025-06-11
dyninst: extract symbols from Go binaries for SymDB

andreimatei:
Walk through the DWARF debug info for a Go binary and extract the symbols data that we'll export to SymDB.

### DIFF
--- a/pkg/dyninst/irgen/irgen.go
+++ b/pkg/dyninst/irgen/irgen.go
@@ -26,6 +26,7 @@ import (
 	"debug/dwarf"
 	"errors"
 	"fmt"
+	"github.com/DataDog/datadog-agent/pkg/dyninst/dwarf/dwarfutil"
 	"iter"
 	"maps"
 	"math"
@@ -1197,52 +1198,11 @@ func (v *rootVisitor) computeLocations(
 	typ ir.Type,
 	locField *dwarf.Field,
 ) ([]ir.Location, error) {
-	totalSize := typ.GetByteSize()
-	var locations []ir.Location
-	switch locField.Class {
-	case dwarf.ClassLocListPtr:
-		offset, ok := locField.Val.(int64)
-		if !ok {
-			return nil, fmt.Errorf(
-				"unexpected location field type: %T", locField.Val,
-			)
-		}
-		loclist, err := v.loclistReader.Read(unit, offset, totalSize)
-		if err != nil {
-			return nil, err
-		}
-		if len(loclist.Default) > 0 {
-			return nil, fmt.Errorf("unexpected default location pieces")
-		}
-		locations = loclist.Locations
-
-	case dwarf.ClassExprLoc:
-		instr, ok := locField.Val.([]byte)
-		if !ok {
-			return nil, fmt.Errorf(
-				"unexpected location field type: %T", locField.Val,
-			)
-		}
-		pieces, err := loclist.ParseInstructions(instr, v.object.PointerSize(), totalSize)
-		if err != nil {
-			return nil, err
-		}
-		// BUG: This should take into consideration the ranges of the current
-		// block, not necessarily the ranges of the subprogram.
-		for _, r := range subprogramRanges {
-			locations = append(locations, ir.Location{
-				Range:  r,
-				Pieces: pieces,
-			})
-		}
-	default:
-		return nil, fmt.Errorf(
-			"unexpected %s class: %s",
-			locField.Attr, locField.Class,
-		)
-	}
-
-	return locations, nil
+	// BUG: We shouldn't pass subprogramRanges below; we should take into
+	// consideration the ranges of the current block, not necessarily the ranges
+	// of the subprogram.
+	return dwarfutil.ProcessLocations(
+		locField, unit, v.loclistReader, subprogramRanges, typ.GetByteSize(), v.object.PointerSize())
 }
 
 // maybeGetAttr is a helper function that returns the value of an attribute if
--- a/pkg/dyninst/symdb/func_name.go
+++ b/pkg/dyninst/symdb/func_name.go
@@ -1,0 +1,145 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-present Datadog, Inc.
+
+//go:build linux_bpf
+
+package symdb
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+// Utilities for parsing Go function names.
+
+var (
+	// The parsing goes as follows:
+	// - an optional package name: (((.*/)?.*?)\.)?. Consume (greedily)
+	// everything up to the last slash, and then non-greedily up to the
+	// following dot.
+	// - an optional type name: (\(?\*?(.*?)\)?\.)?. The type name maybe be
+	// between parens and also start with a '*' if it's a pointer receiver;
+	// otherwise the type name is not wrapped in parens. Note that we won't
+	// parse correctly if there is a type name and no package name. Hopefully that
+	// doesn't exist.
+	//   - note that we don't include the optional '*' in the capture group. We
+	//     don't differentiate between pointer and value receivers.
+	//   - note the lazy capture group for the type name: .*?. It's lazy because
+	//     otherwise it also captures the following ')'
+	//   - the function name: (.*)
+	parseFuncNameRE = regexp.MustCompile(`^((?P<pkg>(.*/)?.*?)\.)?(\(?\*?(?P<typ>.*?)\)?\.)?(?P<name>.*)$`)
+	pkgIdx          = parseFuncNameRE.SubexpIndex("pkg")
+	typIdx          = parseFuncNameRE.SubexpIndex("typ")
+	nameIdx         = parseFuncNameRE.SubexpIndex("name")
+
+	// Recognize anonymous functions declared inside a function that is not a method.
+	// They look like:
+	// github.com/.../pkg.myFunc.func1
+	// internal/bytealg.init.0
+	parseAnonymousFuncNameRE = regexp.MustCompile(`^((?P<pkg>(.*/)?\w*?)\.)?(?P<func>(\w*?\.)?((func|gowrap|(\d+)).*))$`)
+	anonPkgIdx               = parseFuncNameRE.SubexpIndex("pkg")
+	anonNameIdx              = parseFuncNameRE.SubexpIndex("name")
+
+	// Recognize funky functions corresponding to anonymous functions defined
+	// inside inlined functions. These have names like:
+	// runtime.gcMarkDone.forEachP.func5
+	// "forEachP.func5" is an anonymous function defined inside forEachP, and
+	// "runtime.gcMarkDone" is the functions inside which "forEachP" is inlined.
+	parseAnonFuncInsideInlinedFunc = regexp.MustCompile(`^((?P<pkg>(.*/)?\w*?)\.)\w+\.\w+\.func(\d)+$`)
+)
+
+type FuncName struct {
+	Package string
+	// Type is the type of the receiver, if any. Empty if this function is not a
+	// method. The type is not a pointer type even if the method has a
+	// pointer-receiver; the base type is returned, without the '*'.
+	Type string
+	Name string
+	// QualifiedName looks like
+	// github.com/cockroachdb/cockroach/pkg/kv/kvserver.(*raftSchedulerShard).worker
+	QualifiedName string
+
+	// GenericFunction is set if the function takes type arguments. We don't
+	// support parsing these functions at the moment, so no other fields except
+	// QualifiedName are set.
+	GenericFunction bool
+}
+
+func (f *FuncName) Empty() bool {
+	return *f == (FuncName{})
+}
+
+// ParseFuncName parses a Go qualified function name. For a qualifiedName name
+// like:
+// github.com/cockroachdb/cockroach/pkg/kv/kvserver.(*raftSchedulerShard).worker
+// the package is: github.com/cockroachdb/cockroach/pkg/kv/kvserver
+// the type is: raftSchedulerShard (note that it doesn't include the '*' signifying a pointer receiver).
+// the name is: worker
+//
+// Returns (zero value, nil) if the function should be ignored.
+//
+// Cases we need to support:
+// github.com/cockroachdb/cockroach/pkg/kv/kvserver.(*raftSchedulerShard).worker
+// github.com/cockroachdb/cockroach/pkg/ccl/changefeedccl/kvfeed.rangefeedFactory.Run
+// github.com/klauspost/compress/zstd.sequenceDecs_decode_amd64
+// indexbytebody
+// github.com/.../pkg.myFunc.func0
+// internal/bytealg.init.0
+func ParseFuncName(qualifiedName string) (FuncName, error) {
+	// Filter out generic functions, e.g.
+	// os.init.OnceValue[go.shape.interface { Error() string }].func3
+	// Note that this name is weird -- os.init is neither a package nor a type,
+	// but rather it has something to do with the generic function's caller.
+	if strings.ContainsRune(qualifiedName, '[') {
+		return FuncName{
+			QualifiedName:   qualifiedName,
+			GenericFunction: true,
+		}, nil
+	}
+
+	// Ignore anonymous functions declared inside inlined functions. These are
+	// DWARF entries corresponding to anonymous functions that are defined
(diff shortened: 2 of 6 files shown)


pkg/dyninst/symdb/symdb.go
```diff
@@ -0,0 +1,775 @@
...
+
+func NewSymDBBuilder(
+	dwarfData *dwarf.Data, sym *gosym.Table, loclistReader *object.LoclistReader, pointerSize int,
+) *SymDBBuilder {
+	return &SymDBBuilder{
+		dwarfData:     dwarfData,
+		sym:           sym,
+		loclistReader: loclistReader,
+		// TODO: Should we use entryReader.AddressSize() as pointerSize?
+		pointerSize: pointerSize,
```
grantseltzer: What's used now? I think the pointer size (8) should be used.

grantseltzer:
I'm not great with regex, so would this filter out funky symbol names at the top of the symbol table like `runtime..interfaceSwitch.0`, `debugCall64`, or `$f64.41012e0be826d695` ?

pkg/dyninst/symdb/func_name.go
```diff
@@ -0,0 +1,141 @@
...
+
+	// Recognize funky functions corresponding to anonymous functions defined
+	// inside inlined functions. These have names like:
+	// runtime.gcMarkDone.forEachP.func5
+	// "forEachP.func5" is an anonymous function defined inside forEachP, and
+	// "runtime.gcMarkDone" is the functions inside which "forEachP" is inlined.
+	parseAnonFuncInsideInlinedFunc = regexp.MustCompile(`^((?P<pkg>(.*/)?\w*?)\.)\w+\.\w+\.func(\d)+$`)
+)
+
+type FuncName struct {
```
ajwerner, line 54: nit: comment, also, for whom is this infrastructure being exported?

pkg/dyninst/symdb/func_name.go
```diff
@@ -0,0 +1,141 @@
...
+//
+// This function should be kept in sync with parse_func_name in Rust.
+//
+// Cases we need to support:
+// github.com/cockroachdb/cockroach/pkg/kv/kvserver.(*raftSchedulerShard).worker
+// github.com/cockroachdb/cockroach/pkg/ccl/changefeedccl/kvfeed.rangefeedFactory.Run
+// github.com/klauspost/compress/zstd.sequenceDecs_decode_amd64
+// indexbytebody
+// github.com/.../pkg.myFunc.func0
+// internal/bytealg.init.0
```
ajwerner, line 90: what about the deeper nested functions? their deferwraps?

`github.com/cockroachdb/cockroach/pkg/server.(*apiV2Server).execSQL.func8.1.3.2`
`github.com/.../logical.(*logicalReplicationWriterProcessor).flushBuffer.Group.GoCtx.func7.1.deferwrap1`

Look at this one: 
`github.com/cockroachdb/cockroach/pkg/server.(*topLevelServer).startPersistingHLCUpperBound.func1.(*Node).SetHLCUpperBound.1` ([code](https://github.com/cockroachdb/cockroach/blob/6bb7514c9db7662f51d37fb12935bd2d8735f3a3/pkg/server/clock_monotonicity.go#L243-L245))

Even worse, sometimes there's nested `func` labels:
`github.com/cockroachdb/cockroach/pkg/ccl/streamingccl/streamproducer.(*eventStream).startStreamProcessor.(*eventStream).startStreamProcessor.func1.func3.1`

pkg/dyninst/symdb/func_name_test.go
```diff
@@ -0,0 +1,52 @@
+package symdb_test
+
+import (
+	"github.com/DataDog/datadog-agent/pkg/dyninst/symdb"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestParseFuncName(t *testing.T) {
```
ajwerner, line 9: my 2c is that you'd benefit from a much larger list of functions to test

pkg/dyninst/symdb/func_name_test.go
```diff
@@ -0,0 +1,52 @@
+package symdb_test
+
+import (
+	"github.com/DataDog/datadog-agent/pkg/dyninst/symdb"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
```
ajwerner, line 7: linter will yell about your import ordering. Here's the vscode setting to have gopls fix 'em up automatically:
```
"gopls": {
        "formatting.local": "github.com/DataDog/datadog-agent",
},
```

pkg/dyninst/symdb/func_name_test.go
```diff
@@ -0,0 +1,52 @@
+package symdb_test
```
ajwerner, line 1: missing headers and build tags

pkg/dyninst/symdb/symdb.go
```diff
@@ -0,0 +1,775 @@
...
+}
+
+// SymDBBuilder walks the DWARF data for a binary, extracting symbols in the
+// SymDB format.
+type SymDBBuilder struct {
+	// The DWARF data to extract symbols from.
+	dwarfData *dwarf.Data
+	// The Go symbol table for the binary, used to resolve PC addresses to
+	// functions.
+	sym *gosym.Table
```
ajwerner, line 102: we're going to need to reimplement this thing if we care about memory usage...

pkg/dyninst/symdb/symdb.go
ajwerner: I forwarded the definition of Architecture into `object` and it has a `PointerSize()` method. You'll be using the `object.File` anyway I hope, so just use that.

pkg/dyninst/symdb/symdb.go
```diff
@@ -0,0 +1,775 @@
...
+	entryReader := b.dwarfData.Reader()
+
+	// Recognize compile units, which are the top-level entries in the DWARF
+	// data corresponding to Go packages.
+	for entry, err := entryReader.Next(); entry != nil; entry, err = entryReader.Next() {
+		if err != nil {
+			return Symbols{}, err
+		}
+
+		if entry.Tag != dwarf.TagCompileUnit {
```
ajwerner, line 259: is this ever true?

pkg/dyninst/symdb/symdb.go
```diff
@@ -0,0 +1,775 @@
...
+					return Package{}, fmt.Errorf(
+						"compile unit %s has invalid nil file entry at index %d", name, i)
+				}
+				files = append(files, "")
+				continue
+			}
+			files = append(files, strings.Clone(file.Name))
+		}
+	}
+	b.filesInCurrentCompileUnit = files
```
ajwerner, line 358: you clear the current compile unit upon return, should you clear this too?

pkg/dyninst/symdb/func_name.go
```diff
@@ -0,0 +1,141 @@
...
+	parseAnonymousFuncNameRE = regexp.MustCompile(`^((?P<pkg>(.*/)?\w*?)\.)?(?P<func>(\w*?\.)?((func|gowrap|(\d+)).*))$`)
+	anonPkgIdx               = parseFuncNameRE.SubexpIndex("pkg")
+	anonNameIdx              = parseFuncNameRE.SubexpIndex("name")
+
+	// Recognize funky functions corresponding to anonymous functions defined
+	// inside inlined functions. These have names like:
+	// runtime.gcMarkDone.forEachP.func5
+	// "forEachP.func5" is an anonymous function defined inside forEachP, and
+	// "runtime.gcMarkDone" is the functions inside which "forEachP" is inlined.
+	parseAnonFuncInsideInlinedFunc = regexp.MustCompile(`^((?P<pkg>(.*/)?\w*?)\.)\w+\.\w+\.func(\d)+$`)
```
ajwerner, line 51: nit: your other regexes end in RE

ajwerner (commented):
Out of curiosity, have you tried running this on a big binary? I'd be very curious to know:
* Peak memory use
* Total time
* Size of output


It's not clear this was ready for review. I haven't been particularly comprehensive, but leaving an initial round of comments after a pass.

pkg/dyninst/symdb/symdb.go
andreimatei: switched to using `object.ElfFile`. Why do we need the `object.File` interface?

pkg/dyninst/symdb/func_name.go
andreimatei: > what about the deeper nested functions? their deferwraps?

These cases are handled; they fall into the regexps. I wouldn't list too many examples in this function comment, but I've now added tests for them.

> Look at this one:
`github.com/cockroachdb/cockroach/pkg/server.(*topLevelServer).startPersistingHLCUpperBound.func1.(*Node).SetHLCUpperBound.1`
> Even worse, sometimes there's nested func labels:
`github.com/cockroachdb/cockroach/pkg/ccl/streamingccl/streamproducer.(*eventStream).startStreamProcessor.(*eventStream).startStreamProcessor.func1.func3.1`

The nested `func` labels are not a particular problem by themselves, but these functions are examples of anonymous functions called instances of inlined functions (e.g. `(*eventStream).startStreamProcessor.func1` is inlined inside `github.com/cockroachdb/cockroach/pkg/ccl/streamingccl/streamproducer.(*eventStream).startStreamProcessor` and calls `func3.1`). There's code to ignore such functions cause I don't know what to do with them, although I think I don't catch all of them.

pkg/dyninst/symdb/func_name_test.go
andreimatei: added some more. I'll add more as we look at the generated symbols from some binaries and find more funky cases that we want to ignore or deal with in particular

pkg/dyninst/symdb/symdb.go
andreimatei: I don't know, but I'd keep the code as is. No?

pkg/dyninst/symdb/symdb.go
andreimatei: done

pkg/dyninst/symdb/func_name_test.go
andreimatei: done

pkg/dyninst/symdb/symdb.go
```diff
@@ -0,0 +1,772 @@
...
+	// know all the methods of their types).
+	//for _, t := range b.types {
+	//	res.Types = append(res.Types, t)
+	//}
+
+	return res, nil
+}
+
+// DW_INL_inlined is the value for AttrInline marking the function as inlined.
+const DW_INL_inlined = 1
```
piob-io, line 411: I would make this dw_inl_inlined so it doesn't get exported.

pkg/dyninst/symdb/symdb.go
piob-io: Now we have our own gosym implementation. I think it has all that we need here, lmk otherwise.

pkg/dyninst/symdb/func_name.go
andreimatei: unexported

pkg/dyninst/symdb/func_name.go
```diff
@@ -0,0 +1,145 @@
...
+	// - an optional package name: (((.*/)?.*?)\.)?. Consume (greedily)
+	// everything up to the last slash, and then non-greedily up to the
+	// following dot.
+	// - an optional type name: (\(?\*?(.*?)\)?\.)?. The type name maybe be
+	// between parens and also start with a '*' if it's a pointer receiver;
+	// otherwise the type name is not wrapped in parens. Note that we won't
+	// parse correctly if there is a type name and no package name. Hopefully that
+	// doesn't exist.
+	//   - note that we don't include the optional '*' in the capture group. We
+	//     don't differentiate between pointer and value receivers.
```
piob-io, line 29: I'm curious why we don't want to differentiate?

pkg/dyninst/symdb/symdb_test.go
piob-io, line 1: Consider also adding snapshot test that would print whole symdb for selected testprogs we have.

pkg/dyninst/symdb/symdb.go
```diff
@@ -0,0 +1,772 @@
...
+// Scope represents a function or another lexical block.
+type Scope struct {
+	StartLine int
+	EndLine   int
+	Scopes    []Scope
+	Variables []Variable
+}
+
+// Variable represents a variable or function argument.
+type Variable struct {
```
piob-io, line 80: Shall we annotate return values too?

pkg/dyninst/symdb/symdb.go
```diff
@@ -0,0 +1,772 @@
...
+		return res, nil
+	}
+}
+
+func (b *SymDBBuilder) exploreInlinedSubroutine(inlinedEntry *dwarf.Entry, reader *dwarf.Reader) error {
+	if inlinedEntry.Tag != dwarf.TagInlinedSubroutine {
+		return fmt.Errorf("expected TagInlinedSubroutine, got %s", inlinedEntry.Tag)
+	}
+	// TODO: accumulate variable availability data and join it with the other
+	// inlined and out-of-line instances.
```
piob-io, line 566: We also should resolve file/line of the inlined function declaration. The reason to do it here is that the inline function definition doesn't have a file, and we need to run line program for any inlined instantiation. Expand todo?

pkg/dyninst/symdb/symdb.go
```diff
@@ -0,0 +1,772 @@
...
+		}
+
+		// We recognize formal parameters, variables, lexical blocks and inlined subroutines.
+		switch child.Tag {
+		case dwarf.TagFormalParameter:
+			v, err := b.parseVariable(child, funcTable)
+			if err != nil {
+				return exploreCodeResult{}, err
+			}
+			v.FunctionArgument = true
```
piob-io, line 741: TagFormalParamter also includes named return values, which can be recognized by also having a DW_AT_variable_parameter tag with value set to 1. I think these should have FunctionArgument = false.

piob-io (commented):
Neat!

pkg/dyninst/irgen/irgen.go
```diff
@@ -26,6 +26,7 @@ import (
 	"debug/dwarf"
 	"errors"
 	"fmt"
+	"github.com/DataDog/datadog-agent/pkg/dyninst/dwarf/dwarfutil"
```
tylfin, line 29: nit: format imports

pkg/dyninst/symdb/func_name.go
andreimatei: Because, for SymDB purposes, we consider both functions with pointer receiver and functions with value receiver to be methods on the same type -- the type without the leading `*`. I'm not sure in general where listing a type's methods is used for exactly, but that's the common way to organize methods in the SymDB data -- they all hang from a class.
To put it differently, where/why would we differentiate?

pkg/dyninst/symdb/symdb_test.go
andreimatei: yeah; I'm thinking about better testing

pkg/dyninst/irgen/irgen.go
andreimatei: done

andreimatei (commented):
Closing this and moving the review to https://github.com/DataDog/datadog-agent/pull/38679 because I messed up and created this PR from a forked repo, which apparently is no bueno.

pkg/dyninst/symdb/symdb.go
andreimatei: Maybe, but how would such an annotation be used? I'd wait until we need it.

pkg/dyninst/symdb/symdb.go
andreimatei: moved to `gosym`

pkg/dyninst/symdb/symdb.go
andreimatei: meh, this matches how it's spelled elsewhere. And I don't really mind it being exported.

pkg/dyninst/symdb/symdb.go
andreimatei: done

pkg/dyninst/symdb/symdb.go
andreimatei: explained the deal in TODO

andreimatei:
closing in favor of #39020. I had to create a new PR because I screwed up and created this PR from a fork of the repo.

pkg/dyninst/symdb/func_name.go
piob-io: Makes sense, thanks! The UI offers ability to pick a type first, to then offer suggestions on its methods.

PR #37878, opened 2025-06-12
Add 'preserve_root_device' config flag and mountinfo parser for raw device names

jose-manuel-almaza:
### What does this PR do?
This PR introduces a new boolean configuration flag, `resolve_root_device` (default false), which lets users opt-in to seeing the original mount sources (e.g. `/dev/root`) exactly as they appear in `/proc/self/mountinfo`. When set to true, we fall back to `gopsutil`’s resolved device paths (e.g. `/dev/nvme0n1p2`).
### Motivation
When migrating our disk checks from Python’s psutil to Go’s gopsutil, we discovered that by default:
- **Raw sources get “resolved”**: `gopsutil` never reports `/dev/root` or `overlay`—it always looks up the underlying block device (e.g. `/dev/nvme0n1p2`) by reading `/proc/1/mountinfo` (falling back to `/proc/self/mountinfo`).
- **Context is lost in containers/overlays**: in Docker, Kubernetes or other overlay fs setups, users expect to see the literal mount source (`/dev/root`, `overlay`) as shown in `/proc/self/mountinfo`, not the host’s NVMe device.

These differences led to inconsistent results inside containers (see [here](https://ddstaging.datadoghq.com/notebook/12587989/diskv2-check-comparison)). By adding this new flag and mountinfo parser, we preserve the original mount source names by default, restoring parity with `psutil`, while still allowing fully-resolved host paths when desired setting `resolve_root_device` to `true`


clarkb7 (approved):
looks like issue may not be applicable to Windows
same value used for mountpoint and device in both gopsutil and python psutil

https://github.com/shirou/gopsutil/blob/ad10d4d2575ddf908388cf0c4005a07bcea8d812/disk/disk_windows.go#L151-L156

https://github.com/giampaolo/psutil/blob/3d21a43a47ab6f3c4a08d235d2a9a55d4adae9b1/psutil/arch/windows/disk.c#L325-L326

cmd/agent/dist/conf.d/disk.d/conf.yaml.default
```diff
@@ -249,4 +249,9 @@ instances:
     ## logic will be used in order: /proc/1/mountinfo → /proc/self/mountinfo → /proc/mounts.
     #
     # proc_mountinfo_path: /proc/self/mounts
+
+    ## @param preserve_root_device - boolean - optional - default: true
+    ## Instruct the check to report '/dev/root' from /proc/self/mounts instead of resolve device.
+    #
+    # preserve_root_device: true
```
pgimalac: **🥜 nitpick:** ‏What do you think of `resolve_root_device` instead ? (and swap the logic)
I think it would be clearer than "preserve root device"

pgimalac: Maybe worth mentioning this is linux only

pkg/collector/corechecks/system/disk/diskv2/disk.go
```diff
@@ -422,8 +424,21 @@ func (c *Check) collectPartitionMetrics(sender sender.Sender) error {
...
 	}
-	log.Debugf("partitions %s", partitions)
+	rootDevices := make(map[string]string)
+	if runtime.GOOS == "linux" && c.instanceConfig.PreserveRootDevice {
+		rootDevices, err = c.loadRootDevices()
+		if err != nil {
+			log.Warnf("Error reading raw devices: %s", err)
+			rootDevices = map[string]string{}
+		}
+	}
```
pgimalac: Is this ever called on Windows ? If not you can drop the `loadRootDevices` windows version

pkg/collector/corechecks/system/disk/diskv2/disk.go
```diff
@@ -422,8 +424,21 @@ func (c *Check) collectPartitionMetrics(sender sender.Sender) error {
...
+	log.Debugf("partitions '%s'", partitions)
```
pgimalac: The partitions are already logged one by one inside the loop

pkg/collector/corechecks/system/disk/diskv2/disk_nix.go
```diff
@@ -185,3 +189,249 @@ func (c *Check) sendInodesMetrics(sender sender.Sender, usage *gopsutil_disk.Usa
...
+	var lines []string
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		lines = append(lines, scanner.Text())
+	}
+	if err := scanner.Err(); err != nil {
+		return nil, err
+	}
+	return lines, nil
+}
```
pgimalac: **🥜 nitpick:** ‏We could use an iterator to avoid reading the whole file at once and returning a slice

pkg/collector/corechecks/system/disk/diskv2/disk_nix.go
```diff
@@ -185,3 +189,249 @@ func (c *Check) sendInodesMetrics(sender sender.Sender, usage *gopsutil_disk.Usa
...
+
+func readMountFile(root string) (lines []string, useMounts bool, filename string, err error) {
+	filename = path.Join(root, "mountinfo")
+	lines, err = readAllLines(filename)
+	if err != nil {
+		// if kernel does not support 1/mountinfo, fallback to 1/mounts (<2.6.26)
```
pgimalac: We don't support such old kernels anymore, Go itself supports starting kernel 3.2, so we don't need this special handling

pkg/collector/corechecks/system/disk/diskv2/disk_nix.go
```diff
@@ -185,3 +189,249 @@ func (c *Check) sendInodesMetrics(sender sender.Sender, usage *gopsutil_disk.Usa
...
+		// fallback to /proc/self
+		root = filepath.Join(hostProc, "self")
+		lines, useMounts, filename, err = readMountFile(root)
+	}
+	if err != nil {
+		return nil, fmt.Errorf("reading mount file: %w", err)
+	}
+	// Build the map
+	rootDevices := make(map[string]string)
+	log.Debugf("mountfile [%s] lines: '%s'", filename, lines)
```
pgimalac: Do we really need to log the whole file ?

pkg/collector/corechecks/system/disk/diskv2/disk_nix.go
pgimalac, line 1: **❓ question:** Could we use https://pkg.go.dev/github.com/moby/sys/mountinfo ? We're already importing it in the agent‏

pkg/collector/corechecks/system/disk/diskv2/disk.go
pgimalac, line 1: Can you add tests for that new logic ?

pkg/collector/corechecks/system/disk/diskv2/disk.go
jose-manuel-almaza: Not called in runtime but you need it for compilation time. Two different things.

cmd/agent/dist/conf.d/disk.d/conf.yaml.default
jose-manuel-almaza: Done

pkg/collector/corechecks/system/disk/diskv2/disk_nix.go
jose-manuel-almaza: I agree. I'll remove that comment but we can still see errors when reading `/proc/*/mountinfo` even on new kernels, for example:
- Permission errors in unprivileged containers or restricted environments
- I/O errors if the proc-fs is in an odd state
- Context mismatches (e.g. reading `/proc/1/mountinfo` when PID 1 is in a different namespace)

pkg/collector/corechecks/system/disk/diskv2/disk_nix.go
jose-manuel-almaza: Moved to perline-log

pkg/collector/corechecks/system/disk/diskv2/disk_nix.go
jose-manuel-almaza: The extra code we added where we read mountinfo is very specific to addressing a specific issue we're having with migrating the check from Python (psutil) to Go (gopsutil). I think using this package would obfuscate our goal.

rahulkaukuntla (approved):

pkg/collector/corechecks/system/disk/diskv2/disk.go
jose-manuel-almaza: Done

pkg/collector/corechecks/system/disk/diskv2/disk.go
pgimalac: Ah yes we're calling it from the main file
I guess we could have a `darwin` version too and drop the linux runtime check, but doesn't change much

pkg/collector/corechecks/system/disk/diskv2/disk_nix.go
pgimalac: Ah in that case it definitely makes sense !

pkg/collector/corechecks/system/disk/diskv2/disk.go
```diff
@@ -723,6 +745,8 @@ func newCheck() check.Check {
 			Timeout:              5,
 			// Match psutil exactly setting default value (https://github.com/giampaolo/psutil/blob/3d21a43a47ab6f3c4a08d235d2a9a55d4adae9b1/psutil/_pslinux.py#L1277)
 			ProcMountInfoPath: "/proc/self/mounts",
+			// Match psutil reporting '/dev/root' from /proc/self/mounts by default
+			ResolveRootDevice: true,
```
pgimalac: Should this be false by default ?

pkg/collector/corechecks/system/disk/diskv2/disk.go
```diff
@@ -111,13 +114,18 @@ func compileRegExp(expr string, ignoreCase bool) (*regexp.Regexp, error) {
 	return re, err
 }
 
+// StatFunc type alias
+type StatFunc func(path string, st *StatT) error
```
pgimalac: Can we unexport this type to avoid it being used elsewhere ?

pkg/collector/corechecks/system/disk/diskv2/disk_nix.go
```diff
@@ -8,20 +8,30 @@
...
 
 	"github.com/DataDog/datadog-agent/pkg/aggregator/sender"
 	"github.com/DataDog/datadog-agent/pkg/util/log"
 	gopsutil_disk "github.com/shirou/gopsutil/v4/disk"
+	"github.com/spf13/afero"
+	"golang.org/x/sys/unix"
 )
 
+// StatT type alias
+type StatT = unix.Stat_t
```
pgimalac: Same here

pgimalac (approved):
LGTM. I didn't compare with psutil implementation though, just reviewed the added code

pkg/collector/corechecks/system/disk/diskv2/disk.go
jose-manuel-almaza: Yes. Fixed

PR #37879, opened 2025-06-12
[AGENTRUN-346] Improve test coverage of log package

Enzu83:
### What does this PR do?

This PR adds unit tests to the `pkg/util/log` package, mostly in `log_test.go`. It also creates two test files for the log wrapper and the klog redirect logger.

More functions are properly tested. The test coverage is improved from 72.2% to 91.8%.

### Motivation

The `pkg/util/log` package wasn't properly tested:
- Some functions weren't tested.
- The test coverage for this package was too low.

The test coverage can be verified using `go test -cover ./pkg/util/log`.


pkg/util/log/log_test.go
```diff
@@ -614,3 +614,427 @@ func TestLoggerScrubbingCount(t *testing.T) {
...
+
+	Trace("message")
+	Debug("message")
+	Info("message")
+	Warn("message")
+	Error("message")
+	Critical("message")
+
+	// should print all logs except trace
+	Flush()
```
louis-cqrl: are we testing anything here ?

pkg/util/log/log_test.go
```diff
@@ -614,3 +614,427 @@ func TestLoggerScrubbingCount(t *testing.T) {
...
+	SetupLogger(Disabled(), DebugStr)
+
+	Trace("message")
+	Debug("message")
+	Info("message")
+	Warn("message")
+	Error("message")
+	Critical("message")
+
+	Flush() // should not print any logs
```
louis-cqrl: same here

Enzu83: Same thing as the other conversation, there were just prints before.
I added testing for the prints.

pkg/util/log/log_test.go
Enzu83: You're right, we weren't.
I created a function to capture the stdout and check if anything has been printed by the logger.

pkg/util/log/log_test.go
```diff
@@ -44,6 +46,21 @@ func createExtraTextContext(string) seelog.FormatterFunc {
...
+
+	f()
+
+	w.Close()
+	var buf bytes.Buffer
+	io.Copy(&buf, r)
+	os.Stdout = old
+
+	return buf.String()
+}
```
louis-cqrl: Manually redirecting os.Stdout can be fragile and affect parallel tests, moreover this isn't very idiomatic in go.

You should take a look at [logger.go](https://github.com/DataDog/datadog-agent/blob/louis-cqrl/export-started-metric/pkg/util/log/logger.go) to see if there is a better way to do it. `seelog` library also provide native way to override default logger.

louis-cqrl (changes requested):

pkg/util/log/log_test.go
Enzu83: I see. I removed the tests that called this function, as they were only verifying the behavior of the seelog Default and Disabled loggers.

louis-cqrl (approved):

PR #37884, opened 2025-06-12
Add support for live messages feature for Data Streams Monitoring

piochelepiotr:
### What does this PR do?

We are working on a new feature called "Live Messages" for Data Streams, for Kafka.
The feature will enable users to retrieve messages from Kafka on-demand, and send them as logs to Datadog.

The config provider listens to updates from remote configuration, on a new product (DSM_LIVE_MESSAGES), and sends integrations to schedule & unschedule via the stream interface.
The [Python integration](https://github.com/DataDog/integrations-core/pull/20512) then reads that configuration and logs messages.

On top of the tests present in `live_messages_test.go`, the agent is running in our staging environment at Datadog, and the feature can be used there.

I needed a way to pass the config from the agent, down to the Python check.
I explored a few ways of doing so:
1. RC loader. But checking with @GustavoCaso, it had several drawbacks:
- Added security risks by giving more access to the Datadog agent from Python checks
- Complexity in supporting an ever growing RC Loader interface
- RC loader might be deprecated in the not so far future, so adding more features to it wouldn't make sense.
2. That's why we worked on this way with @GustavoCaso , using the instance config of the kafka_consumer check. Note that once a message is retrieved, the config can be overridden without issue. The messages should only be retrieved once. So I'm not worried that the config might be overridden via remote configuration with a fresh config after the messages are retrieved.


pkg/version/base.go
```diff
@@ -22,7 +22,9 @@ var Commit string
 // used to serialize to protobuf
 var AgentPayloadVersion string
 
-var agentVersionDefault = "6.0.0"
+var agentVersionDefault = "7.66.0"
```
piochelepiotr: todo: remove that before merging

release.json
```diff
@@ -13,7 +13,7 @@
         "WINDOWS_DDPROCMON_DRIVER": "release-signed",
         "WINDOWS_DDPROCMON_VERSION": "1.2.0",
         "WINDOWS_DDPROCMON_SHASUM": "e5525959521254ba8218f64f644ad19cdeb720aa48ee83eb5e402a8d1643e32e",
-        "INTEGRATIONS_CORE_VERSION": "b5f3a28bc2f0a507fc3a563a43d1b9b861f5a1fe"
+        "INTEGRATIONS_CORE_VERSION": "3b2a6738253cd9d5e1335f6ce8a340a12c3daaae"
```
piochelepiotr: todo: remove that before merging.

drichards-87 (approved):

jesssu7 (approved):

gitstevenpham (approved):

cmd/agent/subcommands/run/command.go
```diff
@@ -588,6 +589,12 @@ func startAgent(
 	if pkgconfigsetup.IsRemoteConfigEnabled(pkgconfigsetup.Datadog()) {
 		// Subscribe to `AGENT_TASK` product
 		rcclient.SubscribeAgentTask()
+		controller := datastreams.NewController(ac)
+		subscribe := func() {
+			rcclient.Subscribe(data.ProductDataStreamsLiveMessages, controller.Update)
+		}
+		go controller.ManageSubscriptionToRC(subscribe)
```
GustavoCaso: Nit: Maybe I would pass the `rcclient` component to the NewController function, and encapsulate the logic to pull for autodiscovery into the `NewController` and we do not have to expose `ManageSubscriptionToRC` 

Something like this:

```go
controller := datastreams.NewController(ac, rcclient)
ac.AddConfigProvider(controller, false, 0)
```

And in the controller:

```go
// NewController creates a new Controller instance
func NewController(ac autodiscovery.Component, rcClient rcclient.Component) *Controller {
	c := &Controller{
		ac:            ac,
		configChanges: make(chan integration.ConfigChanges, 10),
		rcClient:      rcClient,
	}

	go c.manageSubscriptionToRC()

	return c
}
```

cmd/agent/subcommands/run/command.go
```diff
@@ -588,6 +589,12 @@ func startAgent(
...
+		ac.AddConfigProvider(controller, false, 0)
```
GustavoCaso, line 613: Pinging someone from the container platform to validate the values we use are sane. @davidor 
I'm not sure having a zero pollInterval is advised

pkg/datastreams/live_messages.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+	}
+	return false
+}
+
+// NewController creates a new Controller instance
+func NewController(ac autodiscovery.Component) *Controller {
+	return &Controller{
+		ac:            ac,
+		configChanges: make(chan integration.ConfigChanges, 10),
```
GustavoCaso, line 104: Not a blocker, but I want to know if there's a particular reason why 10 was chosen?

pkg/datastreams/live_messages.go
```diff
@@ -0,0 +1,169 @@
...
+	for _, integrationConfig := range cfgs {
+		if integrationConfig.Name != kafkaConsumerIntegrationName {
+			continue
+		}
+		configChange.Unschedule = append(configChange.Unschedule, integrationConfig)
+		updatedConfig := integrationConfig
+		updatedConfig.Instances = make([]integration.Data, 0, len(updatedConfig.Instances))
+		if updatedConfig.LogsConfig == nil {
+			updatedConfig.LogsConfig = integration.Data("[{\"type\":\"integration\",\"service\":\"kafka_consumer\",\"source\":\"kafka_consumer\"}]")
+		}
```
GustavoCaso, line 144: Should we store this in a constant?

pkg/datastreams/live_messages.go
```diff
@@ -0,0 +1,169 @@
...
+		for _, instance := range integrationConfig.Instances {
+			updatedInstance := instance
+			p := &updatedInstance
+			err := p.SetField("live_messages_configs", remoteConfigs)
+			if err != nil {
+				log.Error("Error setting field")
```
GustavoCaso:
```suggestion
				log.Error("Live messages update: Error setting field")
```

GustavoCaso: Also, for my understanding, if we can not set `live_messages_configs`, would that mean that the new integrations would not log, right?

cmd/agent/subcommands/run/command.go
piochelepiotr: 👍 Just for context, I want to use the streaming config client. So I believe this value is not used.

pkg/datastreams/live_messages.go
piochelepiotr: No reason tbh, something small enough not to use memory, but I don't want to block anything because a few config updates arrive at the same time.

pkg/datastreams/live_messages.go
piochelepiotr: Good point, updated.

pkg/datastreams/live_messages.go
piochelepiotr: Correct. Nothing will be logged in this case.

pkg/datastreams/live_messages.go
```diff
@@ -0,0 +1,176 @@
...
+	for _, config := range ac.GetAllConfigs() {
+		if config.Name == kafkaConsumerIntegrationName {
+			return true
+		}
+	}
+	return false
+}
+
+// NewController creates a new Controller instance
+func NewController(ac autodiscovery.Component, rcclient rcclient.Component) *Controller {
```
GustavoCaso: Not: We could keep the `Controller` struct private and return a `providers.ConfigProvider` 😄

KevinFairise2 (approved):

pkg/datastreams/live_messages.go
piochelepiotr: Makes sense yes, I've updated the function.

wdhif (commented):
LGTM for @DataDog/container-platform

wdhif (approved):

soberpeach (approved):

cmd/agent/subcommands/run/command.go
dustmop: https://github.com/DataDog/datadog-agent/blob/57f58d32e59cd62a2fbfdcab6f8d8da2b9bcc993/comp/core/autodiscovery/autodiscoveryimpl/autoconfig.go#L413

Looks like the method warns if interval is 0 or less, but only if `shouldPoll` is true. Since it's false here, we're all good.

dustmop (approved):

pkg/datastreams/live_messages.go
```diff
@@ -0,0 +1,181 @@
...
+	for _, config := range ac.GetAllConfigs() {
+		if config.Name == kafkaConsumerIntegrationName {
+			return true
+		}
+	}
+	return false
+}
+
+// NewController creates a new controller instance
+func NewController(ac autodiscovery.Component, rcclient rcclient.Component) providers.ConfigProvider {
```
GustavoCaso: `ConfigProvider` is in a new package `"github.com/DataDog/datadog-agent/comp/core/autodiscovery/providers/types"`

pkg/datastreams/live_messages.go
```diff
@@ -0,0 +1,181 @@
...
+// String returns the name of the provider.  All Config instances produced
+// by this provider will have this value in their Provider field.
+func (c *controller) String() string {
+	return names.DataStreamsLiveMessages
+}
+
+// GetConfigErrors returns a map of errors that occurred on the last Collect
+// call, indexed by a description of the resource that generated the error.
+// The result is displayed in diagnostic tools such as `agent status`.
+func (c *controller) GetConfigErrors() map[string]providers.ErrorMsgSet {
```
GustavoCaso: `ErrorMsgSet` is in a new package "github.com/DataDog/datadog-agent/comp/core/autodiscovery/providers/types"

pkg/datastreams/live_messages.go
```diff
@@ -0,0 +1,181 @@
...
+	return nil
```
GustavoCaso: I would return an empty map rather than `nil`. We never know what the callers of  `GetConfigErrors` will do, sending an empty map `map[string]providers.ErrorMsgSet{}`, warranty that we do not panic in the future

GustavoCaso (approved):
Amazing work 🎉 
I left a few comments that require addressing before merging, outside of that, excited to see this working in production

PR #37886, opened 2025-06-12
Avoid bazel symlinks in modules.validate task

CelianR:
### What does this PR do?

`glob` follows symlinks and makes this task to fail when using bazel, ignored bazel directories.

### DIFF
--- a/tasks/modules.py
+++ b/tasks/modules.py
@@ -171,6 +171,10 @@ def validate(ctx: Context, base_dir='.', fix_format=False):
 
     # Backward check for go.mod (ensure there is a module for each go.mod)
     for go_mod in glob(str(base_dir / '**/go.mod'), recursive=True):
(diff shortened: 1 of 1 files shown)


tasks/modules.py
```diff
@@ -171,6 +171,10 @@ def validate(ctx: Context, base_dir='.', fix_format=False):
 
     # Backward check for go.mod (ensure there is a module for each go.mod)
     for go_mod in glob(str(base_dir / '**/go.mod'), recursive=True):
+        # Ignore bazel generated symlinks
+        if go_mod.startswith(str(base_dir / 'bazel')):
```
KevinFairise2, line 175: How do you end up with a `go.mod` file inside a bazel directory?

KevinFairise2 (approved):

PR #37889, opened 2025-06-12
Add .log extension on install-info file in flare

pgimalac:
### What does this PR do?
Add .log extension on install-info file in flare.

### Motivation
JIRA refuses to open files without known file type, in particular `install_info`. Adding a `.log` should fix that.

CI


dustmop (approved):
Nice catch, thanks for the fix!

PR #37903, opened 2025-06-12
fix password prompt during kmt.launch-stack

brycekahle:
### What does this PR do?

Fixes a problem with `kmt.launch-stack` where the `Enter Password:` prompt was not visible. This was due to nesting of running an invoke command within an invoke task. Replaced with a direct call instead.

### Motivation

Confusing user experience, thinking the command was hanging.


ofek (approved):

usamasaqib (approved):

mbertrone (approved):

usamasaqib:
@brycekahle is this good to go?

PR #37905, opened 2025-06-12
[CWS] fix fentry tail call detection by looking at kernel BTF

paulcacheux:
### What does this PR do?

In kernel 6.11 and higher, [fentry had a breaking change](https://github.com/torvalds/linux/commit/28ead3eaabc16ecc907cfb71876da028080f6356) around tail calls that break our support. We merged https://github.com/DataDog/datadog-agent/pull/36918 to detect this, but since the breaking change is getting backported ( :( ) this PR improves the detection by using BTF kernel data.

### DIFF
--- a/pkg/security/probe/constantfetch/available.go
+++ b/pkg/security/probe/constantfetch/available.go
@@ -104,3 +104,41 @@ func GetBTFFunctionArgCount(funcName string) (int, error) {
 
 	return len(proto.Params), nil
 }
+
+// AreFentryTailCallsBroken checks if fentry tail calls are broken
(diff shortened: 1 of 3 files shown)


YoannGh (approved):
@DataDog/ebpf-platform might want to move the `AreFentryTailCallsBroken` function to their [pkg/ebpf/kernel_bugs.go](https://github.com/DataDog/datadog-agent/blob/main/pkg/ebpf/kernel_bugs.go) file so that it can be used by other ebpf users, but this can be done in a follow up PR

PR #37906, opened 2025-06-12
[CWS] Implement fsmount syscall (fix runc init path resolution)

kovagsm:
### What does this PR do?

Adds support for the `fsmount` syscall, both to the mount resolver and also creates fsmount event. 

### Motivation

This syscall is used extensively by programs such as `runc` and the lack of coverage was causing multiple mountpoint resolution errors, that resulted in abnormal paths.

* Manually, locally
* Deployed on a cluster
* Automated tests

* N/A

### Additional Notes

This syscall is relatively new and documentation is lacking


maycmlee (approved):

pkg/security/ebpf/c/include/hooks/mount.h
```diff
@@ -407,4 +407,78 @@ TAIL_CALL_TRACEPOINT_FNC(handle_sys_mount_exit, struct tracepoint_raw_syscalls_s
...
+    };
+
+    cache_syscall(&syscall);
+
+    return 0;
+}
+
+
+HOOK_SYSCALL_EXIT(fsmount) {
+    struct syscall_cache_t *syscall = peek_syscall(EVENT_FSMOUNT);
```
safchain: nit: pop instead of peek

paulcacheux: it's not a nit, it's pretty important to not fill the syscall cache

paulcacheux: so yeah good catch !

pkg/security/serializers/serializers_linux.go
```diff
@@ -495,6 +495,17 @@ type MountEventSerializer struct {
 	MountSourcePathResolutionError string `json:"source.path_error,omitempty"`
 }
 
+// FsmountEventSerializer serializes an fsmount event to JSON
+// easyjson:json
+type FsmountEventSerializer struct {
```
safchain: nit: you can remove it

pkg/security/tests/fsmount_test.go
```diff
@@ -0,0 +1,107 @@
...
+		_ = fsconfig(fsfd, unix.FSCONFIG_CMD_CREATE, nil, nil, 0)
+
+		mountfd, err := unix.Fsmount(fsfd, 0, 0)
+		if err != nil {
+			assert.Fail(t, "fsmount failed")
+			return
+		}
+		defer unix.Close(mountfd)
+
+		time.Sleep(500 * time.Millisecond)
```
safchain: nit: maybe useGetProbeEvent

pkg/security/ebpf/c/include/hooks/mount.h
kovagsm: addressed

pkg/security/serializers/serializers_linux.go
kovagsm: addressed

pkg/security/tests/fsmount_test.go
kovagsm: addressed

pkg/security/probe/probe_ebpf.go
```diff
@@ -1060,6 +1060,9 @@ func (p *EBPFProbe) handleEvent(CPU int, data []byte) {
 			seclog.Errorf("failed to insert exec event: %s (pid %d, offset %d, len %d)", err, event.PIDContext.Pid, offset, len(data))
 			return
 		}
+
+		event.Exec.FileEvent.PathnameStr = p.fieldHandlers.ResolveFilePath(event, &event.Exec.FileEvent)
```
safchain: Why do you need to call this two function here, the path should have been resolved in the AddExecEntry ?

kovagsm: Not needed. It was an attempt to solve some still failing resolutions for `runc init` in the case it was a race condition. I removed it

pkg/security/resolvers/mount/resolver.go
```diff
@@ -372,6 +372,11 @@ func (mr *Resolver) _getMountPath(mountID uint32, device uint32, pid uint32, cac
 	}
 	cache[mountID] = true
 
+	if mount.Origin == model.MountOriginFsmount {
+		// Detached mount
+		return "/", source, mount.Origin, nil
```
spikat, line 377: How should we notify the user that this is not the "real" / ? WDYT of returning something like "detached:/" (except the terrible name :)) ?

pkg/security/tests/fsmount_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+			mountfd, err := unix.Fsmount(fsfd, 0, 0)
+			if err != nil {
+				return err
+			}
+			defer unix.Close(mountfd)
+
+			return nil
+		}, func(event *model.Event) bool {
+			assert.NotEqual(t, event.Fsmount.MountID, 0, "Mount id is zero")
```
spikat: Should we try to test other fields here, like flags, mountAttrs etc?

pkg/security/tests/fsmount_test.go
spikat, line 1: IMHO we should also try to test the create of a directory+file in the created tmpfs, and then check that if we open the file the resolution ~"worked", WDYT?

pkg/security/resolvers/mount/resolver.go
kovagsm: I had a discussion with @safchain today about this, and we agreed to have a boolean field saying whether it's detached or not, this way it's easier to write rules.

pkg/security/tests/fsmount_test.go
kovagsm: added

pkg/security/tests/fsmount_test.go
kovagsm: yup, good suggestion. added!

pkg/security/secl/model/unmarshallers_linux.go
```diff
@@ -476,6 +476,46 @@ func (e *MountEvent) UnmarshalBinary(data []byte) (int, error) {
...
+		return n, ErrNotEnoughData
+	}
+
+	e.Fd = int32(binary.NativeEndian.Uint32(data[0:4]))
+	e.Flags = binary.NativeEndian.Uint32(data[4:8])
+	e.Device = binary.NativeEndian.Uint32(data[8:12])
+	e.MountAttrs = binary.NativeEndian.Uint32(data[12:16])
+	e.MountID = e.RootPathKey.MountID
+	read += 16
+	fmt.Printf("EVENT %+v\n", e)
```
spikat: deub log to be removed

pkg/security/tests/fsmount_test.go
```diff
@@ -0,0 +1,157 @@
...
+			err = cmd.Run()
+
+			if err != nil {
+				return err
+			}
+			defer unix.Close(mountfd)
+
+			return nil
+		}, func(event *model.Event, _ *rules.Rule) {
+			fmt.Printf("%+v\n", event.Fsmount)
```
spikat: debug log to be removed

pkg/security/tests/fsmount_test.go
```diff
@@ -0,0 +1,157 @@
...
+			err = cmd.Run()
+
+			if err != nil {
+				return err
+			}
+			defer unix.Close(mountfd)
+
+			return nil
+		}, func(event *model.Event, _ *rules.Rule) {
+			fmt.Printf("%+v\n", event)
```
spikat: and another one :)

spikat (approved):

PR #37911, opened 2025-06-12
[dyninst] Support probe throttling

piob-io:
### What does this PR do?

Introduce per-probe throttling. To prevent needless probe execution, we need to perform throttling inside the eBPF program.

#### Requirements

We want to be able to limit probe execution on roughly 1 second time scale. Practically snapshots are limited to 1ps, no-snapshot logs to 5kps, and metric probes (we don't support these yet) are unlimited.

Most performance sensitive scenario is a probe injected into busyloop, executed on every core. The cheapest eBPF program execution path should be throttler rejection. Acceptance path should be cheap, but it's less of a concern given the cost of executing the probe itself. Both paths must minimize contention, as probes may be executed across cores (and currently, with exception of output ringbuffer, are fully independent).

eBPF programs cannot block nor wait. Limited (less than 1%) over-throttling is allowed.

#### Solution

We introduce following throttling schema. There are two parameters - period and budget. Whenever "budget" probes are executed, we throttle following probes. We refresh the "budget" after "period" elapsed since previous refresh.

Throttle-reject fast path involves checking if budget is non-positive, and whether period has not elapsed since previous refresh. It only involves memory reads, causing no contention.

Throttle-accept fast path involves atomically decrementing budget variable. This should be acceptable, as it will cause no more contention than submitting output to the ringbuf. There is an implementation approach that would eliminate contention, where each core reserves fraction of the budget for itself, but the complexity outweighs the benefits.

Slow path involves refreshing budget variable. It requires compare-and-swap to only refresh the budget once within a period. We make an assumption that ktime_ns reads practically always return different values. Compare-and-swap may fail, in which case we repeat whole throttling evaluation procedure again, and if it fails again, we conservatively make reject decision. Two consecutive compare-and-swap failures imply that budget was expired "immediately" between previous refresh and probe retry. In which case the conservative rejection is correct. We may over-throttle - any probe that checks the budget between the compare-and-swap on the last refresh variable, before budget is refreshed, is rejected. Practically, this requires probing a hot function - to cause a race, in which case incorrectly throttling one call, will allow some future call to stay within probing budget.

To provide throttling metric, eBPF programs has per-cpu throttled_events counter (unsynchronized memory write).

See implementation in ebpf/throttler.h.

Benchmarks, using kmt amd64 ubuntu24 machine, with 4 vcpus. We run busyloops on 4 threads, and apply 500 probes per 100ms limit. We measure number of busyloops iterations in twenty 10s rounds, to calculate busyloop single iteration latency, which is practically the cost of ebpf program execution (the busyloop itself only performs function call, to allow for an easy injection point).

The no-op ebpf program costs (= interrupt latency): 
avg=0.604338us stddev=0.000763us

Ebpf program with only thottler (no stack machine execution, nothing else):
avg=0.623178us stddev=0.002300us 

Giving ~19ns throttling latency. Nota bene, ~15ns of that comes from reading ktime (for ebpf latency expectations, see: https://datadoghq.atlassian.net/wiki/spaces/~712020ffd5f306baa1460395eea88b46dbe046/pages/5111545950/Ebpf+low-level+benchmarks). 

### Motivation

https://datadoghq.atlassian.net/browse/DEBUG-4012

### DIFF
--- a/pkg/dyninst/ebpf/event.c
+++ b/pkg/dyninst/ebpf/event.c
@@ -10,14 +10,22 @@
 #include "murmur2.h"
 #include "walk_stack.h"
 #include "scratch.h"
+#include "throttler.h"
 
 char _license[] SEC("license") = "GPL";
 
 extern const probe_params_t probe_params[];
 extern const uint32_t num_probe_params;
 
+struct {
+  __uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+  __uint(max_entries, 1);
+  __type(key, uint32_t);
+  __type(value, uint64_t);
+} throttled_events SEC(".maps");
+
 SEC("uprobe") int probe_run_with_cookie(struct pt_regs* regs) {
-  uint64_t start = bpf_ktime_get_ns();
+  uint64_t start_ns = bpf_ktime_get_ns();
 
   const uint64_t cookie = bpf_get_attach_cookie(regs);
   if (cookie >= num_probe_params) {
@@ -25,6 +33,14 @@ SEC("uprobe") int probe_run_with_cookie(struct pt_regs* regs) {
   }
   const probe_params_t* params = &probe_params[cookie];
 
+  if (should_throttle(params->throttler_idx, start_ns)) {
+    uint32_t zero = 0;
+    uint64_t* cnt = bpf_map_lookup_elem(&throttled_events, &zero);
+    if (cnt) {
+      ++*cnt;
+    }
+    return 0;
+  }
   global_ctx_t global_ctx;
   global_ctx.stack_machine = stack_machine_ctx_load();
   if (!global_ctx.stack_machine) {
@@ -52,7 +68,7 @@ SEC("uprobe") int probe_run_with_cookie(struct pt_regs* regs) {
   *header = (event_header_t){
       .data_byte_len = sizeof(event_header_t),
       .stack_byte_len = 0, // set this if we collect stacks
-      .ktime_ns = start,
+      .ktime_ns = start_ns,
   };
 
   __maybe_unused int process_steps = 0;
@@ -91,7 +107,6 @@ SEC("uprobe") int probe_run_with_cookie(struct pt_regs* regs) {
     stack_hash = 0;
   }
   global_ctx.regs = &global_ctx.stack_walk->regs;
-
   frame_data_t frame_data = {
 // Stack layout is slightly different in Go between arm64 and x86_64.
 // Established based on following documentation and machine code reads:
--- a/pkg/dyninst/testprogs/probes.go
+++ b/pkg/dyninst/testprogs/probes.go
@@ -9,11 +9,14 @@ package testprogs
 
 import (
 	"encoding/json"
+	"fmt"
 	"os"
 	"path"
 	"testing"
 
 	"gopkg.in/yaml.v3"
+
+	"github.com/stretchr/testify/require"
 
 	"github.com/DataDog/datadog-agent/pkg/dyninst/config"
 )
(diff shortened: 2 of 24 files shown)


pkg/dyninst/testprogs/probes.go
```diff
@@ -24,31 +24,39 @@ type probeYaml struct {
 }
 
 // GetProbeCfgs returns the probe configurations for binary of a given name.
-func GetProbeCfgs(t *testing.T, name string) []config.Probe {
+func GetProbeCfgs(name string) ([]config.Probe, error) {
```
ajwerner, line 37: nit: what if we add a `Must` form of each of these that takes a testing.T and asserts no error. This is pretty common elsewhere as a pattern. It could also use a generic helper :shrug: 

```go
func must[V any](t testing.TB, f func() (V, error)) V {
    v, err := f()
    require.NoError(t, err)
    return v
}

func MustGetProbeCfgs(t testing.TB, name string) []config.Probe {
    return must(t, func() ([]config.Probe, error) {return GetProbeCfgs(name) })
}

func MustGetCommonConfigs(t testing.TB) []Config {
   return must(t, GetCommonConfigs)
}
// ...
```

pkg/dyninst/testutil/util.go
```diff
@@ -0,0 +1,161 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+//go:build linux_bpf
+
+// Package testutil provides utilities for dyninst integration testing.
+package testutil
```
ajwerner: To me there's something odd about a package called testutil depending on like everything... I think of testutil as usually depending on almost nothing. `PrepTmpDir` and `StartProcess` seem fine, but the others feel like they belong somewhere else. This comment is sort of moot though, because I hope after [#37841](https://github.com/DataDog/datadog-agent/pull/37841) the actuator is what you'd use. If it's not and we want to keep these around, what about yet another package. Maybe `integrationtest` even `dyninsttest` (drawing on inspiration from https://pkg.go.dev/net/http/httptest)?

pkg/dyninst/ebpf/event.c
```diff
@@ -10,21 +10,37 @@
...
   if (cookie >= num_probe_params) {
     return 0;
   }
   const probe_params_t* params = &probe_params[cookie];
 
+  if (should_throttle(params->throttler_idx, start_ns)) {
+    uint32_t zero = 0;
+    uint64_t* cnt = bpf_map_lookup_elem(&throttled_events, &zero);
+    if (cnt) {
+    ++*cnt;
```
ajwerner: nit: indentation

ajwerner (approved):
Neat!

pkg/dyninst/testutil/util.go
piob-io: Well, I think yes, more commonly there is a test util for unit tests, that's low-level. But it's useful to have testutils for integration tests sometimes too. We should definitely use actuator in the integration_test, but I think there will be tests, like throttler test, that orchestrate the situation more directly - e.g throttler test needs to modify throttling params, and I prefer not to allow injection to the production code just because test needs it, I rather have a test execute what it needs at the level it needs. I like the dyninsttest (somehow despite the repeated word in code path :P), moved.

pkg/dyninst/ebpf/event.c
piob-io: Done

pkg/dyninst/testprogs/probes.go
piob-io: Done, without the generic, I think it's not worth the readability hit.

piob-io:
Thank you!

PR #37919, opened 2025-06-13
chore: Datadog URL connectivity

san-jos:
### What does this PR do?
Add connectivity checks to the inventory payload.

image

### Motivation
[FA-1019](https://datadoghq.atlassian.net/browse/FA-1019)

Unit / Manual

### Additional Notes

[FA-1019]: https://datadoghq.atlassian.net/browse/FA-1019?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


comp/connectivitychecker/def/component.go
pgimalac, line 1: How different is that from the diagnose connectivity check ?
Could we reuse code ? (eg. avoid duplicating the list of endpoints)

san-jos: The need is to have light connectivy checks has they are going to be run every 10 mins covering some checks covered in the core endpoint, some in the evp but not all of them.
I'll check what can be reused

comp/connectivitychecker/fx/fx.go
```diff
@@ -0,0 +1,34 @@
...
+		),
+		fxutil.ProvideOptional[connectivitychecker.Component](),
+
+		// connectivitychecker is a component with no public method, therefore nobody depends on it and FX only instantiates
+		// components when they're needed. Adding a dummy function that takes our Component as a parameter forces
+		// the instantiation of ssistatus. This means that simply using 'ssistatusfx.Module()' will run our
+		// component (which is the expected behavior).
+		//
+		// This prevents silent corner case where including 'ssistatus' in the main function would not actually
+		// instantiate it. This also removes the need for every main using ssistatus to add the line below.
```
pgimalac: I know this is copy pasted in various components, we could try to rephrase to avoid this copy pasting issue 😄 
```suggestion
		// connectivitychecker is a component with no public method, therefore nobody depends on it and FX only instantiates
		// components when they're needed. Adding a dummy function that takes our Component as a parameter forces
		// the instantiation of connectivitychecker. This means that simply using 'connectivitycheckerfx.Module()' will run our
		// component (which is the expected behavior).
		//
		// This prevents silent corner case where including 'connectivitychecker' in the main function would not actually
		// instantiate it. This also removes the need for every main using connectivitychecker to add the line below.
```

comp/connectivitychecker/impl/connectivitychecker.go
```diff
@@ -0,0 +1,145 @@
...
+	default:
+		// Continue with setting data
+	}
+
+	// Send results to inventory agent
+	c.inventoryAgent.Set("connectivity", diagnoses)
+	c.log.Debug("Connectivity check completed successfully")
+}
+
+func (c *inventoryImpl) Start(_ context.Context) error {
```
pgimalac: Does it need to be exported ?

comp/connectivitychecker/impl/connectivitychecker.go
```diff
@@ -0,0 +1,145 @@
...
+	c.startTimer(initialDelay)
+	return nil
+}
+
+func (c *inventoryImpl) Stop(_ context.Context) error {
```
pgimalac: Same

comp/connectivitychecker/impl/connectivitychecker.go
```diff
@@ -0,0 +1,145 @@
...
+)
+
+const (
+	initialDelay = 30 * time.Second
+	interval     = 10 * time.Minute
+)
+
+// Requires defines the dependencies for the connectivitychecker component
+type Requires struct {
+	// Remove this field if the component has no lifecycle hooks
```
pgimalac:
```suggestion
```

comp/connectivitychecker/mock/mock.go
pgimalac, line 1: The component interface is empty so a mock is probably not needed ?

comp/connectivitychecker/def/component.go
pgimalac: Are we fine with all agents making requests to that set of endpoints every 10 minutes ? (in particular the POST requests)

pgimalac (approved):

comp/connectivitychecker/def/component.go
san-jos: There is no post request. Only head and get.

coignetp (approved):

pkg/diagnose/connectivity/http.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/DataDog/datadog-agent/comp/core/config"
+	log "github.com/DataDog/datadog-agent/comp/core/log/def"
+	forwarder "github.com/DataDog/datadog-agent/comp/forwarder/defaultforwarder"
+	"github.com/DataDog/datadog-agent/pkg/util/scrubber"
+)
+
+type method string
+
+const (
+	head method = "HEAD"
```
dustmop: I'd recommend you use the net/http constants instead: https://pkg.go.dev/net/http#pkg-constants

pkg/diagnose/connectivity/hooks.go
```diff
@@ -26,10 +26,24 @@ import (
 // During a request, the http.Client will call the functions of the ClientTrace at specific moments
 // This is useful to get extra information about what is happening and if there are errors during
 // connection establishment, DNS resolution or TLS handshake for instance
-func createDiagnoseTraces(httpTraces *[]string) *httptrace.ClientTrace {
+func createDiagnoseTraces(httpTraces *[]string, errorOnly bool) *httptrace.ClientTrace {
```
dustmop, line 30: Could you add a line documenting how `errorOnly` works here?

pkg/diagnose/connectivity/http.go
```diff
@@ -0,0 +1,112 @@
...
+func sendGet(ctx context.Context, client *http.Client, url string, headers map[string]string) (statusCode int, body []byte, logURL string, err error) {
+	return sendRequest(ctx, client, url, get, nil, headers)
+}
+
+// sendHTTPRequestToEndpoint sends an HTTP Request with the method and payload inside the endpoint information
+func sendPost(ctx context.Context, client *http.Client, url string, payload []byte, headers map[string]string) (statusCode int, body []byte, logURL string, err error) {
+	return sendRequest(ctx, client, url, post, payload, headers)
+}
+
+// sendRequest
```
dustmop: This seems like there was about to be a comment here, could you include one?

pkg/diagnose/connectivity/inventoryendpoint.go
```diff
@@ -0,0 +1,343 @@
...
+	pkgconfigsetup "github.com/DataDog/datadog-agent/pkg/config/setup"
+	"github.com/DataDog/datadog-agent/pkg/config/utils"
+)
+
+type separator string
+
+// URL types for different services
+const (
+	dot  separator = "."
+	dash separator = "-"
```
dustmop: I think the need for these two separators is causing more complexity here than it is worth. Only `browser-intake-` is using the dash. At the same time, some prefixes end in a dot (like `"llmobs-intake."`) while others (like `
`"orchestrator") don't. Although that gets [handled here](https://github.com/DataDog/datadog-agent/pull/37919/files#diff-820d736bbf58fd9715c761e38cdb20eab5ec137f14efa72be2a15718494e2820R168), the logic is difficult to understand and validate. I would recommend the following cleanup:

1) Add a field called `routeDashPrefix` to handle `browser-intake` and its need for special casing. Get rid of the `separator` and use the dot everywhere else.
2) Remove the trailing dot everywhere for `routePrefix` and `configPrefix`. It isn't necessary because the code is handling the dot separator, and the inconsistency is confusing.
3) Add a method called `joinPrefix` or `joinSubdomain` or something like that. Use that to join a prefix and a domain string, adding the dot if necessary.

pkg/diagnose/connectivity/inventoryendpoint.go
```diff
@@ -0,0 +1,343 @@
...
+		{routePrefix: "intake.profile", routePath: "probe", method: get},
+		{routePrefix: "app", routePath: "probe", versioned: true, method: get},
+	}
+}
+
+type resolvedEndpoint struct {
+	url           string
+	base          string
+	method        method
+	apiKey        string
```
dustmop, line 71: This won't function properly due to the possibility of an [API Key refresh](https://docs.datadoghq.com/agent/configuration/secrets-management/?tab=linux#refreshing-apiapp-keys-at-runtime). The agent has the ability to refresh API keys, either manually or automatically by refreshing at a regular interval. As a result, copying the string of the API Key means that running `datadog-agent secrets refresh` will cause this diagnostic to start failing. As a fix, you can follow our [recommendations in the docs](https://datadoghq.atlassian.net/wiki/spaces/ACFG/pages/4906156042/API+APP+key+refresh#How-to-handle-API/APP-key-refresh-in-your-product), to either 1) use a functor instead such that the key is never stored and instead will always get fetched from the config object, or 2) register a listener with `config.OnUpdate` to trigger a callback when the key is changed.

pkg/diagnose/connectivity/inventoryendpoint.go
```diff
@@ -0,0 +1,343 @@
...
+	return "Success", nil
+}
+
+func isSuccessStatusCode(endpoint resolvedEndpoint, statusCode int) bool {
+	switch endpoint.method {
+	case head:
+		if statusCode == http.StatusTemporaryRedirect || statusCode == http.StatusPermanentRedirect {
+			return endpoint.limitRedirect
+		}
+		return statusCode >= http.StatusOK && statusCode < http.StatusMultipleChoices
```
dustmop: If you want to check for a `2xx` code, I might suggest using 300 instead of `StatusMultipleChoices`. I'm not sure the constant helps the understandability of this code.

pkg/diagnose/connectivity/inventoryendpoint.go
```diff
@@ -0,0 +1,343 @@
...
+		return apiKey
+	}
+	return defaultAPIKey
+}
+
+type domain struct {
+	site            string
+	mainAPIKey      string
+	infraEndpoint   string
+	useCustomAPIKey bool
```
dustmop: Very minor naming nit: I feel like "custom" here is a bit-misleading. We want a name that just indicates that it's not the "main" API Key but rather something used by an additional endpoint (such as a region failover or dual shipping). How about `useAltAPIKey`?

dustmop (commented):
This is great! Thanks for working on combining parts with the `diagnose` command as @pgimalac suggested. Before approving I just have some comments, particularly around how this interacts with the api key refresh feature.

pkg/diagnose/connectivity/inventoryendpoint.go
dustmop: Spoke offline and my comment here is not actually applicable. These data structures are built fresh every time this connectivity check is run, so there's no need to specially handle an update from the config.

dustmop (approved):
Thanks for the fixes! Approving

coignetp (approved):

PR #37921, opened 2025-06-13
WLM: process: add Service Discovery logic

Yumasi:
### What does this PR do?

This PR adds Service Discovery info collection to the recently added Process collector. The collector now uses the new `/services` endpoint to fetch SD info about running services, and store it in a new Service field of the Process entity. The collector also performs part of the previous SD logic with handling of retries on PIDs that do have Service info, and heartbeats for Service updates.

### Motivation

DSCVR-141

Change covered by unit tests, and some manual testing


comp/core/workloadmeta/def/types.go
```diff
@@ -1196,6 +1201,63 @@ func printHistory(out io.Writer, history *v1.History) {
...
+	// AdditionalGeneratedNames contains other potential names for the service
+	AdditionalGeneratedNames []string
+
+	// ContainerServiceName is the service name from container tags
+	ContainerServiceName string
+
+	// ContainerServiceNameSource indicates the source of the container service name
+	ContainerServiceNameSource string
+
+	// ContainerTags are the container tags associated with the service
```
gabedos: Are these values the container tags from the tagger? I think it might be better practice to fetch those tags from the tagger component directly using the containerID you already have stored in this struct here. This will avoid duplicating data and also ensuring it's consistent across the Agent.

There could be instances in which container tags are updated in the tagger even if there are no actual changes to the container itself. One instance is if the container runtime collector processes the container first and gets the tags from there, and then at a later time the kubelet collector processes the container adding other k8s related tags that may come from pod-level attributes like annotations. A user could also update a pod's annotations in place which could change tags. Albeit these are pretty rare conditions.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -203,6 +242,198 @@ func (c *collector) detectLanguages(processes []*procutil.Process) []*languagemo
...
+		DDService:                  modelService.DDService,
+		DDServiceInjected:          modelService.DDServiceInjected,
+		CheckedContainerData:       modelService.CheckedContainerData,
+		Ports:                      modelService.Ports,
+		APMInstrumentation:         modelService.APMInstrumentation,
+		Language:                   modelService.Language,
+		Type:                       modelService.Type,
+		CommandLine:                modelService.CommandLine,
+		StartTimeMilli:             modelService.StartTimeMilli,
+		ContainerID:                modelService.ContainerID,
```
vitkyrka: Can you remove all the fields which are unused or duplicated with what's already in Process? CheckedContainerData, ContainerID, StartTimeMilli, Language, etc.

comp/core/workloadmeta/def/types.go
```diff
@@ -1252,7 +1325,9 @@ func (p Process) String(_ bool) string {
 	if p.Language != nil {
 		_, _ = fmt.Fprintln(&sb, "Language:", p.Language.Name)
 	}
-
+	if p.Service != nil {
+		_, _ = fmt.Fprintln(&sb, "Service:", p.Service.GeneratedName)
```
vitkyrka: Maybe print all fields?

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -203,6 +242,198 @@ func (c *collector) detectLanguages(processes []*procutil.Process) []*languagemo
...
+		service := pidsToService[pid]
+		if service == nil {
+			c.handleServiceRetries(pid)
+			continue
+		}
+
+		// Update the last heartbeat to current time
+		service.LastHeartbeat = now.Unix()
+
+		entity := &workloadmeta.Process{
```
vitkyrka, line 337: Can this result in a WLM Process entry getting created without all the other fields of Process (eg cmdline) filled in properly?

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -233,12 +464,30 @@ func (c *collector) collect(ctx context.Context, collectionTicker *clock.Ticker)
...
+			wlmServiceEntities := c.updateServices(alivePids)
+			if len(wlmServiceEntities) > 0 {
+				c.processEventsCh <- &Event{
+					Type:    EventTypeServiceDiscovery,
+					Created: wlmServiceEntities,
+				}
+			}
+
+			cleanPidMaps(alivePids, c.ignoredPids)
+			cleanPidMaps(alivePids, c.serviceRetries)
```
gengnamstyle: Currently service-discovery has a collection interval of 60s by default and we have a 10s collection interval by default.

I don't recall service-discovery planning to increase their collection frequency to 10s, and I'm unsure of what the performance implications of collecting 6x as frequent.

As a result, I currently lean towards moving this piece of code into a separate `collect` function and perhaps renaming this function. Just throwing it out there, perhaps?

`collectProcesses`
`collectServices`
 
 and `collectServices` can read from `c.lastCollectedProcesses` instead of re-querying with 
 ```
 procs, err := c.processProbe.ProcessesByPID(c.clock.Now(), false)
 ```
 
 after access to `c.lastCollectedProcesses` is thread safe. 
 
I think the caveat of this approach is that depending on future configuration there could be a case where "process collection is disabled "and we only want "service discovery enabled". In this case we may need some more checks that allow `collectServices` to query processes.

comp/core/workloadmeta/def/types.go
Yumasi: Oh, that's a mistake on my part, we should not store these in WLM indeed. I'll remove those fields

comp/core/workloadmeta/collectors/internal/process/process_collector.go
Yumasi: If there is no Process entry in WLM, then yes. One way I could see this happen is if the `stream` goroutine process the SD event before the one from process collection. In this case, for a brief moment, we would be in the state you describe until the event from process collection is processed by the `stream` goroutine, which will then merge the two Process entities.

This PR does not handle the case where process collection is disabled though. In that case, our part of the collector will need to fill up the rest of the fields we care about, but I'd prefer doing so in a separate PR.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
Yumasi: I think that's a good idea, and I have done so in the latest commit.

> I think the caveat of this approach is that depending on future configuration there could be a case where "process collection is disabled "and we only want "service discovery enabled". In this case we may need some more checks that allow collectServices to query processes.

Yes, there is still the case of service discovery only that needs to be solve, but I was thinking of doing so in a separate PR.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -481,10 +485,12 @@ func (c *collector) collectServices(ctx context.Context, collectionTicker *clock
 		select {
 		case <-collectionTicker.C:
 			// Get alive PIDs from last collected processes
+			c.mux.RLock()
 			alivePids := make(core.PidSet, len(c.lastCollectedProcesses))
 			for pid := range c.lastCollectedProcesses {
```
smeso-ddg, line 488: Isn't it possible for `lastCollectedProcesses` to be nil here?

comp/core/workloadmeta/collectors/internal/process/process_collector.go
gengnamstyle: awesome yeah I agree.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -10,29 +10,45 @@ package process
...
 )
 
 const (
-	collectorID       = "process-collector"
-	componentName     = "workloadmeta-process"
-	cacheValidityNoRT = 2 * time.Second
+	collectorID               = "process-collector"
+	componentName             = "workloadmeta-process"
+	cacheValidityNoRT         = 2 * time.Second
+	serviceCollectionInterval = 60 * time.Second
```
gengnamstyle: nit: potentially adding a comment here to indicate that this is likely to become a configuration in the future?

comp/core/workloadmeta/collectors/internal/process/process_collector.go
Yumasi: It is possible. It should not cause any panic the way its written now, but now that I think about it, it will do a small allocation for alivePids for nothing. I'll add an early nil check to prevent that. Thanks!

smeso-ddg: What I mean is that, if this code happen to acquire the lock before collectProcesses replaces/set lastCollectedProcesses, your changes here will be discarded by collectProcesses. Or am I missing something?

gengnamstyle: I don't seem to see why the "changes" (results?) would be discarded as you suggest? 

I agree that it is possible that `collectServices` acquires the lock before `collectProcesses` sets `lastCollectedProcesses`, but I think this only leads to potentially collecting services on 10 second old process data (assuming we are collecting processes every 10seconds)

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -92,7 +134,7 @@ func GetFxOptions() fx.Option {
 func (c *collector) isEnabled() bool {
 	// TODO: implement the logic to check if the process collector is enabled based on dependent configs (process collection, language detection, service discovery)
 	// hardcoded to false until the new collector has all functionality/consolidation completed (service discovery, language collection, etc)
-	return false
+	return true
```
daniel-taf: I don't think we want this to be `true` yet.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
daniel-taf: For a process to be a service, does it need to be alive for a certain amount of time? I remember a long time that used to be a requirement. If that is the case, then it seems this situation couldn't happen.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
Yumasi: > What I mean is that, if this code happen to acquire the lock before collectProcesses replaces/set lastCollectedProcesses, your changes here will be discarded by collectProcesses. Or am I missing something?

So, normally, the Merge function should take care of this, as it should be called when an entity is set/updated in the store. `collectProcesses` won't have a service field set in its event, in which case the merge function should preserve the one in the store.

smeso-ddg: I think you can discard my comment then :)

comp/core/workloadmeta/collectors/internal/process/process_collector.go
Yumasi: It does need to be alive for one minute. The check could see an entry in WLM with only the Service info filled, but it would not yet consider it a service, but a potential service (meaning no event send to the backend yet). So the next check iteration should have access to a filled up Entity.

Yumasi: It does, but this handled in the check, so it is possible that it sees the entity with only the Service fields set.

Yumasi: But actually, we'll probably change that and just have `collectServices` check the start time in the filtering of PIDs to collect, and only ask SP to collect info for processes older than a minute. That will be in a later PR though.

daniel-taf (approved):

comp/core/workloadmeta/def/types.go
```diff
@@ -1196,6 +1201,39 @@ func printHistory(out io.Writer, history *v1.History) {
...
+	Ports []uint16
+
+	// APMInstrumentation indicates the APM instrumentation status
+	APMInstrumentation string
+
+	// Type is the service type (e.g., "web_service")
+	Type string
+
+	// LastHeartbeat is the timestamp of the last heartbeat update (Unix seconds)
+	LastHeartbeat int64
```
gabedos: This value is updated every time the service collector executes (every 60 seconds). This would then cause wmeta to detect a "change" in the process entity when nothing _actually_ changed and all subscribers would be notified.

What is the reason behind storing the lastHeartbeat of the service? Would it be possible to only update this value when there is also a real change the service entities?

Yumasi: > This value is updated every time the service collector executes (every 60 seconds).

`collectServices` should only update entities that needed to be updated in a specific cycle. So this value won't be updated every 60s but at every heartbeat (which is every 15 minutes).

> What is the reason behind storing the lastHeartbeat of the service? Would it be possible to only update this value when there is also a real change the service entities?

We planned for our check to use `ListProcessesWithFilter` to get services from WLM, and fits perfectly to how SD logic is implemented which needs that timestamp to know when an entry was last checked and generate a heartbeat event to the backend (which should eventually get rid of the service data after a while if it does not get it).

gabedos: Gotcha that is better only occurring every 15 minutes.

Curious if the existence of the process itself in wmeta still proves that the service is alive? Once the process is no longer running, there should be an unset event which would remove it from wmeta. Is there some difference between a process being alive and a service being alive?

Yumasi: > Curious if the existence of the process itself in wmeta still proves that the service is alive?

In the context of the agent, a service is a single process (so if you have say N workers for e.g nginx, the agent would report N nginx-worker services). So if a process is gone from WLM (and the service info that goes with it), then we would send a service-stop event to the backend.
Now, one could argue that a process could still be alive but no longer fit in the definition of a service (for example, no more listening ports). What would happen is that `collectServices` won't be able to update the service info for it, the last heartbeat field will be left untouched, so our check will never send heartbeats for it to the backend. If that goes on for too long, the backend will eventually get rid of the service on its side, even without receiving the service-stop event.

Yumasi: After discussing this with the team, we actually figured that we can remove the lastHeartbeat field. The collector can cache this info without putting it in the store, and the check also has some heartbeat tracking as well on its end. So, I've removed the field altogether and simplified the pid filtering logic accordingly.

comp/core/workloadmeta/def/types.go
```diff
@@ -1252,7 +1296,17 @@ func (p Process) String(_ bool) string {
 	if p.Language != nil {
 		_, _ = fmt.Fprintln(&sb, "Language:", p.Language.Name)
 	}
-
+	if p.Service != nil {
```
gabedos, line 1299: Optional nit: maybe make all/some of the service information print with the verbose command when users are specifically looking for extra details. Flares will still return verbose information from `workload-list`.

gabedos (approved):

comp/core/workloadmeta/def/types.go
Yumasi: Good point, I'll fix it before merging, thanks!

PR #37945, opened 2025-06-13
[pkg/dyninst] Support pointer chasing limit

piob-io:
### What does this PR do?

Respect MaxReferenceDepth as a pointer chasing limit. Pointers (and interfaces in the futures) count towards the limit. Strings, slices (and maps in the future) do not count towards the limit.

### Motivation

https://datadoghq.atlassian.net/browse/DEBUG-4066

### DIFF
--- a/pkg/dyninst/ebpf/stack_machine.h
+++ b/pkg/dyninst/ebpf/stack_machine.h
@@ -214,29 +214,37 @@ inline __attribute__((always_inline)) bool sm_return(stack_machine_t* sm) {
   return true;
 }
 
-static bool sm_enqueue_pointer(global_ctx_t* ctx,
-                               data_item_header_t data_item) {
-  data_item_header_t* elem =
-      pointers_queue_push(&ctx->stack_machine->pointers_queue);
-  if (elem == NULL) {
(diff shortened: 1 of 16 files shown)


pkg/dyninst/ebpf/stack_machine.h
```diff
@@ -285,20 +268,36 @@ sm_memoize_pointer(__maybe_unused global_ctx_t* ctx, type_t type,
 
 inline __attribute__((always_inline)) bool
 sm_record_pointer(global_ctx_t* ctx, type_t type, target_ptr_t addr,
+                  bool decrease_ttl,
```
ajwerner, line 271: decrease_ttl is a bit tricky here. It's both about pushing to the front and about decreasing the ttl. Maybe some commentary? The idea that it's capturing is something like dereferencing the pointer immediately vs deferring it.

ajwerner (approved):
LGTM

pkg/dyninst/ebpf/stack_machine.h
piob-io: Discussed offline

PR #37948, opened 2025-06-13
[AGENTCFG-276] Configure Agent Install Info via API

rahulkaukuntla:
### What does this PR do?
We are exposing an API in the agent codebase to explicitly set the install_method value.

Here is the provided context: on ECS, the agent is deployed as part of an ECS task, and for years customers have been copy/pasting the example task definition (in JSON). As a result, all agents deployed this way report DOCKER as the install_method, because that’s the default in the container image.

However, the agent can detect when it’s running on ECS, and  updates the install method at runtime using logic in [pkg/config/env](https://github.com/DataDog/datadog-agent/tree/main/pkg/config/env). Having an explicit API to override or set this value would help ensure more accurate reporting and give users more control.


pkg/util/installinfo/install_info.go
```diff
@@ -166,3 +247,92 @@ func logVersionHistoryToFile(versionHistoryFilePath, installInfoFilePath, agentV
 		return
 	}
 }
+
+// HandleSetInstallInfo is an HTTP handler for setting install info at runtime
+func HandleSetInstallInfo(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodPost && r.Method != http.MethodPut {
+		http.Error(w, "Method not allowed", http.StatusMethodNotAllowed)
+		return
+	}
```
GustavoCaso: We are already setting the `Methods` filter in the handler definition https://github.com/DataDog/datadog-agent/pull/37948/files#diff-2d0c2690e3abf0afdb94956a349c583b9e1cf4a348c2a9213982d0ba95d4c490R42

We can probably remove this extra check. 

Also, it seems that you also support PUT requests, so we might want to add that in the mux filter `Methods("POST", "PUT")`

pkg/util/installinfo/install_info.go
```diff
@@ -166,3 +247,92 @@ func logVersionHistoryToFile(versionHistoryFilePath, installInfoFilePath, agentV
...
+
+	respondWithSuccess(w, "Install info set successfully")
+}
+
+// HandleGetInstallInfo is an HTTP handler for getting current install info
+func HandleGetInstallInfo(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodGet {
+		http.Error(w, "Method not allowed", http.StatusMethodNotAllowed)
+		return
+	}
```
GustavoCaso: Same comment about not needing this extract check if we are using the mux filters

pkg/util/installinfo/install_info.go
```diff
@@ -166,3 +247,92 @@ func logVersionHistoryToFile(versionHistoryFilePath, installInfoFilePath, agentV
...
+}
+
+// HandleClearInstallInfo is an HTTP handler for clearing runtime install info override
+func HandleClearInstallInfo(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodDelete {
+		http.Error(w, "Method not allowed", http.StatusMethodNotAllowed)
+		return
+	}
```
GustavoCaso: Same comment about not needing this extract check if we are using the mux filters

pkg/util/installinfo/install_info.go
```diff
@@ -58,12 +80,69 @@ func GetFilePath(conf model.Reader) string {
...
+		return installInfo, nil
+	}
 	if installInfo, ok := getFromEnvVars(); ok {
 		return installInfo, nil
 	}
 	return getFromPath(GetFilePath(conf))
 }
 
+// SetRuntimeInstallInfo sets the install info at runtime, overriding file and env var values
+func SetRuntimeInstallInfo(info *InstallInfo) error {
```
GustavoCaso: I would make this function private at first. We want to reduce the surface area of the package and ways that people can tweak the installation info. It seems that we want to use the HTTP server for that, in that case I would make this function private and only expose the HTTP handlers

pkg/util/installinfo/install_info.go
```diff
@@ -50,6 +53,25 @@ type versionHistoryEntries struct {
...
+
+// SetInstallInfoRequest represents the JSON payload for setting install info
+type SetInstallInfoRequest struct {
+	Tool             string `json:"tool"`
+	ToolVersion      string `json:"tool_version"`
+	InstallerVersion string `json:"installer_version"`
+}
+
+// SetInstallInfoResponse represents the response after setting install info
+type SetInstallInfoResponse struct {
```
GustavoCaso, line 70: Same comment about privacy. The structs seems to be only relevant in the context of this package, so I would make them private at first

pkg/util/installinfo/install_info_test.go
```diff
@@ -322,3 +326,399 @@ func TestScrubFromPath(t *testing.T) {
 	assert.Equal(t, "2.5.0 password=********", info.ToolVersion)
 	assert.Equal(t, "3.7.1 password=********", info.InstallerVersion)
 }
+
+func TestSetRuntimeInstallInfo(t *testing.T) {
```
GustavoCaso, line 327: We could run the tests on the API component. We are testing the handlers here, but we are not testing that the API routes are available for the API server. Unless you are planning on adding more tests on a following PR I think it would be better to move the test to the API server component which is the one that expose the routes

comp/api/api/apiimpl/internal/agent/agent.go
```diff
@@ -37,7 +38,9 @@ func SetupHandlers(
 	r.HandleFunc("/status/health", getHealth).Methods("GET")
 	r.HandleFunc("/{component}/status", componentStatusHandler).Methods("POST")
 	r.HandleFunc("/{component}/configs", componentConfigHandler).Methods("GET")
-
+	r.HandleFunc("/api/v1/install-info/get", installinfo.HandleGetInstallInfo).Methods("GET")
```
GustavoCaso: We do not version routes for the agent handler. These routes are meant to be used privately by the Agent. Following the previous routes I would remove the `/api/v1/` prefix

GustavoCaso:
@rahulkaukuntla, the code looks good. I left a few comments, but my biggest concern is about the API users. Who is going to be using this API, us? Will it be used from the same Agent process?  I think having a GET and SET handler, if we decide to use the HTTP route, is more than enough. I'm not sure exposing a Clear handler is something we want, as it could potentially remove important information. 

Looking forward to your reply 😄

pkg/util/installinfo/install_info.go
rahulkaukuntla: Removed filter, and added "PUT" in [9b8b67d](https://github.com/DataDog/datadog-agent/pull/37948/commits/9b8b67d5a0738940f90bae94598c504734405e42).

pkg/util/installinfo/install_info.go
rahulkaukuntla: done in [9b8b67d](https://github.com/DataDog/datadog-agent/pull/37948/commits/9b8b67d5a0738940f90bae94598c504734405e42)

pkg/util/installinfo/install_info.go
rahulkaukuntla: removed in [9b8b67d](https://github.com/DataDog/datadog-agent/pull/37948/commits/9b8b67d5a0738940f90bae94598c504734405e42)

pkg/util/installinfo/install_info.go
rahulkaukuntla: Done in [9b8b67d](https://github.com/DataDog/datadog-agent/pull/37948/commits/9b8b67d5a0738940f90bae94598c504734405e42). I only keep the handlers public, and moved the handler tests to the api component.

pkg/util/installinfo/install_info.go
rahulkaukuntla: Talked offline about this--since the handler tests use these structs, I am keeping these public for now.

pkg/util/installinfo/install_info_test.go
rahulkaukuntla: Agreed--I moved the handler tests to the api component in [9b8b67d](https://github.com/DataDog/datadog-agent/pull/37948/commits/9b8b67d5a0738940f90bae94598c504734405e42).

comp/api/api/apiimpl/internal/agent/agent.go
rahulkaukuntla: Removed the prefixes.

rahulkaukuntla:
> @rahulkaukuntla, the code looks good. I left a few comments, but my biggest concern is about the API users. Who is going to be using this API, us? Will it be used from the same Agent process? I think having a GET and SET handler, if we decide to use the HTTP route, is more than enough. I'm not sure exposing a Clear handler is something we want, as it could potentially remove important information.
> 
> Looking forward to your reply 😄

Hi Gustavo, thank you for the thorough review! We agreed offline that a clear handler isn't really necessary, as the use case is customers overriding their agent setting their install info to certain values.

comp/api/api/apiimpl/internal/agent/agent_test.go
```diff
@@ -63,3 +67,435 @@ func TestSetupHandlers(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			var body []byte
+			var err error
+
+			if tt.payload != nil {
+				body, err = json.Marshal(tt.payload)
+				require.NoError(t, err, "Failed to marshal test payload")
+			}
+
+			req, err := http.NewRequest(tt.method, ts.URL+tt.route, bytes.NewReader(body))
```
GustavoCaso: Nit: should we use  `url.JoinPath(ts.URL, tt.route)`

comp/api/api/apiimpl/internal/agent/agent_test.go
```diff
@@ -63,3 +67,435 @@ func TestSetupHandlers(t *testing.T) {
...
+					assert.True(t, exists, "Response should contain 'message' field")
+					assert.NotEmpty(t, message, "Response message should not be empty")
+				}
+			}
+
+			// Additional validation for method not allowed responses
+			if tt.expectedCode == 405 {
+				// Gorilla Mux returns 405 for unsupported methods
+				assert.Equal(t, 405, resp.StatusCode, "Should return Method Not Allowed for unsupported methods")
+			}
```
GustavoCaso: Nit: Normally, I prefer to create assert function as part of the tests struct. It might be a more repetitive, but I find that it causes less back and forth when looking at the test.  Especially if in the `t.Run` function there is a lot conditional logic base on the test struct values 😄 

Example: 
```go
{
	name:         "install info get with GET method",
	route:        "/install-info/get",
	method:       "GET",
	payload:      nil,
	assertFunnc:  func(resp *http.Response) {
		assert.Equal(t, 200, resp.StatusCode, "excepted status code 200 for GET /install-info/get got: %d", resp.StatusCode)
		
		responseBody, err := io.ReadAll(resp.Body)
		require.NoError(t, err, "Failed to read response body")
		
		// Verify Content-Type header is set for JSON responses
		contentType := resp.Header.Get("Content-Type")
		assert.Equal(t, "application/json", contentType, "Response should have JSON content type")
		
		var installInfo installinfo.InstallInfo
		err = json.Unmarshal(responseBody, &installInfo)
		assert.NoError(t, err, "Response should be valid InstallInfo JSON")

		assert.Equal(t, "test-tool", installInfo.Tool, "Tool should match test data")
		assert.Equal(t, "1.0.0", installInfo.ToolVersion, "ToolVersion should match test data")
		assert.Equal(t, "test-installer-1.0", installInfo.InstallerVersion, "InstallerVersion should match test data")
	}
	description:  "GET request to install-info/get should succeed with test config",
},
```

It might be a bit more repetitive, but I find it more readable. 

Let me know what you think

comp/api/api/apiimpl/internal/agent/agent_test.go
```diff
@@ -63,3 +67,435 @@ func TestSetupHandlers(t *testing.T) {
...
+	setPayload := installinfo.SetInstallInfoRequest{
+		Tool:             "ECS",
+		ToolVersion:      "1.0",
+		InstallerVersion: "test-installer",
+	}
+
+	body, err := json.Marshal(setPayload)
+	require.NoError(t, err)
+
+	req, err := http.NewRequest("POST", ts.URL+"/install-info/set", bytes.NewReader(body))
```
GustavoCaso: Same comment regarding the usage of `url.JoinPath`

comp/api/api/apiimpl/internal/agent/agent_test.go
```diff
@@ -63,3 +67,435 @@ func TestSetupHandlers(t *testing.T) {
...
+	require.NoError(t, err)
+	req.Header.Set("Content-Type", "application/json")
+
+	resp, err := ts.Client().Do(req)
+	require.NoError(t, err)
+	defer resp.Body.Close()
+
+	assert.Equal(t, 200, resp.StatusCode, "Setting install info should succeed")
+
+	req, err = http.NewRequest("GET", ts.URL+"/install-info/get", nil)
```
GustavoCaso: Same comment regarding the usage of `url.JoinPath`

comp/api/api/apiimpl/internal/agent/agent_test.go
```diff
@@ -63,3 +67,435 @@ func TestSetupHandlers(t *testing.T) {
...
+					body, err = json.Marshal(tt.payload)
+					require.NoError(t, err)
+				}
+			}
+
+			req := httptest.NewRequest(tt.method, "/install-info/set", bytes.NewReader(body))
+			req.Header.Set("Content-Type", "application/json")
+			w := httptest.NewRecorder()
+
+			installinfo.HandleSetInstallInfo(w, req)
```
GustavoCaso: Should we use `resp, err := ts.Client().Do(req)` instead of calling the handler directly?

comp/api/api/apiimpl/internal/agent/agent_test.go
```diff
@@ -63,3 +67,435 @@ func TestSetupHandlers(t *testing.T) {
...
+				w := httptest.NewRecorder()
+				req := httptest.NewRequest("POST", "/install-info/set", bytes.NewReader(body))
+				req.Header.Set("Content-Type", "application/json")
+				installinfo.HandleSetInstallInfo(w, req)
+			}
+
+			w := httptest.NewRecorder()
+			req := httptest.NewRequest("GET", "/install-info/get", nil)
+
+			installinfo.HandleGetInstallInfo(w, req)
```
GustavoCaso: Same question as: `Should we use resp, err := ts.Client().Do(req) instead of calling the handler directly?`

comp/api/api/apiimpl/internal/agent/agent_test.go
rahulkaukuntla: This is a great suggestion! It's a lot easier to tell what is happening in this test. I've made the change in [e1efcfa](https://github.com/DataDog/datadog-agent/pull/37948/commits/e1efcfaa33afbbb8537b1cd6ec46dd13d861a982).

comp/api/api/apiimpl/internal/agent/agent_test.go
rahulkaukuntla: Done in [e1efcfa](https://github.com/DataDog/datadog-agent/pull/37948/commits/e1efcfaa33afbbb8537b1cd6ec46dd13d861a982).

comp/api/api/apiimpl/internal/agent/agent_test.go
rahulkaukuntla: Done in [e1efcfa](https://github.com/DataDog/datadog-agent/pull/37948/commits/e1efcfaa33afbbb8537b1cd6ec46dd13d861a982).

comp/api/api/apiimpl/internal/agent/agent_test.go
rahulkaukuntla: Done in [e1efcfa](https://github.com/DataDog/datadog-agent/pull/37948/commits/e1efcfaa33afbbb8537b1cd6ec46dd13d861a982).

comp/api/api/apiimpl/internal/agent/agent_test.go
rahulkaukuntla: Sounds good, I've switched the call in [79c5bc6](https://github.com/DataDog/datadog-agent/pull/37948/commits/79c5bc630d6c7d3c10331bb15917775ec3dce8a4).

comp/api/api/apiimpl/internal/agent/agent_test.go
rahulkaukuntla: Switched in [79c5bc6](https://github.com/DataDog/datadog-agent/pull/37948/commits/79c5bc630d6c7d3c10331bb15917775ec3dce8a4).

comp/api/api/apiimpl/internal/agent/agent.go
```diff
@@ -37,7 +38,8 @@ func SetupHandlers(
 	r.HandleFunc("/status/health", getHealth).Methods("GET")
 	r.HandleFunc("/{component}/status", componentStatusHandler).Methods("POST")
 	r.HandleFunc("/{component}/configs", componentConfigHandler).Methods("GET")
-
+	r.HandleFunc("/install-info/get", installinfo.HandleGetInstallInfo).Methods("GET")
```
dustmop: Is there a need for these two routes to be different paths? Alternatively, could we use `/install-info` for both and distinguish the handler based upon the method? Or is there already something that exists that is depending upon these exact paths.

pkg/util/installinfo/install_info.go
```diff
@@ -58,12 +80,60 @@ func GetFilePath(conf model.Reader) string {
...
+
+	log.Infof("Runtime install info set: tool=%s, tool_version=%s, installer_version=%s",
+		runtimeInstallInfo.Tool, runtimeInstallInfo.ToolVersion, runtimeInstallInfo.InstallerVersion)
+
+	return nil
+}
+
+// getRuntimeInstallInfo returns the current runtime override if set
+func getRuntimeInstallInfo() *InstallInfo {
+	return getRuntimeOverride()
```
dustmop: Let's combine these two functions, doesn't seem worth it to have one just call the other.

pkg/util/installinfo/install_info.go
```diff
@@ -58,12 +80,60 @@ func GetFilePath(conf model.Reader) string {
...
+	}
+
+	if info.Tool == "" || info.ToolVersion == "" || info.InstallerVersion == "" {
+		return fmt.Errorf("install info must have tool, tool_version, and installer_version set")
+	}
+
+	runtimeInfoMutex.Lock()
+	defer runtimeInfoMutex.Unlock()
+
+	runtimeInstallInfo = scrubFields(&InstallInfo{
```
dustmop, line 107: It's not a major issue, but worth noting: the other methods of getting the install-info (env vars or the file) scrub at `get` time, whereas here we're scrubbing at `set` time. Nothing exists at the moment that needs to retrieve unscrubbed info, but that may be the case in the future. I would recommend adding a comment to this function just to point out that the assigned data gets scrubbed.

dustmop (commented):
Overall looks great, but I have just a few small comments.

comp/api/api/apiimpl/internal/agent/agent.go
rahulkaukuntla: Per our conversation, I just have a `/install-info` endpoint. As a side effect, I had to modify some test cases--405 are now reported as 400 (doesn't seem like a big issue), and since get and post requests are sent to the same endpoint, previously "faulty" get requests return 200 now since the Methods handler redirects it correctly now, so I removed those test cases as they were misleading.

pkg/util/installinfo/install_info.go
rahulkaukuntla: Done.

pkg/util/installinfo/install_info.go
rahulkaukuntla: Comment added to the setRuntimeInstallInfo method.

comp/api/api/apiimpl/internal/agent/agent.go
dustmop: That makes sense. The different methods from the endpoint mean we shouldn't need to test it the same way.

dustmop (approved):

comp/api/api/apiimpl/internal/agent/agent_test.go
```diff
@@ -63,3 +70,447 @@ func TestSetupHandlers(t *testing.T) {
...
+				assert.Equal(t, "test-installer-1.0", installInfo.InstallerVersion)
+			},
+		},
+		{
+			name:         "install info get with POST method",
+			route:        "/install-info",
+			method:       "POST",
+			payload:      nil,
+			expectedCode: 400,
+			description:  "POST request to install-info should be rejected",
```
GustavoCaso: The descriptions for these tests and the one below (line 146) may miss some information on why they are rejected. Are they rejected because the body is empty? If so, it would be good to add that as part of the description. 😄

comp/api/api/apiimpl/internal/agent/agent_test.go
```diff
@@ -63,3 +70,447 @@ func TestSetupHandlers(t *testing.T) {
...
+			description:  "DELETE request to install-info should be rejected",
+		},
+		{
+			name:         "install info set with POST method and empty payload",
+			route:        "/install-info",
+			method:       "POST",
+			payload:      nil,
+			expectedCode: 400,
+			description:  "POST request to install-info with empty payload should return bad request",
+		},
```
GustavoCaso: These two tests are duplicated. We already have `install info get with DELETE method` line 149 and `install info get with POST method` line 133. We can remove them 😄

comp/api/api/apiimpl/internal/agent/agent_test.go
```diff
@@ -63,3 +70,447 @@ func TestSetupHandlers(t *testing.T) {
...
+				tt.description, tt.method, tt.route, resp.StatusCode, tt.expectedCode, string(responseBody))
+
+			if tt.assertFunc != nil {
+				tt.assertFunc(t, resp, responseBody)
+			}
+		})
+	}
+}
+
+func TestInstallInfoAPIRoutesWithRuntimeInfo(t *testing.T) {
```
GustavoCaso: We could probably merge these assertions in the above test function `TestInstallInfoAPIRoutes` we can add them as part of the tests: `POST request to install-info with valid payload should succeed` and `PUT request to install-info with valid payload should succeed` 

That way things are organized in a single test 😄

comp/api/api/apiimpl/internal/agent/agent_test.go
```diff
@@ -63,3 +70,447 @@ func TestSetupHandlers(t *testing.T) {
...
+	var installInfo installinfo.InstallInfo
+	err = json.Unmarshal(responseBody, &installInfo)
+	assert.NoError(t, err, "Response should be valid InstallInfo JSON")
+
+	assert.Equal(t, "ECS", installInfo.Tool, "Tool should match what was set")
+	assert.Equal(t, "1.0", installInfo.ToolVersion, "ToolVersion should match what was set")
+	assert.Equal(t, "test-installer", installInfo.InstallerVersion, "InstallerVersion should match what was set")
+}
+
+func TestHandleSetInstallInfo(t *testing.T) {
```
GustavoCaso: Same comment as above. We can probably merge these assertions in the `TestInstallInfoAPIRoutes` test function 😄

comp/api/api/apiimpl/internal/agent/agent_test.go
```diff
@@ -63,3 +70,447 @@ func TestSetupHandlers(t *testing.T) {
...
+				require.NoError(t, err)
+				assert.Equal(t, "ECS", info.Tool)
+				assert.Equal(t, "1.0", info.ToolVersion)
+				assert.Equal(t, "ecs-task-def-v1", info.InstallerVersion)
+			}
+		})
+	}
+}
+
+func TestHandleGetInstallInfo(t *testing.T) {
```
GustavoCaso: Same comment as above. We can probably merge these assertions in the TestInstallInfoAPIRoutes test function 😄

GustavoCaso (commented):
Great work @rahulkaukuntla 
I still think we should probably cleanup a bit the tests file, as there is a bit of duplication. Let me know if you have any question

comp/api/api/apiimpl/internal/agent/agent_test.go
rahulkaukuntla: Good idea, i updated the test descriptions to be more specific about why POST/PUT requests are rejected (e.g., "POST request to install-info should be rejected" became "POST request to install-info with empty payload should return bad request")

comp/api/api/apiimpl/internal/agent/agent_test.go
rahulkaukuntla: Sounds good, I've removed them (i removed the one on [line 133](https://github.com/DataDog/datadog-agent/pull/37948/commits/08196fdc6be5637e66da15e68abe9314b3a6588b#diff-bc3bb0f7b3c1274f60740f0d6659065744cff04d050c9eddcb7136bb5dbf6e1dL133) instead actually, since the description was slightly less descriptive 😅).

comp/api/api/apiimpl/internal/agent/agent_test.go
rahulkaukuntla: Yes, this would make things more simple...I've made the changes.

comp/api/api/apiimpl/internal/agent/agent_test.go
rahulkaukuntla: Done.

comp/api/api/apiimpl/internal/agent/agent_test.go
rahulkaukuntla: Done.

GustavoCaso (approved):
Great work 🎉

PR #37951, opened 2025-06-13
[OTAGENT-428] Converter: granular management of auto-enrich features 

tim-liakh-datadog:
### What does this PR do?
This PR gives the user the ability to select which auto-enrichments they want from the converter.
### Motivation
Previously, the user would be forced to enable or disable all of them. 

To validate my changes I ran the unit tests and manually checked that it performed as expected with environment variables and in datadog.yaml. In addition, I created new unit tests to make sure it continues to work correctly in the future. 
Creates additional documentation which might confuse user.


joepeeples (approved):

pkg/config/setup/config.go
```diff
@@ -927,6 +927,22 @@ func InitConfig(config pkgconfigmodel.Setup) {
 	config.BindEnvAndSetDefault("otelcollector.submit_dummy_metadata", false) // dev flag - to be removed
 	config.BindEnvAndSetDefault("otelcollector.converter.enabled", true)
 	config.BindEnvAndSetDefault("otelcollector.flare.timeout", 60)
+	config.BindEnvAndSetDefault("otelcollector.converter.features", []string{"infraattributes", "prometheus", "core", "pprof", "zpages", "health_check", "ddflare"}, "DD_OTELCOLLECTOR_FEATURES")
```
liustanley:
```suggestion
	config.BindEnvAndSetDefault("otelcollector.converter.features", []string{"infraattributes", "prometheus", "core", "pprof", "zpages", "health_check", "ddflare"})
```

The environment variable name can be removed here, the config package will automatically set the environment variable name according to its convention (`DD_OTELCOLLECTOR_CONVERTER_FEATURES`)

comp/otelcol/converter/impl/autoconfigure.go
```diff
@@ -23,9 +24,28 @@ type component struct {
...
 func (c *ddConverter) enhanceConfig(conf *confmap.Conf) {
-	// extensions
+	var enabledFeatures []string
+
+	// If not specified, assume all features are enabled
+	if c.coreConfig != nil {
+		enabledFeatures = c.coreConfig.GetStringSlice("otelcollector.converter.features")
+	} else {
+		enabledFeatures = []string{"infraattributes", "prometheus", "core", "pprof", "zpages", "health_check", "ddflare"}
+	}
```
liustanley: We don't need to check if `c.coreConfig` is nil here, it will always be present when this function is called. There's also no need to set `enabledFeatures` to all features here since you already set this by default in `pkg/config/setup/config.go`.

pkg/config/setup/config.go
liustanley: Would also recommend declaring the full list of default features as a constant in `autoconfigure.go` and importing it here. That way all of the converter features are in one place and can be modified more easily in the future.

pkg/config/setup/config.go
```diff
@@ -927,6 +927,22 @@ func InitConfig(config pkgconfigmodel.Setup) {
...
+	config.ParseEnvAsStringSlice("otelcollector.converter.features", func(s string) []string {
+		// Either commas or spaces can be used as separators.
+		// Comma takes precedence as it was the only supported separator in the past for other implementations.
+		// Mixing separators is not supported.
```
dustmop: Is it a hard requirement to not allow mixing separators, or a consequence of the implementation? If you're okay with allowing either, you could simplify this code by using [FieldsFunc](https://pkg.go.dev/strings#FieldsFunc). We do something similar here: https://github.com/DataDog/datadog-agent/blob/56572b4981ca461c28ffb8987cd9a7414007c523/pkg/fleet/installer/setup/common/setup.go#L72

comp/otelcol/converter/impl/converter_test.go
```diff
@@ -305,6 +305,48 @@ func TestConvert(t *testing.T) {
...
+		{
+			name:           "features/infraattributes-only",
+			provided:       "features/infraattributes-only/config.yaml",
+			expectedResult: "features/infraattributes-only/config-result.yaml",
+			agentConfig:    "features/infraattributes-only/acfg.yaml",
+		},
+		{
+			name:           "features/no-features",
+			provided:       "features/no-features/config.yaml",
+			expectedResult: "features/no-features/config-result.yaml",
```
liustanley: nit: you could use `config.yaml` for both here and delete `features/no-features/config-result.yaml`

comp/otelcol/converter/impl/testdata/features/all-features/config.yaml
```diff
@@ -0,0 +1,24 @@
+exporters:
+  datadog:
+    api: 
+      key: ggggg77777
+      site: datadoghq.eu
```
liustanley: Should remove these two lines in this test

comp/otelcol/converter/impl/converter_test.go
```diff
@@ -305,6 +305,48 @@ func TestConvert(t *testing.T) {
 			expectedResult: "dd-core-cfg/env/empty-profiler-options/config-result.yaml",
 			agentConfig:    "dd-core-cfg/env/empty-profiler-options/acfg.yaml",
 		},
+		{
```
liustanley, line 308: Could you also add a test case where `features` is not defined in `acfg.yaml` to verify that all features still get used?

liustanley (changes requested):
Nice job on the implementation and tests! Mostly leaving comments on smaller things but the rest of the PR looks good. Would also recommend adding documentation in the converter README for this feature.

comp/otelcol/converter/impl/converter_test.go
tim-liakh-datadog: Good point, I just added it.

comp/otelcol/converter/impl/testdata/features/all-features/config.yaml
tim-liakh-datadog: I believe these two arguments are required for the "core" feature to populate under `
exporters:
  datadog:
    api:
`. I did notice I accidentally had the `config.yaml` pre-populated with the key and site info which I just fixed.

comp/otelcol/converter/impl/converter_test.go
tim-liakh-datadog: That makes sense, removed.

comp/otelcol/converter/impl/autoconfigure.go
tim-liakh-datadog: I was running into the unit tests previously written having c.coreConfig be nil. Maybe this is something we can look at together.

pkg/config/setup/config.go
tim-liakh-datadog: Oh cool! Didn't know that was a feature, updated it 👍

pkg/config/setup/config.go
liustanley: I don't think it's a hard requirement, but we went with this approach to be consistent with the implementation of [apm_config.features](https://github.com/DataDog/datadog-agent/blob/8b78abb0c5a2b38d6365cc1b19f06e3e35e54692/pkg/config/setup/apm.go#L192-L206) which mirrors what we wanted to do with this config option. I think it might make sense to not allow mixing separators, but would also be fine with allowing mixing separators if this is commonly done.

comp/otelcol/converter/impl/autoconfigure.go
liustanley: I think this is something we should fix in the converter tests, if there is no agent config file provided the converter should be supplied `config.NewMock`.

pkg/config/setup/config.go
tim-liakh-datadog: I wasn't able to do it inside of autoconfigure, but made a new file to hold the constant. Do you think this would be fine as well?

pkg/config/setup/config.go
tim-liakh-datadog: That's right, I initially implemented it that way to match apm_config, but I don't see a reason as to why not to support it. I made a change to allow for both.

comp/otelcol/converter/impl/autoconfigure.go
tim-liakh-datadog: Implemented this, had to modify some tests and modify the creation of the ocb environment mock file

pkg/config/setup/config.go
tim-liakh-datadog: Spoke with Stanley, reverted back to the use without the constant.

comp/otelcol/converter/impl/autoconfigure.go
```diff
@@ -23,9 +24,16 @@ type component struct {
 	Config       any
 }
 
+// Applies selected feature changes
 func (c *ddConverter) enhanceConfig(conf *confmap.Conf) {
-	// extensions
+	var enabledFeatures []string = c.coreConfig.GetStringSlice("otelcollector.converter.features")
```
liustanley:
```suggestion
	var enabledFeatures = c.coreConfig.GetStringSlice("otelcollector.converter.features")
```

nit: the type can be omitted here since you're instantiating in the same line

comp/otelcol/converter/impl/autoconfigure.go
```diff
@@ -23,9 +24,23 @@ type component struct {
 	Config       any
 }
 
+// Applies selected feature changes
 func (c *ddConverter) enhanceConfig(conf *confmap.Conf) {
-	// extensions
+	var enabledFeatures []string
+
+	// If not specified, assume all features are enabled (ocb tests will not have enabled)
```
liustanley:
```suggestion
	// If not specified, assume all features are enabled (ocb tests will not have coreConfig)
```

comp/otelcol/converter/impl/autoconfigure.go
```diff
@@ -23,9 +24,23 @@ type component struct {
...
+	if c.coreConfig != nil {
+		enabledFeatures = c.coreConfig.GetStringSlice("otelcollector.converter.features")
+	} else {
+		enabledFeatures = []string{"infraattributes", "prometheus", "pprof", "zpages", "health_check", "ddflare"}
+	}
+
+	// extensions (pprof, zpages, health_check, ddflare)
 	for _, extension := range extensions {
+		if !slices.Contains(enabledFeatures, extension.Name) {
```
liustanley: nit: these two if statements could be combined for simplicity:

```
if !slices.Contains(enabledFeatures, extension.Name) || extensionIsInServicePipeline(conf, extension) {
  continue
}
```

liustanley (approved):
LGTM, just two small nits

tim-liakh-datadog:
@dustmop When you have a moment could you please review again to make sure it is still good to go. I made a good amount of changes.

dustmop (approved):

comp/otelcol/converter/impl/autoconfigure.go
tim-liakh-datadog: Done 👍

PR #37952, opened 2025-06-13
Add sample service, adds decoding path to integration test

grantseltzer:
### What does this PR do?

- Adds the sample service
- Adds call to decoding to integration test

### Motivation

More testing

Ran the tests

### Additional Notes


pgimalac (approved):

pkg/dyninst/testprogs/progs/go.mod
piob-io, line 1: Seems like this file was not marked as moved?

piob-io (commented):
Could you comment on the move of directories into new testing/ directory? I am not sure about the purpose of the new directory. E.g. do tests themselves like integration_test and throttler_test belong there and why?

The original change - adding sample service, was already quite big, making it harder to review and merge. I'd suggest separating that from any other bigger changes, like code moves. Especially code moves are easier to review if they are no-ops, as we don't have to worry about missing logical changes.

grantseltzer:
> Could you comment on the move of directories into new testing/ directory? I am not sure about the purpose of the new directory. E.g. do tests themselves like integration_test and throttler_test belong there and why?
> 
> The original change - adding sample service, was already quite big, making it harder to review and merge. I'd suggest separating that from any other bigger changes, like code moves. Especially code moves are easier to review if they are no-ops, as we don't have to worry about missing logical changes.

@piob-io You make a good point, this is too difficult to review. I've reverted the changes to directory structure and made this PR just focusing on adding the sample service, and adding decoding to integration test.

The decoding and comparison of output works, but as mentioned on slack (pasted below), it can only work for single probe tests at a time as I need to add event header info (and fix propagation of it) to decoding.

>It doesn't yet have comparison of output in its entirety because I need to improve decoding to include header details (so we can correlate types/events with the probe ID).

piob-io:
Not sure if you want to merge this as-is, but let's make the test pass. If the failures are due to the shortcomings you mentioned, let's disable the test cases that don't work yet.

grantseltzer:
> Not sure if you want to merge this as-is, but let's make the test pass. If the failures are due to the shortcomings you mentioned, let's disable the test cases that don't work yet.

@piob-io If I can get event/program IDs working today and merge that first, we'll do that. But otherwise I'll just disable the test. Will keep ya posted.

grantseltzer:
Once https://github.com/DataDog/datadog-agent/pull/38027 merges, i'll rebase and fix the currently failing tests and uncomment the test for all probes at once (since we will be able to correlate events with their probe definitions).

pkg/dyninst/testprogs/test_progs.go
```diff
@@ -63,60 +63,64 @@ func MustGetBinary(t *testing.T, name string, cfg Config) string {
...
 var (
-	globalState     state
+	globalState     State
 	globalStateErr  error
 	globalStateOnce sync.Once
 )
 
-func getState() (*state, error) {
+// GetState returns the global state of the testprogs package.
+func GetState() (*State, error) {
```
piob-io: Why export this and other internals of this package?

grantseltzer: It may be a remnant from when I rearranged all the directories based on feedback and then moved them back based on further feedback. I don't see any problem with having it exported, it's a test utility.

piob-io: We should keep the same high code quality regardless. Tests should be easy to read, and if we start messing with internal state of utilities, it won't. Please revert.

grantseltzer: Sure that's fine.

pkg/dyninst/testprogs/test_progs.go
```diff
@@ -108,6 +109,8 @@ type state struct {
 	haveSources bool
 	// The directory where the probe configs are stored.
 	probesCfgsDir string
+	// ExpectedOutputDir is the directory where the expected output files are stored.
+	expectedOutputDir string
```
piob-io: This looks unused now too.

piob-io (approved):
Thank you!

PR #37960, opened 2025-06-16
Remove klog v1

pgimalac:
### What does this PR do?
k8s.io/klog (v1) is not used anymore (only import is from us to set a custom logger), so we should be able to remove it.

### Motivation
Remove unneeded and outdated dependencies.

CI


dustmop (approved):
Nice!

gabedos (approved):

wdhif (approved):

PR #37961, opened 2025-06-16
[CWS-3909] Move the cgroups tag resolution in the tags resolver

mftoure:
### What does this PR do?
This PR aims to  move the cgroups tags resolution in the tags resolver


pkg/security/resolvers/tags/resolver_linux.go
```diff
@@ -117,30 +131,108 @@ func (t *LinuxResolver) checkTags(pendingWorkload *Workload) {
...
+	servicePath := filepath.Join(systemdSystemDir, systemdService)
+
+	// Try to resolve version using version resolver
+	if t.versionResolver != nil {
+		serviceVersion = t.versionResolver(servicePath)
+	}
+
+	tags := []string{
+		"service:" + systemdService,
+		"version:" + serviceVersion,
```
YoannGh: I think it would be better not to add a `version` tag if we failed to get the service version

pkg/security/resolvers/resolvers_ebpf.go
```diff
@@ -103,7 +103,19 @@ func NewEBPFResolvers(config *config.Config, manager *manager.Manager, statsdCli
...
 
-	tagsResolver := tags.NewResolver(opts.Tagger, cgroupsResolver)
+	// Create version resolver function that uses SBOM resolver if available
+	var versionResolver func(servicePath string) string
+	if config.RuntimeSecurity.SBOMResolverEnabled && sbomResolver != nil {
+		versionResolver = func(servicePath string) string {
+
+			if pkg := sbomResolver.ResolvePackage("", &model.FileEvent{PathnameStr: servicePath}); pkg != nil {
+				return pkg.Version
+			}
```
spikat, line 113: Q: I don't know much about sbom resolver but, did this func could blocks anyhow?

pkg/security/resolvers/tags/resolver_linux.go
```diff
@@ -89,7 +93,15 @@ func (t *LinuxResolver) Start(ctx context.Context) error {
...
 func needsTagsResolution(workload *Workload) bool {
-	return len(workload.ContainerID) != 0 && !workload.Selector.IsReady()
+	// Container workloads need tags resolution if they don't have a ready selector
+	if len(workload.ContainerID) != 0 {
+		return !workload.Selector.IsReady()
+	}
+	// Cgroup workloads (non-container) always need tags resolution if they have a cgroup ID
+	if len(workload.CGroupID) != 0 {
+		return true
+	}
```
spikat: I'm not sure of the implication. Does this means that we can't and won't have any cache mechanism for cgroups tags resolution? If no, do you think we should be concerned about the impact of making non-stop calls to the sbom resolver?

pkg/security/resolvers/tags/resolver_linux.go
```diff
@@ -117,30 +131,107 @@ func (t *LinuxResolver) checkTags(pendingWorkload *Workload) {
...
+		workload.Selector.Tag = utils.GetTagValue("image_tag", newTags)
+		if len(workload.Selector.Image) != 0 && len(workload.Selector.Tag) == 0 {
+			workload.Selector.Tag = "latest"
+		}
+	} else {
+		// For cgroup workloads, set service information as the selector
+		serviceName := utils.GetTagValue("service", newTags)
+		if len(serviceName) != 0 {
+			workload.Selector.Image = serviceName
+			workload.Selector.Tag = utils.GetTagValue("version", newTags)
```
spikat, line 160: Should we expect a version tag being always present or should we set a default value here? Or Does this means if not set we'll try later on?

pkg/security/resolvers/tags/resolver.go
```diff
@@ -46,14 +46,38 @@ type DefaultResolver struct {
...
+		return GetTagsOfContainer(t.tagger, v)
+	case containerutils.CGroupID:
+		if len(v) == 0 {
+			return nil, nil
+		}
+		// CGroup resolution is only supported on Linux
+		return nil, nil
+	default:
+		return nil, nil
+	}
```
spikat: Should return an error for line 68, 74 and 79 ?

pkg/security/resolvers/resolvers_ebpf.go
mftoure: I think it may be blocking if the hostSBOM is not computed yet.

pkg/security/resolvers/tags/resolver_linux.go
mftoure: I should use the same logic for cgroups

pkg/security/resolvers/tags/resolver_linux.go
mftoure: If the version tag is empty at this point, there is not much we can do. Which I don't think is a problem, as we can just use the image as the workload selector.

pkg/security/resolvers/resolvers_ebpf.go
spikat: discussed offline, it should not block

spikat (approved):

PR #37963, opened 2025-06-16
[AGENTRUN-46] Remove authtoken and tls config globals from pkg/api/util

misteriaud:
### What does this PR do?

Refactors `pkg/api/util` to remove global variables (`authtoken` and TLS configuration).  
These values are now passed explicitly to the relevant functions.

Key changes
- Split `pkg/api/util` into general logic and DCA-specific logic.
- Move auth-artifact initialization to `comp/core/ipc/imp`.
- Replace `pkg/api/util` globals with `ipc.Component` in:
  - JMXFetch
  - ddflareextension
  - `comp/trace`
- Update the Cluster Agent API to use `pkg/api/uti_dca.go`.
- Update **`GetHostname`** in `pkg/security/utils/hostnameutils/hostname.go` and pass an `ipc.Component` through all transitive calls.

### Motivation
Removing usage of `pkg/api/util` globals prevent any use of half-initialized variables.
- **Maintainability:** Using globals makes the codebase harder to reason about, test, and maintain. Passing configuration explicitly increases transparency and modularity.
- **Testability:** Removing globals allows for easier unit testing and isolation of logic, reducing side effects and facilitating dependency injection.
- **Clarity:** This change makes dependencies explicit and avoids potential issues with state leakage or unintended side effects.

Most of the changes are covered by unit-tests and e2e tests.


pkg/jmxfetch/state.go
```diff
@@ -95,6 +96,6 @@ func StopJmxfetch() {
 }
 
 // InitRunner inits the runner and injects the dogstatsd server component
-func InitRunner(server dogstatsdServer.Component, logger jmxlogger.Component) {
-	state.runner.initRunner(server, logger)
+func InitRunner(server dogstatsdServer.Component, logger jmxlogger.Component, ipc ipc.Component) {
```
vickenty, line 99: Could you also please update the comment, since there are more components now than just dogstatsd?

vickenty (approved):

ichinaski (approved):

paulcacheux (approved):

adel121 (approved):
LGTM for @DataDog/container-platform owned files.

comp/otelcol/ddflareextension/impl/extension.go
```diff
@@ -184,7 +186,7 @@ func NewExtension(ctx context.Context, cfg *Config, telemetry component.Telemetr
 
 	// auth = providedConfigSupported; if value true, component was likely built by Agent and has
```
mackjmr: you can remove this comment now

mackjmr (approved):
LGTM

iliakur (approved):

wiyu (approved):
👍  from Container-Experiences

cmd/agent/subcommands/processchecks/command_test.go
```diff
@@ -11,34 +11,44 @@ import (
...
+	// with the `ipcfx.ModuleReadWrite()` module.
+	fxutil.Test[ipc.Component](t,
+		ipcfx.ModuleReadWrite(),
+		core.MockBundle(),
+		fx.Replace(configComponent.MockParams{
+			Params: configComponent.Params{
+				ConfFilePath: configPath,
+			},
+		}),
+	)
```
hush-hush: We're trying to move away from the MockParams as they have a lot of issues.

Could we update the code with:
```suggestion
    configComponent.NewMockFromYAMLFile(t, configPath)

	// Check command should work when an Agent is running, so we need to
	// ensure we have exisiting IPC auth artifacts.
	// This is done by building the IPC component
	// with the `ipcfx.ModuleReadWrite()` module.
	fxutil.Test[ipc.Component](t,
		ipcfx.ModuleReadWrite(),
		core.MockBundle(),
	)
```

The config mock is design to be created multiple times per check without issue (we don't reset the mock if it's already mocked, this allow multiple helpers to be combined).

hush-hush (changes requested):

hmahmood (approved):

pkg/jmxfetch/state.go
misteriaud: Done in b4bb924fb2830b5d28cea0bdadec6fc2d5226c8b.

comp/otelcol/ddflareextension/impl/extension.go
misteriaud: Done in 3698838973d775e08480460ad9a0f9800b1d5663.

cmd/agent/subcommands/processchecks/command_test.go
misteriaud: Good to know ! 
Done in 77db77a314ed2961a19df9edef74151c2c4246d4.
(I also made the change in `cmd/agent/subcommands/processchecks/command_test.go`)

jose-manuel-almaza (approved):

val06 (approved):
LGTM for ebpf-platform

hush-hush (approved):

tools/NamedPipeCmd/main.go
```diff
@@ -105,30 +98,17 @@ func main() {
...
-	fprintf("Setting up options. ")
-
-	options := &util.ReqOptions{Conn: 1}
-	if options.Authtoken == "" {
-		options.Authtoken = util.GetAuthToken()
-	}
-
-	if options.Ctx == nil {
-		options.Ctx = context.Background()
-	}
```
julien-lebot, line 117: As discussed offline, this code was never using the AuthToken in the first place since that required a call to SetAuthToken. Therefore it was using an empty token, so it's fine to remove it.

julien-lebot (approved):

tim-liakh-datadog (approved):

PR #37975, opened 2025-06-16
[USMON-1452] usm: merge statis tags

guyarb:
### What does this PR do?

Merging static tags when combining stats

### Motivation

Missed in most of the protocols (besides postgres)

### DIFF
--- a/pkg/network/protocols/redis/stats_test.go
+++ b/pkg/network/protocols/redis/stats_test.go
@@ -45,7 +45,7 @@ func TestCombineWith(t *testing.T) {
 
 	stats2.AddRequest(false, 10, 1, 10.0)
 	stats3.AddRequest(false, 15, 2, 15.0)
-	stats4.AddRequest(false, 20, 5, 20.0)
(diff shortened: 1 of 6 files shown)


pkg/network/protocols/redis/stats_test.go
```diff
@@ -59,6 +59,7 @@ func TestCombineWith(t *testing.T) {
 		assert.Equal(t, 45, s.Count)
 		assert.Equal(t, float64(45), s.Latencies.GetCount())
 		assert.Equal(t, 10.0, s.FirstLatencySample)
+		assert.Equal(t, uint64(1|2|4), s.StaticTags)
```
amitslavin, line 62: I guess it should be 1|2|5 as you inserted in line 48?

amitslavin (approved):

pkg/network/protocols/redis/stats_test.go
guyarb: the issue is line 48, it should have been 4
but since 5 == 4 | 1, there's no failure
anyway, pushed a fix

PR #37979, opened 2025-06-16
[fleet] telem - fix unsupported type

raphaelgavache:
Add a default value instead of unsupported type, this works with []string, []int, types that implement string method


BaptisteFoy (approved):
nice

PR #5500, opened 2020-05-12
Add ecs and kube cluster name tags

ahmed-mez:
### What does this PR do?

- Add `ecs_cluster_name` and `kube_cluster_name` in addition to `cluster_name`
- Add a new config `disable_cluster_name_tag_key` to disable sending the `cluster_name` tag

### Motivation

- `cluster_name` can refer to other clusters in some integrations, not the orchestrator cluster
- `ecs_cluster_name` and `kube_cluster_name` are needed for the orchestrator explorer

PR #37992, opened 2025-06-16
chore: DCA Static tags in tagger global entity clean up [CONTP-838]

gabedos:
### What does this PR do?

Moves tagging logic from cluster agent dispatcher to the cluster agent's tagger initialization.

### Motivation

Separation of concerns. Keep tagging logic within the tagger and apply the global configuration via `tagger.GetGlobalTags()`

CI and unit tests.

Additionally manually deployed the cluster agent with `dd_cluster_checks_cluster_tag_name` set to a custom value and verified that it was in the global entity and checkconfig for KSM. Also deployed DCA with `dd_disable_cluster_name_tag_key` to verify that tags aren't on the global entity or ksm check when this is set to true.

N/A

### Additional Notes

Need to move the workloadfilter create methods into a separate package to avoid pulling in workloadmeta dependencies that caused circular dependencies with the workloadmeta collectors (particularly SBOM)


test/new-e2e/tests/containers/k8s_test.go
```diff
@@ -561,6 +561,7 @@ func (suite *k8sSuite) TestNginx() {
 		},
 		Expect: testMetricExpectArgs{
 			Tags: &[]string{
+				`^cluster_name:`,
```
davidor: Question: Were these being emitted before and just not tested, or were they not emitted at all?

sblumenthal: This block doesn't have `AcceptUnexpectedTags: true,` in it from what I can see, so this would be a new tag. I wonder if it would make sense to simply remove the `disable_cluster_name_tag_key` check that was added in this PR, since it would only be a net new tag and that config is in place to not break old behavior with the historical tag that was in place

pkg/util/tags/static_tags.go
```diff
@@ -101,6 +104,24 @@ func GetGlobalEnvTags(config config.Reader) map[string][]string {
...
+
+	// CLUSTER_NAME and KUBE_CLUSTER_NAME
+	hname, _ := hostname.Get(context.TODO())
+	clusterTagValue := clustername.GetClusterName(context.TODO(), hname)
+	clusterTagName := config.GetString("cluster_checks.cluster_tag_name")
+	if clusterTagValue != "" {
+		if clusterTagName != "" && !config.GetBool("disable_cluster_name_tag_key") {
+			tags = append(tags, fmt.Sprintf("%s:%s", clusterTagName, clusterTagValue))
+			log.Info("Adding both tags cluster_name and kube_cluster_name. You can use 'disable_cluster_name_tag_key' in the Agent config to keep the kube_cluster_name tag only")
+		}
```
sblumenthal: See @davidor's comment, since this appears to only be a net new tag, I don't know that this check is needed here. It is not going to break old behavior

test/new-e2e/tests/containers/k8s_test.go
gabedos: Hey all,

The `disable_cluster_name_tag_key` option was added back in 2020 in this PR https://github.com/DataDog/datadog-agent/pull/5500 . From what I can gather, some customers may explicitly not want this generic `cluster_name` tag from being associated with their metrics. There can be a variety of cluster types like Kafka, Kubernetes, etc. In an ideal world, I believe we would completely deprecate the `cluster_name` tag as per Vincent/Cedric.

I don't believe I changed any of the logic for generating the tags. The intention was just to move it from the dca dispatcher logic and into the tagger instead.

I'm unsure how the tags are configured for the KSM check specifically, but it appears that originally it wasn't pulling the `cluster_name` tag defined in the dispatcher, but now it does via tagger. Many other checks in our e2e suite expect the `cluster_name` tag already, but for some reason only KSM didn't. This partially goes to prove why we need tags to be centralized in the tagger's global entity to make sure every check/telemetry emitted consistently has the full set of DCA global tags.

davidor: I get that this might not have been done on purpose and could just be because of how the code is structured. But based on what I know, I think we should avoid adding the tag in KSM. It wasn’t there before, and it’s a tag we’ve been trying to phase out in favor of `kube_cluster_name`, so I don’t see a reason to add it.

gabedos: Understood, I've made an update to maintain the partial support for the deprecated-ish`cluster_name` tag setup in the dispatcher. The tagger init will handle `kube_cluster_name`, `orch_cluster_id`, etc.

davidor (approved):

PR #37985, opened 2025-06-16
dyninst/irgen: make output deterministic

ajwerner:
This was flaking because the inlined subprograms could alternate.

See https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/984091989.

Part of https://datadoghq.atlassian.net/browse/DEBUG-3886.

PR #37994, opened 2025-06-16
Go DI decoding of strings, slices, and arrays

grantseltzer:
### What does this PR do?

Adds decoding functionality for strings, slices, and arrays for Go DI

### Motivation

We need to decode slices, arrays, and strings

Ran the integration test changes introduced in #37952

### DIFF
--- a/pkg/dyninst/decode/decoder.go
+++ b/pkg/dyninst/decode/decoder.go
@@ -317,7 +317,6 @@ func (d *Decoder) encodeValue(
 			return errors.New("passed data not long enough for slice header")
 		}
 		address := binary.NativeEndian.Uint64(data)
-		sliceLength := binary.NativeEndian.Uint64(data[8:16])
 		elementType := v.Data.ID
 		elementSize := int(v.Data.Element.GetByteSize())
 		sliceDataItem, ok := dataItems[typeAndAddr{
@@ -331,6 +330,7 @@ func (d *Decoder) encodeValue(
 		if err != nil {
 			return err
 		}
+		sliceLength := int(sliceDataItem.Header().Length) / elementSize
 		sliceData := sliceDataItem.Data()
 		for i := range int(sliceLength) {
 			elementData := sliceData[i*elementSize : (i+1)*elementSize]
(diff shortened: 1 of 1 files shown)


pkg/dyninst/decode/decoder.go
```diff
@@ -263,21 +288,67 @@ func (d *Decoder) encodeValue(
...
+		}]
+		if !ok {
+			return errors.New("slice data item not found in data items")
+		}
+		err := enc.WriteToken(jsontext.BeginArray)
+		if err != nil {
+			return err
+		}
+		sliceData := sliceDataItem.Data()
+		for i := range int(sliceLength) {
```
piob-io, line 335: Slice may be truncated, in which case the length in the header will be the untruncated one, and for element extraction we should use the truncated data item length.

pkg/dyninst/decode/decoder.go
```diff
@@ -291,7 +362,7 @@ func (d *Decoder) encodeValue(
 			}
 		}
 		stringValue, ok := dataItems[typeAndAddr{
-			irType: uint32(irType.GetID()),
+			irType: uint32(v.Data.GetID()),
 			addr:   address,
 		}]
 		if !ok {
```
piob-io, line 363: Ditto below about length mismatch due to truncation.

piob-io (approved):
Thanks!

pkg/dyninst/decode/decoder.go
grantseltzer: So `truncatedNumberOfSliceElements := dataItem.Header().Length / elementSize` ?

piob-io: Yes

ajwerner (commented):
Your test flake was https://github.com/DataDog/datadog-agent/pull/37985 which I am actively trying to merge...

pkg/dyninst/decode/decoder.go
grantseltzer: I've addressed this if you can take another look. Do arrays ever get truncated?

piob-io: Lg, thanks. No, arrays are not truncated. Slices and strings are always pointers to underlying data, offering an easy place to truncate - by reading less data behind the pointer. Arrays are either embedded in structs or declared directly , and in the latter case they can be stored in registers or stack, or go can decide to put them on heap. So there is not common approach to truncate them. But also commonly they aren't that large. So for now it's not a worry.

PR #37996, opened 2025-06-17
fix(installer): Install agent before adding the user to additional groups

BaptisteFoy:
### What does this PR do?
This PR installs the agent before adding the user to additional groups in the installer install scripts.
It also moves the copyInstallerToBin method to be unix-only.

### Motivation
Some installs for EMR with logs failed because we were trying to add the dd-agent user to additional groups before that user was even created.

E2E, manual QA

### DIFF
--- a/pkg/fleet/installer/setup/common/setup.go
+++ b/pkg/fleet/installer/setup/common/setup.go
@@ -130,7 +130,7 @@ func (s *Setup) Run() (err error) {
 		}
 	}
 	if err = s.postInstallPackages(); err != nil {
-		return fmt.Errorf("failed during pre-package installation: %w", err)
+		return fmt.Errorf("failed during post-package installation: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


pkg/fleet/installer/setup/common/setup.go
```diff
@@ -122,16 +122,16 @@ func (s *Setup) Run() (err error) {
...
-	}
 	for _, p := range packages {
 		url := oci.PackageURL(s.Env, p.name, p.version)
 		err = s.installPackage(p.name, url)
 		if err != nil {
 			return fmt.Errorf("failed to install package %s: %w", url, err)
 		}
 	}
+	if err = s.postInstallPackages(); err != nil {
+		return fmt.Errorf("failed during pre-package installation: %w", err)
```
san-jos: **🥜 nitpick:** ‏error message post-package

san-jos (approved):

PR #38001, opened 2025-06-17
Extend setsockopt hooks with Detailed Filter and Socket Data

theop-dd:
### What does this PR do?
Adds additional information and data related to the setsockopt hooks.

Socket Information:
- Socket type
- Socket family
- Socket protocol

Filter Information:
- Disassembled filter
- Filter hash

### Motivation
To more accurately detect BPF doors, additional data was required from the setsockopt syscall. Specifically, we need to retrieve details about the socket to which the filter is attached, as well as key information from the filter itself. This data enables us to define more precise detection rules based on specific filter characteristics in the future. 

The setsockopt test was updated to verify the tracking of the newly added information.


theop-dd:
need to create a new branch for conflict handling

PR #38006, opened 2025-06-17
Add log discovery based on service information in WLM

vitkyrka:
### What does this PR do?

Add an autodiscovery provider which listens for new processes/services from WorkloadMeta and schedules log configs when they have logs to tail.

Note that the provider needs to be enabled via a config change, it is not enabled by default.

### Motivation

https://datadoghq.atlassian.net/wiki/spaces/DSCVR/pages/5212537551
https://datadoghq.atlassian.net/browse/DSCVR-165

Test included in the PR.

Also tested manually (see linked wiki page), that currently requires code changes to enable the process collector.


vitkyrka:
The extra imports in `otel-agent` will be handled by https://github.com/DataDog/datadog-agent/pull/38784

comp/core/autodiscovery/providers/process_log.go
```diff
@@ -0,0 +1,328 @@
...
+			case <-ctx.Done():
+				p.workloadmetaStore.Unsubscribe(inCh)
+				return
+			case evBundle, ok := <-inCh:
+				if !ok {
+					return
+				}
+
+				outCh <- p.processEvents(evBundle)
+				evBundle.Acknowledge()
```
davidor, line 93: I don’t know all the details about this provider, so I’m asking just in case: should the ack happen before processing the events? If it can’t, there should probably be a comment explaining why.

Here’s the comment that mentions `Acknowledge()` should be called as soon as possible: https://github.com/DataDog/datadog-agent/blob/de374ec620e62698c4286f0bc5fa312e92bd13e5/comp/core/workloadmeta/def/types.go#L1430

comp/core/autodiscovery/providers/process_log.go
```diff
@@ -0,0 +1,328 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
```
davidor, line 1: Should agent-discovery own or co-own this file?

comp/core/autodiscovery/providers/process_log_test.go
```diff
@@ -0,0 +1,907 @@
...
+		Type:   workloadmeta.EventTypeSet,
+		Entity: process,
+	}
+	setBundle := workloadmeta.EventBundle{
+		Events: []workloadmeta.Event{setEvent},
+	}
+	changes := p.processEventsNoVerifyReadable(setBundle)
+	t.Logf("Changes: %+v", changes)
+	t.Logf("Schedule count: %d", len(changes.Schedule))
+	t.Logf("Unschedule count: %d", len(changes.Unschedule))
```
davidor: Can we delete these lines and rely on the asserts to reduce noise?

comp/core/autodiscovery/providers/process_log_test.go
```diff
@@ -0,0 +1,907 @@
...
+	found1, found2 = false, false
+	for _, config := range changes.Unschedule {
+		if config.Name == "process-test-service-_var_log_test.log" {
+			found1 = true
+		} else if config.Name == "process-test-service-2-_var_log_test2.log" {
+			found2 = true
+		}
+	}
+	assert.True(t, found1)
+	assert.True(t, found2)
```
davidor: Nit: I think there's a cleaner pattern for this kind of check:

```go
	var actualNames []string
	for _, c := range changes.Unschedule {
		actualNames = append(actualNames, c.Name)
	}

	expectedNames := []string{
		"process-test-service-_var_log_test.log",
		"process-test-service-2-_var_log_test2.log",
	}

	assert.Equal(t, expectedNames, actualNames) // Or assert.ElementsMatch if order is not important
```

comp/core/autodiscovery/providers/process_log_test.go
```diff
@@ -0,0 +1,907 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-present Datadog, Inc.
+
+//go:build linux
```
davidor: Do we need this? I'm not sure if there's anything linux-specific on this particular file.

If it's needed it should probably also be in `comp/core/autodiscovery/providers/process_log.go`.

comp/core/autodiscovery/providers/process_log.go
vitkyrka: You're right, from the comments this should be acking before. Fixed.

comp/core/autodiscovery/providers/process_log.go
vitkyrka: Added to CODEOWNERS.

comp/core/autodiscovery/providers/process_log_test.go
vitkyrka: Done.

comp/core/autodiscovery/providers/process_log_test.go
vitkyrka: There are a couple tests which uses things like `chmod(0)` which don't do what we want on Windows. I've skipped them explicitly and removed the build tag.

comp/core/autodiscovery/providers/process_log_test.go
vitkyrka: I've reworked this and other places to use asserts on a map to clean up these loops (the suggested pattern wouldn't have worked for the other test cases which had to assert some property of the items too).

ddrthall (approved):

comp/core/autodiscovery/providers/process_log.go
```diff
@@ -0,0 +1,328 @@
...
+
+	filter := workloadmeta.NewFilterBuilder().
+		AddKindWithEntityFilter(workloadmeta.KindProcess, func(e workloadmeta.Entity) bool {
+			process, ok := e.(*workloadmeta.Process)
+			if !ok {
+				return false
+			}
+
+			// Ignore containers since log files inside them usually can't be
+			// accessed from here since they are in a different namespace.
```
adel121, line 76: Question:

If I understand correctly, the logs agent accesses logs by looking at files under `/var/log/pods` 

We also have `/var/log/containers/` 

So why is being in a different namespace an issue when it comes to collecting logs? (My thought is that `/var/log/containers/` are host level files that are not scoped to namespaces, so in theory they should be accessible by the log agent)

comp/core/autodiscovery/providers/process_log.go
```diff
@@ -0,0 +1,328 @@
...
+			}
+		}
+	}()
+
+	return outCh
+}
+
+func (p *processLogConfigProvider) generateServiceLogKey(logPath string) string {
+	return strings.ReplaceAll(logPath, "/", "_")
+}
```
adel121: Just to be sure: 

I assume this should return a unique key (correct me if I wrong). 

Isn't there a chance for collision here? 

like if we had something like:

`some-dir/logs/one`

and

`some-dire_log_one` 


Both will get the same key.

adel121 (commented):
Looks good 👍 

Left 2 minor comments/questions.

comp/core/autodiscovery/providers/process_log.go
vitkyrka: You're correct that container runtimes store the stderr/stdout from containers in paths like `/var/log/containers` on the host, but in this case we are ignoring logs files written _inside_ containers (not via the container's stdout/stderr). See also https://datadoghq.atlassian.net/wiki/spaces/DSCVR/pages/5212537551/#Log-files-inside-containers

comp/core/autodiscovery/providers/process_log.go
vitkyrka: You're right. I realized there's no need to do this massaging of the path so I just used the raw path now which should fix this.

comp/core/autodiscovery/providers/process_log.go
adel121: makes sense 

Thanks for the clarification.

adel121 (approved):

CelianR (approved):

PR #38010, opened 2025-06-17
feat: create new fingerprinting feature to track truncation and rotation

lucastemb:
### What does this PR do?

This PR implements improvements to log rotation detection by introducing a fingerprinting mechanism that enhances Datadog Agent’s ability to accurately identify log file rotations. The change replaces or supplements the current inode-based detection logic with a content-based fingerprint, improving resilience to edge cases such as inode reuse or file truncation.

### Motivation
The existing log rotation detection mechanism can misidentify log file continuations or rotations under certain conditions, particularly in environments where inode reuse is common or files are renamed without changes to metadata. These failures lead to incorrect log collection behavior such as duplicated logs, data loss, or broken pipelines.

This PR addresses those issues by implementing a fingerprinting strategy based on content hashes of the beginning of the file, which remains stable across common log rotation patterns. This change aims to:

- Improve rotation detection accuracy
- Reduce false positives in file change detection
- Enable safe and performant tracking of log files over time

The design and rationale are described in detail in [RFC: Log Rotation Detection Improvements via Fingerprinting](https://datadoghq.atlassian.net/wiki/spaces/AL/pages/5095229352).


comp/logs/auditor/impl/auditor.go
```diff
@@ -234,7 +257,7 @@ func (a *registryAuditor) run() {
 			}
 			// update the registry with the new entry
 			for _, msg := range payload.MessageMetas {
-				a.updateRegistry(msg.Origin.Identifier, msg.Origin.Offset, msg.Origin.LogSource.Config.TailingMode, msg.IngestionTimestamp)
+				a.updateRegistry(msg.Origin.Identifier, msg.Origin.Offset, msg.Origin.LogSource.Config.TailingMode, msg.IngestionTimestamp, msg.Origin.Fingerprint, msg.Origin.LogSource.Config.FingerprintConfig)
```
ddrthall: We might be fingerprinting only via global configs, rather than the source configs being persisted here.

pkg/logs/launchers/file/launcher.go
```diff
@@ -170,26 +179,45 @@ func (s *Launcher) scan() {
...
+				didRotate, err = tailered.DidRotate()
+			}
 			if err != nil {
 				log.Debugf("failed to detect log rotation: %v", err)
 				continue
 			}
 			if didRotate {
 				// restart tailer because of file-rotation on file
-				succeeded := s.restartTailerAfterFileRotation(tailer, file)
+				succeeded := s.restartTailerAfterFileRotation(tailered, file)
```
ddrthall, line 206: This is dangerous to call in the checksum enabled case, as restarting a tailer here might cause log duplication. If we know we don't want a new tailer due to the inability to fingerprint we shouldn't start one up.

pkg/logs/launchers/file/position.go
```diff
@@ -20,13 +21,20 @@ func Position(registry auditor.Registry, identifier string, mode config.TailingM
...
+	}
+	newFingerprint := tailerfile.ComputeFingerprint(filePath, fingerprintConfig)
 
 	switch {
 	case mode == config.ForceBeginning:
 		offset, whence = 0, io.SeekStart
 	case mode == config.ForceEnd:
 		offset, whence = 0, io.SeekEnd
-	case value != "":
+	case value != "" && fingerprintConfig != nil && previousFingerprint == newFingerprint: //and fingerprint valid (fingerprint stored oldconfig and recalculate oldconfig the same)
```
ddrthall: We don't want to break auditor functionality for newly upgraded agents (which can't have a stored fingerprint config) and agents without checksums enabled (which look like they are assigned a nil config in the current code version).

pkg/logs/launchers/file/launcher.go
lucastemb: Just now realizing that if `restartTailerAfterFileRotation` does not succeed, we lose that old file information for both the checksum and default case. When it's tailed on second pass, it will be treated as a new file, not as a rotated one. 🤔

pkg/config/setup/config.go
```diff
@@ -1560,6 +1583,15 @@ func logsagent(config pkgconfigmodel.Setup) {
 	config.BindEnvAndSetDefault("logs_config.socks5_proxy_address", "")
 	// disable distributed senders
 	config.BindEnvAndSetDefault("logs_config.disable_distributed_senders", false)
+	// determines fingerprinting strategy to detect rotation and truncation
+	config.BindEnvAndSetDefault("logs_config.rotation_detection_strategy", DefaultRotationDetectionStrategy)
+	// consolidated fingerprint configuration with defaults
+	config.BindEnvAndSetDefault("logs_config.fingerprint_config", map[string]interface{}{
+		"max_lines":            DefaultFingerprintingMaxLines,
```
ddrthall: What do you think of something like
```fingerprint_config:
    type: "lines"
    count: 1
    max_bytes: 100000000000
    count_to_skip: 0
```
and
```
fingerprint_config:
    type: "bytes"
    count: 2048
    count_to_skip: 24
```

This reduces the overloading on max_bytes, where it means radically different things in different contexts

comp/logs/agent/config/integration_config.go
```diff
@@ -358,6 +386,29 @@ func (c *LogsConfig) validateTailingMode() error {
 	return nil
 }
 
+func (c *LogsConfig) validateFingerprintConfig() error {
+	// If log level config has empty or incomplete values, populate the FingerprintConfig with what is in the global config
+	globalConfig, err := GlobalFingerprintConfig(pkgconfigsetup.Datadog())
```
ddrthall, line 413: Having the global config override defaults on a point-by-point basis can be confusing for users. E.g. if I'm a developer setting values for my app in an integration_config, I want to be able to set the number of lines to scan to 2 without being surprised by a change in the datadog global config months later that drops in a lines to skip of 2. 
If an integration fingerprint config is set, it should completely override the global config.

brett0000FF (approved):

releasenotes/notes/fingerprinting-logic-a51a9e92474ba9ef.yaml
```diff
@@ -0,0 +1,10 @@
+# Each section from every release note are combined when the
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+features:
+  - |
+    Adds the logic required to compute checksum fingerprints to more accurately detect instances of rotation and truncation.
```
hush-hush: We should mentioned that this feature is for logs. The changelog is meant for users/customers so we have to be precise.

comp/logs/agent/config/config.go
```diff
@@ -160,6 +165,97 @@ func IsExpectedTagsSet(coreConfig pkgconfigmodel.Reader) bool {
...
+		log.Debugf("GlobalFingerprintConfig: unmarshaling from string: %s", s)
+		err = json.Unmarshal([]byte(s), &config)
+	} else {
+		log.Debugf("GlobalFingerprintConfig: using structure.UnmarshalKey")
+		// Fallback to structure.UnmarshalKey
+		err = structure.UnmarshalKey(coreConfig, "logs_config.fingerprint_config", &config, structure.ConvertEmptyStringToNil)
+	}
+	if err != nil {
+		return nil, err
+	}
```
hush-hush: Why not just UnmarshalKey the config here ?

```suggestion
	config := &FingerprintConfig{}

		err = structure.UnmarshalKey(coreConfig, "logs_config.fingerprint_config", &config, structure.ConvertEmptyStringToNil)
	if err != nil {
		return nil, err
	}
```

comp/logs/agent/config/config.go
```diff
@@ -160,6 +165,97 @@ func IsExpectedTagsSet(coreConfig pkgconfigmodel.Reader) bool {
...
+		return ErrEmptyFingerprintConfig
+	}
+
+	// Check if any fingerprint config fields are set
+	hasAnyConfig := config.FingerprintStrategy != "" || config.Count != 0 || config.CountToSkip != 0 || config.MaxBytes != 0
+
+	if !hasAnyConfig {
+		// If no fields are set, return empty error
+		return ErrEmptyFingerprintConfig
+	}
```
hush-hush: You should not have to validate such things. The config will either return the defaults or the value set by the customer.

In the rest of the Agent we treat `0` and `empty string` values explicitly set by customer as disabling the feature. In general, we should never ignore configuration set by customers.

A better logic would be to rely on the config returning the default and check if the feature is enabled or not. Function like `applyFingerprintDefaults` should not exists as the code should behave the same no matter the source of a config (default, customer, RC, CLI, env vars, ...).

hush-hush (changes requested):

comp/logs/agent/config/config.go
```diff
@@ -160,6 +165,97 @@ func IsExpectedTagsSet(coreConfig pkgconfigmodel.Reader) bool {
...
+
+	return config, nil
+}
+
+// ValidateFingerprintConfig validates the fingerprint config and returns an error if the config is invalid
+func ValidateFingerprintConfig(config *FingerprintConfig) error {
+	if config == nil {
+		return ErrEmptyFingerprintConfig
+	}
+
```
hush-hush: If we were to not take a pointer we would not have to worry about `nil` here. Seems like nowhere in the code we pass a nil pointer to this func.

comp/logs/agent/config/config.go
lucastemb: @hush-hush Just to clarify, does this suggested change mean we don't want to allow for deserialization from environment variables for our fingerprintingConfig? I followed the conventions that were established in the GlobalProcessingRules function, so not sure if you see it best to just remove lines (189-192)

hush-hush: Yes, you should not manually parse env vars in your code, the config does that. The configuration is in charge of aggregating all the sources we have for configuration and offer a single view to the rest of the code base. This is important because we have sources of config with higher priority than env vars (CLI for example).

If a setting you have has complex type associated to it you can use one of the `ParseEnvAs*` method to tell the config how to load env vars for it. In your case you could use [this](https://github.com/DataDog/datadog-agent/blob/main/pkg/config/model/types.go#L247) for it.
But actually a better design would be use dedicated settings since you know exactly what field you're waiting for:

ie:
```
	config.BindEnvAndSetDefault("logs_config.fingerprint_config.count", DefaultFingerprintingMaxLines)
	config.BindEnvAndSetDefault("logs_config.fingerprint_config.max_bytes", DefaultFingerprintingMaxBytes)
	config.BindEnvAndSetDefault("logs_config.fingerprint_config.count_to_skip", DefaultLinesOrBytesToSkip)
	config.BindEnvAndSetDefault("logs_config.fingerprint_config.fingerprint_strategy", DefaultFingerprintStrategy)
```
You can still use `UnmarshalKey` to load a chunk of the configuration into a struct.

rdesgroppes:
There are too many changes to review:
image

PR #38027, opened 2025-06-17
Go DI: Surface program ID in events 

grantseltzer:
### What does this PR do?

Program and event IDs were not being surfaced in events off the ringbuffer in Go DI. This fixes this.

### Motivation

We need these IDs to correlate events with the IR definitions.

Ran tests locally.


ajwerner (commented):
For one, I don't think we need the event ID in the header given the root data item already implies it (let's instead remove it from the header -- no reason to waste those bytes). The ProgramID is good and all but let's just shove it into a constant global variable as opposed to puitting it into the params; by definition the prog_id is the same for everything in the program.

Also, when you change the output, I think you'll need to rewrite the tests with `REWRITE=1`

grantseltzer:
> For one, I don't think we need the event ID in the header given the root data item already implies it (let's instead remove it from the header -- no reason to waste those bytes). 

Fair enough.

>The ProgramID is good and all but let's just shove it into a constant global variable as opposed to puitting it into the params; by definition the prog_id is the same for everything in the program.

Done.

> Also, when you change the output, I think you'll need to rewrite the tests with `REWRITE=1`

Done.

ajwerner (approved):
LGTM

PR #38029, opened 2025-06-17
[clusteragent/autoscaling] Add indexing to autoscaling workload store

jennchenn:
### What does this PR do?

Adds support for indexers to the autoscaling store and implements a map indexer on the workload store

### Motivation

The call to get the autoscaler prior to patching it was slow:
```
	podAutoscalers := pa.store.GetFiltered(func(podAutoscaler model.PodAutoscalerInternal) bool {
		if podAutoscaler.Namespace() == pod.Namespace &&
			podAutoscaler.Spec().TargetRef.Name == ownerRef.Name &&
			podAutoscaler.Spec().TargetRef.Kind == ownerRef.Kind &&
			podAutoscaler.Spec().TargetRef.APIVersion == ownerRef.APIVersion {
			return true
		}
		return false
	})
```
in large clusters with many workload (e.g. 1000+), we also had a memory overhead as we allocated a slice with the size of the number of workloads in the store to filter on. In most cases, we expect this to only return one autoscaler (i.e. one autoscaler<>one workload); adding indexing allows us to easily search through and quickly return the autoscaler we want to scale on.

1. Deploy these changes
2. Verify that recommendations are being applied as usual

Also added unit tests to verify the change.

To allow for indexing, we are allocating a map of owner=>pod autoscaler IDs for each autoscaler. In large clusters with many autoscalers this can cause a memory overhead. 

### Additional Notes

Do we want to allow for

### DIFF
--- a/pkg/clusteragent/autoscaling/store.go
+++ b/pkg/clusteragent/autoscaling/store.go
@@ -36,18 +36,22 @@ type Store[T any] struct {
 	lock          sync.RWMutex
 	observers     map[storeOperation][]ObserverFunc
 	observersLock sync.RWMutex
+	indexMap      map[any][]string
+	getIndexKey   func(*T) any
 }
 
 type storeOperation int
 
 // NewStore creates a new NewStore
-func NewStore[T any]() *Store[T] {
+func NewStore[T any](getIndexKey func(*T) any) *Store[T] {
 	return &Store[T]{
 		store: make(map[string]T),
 		observers: map[storeOperation][]ObserverFunc{
 			setOperation:    make([]ObserverFunc, 0),
 			deleteOperation: make([]ObserverFunc, 0),
 		},
+		indexMap:    make(map[any][]string),
+		getIndexKey: getIndexKey,
 	}
 }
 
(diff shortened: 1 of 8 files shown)


pkg/clusteragent/autoscaling/store.go
```diff
@@ -98,6 +116,26 @@ func (s *Store[T]) GetFiltered(filter func(T) bool) []T {
...
+			objects = append(objects, obj)
+		}
+	}
+
+	return objects
+}
+
 // Update updates all objects in the store with the result of the `updator` function.
 // Updator func is expected to return the new object and a boolean indicating if the object has changed.
 // The object is updated only if boolean is true, observers are notified only for updated objects after all objects have been updated.
```
clamoriniere, line 141: I'm wondering if also the Update() function must be modified.
From what I understood we use the ownerreference to generate they indexKey. However the owner-reference contains the APIGroup information, and I'm sure how it should work when an object is move to a different apigroup version.

vboulineau:
Keeping it open as still needed but delayed

PR #38030, opened 2025-06-17
David ioinovitch/ntp ad discovery

davidioin:
### What does this PR do?
Enhaces Windows NTP check for `use_local_defined_servers: true`, enabling it to correctly discover and query the Primary Domain Controller (PDC) for time on domain-joined machines. It includes an intelligent fallback to registry-configured NTP servers if the PDC is unavailable. Additionally, it improves startup resilience, reduces log noise by making server list comparisons order-insensitive, and clarifies error messages.

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1549
Fix offset reported by domain joined hosts as previous check did not account for NT5DS time synchronization in domain environments, leading to inaccurate ntp.offset metrics.

The changes were validated through specific unit tests and manual end-to-end tests
Manual test was performed on 2 Windows hyper V hosts, one being a Domain Controller and the other a Domain Client connected to the DC. 
Tested to ensure:
On a domain-joined host, the system successfully identified and queried the Primary Domain Controller (PDC).
On a domain-joined host that was disconnected from the network, the system successfully fell back to using registry-configured NTP servers.
On a non-domain-joined host, the system correctly utilized the registry for NTP server information.
On a domain controller, the system correctly avoided using itself as a time source.


pkg/collector/corechecks/net/ntp/ntp.go
```diff
@@ -109,7 +123,7 @@ func (c *ntpConfig) parse(data []byte, initData []byte, getLocalServers func() (
 		if err != nil {
 			return err
 		}
-		log.Debugf("Detected local defined servers: [ %s ]", strings.Join(defaultDatadogPool, ", "))
```
clarkb7: nice catch

pkg/collector/corechecks/net/ntp/ntp.go
```diff
@@ -150,7 +164,7 @@ func (c *ntpConfig) parse(data []byte, initData []byte, getLocalServers func() (
 // Configure configure the data from the yaml
 func (c *NTPCheck) Configure(senderManager sender.SenderManager, integrationConfigDigest uint64, data integration.Data, initConfig integration.Data, source string) error {
 	cfg := new(ntpConfig)
-	err := cfg.parse(data, initConfig, getLocalDefinedNTPServers)
+	err := cfg.parse(data, initConfig, getLocalDefinedNTPServersFunc)
 	if err != nil {
 		log.Errorf("Error parsing configuration file: %s", err)
 		return err
```
clarkb7, line 171: `DsGetDCName` could potentially fail (maybe no internet connectivity, maybe Agent started before netlogon service, etc.), and we wouldn't want that to prevent the check from running now that we'll retry upon next check iteration anyway. Lets update the call to `getLocalServers` inside `cfg.parse` to log a warning instead of returning an error.

current code, I can't comment directly there b/c there's no changed lines
```
		localNtpServers, err = getLocalServers()
		if err != nil {
			return err
		}
```

pkg/collector/corechecks/net/ntp/ntp_local_defined_server_windows.go
```diff
@@ -11,26 +11,141 @@ import (
...
+	DomainGuid                  syscall.GUID
+	DomainName                  *uint16
+	DnsForestName               *uint16
+	Flags                       uint32
+	DcSiteName                  *uint16
+	ClientSiteName              *uint16
+}
+
 func getLocalDefinedNTPServers() ([]string, error) {
+	// Use native Windows registry API to get the Type value
```
clarkb7, line 61: Link to MSDN for future reference

```suggestion
	// Use native Windows registry API to get the Type value
	// https://learn.microsoft.com/en-us/windows-server/networking/windows-time-service/windows-time-service-tools-and-settings?tabs=parameters
```

pkg/collector/corechecks/net/ntp/ntp_local_defined_server_windows.go
```diff
@@ -11,26 +11,141 @@ import (
...
+
+	// Standard registry method for NTP servers
+	return getNTPServersFromRegistry(k)
+}
+
+func discoverPDC() ([]string, error) {
+	// Use DsGetDcName to find the PDC Emulator
+	var dcInfo *DOMAIN_CONTROLLER_INFO
+
+	// Call DsGetDcName with DS_PDC_REQUIRED flag to get the PDC
```
clarkb7, line 114: should we include the `DS_TIMESERV_REQUIRED` flag?

pkg/collector/corechecks/net/ntp/ntp_local_defined_server_windows.go
```diff
@@ -43,11 +158,13 @@ func getNptServersFromRegKeyValue(regKeyValue string) ([]string, error) {
 	var servers []string
 	for _, f := range fields {
 		server := strings.Split(f, ",")[0]
-		servers = append(servers, server)
+		if server != "" {
```
clarkb7, line 167: nice catch.

pkg/collector/corechecks/net/ntp/ntp_local_defined_server_windows_test.go
```diff
@@ -13,25 +13,95 @@ import (
...
+		},
+		{
+			name:            "Whitespace only",
+			regKeyValue:     "   \t\n   ",
+			expectedServers: nil,
+			expectError:     true,
+		},
+		{
+			name:            "Server with port",
+			regKeyValue:     "ntp.example.com:123,0x9",
```
clarkb7: do we support a custom port in the host field? It looks like the check has a separate option for it.

pkg/collector/corechecks/net/ntp/ntp_local_defined_server_windows.go
```diff
@@ -11,26 +11,141 @@ import (
...
+		return nil, fmt.Errorf("DsGetDcName failed with error code: %d (%v)", ret, err)
+	}
+
+	// Ensure we free the buffer when done
+	defer procNetApiBufferFree.Call(uintptr(unsafe.Pointer(dcInfo)))
+
+	// Convert the PDC name to a Go string
+	if dcInfo.DomainControllerName != nil {
+		pdcName := windows.UTF16PtrToString(dcInfo.DomainControllerName)
+		// Remove leading backslashes if present
```
clarkb7, line 134: In what cases does this API return a name with leading slashes?

pkg/collector/corechecks/net/ntp/ntp_local_defined_server_windows.go
```diff
@@ -11,26 +11,141 @@ import (
...
+	// Standard registry method for NTP servers
+	return getNTPServersFromRegistry(k)
+}
+
+func discoverPDC() ([]string, error) {
+	// Use DsGetDcName to find the PDC Emulator
+	var dcInfo *DOMAIN_CONTROLLER_INFO
+
+	// Call DsGetDcName with DS_PDC_REQUIRED flag to get the PDC
+	ret, _, err := dsGetDcNameCall(
```
clarkb7, line 115: lets pass `DS_AVOID_SELF` so a DC doesn't try to use itself as a time source

clarkb7, line 115: should we include the `DS_BACKGROUND_ONLY` flag?

pkg/collector/corechecks/net/ntp/ntp.go
```diff
@@ -76,6 +76,20 @@ func (c *NTPCheck) String() string {
 
 // for testing
 var getCloudProviderNTPHosts = cloudproviders.GetCloudProviderNTPHosts
+var getLocalDefinedNTPServersFunc = getLocalDefinedNTPServers
+
+// stringSlicesEqual compares two string slices for equality
+func stringSlicesEqual(a, b []string) bool {
```
clarkb7, line 82: this function also asserts the order of the elements is the same. I think for this check we don't care about the order.

pkg/collector/corechecks/net/ntp/ntp_local_defined_server_windows.go
davidioin: The goal of the NT5DS logic is to find the Primary Domain Controller (PDC), which is the authoritative time source for the domain. We already use the DS_PDC_REQUIRED flag, which guarantees the returned server is the PDC. According to Microsoft's documentation, the PDC is implicitly a time server for the domain. The DS_TIMESERV_REQUIRED flag is for finding any domain controller that is a time server, not necessarily the main one. Since our goal is to find the most authoritative source (the PDC), our current approach is correct and more specific. Adding the other flag is unnecessary.
Implementation: No code change is needed.

**answer from cursor, looks right to me

pkg/collector/corechecks/net/ntp/ntp_local_defined_server_windows.go
davidioin: Screenshot 2025-06-26 155902

pkg/collector/corechecks/net/ntp/ntp_local_defined_server_windows.go
davidioin: I'd say no: documentation warns that it can return old stale data, and that "Note that if the DS_BACKGROUND_ONLY flag is specified, DsGetDcName will never try to rediscover the domain controller name, since the point of that flag is to force DsGetDcName to use the cached domain controller name even if it is expired." (defeating the purpose of the PDC discovery) so I'd say better to perform a live query like we have atm.

pkg/collector/corechecks/net/ntp/ntp.go
```diff
@@ -107,7 +121,7 @@ func (c *ntpConfig) parse(data []byte, initData []byte, getLocalServers func() (
 	if c.instance.UseLocalDefinedServers {
 		localNtpServers, err = getLocalServers()
 		if err != nil {
-			return err
+			log.Warnf("Could not get local NTP servers, falling back to configured hosts: %v", err)
 		}
 		log.Debugf("Detected local defined servers: [ %s ]", strings.Join(localNtpServers, ", "))
```
clarkb7: don't log the success message when there is an error

```suggestion
			log.Warnf("Could not get local NTP servers, falling back to configured hosts: %v", err)
		} else {
		    log.Debugf("Detected local defined servers: [ %s ]", strings.Join(localNtpServers, ", "))
		}
```

pkg/collector/corechecks/net/ntp/ntp_local_defined_server_windows.go
```diff
@@ -7,30 +7,151 @@
...
+// Constants for DsGetDcName
+// Name is intended to match the Windows API name see MSDN: https://learn.microsoft.com/en-us/windows/win32/api/dsgetdc/nf-dsgetdc-dsgetdcnamew
+const (
+	DS_PDC_REQUIRED    = 0x00000080
+	DS_RETURN_DNS_NAME = 0x40000000
+	DS_AVOID_SELF      = 0x00004000
 )
 
+// DOMAIN_CONTROLLER_INFO structure
+// Name is intended to match the Windows API name see MSDN: https://learn.microsoft.com/en-us/windows/win32/api/dsgetdc/ns-dsgetdc-domain_controller_infoa
```
clarkb7: link to the wide char version of the API

```suggestion
// Name is intended to match the Windows API name see MSDN: https://learn.microsoft.com/en-us/windows/win32/api/dsgetdc/ns-dsgetdc-domain_controller_infow
```

brett0000FF (approved):

releasenotes/notes/ntp-windows-resiliency-fixes-17c350b623dc859d.yaml
```diff
@@ -0,0 +1,21 @@
...
+    The NTP check on Windows is now more resilient. If a machine is configured to use an Active Directory domain controller for time (`NT5DS`) but cannot reach it, the check will now automatically fall back to using the NTP servers listed in the Windows registry instead of failing completely.
+  - |
+    The check is now smarter about detecting changes to the NTP server list. It will no longer create a log entry if the discovered servers are the same but in a different order, reducing log noise.
+  - |
+    When discovering a domain controller for time synchronization, the check will now avoid using the machine itself if it is also a domain controller.
+  - |
+    Fixed a misleading log message that occurred when NTP domain controller discovery failed on Windows. The log will no longer include ``(The operation completed successfully.)`` alongside the error code.
+  - |
+    The NTP check will no longer fail to start if the initial discovery of local NTP servers fails. It will instead log a warning and retry on the next run.
+
```
clarkb7: lets try to consolidate these.
Information regarding the new domain controller detection behavior in one entry.
Information regarding the new retry behavior in one entry.

clarkb7 (approved):

pkg/collector/corechecks/net/ntp/ntp.go
louis-cqrl, line 82: can't we use [`slice.Equal`](stringSlicesEqual)?

PR #38032, opened 2025-06-18
remove debug agent.exe from windows container

clarkb7:
### What does this PR do?
Replaces the duplicate, debug build, `bin/agent.exe` with the actual, release/stripped, `bin/agent/agent.exe` in the Windows container.

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1352
Reduce container size to pass static size checks.

check windows container size is reduced, view static size check report

check agent still runs in container
```
docker run --rm -it -e "DD_API_KEY=XXX" -e "DD_HOSTNAME=XXX" --isolation process datadog/agent-dev:branden-clark-remove-debug-agent-py3-win-servercore-ltsc2022
```
the duplicate `agent.exe` is coming from the zip package, so this is a bandaid just for the container to get the static size check to pass. We should fix it in the zip package itself, but we have to investigate what else depends on it first.

### DIFF
--- a/.gitlab/container_build/docker_windows.yml
+++ b/.gitlab/container_build/docker_windows.yml
@@ -23,6 +23,9 @@
     - Expand-Archive datadog-agent-latest.amd64.zip
     - Remove-Item datadog-agent-latest.amd64.zip
     - Get-ChildItem -Path datadog-agent-* | Rename-Item -NewName "Datadog Agent"
+    # Omnibus uses a slightly different path than the actual install
+    # TODO: This overwrites an extra agent.exe, we should also remove it from the zip
+    - mv "Datadog Agent\bin\agent\agent.exe" "Datadog Agent\bin\agent.exe"
     - popd
 
     - get-childitem ${BUILD_CONTEXT}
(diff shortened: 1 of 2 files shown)


.gitlab/container_build/docker_windows.yml
```diff
@@ -23,6 +23,9 @@
     - Expand-Archive datadog-agent-latest.amd64.zip
     - Remove-Item datadog-agent-latest.amd64.zip
     - Get-ChildItem -Path datadog-agent-* | Rename-Item -NewName "Datadog Agent"
+    # Omnibus uses a slightly different path than the actual install
+    # TODO: This overwrites an extra agent.exe, we should also remove it from the zip
```
Pythyu, line 27: Just to clarify, Is the `TODO` for a separate PR or an incoming extra change in this PR ?

Pythyu (approved):
LGTM

.gitlab/container_build/docker_windows.yml
clarkb7: separate PR/future work. I'm not sure what's creating `bin/agent.exe` in the zip package.

jack0x2 (approved):

buraizu (approved):

PR #38036, opened 2025-06-18
Extend setsockopt Tracking with Detailed Filter and Socket Data

theop-dd:
### What does this PR do?
Adds additional information and data related to the setsockopt hook.

Socket Information:
- Socket type
- Socket family
- Socket protocol

Filter Information:
- Disassembled filter
- Filter hash
- Truncated Filter : add a flag and a metric associated to it to check if the retrieved filters are truncated or complete.

In addition, this PR provides new helpers that can be used:
- get_protocol_from_sock is able to get the socket protocol field from the struct sock and handle the different offset depending on the kernel version.

### Motivation

To more accurately detect BPF doors, additional data was required from the setsockopt syscall. Specifically, we need to retrieve details about the socket to which the filter is attached, as well as key information from the filter itself. This data enables us to define more precise detection rules based on specific filter characteristics.

The setsockopt test was updated to verify the tracking of the newly added information.


pkg/security/ebpf/c/include/hooks/setsockopt.h
```diff
@@ -3,64 +3,158 @@
...
+    fill_container_context(entry, &event->container);
+    fill_span_context(&event->span);
+
+    if (syscall->setsockopt.filter_len > MAX_BPF_FILTER_LEN){
+        return 0;
+    }
+    send_event_with_size_ptr(ctx, EVENT_SETSOCKOPT, event, (offsetof(struct setsockopt_event_t, bpf_filters_buffer) + sizeof(struct sock_filter) * syscall->setsockopt.filter_len) );
 
     // if the tail call fails, we need to pop the syscall cache entry
     pop_syscall(EVENT_SETSOCKOPT);
```
YoannGh: Unrelated to the changes of your PR but you don't need to pop the syscall here as you already do it at the beginning of the function

docs/cloud-workload-security/backend_linux.md
```diff
@@ -4747,8 +4803,14 @@ Workload Protection events for Linux systems have the following JSON schema:
...
 | Field | Description |
 | ----- | ----------- |
+| `socket_type` | Socket file descriptor |
+| `socket_protocol` | Socket protocol |
+| `socket_family` | Socket family |
 | `level` | Level at which the option is defined |
 | `optname` | Name of the option being set |
+| `filter_len` | Length of the filter |
+| `filter` | Filter data |
+| `filter_hash` | Filter Hash |
```
buraizu:
```suggestion
| `filter_hash` | Filter hash |
```

docs/cloud-workload-security/backend_linux.schema.json
```diff
@@ -1715,20 +1715,48 @@
...
+          "type": "integer",
+          "description": "Length of the filter"
+        },
+        "filter": {
+          "type": "string",
+          "description": "Filter data"
+        },
+        "filter_hash": {
+          "type": "string",
+          "description": "Filter Hash"
```
buraizu:
```suggestion
          "description": "Filter hash"
```

buraizu (approved):
Approving with a couple of minor edits requested for casing consistency

pkg/security/ebpf/c/include/hooks/setsockopt.h
theop-dd: You are completely right, I'll delete that 👍

pkg/security/ebpf/c/include/events_definition.h
```diff
@@ -538,8 +540,13 @@ struct setsockopt_event_t {
     struct container_context_t container;
     struct syscall_t syscall;
 
+    short socket_type;
+    u16 socket_protocol; // socket protocol
+    u16 socket_family;
     int level;
     int optname;
+    unsigned short filter_len;
```
spikat: Maybe we should align variables here? Same in `include/structs/syscalls.h`

pkg/security/ebpf/c/include/hooks/setsockopt.h
```diff
@@ -3,64 +3,150 @@
...
+    if (ret < 0) {
+        return 0;
+    }
+    unsigned short prog_len = prog.len;
+    syscall->setsockopt.filter_len = prog_len;
+    int key = 0;
+    struct setsockopt_event_t *event = bpf_map_lookup_elem(&setsockopt_event, &key);
+    if (event && 0 < prog.len && prog.len <= MAX_BPF_FILTER_LEN) { 
+        bpf_probe_read(&event->bpf_filters_buffer, (sizeof(struct sock_filter) * prog_len) &MAX_BPF_FILTER_LEN , prog.filter);  
+    }
```
spikat: Could we add a "truncated" boolean and return the first MAX_BPF_FILTER_LEN bytes of the filter, instead of just returning nothing? Also, having a metric on that to track how often it happen will be nice

pkg/security/ebpf/c/include/constants/offsets/network.h
```diff
@@ -20,6 +20,18 @@ __attribute__((always_inline)) u16 get_skc_num_from_sock_common(struct sock_comm
     bpf_probe_read(&skc_num, sizeof(skc_num), (void *)sk + sock_common_skc_num_offset);
     return htons(skc_num);
 }
+__attribute__((always_inline)) u16 get_protocol_from_sock(struct sock *sk) {
+    u64 sock_sk_protocol_offset;
+    LOAD_CONSTANT("sock_sk_protocol_offset", sock_sk_protocol_offset);
+    if (sock_sk_protocol_offset == 0) {
+        LOAD_CONSTANT("sock_sk_txhash_offset", sock_sk_protocol_offset);
+        sock_sk_protocol_offset += 9; // sk_protocol is 5 bytes after sk_txhash in a bitfield
```
spikat: I remember having issues (on some kernels?) with offsets on bitfields. Curious to see if the KMT tests will pass :)

pkg/security/probe/constantfetch/constant_names.go
```diff
@@ -108,6 +108,7 @@ const (
 	OffsetNameFlowI6StructSADDR         = "flowi6_saddr_offset"
 	OffsetNameFlowI4StructULI           = "flowi4_uli_offset"
 	OffsetNameFlowI6StructULI           = "flowi6_uli_offset"
+	OffsetNameSockStructSKProtocol      = "sock_sk_protocol_offset"
```
spikat: When we add a new offset we must add a fallback method if known possible values (or at least one)

pkg/security/tests/setsockopt_test.go
spikat, line 1: Could you add a test with a filter of a len > of the limit?

pkg/security/secl/model/model_unix.go
```diff
@@ -1009,6 +1009,13 @@ type SetrlimitEvent struct {
...
+	SocketFamily       uint16 `field:"socket_family"`                                                   // SECLDoc[socket_family] Definition:`Socket family`
+	Level              uint32 `field:"level"`                                                           // SECLDoc[level] Definition:`Socket level`
+	OptName            uint32 `field:"optname"`                                                         // SECLDoc[optname] Definition:`Socket option name`
+	FilterLen          uint16 `field:"filter_len"`                                                      // SECLDoc[filter_len] Definition:`Length of the filter`
+	RawFilter          []byte `field:"-"`                                                               // Internal field, not exposed to users
+	FilterInstructions string `field:"filter_instructions,handler:ResolveSetSockOptFilterInstructions"` // SECLDoc[filter_instructions] Definition:`Filter instructions`
+	FilterHash         string `field:"filter_hash,handler:ResolveSetSockOptFilterHash:"`                // SECLDoc[filter_hash] Definition:`Hash of the socket filter`
```
spikat: You should mention which hash algo is used here

pkg/security/serializers/serializers_linux.go
```diff
@@ -552,10 +552,22 @@ type SyscallArgsSerializer struct {
...
+	// Socket protocol
+	SocketProtocol uint16 `json:"socket_protocol"`
+	// Socket family
+	SocketFamily uint16 `json:"socket_family"`
 	// Level at which the option is defined
 	Level uint32 `json:"level"`
 	// Name of the option being set
 	OptName uint32 `json:"optname"`
+	// Length of the filter
+	FilterLen uint16 `json:"filter_len"`
```
spikat: This one should be `omitempty` as well

pkg/security/secl/model/consts_map_names_linux.go
```diff
@@ -84,6 +84,7 @@ var bpfMapNames = []string{
 	"security_profil",
 	"selinux_enforce",
 	"selinux_write_b",
+	"setsockopt_even",
```
spikat, line 87: +`t`? :smile:

pkg/security/probe/field_handlers_ebpfless.go
```diff
@@ -545,3 +549,43 @@ func (fh *EBPFLessFieldHandlers) ResolveConnectHostnames(_ *model.Event, e *mode
...
+	}
+
+	instructions, _ := bpf.Disassemble(raw)
+
+	for i, inst := range instructions {
+		e.FilterInstructions += fmt.Sprintf("%03d: %s\n", i, inst)
+	}
+
+	return e.FilterInstructions
+}
```
spikat: Unless you already added the setsockopt support in ebpfless I guess you should clean those handlers :)

pkg/security/ebpf/c/include/hooks/setsockopt.h
```diff
@@ -3,64 +3,150 @@
...
+    if (ret < 0) {
+        return 0;
+    }
+    unsigned short prog_len = prog.len;
+    syscall->setsockopt.filter_len = prog_len;
+    int key = 0;
+    struct setsockopt_event_t *event = bpf_map_lookup_elem(&setsockopt_event, &key);
+    if (event && 0 < prog_len && prog_len <= MAX_BPF_FILTER_LEN) { 
+        bpf_probe_read(&event->bpf_filters_buffer, (sizeof(struct sock_filter) * prog_len) &MAX_BPF_FILTER_LEN , prog.filter);  
+    }
```
spikat: Could we add a "truncated" boolean and retrieve the maximum length of the filter here?
Also, having a metric to track how often it happens could be useful

pkg/security/tests/setsockopt_test.go
```diff
@@ -23,8 +23,31 @@ func TestSetSockOpt(t *testing.T) {
...
+			&& setsockopt.socket_protocol == 6 
+			&& setsockopt.socket_family == AF_INET 
+			&& setsockopt.filter_hash == "627019f67a3853590209488302dd51282834c4f9f9c1cc43274f45c4bfd9869f"`,
+		},
+		{
+			ID: "test_rule_setsockopt_udp",
+			Expression: `setsockopt.level == SOL_SOCKET 
+			&& setsockopt.optname == SO_ATTACH_FILTER 
+			&& setsockopt.socket_type == SOCK_DGRAM 
+			&& setsockopt.socket_protocol == 17
```
spikat: Could we use some constants here instead of specify the protocol by its value?

pkg/security/ebpf/c/include/events_definition.h
theop-dd: Do you mean the types?

pkg/security/secl/model/consts_map_names_linux.go
theop-dd: oops :)

pkg/security/probe/field_handlers_ebpfless.go
theop-dd: Last I checked it was not building if I remove it.

pkg/security/ebpf/c/include/events_definition.h
spikat: yeah, here we have: 16+16+16 (not aligned in 32bits), then 32+32 (so, unaligned) then another 16. Ideally we want somethings like 16/16/16/16+32/32

pkg/security/probe/field_handlers_ebpfless.go
spikat: You have to provide them yep, but empty more as mock

pkg/security/secl/model/consts_map_names_linux.go
spikat: discussed offline, the file is generated and the content truncated

pkg/security/probe/field_handlers_ebpfless.go
theop-dd: okay yes I'll do that

pkg/security/ebpf/c/include/constants/offsets/network.h
theop-dd: This trick seems to have fix a lot of bugs but I'm not sure I have 0 fail, need to double check ;)

pkg/security/ebpf/c/include/events_definition.h
theop-dd: ok yes, I can swap filter_len to have 4 * 16 at the beginning

pkg/security/probe/constantfetch/constant_names.go
theop-dd: But if I add a fallback I think it will change how I handle the issue no?
Because at the moment, if I can't get the offset, it will be 0 and I compute it manually (in get_protocol_from_sock)

pkg/security/secl/model/model_unix.go
theop-dd: yes you're right, I have added it (it is sha256)

pkg/security/secl/model/consts_map_names_linux.go
theop-dd: After checking it's due to a generated file. We need to modify how the generation is done to avoid truncation (max = 15)

pkg/security/tests/setsockopt_test.go
theop-dd: Yes probably, I forgot to do it here you're right. Let me check

pkg/security/ebpf/c/include/hooks/setsockopt.h
theop-dd: I've added a metric and the boolean. In addition, I fixed how I was handling the size of the filter and the size of what was actually read.

pkg/security/tests/setsockopt_test.go
theop-dd: Done

pkg/security/probe/probe_ebpf.go
```diff
@@ -1479,6 +1487,10 @@ func (p *EBPFProbe) handleEvent(CPU int, data []byte) {
 			seclog.Errorf("failed to decode setsockopt event: %s (offset %d, len %d)", err, offset, len(data))
 			return
 		}
+		seclog.Debugf("setsockopt event truncated")
+		if p.BPFFilterTruncated != nil {
```
YoannGh:
```suggestion
		if event.SetSockOpt.IsFilterTruncated {
```

pkg/security/probe/field_handlers_ebpf.go
```diff
@@ -909,3 +911,44 @@ func (fh *EBPFFieldHandlers) ResolveAcceptHostnames(_ *model.Event, e *model.Acc
 
 	return e.Hostnames
 }
+
+// ResolveSetSockOptFilterHash resolves the filter hash of a setsockopt event
+func (fh *EBPFFieldHandlers) ResolveSetSockOptFilterHash(_ *model.Event, e *model.SetSockOptEvent) string {
```
YoannGh, line 916: It would be nice to check whether the event `FilterHash` field is already set here to avoid recomputing it every time `ResolveSetSockOptFilterHash` is called

pkg/security/probe/field_handlers_ebpf.go
```diff
@@ -909,3 +911,44 @@ func (fh *EBPFFieldHandlers) ResolveAcceptHostnames(_ *model.Event, e *model.Acc
...
+	h := sha256.New()
+	h.Write(e.RawFilter)
+	bs := h.Sum(nil)
+	e.FilterHash = fmt.Sprintf("%x", bs)
+	return e.FilterHash
+}
+
+// ResolveSetSockOptFilterInstructions resolves the filter instructions of a setsockopt event
+func (fh *EBPFFieldHandlers) ResolveSetSockOptFilterInstructions(_ *model.Event, e *model.SetSockOptEvent) string {
```
YoannGh, line 928: Same thing for the `FilterInstructions` field

pkg/security/serializers/serializers_base.go
```diff
@@ -136,6 +136,19 @@ type NetworkContextSerializer struct {
 	NetworkDirection string `json:"network_direction"`
 }
 
+// BPFFilterSerializer serializes a BPF filter to JSON
+// easyjson:json
+type BPFFilterSerializer struct {
```
YoannGh: This type doesn't seem to be used so you can probably remove it

pkg/security/ebpf/c/include/hooks/setsockopt.h
```diff
@@ -3,64 +3,152 @@
...
 }
 
 int __attribute__((always_inline)) sys_set_sock_opt_ret(void *ctx, int retval) {
     struct syscall_cache_t *syscall = pop_syscall(EVENT_SETSOCKOPT);
-
     if (!syscall) {
         return 0;
     }
+    int key = 0;
+    struct setsockopt_event_t *event = bpf_map_lookup_elem(&setsockopt_event,&key);
```
YoannGh, line 38: Is it possible to use `__builtin_memset` on the setsockopt event here in order to reset the event fields and make sure we're not using any data that was set for a previous event?

pkg/security/serializers/serializers_base.go
theop-dd: Yes you're right, I was keeping it as a backup for this PR but forgot to remove it at the end

pkg/security/ebpf/c/include/hooks/setsockopt.h
theop-dd: Due to the size of the map the builtin_memset seems to impossible to use. However, every field is supposed to be properly overwritten so it's not supposed to cause any issue

pkg/security/probe/field_handlers_ebpf.go
theop-dd: Yes you are right!

pkg/security/probe/field_handlers_ebpf.go
theop-dd: I agree

pkg/security/secl/model/unmarshallers_linux.go
```diff
@@ -1555,10 +1555,25 @@ func (e *SetSockOptEvent) UnmarshalBinary(data []byte) (int, error) {
 	if len(data) < 8 {
```
YoannGh:
```suggestion
	if len(data) < 24 {
```

YoannGh (approved):

pkg/security/serializers/serializers_linux.go
```diff
@@ -1309,9 +1324,68 @@ func newSecurityProfileContextSerializer(event *model.Event, e *model.SecurityPr
...
+			SocketType:         model.SocketType(e.SetSockOpt.SocketType).String(),
+			SocketProtocol:     model.SocketProtocol(e.SetSockOpt.SocketProtocol).String(),
+			SocketFamily:       model.SocketFamily(e.SetSockOpt.SocketFamily).String(),
+			Level:              model.SetSockOptLevel(e.SetSockOpt.Level).String(),
+			OptName:            fmt.Sprintf("%d", e.SetSockOpt.OptName),
+			IsFilterTruncated:  e.SetSockOpt.IsFilterTruncated,
+			FilterLen:          e.SetSockOpt.FilterLen,
+			FilterInstructions: e.FieldHandlers.ResolveSetSockOptFilterInstructions(e, &e.SetSockOpt),
+			FilterHash:         e.FieldHandlers.ResolveSetSockOptFilterHash(e, &e.SetSockOpt),
+		}
```
spikat, line 1388: for the love of god, please only set the OptName in the switch case, all of the other fields are set the same way

pkg/security/probe/field_handlers_ebpf.go
```diff
@@ -909,3 +911,50 @@ func (fh *EBPFFieldHandlers) ResolveAcceptHostnames(_ *model.Event, e *model.Acc
...
+
+			raw = append(raw, bpf.RawInstruction{
+				Op: Code,
+				Jt: Jt,
+				Jf: Jf,
+				K:  K,
+			})
+		}
+
+		instructions, _ := bpf.Disassemble(raw)
```
spikat, line 951: should we check the second returned value here?

pkg/security/probe/probe_ebpf.go
```diff
@@ -1479,6 +1487,10 @@ func (p *EBPFProbe) handleEvent(CPU int, data []byte) {
 			seclog.Errorf("failed to decode setsockopt event: %s (offset %d, len %d)", err, offset, len(data))
 			return
 		}
+		seclog.Debugf("setsockopt event truncated")
```
spikat, line 1490: not sure we want to keep this debug log

pkg/security/ebpf/c/include/constants/offsets/network.h
```diff
@@ -20,6 +20,17 @@ __attribute__((always_inline)) u16 get_skc_num_from_sock_common(struct sock_comm
     bpf_probe_read(&skc_num, sizeof(skc_num), (void *)sk + sock_common_skc_num_offset);
     return htons(skc_num);
 }
+__attribute__((always_inline)) u16 get_protocol_from_sock(struct sock *sk) {
+    u64 sock_sk_protocol_offset;
+    LOAD_CONSTANT("sock_sk_protocol_offset", sock_sk_protocol_offset);
```
spikat, line 25: you should first check the returned value here

pkg/security/ebpf/c/include/constants/offsets/network.h
```diff
@@ -20,6 +20,17 @@ __attribute__((always_inline)) u16 get_skc_num_from_sock_common(struct sock_comm
...
+    u16 protocol = 0;
+    if ((void *)sk + sock_sk_protocol_offset > 0 && sock_sk_protocol_offset + sizeof(protocol) < sizeof(struct sock)) {
```
spikat, line 27: not sure the first part of the condition (`(void *)sk + sock_sk_protocol_offset > 0`) makes sense.

spikat (commented):
Please address those comments in a new separate PR

pkg/security/probe/probe_ebpf.go
theop-dd: Okay I remove it

pkg/security/ebpf/c/include/constants/offsets/network.h
theop-dd: yes you're right. But on the other functions you were not doing it for the LOAD CONSTANT, that's why I kept the same syntax

pkg/security/probe/field_handlers_ebpf.go
theop-dd: Yes, it's a flag to check if everything was decoded, we can manage it as an error maybe

pkg/security/serializers/serializers_linux.go
theop-dd: Yes I agree

pkg/security/ebpf/c/include/constants/offsets/network.h
theop-dd: Yes it's probably useless. Locally it works without, let's see on other kernel versions

pkg/security/probe/field_handlers_ebpf.go
spikat: yeah, and avoid the parsing of the filter in this case ('cause we don't know where it stopped right?)

pkg/security/ebpf/c/include/constants/offsets/network.h
spikat: yeah you're right (only on few places we check the return of a load constant). not mandatory for this one, as you want :)

PR #38041, opened 2025-06-18
[CWS] fix container id from unknown container runtime

safchain:
### What does this PR do?

Use the user space container id/container manager userspace side when the manager is not systemd. Recent versions of containerd break the container runtime kernel side detection.

Add unit tests.

### DIFF
--- a/pkg/security/resolvers/process/resolver_ebpf.go
+++ b/pkg/security/resolvers/process/resolver_ebpf.go
@@ -1528,13 +1528,19 @@ func (p *EBPFResolver) UpdateProcessCGroupContext(pid uint32, cgroupContext *mod
 		return false
 	}
 
-	pce.Process.CGroup = *cgroupContext
-	pce.CGroup = *cgroupContext
-	if cgroupContext.CGroupFlags.IsContainer() {
-		containerID, _ := containerutils.FindContainerID(cgroupContext.CGroupID)
+	// Assume that the container runtime from the kernel side may be incorrect or missing
+	// In that case fallback to the userland container runtime.
+	if !cgroupContext.CGroupFlags.IsSystemd() && cgroupContext.CGroupID != "" {
+		containerID, cgroupFlags := containerutils.FindContainerID(cgroupContext.CGroupID)
+		cgroupContext.CGroupFlags = cgroupFlags
+
 		pce.ContainerID = containerID
 		pce.Process.ContainerID = containerID
 	}
+
+	pce.Process.CGroup = *cgroupContext
+	pce.CGroup = *cgroupContext
+
 	return true
 }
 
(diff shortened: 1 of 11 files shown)


pkg/security/secl/containerutils/cgroup.go
```diff
@@ -42,13 +44,16 @@ var RuntimeToken = []struct {
 	{"crio-", CGroupManagerCRIO},
 	{"libpod-", CGroupManagerPodman},
 	{"ecs/", CGroupManagerECS},
+
+	// default to containerd in case of kubepods
+	{"kubepods", CGroupManagerCRI},
```
lebauce, line 48: Is it necessary ? CGroups on Kubernetes using CRI-O seems to also have the "kubepods" prefix

pkg/security/secl/containerutils/cgroup.go
```diff
@@ -23,12 +23,14 @@ const (
...
+
+	CgroupManagerUndefined CGroupManager = 255
 )
 
 // CGroup flags
 const (
-	SystemdService CGroupFlags = (0 << 8)
-	SystemdScope   CGroupFlags = (1 << 8)
+	SystemdService CGroupFlags = iota + (1 << 8)
+	SystemdScope
```
YoannGh, line 32: We need to update these flags (as well as the `CGroupManagerMask`) on the eBPF side as well

pkg/security/secl/containerutils/cgroup.go
safchain: right but as this is a list the and crio- entry will have matched before. This entry is more a fallback

pkg/security/secl/containerutils/cgroup.go
safchain: done

pkg/security/secl/containerutils/cgroup.go
```diff
@@ -23,12 +23,14 @@ const (
 	CGroupManagerCRI                              // containerd
 	CGroupManagerSystemd                          // systemd
 	CGroupManagerECS                              // ecs
+
+	CgroupManagerUndefined CGroupManager = 255 // undefined
```
safchain: nit : use unknown with a 0 value

pkg/security/resolvers/process/resolver_ebpf.go
```diff
@@ -1528,13 +1528,19 @@ func (p *EBPFResolver) UpdateProcessCGroupContext(pid uint32, cgroupContext *mod
...
 	}
 
-	pce.Process.CGroup = *cgroupContext
-	pce.CGroup = *cgroupContext
-	if cgroupContext.CGroupFlags.IsContainer() {
-		containerID, _ := containerutils.FindContainerID(cgroupContext.CGroupID)
+	// Assume that the container runtime from the kernel side may be incorrect or missing
+	// In that case fallback to the userland container runtime.
+	if !cgroupContext.CGroupFlags.IsSystemd() && cgroupContext.CGroupID != "" {
+		containerID, cgroupFlags := containerutils.FindContainerID(cgroupContext.CGroupID)
```
safchain: nit : from bob, remove a few lines

YoannGh (approved):

PR #38047, opened 2025-06-18
[APM] Fix sensitive information info scrubbing

ichinaski:
### What does this PR do?

Replace the current `ScrubBytes` approach -- which is prone to break JSON syntax -- with a YAML scrubber.

Changes:

• Replace direct JSON byte scrubbing with structure-aware YAML round-trip processing for better config data integrity
• Add `yaml:"-"` tags to function pointer fields in AgentConfig to prevent YAML serialization errors
• Update unit tests with complex scrubbing patterns and YAML marshaling behavior expectations

### Motivation

Current `ScrubBytes` approach can sometimes break the JSON-structure when scrubbing complex or nested patterns.


ajgajg1134 (approved):
Looks great!

PR #38052, opened 2025-06-18
[NDINT-184] Consistent NetFlow flush interval

nathalieabello:
### What does this PR do?
- Ensures flows are always flushed after the full 5-minute interval, including first-time flows.
- Removes asynchronous timing dependencies by passing a controlled now from the aggregator to the accumulator. This makes nextFlush computation deterministic and tests reliable.

### Motivation
- Current behavior inconsistently flushes flows: sometimes after <10s, sometimes after 5 minutes.
- Immediate flushing makes reverse DNS enrichment unreliable for newly seen flows.

- Updated accumulator tests to align with new flush timing.
- Verified all relevant integration tests pass.

- Customers need to wait 5 minutes to start seeing NetFlow data after enabling it. 
- flowAccumulator now depends on time passed in by the aggregator. Incorrect usage (i.e. passing a bad now) isn't validated internally.

### Additional Notes

- An alternate approach that was considered was injecting a timeNow function into flowAccumulator to enable testability without changing method signatures. Chose current approach for its simplicity and better encapsulation.


comp/netflow/flowaggregator/flowaccumulator.go
```diff
@@ -9,14 +9,17 @@ import (
...
 	log "github.com/DataDog/datadog-agent/comp/core/log/def"
 	"github.com/DataDog/datadog-agent/comp/netflow/common"
 	"github.com/DataDog/datadog-agent/comp/netflow/portrollup"
 	rdnsquerier "github.com/DataDog/datadog-agent/comp/rdnsquerier/def"
-	"go.uber.org/atomic"
 )
 
-var timeNow = time.Now
+const (
+	tickerDuration = 10 * time.Millisecond
```
nathalieabello, line 21: Is this timeDuration okay? Should I make it shorter / longer?

PR #38056, opened 2025-06-18
[AGENTRUN-337] Add e2e tests for networkv2 check

jeremy-hanna:
### What does this PR do?

This PR adds e2e tests for the default configuration of the networkv2 check, comparing it against the Python implementation to ensure all metrics and tags are correct. Subsequent PRs around configuration variants to follow.


pkg/collector/corechecks/net/networkv2/network.go
```diff
@@ -145,8 +145,15 @@ func (c *NetworkCheck) Run() error {
 				c.submitProtocolMetrics(sender, counters[protocol])
 			}
 		}
+		if tcpStats, ok := counters["Tcp"]; ok {
+			if val, ok := tcpStats.Stats["CurrEstab"]; ok {
+				sender.Gauge("system.net.tcp.current_established", float64(val), "", nil)
```
jeremy-hanna, line 150: we only submitted metrics before and [missed the gauge](https://github.com/DataDog/integrations-core/blob/master/network/datadog_checks/network/check_linux.py#L267) which is just this single entry

pkg/collector/corechecks/net/networkv2/network.go
```diff
@@ -145,8 +145,15 @@ func (c *NetworkCheck) Run() error {
...
+			}
+		}
 	}
 
+	submitInterfaceSysMetrics(sender)
```
jeremy-hanna, line 155: [this](https://github.com/DataDog/integrations-core/blob/master/network/datadog_checks/network/check_linux.py#L330-L348) was not implemented earlier and we missed some default metrics

pkg/collector/corechecks/net/networkv2/network.go
```diff
@@ -662,45 +714,50 @@ func runCommand(cmd []string, env []string) (string, error) {
 }
 
 func collectConntrackMetrics(sender sender.Sender, conntrackPath string, useSudo bool, procfsPath string, blacklistConntrackMetrics []string, whitelistConntrackMetrics []string) {
-	if conntrackPath != "None" {
```
jeremy-hanna, line 665: the guard clause should only be fore the [ConntrackStats metrics](https://github.com/DataDog/integrations-core/blob/master/network/datadog_checks/network/check_linux.py#L273-L274) and not subsequent files (might be a bug, but this is a bug for bug migration of behavior)

pkg/collector/corechecks/net/networkv2/network.go
```diff
@@ -662,45 +714,50 @@ func runCommand(cmd []string, env []string) (string, error) {
...
-					continue
-				}
+	// By default, only max and count are reported. However if the blacklist is set,
+	// the whitelist is losing its default value
+	for _, metricName := range availableFiles {
+		if len(blacklistConntrackMetrics) > 0 {
+			if slices.ContainsFunc(blacklistConntrackMetrics, func(s string) bool {
+				return strings.Contains(metricName, s)
+			}) {
+				continue
```
jeremy-hanna, line 740: we are not doing a string comparison but a substr search [for blocklists/allowlists](https://github.com/DataDog/integrations-core/blob/master/network/datadog_checks/network/check_linux.py#L302) in the python check

pgimalac: In the Python code we use `pattern_filter` function which considers the arguments as regexes, it's not a substring search ?

pgimalac: Also `pattern_filter` checks for whitelist first, and fallbacks to blacklist otherwise

jeremy-hanna: hmm you're right, is it alright if I follow this up with a new PR to get these passing e2e tests in?

pgimalac (approved):

PR #38065, opened 2025-06-19
Add started and running metrics to COAT

louis-cqrl:
### What does this PR do?

This PR aims to add `runtime.started` and `runtime.running` metrics to COAT.

- `datadog.agent.started` is defined in [pkg/aggregator/demultiplexer_agent.go](https://github.com/DataDog/datadog-agent/blob/ac84213ad42c883075b240f30f64ca7c56af2211/pkg/aggregator/demultiplexer_agent.go#L240)
- `datadog.agent.running` is defined in [pkg/aggregator/aggregator.go](https://github.com/DataDog/datadog-agent/blob/359ca2397956d648e63795b358e6e2ee5eae639e/pkg/aggregator/aggregator.go#L609)

Since these metrics are not Prometheus compliant, we need to declare 2 new counters for each metric in order to sent them to COAT

### Motivation

Create dashboards and monitors to evaluate the impact of an Agent update to customer environment.

Metrics present in both `sudo datadog-agent diagnose show-metadata agent-telemetry` and on COAT org on Datadog app

### Additional Notes

By default, COAT emits every registered metric on a 15-minute cadence until the host shuts down. This change customizes that behavior for two special metrics:

- started:
Emitted once immediately after the agent successfully starts (override the default 15-minute schedule) to avoid sending redundant startup payloads.

- running:
Emitted once on successful startup to signal liveness, then resumes the default 15-minute schedule thereafter to continuously report host liveliness.

These tweaks ensure that startup notifications are sent exactly once, while ongoing health data still flows every quarter hour.


pkg/aggregator/aggregator.go
```diff
@@ -168,6 +168,10 @@ var (
 	tlmChecksContextsBytesByMtype = telemetry.NewGauge("aggregator", "checks_contexts_bytes_by_mtype",
 		[]string{"shard", "metric_type", tags.BytesKindTelemetryKey}, "Estimated count of bytes taken by contexts in the check aggregator, by metric type")
 
+	agentRunning = telemetry.NewGauge("runtime", "running",
+		[]string{}, "Indicates if the agent is running",
+	)
+
```
vickenty: The values never change and do not explain anything about the aggregator itself. Can we please implement these new metrics directly in the agenttelemetry component, as to not to pollute aggregator with this unrelated function?

(Yes, aggregator reports similar metrics to the user org, because it was a natural place to inject metrics without a clear user configured source. No, this isn't a good reason to place new metrics here as well.)

vickenty (dismissed):

hestonhoffman (approved):
Docs 👍

pkg/aggregator/aggregator.go
louis-cqrl: > The values never change

The running value needs to stay at `1`; the idea is to confirm the agent is still running, so it does not need to change its value.

> and do not explain anything about the aggregator itself. Can we please implement these new metrics directly in the agenttelemetry component, as to not to pollute aggregator with this unrelated function?

I do not have the feeling that the agenttelemetry component is a better place to hold these metrics. The goal of these two metrics is to behave exactly the same as the `datadog.agent.started` and `datadog.agent.running` metrics since we cannot export them as is. Mimicking the same pipeline in another component might add some overhead and code complexity that is not needed.

Maybe @iglendd may have better input on this.

vickenty: My logic was agenttelemetry is the component responsible for COAT in the agent, so I think it is reasonable to put COAT-specific things there. But I don't have a strong opinion, we could do it at `cmd` level as well where we init the components, since these metrics are about the entire agent. But I do strongly believe that topical components, like aggregator, should only report telemetry about the component itself, and the entire agent.

iglendd: I do not think that COAT should be emitting this metric either. No can it. I looks like one use case it to detect then Agent started but never reach running state When and where we know for sure that the Agent is running? I do not know, it is a good question as well if it aggregators responsibility. But I think that @louis-cqrl is only a messenger here because that decision had been made a while back in the form of updating a different metric with the same semantic intent (there is a limitation why COAT cannot reuse existing metric otherwise it would not be necessary).

rahulkaukuntla (approved):

cmd/agent/subcommands/run/command.go
```diff
@@ -580,6 +596,9 @@ func startAgent(
 
 	http.Handle("/telemetry", telemetryHandler)
 
+	agentStarted.Inc()
+	defer agentRunning.Set(1)
```
jose-manuel-almaza: Do we set this even if we return an error?

louis-cqrl: Having an error is not an issue I think, it won't send a metric which is already a data to not receive the metric, running is defer so running will be sent only if everything work and started needs to be sent in any case

cmd/agent/subcommands/run/command.go
```diff
@@ -183,6 +183,22 @@ import (
 	ddruntime "github.com/DataDog/datadog-agent/pkg/runtime"
 )
 
+var (
+	agentStarted = pkgTelemetry.NewCounter(
```
GustavoCaso: Can we not use the global `pkgTelemetry` instead of the telemetry component? It would require defining the metrics in the `run` function. Since this metrcis are only used in that function I believe there is no issue on doing that + we do not relay in the global `pkgTelemetry`

cmd/agent/subcommands/run/command.go
GustavoCaso: Is the running metric something we need to keep sending on an interval to validate that the agent is running? It seems we sent this metric when `startAgent` exists. I believe we need to keep sending the metric similar to how the aggregator does.

GustavoCaso (changes requested):
Left a few suggestion that I believe need to be addressed before merging

releasenotes/notes/add-startup-running-COAT-e2342c64da7061c2.yaml
```diff
@@ -0,0 +1,12 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+other:
+  - |
+    Add a new metric counter to the Agent telemetry for the startup and running states.
```
GustavoCaso:
```suggestion
    Add a new metric to the Agent telemetry for the startup and running states.
```
One is a counter and the other is a gauge. If you don't want to specify both types in the changelog which I don't think is important, we can remove the metric type altogether

cmd/agent/subcommands/run/command.go
```diff
@@ -353,6 +353,27 @@ func run(log log.Component,
...
+		"Establish if the agent is running",
+	)
+
+	// agentStarted is a counter that increments to 1 when the agent starts successfully.
+	// This metric is only sent once at startup to indicate successful initialization.
+	agentStarted.Inc()
+	// agentRunning is a gauge set to 1 when the agent is running.
+	// This metric is sent periodically to indicate that the agent is still alive and operational.
+	// It is registered in the Prometheus registry for  the COAT monitoring frameworks.
+	agentRunning.Set(1)
```
GustavoCaso:
```suggestion
	// agentStarted and agentRunning are metrics used for Cross-org Agent Telemetry (COAT)
	// for more details on the scheduling configuration, check comp/core/agenttelemetry/impl/config.go
	agentStarted.Inc()
	agentRunning.Set(1)
```

louis-cqrl:
```suggestion
	// It is registered in the Prometheus registry for the COAT monitoring frameworks.
	agentRunning.Set(1)
```

GustavoCaso (approved):
Fantastic work 🎉 
Left two suggestions regarding some wording. Feel free to disregard them if you disagree with them. 

🚀

jose-manuel-almaza (approved):

PR #37733, opened 2025-06-06
Adding a new CI job to run unit-test with the new unmarshal implementation

hush-hush:
### What does this PR do?

Adding a new CI job to run unit-test with the new unmarshal implementation

### Describe how you validated your changes

Running the CI is enough.

PR #38070, opened 2025-06-19
Remove tests_nodetreemodel from MQ

KevinFairise2:
### What does this PR do?

Remove tests_nodetreemodels from the merge queue. These jobs are allowed to fail and should not increase duration of the MQ


chouetz (approved):

chouetz:
@KevinFairise2 we need to update https://github.com/DataDog/datadog-agent/pull/37733 as well

PR #33075, opened 2025-01-17
[ASCII-2671] Add check loader name tag on `checks.execution_time` metric

pgimalac:
### What does this PR do?
Add the check loader name as a tag on the `checks.execution_time` metrics.

### Motivation
In theory each check should be either on the Go loader or the Python loader.
However some python checks are being rewritten in Go, and with custom checks customers could use the same name as a Go check.
Knowing the loader used to run checks would help a lot to know if we risk breaking users with some checks-related changes. 

### Describe how you validated your changes

Ran the agent locally, with the following configs:
```
telemetry.enabled: true
telemetry.checks: "*"
```
Generated a flare, and checked that `telemetry.log` contained the metric with the tag, with the correct loader name, eg.
```
checks__execution_time{check_loader="core",check_name="uptime"} 0
checks__execution_time{check_loader="python",check_name="network"} 46
```

Checked that the tag appeared in instrumentation telemetry org.

Also updated `pkg/collector/check/stats/stats_test.go` to have a unit test ensuring the correct check loader name was used.

PR #38071, opened 2025-06-19
[AGENTRUN-356] Make Go check loader have priority over Python

pgimalac:
### What does this PR do?
Make the Go check loader have priority over the Python check loader.

### Motivation
This will eventually allow not loading Python by default. For now, even loading a Go check requires Python, because we need to check if there is a Python version for that check first.

We know that 5 checks have both a Go and a Python version:
- network
- disk
- kubelet
- snmp
- win32_event_log

Ensure that the check(s) mentioned above that you own still behave in the same way, in particular the same version (Go/Python) runs, for each flavor / platform where it makes sense.

I tested the `network` and `disk` checks for Agent Runtimes, with the default flavor of the agent, with the iot agent, and on k8s. In each case we were still running the same expected version (Go for IoT, Python otherwise).

### Additional Notes

The SNMP check had a special handling to try loading with Go first then Python, so the behavior is the same.
Each of the other checks has a configuration to pick whether to run the Go or the Python version, so there should be no impact on that side (it doesn't depend on loader priority).

We also checked whether users were using custom Python checks with the same name as a Go check, using a telemetry added in Agent 7.64.0 ([PR](https://github.com/DataDog/datadog-agent/pull/33075)), and didn't find any such case.


julien-lebot (approved):

releasenotes/notes/go-loader-priority-21f12dce33664351.yaml
```diff
@@ -0,0 +1,13 @@
...
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+enhancements:
+  - |
+    If a check has both a Go and a Python version, the Go version now has priority by default.
+    It should not have any visible impact for users, but if needed it can be
+    disabled by setting `prioritize_go_check_loader` configuration to `false`.
```
maycmlee:
```suggestion
    This change should not have any visible impact, but if needed, you can
    disable this configuration by setting `prioritize_go_check_loader` to `false`.
```

maycmlee (commented):
Left some suggestions

Enzu83:
Is there a specific reason for the Python check loader to be prioritized by the scheduler?
Why do we need to check if a Python version exists for every check?

pgimalac:
@Enzu83 here is [an RFC](https://datadoghq.atlassian.net/wiki/spaces/ARUN/pages/4644635779/RFC+Avoid+requiring+Python+to+load+Go+checks) explaining why we're doing for this change.

> Is there a specific reason for the Python check loader to be prioritized by the scheduler?

I believe initially agent 6 only had Python checks (inherited from Agent 5), so there was only a Python loader.
Then we added an IoT agent, which doesn't ship with Python, so we needed a basic Go version of some checks to provide system metrics. 
If Python is available we should use it as it contains a better version of the checks, otherwise we fallback to the Go version.
Also doing it the other way could have broken users: if a user has a custom Python check named the same way as one of the Go checks, then the Go version would have shadowed it (note that I was able to check that there is no such case).

> Why do we need to check if a Python version exists for every check?

That's just defined by the loader priority, as long as Python checks have priority over Go checks (which is the case until this PR) we need to try loading with Python then with Go. 
For Go the list of checks is known and straightforward, but for Python we can't know easily if a check has a Python implementation. It requires importing the Python module with the same name as the check (eg. `snmp` or `datadog_checks.snmp`), and searching that module to find a class which extends the base check class (+ a few other minor conditions).
I tried some alternative way to know if a check has a Python version without needing to load Python (eg. https://github.com/DataDog/datadog-agent/pull/32922), but it's a difficult problem to do correctly (lots of caveats).

Anyway with this PR, we make Go checks have priority over Python, which means that if a check has a Go implementation we don't need to try loading it with Python, we can just stop there.
This will enable not loading Python at all by default, once all the default Python checks are migrated to Go.

pkg/config/setup/config.go
```diff
@@ -318,7 +318,7 @@ func InitConfig(config pkgconfigmodel.Setup) {
 	config.BindEnvAndSetDefault("network_check.use_core_loader", false)
 
 	// If true, then the go loader will be prioritized over the python loader.
-	config.BindEnvAndSetDefault("prioritize_go_check_loader", false)
+	config.BindEnvAndSetDefault("prioritize_go_check_loader", true)
```
jeremy-hanna, line 321: won't this set the priority to use the IoT go disk / network checks over the python ones? I'm not sure we can do this yet as those lack a lot of features we're in the process of porting over with the v2 checks

pgimalac: No because those two checks have a configuration to determine whether to run the Go or Python version, and it default to Python for now, see https://github.com/DataDog/datadog-agent/pull/36526 for details.

That's what I'm referring to here:
> The SNMP check had a special handling to try loading with Go first then Python, so the behavior is the same.
Each of the other checks has a configuration to pick whether to run the Go or the Python version, so there should be no impact on that side (it doesn't depend on loader priority).

Also for the record I tested this change locally and on internal k8s deployments and we're still running the Python version for both checks.

Enzu83:
@pgimalac Thanks for the links and the context!
Everything is clear for me now.

pkg/config/setup/config.go
jeremy-hanna: ah gotcha, do we need [this additional configuration](https://github.com/DataDog/datadog-agent/pull/36526/files#diff-a553cd581968cebebcde0f68ac7e3f4faea0bf7be8a78f3f398d835a7e96b2b8R158-R161) in the networkv2 check then?

jeremy-hanna (approved):

pkg/config/setup/config.go
pgimalac: Yep I think so

maycmlee (approved):

sblumenthal (approved):

dustmop (approved):

TCheruy (approved):

PR #38072, opened 2025-06-19
[EUDM] Add Agent module to collect software inventory

julien-lebot:
### What does this PR do?
This PR adds a new Windows software inventory collection module to the Datadog Agent.
This new component collects installed software information on Windows systems through System-Probe.

The status page is updated with a new section for Software Inventory:
image

The status command is also updated to report the status of the Software Inventory:
```
===========================
Software Inventory Metadata
===========================

Detected 549 installed software entries.
The full list is not displayed in this status output, but is available in a flare, the Agent status page in the GUI, or in the Datadog Application.
```

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1544

- **Unit tests**: Comprehensive test coverage for the component, collectors, and integration points
- **Integration tests**: Windows-specific integration tests for software inventory collection
- **GUI integration**: Verified the status templates render correctly in the Agent GUI

#### Operating System support
This is the initial implementation focused on Windows systems. Future iterations may extend to other operating systems.

#### Configuration flag
The System-Probe flag that was controlling the feature moved to the Agent.

#### SCSS/CSS
The CSS file was generated using the SCSS file, but the generated file gets checked-in anyway because there is no tooling to re-generate it during build.


comp/metadata/inventorysoftware/impl/inventorysoftware.go
```diff
@@ -0,0 +1,194 @@
...
+	"github.com/DataDog/datadog-agent/pkg/system-probe/config/types"
+	httputils "github.com/DataDog/datadog-agent/pkg/util/http"
+)
+
+const flareFileName = "inventorysoftware.json"
+
+// SysProbeClient is an interface for the sysprobeclient used for dependency injection and testing.
+// This interface abstracts the communication with the System Probe to retrieve software inventory data,
+// allowing for easier testing and dependency injection.
+type SysProbeClient interface {
```
alexn-dd: IIRC, Bryce has created a system probe client wrapper shared across various components.

julien-lebot: After searching through the codebase, I couldn't find another shared wrapper for `sysprobeclient.CheckClient` that would fit our needs here. If you're aware of a specific implementation by Bryce that I missed, could you point me to that file or component?

cmd/agent/subcommands/run/command_windows.go
```diff
@@ -254,6 +258,7 @@ func getPlatformModules() fx.Option {
 		fx.Invoke(func(_ etwtracer.Component) {}),
 		fx.Invoke(func(_ windowseventlog.Component) {}),
 		fx.Invoke(func(_ winregistry.Component) {}),
+		fx.Invoke(func(_ inventorysoftware.Component) {}),
```
GustavoCaso: You could add `fx.Invoke(func(_ inventorysoftware.Component) {}),` as part of the `inventorysoftwarefx.Module()`, that way you don't have call it here. That way, ensuring the module is included and loaded is done in a single place. 

You can look at the `configsync` component https://github.com/DataDog/datadog-agent/blob/main/comp/core/configsync/configsyncimpl/module.go#L48

comp/metadata/inventorysoftware/impl/inventorysoftware.go
```diff
@@ -0,0 +1,194 @@
...
+	Endpoint api.AgentEndpointProvider
+}
+
+// New creates a new inventory software component with the default sysprobeclient
+func New(reqs Requires) (Provides, error) {
+	return NewWithClient(reqs, nil)
+}
+
+// NewWithClient creates a new inventory software component with a custom sysprobeclient
+func NewWithClient(reqs Requires, client SysProbeClient) (Provides, error) {
```
GustavoCaso: To reduce package exposure, could we make this function private? Looking at the changes, we are only using them in the tests 😄

comp/metadata/inventorysoftware/impl/inventorysoftware.go
```diff
@@ -0,0 +1,194 @@
...
+	if client == nil {
+		client = &sysProbeClientWrapper{
+			client: sysprobeclient.GetCheckClient(pkgconfigsetup.SystemProbe().GetString("system_probe_config.sysprobe_socket")),
```
GustavoCaso: Could we use the systemprobe config component instead of the global accessor: `pkgconfigsetup.SystemProbe()`?

comp/metadata/inventorysoftware/def/component.go
hush-hush, line 1: Could we remove the `inventory` part ? It was the old name for REDAPL, we stopped using it in the newest payload.

comp/metadata/inventorysoftware/def/component.go
```diff
@@ -0,0 +1,24 @@
...
+// Component is the interface for the inventory software component.
+// This component is responsible for collecting software inventory data from the host system
+// and providing it in a structured format for reporting and monitoring purposes.
+// The component supports both automatic periodic collection and manual refresh triggers.
+type Component interface {
+	// Refresh triggers a new payload to be sent while still respecting the minimal interval between two updates.
+	// This method allows manual triggering of software inventory collection, useful for immediate
+	// inventory updates or testing purposes. The component will respect any configured minimum
+	// intervals to prevent excessive data collection.
+	Refresh()
```
hush-hush: Is this used anywhere ? The `Refresh` method form other metadata payload is needed because metadata might be changed by other part of the code (for example when a new check is being run we refresh the metadata payload related to them).

comp/metadata/inventorysoftware/impl/inventorysoftware.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+	}
+	hname, _ := reqs.Hostname.Get(context.Background())
```
GustavoCaso: Do you think we should check if there was an error?

comp/metadata/inventorysoftware/impl/inventorysoftware.go
hush-hush: Does this needs to be public ?

comp/metadata/inventorysoftware/impl/inventorysoftware.go
hush-hush: Why do we have 2 function for this ?

comp/metadata/inventorysoftware/impl/inventorysoftware.go
```diff
@@ -0,0 +1,194 @@
...
+	return NewWithClient(reqs, nil)
+}
+
+// NewWithClient creates a new inventory software component with a custom sysprobeclient
+func NewWithClient(reqs Requires, client SysProbeClient) (Provides, error) {
+	if client == nil {
+		client = &sysProbeClientWrapper{
+			client: sysprobeclient.GetCheckClient(pkgconfigsetup.SystemProbe().GetString("system_probe_config.sysprobe_socket")),
+		}
+	}
```
hush-hush: It seems that this will not be nil only for tests. If so, could we move this within tests ? Maybe having a `createClient` method that is mocked by the tests ?

comp/metadata/inventorysoftware/impl/inventorysoftware.go
```diff
@@ -0,0 +1,194 @@
...
+
+	is.cachedInventory = installedSoftware
+
+	return nil
+}
+
+// getPayload creates and returns a new software inventory payload.
+// This method triggers a refresh of the cached data and returns a properly
+// formatted payload for transmission to the backend.
+func (is *inventorySoftware) getPayload() marshaler.JSONMarshaler {
```
GustavoCaso: Is `getPayload` supposed to return something even when the `software_inventory` is disabled? Currently, it returns an empty software list + hostname. Would it make more sense that we return nothing?

comp/metadata/inventorysoftware/impl/inventorysoftware.go
```diff
@@ -0,0 +1,194 @@
...
+		client = &sysProbeClientWrapper{
+			client: sysprobeclient.GetCheckClient(pkgconfigsetup.SystemProbe().GetString("system_probe_config.sysprobe_socket")),
+		}
+	}
+	hname, _ := reqs.Hostname.Get(context.Background())
+	is := &inventorySoftware{
+		log:            reqs.Log,
+		sysProbeClient: client,
+		hostname:       hname,
+		enabled:        reqs.Config.GetBool("software_inventory.enabled"),
```
hush-hush: Same here, let's not use `inventory` which is not a known concept for users.

WDYT:
```suggestion
		enabled:        reqs.Config.GetBool("software_monitoring.enabled"),
```

comp/metadata/inventorysoftware/impl/inventorysoftware.go
```diff
@@ -0,0 +1,194 @@
...
+
+// refreshCachedValues updates the cached software inventory data by collecting
+// fresh data from the System Probe. This method respects the enabled flag
+// and will skip collection if the feature is disabled in the configuration.
+func (is *inventorySoftware) refreshCachedValues() error {
+	if !is.enabled {
+		is.log.Debugf("Software inventory is disabled in agent configuration")
+		return nil
+	}
+	is.log.Infof("Collecting Software Inventory")
```
hush-hush: This will be printed every 10min by default. We should move this to `Debug`

comp/metadata/inventorysoftware/impl/inventorysoftware.go
```diff
@@ -0,0 +1,194 @@
...
+
+	installedSoftware, err := is.sysProbeClient.GetCheck(sysconfig.InventorySoftwareModule)
+	if err != nil {
+		return is.log.Errorf("error getting software inventory: %v", err)
+	}
+
+	is.cachedInventory = installedSoftware
```
hush-hush: Why cache `installedSoftware` if it's pulled again every time we call `getPayload` ? Could we simply return it to the caller ?

comp/metadata/inventorysoftware/impl/inventorysoftware.go
```diff
@@ -0,0 +1,194 @@
...
+		sysProbeClient: client,
+		hostname:       hname,
+		enabled:        reqs.Config.GetBool("software_inventory.enabled"),
+	}
+	// Always provide the component because FX dependency injection expects
+	// status providers, metadata providers, etc. to be available, not wrapped
+	// in option.Option. The enabled flag only controls whether we call the System
+	// Probe software inventory endpoint to collect actual data.
+	// Note that there is a second way to disable this feature, through InventoryPayload.Enabled.
+	// 'enable_metadata_collection' and 'inventories_enabled' both need to be set to true.
```
GustavoCaso: If the software inventory is not enabled, we can return `nil` flare providers, status provider, etc. The components that process those filter out nil providers. If you are curious, you can search for `fxutil.GetAndFilterGroup`

That way if is disable it won't show in the status output, flare, etc 😄

comp/metadata/inventorysoftware/impl/inventorysoftware_test.go
```diff
@@ -0,0 +1,171 @@
...
+	sp := &mock.SysProbeClient{}
+
+	// Create dependencies manually for the test
+	logComp := logmock.New(t)
+	hostnameComp := &mock.Hostname{}
+
+	// Get config using fxutil.Test
+	configComp := fxutil.Test[config.Component](t,
+		config.MockModule(),
+		fx.Replace(config.MockParams{Overrides: confOverrides}),
```
hush-hush:
```suggestion
	configComp := config.NewMock(t)
	configComp.SetWithoutSource("software_inventory.enabled", true)
```

comp/metadata/inventorysoftware/impl/inventorysoftware_test.go
```diff
@@ -0,0 +1,171 @@
...
+	"github.com/DataDog/datadog-agent/comp/core/config"
+	"github.com/DataDog/datadog-agent/comp/core/flare/helpers"
+	logmock "github.com/DataDog/datadog-agent/comp/core/log/mock"
+	"github.com/DataDog/datadog-agent/pkg/inventory/software"
+	"github.com/DataDog/datadog-agent/pkg/serializer"
+	serializermock "github.com/DataDog/datadog-agent/pkg/serializer/mocks"
+	sysconfig "github.com/DataDog/datadog-agent/pkg/system-probe/config"
+	"github.com/DataDog/datadog-agent/pkg/util/fxutil"
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/fx"
```
hush-hush: Implementation folder should not contain any reference to FX.

cmd/system-probe/modules/software_inventory_windows.go
```diff
@@ -36,14 +36,14 @@ type softwareInventoryModule struct {
 func (sim *softwareInventoryModule) Register(httpMux *module.Router) error {
 	httpMux.HandleFunc("/check", utils.WithConcurrencyLimit(1, func(w http.ResponseWriter, _ *http.Request) {
 		log.Infof("Got check request in software inventory")
-		inventory, warn, err := winsoftware.GetSoftwareInventory()
+		inventory, warn, err := software.GetSoftwareInventory()
```
GustavoCaso:
```suggestion
		inventory, warnings, err := software.GetSoftwareInventory()
```

comp/metadata/inventorysoftware/impl/inventorysoftware_test.go
```diff
@@ -0,0 +1,171 @@
...
+		fx.Provide(func() serializer.MetricSerializer { return serializermock.NewMetricSerializer(t) }),
+	)
+
+	// Create the Requires struct manually
+	reqs := Requires{
+		Log:        logComp,
+		Config:     configComp,
+		Serializer: serializerComp,
+		Hostname:   hostnameComp,
+	}
```
hush-hush:
```suggestion
	// Create the Requires struct manually
	reqs := Requires{
		Log:        logComp,
		Config:     configComp,
		Serializer: serializermock.NewMetricSerializer(t),
		Hostname:   hostnameComp,
	}
```

cmd/system-probe/modules/software_inventory_windows.go
```diff
@@ -36,14 +36,14 @@ type softwareInventoryModule struct {
...
 		if err != nil {
 			log.Errorf("Error getting software inventory: %v", err)
 			w.WriteHeader(500)
 			return
 		}
-		if warn != nil {
-			log.Warnf("warning: %v", warn)
+		for _, warning := range warn {
```
GustavoCaso:
```suggestion
		for _, warning := range warnings {
```

comp/metadata/inventorysoftware/impl/inventorysoftware_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+	// Call the constructor directly with the mock client
+	provides, err := NewWithClient(reqs, sp)
+	assert.NoError(t, err)
```
hush-hush: Should we use `require` here ? Is there any reason to continue the test if we can't create a new client ?

comp/metadata/inventorysoftware/impl/inventorysoftware.go
julien-lebot: Oh thank you that's helpful, I'll change it 🙇

comp/metadata/inventorysoftware/impl/inventorysoftware.go
julien-lebot: Yes of course!

comp/metadata/inventorysoftware/impl/inventorysoftware.go
GustavoCaso: I believe the code could look something like: 

```go

if !is.enabled {
	return Provides{
		Comp:                 is,
	}, nil
}

is.InventoryPayload = util.CreateInventoryPayload(reqs.Config, reqs.Log, reqs.Serializer, is.getPayload, flareFileName)
is.log.Infof("Starting the inventory software component")
return Provides{
	Comp:                 is,
	Provider:             is.InventoryPayload.MetadataProvider(),
	FlareProvider:        is.FlareProvider(),
	StatusHeaderProvider: status.NewHeaderInformationProvider(is),
	Endpoint:             api.NewAgentEndpointProvider(is.writePayloadAsJSON, "/metadata/software", "GET"),
}, nil
```

cmd/agent/subcommands/run/command_windows.go
julien-lebot: Thanks I can start using that, especially since that component is cross-platform.

comp/metadata/inventorysoftware/impl/inventorysoftware_test.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func newInventorySoftware(t *testing.T, confOverrides map[string]any) (*inventorySoftware, *mock.SysProbeClient) {
+	// Set default to enabled for tests
+	if confOverrides == nil {
+		confOverrides = map[string]any{}
+	}
+	if _, exists := confOverrides["software_inventory.enabled"]; !exists {
+		confOverrides["software_inventory.enabled"] = true
+	}
```
hush-hush: Seems like `confOverrides` is always nil. We can remove this and simply set `software_inventory.enabled`  to true all the time in `getProvides`.

comp/metadata/inventorysoftware/impl/inventorysoftware.go
```diff
@@ -0,0 +1,194 @@
...
+		StatusHeaderProvider: status.NewHeaderInformationProvider(is),
+		Endpoint:             api.NewAgentEndpointProvider(is.writePayloadAsJSON, "/metadata/software", "GET"),
+	}, nil
+}
+
+// refreshCachedValues updates the cached software inventory data by collecting
+// fresh data from the System Probe. This method respects the enabled flag
+// and will skip collection if the feature is disabled in the configuration.
+func (is *inventorySoftware) refreshCachedValues() error {
+	if !is.enabled {
```
GustavoCaso: If we apply this suggestion https://github.com/DataDog/datadog-agent/pull/38072#discussion_r2225919690 we can remove this check of `is.enabled` 😄

comp/metadata/inventorysoftware/mock/mock.go
```diff
@@ -0,0 +1,64 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+//go:build test
+
+// Package mock contains mock used to test the inventorysoftware component
+package mock
```
hush-hush: The mock package of a component should contain a mock of itself it offers to the rest of the code base not mock use in its tests internal tests. Most metadata payload don't offer a mock since they don't have any public method in their component interface.

pkg/config/setup/config.go
```diff
@@ -1199,6 +1199,10 @@ func agent(config pkgconfigmodel.Setup) {
 
 	// Core agent (disabled for Error Tracking Standalone, Logs Collection Only)
 	config.BindEnvAndSetDefault("core_agent.enabled", true)
+
+	// Software Inventory
+	config.BindEnvAndSetDefault("software_inventory.enabled", false, "DD_SOFTWARE_INVENTORY_ENABLED")
```
hush-hush: The aliasing for env vars is automatic (format is DD_+< setting name in capital with dot replace by underscore >)

```suggestion
	config.BindEnvAndSetDefault("software_inventory.enabled", false)
```

pkg/inventory/software/collector.go
```diff
@@ -0,0 +1,136 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
```
hush-hush, line 1: Same here could we rename this to not use `inventory` ? Maybe just `pkg/softwaremonitoring` ?

pkg/inventory/software/config_test.go
```diff
@@ -16,25 +16,25 @@ import (
 
 func TestSoftwareInventoryModuleConfig(t *testing.T) {
 	t.Run("via YAML", func(t *testing.T) {
-		cfg := configmock.NewSystemProbe(t)
+		cfg := configmock.New(t)
```
hush-hush, line 19: Do we want the config from `datadog.yaml` or `system-probe.yaml` ? (`New` gives you datadog.yaml and `NewSystemProbe` give you the other).

cmd/agent/subcommands/run/command_windows.go
hush-hush: Do we want to have this only for Windows ?

hush-hush (changes requested):

comp/metadata/inventorysoftware/impl/inventorysoftware.go
julien-lebot: It allows injecting a `SysProbeClient` for unit testing. As per @GustavoCaso comment I'll make the DI constructor private.

comp/metadata/inventorysoftware/mock/mock.go
julien-lebot: Thanks for the clarification, it was not obvious from the documentation that this was for external consumption only. I can move the mock back into the `impl` folder.

comp/metadata/inventorysoftware/def/component.go
julien-lebot: Not right now no, I can remove it from the interface

comp/metadata/inventorysoftware/def/component.go
julien-lebot: From the name of the folder or from the package name (or both)?
Will it be obvious what `comp/metadata/software` does?

comp/metadata/inventorysoftware/impl/inventorysoftware.go
julien-lebot: It's not pulled again every time: the status code only refreshes the list if it's empty - otherwise it reflects the current state to avoid re-collection every time the Agent status runs.

comp/metadata/inventorysoftware/impl/inventorysoftware.go
julien-lebot: Yeah there's a `getProvides` function, I can merge this one with it and remove the extra constructor.

GustavoCaso:
@julien-lebot could we provide a screenshot in the PR description of the status output and the GUI with the new component working? Mostly, to have a visual reference of the new changes 😄

cmd/agent/subcommands/run/command_windows.go
julien-lebot: Yes, because it's a no-op on Linux/Mac for now.

pkg/inventory/software/config_test.go
julien-lebot: From `datadog.yaml`, the config changed file.

comp/metadata/inventorysoftware/def/component.go
hush-hush: Now that I understand that the product is called `software inventory`, could we rename both the folder and package to `softwareinventory` ?

comp/metadata/inventorysoftware/impl/inventorysoftware.go
hush-hush: Thanks for explaining to me that's the name of the product and has no link with the old naming convention for REDAPL.Ignore this comment :+1: .

pkg/config/setup/system_probe.go
```diff
@@ -441,9 +441,6 @@ func InitSystemProbeConfig(cfg pkgconfigmodel.Config) {
 	cfg.BindEnvAndSetDefault(join(gpuNS, "max_streams"), 100)
 	cfg.BindEnvAndSetDefault(join(gpuNS, "max_stream_inactivity_seconds"), 30) // 30 seconds by default, includes two checks at the default interval of 15 seconds
 
-	// Software Inventory
-	cfg.BindEnvAndSetDefault(join(swNS, "enabled"), false)
```
GustavoCaso, line 445: Would this change be considered a breaking change? 
What would happen to customers using this flag to turn the feature on? Should we add an extra check in `pkg/config/setup/config.go` that reads the old environment variable and logs a deprecation warning if it is still being used?

comp/metadata/inventorysoftware/def/component.go
julien-lebot: Yes for sure

.github/CODEOWNERS
```diff
@@ -334,6 +334,7 @@
 /comp/metadata/clusteragent @DataDog/container-platform
 /comp/metadata/haagent @DataDog/ndm-core
 /comp/metadata/hostgpu @DataDog/ebpf-platform
+/comp/metadata/inventorysoftware @DataDog/windows-agent
```
iglendd: Is this directory accurate anymore (looks like this reference says otherwise ""github.com/DataDog/datadog-agent/comp/metadata/softwareinventory/def". May be it was true before?

comp/metadata/inventorysoftware/impl/inventorysoftware.go
julien-lebot: Done

comp/metadata/inventorysoftware/impl/inventorysoftware.go
julien-lebot: Checked!

comp/metadata/inventorysoftware/impl/inventorysoftware.go
julien-lebot: It should not be called anymore, since `InventoryPayload` wont' be initialized if the config is false (therefore `InventoryEnabled(conf)` will return false).

comp/metadata/softwareinventory/impl/inventorysoftware.go
```diff
@@ -0,0 +1,201 @@
...
+
+	installedSoftware, err := is.sysProbeClient.GetCheck(sysconfig.SoftwareInventory)
+	if err != nil {
+		return is.log.Errorf("error getting software inventory: %v", err)
+	}
+
+	is.cachedInventory = installedSoftware
+
+	return nil
+}
```
iglendd: Where sysProbeClient.GetCheck is defined. I cannot find signature which takes bool. I see this signature GetCheck(module types.ModuleName) ([]software.Entry, error) though

comp/metadata/inventorysoftware/impl/inventorysoftware_test.go
julien-lebot: I've updated the unit tests, there is now a test that tests when software inventory is disabled.

comp/metadata/inventorysoftware/impl/inventorysoftware_test.go
julien-lebot: I agree, I've updated the code.

comp/metadata/inventorysoftware/mock/mock.go
julien-lebot: Moved mock back into the `impl` folder.

pkg/inventory/software/collector.go
julien-lebot: In light of our discussion regarding software inventory, does this comment still hold?

pkg/config/setup/system_probe.go
julien-lebot: So, technically yes - we did release with this flag. However it was not advertised (or advertised by accident) to the customer as this feature is still a WIP. We should be fine breaking this.

comp/metadata/softwareinventory/impl/status_templates/inventory.tmpl
```diff
@@ -0,0 +1,11 @@
+{{- $swMap := index . "software_inventory_metadata" }}
+{{- if not $swMap }}
+No software inventory metadata - is System Probe running?
```
iglendd: Why would not be System Probe running since we configure it to always run for EUDM? I feel that it could be only due to errors, no?

iglendd (approved):
LG

comp/metadata/softwareinventory/impl/inventorysoftware.go
julien-lebot: The main function is here:

https://github.com/DataDog/datadog-agent/blob/335c148a42295408e63de344fb4471e4c23949fd/pkg/system-probe/api/client/client.go

What you are looing at here is our wrapper `func (w *sysProbeClientWrapper) GetCheck(module types.ModuleName) ([]software.Entry, error)`

comp/metadata/softwareinventory/impl/status_templates/inventory.tmpl
julien-lebot: That's true, let me remove that - we should not display anything when `software_inventory` is not enabled.

.github/CODEOWNERS
julien-lebot: It should be `softwareinventory` now

comp/metadata/softwareinventory/impl/status_templates/inventory.tmpl
julien-lebot: Actually, this can happen if software inventory is enabled but system probe is not running (crashed?) so that could be a good diagnostic - wdyt?

pkg/config/setup/config.go
```diff
@@ -1200,6 +1200,10 @@ func agent(config pkgconfigmodel.Setup) {
 
 	// Core agent (disabled for Error Tracking Standalone, Logs Collection Only)
 	config.BindEnvAndSetDefault("core_agent.enabled", true)
+
+	// Software Inventory
+	config.BindEnvAndSetDefault("software_inventory.enabled", false, "SOFTWARE_INVENTORY_ENABLED")
```
hush-hush: Every env vars we support starts with `DD_`, I don't think we should create an exception here. `BindEnvAndSetDefault` will automatically add support for `DD_SOFTWARE_INVENTORY_ENABLED`.

pkg/config/setup/system_probe.go
julien-lebot: This flag now exists on the Agent side, so there should not be a big difference for the customer - they just need to use the `datadog.yaml` instead. Since that feature was not intended for a general audience I think it's ok to do this kind of breaking change.

cmd/agent/subcommands/run/command_windows.go
```diff
@@ -119,6 +121,7 @@ func StartAgentWithDefaults(ctxChan <-chan context.Context) (<-chan error, error
 			_ inventoryagent.Component,
 			_ inventoryhost.Component,
 			_ inventoryotel.Component,
+			_ softwareinventory.Component,
```
GustavoCaso: I believe that with the trick of adding `fx.Invoke(func(_ softwareinventory.Component) {}),` to the `Module()` function, we no longer have to declare the component here 😄

comp/metadata/softwareinventory/impl/status.go
```diff
@@ -0,0 +1,92 @@
...
+
+// populateStatus populates the status map with software inventory data.
+// This method processes the cached inventory data and formats it for display
+// in the status output. It handles date formatting and organizes the data
+// by software ID for easy lookup.
+func (is *softwareInventory) populateStatus(status map[string]interface{}) {
+	data := map[string]interface{}{}
+	if is.cachedInventory == nil {
+		_ = is.refreshCachedValues()
+	}
```
GustavoCaso: We are making a shallow error if the `cachedInventory` should be used as part of the `software_inventory_metadata`, so if the feature is enabled but the `refreshCachedValues` fails, at least we display some information.

GustavoCaso (approved):
Awesome work 🎉

comp/metadata/softwareinventory/impl/status.go
julien-lebot: I've added a warning if the err is not nil

cmd/agent/subcommands/run/command_windows.go
julien-lebot: I've removed the component and tested to make sure it still works !

pkg/inventory/software/collector.go
hush-hush: I think so, maybe a single folder `pkg/softwareinventory`, unless we plan other kind of inventory ?

hush-hush (changes requested):
Added one last comment.

pkg/config/setup/config.go
julien-lebot: Fixed!

cmd/system-probe/subcommands/run/command.go
```diff
@@ -97,7 +97,13 @@ func Commands(globalParams *command.GlobalParams) []*cobra.Command {
...
 			return fxutil.OneShot(run,
 				fx.Supply(config.NewAgentParams("", config.WithConfigMissingOK(true))),
-				fx.Supply(sysprobeconfigimpl.NewParams(sysprobeconfigimpl.WithSysProbeConfFilePath(globalParams.ConfFilePath), sysprobeconfigimpl.WithFleetPoliciesDirPath(globalParams.FleetPoliciesDirPath))),
+				// Force FX to load Datadog configuration before System Probe config.
+				// This is necessary because the 'software_inventory.enabled' setting is defined in the Datadog configuration.
+				// Without this explicit dependency, FX might initialize System Probe's config first, causing pkgconfigsetup.Datadog().GetBool()
+				// to return default values instead of the actual configuration.
+				fx.Provide(func(_ config.Component) sysprobeconfigimpl.Params {
+					return sysprobeconfigimpl.NewParams(sysprobeconfigimpl.WithSysProbeConfFilePath(globalParams.ConfFilePath), sysprobeconfigimpl.WithFleetPoliciesDirPath(globalParams.FleetPoliciesDirPath))
+				}),
```
val06, line 106: is that the only possible solution? I don't like that much introducing this dependency that could impose some limitations later.

val06 (commented):
left comments, reviewed only ebpf-platform owned files

cmd/system-probe/subcommands/run/command.go
julien-lebot: What kind of limitations are you thinking about specifically? Also, @hush-hush is working on a more permanent fix that would make this dependency part of the default config, but until that lands, this change is necessary to ensure correct load order.

hush-hush: The solution here is a workaround more that a final solution.

For context we have 2 comp for loading configuration file: `config` loads datadog.yaml and `sysprobeconfig` loads system-probe.yaml.

The core issue is that `sysprobeconfig` does more that loading the file it also includes part of the init whole init logic for system-probe (ie this https://github.com/DataDog/datadog-agent/pull/38072/files#diff-0daece1cfe8277578a50f7c867a3263c7b21fda3600c4a5f403c0bfc1b053df8L98). This part of the code now requires the config from datadog.yaml and we need to express that dependency somewhere.

We have a few option here but they require more work and I think could be done in a following PR since @julien-lebot is leaving soon for OOO time:

1. Split the system-prove init logic out of `sysprobeconfig` and into its own comp. This would be the ideal solution as it splits things in well defined logical blocks (1 comp load datadog.yaml, one comp load system-probe.yaml and one consume both to init system-prove). This would also require the most work.
2. We add the `config` comp as a deps for `sysprobeconfig`. This would enforce these dependencies for everyone using `sysprobeconfig`. I think this might be the best solution as it's easy to implement and I think we already read datadog.yaml everywhere we load system-probe.yaml
3. We update system-probe to allow enabling product dynamically (ie this https://github.com/DataDog/datadog-agent/pull/38072/files#diff-0daece1cfe8277578a50f7c867a3263c7b21fda3600c4a5f403c0bfc1b053df8R181 can be achieve after sysprobeconfig from other comp.

I would push for `2.` personally.

hush-hush (approved):

cmd/system-probe/subcommands/run/command.go
val06: reading that again, this affect only the configs order and not the modules order so it is less severe, I would still avoid having those dependencies ad-hoc in the code when possible, but your comment about @hush-hush upcoming improvement is enough to address that.

val06 (approved):
LGTM for ebpf-platform files

PR #38075, opened 2025-06-19
fix(agent): Report JMX checks again in the metadata payloads

BaptisteFoy:
### What does this PR do?
This PR enables collection of JMX integrations status & configuration, respectively in the `agent_check` & `inventorychecks` payloads.

### Motivation
[FRFA-94], [FRFA-117]

I tested all of that manually; by creating a VM with a JMX server.

I'm not 100% sure if what I did is the right way to handle "config ID". We need to have the same "config ID" for both payloads; and other integrations usually use a config hash; but we don't have access to that from the `agent_check` payload. Instead; I'm using the `name` field from the config (which we have). It is either populated by the customer or automatically generated from the integration type, the host, and the port. If there are multiple instances of a single check with the same name it's considered the same integration by the agent, so we can consider it the same in the payloads too.

### Additional Notes
Support was removed as a side effect https://github.com/DataDog/datadog-agent/pull/26243/s 😞 

[FRFA-94]: https://datadoghq.atlassian.net/browse/FRFA-94?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FRFA-117]: https://datadoghq.atlassian.net/browse/FRFA-117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


coignetp (approved):

pkg/jmxfetch/state.go
```diff
@@ -98,3 +100,38 @@ func StopJmxfetch() {
...
+		if err != nil {
+			return nil, fmt.Errorf("unable to parse JMX configuration: %w", err)
+		}
+
+		c := map[string]interface{}{}
+		c["init_config"] = GetJSONSerializableMap(rawInitConfig)
+		instances := []integration.JSONMap{}
+		for _, instance := range config.Instances {
+			var rawInstanceConfig integration.JSONMap
+			err := yaml.Unmarshal(instance, &rawInstanceConfig)
```
vickenty, line 122: Would it make sense to sanitize usernames and passwords from the check config, or does it happen automatically?

hestonhoffman (approved):

comp/metadata/inventorychecks/inventorychecksimpl/integrations_nojmx.go
```diff
@@ -0,0 +1,13 @@
...
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+//go:build !jmx
+
+package inventorychecksimpl
+
+func (ic *inventorychecksImpl) getJMXChecksMetadata() (jmxMetadata map[string][]metadata) {
+	// This function is a no-op when JMX is not enabled.
+	return nil
```
misteriaud: **💬 suggestion:** ‏Maybe it would be better to return an empty map to avoid possible nil dereference in the unfortunate situation of post-processing payload produced by this no-op impl ?

comp/collector/collector/collectorimpl/agent_check_metadata.go
```diff
@@ -115,6 +115,42 @@ func (c *collectorImpl) GetPayload(ctx context.Context) *Payload {
...
+				}
+				for _, checkRaw := range checks {
+					check, ok := checkRaw.(map[string]interface{})
+					if !ok {
+						continue
+					}
+					checkStatus, ok := check["status"].(string)
+					if !ok {
+						checkStatus = "OK"
+					}
```
misteriaud, line 136: **❓ question:** ‏Why do we consider a `check["status"]` not being a string a `"OK"` status ?

misteriaud (commented):
I left some comments, but overall LGTM

rahulkaukuntla (approved):

comp/collector/collector/collectorimpl/agent_check_metadata.go
coignetp: The default status of a check is `OK` ; unless there is an error or a warning in the status (which would be a string)

vickenty (approved):

pkg/jmxfetch/state.go
coignetp: Passwords are scrubbed automatically

comp/collector/collector/collectorimpl/agent_check_metadata.go
misteriaud: Thanks for the explanation ! Could you add a comment or rename the `ok` variable please, just so future contributors understand the why.

misteriaud (approved):

PR #38078, opened 2025-06-19
[CWS] use unspec instead of ok for rawpacket

safchain:
### What does this PR do?

Let other classifiers classify the packet

### DIFF
--- a/pkg/security/ebpf/probes/rawpacket/pcap.go
+++ b/pkg/security/ebpf/probes/rawpacket/pcap.go
@@ -31,9 +31,12 @@ const (
 	packetCaptureSize = 256
 
 	// state of tc action
-	tcActOk     = 0  // will terminate the packet processing pipeline and allows the packet to proceed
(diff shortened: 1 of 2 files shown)


paulcacheux (approved):

pkg/security/ebpf/probes/rawpacket/pcap.go
```diff
@@ -29,6 +29,11 @@ const (
 
 	// packetCaptureSize see kernel definition
 	packetCaptureSize = 256
+
+	// state of tc action
+	tcActOk     = 0  // will terminate the packet processing pipeline and allows the packet to proceed
```
paulcacheux: won't this trip the linter ? like unused variable

YoannGh (approved):

pkg/security/ebpf/probes/rawpacket/pcap.go
safchain: this is a const, so I hope not

PR #38083, opened 2025-06-19
Pre-build binaries for e2e tests

KevinFairise2:
### What does this PR do?

This PR put the build of e2e test binaries in a dedicated job. So we build them all once and then we can just execute them.
It has several advantages:
- We can build the binary as soon as the pipeline start which is some time saved in the test execution (not really relevant with a simple build but will be more interesting when the build is done with orchestrion instrumentation)
- It reduces the resource usage of the e2e tests jobs. We move from a lot of long-running jobs that have high resource to one big job that is relatively short, and a lot of job that are long-running but with less resource required. The PR does not change the resource given that they all require different resouces depending on the resource they are creating. The goal will be to activate the VPA on e2e stage, so that all the jobs have their requests automatically adjusted.

### Motivation

We want to instrument our e2e tests jobs using orchestrion. If we do that without changing anything it adds between 5 and 15 minutes to the e2e tests jobs which will make the whole pipeline longer. This change is a mitigation of that issue and is explained here: https://datadoghq.atlassian.net/wiki/x/u4ehMwE

### Additional Notes

REVERT INSTRUCTION

If this PR ever bring some unexpect failures in the CI and need to be reverted, it can be reverted by setting `PRE_BUILT_BINARIES_FLAG` to `""` in `.new_e2e_template` of `e2e.yml` file and by removing the `go_e2e_test_binaries` from `.needs_new_e2e_template`


.gitlab/e2e/e2e.yml
```diff
@@ -92,6 +93,35 @@
...
+    KUBERNETES_CPU_REQUEST: 8
+    KUBERNETES_MEMORY_REQUEST: 64Gi
+    KUBERNETES_MEMORY_LIMIT: 64Gi
+  before_script:
+    - !reference [.retrieve_linux_go_e2e_deps]
+  script:
+    - dda inv -- -e new-e2e-tests.build-binaries --output-dir test-binaries -p $KUBERNETES_CPU_REQUEST --manifest-file-path manifest.json
+    - tar czf test-binaries.tar.gz test-binaries
+  artifacts:
+    expire_in: 1 day
```
CelianR: Question: is it enough? Usually I see 2 weeks

.gitlab/e2e/e2e.yml
```diff
@@ -441,8 +463,7 @@ new-e2e-cws:
   variables:
     TARGETS: ./tests/cws
     TEAM: csm-threats-agent
-    # Temporarily disable the test on FIPS pipeline, as remote-configuration is not available with FIPS Agent yet
-    # ON_NIGHTLY_FIPS: "true"
+    ON_NIGHTLY_FIPS: "true"
```
CelianR: Is it expected? We disabled this since remote config is not available with fips

.gitlab/e2e/e2e.yml
```diff
@@ -696,8 +714,7 @@ new-e2e-ha-agent:
   variables:
     TARGETS: ./tests/ha-agent
     TEAM: ndm-core
-    # Temporarily disable the test on FIPS pipeline, as remote-configuration is not available with FIPS Agent yet
-    # ON_NIGHTLY_FIPS: "true"
+    ON_NIGHTLY_FIPS: "true"
```
CelianR: Same question here

internal/tools/gotest-custom/main.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	wg := sync.WaitGroup{}
+	wg.Add(len(binaries))
+	outputs := make([]*bytes.Buffer, len(binaries))
+	for idx := range outputs {
+		outputs[idx] = &bytes.Buffer{}
+	}
+	errChannel := make(chan error, len(binaries))
+	// printLock := sync.Mutex{}
```
CelianR:
```suggestion
```

internal/tools/gotest-custom/main.go
```diff
@@ -0,0 +1,248 @@
...
+	cwd, err := os.Getwd()
+	if err != nil {
+		fmt.Printf("Error getting current working directory: %v\n", err)
+		os.Exit(1)
+	}
+
+	//
```
CelianR:
```suggestion
```

CelianR (approved):
Great work!

.gitlab/e2e/e2e.yml
KevinFairise2: Bumped to 1 week to make sure it does not break retries of the e2e jobs. But want to avoid too long retention because the artifact is a bit large.

.gitlab/e2e/e2e.yml
KevinFairise2: Definitely not expected. Reverted these changes

pgimalac (approved):
LGTM for Agent Runtimes files

PR #38092, opened 2025-06-20
[APM] Populate Lang on Stats Payload

ichinaski:
### What does this PR do?

This field is defined on ClientStatsPayload, but never used.

This PR modified both the Concentrator and the ClientStatsAggregator to
ensure we populate it.

Having this field accessible in the backend will allow us to compute
more granular metrics on CSS adoption.


ajgajg1134 (approved):
This looks good to me, have we spoken to the stats team to ensure this won't cause any cardinality issues / give them a heads up? I can't imagine this would cause problems but would be good to make sure they're aware :)

ichinaski:
> This looks good to me, have we spoken to the stats team to ensure this won't cause any cardinality issues / give them a heads up? I can't imagine this would cause problems but would be good to make sure they're aware :)

This won't have any impact on actual trace metrics. It's just attached to the payload, and the backend can leverage it if/when needed -- in our case, to track CSS adoption. Cardinality on such a metric is fine (checked with backend team).

PR #38096, opened 2025-06-20
[EBPF] gpu: only set the NVML library instance if it's correctly initialized

gjulianm:
### What does this PR do?

Change the NVML library wrapper to only set the library if the initialization works.

### Motivation

If the initialization fails, the library object will still be set but the capabilities will not be populated. This means that the next initialization attempt will succeed, as the library object is set. It will lead to error messages like "nvmlXXX function not available" when trying to call functions which can be confusing. Instead, this returns an error on all initializations if the NVML init fails.

Validated in an environment where NVML cannot init, added unit tests too.


pkg/gpu/safenvml/lib.go
```diff
@@ -151,7 +151,7 @@ func (s *safeNvml) DeviceGetHandleByIndex(idx int) (SafeDevice, error) {
 
 // populateCapabilities verifies nvml API symbols exist in the native library (libnvidia-ml.so).
 // It returns an error only if a critical symbol is missing (to properly initialize device list and create a new safe device wrapper)
-func (s *safeNvml) populateCapabilities() error {
+func (s *safeNvml) populateCapabilities(lib nvml.Interface) error {
```
val06: why? this is a method of safeNvml that already has this field in the struct.
if it is for mocks, can't we just init the "fail" mock for the safeNvml struct?

val06 (commented):
reviewed

pkg/gpu/safenvml/lib.go
gjulianm: It's not because of the mocks, but because this method is called while the library hasn't been set in the struct yet. If we set it in the struct and then call `populateCapabilities` and that fails, we'll have a partially initialized library.

val06: let's change this function to be a helper func and not part of the safeNvml struct's methods. it can return `map[string]struct, error` return values

val06 (approved):

PR #38098, opened 2025-06-20
Fix config in OTLP ingest

mackjmr:
### What does this PR do?

- Add otlp_config.enabled field.
- Fix bug where empty otlp_config.receiver would trigger otlp ingest to be enabled.


comp/otelcol/otlp/configcheck/configcheck.go
```diff
@@ -75,7 +75,15 @@ var intConfigs = map[string]struct{}{
 
 // IsEnabled checks if OTLP pipeline is enabled in a given config.
 func IsEnabled(cfg config.Reader) bool {
-	return hasSection(cfg, coreconfig.OTLPReceiverSubSectionKey)
+	if cfg.Get("otlp_config.enabled") == true {
+		return true
+	}
```
hush-hush: If explicitly set to false by the user we should not enable OTLP I think

```suggestion
	if cfg.Get("otlp_config.enabled") == true {
		return true
	} else if  cfg.IsConfigured("otlp_config.enabled") {
		return false
	}
```

hush-hush (commented):
Should we update the configuration template too to add `otlp_config.enabled` ?

mackjmr:
thanks @hush-hush, added to config template.

hestonhoffman (approved):

dustmop (approved):
LGTM for agent-configuration

pkg/config/config_template.yaml
```diff
@@ -4453,6 +4453,13 @@ api_key:
 #
 # otlp_config:
 
+  ## @param enabled - boolean - optional - default: false
+  ## @env DD_OTLP_CONFIG_ENABLED - boolean - optional - default: false
+  ## Set to true to enable the OTLP ingest (or set any field in the the otlp_config.receiver.* section).
+  #
+  # enabled: true
```
hush-hush, line 4460: We should use the default as the example.

hush-hush (changes requested):

mackjmr:
Closing this PR. This change will have a larger impact than expected, as the empty section was not only supported for `otlp_config.receiver`, but it was also supported for subsections (this behavior is used). For example,
```
otlp_config:
  receiver:
    protocols:
      http:
      grpc:
```

I'll write a one pager about the issue.

PR #38099, opened 2025-06-20
[AGENTRUN-500] Add libraries folder to the runtime search path to have CMake find RTLoader libraries on macOS

Enzu83:
### What does this PR do?

This PR fixes the problem of `no LC_RPATH's found` by adding the RTLoader libraries folder to the runtime search path `rpath`.

### Motivation

I was unable to run the RTLoader tests due to a `no LC_RPATH's found` issue, which occurred when executing `dda inv rtloader.test`. For some reason, CMake is no longer able to locate the RTLoader libraries.

Here's the error:

```
dyld[28146]: Library not loaded: @rpath/libdatadog-agent-rtloader.2.dylib
  Referenced from: <13C11ED4-AE0C-3BC4-FB03-9F36D382F1B0> /private/var/folders/lf/k2z6_b753rng0ckph1wclg1c0000gp/T/go-build2770284591/b250/containers.test
  Reason: no LC_RPATH's found
signal: abort trap
```

### DIFF
--- a/rtloader/test/CMakeLists.txt
+++ b/rtloader/test/CMakeLists.txt
@@ -26,7 +26,7 @@ if (WIN32)
 elseif(APPLE)
     set (CGO_LDFLAGS -L${PROJECT_BINARY_DIR}/rtloader -ldatadog-agent-rtloader -ldl -rpath ${PROJECT_BINARY_DIR}/../../dev/lib)
 else()
-	set (CGO_LDFLAGS -L${PROJECT_BINARY_DIR}/rtloader -ldatadog-agent-rtloader -ldl)
+    set (CGO_LDFLAGS -L${PROJECT_BINARY_DIR}/rtloader -ldatadog-agent-rtloader -ldl)
 endif()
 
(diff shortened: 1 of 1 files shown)


rtloader/test/CMakeLists.txt
```diff
@@ -23,6 +23,8 @@ set (CGO_CFLAGS \"-I${CMAKE_SOURCE_DIR}/include -I${CMAKE_SOURCE_DIR}/common -Wn
 
 if (WIN32)
     set (CGO_LDFLAGS -L${PROJECT_BINARY_DIR}/rtloader -ldatadog-agent-rtloader -lstdc++ -static)
+elseif(APPLE)
+    set (CGO_LDFLAGS -L${PROJECT_BINARY_DIR}/rtloader -ldatadog-agent-rtloader -ldl -rpath ${PROJECT_BINARY_DIR}/../../dev/lib)
 else()
     set (CGO_LDFLAGS -L${PROJECT_BINARY_DIR}/rtloader -ldatadog-agent-rtloader -ldl)
 endif()
```
louis-cqrl, line 30: can you check that windows and linux do not need this parameter as well ?

louis-cqrl (approved):

PR #38100, opened 2025-06-20
Add First and Last Seen per version to activity tree nodes

sarrazaghbib-dev:
### What does this PR do?
This PR introduces a shared NodeBase that every node embeds, which maintains a small map of version → timestamps.
It also provides two helpers :
- `func (b *NodeBase) Record(version string, now uint64)`: updates FirstSeen/LastSeen for each node per version
- `func (b *NodeBase) EvictVersion(version string) ` : deletes the corresponding version

### Motivation
Currently, none of the node types in our activity tree (e.g. ProcessNode, ExecNode, NetworkActivityNode, etc.) store per-version (image-tag) timestamps. To instore a future eviction logic of nodes by version we first need to be able to determine when was each node last seen by version. FirstSeen is also meant for future context enrichment of each node.


pkg/security/process_list/base_node.go
spikat, line 1: you should not modify anything in `process_list/`, ATM it's not used at all. Please move your base_node.go file to activity_tree directory

pkg/security/process_list/exec_node.go
spikat, line 1: no need to modify this file

pkg/security/process_list/process_node.go
spikat, line 1: no need to modify this file

pkg/security/process_list/base_node.go
```diff
@@ -0,0 +1,44 @@
+package processlist
+
+import (
+	"time"
+)
+
+// VersionTimes holds the first and last seen timestamps for a specific version (image tag).
```
spikat: please modify every occurrence of "Version" by "ImageTags" (as it's already the name used for that on other code parts)

pkg/security/process_list/base_node.go
```diff
@@ -0,0 +1,44 @@
...
+	if vt, ok := b.Seen[imageTag]; ok {
+		if now.Before(vt.FirstSeen) {
+			vt.FirstSeen = now
+		}
+		if now.After(vt.LastSeen) {
+			vt.LastSeen = now
+		}
+		return
+	}
+	b.Seen[imageTag] = &VersionTimes{FirstSeen: now, LastSeen: now}
```
spikat: Too complicated and redundant:
* You should not set `FirstSeen` if `ok == true`
* You should set directly set LastSeen without checking `now.After()` (I don't see a reason why we need the check)

pkg/security/process_list/base_node.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// EvictVersion removes the stored timestamps for an imageTag.
+func (b *NodeBase) EvictVersion(imageTag string) {
+	delete(b.Seen, imageTag)
+}
```
spikat: IMHO we should:
* return true/false if we removed an imageTag or not
* add another eviction helper based on a "before" timestamp filter rather than an imageTag (even if for now we won't use it on this PR)

pkg/security/security_profile/activity_tree/process_node.go
```diff
@@ -58,14 +62,18 @@ func NewProcessNode(entry *model.ProcessCacheEntry, generationType NodeGeneratio
...
+	node := &ProcessNode{
 		Process:        entry.Process,
 		GenerationType: generationType,
 		Files:          make(map[string]*FileNode),
 		DNSNames:       make(map[string]*DNSNode),
 		IMDSEvents:     make(map[model.IMDSEvent]*IMDSNode),
 		NetworkDevices: make(map[model.NetworkDeviceContext]*NetworkDeviceNode),
 	}
+	node.NodeBase = processlist.NewNodeBase()
+	node.Record("", now)
```
spikat: This makes no sense.

You should look at places where we put or remove ImageTags and act accordingly. IMHO you should use of the Record method in:
* AppendImageTag
* applyImageTagOnLineageIfNeeded
* TagAllNodes

And  use the evict method in EvictImageTag

Plus, apply the same logic on other node types

pkg/security/security_profile/activity_tree/process_node.go
```diff
@@ -217,6 +225,10 @@ func (pn *ProcessNode) Matches(entry *model.Process, matchArgs bool, normalize b
 
 // InsertSyscalls inserts the syscall of the process in the dump
 func (pn *ProcessNode) InsertSyscalls(e *model.Event, imageTag string, syscallMask map[int]int, stats *Stats, dryRun bool) bool {
+	if !dryRun {
```
spikat: I'm not sure we should check the dryRun (here and on all the other places). My point being: if we look for something, that because the had SEEN the activity and are wondering if we already have the node present (like to auto-suppress a rule or generate an anomaly detection). Even if we don't add new node, IMHO we should update the LastSeen time stamp.

pkg/security/security_profile/activity_tree/process_node.go
```diff
@@ -217,6 +225,10 @@ func (pn *ProcessNode) Matches(entry *model.Process, matchArgs bool, normalize b
...
+		pn.updateTimes(e)
```
spikat: I wonder if we should update the process node when inserting/modifying a subnode on it (here and on all the other places). If the process "last seen" means "last seen executed" we should not. @safchain WDYT?

pkg/security/security_profile/activity_tree/process_node.go
```diff
@@ -523,3 +555,12 @@ func (pn *ProcessNode) EvictImageTag(imageTag string, DNSNames *utils.StringKeys
...
 }
+
+func (pn *ProcessNode) UpdateLastSeen(version string) {
+	pn.Record(version, time.Now())
+}
+
+func (pn *ProcessNode) updateTimes(event *model.Event) {
+	eventTime := event.ResolveEventTime()
+	pn.Record(event.ContainerContext.Tags[0], eventTime)
+}
```
spikat: Instead of having multiple definition of the same function for every node type, you can move it to your baseNode one.
Also, you must not randomly use a tag as an imageTag, but rather use the right imageTag depending of where you call the method.

pkg/security/security_profile/activity_tree/process_node.go
```diff
@@ -523,3 +555,12 @@ func (pn *ProcessNode) EvictImageTag(imageTag string, DNSNames *utils.StringKeys
 	pn.Children = newChildren
 	return false
 }
+
+func (pn *ProcessNode) UpdateLastSeen(version string) {
+	pn.Record(version, time.Now())
+}
```
spikat: I don't see the point of this function if we already have the "updateTimes" one

pkg/security/security_profile/tests/base_node_test.go
```diff
@@ -0,0 +1,396 @@
...
+		}
+
+		// Verify the version exists and has been updated
+		assert.True(t, node.HasVersion("concurrent-tag"))
+		
+		versionTimes, exists := node.Seen["concurrent-tag"]
+		assert.True(t, exists)
+		assert.Greater(t, versionTimes.LastSeen.Unix(), versionTimes.FirstSeen.Unix())
+	})
+}
```
spikat: No need for that, it should not be concurrent

pkg/security/security_profile/tests/base_node_test.go
```diff
@@ -0,0 +1,396 @@
...
+	t.Run("full_lifecycle", func(t *testing.T) {
+		// Create a new process node
+		node := activity_tree.NewProcessNode(createTestProcessCacheEntry(), activity_tree.Runtime, nil)
+
+		// Step 1: Record initial time
+		initialTime := time.Date(2023, 1, 1, 12, 0, 0, 0, time.UTC)
+		node.Record("test-image", initialTime)
+		
+		assert.True(t, node.HasVersion("test-image"))
+		assert.Equal(t, []string{}, node.ImageTags) // No image tags yet
```
spikat: doesn't make sense to have a version timestamps without having the version in nodes.

pkg/security/security_profile/tests/base_node_test.go
```diff
@@ -0,0 +1,396 @@
...
+
+			// Update last seen (this is the public method that updates times)
+			node.UpdateLastSeen("test-image")
+
+			// Verify the node still exists and hasn't been corrupted
+			assert.NotNil(t, node)
+			assert.True(t, node.HasVersion("test-image"))
+		})
+	}
+}
```
spikat: this test doesn't make any sense, it even don't use the "eventTime". please either stop using cursor or at least ENSURE to read what it generates before comiting it.

pkg/security/security_profile/tests/base_node_test.go
spikat, line 1: Most of the tests are aiming at checking the ~same stuff without adding real value. Please once the other parts get fixed, take time to have a proper manual look at those tests, and adapt/improve/remove them when needed.

pkg/security/security_profile/activity_tree/process_node.go
```diff
@@ -33,6 +36,7 @@ type ProcessNodeParent interface {
 
 // ProcessNode holds the activity of a process
 type ProcessNode struct {
+	processlist.NodeBase
```
spikat: The idea was also to replace the `ImageTags` array each node stores. We should not have two different ways of storing "versions" on nodes.

pkg/security/security_profile/activity_tree/process_node.go
sarrazaghbib-dev: I assumed dry-runs wouldn’t count as real observations, but you’re right: any time we consult a node dryRun or not it reflects an actual operation.

sarrazaghbib-dev: I assumed dry-runs wouldn’t count as real observations, but you’re right: any time we consult a node dryRun or not it reflects an actual operation.

pkg/security/security_profile/tests/base_node_test.go
sarrazaghbib-dev: deleted the entire test file and made new tests

pkg/security/security_profile/activity_tree/process_node.go
```diff
@@ -216,15 +221,13 @@ func (pn *ProcessNode) Matches(entry *model.Process, matchArgs bool, normalize b
...
+func (pn *ProcessNode) InsertSyscalls(e *model.Event, imageTag string, syscallMask map[int]int, stats *Stats, dryRun bool) bool {	
 	var hasNewSyscalls bool
 newSyscallLoop:
 	for _, newSyscall := range e.Syscalls.Syscalls {
 		for _, existingSyscall := range pn.Syscalls {
 			if existingSyscall.Syscall == int(newSyscall) {
-				if imageTag != "" && !slices.Contains(existingSyscall.ImageTags, imageTag) {
-					existingSyscall.ImageTags = append(existingSyscall.ImageTags, imageTag)
-				}
+					existingSyscall.Record(imageTag, e.ResolveEventTime())
```
safchain: before imageTag != "" was checked. Do we still need it ?

pkg/security/security_profile/activity_tree/dns_node.go
```diff
@@ -10,30 +10,30 @@ package activitytree
...
-
+	NodeBase
+	MatchedRules   []*model.MatchedRule
 	GenerationType NodeGenerationType
 	Requests       []model.DNSEvent
 }
 
 // NewDNSNode returns a new DNSNode instance
 func NewDNSNode(event *model.DNSEvent, rules []*model.MatchedRule, generationType NodeGenerationType, imageTag string) *DNSNode {
+	now := time.Now()
```
spikat: when possible we should use the event timestamp for consistency (and to avoid calling time.Now() too many times for a single event)

pkg/security/security_profile/activity_tree/base_node.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// Record updates FirstSeen/LastSeen for the given imageTag at time 'now'.
+func (b *NodeBase) Record(imageTag string, now time.Time) {
+	if vt, ok := b.Seen[imageTag]; ok {
+		vt.LastSeen = now
+		return
+	}
+	b.Seen[imageTag] = &ImageTagTimes{FirstSeen: now, LastSeen: now}
+}
```
spikat: We should add a safe guard here to check if the imageTag is != "". Nodes can be created/inserted for activity dumps where we may not have the correct imageTag (so, insert funcs will be called with no tags)

pkg/security/security_profile/activity_tree/dns_node.go
```diff
@@ -53,20 +53,18 @@ func dnsFilterSubdomains(name string, maxDepth int) string {
 }
 
 func (dn *DNSNode) appendImageTag(imageTag string) {
-	dn.ImageTags, _ = AppendIfNotPresent(dn.ImageTags, imageTag)
+	dn.Record(imageTag, time.Now())
 }
```
spikat: we should add another `time.Time` argument to all "appendImageTag" to try to use the same one everywhere and void calling time.Now() for nothing

pkg/security/security_profile/activity_tree/dns_node.go
spikat, line 1: NB: comments on this file are also valid for others type of nodes

pkg/security/security_profile/activity_tree/process_node.go
sarrazaghbib-dev: I think it would be better to keep the check to avoid having an empty string as an imageTag (I'll add it back)

pkg/security/security_profile/activity_tree/flow_node.go
```diff
@@ -9,48 +9,50 @@
...
 // NewFlowNode returns a new FlowNode instance
 func NewFlowNode(flow model.Flow, generationType NodeGenerationType, imageTag string) *FlowNode {
+	now := time.Now()
 	node := &FlowNode{
 		GenerationType: generationType,
 		Flow:           flow,
 	}
+	node.NodeBase = NewNodeBase()
+	node.Record(imageTag, now)
 	node.appendImageTag(imageTag)
```
spikat: duplicate

pkg/security/security_profile/activity_tree/flow_node.go
```diff
@@ -9,48 +9,50 @@
...
+	if node.IsEmpty() {
+		return true
 	}
 	return false
 }
 
 func (node *FlowNode) addFlow(flow model.Flow, imageTag string) {
 	if imageTag != "" {
-		node.appendImageTag(imageTag)
+		node.Record(imageTag, time.Now())
```
spikat: you made appendImageTag to call Record, it was a good idea, stick with it

pkg/security/security_profile/activity_tree/imds_node.go
```diff
@@ -9,42 +9,45 @@
...
+	imds.EvictImageTag(imageTag)
+	if imds.IsEmpty() {
+		return true
 	}
 	return false
 }
+
+
+	
+
```
spikat: please don't add those kind of changes

pkg/security/security_profile/activity_tree/network_device_node.go
spikat, line 1: you didn't modify appendImageTag

spikat (commented):
I did not finish the review and stopped after just some files, please have another look at what you committed and most importantly: at the CI jobs

pkg/security/security_profile/activity_tree/network_device_node.go
sarrazaghbib-dev: It doesn't need to be changed as it uses flow.appendImageTag that is updated to use the Record function

pkg/security/security_profile/activity_tree/imds_node.go
sarrazaghbib-dev: sorry ! I've removed a function there didn't notice that it kept empty lines

pkg/security/security_profile/activity_tree/flow_node.go
sarrazaghbib-dev: appendImageTag is a function will has been using to update the tags list and i've modified it after adding my record functions hence the inconsistency of use. I'll uniformise the code to use appendImageTag only.

pkg/security/security_profile/activity_tree/dns_node.go
sarrazaghbib-dev: Then probably we should just use Record instead ? I think `appendImageTag` would just be a wrapper function if we add the Time.time argument

pkg/security/security_profile/activity_tree/dns_node.go
sarrazaghbib-dev: DNSEvent as well as some other events such as IMDSEvent  have no field or method ResolveEventTime that's why I've used time.Now(). Maybe that wasn't the right way to retrieve the event timestamp.

pkg/security/security_profile/activity_tree/process_node.go
sarrazaghbib-dev: As @spikat suggested, I'll add a safe guard in the `Record` function in order to verify that imageTag != "" hence I'll just remove the rest of similar checks in the code.

pkg/security/security_profile/activity_tree/activity_tree.go
```diff
@@ -493,7 +494,7 @@ func (at *ActivityTree) buildBranchAndLookupCookies(entry *model.ProcessCacheEnt
 		if cs.isSet() {
 			fastMatch, found = at.CookieToProcessNode.Get(cs)
 			if found {
-				fastMatch.applyImageTagOnLineageIfNeeded(imageTag)
+				fastMatch.applyImageTagOnLineageIfNeeded(imageTag, fastMatch.Process.ExecTime)
```
safchain: I would remove the second argument as the Process.ExecTime file is part of fastMatch

pkg/security/security_profile/activity_tree/file_node.go
```diff
@@ -22,13 +22,13 @@ import (
...
 type FileNode struct {
+	NodeBase
 	MatchedRules   []*model.MatchedRule
 	Name           string
-	ImageTags      []string
 	IsPattern      bool
 	File           *model.FileEvent
 	GenerationType NodeGenerationType
-	FirstSeen      time.Time
+	FirstSeen      time.Time // I'm leaving this field here for now, but it should be removed (redundant)
```
safchain: Either we remove it or we keep the line 113-115 otherwise we'll just have an empty value

pkg/security/security_profile/activity_tree/syscalls_node.go
```diff
@@ -8,39 +8,29 @@
...
-		return false
-	}
-	if len(imageTags) == 0 {
-		return true
-	}
-	sn.ImageTags = imageTags
-	return false
+	sn.EvictImageTag(imageTag)
+	return sn.IsEmpty()
 }
```
spikat: nit: I see that every node type keeps its own `evictImageTag()`, calling the base node `EvictImageTag()` plus `IsEmpty()`: should we change the way the NodeBase eviction func behaves and remove all the eviction funcs from nodes? WDYT?

pkg/security/security_profile/activity_tree/activity_tree_proto_dec_v1.go
```diff
@@ -332,7 +340,7 @@ func protoDecodeIMDSNode(in *adproto.IMDSNode) *IMDSNode {
 
 	node := &IMDSNode{
 		MatchedRules: make([]*model.MatchedRule, 0, len(in.MatchedRules)),
-		ImageTags:    in.ImageTags,
+		NodeBase:     NewNodeBase(),
```
spikat, line 375: The decoding part for the BaseNode is missing for: IMDS, NetworkDevice, DNS, Files and Syscalls nodes.

pkg/security/security_profile/activity_tree/syscalls_node.go
sarrazaghbib-dev: I agree I'd make the NodeBase EvictImageTag() have the same behaviour as the nodes' evictImageTag() and only keep proper evictImageTag() for dns and socket nodes where there's a more processing to do

pkg/security/security_profile/activity_tree/activity_tree.go
sarrazaghbib-dev: done

pkg/security/security_profile/activity_tree/file_node.go
sarrazaghbib-dev: removed it

safchain (approved):

pkg/security/security_profile/activity_tree/base_node.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	b.Seen[imageTag] = &ImageTagTimes{FirstSeen: timestamp, LastSeen: timestamp}
+}
+
+// RecordWithTimestamps sets both FirstSeen and LastSeen for the given imageTag with the provided timestamps.
+func (b *NodeBase) RecordWithTimestamps(imageTag string, firstSeen, lastSeen time.Time) {
+	b.Seen[imageTag] = &ImageTagTimes{FirstSeen: firstSeen, LastSeen: lastSeen}
+}
+
+// EvictImageTag removes the stored timestamps for an imageTag returns true if the imageTag was present
```
spikat: the comment is not accurate

spikat (approved):

PR #38101, opened 2025-06-20
[dyninst/gosym] Implement symbolication using pclntab

piob-io:
### Motivation

https://datadoghq.atlassian.net/browse/DEBUG-3994

Using pclntab is more efficient and more portable than using dwarf.
Use our own pclntab parser, ported from side-eye. It's much more efficient than golang/gosym, and supports inlined tree.
Mmap relevent elf sections, here we don't have to worry about compression.

Will plug this into decoding and the integration test once #37952 gets merged.

```
BenchmarkParseGoSymbolTable/ParseModuleData
103011             11144 ns/op             256 B/op          5 allocs/op
BenchmarkParseGoSymbolTable/ParseGoSymbolTable
7364719               162.3 ns/op           436 B/op          4 allocs/op
BenchmarkParseGoSymbolTable/ListFunctions
4186            284563 ns/op          276095 B/op       9827 allocs/op
BenchmarkParseGoSymbolTable/LocatePC
949910              1223 ns/op             683 B/op         26 allocs/op
```


usamasaqib (approved):

pkg/dyninst/gosym/symtab.go
```diff
@@ -0,0 +1,1187 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+//go:build linux_bpf
+
+// Package gosym provides SymbolTable and PcLnTable utilities.
```
ajwerner: nit: note that it's inspired by the other gosym library but say why it's better. Probably also include a link

pkg/dyninst/gosym/symtab.go
```diff
@@ -0,0 +1,1187 @@
...
+type GoLocation struct {
+	// The function name.
+	Function string
+	// The source file
+	File string
+	// The source line number
+	Line uint32
+}
+
+// GoSymbolTable represents a Go symbol table consisting of function symbols
```
ajwerner: uber-nit: periods at the end of sentences here and elsewhere

pkg/dyninst/object/mmap.go
```diff
@@ -0,0 +1,96 @@
...
+		syscall.PROT_READ,
+		syscall.MAP_SHARED,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to mmap section: %w", err)
+	}
+	md := &MMappedData{
+		Data:   mmaped[offsetDelta:],
+		mmaped: mmaped,
+	}
```
ajwerner, line 83: I looked at the impl of how the Munmap impl works and interestingly it does it based on the end of the mapped range, so you can safely call it with the slice after you truncated. Not that it matters... https://cs.opensource.google/go/go/+/master:src/syscall/syscall_unix.go;l=77-78;drc=49cdf0c42e320dfed044baa551610f081eafb781

pkg/dyninst/gosym/symtab.go
```diff
@@ -0,0 +1,1187 @@
...
+		}, nil
+
+	// --- Go 1.2 ---
+	case ver12:
+		var nfunctab uint32
+		if ptrSize == 8 {
+			if len(data) < 8+8 {
+				return nil, fmt.Errorf("pclntab too short for nfunctab")
+			}
+			nfunctab = uint32(binary.LittleEndian.Uint64(data[8 : 8+8]))
```
ajwerner, line 399: `8+8`?

pkg/dyninst/gosym/symtab.go
```diff
@@ -0,0 +1,1187 @@
...
+		if pc < midPC {
+			right = mid
+		} else {
+			left = mid + 1
+		}
+	}
+
+	if left == 0 {
+		return nil
+	}
```
ajwerner: this is fine, but I think go's [sort.Search](https://pkg.go.dev/sort#Search) interface is reasonably useful here.

pkg/dyninst/gosym/symtab.go
```diff
@@ -0,0 +1,1187 @@
...
+	return funcOff, nil
+}
+
+// funcInfo represents single function data in the pclntab.
+type funcInfo struct {
+	lt   *lineTable
+	data []byte
+}
+
+func (fi *funcInfo) field(n uint32) *uint32 {
```
ajwerner: nit: I think this pointer to uint32 is excessive -- let's just do `(uint32, bool)`

ajwerner (approved):
LGTM

pkg/dyninst/gosym/symtab.go
piob-io: Done

pkg/dyninst/gosym/symtab.go
piob-io: Done

pkg/dyninst/object/mmap.go
piob-io: Interesting, gtk. I will keep this as is - the data field is public and can be further subsliced by the consumer for convenience.

pkg/dyninst/gosym/symtab.go
piob-io: I'll keep it to highlight the underlying byte layout.

pkg/dyninst/gosym/symtab.go
piob-io: sort.Search doesn't support functions that return error. I could panic and recover, or maybe some other workaround, but I think this code is simplest.

pkg/dyninst/gosym/symtab.go
piob-io: Got rid of almost all of *(u)int uses, left two *uint fields in the inlinedCall struct.

piob-io:
Thank you!

PR #38103, opened 2025-06-20
Fix the static quality gate exception invoke task

Pythyu:
### What does this PR do?

This PR fix the `quality-gates.exception-threshold-bump` invoke task which was using `filename` as both a filename and a path. This PR adds `report_path` to fill that role.

Tested on the latest exception request.

### Additional Notes

Introduced a [documentation](https://datadoghq.atlassian.net/wiki/spaces/ABLD/pages/5224269457/Static+quality+gates+exception+process) along side this PR to explicit the process of using this task.

### DIFF
--- a/tasks/static_quality_gates/lib/gates_lib.py
+++ b/tasks/static_quality_gates/lib/gates_lib.py
@@ -168,7 +168,9 @@ class GateMetricHandler:
             )
         return None
 
-    def generate_relative_size(self, ctx, filename="static_gate_report.json", report_path="static_gate_report.json", ancestor=None):
+    def generate_relative_size(
+        self, ctx, filename="static_gate_report.json", report_path="static_gate_report.json", ancestor=None
+    ):
         if ancestor:
             # Fetch the ancestor's static quality gates report json file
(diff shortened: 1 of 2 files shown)


rdesgroppes (approved):

tasks/static_quality_gates/lib/gates_lib.py
```diff
@@ -167,17 +168,19 @@ def _add_gauge(self, timestamp, common_tags, gate, metric_name, metric_key):
             )
         return None
 
-    def generate_relative_size(self, ctx, filename="static_gate_report.json", ancestor=None):
+    def generate_relative_size(
+        self, ctx, filename="static_gate_report.json", report_path="static_gate_report.json", ancestor=None
```
alopezz, line 172: Non-blocking, specially if this has any sort of urgency but could we make it clearer how these parameters differ? Perhaps name them differently?

Also how is this `filename` parameter different from the one passed to this object's constructor?

Pythyu: I'll precise on a second PR

PR #38108, opened 2025-06-23
usm: move is tls static tags

guyarb:
### What does this PR do?

Moves USM static tags management to usm package

### Motivation

We want to make usm internal metrics aware of the tls library it came from

### Additional Notes

Recommended to review the PR commit be commit


usamasaqib (approved):

pkg/network/protocols/tls/tags.go
```diff
@@ -0,0 +1,135 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2024-present Datadog, Inc.
```
amitslavin, line 4: Fix the year

amitslavin (approved):

pkg/network/protocols/tls/tags.go
guyarb: the year is still 2024, I just renamed the file

akarpz (approved):

PR #38112, opened 2025-06-23
[SYNTH-19906] Add support for icmp linux on IPv4 and IPv6

Fabrice-Matrat:
### What does this PR do?
This PR adds a ICMP traceroute driver for IPv4 and IPv6
### Motivation
This PR adds IPv6/IPv4 support on icmp for the existing traceroute methods

1. Build for an ec2 box with IPv6 enabled:
```
GOARCH=amd64 go build github.com/DataDog/datadog-agent/pkg/networkpath/traceroute/icmp/portable_icmp
```
2. `scp` it to the ec2 box
3. Traceroute to Google's public DNS servers: `sudo -E ./portable_icmp 2001:4860:4860::8888`

Also this isn't tested like the TCP driver yet

### DIFF
--- a/pkg/networkpath/traceroute/icmp/icmp_driver.go
+++ b/pkg/networkpath/traceroute/icmp/icmp_driver.go
@@ -1,0 +1,306 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-present Datadog, Inc.
+
+package icmp
+
+import (
+	"encoding/binary"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"golang.org/x/net/icmp"
+	"golang.org/x/net/ipv4"
+	"net/netip"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/DataDog/datadog-agent/pkg/networkpath/traceroute/common"
+	"github.com/DataDog/datadog-agent/pkg/networkpath/traceroute/packets"
+	"github.com/DataDog/datadog-agent/pkg/util/log"
+	"github.com/google/gopacket"
+	"github.com/google/gopacket/layers"
+)
+
+//nolint:unused // This is used, but not on all platforms yet
+var curEchoID atomic.Uint32
+
+//nolint:unused // This is used, but not on all platforms yet
+func nextEchoID() uint16 {
+	next := curEchoID.Add(1)
+	return uint16(next)
+}
+
+type icmpDriver struct {
+	sink packets.Sink
+
+	source packets.Source
+	buffer []byte
+	parser *packets.FrameParser
+	// mu guards against concurrent access to sentProbes
+	mu         sync.Mutex
+	sentProbes map[uint8]time.Time
+	localAddr  netip.Addr
+	params     Params
+	echoID     uint16
+	isIPV6     bool
+}
+
+//nolint:unused // This is used, but not on all platforms yet
+func newICMPDriver(params Params, localAddr netip.Addr, sink packets.Sink, source packets.Source) (*icmpDriver, error) {
+	retval := &icmpDriver{
+		sink:       sink,
+		source:     source,
+		buffer:     make([]byte, 1024),
+		parser:     packets.NewFrameParser(),
+		sentProbes: make(map[uint8]time.Time),
+		localAddr:  localAddr,
+		params:     params,
+		echoID:     nextEchoID(),
+		isIPV6:     localAddr.Is6(),
+	}
+	return retval, nil
+}
+
+func (s *icmpDriver) Close() {
+	s.source.Close()
+	s.sink.Close()
+}
+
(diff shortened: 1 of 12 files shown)


pkg/networkpath/traceroute/runner/runner.go
```diff
@@ -147,6 +148,13 @@ func (r *Runner) RunTraceroute(ctx context.Context, cfg config.Config) (payload.
 			tracerouteRunnerTelemetry.failedRuns.Inc()
 			return payload.NetworkPath{}, err
 		}
+	case payload.ProtocolICMP:
+		log.Tracef("Running ICMP traceroute for: %+v", cfg)
+		pathResult, err = r.runICMP(cfg, hname, dest, maxTTL, timeout)
+		if err != nil {
+			tracerouteRunnerTelemetry.failedRuns.Inc()
```
hmahmood, line 155: Do we want to add a tag for the type of traceroute run? Same for the `runs` metric.

pkg/networkpath/traceroute/common/common.go
```diff
@@ -73,7 +73,13 @@ func LocalAddrForHost(destIP net.IP, destPort uint16) (*net.UDPAddr, net.Conn, e
 	// this is a quick way to get the local address for connecting to the host
 	// using UDP as the network type to avoid actually creating a connection to
 	// the host, just get the OS to give us a local IP and local ephemeral port
-	conn, err := net.Dial("udp", net.JoinHostPort(destIP.String(), strconv.Itoa(int(destPort))))
+	var proto string
+	if destIP.To4() != nil {
+		proto = "udp4"
```
pimlu: Does plain "udp" not work here? The docs suggest it will auto select udp4 or udp6 https://pkg.go.dev/net#Dial

pkg/networkpath/traceroute/icmp/icmp_driver.go
```diff
@@ -0,0 +1,274 @@
...
+func newICMPDriver(params Params, localAddr netip.Addr, sink packets.Sink, source packets.Source) (*icmpDriver, error) {
+	retval := &icmpDriver{
+		sink:       sink,
+		source:     source,
+		buffer:     make([]byte, 1024),
+		parser:     packets.NewFrameParser(),
+		sentProbes: make(map[uint8]time.Time),
+		localAddr:  localAddr,
+		params:     params,
+		echoID:     uint16(os.Getpid() & 0xffff),
```
pimlu: This 0xffff method is based off prior art of traceroute tools I presume. It works for the CLI, since every time you run `traceroute` it makes a new PID. However, for the Datadog agent, which is a long-running daemon, the PID will always be the same.

One way to globally allocate echoIDs is to do

```
var curEchoID atomic.Uint32

func NextEchoID() uint16 {
	next := curPacketID.Add(1)
	return uint16(next)
}
```

pkg/networkpath/traceroute/icmp/icmp_packet.go
```diff
@@ -0,0 +1,91 @@
...
+	ipPair packets.IPPair
+}
+
+func (s *icmpPacketGen) generatePacketV4(ttl uint8, echoID uint16) (*layers.IPv4, *layers.ICMPv4, error) {
+	ipLayer := &layers.IPv4{
+		Version:  4,
+		TTL:      ttl,
+		SrcIP:    s.ipPair.SrcAddr.AsSlice(),
+		DstIP:    s.ipPair.DstAddr.AsSlice(),
+		Id:       uint16(os.Getpid() & 0xffff),
```
pimlu: Similar comment as [here](https://github.com/DataDog/datadog-agent/pull/38112/files#r2162538498), although it might be fine to leave this unchanged, since EchoID should already be unique. Ideally, we could randomize packet ID and check TTL exceeded packets in `handleProbeLayers` just in case, similar to existing traceroute tools.

pkg/networkpath/traceroute/runner/runner.go
Fabrice-Matrat: I think when we will extract the code to [datadog-traceroute](https://github.com/DataDog/datadog-traceroute) we will change where telemetry is used to not have dependency on it. So might be better to do it at that point

pkg/networkpath/traceroute/icmp/icmp_driver.go
Fabrice-Matrat: Done

pkg/networkpath/traceroute/icmp/icmp_packet.go
Fabrice-Matrat: we hardcode it also in udp driver, and a bit harder for IPv6 anyway so will leave it as is for the moment or I could use the same echoID just to have something random but no check yet on the handleProbeLayers

pkg/networkpath/traceroute/common/common.go
Fabrice-Matrat: done

pkg/networkpath/traceroute/icmp/icmp_packet.go
pimlu: Okay, that's fine then

pimlu (approved):

PR #38128, opened 2025-06-23
[AGENTRUN-502] Create non-scrubbed list to avoid hardcoded ignored list in flare intake

louis-cqrl:
### What does this PR do?

This pull request introduces an automated approach to managing non-scrubbed files during the Flare intake process. Instead of relying on a hardcoded ignore list in the backend, this update generates and maintains a JSON file containing all non-scrubbed files. This ensures that only non-text files are excluded from scrubbing dynamically, improving maintainability and reducing the risk of errors caused by manual updates.

### Motivation

Avoid to scrub profiles

Generated a flare manually: `agent flare --profile 30`

### Additional Notes

Log file are present in the list since logs are scrubbed at runtime, however all log files will still be scrubbed on the intake side.

### DIFF
--- a/comp/core/flare/helpers/builder.go
+++ b/comp/core/flare/helpers/builder.go
@@ -160,7 +160,7 @@ func (fb *builder) Save() (string, error) {
 		fb.Lock()
 		defer fb.Unlock()
 		if len(fb.nonScrubbedFiles) == 0 {
-			return []byte(`{"files": [], "message": "No files were added without scrubbing"}`), nil
+			return []byte(`{"files": [], "message": "All files were scrubbed"}`), nil
 		}
 
 		// Convert map keys to slice for JSON output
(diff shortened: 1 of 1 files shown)


rahulkaukuntla (approved):
LGTM!

comp/core/flare/helpers/builder.go
```diff
@@ -151,6 +156,33 @@ func (fb *builder) Save() (string, error) {
 		return fb.permsInfos.commit()
 	})
 
+	_ = fb.AddFileFromFunc("non_scrubbed_files.json", func() ([]byte, error) {
+		fb.Lock()
+		defer fb.Unlock()
+		if len(fb.nonScrubbedFiles) == 0 {
+			return []byte(`{"files": [], "message": "No files were added without scrubbing"}`), nil
```
sgnn7: Drive by
```suggestion
			return []byte(`{"files": [], "message": "All files were scrubbed"}`), nil
```

PR #38147, opened 2025-06-23
dyninst/output: improve and test events

ajwerner:
There were a couple of cases we didn't handle correctly. Along the way I rename the file to just event.go and add testing to cover all the cases.

Relates to https://datadoghq.atlassian.net/browse/DEBUG-3889.


piob-io (approved):
Lg, thanks!

PR #38149, opened 2025-06-23
Go DI: Add stack hashes map

grantseltzer:
### What does this PR do?

Adds a map for resolve stack trace hashes to actual stack traces.

### Motivation

bpf will only upload unique stack traces once. 

Integration test

### DIFF
--- a/pkg/dyninst/decode/decoder.go
+++ b/pkg/dyninst/decode/decoder.go
@@ -31,6 +31,7 @@ type probeEvent struct {
 // It is not guaranteed to be thread-safe.
 type Decoder struct {
 	program     *ir.Program
+	stackHashes map[uint64][]uint64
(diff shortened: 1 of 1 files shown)


ajwerner (approved):
This is fine for now. Once we hook up the symbolicator (which should happen soon), perhaps it'll make sense to cache the symbolicated stack trace instead of the raw PCs. I look forward to seeing this in the testing. Perhaps we'll want to find ways to exercise probes being triggered more than once

piob-io:
> This is fine for now. Once we hook up the symbolicator (which should happen soon), perhaps it'll make sense to cache the symbolicated stack trace instead of the raw PCs. I look forward to seeing this in the testing. Perhaps we'll want to find ways to exercise probes being triggered more than once

Yup, I will hook up the symbolicator.

PR #38159, opened 2025-06-23
[pkg/dyninst] Implement loclist parsing

piob-io:
### Motivation

https://datadoghq.atlassian.net/browse/DEBUG-3989

To avoid evaluating dwarf stack expression, and better handle go bugs.

### DIFF
--- a/pkg/dyninst/compiler/sm/ops.go
+++ b/pkg/dyninst/compiler/sm/ops.go
@@ -63,7 +63,7 @@ type ExprSaveOp struct {
 
 type ExprDereferenceCfaOp struct {
 	baseOp
-	Offset       uint32
+	Offset       int32
 	Len          uint32
 	OutputOffset uint32
 }
--- a/pkg/dyninst/dwarf/loclist/loclist.go
+++ b/pkg/dyninst/dwarf/loclist/loclist.go
@@ -1,0 +1,368 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+//go:build linux_bpf
+
+// Package loclist supports parsing DWARF loclist instructions.
+package loclist
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"math"
+
+	"github.com/DataDog/datadog-agent/pkg/dyninst/ir"
+)
+
+// DWARF opcodes
+// https://dwarfstd.org/doc/DWARF5.pdf
+// All range constants are inclusive on both ends.
+//
+//revive:disable:var-naming
+const (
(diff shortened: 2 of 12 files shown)


pkg/dyninst/dwarf/loclist/loclist.go
```diff
@@ -0,0 +1,366 @@
...
+	}
+
+	// We want to find any pair of adjacent pieces that are in the same location
+	for currentByteSize(loclists) > expectedByteSize {
+		if len(loclists) == 0 || len(loclists[0].Pieces) == 0 {
+			break
+		}
+
+		numPieces := len(loclists[0].Pieces)
+		var foundIndex *int
```
ajwerner: nit: I recognize that these indexes are small and go has that clever little hack to avoid allocating small integers, but I still think it'd be more idiomatic to have a boolean and make this just a regular old `int`.

pkg/dyninst/compiler/sm/ops.go
```diff
@@ -63,7 +63,7 @@ type ExprSaveOp struct {
 
 type ExprDereferenceCfaOp struct {
 	baseOp
-	Offset       uint32
+	Offset       int32
```
ajwerner, line 66: We'll need local variables and line probes to exercise this I imagine. Good catch.

pkg/dyninst/ir/location.go
```diff
@@ -0,0 +1,52 @@
...
+// Cfa represents DW_OP_fbreg and DW_OP_call_frame_cfa.
+type Cfa struct {
+	Offset int32
+}
+
+func (c Cfa) pieceOp() {}
+
+// Register represents DW_OP_reg* and DW_OP_breg*.
+type Register struct {
+	Idx   uint8
```
ajwerner: nit: this field name doesn't render particularly nicely in the output. What about `RegNum`, `RegNo`, or even `Register`?

ajwerner (approved):
Cool! I guess if there's bugs anywhere we'll find out pretty quickly when things don't work. Regarding testing, I could imagine some sort of oracle test that sees what delve's code decodes and whether they disagree, but that code has an annoying interface so certainly not blocking.

pkg/dyninst/dwarf/loclist/loclist.go
piob-io: Done, it was all ai :P

pkg/dyninst/compiler/sm/ops.go
piob-io: Actually, this is purely cosmetic - both propagate the bits all the same that stay as-is until they get used in stack machine. It matters that there we use signed int, and we do. The tests cover the negative cases, we need line probes for positive ones :upside_down_face:

pkg/dyninst/ir/location.go
piob-io: Ah, indeed. RegNo.

ajwerner (approved):

PR #38170, opened 2025-06-24
dyninst: fix frameless function CFA computation

ajwerner:
This patch does two things:
1) It propagates the frameless nature of functions.
2) It appropriately initializes the CFA based on that frameless nature.

Go leaf functions that don't use stack space do not have a prologue. We
handled this computation for deciding where to put the probe, but we
didn't translate the CFA offset calculation. Note that this fix remains
incomplete in that it's possible for inlined functions to appear underneath
a frameless parent. Plumbing for that calculation is left for later.

Resolves [DEBUG-4096](https://datadoghq.atlassian.net/browse/DEBUG-4096)

[DEBUG-4096]: https://datadoghq.atlassian.net/browse/DEBUG-4096?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/dyninst/ebpf/event.c
+++ b/pkg/dyninst/ebpf/event.c
@@ -78,6 +78,8 @@ SEC("uprobe") int probe_run_with_cookie(struct pt_regs* regs) {
   global_ctx.stack_walk->regs = *regs;
   global_ctx.stack_walk->stack.pcs.pcs[0] = regs->DWARF_PC_REG;
   global_ctx.stack_walk->stack.fps[0] = regs->DWARF_BP_REG;
+  // TODO: For arm64 we need to use the link register to properly set up the
+  // stack trace.
   if (params->frameless) {
     if (bpf_probe_read_user(&global_ctx.stack_walk->stack.pcs.pcs[1],
                             sizeof(global_ctx.stack_walk->stack.pcs.pcs[1]),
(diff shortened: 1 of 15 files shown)


pkg/dyninst/ebpf/event.c
```diff
@@ -109,18 +111,45 @@ SEC("uprobe") int probe_run_with_cookie(struct pt_regs* regs) {
...
+// the beginning of the prologue adjustment and can assume the registers are
+// still set up for the caller's frame.
+//
+// See https://github.com/golang/go/issues/74357.
+    *(volatile uint64_t*)(&frame_data.cfa) = global_ctx.regs->DWARF_BP_REG + 8;
 #elif defined(bpf_target_x86)
-    .cfa = global_ctx.regs->DWARF_BP_REG + 16,
+// On x86, if the function is frameless, the stack pointer is pointing to the
+// return pc, so one word above that is our CFA. If it's not frameless, then
+// the base pointer is pointing to our frame pointer which is 16 bytes above
```
piob-io: below (since we need to add to it, to point higher)?

pkg/dyninst/ebpf/event.c
```diff
@@ -109,18 +111,45 @@ SEC("uprobe") int probe_run_with_cookie(struct pt_regs* regs) {
...
+// our CFA.stack_walk->stack.fps[1] + 8;
```
piob-io: nitcruft

pkg/dyninst/ir/program.go
```diff
@@ -131,6 +131,8 @@ type Event struct {
 	Type *EventRootType
 	// The PC values at which the event should be injected.
 	InjectionPCs []uint64
```
piob-io: I think it's nicer to have a slice of structs, with two fields, pc and frameless, rather than promise that the two slices will always have the same length. We can call it InjectionPoint? InjectionTarget? Injection?

piob-io (approved):
Thank you!

pkg/dyninst/ir/program.go
ajwerner: Done.

pkg/dyninst/ebpf/event.c
ajwerner: Updated.

pkg/dyninst/ebpf/event.c
ajwerner: Fixed

ajwerner (commented):
TFTR!

ajwerner:
See https://github.com/golang/go/issues/74357

PR #36383, opened 2025-04-22
[omnibus][AGENTCFG-242] Adding the sgc binary to omnibus

rahulkaukuntla:
### What does this PR do?
We want to add the [datadog-secret-backend](https://github.com/DataDog/datadog-secret-backend/tree/main) as a side binary to the agent.

PR #38181, opened 2025-06-24
Remove secret binary from heroku agent package

pgimalac:
### What does this PR do?
Remove the secret binary added in https://github.com/DataDog/datadog-agent/pull/36383 from the heroku agent package.

### Motivation
The size increase is not considered worth it for Heroku, and most functionalities from the secret binary don't apply (eg. fetch secret from k8s or cloud providers).

### DIFF
--- a/omnibus/config/software/datadog-agent-dependencies.rb
+++ b/omnibus/config/software/datadog-agent-dependencies.rb
@@ -31,7 +31,7 @@ dependency 'libpcap' if linux_target? and !heroku_target? # system-probe depende
 # Include traps db file in snmp.d/traps_db/
 dependency 'snmp-traps'
 
-dependency 'secret-generic-connector'
+dependency 'secret-generic-connector' if !heroku_target?
 
 dependency 'datadog-agent-integrations-py3'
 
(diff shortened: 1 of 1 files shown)


omnibus/config/software/datadog-agent-dependencies.rb
```diff
@@ -31,7 +31,7 @@
 # Include traps db file in snmp.d/traps_db/
 dependency 'snmp-traps'
 
-dependency 'secret-generic-connector'
+dependency 'secret-generic-connector' if !heroku_target?
```
chouquette:
```suggestion
dependency 'secret-generic-connector' unless heroku_target?
```

chouquette: otherwise I think rubocop or similar will complain sooner or later. That being said, that's definitely a nit and non blocking

chouquette (approved):

omnibus/config/software/datadog-agent-dependencies.rb
pgimalac: Too late I already accepted ;)

PR #38187, opened 2025-06-24
wlm: process: allow SD collector to work when process collection is disabled

Yumasi:
### What does this PR do?

This PR allows the process collector to run when process collection is disabled, only collecting service information and storing Service Discovery only fields.

### Motivation

Handle the case where process collection is disabled.

Unit tests are included, and were run, but are skipped until the configuration is no longer hardcoded.

The change was also tested manually by:
- Changing the return value of isServiceDiscoveryEnabled to true
- Run processes that look like a service: Python HTTP server, postgres, rails server.
- Running the agent and system-probe, and run `agent workload-list`. Only service information should be present on the output. Example:

```
vagrant@agent-dev-ubuntu-22:/git/datadog-agent$ ./bin/agent/agent workload-list -c ./dev/dist/datadog.yaml --sysprobecfgpath ./dev/dist/system-probe.yaml | rg -A 10 'Entity process'
=== Entity process sources(merged):[service_discovery] id: 563793 ===
----------- Entity ID -----------
PID: 563793
Namespace PID: 0
Container ID:
Creation time: 2025-06-24 12:43:52 +0000 UTC
Service Generated Name: agent
===

=== Entity process sources(merged):[service_discovery] id: 921 ===
----------- Entity ID -----------
PID: 921
Namespace PID: 0
Container ID:
Creation time: 2025-06-22 20:42:10 +0000 UTC
Service Generated Name: postgres
===

=== Entity process sources(merged):[service_discovery] id: 993737 ===
----------- Entity ID -----------
PID: 993737
Namespace PID: 0
Container ID:
Creation time: 2025-06-23 12:12:39 +0000 UTC
Service Generated Name: rails_hello
===

=== Entity process sources(merged):[service_discovery] id: 1346514 ===
----------- Entity ID -----------
PID: 1346514
Namespace PID: 0
Container ID:
Creation time: 2025-06-24 01:57:43 +0000 UTC
Service Generated Name: node
===
```


wdhif (approved):

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -154,23 +160,29 @@ func (c *collector) collectionIntervalConfig() time.Duration {
...
-		c.store = store
+		c.containerProvider = containerProvider
+	}
+	c.store = store
+
+	if c.isProcessCollectionEnabled() {
 		go c.collectProcesses(ctx, c.clock.Ticker(c.collectionIntervalConfig()))
+	}
+
+	if c.isServiceDiscoveryEnabled() {
```
gengnamstyle: Could we separate the 2 cases of service collection when process is on/off into it's own goroutine and add another check here to run that specifically? My first thoughts on naming:
- `collectServicesCached`
- `collectServicesDefault`

for example:

```
if c.isProcessCollectionEnabled() {
	go c.collectProcesses(ctx, c.clock.Ticker(c.collectionIntervalConfig()))
}

if c.isServiceDiscoveryEnabled()  && c.isProcessCollectionEnabled() {
	if c.isProcessCollectionEnabled() {
		go c.collectServicesCached(ctx, c.clock.Ticker(serviceCollectionInterval))
	} else {
		go c.collectServicesDefault(ctx, c.clock.Ticker(serviceCollectionInterval))
	}
}
```

My reasoning is that the main functions in `collectServices`
- `getProcessDataForServices`
- `updateServices`
- `findDeletedProcesses`

vary quite a bit when process collection is on/off.

I think this could make the conditions more distinct and make testing a bit easier! I'm open to thoughts though!

Yumasi: I don't have a particularly strong opinion on this. I've tried to apply your suggestion, trying to limit the code duplication of doing so, let me know what you think. The one upside though, to me, is that it allows removing the test skip.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -468,30 +576,62 @@ func (c *collector) collectProcesses(ctx context.Context, collectionTicker *cloc
...
+			if err != nil {
+				log.Errorf("Error getting processes for service discovery: %v", err)
 				continue
 			}
+			if len(alivePids) == 0 {
+				continue // no processes to check
+			}
+
+			wlmServiceEntities := c.updateServicesNoCache(alivePids, procs)
+			deletedProcesses := c.findDeletedProcesses(alivePids)
```
gengnamstyle: When process collection is disabled, I'm wondering if we can apply what we did for the process collector to find the deleted processes mainly for the sake of consistency.

I've heard that querying wlm is not the most performant if we can avoid it, but I haven't benchmarked.

i.e.

```
deletedProcs := processCacheDifference(c.lastCollectedProcesses, procs)
wlmDeletedProcs := deletedProcessesToWorkloadmetaProcesses(deletedProcs)

...


c.lastCollectedProcesses = procs
```

I do prefer your naming of `findDeletedProcesses`, so perhaps the 2 functions could go in there and the `collectProcesses` could also use this :)

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -377,14 +394,84 @@ func (c *collector) updateServices(alivePids core.PidSet) []*workloadmeta.Proces
...
+	// Only detect languages for services when process collection is disabled,
+	// otherwise the collectProcesses goroutine already did it for us.
+	var pidToLanguage map[int32]*languagemodels.Language
+	serviceProcs := make([]*procutil.Process, 0, len(pidsToService))
+	for pid := range pidsToService {
+		if proc, exists := procs[pid]; exists {
+			serviceProcs = append(serviceProcs, proc)
+		}
+	}
+	languages := c.detectLanguages(serviceProcs)
```
daniel-taf, line 419: This will only return languages if the language detection feature is enabled. Does that align well enough when service discovery needs it?

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -468,30 +576,62 @@ func (c *collector) collectProcesses(ctx context.Context, collectionTicker *cloc
...
+				c.processEventsCh <- &Event{
+					Type:    EventTypeServiceDiscovery,
+					Created: wlmServiceEntities,
+					Deleted: deletedProcesses,
+				}
+			}
+
+			cleanPidMaps(alivePids, c.ignoredPids)
+			cleanPidMaps(alivePids, c.serviceRetries)
+			cleanPidMaps(alivePids, c.pidHeartbeats)
```
daniel-taf: nit: since this function is used this way in both collection functions. I wonder if we can consolidate the uses into one function.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
Yumasi: Hmm, I did not know that. Do you think it would be possible to make service discovery enable language detection with it?

daniel-taf: FWIW, its enabled by default, so it may be good enough? If not, we could likely enable the system probe module if service discovery is enabled.

daniel-taf (approved):

PR #38191, opened 2025-06-24
Push apttesting artifact on a dedicated repo

KevinFairise2:
### What does this PR do?

Use `--prefix` of `deb-s3` to upload artifacts for each pipeline on a dedicated folder. It has several advantages:
- Repo are separated so corruption of the shared `pools` should no longer be able to break everything
- Make cleanup way easier, as of today we cannot easily automatically clean up the content of `pools` on the repository because all the pipelines use the same and if there are missing artifacts it will corrupt the whole repo.

### Motivation

Make clean up easier


spencergilbert (approved):

BaptisteFoy (approved):
LGTM for Fleet-owned files

.gitlab/e2e_testing_deploy/e2e_deploy.yml
```diff
@@ -64,10 +64,10 @@ deploy_deb_testing-a7_x64:
...
-    - echo "$APT_SIGNING_KEY_PASSPHRASE" | deb-s3 upload -c "pipeline-$DD_PIPELINE_ID-x86_64" -m 7 -b $DEB_TESTING_S3_BUCKET -a amd64 --sign=$DEB_GPG_KEY_ID --gpg_options="--passphrase-fd 0 --batch --digest-algo SHA512" --preserve_versions --visibility public $OMNIBUS_PACKAGE_DIR/datadog-signing-keys_${DD_PIPELINE_ID}.deb
-    - echo "$APT_SIGNING_KEY_PASSPHRASE" | deb-s3 upload -c "pipeline-$DD_PIPELINE_ID-x86_64" -m 7 -b $DEB_TESTING_S3_BUCKET -a x86_64 --sign=$DEB_GPG_KEY_ID --gpg_options="--passphrase-fd 0 --batch --digest-algo SHA512" --preserve_versions --visibility public $OMNIBUS_PACKAGE_DIR/datadog-signing-keys_${DD_PIPELINE_ID}.deb
+    - echo "$APT_SIGNING_KEY_PASSPHRASE" | deb-s3 upload -c "stable-x86_64" -m 7 -b $DEB_TESTING_S3_BUCKET -a amd64 --sign=$DEB_GPG_KEY_ID --gpg_options="--passphrase-fd 0 --batch --digest-algo SHA512" --preserve_versions --visibility public --prefix datadog-agent/pipeline-${DD_PIPELINE_ID} $OMNIBUS_PACKAGE_DIR/datadog-*_7*amd64.deb
```
chouetz, line 68: Should it be `testing` or `dev` instead of `stable`?

chouetz (approved):

.gitlab/e2e_testing_deploy/e2e_deploy.yml
KevinFairise2: Actually I followed the same convention as in `fips-proxy`. With a dedicated testing repo, and a `stable` branch in that testing repo.
https://github.com/DataDog/fips-proxy/blob/main/.gitlab/kitchen_promote/deb.yml#L34
In the end it does not change anything but I can update it if you think it would be clearer

aboitreaud (approved):

rdesgroppes (approved):

.gitlab/e2e_testing_deploy/e2e_deploy.yml
chouetz: I agree this does not change nothing, and I agree it would probably be better to push in `stable` by default on all and every repo. In the current situation I think it can be confusing to know where your artefact is supposed to be during its lifecycle :)

PR #38199, opened 2025-06-24
[AGENTRUN-337] Refactor diskv2 e2e test

jeremy-hanna:
### What does this PR do?

This is a refactor for the diskv2 check e2e test suite to make a lot of the check comparison logic a common util so we can implement the same test for the networkv2 check work in https://github.com/DataDog/datadog-agent/pull/38056.

There is a missing bit of logic with the linux networkv2 check causing [a failure](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/994800608) with missing metrics so I split the refactor PR out to fix that bug separately from this work.


pgimalac (approved):
LGTM if the tests still pass

test/new-e2e/tests/agent-runtimes/checks/disk/common.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (v *diskCheckSuite) getSuiteOptions() []e2e.SuiteOption {
+	suiteOptions := []e2e.SuiteOption{}
+	suiteOptions = append(suiteOptions, e2e.WithProvisioner(
+		awshost.Provisioner(
+			awshost.WithEC2InstanceOptions(ec2.WithOS(v.descriptor)),
+		),
+	))
+
+	return suiteOptions
```
misteriaud, line 40: **🥜 nitpick:** You can directly return‏ the option

```suggestion
	return []e2e.SuiteOption{
			e2e.WithProvisioner(
				awshost.Provisioner(
					awshost.WithEC2InstanceOptions(ec2.WithOS(v.descriptor)),
				),
			),
	}
```

misteriaud (commented):
LGTM !

PR #38205, opened 2025-06-24
Avoid silence failure, and reduce CPU to avoid OOMkill

KevinFairise2:
### What does this PR do?

Reduce CPU request to avoid OOMKill on job building test binaries. In the end it does not increase the duration of the build job.
And make sure that some build silently fail, which could make some test not executed.

### DIFF
--- a/tasks/new_e2e_tests.py
+++ b/tasks/new_e2e_tests.py
@@ -150,10 +150,13 @@ def build_binaries(
                     if success:
                         success_count += 1
                         # Store the original package path and binary name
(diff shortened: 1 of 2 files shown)


tasks/new_e2e_tests.py
```diff
@@ -150,11 +150,14 @@ def build_binaries(
                     if success:
                         success_count += 1
                         # Store the original package path and binary name
```
CelianR:
```suggestion
```

CelianR (approved):

PR #33666, opened 2025-02-03
Ensure IPC perm are updated

julien-lebot:
### What does this PR do?
This PR adds the missing `ipc_perm.pem` file to the list of files the installer should update the permission.

### Motivation
All files should have their permissions updated when changing user.

### Describe how you validated your changes

Manual QA + E2E test.

PR #38207, opened 2025-06-25
Preserve install-info from setup script instead of overwriting from MSI on windows

sar-shah:
### What does this PR do?
This PR preserves install-info and doesn't let windows MSI overwrite it when the agent installation/upgrade is triggered via fleet installation setup script.
### Motivation
https://datadoghq.atlassian.net/browse/WINA-1431
Added a new e2e test which validates the install-info isn't overwritten, as was the case previously. The test correctly failed without the new changes and is passing after adding the new changes as expected.

### Additional Notes
We introduce a new flag for MSI custom actions - SKIP_INSTALL_INFO, and if this is set, the MSI installer will skip writing the install-info. We did consider not using this flag and just rely on the presence of existing install-info file, but that has limitations due to the file not deleted as part of rollback flow (https://github.com/DataDog/datadog-agent/pull/33666)


estherk15 (approved):

tools/windows/DatadogAgentInstaller/CustomActions/InstallInfoCustomActions.cs
```diff
@@ -11,6 +11,12 @@ public class InstallInfoCustomActions
     {
         private static ActionResult WriteInstallInfo(ISession session)
         {
+            var skipInstallInfo = session.Property("SKIP_INSTALL_INFO");
+            if (!string.IsNullOrEmpty(skipInstallInfo) && skipInstallInfo == "1")
```
julien-lebot: **🔬 nit:** ‏ I think this can be simplified to `if (skipInstallInfo == "1")`

julien-lebot (approved):

tools/windows/DatadogAgentInstaller/CustomActions/InstallInfoCustomActions.cs
sar-shah: Changed!

chouetz (approved):

PR #38209, opened 2025-06-25
Build images in two steps to control what gets stored in registry cache

alopezz:
### What does this PR do?

This change makes it so that builds where we write to the registry-based remote cache do so on a separate build step before building the full final target; this is the only way we have of controlling what gets stored in the remote cache.

With these changes, jobs can specify a target to build to populate the cache, allowing us to not store unnecessary artifacts in it (mainly the artifacts that are built on the same pipeline and which are uncacheable in practice).

### Motivation

This is what #37483 meant to achieve, but that solution didn't work as intended, and it turned out we were still uploading everything. The motivation from that PR still applies to this one.

[ABLD-55](https://datadoghq.atlassian.net/browse/ABLD-55)

I run pipelines for all the possible conditions that we may have, by setting the relevant variables in the same job:
- commit to main: https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/68586905 (both read from and write to remote cache)
- nightly: https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/68579212 (start without cache and write results to cache, used to refresh the cache daily)
- deploy: https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/68618729 (neither read from nor write to remote cache, ensure that deploy pipelines always get the latest available).

### Additional Notes

- The order of stages in the dockerfiles now becomes relevant, the `--target` argument for multi-stage builds results in building all the stages up to and including the target (regardless of the actual dependency graph). That's the reason for the reordering on the Agent dockerfile.
- I've had to split the original `--target` arguments into a separate variable (from being part of `BUILD_ARG`) to be able to pass the same build args under `BUILD_ARG` to both the cache-populating build as well as the final build. This is probably a good thing regardless.
- I had to add intermediate stages for cluster agent and dogstatsd images to be able to apply caching to those under this new implementation.

[ABLD-55]: https://datadoghq.atlassian.net/browse/ABLD-55?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.gitlab/container_build/docker_linux.yml
+++ b/.gitlab/container_build/docker_linux.yml
@@ -11,25 +11,24 @@
       else
         export ECR_RELEASE_SUFFIX=${CI_COMMIT_TAG+-release}
       fi
+    - AGENT_BASE_IMAGE_TAG=registry.ddbuild.io/ci/datadog-agent/agent-base-image${ECR_RELEASE_SUFFIX}:v${CI_PIPELINE_ID}-${CI_COMMIT_SHORT_SHA}-$ARCH
+    - TARGET_TAG=${IMAGE}${ECR_RELEASE_SUFFIX}:v${CI_PIPELINE_ID}-${CI_COMMIT_SHORT_SHA}$TAG_SUFFIX-$ARCH
+    - !reference [.login_to_docker_readonly]
     # Caching setup
-    - DOCKER_CACHE_TARGET="${IMAGE}${TAG_SUFFIX}-${ARCH}:cache"
-    - CACHE_SOURCE="--cache-from type=registry,ref=${DOCKER_CACHE_TARGET}"
+    - DOCKER_CACHE_REGISTRY_TARGET="${IMAGE}${TAG_SUFFIX}-${ARCH}:cache"
+    - CACHE_SOURCE="--cache-from type=registry,ref=${DOCKER_CACHE_REGISTRY_TARGET}"
     - CACHE_TO=""
-    - |
-      DOCKER_NO_CACHE=""
-      for target in ${NO_CACHE_TARGETS}; do
-        DOCKER_NO_CACHE="${DOCKER_NO_CACHE_FILTER} --no-cache-filter ${target}"
-      done
+    - CACHE_TARGET=${CACHE_TARGET:-}
     # Don't use caching on nightlies, to allow for regular cache invalidation,
     # and update the cache on both main and nightly builds
     - |
       if [[ "$BUCKET_BRANCH" == "nightly" ]]; then
         DOCKER_NO_CACHE="--no-cache"
         CACHE_SOURCE=""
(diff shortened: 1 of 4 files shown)


Dockerfiles/agent/Dockerfile
```diff
@@ -169,6 +127,47 @@ COPY s6-services /etc/services.d/
...
+
+# Extract and cleanup:
+#   - unused systemd unit
+#   - GPL sources for embedded software  # FIXME: move upstream
+#   - docs and manpages                  # FIXME: move upstream
+#   - static libraries                   # FIXME: move upstream
+#   - jmxfetch on nojmx build
+
+RUN --mount=from=artifacts,target=/artifacts \
+ find /artifacts -maxdepth 1 -type f -name "datadog-*-$TARGETARCH.tar.xz" ! -name "$DD_AGENT_ARTIFACT" -exec rm {} \; \
```
rdesgroppes, line 158: why not
```suggestion
 find /artifacts -maxdepth 1 -type f -name "datadog-*-$TARGETARCH.tar.xz" ! -name "$DD_AGENT_ARTIFACT" -delete \
```
? (like a few line below)

Update: my bad: this code was simply _moved_ - resolving.

.gitlab/container_build/docker_linux.yml
```diff
@@ -248,7 +275,8 @@ docker_build_fips_agent7_arm64_jmx:
   variables:
     IMAGE: registry.ddbuild.io/ci/datadog-agent/agent-base-image
     BUILD_CONTEXT: Dockerfiles/base-image
-    BUILD_ARG: --target release
+    TARGET_ARG: --target release
+    CACHE_TARGET: release
```
chouquette, line 279: Do those have an effect? The variables seem to be systematically overridden by jobs extending this base

chouquette (approved):
LGTM with a nit-like question

.gitlab/container_build/docker_linux.yml
alopezz: Hmm you do have a point here. Arguably it would be the overrides that aren't really necessary here. I feel like there's actually quite a bit of refactoring that could be done altogether. I'll merge this as is and open a refactoring PR right after.

Dockerfiles/agent/Dockerfile
alopezz: You're not wrong, but indeed I'm just moving things around so this is out of scope for the topic of this PR.

.gitlab/container_build/docker_linux.yml
alopezz: On a closer look, actually no job extending from this template is overriding these 🤔 .

PR #38219, opened 2025-06-25
Use a dedicated output directory per-pipeline for windows builds

chouquette:
### What does this PR do?

Use a per-pipeline artifact folder for windows jobs.

### Motivation

When a windows job is cancelled, it will actually keep running in the background, causing the node that was running the job to see both the cancelled job & the newly scheduled job's artifacts.

See https://datadoghq.atlassian.net/wiki/spaces/DEVX/pages/3468492851/Windows+Gitlab+runners+overview#Artifacts-from-2-different-Agent-versions-were-uploaded-to-my-job%E2%80%99s-artifacts 

#incident-39868


alopezz (commented):
I think we're probably missing references here (and others in this file):
https://github.com/DataDog/datadog-agent/blob/435a850385f0d3ebda5542d5df735325c6d8a4c8/.gitlab/deploy_packages/windows.yml#L64

Also, technically doesn't change _that_ much I think, but for efficiency and consistency this should probably also include the pipeline id.

https://github.com/DataDog/datadog-agent/blob/435a850385f0d3ebda5542d5df735325c6d8a4c8/.gitlab/package_build/installer.yml#L245

alopezz (approved):

julien-lebot (commented):
I think we are missing a couple of changes in those files:
- .gitlab/choco_build/choco_build.yml
- tasks\winbuild.py

tasks/winbuild.py
```diff
@@ -69,7 +67,10 @@ def installer_package(
         url_safe=True,
         include_pipeline_id=True,
     )
+    output_path = os.path.join(os.getcwd(), "omnibus", "pkg")
+    if 'CI_PIPELINE_ID' in os.environ:
+        output_path = os.path.join(output_path, f'pipeline-{os.getenv("CI_PIPELINE_ID")}')
     shutil.copy2(
         os.path.join(OPT_SOURCE_DIR, "datadog-installer\\datadog-installer.exe"),
-        os.path.join(OUTPUT_PATH, f"datadog-installer-{agent_version}-1-x86_64.exe"),
+        os.path.join(output_path, f"datadog-installer-{agent_version}-1-x86_64.exe"),
```
julien-lebot: The call to `Generate-OCIPackage.ps1` is still:
```
 ctx.run('powershell -C "./tasks/winbuildscripts/Generate-OCIPackage.ps1 -package datadog-agent"')
 ```
and
```
ctx.run('powershell -C "./tasks/winbuildscripts/Generate-OCIPackage.ps1 -package datadog-installer"')
```

By default, this script will look in `omnibus/pkg`, not the per-pipeline directory.
I think you need to pass the new output path as the `-omnibusOutput` argument (or update the `Generate-OCIPackage.ps1`) if you want the script to find the MSI in the correct place.

Comment (chouquette) [reply tasks/winbuild.py]: Indeed, but I'm surprised this didn't cause some test failure

This is now fixed

File: tasks/static_quality_gates/lib/gates_lib.py [hunk 2]
```diff
@@ -85,10 +85,9 @@ def find_package_path(flavor, package_os, arch, extension=None):
     separator = '_' if package_os == 'debian' else '-'
     if not extension:
         extension = "deb" if package_os == 'debian' else "rpm"
-    pipeline_match = ""
-    if package_os == "windows":  # Windows builds are subject to artifacts leak and need their pipeline to match the msi
-        pipeline_match = f"{os.environ['CI_PIPELINE_ID']}-1-"
-    glob_pattern = f'{package_dir}/{flavor}{separator}7*{pipeline_match}{arch}.{extension}'
+    if package_os == "windows":
+        package_dir = f"{package_dir}/pipeline-{os.environ['CI_PIPELINE_ID']}"
+    glob_pattern = f'{package_dir}/{flavor}{separator}7*{arch}.{extension}'
```
CelianR, line 90: I'm not sure to understand, don't we need the `pipeline-XXX/` separator here?

julien-lebot (approved):

tasks/winbuild.py
```diff
@@ -9,12 +9,17 @@
...
-# Output directory for package files
-OUTPUT_PATH = os.path.join(os.getcwd(), "omnibus", "pkg")
 # Omnibus stores files here, e.g. C:\opt\datadog-agent, C:\opt\dataog-installer
 OPT_SOURCE_DIR = os.path.join('C:\\', 'opt')
 
 
+def _get_output_path():
+    output_path = os.path.join(os.getcwd(), "omnibus", "pkg")
+    if 'CI_PIPELINE_ID' in os.environ:
+        output_path = os.path.join(output_path, f'pipeline-{os.getenv("CI_PIPELINE_ID")}')
```
CelianR:
```suggestion
        output_path = os.path.join(output_path, f'pipeline-{os.environ["CI_PIPELINE_ID"]}')
```
I prefer to keep this syntax everywhere even though the variable is checked

CelianR (approved):

tasks/static_quality_gates/lib/gates_lib.py
chouquette: only for windows builds (at least for now), but it's already added just above in the `package_dir` variable

alopezz:
Stopped the merge to double check that https://github.com/DataDog/datadog-agent/pull/38219#pullrequestreview-2961551056 has been addressed.

PR #38222, opened 2025-06-25
[CWS-4746] Implement `open_tree` syscall

kovagsm:
### What does this PR do?

* Implements the `open_tree` syscall to keep track of new mounts (recursive or otherwise) created with this API
* Enrich the FileEvent with mount visibility / detachment info and allow filtering events with it
* Cleanup previous fsmount implementation to reuse code

### Motivation
* To improve mount resolver and reduce the number of abnormal paths

* Integration and manual tests


git-thuerk-done (approved):
Good for docs

pkg/security/ebpf/c/include/constants/offsets/filesystem.h
```diff
@@ -117,7 +117,13 @@ int __attribute__((always_inline)) get_path_mount_flags(struct path *path) {
...
+    bpf_probe_read(&mount_id, sizeof(mount_id), (char *)mnt + get_mount_offset_of_mount_id());
+    return mount_id;
+}
+
+int __attribute__((always_inline)) get_mount_parent(void *mnt) {
+    int mount_id;
+
     bpf_probe_read(&mount_id, sizeof(mount_id), (char *)mnt + get_mount_offset_of_mount_id());
     return mount_id;
 }
```
YoannGh: These two functions are doing the same thing, can we remove one of them?

pkg/security/secl/model/unmarshallers_linux.go
```diff
@@ -466,53 +466,33 @@ func (m *Mount) UnmarshalBinary(data []byte) (int, error) {
...
+func (e *MountEvent) UnmarshalBinary(data []byte) (int, error) {
+	n, err := UnmarshalBinary(data, &e.SyscallEvent, &e.SyscallContext, &e.Mount)
 	if err != nil {
-		return read, err
+		return n, err
 	}
-
-	read += n
 	data = data[n:]
+	origin := binary.NativeEndian.Uint32(data[0:4])
```
YoannGh, line 485: Is it possible to check the bounds of data before reading `origin` to avoid any potential crashes?

pkg/security/ebpf/c/include/hooks/mount.h
```diff
@@ -164,21 +199,26 @@ void __attribute__((always_inline)) handle_new_mount(void *ctx, struct syscall_c
...
 
-    resolve_dentry(ctx, prog_type);
-    // if the tail call fails, we need to pop the syscall cache entry
-    pop_syscall(syscall->type);
+        resolve_dentry(ctx, prog_type);
+
+        // if the tail call fails, we need to pop the syscall cache entry
+        pop_syscall(syscall->type);
+    } else {
+        send_detached_event(ctx, syscall);
```
safchain, line 218: should we pop_syscall anyway ?

pkg/security/ebpf/c/include/hooks/mount.h
```diff
@@ -209,38 +250,42 @@ TAIL_CALL_TRACEPOINT_FNC(dr_mount_stage_one_callback, struct tracepoint_syscalls
...
+            // All the other mounts are ultimately attached to the detached mount
+            // That's why they aren't detached but are visible
+            event.mountfields.params.visible = true;
+            event.source = SOURCE_MOUNT;
+            pop_syscall(EVENT_MOUNT);
+        }
 
-        pop_syscall(EVENT_MOUNT);
         send_event(ctx, EVENT_MOUNT, event);
+
```
safchain, line 279: same here should we pop_syscall ?

pkg/security/ebpf/c/include/constants/offsets/filesystem.h
kovagsm: It was some leftover from some tests. removed

pkg/security/ebpf/c/include/hooks/mount.h
kovagsm: Not really, an `open_tree` can create multiple new mounts and if the syscall gets popped here, then the hooks on `attach_mnt` / `__attach_mnt` will return and won't send the other mounts to userspace.

The two syscalls that can reach this branch (fsmount and open_tree) are popping and guaranteeing that it's clean afterwards.

pkg/security/ebpf/c/include/hooks/mount.h
kovagsm: Also not here, since we can have many mounts from a single open_tree.

pkg/security/ebpf/c/include/structs/filesystem.h
```diff
@@ -20,14 +20,10 @@ struct mount_fields_t {
...
-struct fsmount_fields_t {
-    struct path_key_t root_key;
-    s32 fd;
-    u32 flags;
-    dev_t device;
-    unsigned int mount_attrs;
+    struct {
+        u32   visible :  1;  // Is mount visible in the VFS?
+        u32   detached : 1;  // A detached mount is always not visible, but an invisible mount isn't always detached
+    } params;
```
spikat: not a huge fan of this, at the end we can't use an u8 to gain memory and have to stick to a u32, so saving those in like two u16 will be much simpler right?

pkg/security/secl/model/unmarshallers_linux.go
kovagsm: yes, added!

pkg/security/secl/model/events.go
```diff
@@ -183,6 +185,10 @@ func (t EventType) String() string {
 		return "utimes"
 	case FileMountEventType:
 		return "mount"
+	case FileOpenTreeEventType:
+		return "open_tree"
+	case FileFsmountEventType:
+		return "fsmount"
```
spikat: should we remove this one? (+ the same line 122/123)

spikat: there is a lot of other places where we still mention fsmount, shloud we make a complete cleanup of them ?

pkg/security/ebpf/c/include/structs/filesystem.h
kovagsm: I have removed it. It was an attempt of saving space but couldn't do with an u8 like you said so in the end there's added complexity for nothing.

pkg/security/secl/model/events.go
kovagsm: I've removed from here since this event doesn't exist in the go side, but left it in the enum to keep it synchronized with the eBPF enum.

safchain (approved):

pkg/security/secl/model/events.go
spikat: ok I see!

spikat (approved):

YoannGh (approved):

PR #38223, opened 2025-06-25
[AGENTRUN-337] Add windows networkv2 e2e test

jeremy-hanna:
### What does this PR do?

Add a e2e test around the networkv2 windows check using the common check comparison utilities refactored from the diskv2 tests in https://github.com/DataDog/datadog-agent/pull/38199.

### DIFF
--- a/test/new-e2e/tests/agent-runtimes/checks/network/common.go
+++ b/test/new-e2e/tests/agent-runtimes/checks/network/common.go
@@ -1,0 +1,101 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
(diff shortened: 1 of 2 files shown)


test/new-e2e/tests/agent-runtimes/checks/network/common.go
```diff
@@ -0,0 +1,101 @@
...
+	}{
+		{
+			"default",
+			`init_config:
+instances:
+  - use_mount: false
+`,
+			``,
+		},
+	}
```
pgimalac, line 57: Ideally we should add more configurations to test (same for the disk check btw) but not a blocker

pgimalac (approved):

test/new-e2e/tests/agent-runtimes/checks/network/common.go
jeremy-hanna: agreed, the windows network check does not have much configurable variation to it

PR #38227, opened 2025-06-25
[ACIX-681] Move "result JSON File" parsing logic to a shared lib

Ishirui:
### What does this PR do?

Unifies and moves all the logic related to parsing the "result json files" generated by `go test` into its own lib. 

### Motivation

This logic was duplicated in many places, also reading and parsing the same "result json" files multiple times. By moving this into its own lib we avoid this duplication.

This is also required for a bugfix in #37862 that impacts how these "Result JSON" files are handled in `new-e2e-tests.run`

Existing and new unit tests

I tried to keep the behavior the same, but a lot of the functions in `TestWasher` have (imo) poorly defined semantics, ambiguous names and conventions that differ between them. For example:
- `get_flaky_failures` actually also returns tests that passed, not only failures
- Sometimes we consider a test to be failed if one of its parents failed or if there was a package-level failure, sometimes we don't
- There is some ambiguity between what is a "flaky" test and a test that is "marked as flaky"
- ...
This makes it difficult to be sure some edge case has not had its behavior changed. Hopefully the existing unit tests are enough to avoid this.

### DIFF
--- a/tasks/libs/testing/result_json.py
+++ b/tasks/libs/testing/result_json.py
@@ -76,6 +76,7 @@ class ResultJson:
 
     @property
     def package_tests_dict(self) -> dict[str, dict[str, list[ResultJsonLine]]]:
+        """Returns a dictionary of all ResultJsonLines sorted by package and test."""
         if not self._package_tests_dict:
             self._package_tests_dict = self._sort_into_packages_and_tests()
 
@@ -98,6 +99,9 @@ class ResultJson:
 
         for package, tests in self.package_tests_dict.items():
             for test_name, actions in tests.items():
+                if test_name == "_":
+                    # Skip package-level actions, as they are not tests
+                    continue
                 if run_is_failing(actions):
                     failing_tests[package].add(test_name)
 
--- a/tasks/libs/testing/result_json.py
+++ b/tasks/libs/testing/result_json.py
@@ -112,7 +112,8 @@ def run_is_failing(lines: list[ResultJsonLine]) -> bool:
     """Determines the failure status of the test run based on the actions in the lines."""
     is_fail = False
     for line in lines:
-        if line.action == ActionType.FAIL:
+        # Some test lines don't set their action to fail when they panic, but we should also consider that a failure
+        if line.action == ActionType.FAIL or line.output and "panic:" in line.output:
             is_fail = True
         elif line.action == ActionType.PASS:
             # As soon as we find a PASS action, we can conclude the test run is not a failure (retry succeeded)
--- a/tasks/libs/testing/result_json.py
+++ b/tasks/libs/testing/result_json.py
@@ -109,7 +109,11 @@ class ResultJson:
 
 
 def run_is_failing(lines: list[ResultJsonLine]) -> bool:
-    """Determines the failure status of the test run based on the actions in the lines."""
+    """
+    Determines the failure status of the test run based on the actions in the lines.
+    A run is considered failing if it contains a FAIL action or if it has an output with "panic:" in it.
+    Make sure the lines in `lines` all refer to the same test !
+    """
     is_fail = False
     for line in lines:
         # Some test lines don't set their action to fail when they panic, but we should also consider that a failure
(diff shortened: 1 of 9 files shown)


tasks/system_probe.py
```diff
@@ -1879,32 +1879,14 @@ def print_failed_tests(_, output_dir):
...
-
-                            if action == "pass" or action == "fail" or action == "skip":
-                                test_key = f"{package}.{name}"
-                                res = test_results.get(test_key)
-                                if res is None:
-                                    test_results[test_key] = action
-                                    continue
-
-                                if res == "fail":
-                                    print(f"re-ran [{test_platform}] {package} {name}: {action}")
```
brycekahle, line 1900: This functionality has been removed

tasks/system_probe.py
```diff
@@ -1879,32 +1879,14 @@ def print_failed_tests(_, output_dir):
...
-                                if (action == "pass" or action == "skip") and res == "fail":
-                                    test_results[test_key] = action
+        merged_test_json = merge_result_jsons(result_jsons)
+        failing_tests = merged_test_json.failing_tests
 
-        for key, res in test_results.items():
-            if res == "fail":
-                package, name = key.split(".", maxsplit=1)
-                print(color_message(f"FAIL: [{test_platform}] {package} {name}", "red"))
```
brycekahle, line 1907: We are intentionally only outputting tests that failed in all attempts. My understanding of `failing_tests` is those are tests that failed in **any** attempt.

tasks/libs/testing/result_json.py
```diff
@@ -0,0 +1,141 @@
...
+
+@dataclass
+class ResultJson:
+    lines: list[ResultJsonLine]
+    _package_tests_dict: dict[str, dict[str, list[ResultJsonLine]]] = field(
+        init=False, repr=False, default_factory=dict
+    )
+
+    def __post_init__(self):
+        self.lines.sort(key=lambda x: x.time)
```
KevinFairise2, line 50: Does it change something? I'd expect the json file to be sorted already. Curious if it can takes a lot of time if we call it on a big file

tasks/system_probe.py
Ishirui: Your understanding is 100% correct, thank you very much for the clarification ! I must have read through the previous code a bit too fast and did not notice this difference - I'll push a fix restoring the original behavior 👍

brycekahle: For context, the reason we do this is so users can scan the bottom of the Gitlab job log to understand which test failures caused the job to fail, whereas flaky tests that passed on re-run do not cause the job to fail.

brycekahle: Looking again, I don't think this task is used anymore, because we have this functionality is our Go `test-json-review` program. You can just delete this task 😁

Ishirui: Well, that makes it easier haha ! Thanks, will do 👍

KevinFairise2 (approved):
LGTM! We should have a decent amount of test to check the behavior of the logs parsing logic, so if everything is green I am rather confident that it is working.
It will definitely make parsing go test json output easier !

tasks/libs/testing/result_json.py
Ishirui: I added this while implementing the `merge_result_jsons`, depending on the order in which you pass the `ResultJSON` objects to the function it could mess up the order.

The order is not actually important at the moment, as `run_is_failing` only checks that there is some passing line in the run instead of something like "it failed then passed", for which the order _would_ be important.
But I think it is still better to ensure the order is kept, it makes the behavior more predictable (very useful when writing my unit tests haha) and makes changing the logic of `run_is_failing` in the future easier.

As for the performance concern, yeah, it definitely could but I don't think we'll be sorting lists more than a couple thousand lines long at a time. I was curious though so I ran a quick perf test on my laptop:
```python
from time import perf_counter_ns
from random import randint

l = [randint(0, 10_000_000) for _ in range(1_000_000)]

start = perf_counter_ns()
l.sort()
end = perf_counter_ns()
print(f"Sorting done in {(end - start) / 1_000_000} ms")
```
> Sorting done in 109.578875 ms

And that's for one million lines, so I think we're fine :grin:. And even more than that, as you say it will probably be already sorted most of the time, and IIRC `sort` already checks for that and skips doing anything if that's the case 👍

brycekahle (approved):

PR #38228, opened 2025-06-25
[pkg/dyninst] Implement DWARF5 loclist parsing

piob-io:
### Motivation

https://datadoghq.atlassian.net/browse/DEBUG-3991

Also reorganize loclist package, now that it is bigger.

Run tests locally with manually compiled test progs using gotip.
Eventually we will run tests automatically across different go versions.


pkg/dyninst/object/elf.go
```diff
@@ -130,7 +128,19 @@ func NewElfObject(elfFile *safeelf.File) (_ *ElfFile, retErr error) {
 	if err != nil {
 		return nil, err
 	}
+	_, _, dwarfVersion, byteOrder := dlvdwarf.ReadDwarfLengthVersion(ds.Info)
+	if byteOrder != binary.LittleEndian {
+		return nil, fmt.Errorf("unexpected DWARF byte order: %v", byteOrder)
+	}
 	unitVersions := dlvdwarf.ReadUnitVersions(ds.Info)
+	if dwarfVersion >= 5 {
+		// Delve unit offset calculations are 1 byte off.
```
grantseltzer, line 139: Is this a bug in delve code? If so is there an already open issue you can link in comment? Also if that's the case we should have a test for when the delve dependency version gets upgraded, including a fix.

pkg/dyninst/object/elf.go
```diff
@@ -62,23 +63,20 @@ func (e *ElfFile) DwarfSections() *DebugSections {
...
 
 // LoclistReader implements File.
-func (e *ElfFile) LoclistReader() (*LoclistReader, error) {
-	if e.dwarfSections.Loc == nil {
-		return nil, fmt.Errorf("no loclist section found")
+func (e *ElfFile) LoclistReader() (*loclist.Reader, error) {
+	var data []byte
+	if e.dwarfSections.LocLists != nil {
+		data = e.dwarfSections.LocLists
+	} else if e.dwarfSections.Loc != nil {
```
grantseltzer, line 72: Some comments on this block would be nice (i.e. LocLists for Dwarf5+, Loc for 2-4)

grantseltzer (approved):

pkg/dyninst/object/elf.go
piob-io: I'd consider this as a bug, however, this is also kinda an internal utility delve library. Looking at issues and repo, I am uncertain that delve fully supports DWARF5, it's more of, organic growth of features were DWARF5 support was desired. I think we should keep moving away from depending on delve, but I consider that low priority.

All integration tests will fail if these offsets don't match exactly, we are protected on that side. And a fix would be to just drop these lines.

pkg/dyninst/object/elf.go
piob-io: Done

pkg/dyninst/dwarf/loclist/reader.go
```diff
@@ -0,0 +1,291 @@
...
+		return Loclist{}, fmt.Errorf("no unit version found for unit at offset 0x%x", unit.Offset)
+	}
+	if unitVersion < 2 {
+		return Loclist{}, fmt.Errorf("unsupported unit version: %d", unitVersion)
+	}
+
+	var loclist Loclist
+	var err error
+	if unitVersion < 5 {
+		loclist, err = readDwarf2(r.data[offset:], r.ptrSize, typeByteSize)
```
ajwerner: bounds check offset?

pkg/dyninst/dwarf/loclist/reader.go
```diff
@@ -0,0 +1,291 @@
...
+	} else {
+		if r.debugAddr == nil {
+			return Loclist{}, fmt.Errorf("missing debug_addr section")
+		}
+		addrBase, ok := unit.Val(dwarf.AttrAddrBase).(int64)
+		if !ok {
+			return Loclist{}, fmt.Errorf("missing addr_base attribute")
+		}
+		loclist, err = readDwarf5(r.data[offset:], r.ptrSize, typeByteSize, r.debugAddr[addrBase:])
```
ajwerner: bound check offset and addrBase?

pkg/dyninst/object/elf.go
ajwerner: Is it this code with the question mark? https://github.com/go-delve/delve/blob/946e4885b69396512958b2774402e86acd530bbe/pkg/dwarf/parseutil.go#L126-L142

ajwerner: Might be nice to leave a permalink breadcrumb for future folks / if delve fixes it and we update it'll be clear what changed?

pkg/dyninst/dwarf/loclist/fix.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	// We want to find any pair of adjacent pieces that are in the same location
+	for currentByteSize(loclists) > expectedByteSize {
+		if len(loclists) == 0 || len(loclists[0].Pieces) == 0 {
+			break
+		}
+
+		numPieces := len(loclists[0].Pieces)
+		var foundIndex *int
```
ajwerner: my nit from the other pr about using a boolean as opposed to a pointer stands.

ajwerner (approved):
Excellent! Now we need to start testing against go1.25rc1

pkg/dyninst/object/elf.go
piob-io: Yes, that's the code, linked.

pkg/dyninst/dwarf/loclist/fix.go
piob-io: Done

pkg/dyninst/dwarf/loclist/reader.go
piob-io: Done

pkg/dyninst/dwarf/loclist/reader.go
piob-io: Done

PR #38229, opened 2025-06-25
[AGENTCFG-154] Adding an host metadata payload field to inform if the Agent is FIPS-enabled

rahulkaukuntla:
### What does this PR do?
We originally sent this information via agent_metadata, but agent_metadata is no longer part of the v5 payload it sends to sobotka, so we have to manually add this fips_mode field to the v5 payload.

I ran an agent incorporating these changes and pointed it to staging, where `fips_mode:true`. I then was able to see a metric `dd.agent_stats.fips_mode` with the same tag in the metrics explorer ui.

### DIFF
--- a/comp/metadata/host/hostimpl/utils/host.go
+++ b/comp/metadata/host/hostimpl/utils/host.go
@@ -164,9 +164,11 @@ func getProxyMeta(conf model.Reader) *ProxyMeta {
 }
 
 func getFipsMode() bool {
-	if val, err := fips.Enabled(); err == nil {
+	val, err := fips.Enabled()
+	if err == nil {
 		return val
 	}
+	log.Warn("Could not determine if FIPS mode is enabled: ", err)
 	return false
 }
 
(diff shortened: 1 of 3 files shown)


comp/metadata/host/hostimpl/utils/host.go
```diff
@@ -161,6 +163,13 @@ func getProxyMeta(conf model.Reader) *ProxyMeta {
 	}
 }
 
+func getFipsMode() bool {
+	if val, err := fips.Enabled(); err == nil {
+		return val
+	}
```
dustmop, line 170: If err is not enabled, could we get a Warn log here? From looking at the `pkg/fips` package, an error should only be returned on Windows if there's a problem accessing or decoding the registry.

dustmop (approved):
LGTM, with one small change requested.

comp/metadata/host/hostimpl/utils/host.go
rahulkaukuntla: done.

daniel-taf (approved):

PR #38232, opened 2025-06-25
[ AGNTLOG-224 ] Split utilization and capacity logging metrics according to a more granular level

ddrthall:
### What does this PR do?
The existing monitors used to identify and track resource allocation in the logs agent pipeline have been set to monitor the system at a very high level following the separation of the log agent's sender from the processor/strategy pipeline. 
Each main component tracks utilization/capacity as a singular entity rather than accounting for divisions of resources amongst its subcomponents. For example, the strategy pipeline component combines and reports its information as a singular number, despite there being 4 distinct strategy routines running on average. Recapturing the ability to monitor the components on a more granular level has the potential to be much more helpful when debugging certain classes of issues. 

This PR will attach instanceIDs to each of the major components to meaningfully see the behavior of their subcomponents. An example prometheus output in the new format is 
```
# HELP logs_component_utilization__bytes Gauge of the number of bytes currently held in a component and its buffers
# TYPE logs_component_utilization__bytes gauge
logs_component_utilization__bytes{instance="0",name="sender"} 96.86164602172508
logs_component_utilization__bytes{instance="2",name="processor"} 19.157356049314203
logs_component_utilization__bytes{instance="2",name="strategy"} 110.30437914691598
logs_component_utilization__bytes{instance="3",name="processor"} 10.831449356384455
logs_component_utilization__bytes{instance="3",name="strategy"} 121.05683618608943
logs_component_utilization__bytes{instance="q0s0",name="destination_reliable_0"} 81.14695323216918
logs_component_utilization__bytes{instance="q0s0",name="worker"} 96.86164602172508
# HELP logs_component_utilization__items Gauge of the number of items currently held in a component and its buffers
# TYPE logs_component_utilization__items gauge
logs_component_utilization__items{instance="0",name="sender"} 4.035901917571878
logs_component_utilization__items{instance="2",name="processor"} 0.7982231687214254
logs_component_utilization__items{instance="2",name="strategy"} 4.596015797788166
logs_component_utilization__items{instance="3",name="processor"} 0.4513103898493523
logs_component_utilization__items{instance="3",name="strategy"} 5.044034841087059
logs_component_utilization__items{instance="q0s0",name="destination_reliable_0"} 3.3811230513403827
logs_component_utilization__items{instance="q0s0",name="worker"} 4.035901917571878
# HELP logs_component_utilization__ratio Gauge of the utilization ratio of a component
# TYPE logs_component_utilization__ratio gauge
logs_component_utilization__ratio{instance="2",name="processor"} 9.725632976053808e-05
logs_component_utilization__ratio{instance="2",name="strategy"} 0.00010593946922206945
logs_component_utilization__ratio{instance="3",name="processor"} 2.9504501267019528e-05
logs_component_utilization__ratio{instance="3",name="strategy"} 3.797890311443881e-05
logs_component_utilization__ratio{instance="q0s0",name="destination_reliable_0"} 1.2004623758720116e-05
logs_component_utilization__ratio{instance="q0s0",name="worker"} 1.3957554083711945e-06
```

#### Low Level Details
The discrete components of the logs agent are, in operational order:
Processor
Strategy
Sender
Worker
Destination

Each subcomponent of the Processor has a one-to-one relationship with a subcomponent of the Strategy. These pairs are always assigned the same instanceID, which corresponds to the concept of a "pipeline id". 

Sender is a unique structure in this model, as it has no subcomponents and all data in the system inevitably routes through its singleton. It also does no work, and instead acts as a decoupler to allow work generated from a given Processor/Strategy pair to be freely distributed to whichever "Worker" subcomponent is available. As such, it is always designated as instanceID "0" for utilization_bytes/utilization_items, and it has no concept of a utilization_ratio. 

Workers have a variable subcomponent setup, with the concept of worker ids and worker queue assignments. As such, the instanceID for workers allows for grouping on queue ids while also preserving individual worker id uniqueness. Destinations are always assigned to a single worker and already have uniquely identifiable names, so all they require is to have their instanceID match their equivalent worker's instance. 

In addition to automated testing, prometheus metrics were generated and examined for all three of the possible pipeline configurations (HTTP/TCP/Legacy) to confirm that ids were correctly generated across the different possible permutations of subcomponents.

### DIFF
--- a/pkg/logs/metrics/capacity_monitor.go
+++ b/pkg/logs/metrics/capacity_monitor.go
@@ -44,6 +44,9 @@ func newCapacityMonitorWithTick(name, instance string, tickChan <-chan time.Time
 
 // AddIngress records the ingress of a payload
 func (i *CapacityMonitor) AddIngress(pl MeasurablePayload) {
+	if i == nil {
+		return
+	}
 	i.Lock()
 	defer i.Unlock()
 	i.ingress += pl.Count()
@@ -53,6 +56,9 @@ func (i *CapacityMonitor) AddIngress(pl MeasurablePayload) {
 
 // AddEgress records the egress of a payload
 func (i *CapacityMonitor) AddEgress(pl MeasurablePayload) {
+	if i == nil {
+		return
+	}
 	i.Lock()
 	defer i.Unlock()
 	i.egress += pl.Count()
@@ -62,6 +68,9 @@ func (i *CapacityMonitor) AddEgress(pl MeasurablePayload) {
 }
 
 func (i *CapacityMonitor) sample() {
+	if i == nil {
+		return
+	}
 	select {
 	case <-i.tickChan:
 		i.avgItems = ewma(float64(i.ingress-i.egress), i.avgItems)
@@ -76,6 +85,9 @@ func ewma(newValue float64, oldValue float64) float64 {
 }
 
 func (i *CapacityMonitor) report() {
+	if i == nil {
+		return
+	}
 	TlmUtilizationItems.Set(i.avgItems, i.name, i.instance)
 	TlmUtilizationBytes.Set(i.avgBytes, i.name, i.instance)
 }
--- a/pkg/logs/pipeline/processor_only_provider.go
+++ b/pkg/logs/pipeline/processor_only_provider.go
@@ -23,7 +23,7 @@ type processorOnlyProvider struct {
 	processor       *processor.Processor
 	inputChan       chan *message.Message
 	outputChan      chan *message.Message
-	pipelineMonitor *metrics.TelemetryPipelineMonitor
+	pipelineMonitor metrics.PipelineMonitor
 }
 
 // NewProcessorOnlyProvider is used by the logs check subcommand as the feature does not require the functionalities of the log pipeline other then the processor.
@@ -33,7 +33,7 @@ func NewProcessorOnlyProvider(diagnosticMessageReceiver diagnostic.MessageReceiv
 	encoder := processor.JSONEncoder
 	inputChan := make(chan *message.Message, chanSize)
 	pipelineID := "0"
-	pipelineMonitor := metrics.NewTelemetryPipelineMonitor()
+	pipelineMonitor := metrics.NewNoopPipelineMonitor(pipelineID)
 	processor := processor.New(cfg, inputChan, outputChan, processingRules,
 		encoder, diagnosticMessageReceiver, hostname, pipelineMonitor, pipelineID)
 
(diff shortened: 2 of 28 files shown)


drichards-87 (approved):

pkg/logs/pipeline/processor_only_provider.go
```diff
@@ -33,10 +32,10 @@ func NewProcessorOnlyProvider(diagnosticMessageReceiver diagnostic.MessageReceiv
 	outputChan := make(chan *message.Message, chanSize)
 	encoder := processor.JSONEncoder
 	inputChan := make(chan *message.Message, chanSize)
-	pipelineID := 0
-	pipelineMonitor := metrics.NewTelemetryPipelineMonitor(strconv.Itoa(pipelineID))
+	pipelineID := "0"
+	pipelineMonitor := metrics.NewTelemetryPipelineMonitor()
```
soberpeach: Should this be a noop pipeline monitor

pkg/logs/metrics/capacity_monitor.go
```diff
@@ -44,6 +44,9 @@ func newCapacityMonitorWithTick(name, instance string, tickChan <-chan time.Time
 
 // AddIngress records the ingress of a payload
 func (i *CapacityMonitor) AddIngress(pl MeasurablePayload) {
+	if i == nil {
```
soberpeach, line 47: Under what instances would the capacity monitor itself be nil?

pkg/logs/processor/processor.go
```diff
@@ -252,7 +254,7 @@ func (p *Processor) processMessage(msg *message.Message) {
 
 		p.utilization.Stop() // Explicitly call stop here to avoid counting writing on the output channel as processing time
 		p.outputChan <- msg
-		p.pipelineMonitor.ReportComponentIngress(msg, "strategy")
+		p.pipelineMonitor.ReportComponentIngress(msg, metrics.StrategyTlmName, p.instanceID)
```
soberpeach, line 257: Why report strategy ingress here vs in the strategy run loop?

pkg/logs/metrics/capacity_monitor.go
ddrthall: When returned by a noop Pipeline Monitor

pkg/logs/processor/processor.go
ddrthall: We always want to report ingress as soon as the bytes enter the corresponding system, which for strategy happens when the message lands in the outputChan above. This means we can account for bytes trapped in the queue of a given component as well as bytes that are taking longer than normal to process.

soberpeach (approved):

PR #38233, opened 2025-06-25
Go DI: Encode correct JSON format for eventually uploading to Datadog backend

grantseltzer:
### What does this PR do?

Encodes the correct envelope format for turning events from bpf into messages for upload.

Also updates integration test and testdata. Service name is currently being injected, and artificially so in integration test.

Notably program counters are not yet symbolicated.

### Motivation

Uploading events to the datadog backend 

https://datadoghq.atlassian.net/browse/DEBUG-3889

Integration test

### Example:

```
{
    "service": "sample",
    "ddsource": "dd_debugger",
    "ddtags": "",
    "logger": {
        "name": "",
        "method": ""
    },
    "debugger": {
        "snapshot": {
            "id": "fdf8b824-ecbc-47ea-bab0-60c70e12dff8",
            "timestamp": 1750927066975,
            "language": "go",
            "stack": [
                {
                    "function": "main.testSingleRune",
                    "file": "/home/vagrant/datadog-agent/pkg/dyninst/testprogs/progs/sample/basics.go",
                    "line": 15
                },
                {
                    "function": "main.main",
                    "file": "/home/vagrant/datadog-agent/pkg/dyninst/testprogs/progs/sample/main.go",
                    "line": 24
                },
                {
                    "function": "main.main",
                    "file": "/home/vagrant/datadog-agent/pkg/dyninst/testprogs/progs/sample/main.go",
                    "line": 24
                },
                {
                    "function": "runtime.main",
                    "file": "/usr/local/go/src/runtime/proc.go",
                    "line": 283
                },
                {
                    "function": "runtime.goexit",
                    "file": "/usr/local/go/src/runtime/asm_arm64.s",
                    "line": 1223
                }
            ],
            "probe": {
                "location": {
                    "method": "main.testSingleRune"
                }
            },
            "captures": {
                "entry": {
                    "arguments": {
                        "x": {
                            "type": "int32",
                            "value": 1
                        }
                    }
                }
            }
        }
    }
}
```


pkg/dyninst/decode/decoder.go
```diff
@@ -62,10 +73,135 @@ type typeAndAddr struct {
...
+//	    "name": "",
+//	    "method": ""
+//	  },
+//	  "debugger": {
+//	    "snapshot": {
+//	      "id": "242b6daa-506a-11f0-9835-001c421ddea2",
+//	      "timestamp": 1750707725206,
+//	      "language": "go",
+func (d *Decoder) encodeJSONPrologue(enc *jsontext.Encoder) error {
+	err := enc.WriteToken(jsontext.BeginObject)
```
tylfin: For readability, you could build an array then loop to make it slightly more compact:


```
fields := []jsontext.Token{
    jsontext.BeginObject, 
    ...,
}
for _, field := range fields {
    err := enc.WriteToken(field)
    if err != nil { ... }
}
```

grantseltzer: Yea good suggestion, I've been trying to think if there were better ways of handling these. I thought about having a stack of pushing jsontext.EndObject tokens (or just a count of how many are needed). But I don't know how readable that'd be.

tylfin: You could follow a `defer done(enc)` or `defer close(enc)` pattern such that each function that opens any parentheses needs to also close them

pkg/dyninst/decode/decoder.go
```diff
@@ -151,14 +216,70 @@ func (d *Decoder) Decode(event output.Event, out io.Writer) error {
 	if !ok {
 		return errors.New("no probe event found for root type")
 	}
-	err = enc.WriteToken(jsontext.String("probe_id"))
+	err = enc.WriteToken(jsontext.String("probe"))
```
ajwerner: If we keep this approach (which I don't think we should), let's compactify here and everywhere else too. Perhaps make a helper:

```
func writeTokens(toks ...jsontext.Token) error {
```

then this might be:
```go
if err := writeTokens(
    jsontext.BeginObject,
    jsontext.String("id"),
    jsontext.String(p.probe.ID),
    // ...
); err != nil {
    return err
}
```

pkg/dyninst/integration_test.go
```diff
@@ -245,6 +238,85 @@ func testDyninst(t *testing.T, service string, sampleServicePath string, probes
 	}
 }
 
+// Purge:
```
ajwerner: I think I'd structure this differently. Consider:

```go
type jsonRedactor struct {
	matches     func(ptr jsontext.Pointer) bool
	replacement jsontext.Value
}

func redactPtr(toMatch string, replacement jsontext.Value) jsonRedactor {
	return jsonRedactor{
		matches: func(ptr jsontext.Pointer) bool {
			return string(ptr) == toMatch
		},
		replacement: replacement,
	}
}

func redactPtrPrefixSuffix(prefix, suffix string, replacement jsontext.Value) jsonRedactor {
	return jsonRedactor{
		matches: func(ptr jsontext.Pointer) bool {
			return strings.HasPrefix(string(ptr), prefix) && strings.HasSuffix(string(ptr), suffix)
		},
		replacement: replacement,
	}
}

var redactors = []jsonRedactor{
    redactPtr("/debugger/snapshot/stack", []byte(`"<stack>"`)),
    redactPtr("/debugger/snapshot/stack", []byte(`"<id>"`)),
    redactPtr("/debugger/snapshot/timestamp", []byte(`"<ts>"`)),
    redactPtrPrefixSuffix("/debugger/snapshot/captures/", "/Address", []byte(`"<addr>"`)),
}

func redactJSON(t *testing.T, input []byte) (redacted []byte) {
	d := jsontext.NewDecoder(bytes.NewReader(input))
	var buf bytes.Buffer
	e := jsontext.NewEncoder(&buf)
	for {
		tok, err := d.ReadToken()
		if errors.Is(err, io.EOF) {
			break
		}
		require.NoError(t, err)
		kind, idx := d.StackIndex(d.StackDepth())
		require.NoError(t, e.WriteToken(tok))
		if kind != '{' || idx%2 == 0 {
			continue
		}
		ptr := d.StackPointer()
		for _, redactor := range redactors {
			if redactor.matches(ptr) {
				_, err := d.ReadValue()
				require.NoError(t, err)
				require.NoError(t, e.WriteValue(redactor.replacement))
				break
			}
		}
	}
	return buf.Bytes()
}
```

This way we could extend it to validate properties or to do more in the future as we may want to redact other stuff later.

ajwerner (commented):
I think we're doing ourselves a disservice with regards to readability and adaptability by writing all these tokens like this. It makes sense in the dynamic parse as you had before, but for this outer envelope I think it's a lot more harm than good.

Consider:
```go
type snapshotMessage struct {
    Service string `json:"service"`
    DDSource string `json:"ddsource"`
    // ...

}
```

For the parts of the message that are dynamic, you wrap up the relevant state and shove it into the relevant field and implement https://pkg.go.dev/github.com/go-json-experiment/json#MarshalerTo

All of this can hang off the decoder, so you don't need to allocate anything new per call. You  set the relevant fields and then just call `json.Marshal` and good efficient things can happen where we want them to and we can eliminate all this toil.

pkg/dyninst/decode/decoder.go
grantseltzer: When you say "If we keep this approach" do you mean the code structure as it is now or how it'd be even after compactifying as you suggested?

grantseltzer:
> I think we're doing ourselves a disservice with regards to readability and adaptability by writing all these tokens like this. It makes sense in the dynamic parse as you had before, but for this outer envelope I think it's a lot more harm than good.
> 
> Consider:
> 
> ```go
> type snapshotMessage struct {
>     Service string `json:"service"`
>     DDSource string `json:"ddsource"`
>     // ...
> 
> }
> ```
> 
> For the parts of the message that are dynamic, you wrap up the relevant state and shove it into the relevant field and implement https://pkg.go.dev/github.com/go-json-experiment/json#MarshalerTo
> 
> All of this can hang off the decoder, so you don't need to allocate anything new per call. You set the relevant fields and then just call `json.Marshal` and good efficient things can happen where we want them to and we can eliminate all this toil.

I've pushed a change that I believe does what you describe. In addition to looking for code organization and allocation efficiency, I want to clarify from our irl conversation earlier (please tell me if the following is true or not)

1) An ir.Program should not be assumed to be for a single service. For example, you may instrument the same functions in a bunch of different binaries (i.e. probing library functions).

2) A decoder is not specifically for a single service, but the service/process information can be injected by the ProcessInfoResolver that we added to this branch. So it shouldn't be assumed to be specific to anything as such.

3) I still need to figure out exactly what goes into, and/or what is the purpose of, ddsource, logger, and message.

pkg/dyninst/decode/decoder.go
```diff
@@ -30,19 +33,60 @@ type probeEvent struct {
...
 
+type snapshotMessage struct {
+	Service  string   `json:"service"`
+	DDSource string   `json:"ddsource"`
+	DDTags   string   `json:"ddtags"`
+	Logger   Logger   `json:"logger"`
+	Debugger Debugger `json:"debugger"`
+}
+
+type Logger struct {
```
ajwerner: this type shouldn't be exported

pkg/dyninst/decode/decoder.go
```diff
@@ -526,3 +614,47 @@ func (d *Decoder) encodeBaseTypeValue(enc *jsontext.Encoder, irType *ir.BaseType
 		return errors.New("invalid base type")
 	}
 }
+func (s *snapshotMessage) MarshalJSONTo(enc *jsontext.Encoder) error {
```
ajwerner: I think you misunderstood what I was saying. This type should not implement `MarshalerTo`. It's the `Snapshot` field that should hold a type that implements `MarshalerTo`.

pkg/dyninst/decode/decoder.go
```diff
@@ -30,19 +33,60 @@ type probeEvent struct {
...
+	Logger   Logger   `json:"logger"`
+	Debugger Debugger `json:"debugger"`
+}
+
+type Logger struct {
+	Name   string `json:"name"`
+	Method string `json:"method"`
+}
+
+type Debugger struct {
```
ajwerner: shouldn't be an exported type

ajwerner (commented):
Let me make it more concrete for you:

```go
type snapshotMessage struct {/* .. more or less as you had it */}

// Take the arguments to fill in the relevant fields
func (s *snapshotMessage) init(/* ... */) {}

// Clear the per-snapshot fields
func (s *snapshotMessage) clear() {}

type snapshotData struct {
    decoder *Decoder
    event output.Event
}

func (sd *snapshotData) MarshalToJSON(e *jsontext.Encoder) error {
    return sd.d.decodeEvent(sd.event, e)
}

func (d *Decoder) Decode(event, out) error {
    d.snapshotMessage.init(event)
    defer d.snapshotMessage.clear()
    return json.MarshalWrite(out, &d.snapshotMessage)
}
```

pkg/dyninst/decode/decoder.go
```diff
@@ -52,17 +52,37 @@ type snapshotMessage struct {
...
+		decoder: decoder,
+		event:   event,
+	}
+}
+
+func (s *snapshotMessage) clear() {
+	s.Debugger.Snapshot = snapshotData{}
+}
+
+type logger struct {
```
ajwerner: I filed https://datadoghq.atlassian.net/browse/DEBUG-4111 to track collecting the goroutine ID to put here.

pkg/dyninst/decode/decoder.go
```diff
@@ -33,11 +33,11 @@ type probeEvent struct {
...
-	program                *ir.Program
-	stackHashes            map[uint64][]uint64
-	probeEvents            map[ir.TypeID]probeEvent
-	procInfoResolver       ProcessInfoResolver
-	defaultSnapshotMessage *snapshotMessage
+	program          *ir.Program
+	stackHashes      map[uint64][]uint64
+	probeEvents      map[ir.TypeID]probeEvent
+	procInfoResolver ProcessInfoResolver
+	snapshotMessage  *snapshotMessage
```
ajwerner: this doesn't need to be a pointer

pkg/dyninst/decode/decoder.go
```diff
@@ -107,21 +124,9 @@ type typeAndAddr struct {
 }
 
 func (d *Decoder) Decode(event output.Event, out io.Writer) error {
```
ajwerner: I've changed my mind on what I want here. Instead of having the ProcessInfoResolver, let's instead define a structure and pass this as the argument to `Decode`

```go
package decode

type Event struct {
    output.Event
    ServiceName string
    // Whatever else we need...
}
```

pkg/dyninst/decode/decoder.go
```diff
@@ -30,19 +33,79 @@ type probeEvent struct {
...
+	probeEvents     map[ir.TypeID]probeEvent
+	snapshotMessage snapshotMessage
+}
+
+type Event struct {
+	output.Event
+	ServiceName string
+}
+
+// ProcessInfoResolver is used to inject resolution of service info which is
```
ajwerner: this rotted, no?

pkg/dyninst/decode/decoder.go
```diff
@@ -30,19 +33,79 @@ type probeEvent struct {
...
+}
+
+type debugger struct {
+	Snapshot snapshotData `json:"snapshot"`
+}
+
+type snapshotData struct {
+	decoder *Decoder
+	event   output.Event
+}
```
ajwerner: Let's take this one step further:

```
type snapshotData struct {
    ID uuid.UUID `json:"id"`
    Timestamp int `json:"timestamp"`
    Language goLanguage `json:"language"`
}

type goLanguage struct{}

var goLangaugeBytes = []byte(`"go"`)
func (goLanguage) MarshalJSON(e *jsontext.Encoder) ([]byte, error) {
    return goLanguageBytes, nil
}

type stackData struct {
    stackPCs []uint64
    header *output.EventHeader
    // symbolicator? 
}

func probeData struct {
 // id
 // location
 // captures -- some sort of structure that has the root event type, the root event data item, the map... 
} 
```

pkg/dyninst/decode/decoder.go
```diff
@@ -62,14 +125,30 @@ type typeAndAddr struct {
...
+	uu, err := uuid.NewUUID()
+	if err != nil {
+		return err
+	}
+	err = writeTokens(enc,
+		jsontext.BeginObject, // begin snapshot
+		jsontext.String("id"),
+		jsontext.String(uu.String()),
+		jsontext.String("timestamp"),
+		jsontext.Int(time.Now().UnixMilli()),
```
ajwerner: This is wrong. The event header has a timestamp. At least leave a TODO to follow up in https://datadoghq.atlassian.net/browse/DEBUG-4093

pkg/dyninst/decode/decoder.go
```diff
@@ -60,78 +71,56 @@ type typeAndAddr struct {
...
-	if err != nil {
-		return err
-	}
+type snapshotMessage struct {
+	Service  string       `json:"service"`
+	DDSource string       `json:"ddsource"`
+	DDTags   string       `json:"ddtags"`
+	Logger   logger       `json:"logger"`
+	Debugger debuggerData `json:"debugger"`
+	rootData []byte       `json:"-"`
```
ajwerner: nit: no reason to have json tags on private fields -- but also this does not seem to be used, please remove it.

pkg/dyninst/decode/decoder.go
```diff
@@ -60,78 +71,56 @@ type typeAndAddr struct {
...
-	err = enc.WriteToken(jsontext.BeginObject)
-	if err != nil {
-		return err
-	}
-	addressReferenceCount := map[typeAndAddr]output.DataItem{}
-
 	var (
-		rootData []byte
 		rootType *ir.EventRootType
 	)
```
ajwerner: nit: var block with one entry looks weird

pkg/dyninst/decode/decoder.go
```diff
@@ -60,78 +71,56 @@ type typeAndAddr struct {
...
-	for item, err := range event.DataItems() {
+	addressReferenceCount := map[typeAndAddr]output.DataItem{}
```
ajwerner: store this in the snapshotMessage and clear the map in the `clear` call

pkg/dyninst/decode/decoder.go
```diff
@@ -145,393 +134,52 @@ func (d *Decoder) Decode(event output.Event, out io.Writer) error {
...
-	}
-	err = enc.WriteToken(jsontext.String(p.probe.GetID()))
+	var (
+		pcs []uint64
+		err error
+	)
+	header, err := event.Event.Header()
 	if err != nil {
-		return err
+		log.Println("error getting header", err)
```
ajwerner: wrong log library

pkg/dyninst/decode/decoder.go
```diff
@@ -145,393 +134,52 @@ func (d *Decoder) Decode(event output.Event, out io.Writer) error {
...
+		return
 	}
+	s.Debugger.Snapshot.Timestamp = int(header.Ktime_ns)
```
ajwerner: This is more bogus than what was there before. Use time.Now() and fill it in with the appropriate resolution that the backend wants (ms?) but leave a TODO to get the value from Ktime_ns.

pkg/dyninst/decode/decoder.go
```diff
@@ -30,17 +27,31 @@ type probeEvent struct {
 // Decoder decodes the output of the BPF program into a JSON format.
 // It is not guaranteed to be thread-safe.
 type Decoder struct {
-	program     *ir.Program
-	stackHashes map[uint64][]uint64
-	probeEvents map[ir.TypeID]probeEvent
+	program         *ir.Program
+	symbolicator    symbol.Symbolicator
+	stackFrames     map[uint64][]symbol.StackFrame
```
ajwerner: I think it's better to store the pcs here and delegate caching of the strings to the symbolicator. I'd carve out a "class" (as Piotr called it)  called something like  `stackInterner` and have that be a `map[uint64][]uint64`

pkg/dyninst/decode/decoder.go
```diff
@@ -145,393 +134,52 @@ func (d *Decoder) Decode(event output.Event, out io.Writer) error {
...
-
-	// We iterate over the 'Expressions' of the EventRoot which contains
-	// metadata and raw bytes of the parameters of this function.
-	for _, expr := range rootType.Expressions {
-		parameterType := expr.Expression.Type
-		parameterData := rootData[expr.Offset : expr.Offset+parameterType.GetByteSize()]
-		err = enc.WriteToken(jsontext.String(expr.Name))
+	stackFrames, ok := decoder.stackFrames[header.Stack_hash]
+	if !ok {
+		pcs, err = event.StackPCs()
```
ajwerner, line 155: I find the error handling approach here to be broadly surprising. In general I think it's unreasonable to just be logging and returning and then *not* returning that error. Like I think that the end result would be a panic because later you assume that init has succeeded but at various points you'll have returned but without, say, populating the root type. The summary is you must return errors when things that are unexpected and unhandled occur.

This PR would benefit from testing that actually covers the failure cases in addition to the happy cases.

pkg/dyninst/decode/decoder.go
```diff
@@ -145,393 +134,52 @@ func (d *Decoder) Decode(event output.Event, out io.Writer) error {
...
-		return err
+	probe, ok := decoder.probeEvents[rootType.ID]
+	if !ok {
+		log.Println("error getting probe", err)
+		return
 	}
-	err = enc.WriteToken(jsontext.EndObject) // full object
-	if err != nil {
-		return err
+	switch where := probe.probe.GetWhere().(type) {
```
ajwerner, line 170: I think the default case here is an error.

pkg/dyninst/decode/decoder.go
```diff
@@ -145,393 +134,52 @@ func (d *Decoder) Decode(event output.Event, out io.Writer) error {
...
-		return errors.New("strings are not base types")
-	case reflect.Struct:
-		return errors.New("structs are not base types")
-	case reflect.UnsafePointer:
-		return errors.New("unsafe pointers are not base types")
-	default:
-		return errors.New("invalid base type")
-	}
+func (s *snapshotMessage) clear() {
+	s.Debugger.Snapshot = snapshotData{}
```
ajwerner, line 190: this should probably clear some of the other fields like the logger and the service

pkg/dyninst/decode/marshal.go
```diff
@@ -0,0 +1,496 @@
...
+func (ad *argumentsData) MarshalJSONTo(enc *jsontext.Encoder) error {
+	var err error
+	currentlyEncoding := map[typeAndAddr]struct{}{}
+
+	err = writeTokens(enc,
+		jsontext.BeginObject,
+	)
+	if err != nil {
+		return err
+	}
```
ajwerner: all of these here and throughout the rest of the file can be more succinctly written as below. I don't see any upside to all the extra newlines.

```go
if err := writeTokens(enc, jsontext.BeginObject); err != nil {
    return err
}
```
or in the case of longer runs:
```go
if err = writeTokens(enc,
	jsontext.String(expr.Name),
	jsontext.BeginObject,
	jsontext.String("type"),
	jsontext.String(parameterType.GetName()),
	jsontext.String("value"),
); err != nil {
    return err
}
```

pkg/dyninst/decode/marshal.go
```diff
@@ -0,0 +1,496 @@
...
+			}
+			err = d.encodeValue(
+				enc,
+				dataItems,
+				currentlyEncoding,
+				d.program.Types[ir.TypeID(pointedValue.Header().Type)],
+				pointedValue.Data(),
+			)
+			if err != nil {
+				return fmt.Errorf("could not get pointed-at value: %s", err)
```
ajwerner: I don't understand this message. Also, `%s` with an err is always worse than `%w`.

ajwerner (commented):
Here's another pass

pkg/dyninst/decode/decoder.go
grantseltzer: Ah because it needs to be resolved to wall time. Changing back.

pkg/dyninst/decode/decoder.go
grantseltzer: Yea pretty careless. Fixing. I'll work on testing as well.

pkg/dyninst/decode/decoder.go
grantseltzer: Agree to disagree - but i'll change it as a sign of good faith :)

pkg/dyninst/decode/decoder.go
grantseltzer: Removing the tag, but the field itself is used.

pkg/dyninst/decode/decoder.go
grantseltzer: I'm not sure I agree. We pass the stack hash and PC to the symbolicator anyway, so why would caching of PCs live in both the decoder and symbolicator?

Another alternative is to cache 1:1 from PC to Stackframe.

pkg/dyninst/decode/marshal.go
```diff
@@ -0,0 +1,473 @@
...
+
+type captureData struct {
+	Entry capturePointData `json:"entry"`
+}
+
+type capturePointData struct {
+	Arguments argumentsData `json:"arguments"`
+}
+
+type argumentsData struct {
```
ajwerner, line 75: There no reason for json tags on private fields

pkg/dyninst/decode/doc.go
```diff
@@ -8,4 +8,70 @@
...
+//	                            "value": 1
+//	                        }
+//	                    }
+//	                }
+//	            }
+//	        }
+//	    }
+//	}
+//
+// ```
```
ajwerner: In go doc comments you just need to indent for it to be block quoted and pre-formatted -- the \`\`\` doesn't do anything and ultimately renders in quite an ugly way:

> A pair of consecutive backticks (` U+0060) is interpreted as a Unicode left quote (“ U+201C), and a pair of consecutive single quotes (' U+0027) is interpreted as a Unicode right quote (” U+201D).

https://tip.golang.org/doc/comment#paragraphs

pkg/dyninst/decode/marshal.go
```diff
@@ -0,0 +1,473 @@
...
+	Name   string `json:"name"`
+	Method string `json:"method"`
+}
+
+type debuggerData struct {
+	Snapshot snapshotData `json:"snapshot"`
+}
+
+type snapshotData struct {
+	decoder *Decoder `json:"-"`
```
ajwerner: There no reason for json tags on private fields

pkg/dyninst/decode/marshal.go
```diff
@@ -0,0 +1,473 @@
...
+		return writeTokens(enc, jsontext.String("unimplemented - swiss map groups type"))
+	case *ir.EventRootType:
+		return writeTokens(enc, jsontext.String("unimplemented - event root type"))
+	default:
+		return errors.New("invalid type")
+	}
+	return nil
+}
+
+func (d *Decoder) encodeBaseTypeValue(enc *jsontext.Encoder, irType *ir.BaseType, data []byte) error {
```
ajwerner: This method doesn't seem to use the receiver. In general I don't think this should be a method, just make it a function.

pkg/dyninst/decode/marshal.go
```diff
@@ -0,0 +1,473 @@
...
+			irType: uint32(v.Data.GetID()),
+			addr:   address,
+		}]
+		if !ok {
+			return fmt.Errorf("string content not present in data items")
+		}
+		length := stringValue.Header().Length
+		if len(stringValue.Data()) < int(length) {
+			return errors.New("string content not long enough for known length")
+		}
```
ajwerner: If you're going to construct a new value here, give us the numbers for the lengths and the expected length.

pkg/dyninst/decode/marshal.go
```diff
@@ -0,0 +1,473 @@
...
+				}
+				address = binary.NativeEndian.Uint64(data[field.Offset : field.Offset+uint32(field.Type.GetByteSize())])
+			}
+		}
+		stringValue, ok := dataItems[typeAndAddr{
+			irType: uint32(v.Data.GetID()),
+			addr:   address,
+		}]
+		if !ok {
+			return fmt.Errorf("string content not present in data items")
```
ajwerner: This shouldn't fail -- just write out null or something for now and then leave a TODO to populate the reason it's not present when you come back to do that (which is soon, right?)

pkg/dyninst/decode/marshal.go
```diff
@@ -0,0 +1,473 @@
...
+	case *ir.GoChannelType:
+		return writeTokens(enc, jsontext.String("unimplemented - channel type"))
+	case *ir.GoStringHeaderType:
+		var address uint64
+		for _, field := range v.Fields {
+			if field.Name == "str" {
+				if uint32(field.Type.GetByteSize()) != 8 || len(data) < int(field.Offset+field.Type.GetByteSize()) {
+					return errors.New("malformed string field 'str'")
+				}
+				address = binary.NativeEndian.Uint64(data[field.Offset : field.Offset+uint32(field.Type.GetByteSize())])
```
ajwerner: just use `8` given you're already checking for it. Also, variables are your friends. Also, `uint32(field.Type.GetByteSize())` -- isn't this cast meaningless (`GetByteSize() uint32`)?

```
binary.NativeEndian.Uint64(data[field.Offset : field.Offset+8)])
```

pkg/dyninst/decode/marshal.go
```diff
@@ -0,0 +1,473 @@
...
+		}
+		return nil
+	case *ir.GoChannelType:
+		return writeTokens(enc, jsontext.String("unimplemented - channel type"))
+	case *ir.GoStringHeaderType:
+		var address uint64
+		for _, field := range v.Fields {
+			if field.Name == "str" {
+				if uint32(field.Type.GetByteSize()) != 8 || len(data) < int(field.Offset+field.Type.GetByteSize()) {
+					return errors.New("malformed string field 'str'")
```
ajwerner: Tell us what about it is malformed

pkg/dyninst/decode/marshal.go
```diff
@@ -0,0 +1,473 @@
...
+		}
+		stringValue, ok := dataItems[typeAndAddr{
+			irType: uint32(v.Data.GetID()),
+			addr:   address,
+		}]
+		if !ok {
+			return fmt.Errorf("string content not present in data items")
+		}
+		length := stringValue.Header().Length
+		if len(stringValue.Data()) < int(length) {
```
ajwerner: nit: use a variable for the data: `stringData := stringValue.Data()`

pkg/dyninst/decode/marshal.go
```diff
@@ -0,0 +1,473 @@
...
+			return errors.New("string content not long enough for known length")
+		}
+		if err := writeTokens(enc,
+			jsontext.String(string(stringValue.Data()[:length]))); err != nil {
```
ajwerner: this is a rather unfortunate allocation. Instead let's do:

```go
str := unsafe.String(unsafe.SliceData(stringData), length)
if err := writeTokens(enc, jsontext.String(str)); err != nil {
```

pkg/dyninst/decode/marshal.go
```diff
@@ -0,0 +1,473 @@
...
+	default:
+		return errors.New("invalid type")
+	}
+	return nil
+}
+
+func (d *Decoder) encodeBaseTypeValue(enc *jsontext.Encoder, irType *ir.BaseType, data []byte) error {
+	kind, ok := irType.GetGoKind()
+	if !ok {
+		return errors.New("No go kind")
```
ajwerner: Give us more information about the type and its name. Also, nit, error messages shouldn't start with capitals so they compose better

pkg/dyninst/decode/marshal.go
```diff
@@ -0,0 +1,473 @@
...
+		}
+		address := binary.NativeEndian.Uint64(data[0:8])
+		length := binary.NativeEndian.Uint64(data[8:16])
+		if length == 0 {
+			if err := writeTokens(enc,
+				jsontext.BeginArray,
+				jsontext.EndArray); err != nil {
+				return err
+			}
+			return nil
```
ajwerner: `return writeTokens(`

pkg/dyninst/decode/marshal.go
```diff
@@ -0,0 +1,473 @@
...
+		}
+		elementType := v.Data.ID
+		elementSize := int(v.Data.Element.GetByteSize())
+		sliceDataItem, ok := dataItems[typeAndAddr{
+			addr:   address,
+			irType: uint32(elementType),
+		}]
+		if !ok {
+			return errors.New("slice data item not found in data items")
```
ajwerner: This shouldn't be a fatal error -- it can happen if we run out of buffer space.

pkg/dyninst/decode/marshal.go
```diff
@@ -0,0 +1,473 @@
...
+			elementData := data[i*elementSize : (i+1)*elementSize]
+			if err := d.encodeValue(enc,
+				dataItems, currentlyEncoding, v.Element, elementData); err != nil {
+				return err
+			}
+		}
+		if err = writeTokens(enc, jsontext.EndArray); err != nil {
+			return err
+		}
+		return nil
```
ajwerner, line 322: why is it that some of these switch cases return and some fall to the end of the function which has a return nil? Let's remove the `return nil` from the end of the function and force each case to return explicitly.

pkg/dyninst/decode/marshal.go
```diff
@@ -0,0 +1,473 @@
...
+		jsontext.String("line"),
+		jsontext.Int(int64(sl.Line)),
+		jsontext.EndObject,
+	); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (d *Decoder) encodeValue(
```
ajwerner, line 162: I'm not certain I see a ton of value in this being a method -- we don't use `d` for much of anything here and I think it can then become something of a false dependency

pkg/dyninst/decode/decoder.go
```diff
@@ -60,478 +73,119 @@ type typeAndAddr struct {
...
-	}
-	addressReferenceCount := map[typeAndAddr]output.DataItem{}
 
-	var (
-		rootData []byte
-		rootType *ir.EventRootType
-	)
-	for item, err := range event.DataItems() {
+	var rootType *ir.EventRootType
+	s.addressReferenceCount = map[typeAndAddr]output.DataItem{}
```
ajwerner: rather than constructing a new map every time, how about we keep the old one around so we don't have to keep re-allocating the backing arrays?

pkg/dyninst/decode/decoder.go
ajwerner: Okay. I'll drop this.

pkg/dyninst/decode/decoder.go
```diff
@@ -60,478 +73,119 @@ type typeAndAddr struct {
...
-		return errors.New("strings are not base types")
-	case reflect.Struct:
-		return errors.New("structs are not base types")
-	case reflect.UnsafePointer:
-		return errors.New("unsafe pointers are not base types")
-	default:
-		return errors.New("invalid base type")
-	}
+func (s *snapshotMessage) clear() {
+	s.Debugger.Snapshot = snapshotData{}
```
ajwerner, line 190: If we're going to preserve and reuse some map data structures, we'll need to do something fancier here.

pkg/dyninst/integration_test.go
```diff
@@ -194,51 +200,126 @@ func testDyninst(
...
+	return jsonRedactor{
+		matches: func(ptr jsontext.Pointer) bool {
+			return strings.HasPrefix(string(ptr), prefix) && strings.HasSuffix(string(ptr), suffix)
+		},
+		replacement: replacement,
+	}
+}
+
+var redactors = []jsonRedactor{
+	redactPtr("/debugger/snapshot/stack", []byte(`"<stack>"`)),
```
ajwerner: maybe for the stacks we just redact some of the parts of the frame data? It seems to me like the function names shouldn't change and it's nice to see them.

ajwerner (commented):
Getting close!

pkg/dyninst/decode/marshal.go
grantseltzer: That's fine. And yes I'll handle not capture reason very soon.

pkg/dyninst/decode/doc.go
```diff
@@ -8,4 +8,70 @@
...
+//
+// JSON is written directly to the decoder's output stream for efficiency.
+// An example of the JSON output is:
+//
+// ```json
+//
+//	{
+//	    "service": "sample",
+//	    "ddsource": "dd_debugger",
+//	    "ddtags": "",
```
ajwerner: Let's put `Tags` into the Event and propagate them through in this ddtags field

pkg/dyninst/decode/marshal.go
grantseltzer: This is really interesting, haven't put much thought into how Go allocates strings even in cases like this where it's not assigned to a variable directly.

pkg/dyninst/decode/decoder.go
grantseltzer: Is there something in particular you're thinking when you say fancy? Or explicitly zeroing/clearing all the fields under s.Debugger.Snapshot is what you were thinking?

pkg/dyninst/decode/doc.go
```diff
@@ -8,4 +8,66 @@
...
+//	                    "function": "runtime.goexit",
+//	                    "file": "/usr/local/go/src/runtime/asm_arm64.s",
+//	                    "line": 1223
+//	                }
+//	            ],
+//	            "probe": {
+//	                "location": {
+//	                    "method": "main.testSingleRune"
+//	                }
+//	            },
```
ajwerner: This needs the probe ID and version. Turns out this drives a bunch of the ui.

pkg/dyninst/decode/decoder.go
```diff
@@ -60,478 +75,136 @@ type typeAndAddr struct {
...
+	s.Debugger.Snapshot.Probe.Location.Type = ""
+	clear(s.Debugger.Snapshot.Captures.Entry.Arguments.addressReferenceCount)
+	s.Debugger.Snapshot.Captures.Entry.Arguments.rootData = nil
+	s.Debugger.Snapshot.Captures.Entry.Arguments.rootType = nil
+	s.Debugger.Snapshot.Captures.Entry.Arguments.event = Event{}
+	s.Debugger.Snapshot.Captures.Entry.Arguments.decoder = nil
+	s.Logger.Name = ""
+	s.Logger.Method = ""
+
+	s.Service = ""
```
ajwerner: Hmm, this does seem more error prone than before. Perhaps the "right" thing to do is to put all the data structures we intend to reuse on the out snapshotMessage and then clear it wholesale like you have before but then in init, reinitialize the shared pointers from those. Concretely, put the `addressReferenceCount` map (and ideally a data items map we'll reuse) into the `Decoder` and then pass those to init.

pkg/dyninst/decode/decoder.go
```diff
@@ -60,478 +75,136 @@ type typeAndAddr struct {
...
-	case *ir.GoSwissMapHeaderType:
-		return enc.WriteToken(jsontext.String("unimplemented - swiss map header"))
-	case *ir.GoSwissMapGroupsType:
-		return enc.WriteToken(jsontext.String("unimplemented - swiss map groups"))
-	case *ir.EventRootType:
-		return enc.WriteToken(jsontext.String("unimplemented - event root"))
-	default:
-		return errors.New("invalid type")
+	s.DDTags = strings.Join(probe.probe.GetTags(), ",")
+	s.Debugger.Snapshot.Stack.frames = stackFrames
```
ajwerner, line 178: seems like you've neglected to populate `s.Debugger.Snapshot.Probe.ID` (and also you're missing version).

pkg/dyninst/decode/decoder.go
```diff
@@ -60,478 +75,136 @@ type typeAndAddr struct {
...
-		return errors.New("unsafe pointers are not base types")
-	default:
-		return errors.New("invalid base type")
-	}
+func (s *snapshotMessage) clear() {
+	s.Debugger.Snapshot.decoder = nil
+	s.Debugger.Snapshot.ID = uuid.UUID{}
+	s.Debugger.Snapshot.Timestamp = 0
+	s.Debugger.Snapshot.Language = ""
+	s.Debugger.Snapshot.Stack.frames = s.Debugger.Snapshot.Stack.frames[:0]
```
ajwerner: truncating this slice is actually kinda bad because you're leaking the strings in the remaining capacity of the slice.

ajwerner (commented):
Now it's looking extremely close

pkg/dyninst/decode/marshal.go
```diff
@@ -0,0 +1,485 @@
...
+
+func (sl *stackLine) MarshalJSONTo(enc *jsontext.Encoder) error {
+	if err := writeTokens(enc,
+		jsontext.BeginObject,
+		jsontext.String("function"),
+		jsontext.String(sl.Function),
+		jsontext.String("file"),
+		jsontext.String(sl.File),
+		jsontext.String("line"),
+		jsontext.Int(int64(sl.Line)),
```
ajwerner: These need to be `"fileName"` and `"lineNumber"`

https://docs.google.com/document/d/10ViZr1Tgj3L8tWBqHw9krPjraYxXvJE-2FZXOys8NVI/edit?tab=t.0#bookmark=id.ji3xy4t58gn6

pkg/dyninst/decode/marshal.go
```diff
@@ -0,0 +1,485 @@
...
+			}
+		}
+		stringValue, ok := dataItems[typeAndAddr{
+			irType: uint32(v.Data.GetID()),
+			addr:   address,
+		}]
+		if !ok {
+			// TODO: Add not captured reason to the snapshot
+			if err := writeTokens(enc,
+				jsontext.String("NULL")); err != nil {
```
ajwerner: is this definitely the right thing? Should it not be a json `null`?

pkg/dyninst/decode/marshal.go
```diff
@@ -0,0 +1,485 @@
...
+		return writeTokens(enc, jsontext.String("unimplemented - hmap header type"))
+	case *ir.GoHMapBucketType:
+		return writeTokens(enc, jsontext.String("unimplemented - hmap bucket type"))
+	case *ir.GoSwissMapHeaderType:
+		return writeTokens(enc, jsontext.String("unimplemented - swiss map header type"))
+	case *ir.GoSwissMapGroupsType:
+		return writeTokens(enc, jsontext.String("unimplemented - swiss map groups type"))
+	case *ir.EventRootType:
+		return writeTokens(enc, jsontext.String("unimplemented - event root type"))
+	default:
```
ajwerner: I think you're missing types here. Get rid of the default case and have the IDE fill in all the cases. You'll see that you're missing several:

```
	case *ir.GoSliceDataType:
	case *ir.GoStringDataType:
	case *ir.GoSubroutineType:
```
It's unexpected to see `*ir.GoSliceDataType` or `*ir.GoStringDataType` -- so maybe we return errors for those, but `*ir.GoSubroutineType` is pretty common. Let's for now say it's unimplemented. Later we can support at least saying what the function is in terms of its line and name.

pkg/dyninst/decode/marshal.go
```diff
@@ -0,0 +1,485 @@
...
+	case reflect.Ptr:
+		return errors.New("ptrs are not base types")
+	case reflect.Slice:
+		return errors.New("slices are not base types")
+	case reflect.String:
+		return errors.New("strings are not base types")
+	case reflect.Struct:
+		return errors.New("structs are not base types")
+	case reflect.UnsafePointer:
+		return errors.New("unsafe pointers are not base types")
```
ajwerner: how about just:
```go
default:
    return fmt.Errorf("%s is not a base type", kind)
```

pkg/dyninst/integration_test.go
```diff
@@ -194,51 +200,150 @@ func testDyninst(
...
-		}
-
-		if _, ok := tmpMap["stack_frames"]; !ok {
-			t.Error("No stack frames in output")
-		} else {
-			tmpMap["stack_frames"] = ""
-		}
 
-		clearAddressFields(tmpMap)
+		fmt.Println(decodeOut.String())
```
ajwerner: ?

pkg/dyninst/integration_test.go
```diff
@@ -194,51 +200,150 @@ func testDyninst(
...
+			if middle == "" {
+				return false
+			}
+			for _, r := range middle {
+				if r < '0' || r > '9' {
+					return false
+				}
+			}
+			return true
+		},
```
ajwerner: perhaps at this point a regex would be nicer:
```go
func redactPtrRegexp(pat string, replacement jsontext.Value) jsonRedactor {
        re := regexp.MustCompile(pat)
	return jsonRedactor{
               matches: re.MatchString,
               replacement: replacement,
         }
 }
 ```
 
 Then:
 ```go
 redactPtrRegexp(`^/debugger/snapshot/stack/[0-9]+/file$`, []byte(`"<line>"`))
 ```

**Review by ajwerner — COMMENTED** [review]
A bit more feedback as I learn what the frontend wants.

File: pkg/dyninst/decode/marshal.go [hunk 53]
```diff
@@ -0,0 +1,485 @@
...
+	// metadata and raw bytes of the parameters of this function.
+	for _, expr := range ad.rootType.Expressions {
+		parameterType := expr.Expression.Type
+		parameterData := ad.rootData[expr.Offset : expr.Offset+parameterType.GetByteSize()]
+		if err = writeTokens(enc,
+			jsontext.String(expr.Name),
+			jsontext.BeginObject,
+			jsontext.String("type"),
+			jsontext.String(parameterType.GetName()),
+			jsontext.String("value"),
```
ajwerner: I'm starting to feel like we've made a mistake in putting this here -- doesn't it belong in the code that marshals the value?

Like, for example, when it's a pointer and it's a nil value, we don't want to be filling in value, we want `"isNull": true`

ajwerner (commented):
I think generally we've gotten the format wrong when it comes to `encodeValue`. You need to have the envelopes recursively. Like at every level there needs to be an object with `{"type": <type_name>, "value": <value>[, "isNull": true][,"notCapturedReason":<...>}`. We've been dragging this on for a while, but also it feels like a pretty minor and small change. If you can't fit it into this one, we need it ASAP because as it stands compound structures break the UI. Let's get some more tests with some more complex pointer graphs in play ASAP too.

grantseltzer:
>I think generally we've gotten the format wrong when it comes to encodeValue. You need to have the envelopes recursively. Like at every level there needs to be an object with {"type": <type_name>, "value": <value>[, "isNull": true][,"notCapturedReason":<...>}. We've been dragging this on for a while, but also it feels like a pretty minor and small change. If you can't fit it into this one, we need it ASAP because as it stands compound structures break the UI. Let's get some more tests with some more complex pointer graphs in play ASAP too.

Thanks for pointing this out. I've fixed this. Another nuance to it is that structs expect a "fields" key instead of value, slices/arrays expect "elements". Everything else is "value".

pkg/dyninst/decode/decoder.go
grantseltzer: I'm not entirely sure if what I just changed is what you pictured in terms of when things get cleared, but moving addressReferenceCount to decoder does make a lot of sense so me and have done that.

pkg/ebpf/bytecode/runtime/dyninst_event.go
```diff
@@ -0,0 +1,6 @@
+// Code generated by go generate; DO NOT EDIT.
+//go:build linux_bpf
+
+package runtime
+
+var Dyninstevent = newAsset("dyninst_event.c", "e3bb0a4c3ee3064be038d1071cefe2e84bd7ddfff4bbb2fb28fa108e1d622e79")
```
grantseltzer: Is this supposed to be commited? What is this?

ajwerner: No it should have been gitignored

ajwerner (approved):
I think the deal is that file is gone with Piotr’s change and probably the gitignore with it. Delete it and this is ready to go.

PR #38234, opened 2025-06-25
[pkg/dyninst/test] Stabilize integration test output

piob-io:
Surprisingly go-experiment-json by default doesn't sort maps by keys. Even though test logic is not sensitive to these orders, we want to avoid changing snapshot files when relevant code behavior stays the same.

### DIFF
--- a/pkg/dyninst/integration_test.go
+++ b/pkg/dyninst/integration_test.go
@@ -11,6 +11,7 @@ import (
 	"bytes"
 	"context"
 	"embed"
+	"encoding/json"
 	"io"
 	"math"
 	"os"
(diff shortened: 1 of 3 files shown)


tylfin (approved):

ajwerner (approved):

ajwerner:
At some point here we should have the test be sensitive to the output key ordering. I suspect that'll fall out of the evolution of the test organically.

piob-io:
> At some point here we should have the test be sensitive to the output key ordering. I suspect that'll fall out of the evolution of the test organically.

Yup. Once we have "run all probes together" test case, that could define the order in the output file. For now we need to have some source of order.

PR #38239, opened 2025-06-26
dyninst: fixes ahead of plumbing

ajwerner:
#### dyninst/compiler: encapsulate dependency

We'll need to take the ebpf.Config for the module. Also this makes it
more convenient to plumb other options into the compiler.

#### dyninst/rcjson: use "internal" camelCase

Prior to this change, we were using the "external" json API which uses
snake_case. This change adops the "internal" API as used in the remote config
documents. These use camelCase.

#### dyninst/rcjson,irgen: abstract probe configuration

This is in advance of having RC scraping probes. This also
fixes some conformance issues and carves out an explicit
snapshot probe type.

### Motivation

Carving out a PR ahead of some other PRs that set up plumbing for the dyninst module. I wanted to break up the reviews a tad. Hopefully these are broadly uncontroversial.

Relates to https://datadoghq.atlassian.net/browse/DEBUG-4039


pkg/dyninst/irgen/probe_definition.go
```diff
@@ -0,0 +1,101 @@
...
+package irgen
+
+import (
+	"math"
+
+	"github.com/DataDog/datadog-agent/pkg/dyninst/ir"
+	"github.com/DataDog/datadog-agent/pkg/dyninst/rcjson"
+)
+
+// ProbeDefinition abstracts the configuration of a probe.
```
ajwerner, line 17: I considered having this live in yet another package so that code that depends on defining the configuration for probes don't have to also depend on ir generation, but then I decided against it. If y'all think it's worthwhile, I'll suggest that we still leave it for a follow-up.

pkg/dyninst/actuator/configuration.go
```diff
@@ -147,3 +161,18 @@ func WithCompiledCallback(f CompiledCallback) Option {
 		s.compiledCallback = f
 	})
 }
+
+// Compiler is an interface that abstracts the eBPF compiler.
```
piob-io, line 165: Maybe this belongs in the compiler package? Although there is not actual package dependency, logically compiler depends on actuator, to match this interface, while I'd expect actuator to be a "top" package that depends on other things, only in this one direction. If there are other obstacles to make that work, and this is the simplest way to glue these together, this is fine.

pkg/dyninst/irgen/probe_definition.go
piob-io: Discussed offline. We can move interface to ir and implementation to rcjson, up to you when to do that.

piob-io (approved):
Thanks!

pkg/dyninst/actuator/configuration.go
ajwerner: Meh, discussed offline, not doing it here.

ajwerner:
Thanks for the review, will follow up with a fix promptly.

PR #38244, opened 2025-06-26
Refactor agent docker image build jobs in CI

alopezz:
### What does this PR do?

Refactor docker image build jobs separating concerns into templates and reducing repetition.

### Motivation

Somewhat inspired by [this comment](https://github.com/DataDog/datadog-agent/pull/38209/files/b61b4d4dc1be5d32d4192d845ec6635e7b9fe5ed#r2168529028) and by some recent changes I had to do to these jobs, where certain things had to be copy-pasted in many places.

Refactoring this into purpose-specific templates will reduce chances of future copy-paste related errors and reduce the number of places where certain changes might be needed.

I'm trusting the [gitlab ci configuration changes](https://github.com/DataDog/datadog-agent/pull/38244#issuecomment-3008039455) check here, which indicates that no actual jobs were modified, only templates.

### DIFF
--- a/.gitlab/container_build/docker_linux.yml
+++ b/.gitlab/container_build/docker_linux.yml
@@ -124,20 +124,24 @@
     ARCH: arm64
 
 # build agent7 image
-docker_build_agent7:
-  extends: [.docker_build_job_definition_amd64, .docker_build_artifact]
+.docker_build_agent7:
+  extends: [.docker_build_job_definition, .docker_build_artifact]
   rules:
     - !reference [.except_mergequeue]
     - when: on_success
-  needs:
-    - job: datadog-agent-7-x64
   variables:
     IMAGE: registry.ddbuild.io/ci/datadog-agent/agent
     BUILD_CONTEXT: Dockerfiles/agent
-    TAG_SUFFIX: -7
     TARGET_ARG: --target test
-    BUILD_ARG: --build-arg DD_AGENT_ARTIFACT=datadog-agent-7*-amd64.tar.xz
     CACHE_TARGET: release-base
+
+docker_build_agent7:
+  extends: [.docker_build_agent7, .docker_build_job_definition_amd64]
+  needs:
+    - job: datadog-agent-7-x64
+  variables:
+    TAG_SUFFIX: -7
+    BUILD_ARG: --build-arg DD_AGENT_ARTIFACT=datadog-agent-7*-amd64.tar.xz
 
 # TODO: Move this job to .gitlab/deploy_containers/deploy_containers_a7.yml.
 # This cannot be done now because of the following reasons:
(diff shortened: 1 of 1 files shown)


.gitlab/container_build/docker_linux.yml
```diff
@@ -109,36 +109,39 @@
...
-  extends: .docker_build_job_definition
+.docker_build_amd64:
   image: registry.ddbuild.io/ci/datadog-agent-buildimages/docker_x64$CI_IMAGE_DOCKER_X64_SUFFIX:$CI_IMAGE_DOCKER_X64
   tags: ["arch:amd64"]
   variables:
     ARCH: amd64
 
-.docker_build_job_definition_arm64:
+.docker_build_arm64:
   extends: .docker_build_job_definition
```
rdesgroppes: I guess this line should be removed as well?

alopezz: Indeed, I missed it! I'll fix that in a sec.

.gitlab/container_build/docker_linux.yml
```diff
@@ -109,36 +109,38 @@
...
     TARGET_ARG: --target test
-    BUILD_ARG: --build-arg DD_AGENT_ARTIFACT=datadog-agent-7*-amd64.tar.xz
     CACHE_TARGET: release-base
 
+docker_build_agent7:
+  extends: [.docker_build_amd64, .docker_build_agent7]
+  needs:
+    - job: datadog-agent-7-x64
+  variables:
+    TAG_SUFFIX: -7
```
chouquette, line 141: Do we still need to care about agent version?

All jobs seem to include `-7` now, so I wonder if we couldn't have `-7` in the common part, and keep `-fips`/`-jmx` only as the suffix, and rename `docker_build_agent7` to `docker_build_agent`

.gitlab/container_build/docker_linux.yml
```diff
@@ -109,36 +109,38 @@
...
-    - job: datadog-agent-7-x64
   variables:
     IMAGE: registry.ddbuild.io/ci/datadog-agent/agent
     BUILD_CONTEXT: Dockerfiles/agent
-    TAG_SUFFIX: -7
     TARGET_ARG: --target test
-    BUILD_ARG: --build-arg DD_AGENT_ARTIFACT=datadog-agent-7*-amd64.tar.xz
     CACHE_TARGET: release-base
 
+docker_build_agent7:
```
chouquette, line 136: This might need more work, and isn't required here/now, but I wonder if we shouldn't include the arch in the job name, for consistency with the arm jobs

rdesgroppes (approved):
LGTM - thanks for factoring out!

chouquette (approved):
some minor nits, but that's already a very nice cleanup! Thanks!

.gitlab/container_build/docker_linux.yml
alopezz: I do think it would make sense and I also find the inconsistency somewhat jarring. But changing job names might add a layer of annoyance that I prefer to deal with at a separate time 😬 .

chouquette: very fair & understandable :sweat_smile:

.gitlab/container_build/docker_linux.yml
alopezz: I agree that it would be nice but is a bit annoying to deal with given how the final image + tag combination is constructed generically (note that this template is used by every docker build, including those that don't have a version 7 distinction):
https://github.com/DataDog/datadog-agent/blob/4c9bed1c93182c0110c0afb039f55d2c2f321a30/.gitlab/container_build/docker_linux.yml#L15

PR #38246, opened 2025-06-26
[pkg/dyninst/test] Drop expectedToFail

piob-io:
It's an input field in an output file. Simplify, just comment out input file, both offer same level of visibility (characters in yaml file).


ajwerner (approved):
LGTM

PR #38254, opened 2025-06-26
[EBPF-598] gpu: add GPM collector

gjulianm:
### What does this PR do?

Adds support for the GPM collector.

### Motivation

Support advanced metrics for Hopper (and newer) GPUs.

Included unit tests, also validated in an instance with H100 GPUs.

### Additional Notes

MIG support will be added later


pkg/collector/corechecks/gpu/nvidia/gpm.go
```diff
@@ -0,0 +1,227 @@
...
+}
+
+type gpmMetric struct {
+	name       string
+	metricType metrics.MetricType
+}
+
+var allGpmMetrics = map[nvml.GpmMetricId]gpmMetric{
+	nvml.GPM_METRIC_GRAPHICS_UTIL: {
+		name:       "gr_engine_active",
```
gjulianm, line 39: We already have metrics with this name. I want to do a separate PR adding a prioritization mechanism to the collectors, so that in cases of conflicts we discard metrics of the less precise methods.

Maybe in this specific case we just want to remove the other one as we saw the API might not be returning the values we expect, but others (e.g., PCI bandwidth, which is not present here) will probably end up duplicated anyways.

pkg/collector/corechecks/gpu/nvidia/collector_test.go
```diff
@@ -135,3 +134,32 @@ func TestGetDeviceTagsMapping(t *testing.T) {
 		})
 	}
 }
+
+func TestAllCollectorsWork(t *testing.T) {
```
val06, line 138: that's nice! perhaps we can add a test to check the collectors work if one collector fails to initialize (test our best effort behavior for the gpu core-check)

pkg/collector/corechecks/gpu/nvidia/gpm.go
```diff
@@ -0,0 +1,237 @@
...
+	"github.com/hashicorp/go-multierror"
+
+	ddnvml "github.com/DataDog/datadog-agent/pkg/gpu/safenvml"
+	"github.com/DataDog/datadog-agent/pkg/metrics"
+)
+
+const sampleBufferSize = 2
+
+type gpmCollector struct {
+	lib                 ddnvml.SafeNVML
```
val06, line 25: why do we have the lib here? it is better to keep the collector as slim as possible (also for the future refactor), rest of the collectors don't really need the lib reference

pkg/collector/corechecks/gpu/nvidia/gpm.go
val06: hmm I think we should merge the other PR then, to avoid some conflicts and inaccurate values

pkg/collector/corechecks/gpu/nvidia/gpm.go
```diff
@@ -0,0 +1,237 @@
...
+	},
+	nvml.GPM_METRIC_SM_UTIL: {
+		name:       "sm_active",
+		metricType: metrics.GaugeType,
+	},
+	nvml.GPM_METRIC_SM_OCCUPANCY: {
+		name:       "sm_occupancy",
+		metricType: metrics.GaugeType,
+	},
+	nvml.GPM_METRIC_INTEGER_UTIL: {
```
val06, line 50: what is this metric? was there a request to report it?

pkg/collector/corechecks/gpu/nvidia/gpm.go
```diff
@@ -0,0 +1,237 @@
...
+	nvml.GPM_METRIC_ANY_TENSOR_UTIL: {
+		name:       "tensor_active",
+		metricType: metrics.GaugeType,
+	},
+}
+
+func newGPMCollector(device ddnvml.SafeDevice) (c Collector, err error) {
+	support, err := device.GpmQueryDeviceSupport()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get GPM support: %w", err)
```
val06:
```suggestion
		return nil, fmt.Errorf("failed to query for GPM support: %w", err)
```

pkg/gpu/testutil/mocks.go
```diff
@@ -158,6 +160,16 @@ func GetDeviceMock(deviceIdx int, opts ...func(*nvmlmock.Device)) *nvmlmock.Devi
...
 			return false, nvml.SUCCESS
 		},
+		GetSamplesFunc: func(_ nvml.SamplingType, _ uint64) (nvml.ValueType, []nvml.Sample, nvml.Return) {
+			samples := []nvml.Sample{
+				{TimeStamp: 1000, SampleValue: [8]byte{0, 0, 0, 0, 0, 0, 0, 1}},
+				{TimeStamp: 2000, SampleValue: [8]byte{0, 0, 0, 0, 0, 0, 0, 2}},
+			}
+			return nvml.VALUE_TYPE_UNSIGNED_INT, samples, nvml.SUCCESS
+		},
+		GpmQueryDeviceSupportFunc: func() (nvml.GpmSupport, nvml.Return) {
```
val06, line 170: do we always want to return true here or some conditional logic to test unsupported case as well?

pkg/gpu/testutil/mocks.go
```diff
@@ -291,6 +308,54 @@ func WithSymbolsMock(availableSymbols map[string]struct{}) NvmlMockOption {
 	}
 }
 
+// WithMockAllFunctions returns an option that creates basic functions for all nvmlmock.Device.*Func attributes
+// that return nil/zero values. This is useful for ensuring all functions are mocked even if not explicitly set.
+// This is not the default behavior of the mock, as we want explicit errors if we use a function that is not mocked
+// so that we implement the mocked method explicitly, controlling the inputs and outputs. However, in some cases
+// (e.g., testing the collectors) we want to ensure that all functions are mocked without caring too much about the inputs and outputs.
```
val06, line 315: why in case of testing the collectors we want to have all functions mocked? I would expect a strict mocking as well, to ensure the reliability of the test

pkg/gpu/safenvml/lib.go
```diff
@@ -43,6 +43,11 @@ func getNonCriticalAPIs() []string {
 	return []string{
 		"nvmlShutdown",
 		"nvmlSystemGetDriverVersion",
+		"nvmlGpmSampleAlloc",
+		"nvmlGpmSampleFree",
+		"nvmlGpmMetricsGet",
+		"nvmlGpmQueryDeviceSupport",
+		"nvmlGpmSampleGet",
```
val06, line 50: do we need the `nvmlGpmSampleGet` here and in the safenvml.device struct?

pkg/gpu/safenvml/device.go
```diff
@@ -83,6 +83,10 @@ type SafeDevice interface {
 	GetUtilizationRates() (nvml.Utilization, error)
 	// IsMigDeviceHandle returns true if the device is a MIG device or false for a physical device
 	IsMigDeviceHandle() (bool, error)
+	// GpmQueryDeviceSupport returns true if the device supports GPM
+	GpmQueryDeviceSupport() (nvml.GpmSupport, error)
+	// GpmSampleGet gets a sample for GPM
+	GpmSampleGet(sample nvml.GpmSample) error
```
val06, line 89: I think you should also add `GpmMigSampleGet` ([ref](https://docs.nvidia.com/deploy/nvml-api/group__nvmlGpmFunctions.html#group__nvmlGpmFunctions_1ge7880c391d245f640e5cb7633c6db1fe)) to support MIG devices

pkg/collector/corechecks/gpu/nvidia/gpm.go
```diff
@@ -0,0 +1,237 @@
...
+
+	if support.IsSupportedDevice == 0 {
+		return nil, errUnsupportedDevice
+	}
+
+	// Clone the global allGpmMetrics map to avoid mutating global state
+	clonedMetrics := make(map[nvml.GpmMetricId]gpmMetric, len(allGpmMetrics))
+	for key, value := range allGpmMetrics {
+		clonedMetrics[key] = value
+	}
```
val06: use `maps.Copy`

pkg/collector/corechecks/gpu/nvidia/gpm.go
```diff
@@ -0,0 +1,237 @@
...
+	ddnvml "github.com/DataDog/datadog-agent/pkg/gpu/safenvml"
+	"github.com/DataDog/datadog-agent/pkg/metrics"
+)
+
+const sampleBufferSize = 2
+
+type gpmCollector struct {
+	lib                 ddnvml.SafeNVML
+	device              ddnvml.SafeDevice
+	samples             []nvml.GpmSample
```
val06:
```suggestion
	samples             [sampleBufferSize]nvml.GpmSample
```

pkg/collector/corechecks/gpu/nvidia/gpm.go
```diff
@@ -0,0 +1,237 @@
...
+			collector.freeSamples()
+		}
+	}()
+
+	for i := 0; i < sampleBufferSize; i++ {
+		sample, err := collector.lib.GpmSampleAlloc()
+		if err != nil {
+			return nil, fmt.Errorf("failed to allocate GPM sample: %w", err)
+		}
+		collector.samples = append(collector.samples, sample)
```
val06: collector.samples[i] = sample (together with my proposal to bound the slice in the declaration

pkg/collector/corechecks/gpu/nvidia/gpm.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	if err := collector.removeUnsupportedMetrics(); err != nil {
+		return nil, fmt.Errorf("failed to remove unsupported metrics: %w", err)
```
val06: do we need to bail out here?

pkg/collector/corechecks/gpu/nvidia/gpm.go
```diff
@@ -0,0 +1,237 @@
...
+	if len(collector.metricsToCollect) == 0 {
+		return nil, errUnsupportedDevice
+	}
+
+	return collector, nil
+}
+
+func (c *gpmCollector) removeUnsupportedMetrics() error {
+	// Now collect two samples and try to get the metrics, to discard any unsupported ones
+	for i := 0; i < 2; i++ {
```
val06, line 123: replace the 2 with `sampleBufferSize`

pkg/collector/corechecks/gpu/nvidia/gpm.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+}
+
+func (c *gpmCollector) getLastTwoSamples() (nvml.GpmSample, nvml.GpmSample) {
+	// add sampleBufferSize to avoid negative indices
+	lastCollected := (c.nextSampleToCollect - 1 + sampleBufferSize) % sampleBufferSize
+	secondLastCollected := (c.nextSampleToCollect - 2 + sampleBufferSize) % sampleBufferSize
+
+	return c.samples[lastCollected], c.samples[secondLastCollected]
+}
```
val06: this looks a bit clumsy, we have 2 sample anyway, let's rename the function and also return a fixed sized slice instead of 2 separate return values.

also, what is the purpose of this function? if we collect 2 samples anyway, why not just using the samples field in the collector?

pkg/collector/corechecks/gpu/nvidia/gpm.go
```diff
@@ -0,0 +1,237 @@
...
+
+func (c *gpmCollector) calculateGpmMetrics() (*nvml.GpmMetricsGetType, error) {
+	sample1, sample2 := c.getLastTwoSamples()
+	metricsGet := &nvml.GpmMetricsGetType{
+		NumMetrics: uint32(len(c.metricsToCollect)),
+		Version:    nvml.GPM_METRICS_GET_VERSION,
+		Sample1:    sample1,
+		Sample2:    sample2,
+		Metrics:    [98]nvml.GpmMetric{},
```
val06: what is 98?

val06 (changes requested):
reviewed

pkg/collector/corechecks/gpu/nvidia/gpm.go
gjulianm: Unfortunately some of the GPM methods (sample allocation/free) are only exposed in the library, not the device.

pkg/collector/corechecks/gpu/nvidia/gpm.go
gjulianm: It's the utilization of the integer calculation engine. It's not requested, but I added it for completeness as we're reporting the utilization of the other calculation engines (floating point, tensor).

pkg/gpu/testutil/mocks.go
gjulianm: By default, yes. In `gpm_test.go` there's a test where we disable the support and ensure the logic is correct.

pkg/gpu/testutil/mocks.go
gjulianm: For that test we'd need to mock quite a lot of functions (specially for the device collector) without actually caring about the values, because we don't validate that here. I think this is cleaner, as we won't need to update the mocks continuously when we're adding new functions.

pkg/gpu/safenvml/lib.go
gjulianm: This is a bit confusing, `nvmlGpmSampleGet` is a device function but it doesn't have the `nvmlDevice` prefix.

pkg/gpu/safenvml/device.go
gjulianm: I planned to add MIG support separately, as I'm not sure how that works yet.

pkg/collector/corechecks/gpu/nvidia/gpm.go
gjulianm: TIL :D

pkg/collector/corechecks/gpu/nvidia/gpm.go
gjulianm: An error here comes either from collecting samples or from calculating the metrics, both of them are likely permanent errors.

pkg/collector/corechecks/gpu/nvidia/gpm.go
gjulianm: In this case it's just because we want to collect two samples (the two needed for the value calculation), independently of the buffer size.

pkg/collector/corechecks/gpu/nvidia/gpm.go
gjulianm: The samples need to rotate: at each `Collect` call, we get one data sample, and then compute values from that sample and the one that was retrieved in the previous `Collect` call.

pkg/collector/corechecks/gpu/nvidia/gpm.go
gjulianm: `Metrics` is a fixed size array, although I noticed I can just leave it to default initialization.

pkg/gpu/safenvml/lib.go
val06: in the go-nvml there are actually 2 wrappers, I think you can remove this one from the lib and just keep the one in the device_impl.go

pkg/gpu/safenvml/device.go
val06: perhaps add a comment to the PR description then

pkg/collector/corechecks/gpu/nvidia/gpm.go
val06: i prefer a more consistent behavior between our collectors, i think rest of them don't return any error in the removeUnsupportedMetrics func (apart from fields collector).

let's make a decision here and align all collectors. I personally think that it should not return any error and be a best-effort for optimization purposes. wdyt?

pkg/collector/corechecks/gpu/nvidia/gpm.go
val06: you can achieve the rotation via other ways, using a variable to track the active one (the other one will be inactive), or perhaps extending the sample struct by embedding the original and adding `is_last` field

pkg/gpu/safenvml/lib.go
gjulianm: In this case this is just for the list of APIs to search for. The device only has `GpmSampleGet`, in the lib we have `GpmMetricsGet` only.

pkg/gpu/safenvml/device.go
gjulianm: Done

estherk15 (approved):

pkg/collector/corechecks/gpu/nvidia/gpm.go
val06: done [here](https://github.com/DataDog/datadog-agent/pull/38369), resolving

pkg/collector/corechecks/gpu/nvidia/gpm.go
gjulianm: I prefer this method to be honest, tracking the "last collected" and the "next to collect" with a variable forces us to put `if/else` in the places where we access them instead of direct index access. In this case, the only thing that makes it look a bit weirder is the ` + sampleBufferSize` to avoid overflows, but for the rest it looks like access to a circular buffer.

val06: ok, in that case can you add an explicit documentation line above the function to explain the return values, as the convention is not very standard.

and also, perhaps we still should find a bit better name than `getLastTwoSamples`, maybe just `getSamples` or `getMetricsSamples` ?

pkg/collector/corechecks/gpu/nvidia/gpm.go
gjulianm: Sounds good, changed and added a comment

pkg/collector/corechecks/gpu/nvidia/gpm.go
gjulianm: Added the comments. The name looks good to me, `getSamples` or `getMetricsSamples` could look as if it's returning all samples.

val06 (approved):

PR #38256, opened 2025-06-26
Add Virus Total scanning for MSI artifacts

julien-lebot:
### What does this PR do?
This PR adds some jobs that will submit the Datadog Agent MSI to Virus Total for scanning.

### Motivation
https://datadoghq.atlassian.net/browse/WINA-730

### Additional Notes

If the scan job fails it doesn't fail the pipeline.

The job currently fails because there are some false positives.


.gitlab/scan/windows.yml
```diff
@@ -0,0 +1,117 @@
...
+    - chmod +x vt
+    - ./vt version
+    # Find the MSI file for this pipeline
+    - MSI_FILE=$(find "$OMNIBUS_PACKAGE_DIR" -name "$SCAN_ARTIFACT_PATTERN" | head -n 1)
+    - |
+      if [ -z "$MSI_FILE" ]; then
+        echo "No MSI file found for pattern $SCAN_ARTIFACT_PATTERN and pipeline ID $CI_PIPELINE_ID"
+        exit 1
+      fi
+    - VT_API_KEY=$($CI_PROJECT_DIR/tools/ci/fetch_secret.sh $VT_API_KEY) || exit $?; export VT_API_KEY
```
KevinFairise2: If this code can be executed in a linux runner, I think we should move the secret to Vault instead of SSM.

.gitlab/scan/windows.yml
```diff
@@ -0,0 +1,117 @@
...
+              exit 0
+              fi
+          fi
+      done
+
+scan_packages_windows-x64:
+  rules:
+    - !reference [.on_deploy_nightly_repo_branch]
+    - !reference [.on_main]
+    - !reference [.except_mergequeue]
```
KevinFairise2: With these rules the job should be executed on deploy pipelines, and on all main branches pipelines, but not on release and rc pipelines. Is that expected? Would it be interesting to do the scanning on rc and release pipelines?

julien-lebot: Updated thanks!

.gitlab/scan/windows.yml
julien-lebot: Done!

KevinFairise2 (approved):

PR #38260, opened 2025-06-26
dyninst: create norms around setting up logging

ajwerner:
The primary thing here is to set up a utility function to set up logging consistently. The intention is to call this setup function in a TestMain.

Example of the default format:
```
d 16:03:28.177952641 @actuator.go:171| event: eventProgramAttached{programID: 1, processID: {PID:113694}}
d 16:03:28.179447592 @actuator.go:144| sending update: {[] [{PID:113694}]}
d 16:03:28.179522307 @actuator.go:171| event: eventProcessesUpdated{updated: 0, removed: 1}
d 16:03:28.179627972 @actuator.go:168| Received shutdown event
```
Json format:
```
{"time":1750953833754972794,"level":"Debug","msg":"event: eventProgramAttached{programID: 1, processID: {PID:113932}}","path":"actuator/actuator.go","func":"github.com/DataDog/datadog-agent/pkg/dyninst/actuator.(*Actuator).runEventProcessor","line":171}
{"time":1750953833757064633,"level":"Debug","msg":"sending update: {[] [{PID:113932}]}","path":"actuator/actuator.go","func":"github.com/DataDog/datadog-agent/pkg/dyninst/actuator.(*Actuator).HandleUpdate","line":144}
{"time":1750953833757162786,"level":"Debug","msg":"event: eventProcessesUpdated{updated: 0, removed: 1}","path":"actuator/actuator.go","func":"github.com/DataDog/datadog-agent/pkg/dyninst/actuator.(*Actuator).runEventProcessor","line":171}
{"time":1750953833757195137,"level":"Debug","msg":"Received shutdown event","path":"actuator/actuator.go","func":"github.com/DataDog/datadog-agent/pkg/dyninst/actuator.(*Actuator).runEventProcessor","line":168}
```
Relates to https://datadoghq.atlassian.net/browse/DEBUG-4107

### DIFF
--- a/pkg/dyninst/actuator/main_test.go
+++ b/pkg/dyninst/actuator/main_test.go
@@ -1,0 +1,20 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+//go:build linux_bpf
+
+package actuator
+
+import (
+	"os"
+	"testing"
+
+	"github.com/DataDog/datadog-agent/pkg/dyninst/dyninsttest"
+)
+
+func TestMain(m *testing.M) {
+	dyninsttest.SetupLogging()
+	os.Exit(m.Run())
+}
--- a/pkg/dyninst/dyninsttest/util.go
+++ b/pkg/dyninst/dyninsttest/util.go
@@ -10,6 +10,7 @@ package dyninsttest
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"os"
 	"os/exec"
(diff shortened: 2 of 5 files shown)


pkg/dyninst/actuator/main_test.go
piob-io, line 1: If this is common, so be it, but "test_main.go" would make more sense to me - we are not testing the main, we are setting up a main for tests.

pkg/dyninst/dyninsttest/util.go
```diff
@@ -28,13 +29,43 @@ import (
...
 	"github.com/DataDog/datadog-agent/pkg/dyninst/object"
 	"github.com/DataDog/datadog-agent/pkg/dyninst/testprogs"
+	"github.com/DataDog/datadog-agent/pkg/util/log"
 	"github.com/DataDog/datadog-agent/pkg/util/safeelf"
 )
 
+// SetupLogging is used to have a consistent logging setup for all tests.
+// It is best to call this in TestMain.
+func SetupLogging() {
+	logLevel := os.Getenv("DD_LOG_LEVEL")
```
piob-io, line 39: I'm unsure about this being a default... Are all prod deployments mindful to set it to something less than debug? For tests, perhaps this function could take an argument to force enable debug.

pkg/dyninst/main_test.go
piob-io, line 1: ditto

piob-io (approved):

pkg/dyninst/actuator/main_test.go
ajwerner: well it has to be `_test.go` to be a testing only file... This is a well established norm in go: https://github.com/search?q=repo%3Agolang%2Fgo%20path%3Amain_test.go&type=code

pkg/dyninst/dyninsttest/util.go
ajwerner: Fair, changed it to info.

pkg/dyninst/main_test.go
ajwerner: ditto

pkg/dyninst/dyninsttest/util.go
```diff
@@ -28,13 +29,43 @@ import (
...
+	case "":
+		format = defaultFormat
+	case "json":
+		format = `{"time":%Ns,"level":"%Level","msg":"%Msg","path":"%RelFile","func":"%Func","line":%Line}%n`
+	case "json-short":
+		format = `{"t":%Ns,"l":"%Lev","m":"%Msg"}%n`
+	default:
+		format = formatFromEnv
+	}
+	logger, err := log.LoggerFromWriterWithMinLevelAndFormat(
```
tylfin, line 55: I don't know how much it breaks the repository conventions, but if we're changing logging you could consider throwing in [slog](https://go.dev/blog/slog) for structured logging.

ajwerner: We don't have a choice here -- the agent is 100% bought in on the logging framework it uses and we must use it.

PR #38262, opened 2025-06-26
Use a string value in json instead of returning an error for unimplemented types in decoding (+bug fixes)

grantseltzer:
### What does this PR do?

- Use a string value in json instead of returning an error for unimplemented types in decoding

- Fixes two issues with decoding:
    1. There was a check while iterating over data items which would cut off iterating in the case that the index+size of a data item is >= to the size of raw buffer. This was a wrong calculation since the types are being casted from an unsafe pointer, rather than indexing a slice. These checks may not be needed at all given alignment checks elsewhere, but keeping them in case they're needed in packages other than decoding which i'm not as familiar with.
    1. When pointers are cut off due to capture depth, the data item corresponding to its address wouldn't exist. So I tweaked the logic to not error out in the case of a data item not existing in the address reference map.

- Fixes a test data issue in gosym:
    1. The gosym tests had environment specific paths in the testdata. For some reason my env is clearly set up differently from other dev machines, and certainly the CI. I stripped the leading paths on the symbol names to just the filename to avoid this from happening. Updated testdata accordingly.
  
### Motivation

Not getting errors.


ajwerner:
Can we uncomment out some probes now?

grantseltzer:
> Can we uncomment out some probes now?

Ah yea - now that Piotr removed the expectedToFail we should be able to. On it.

ajwerner (approved):
LGTM

ajwerner:
I think you need to rebase, rebuild the programs and object files and then rewrite...

pkg/dyninst/gosym/testdata/snapshot/simple.arch=amd64,toolchain=go1.24.3.out
```diff
@@ -1,17 +1,17 @@
 pc: 0x4a8060
 	main.intArg@pkg/dyninst/testprogs/progs/simple/main.go:41
 pc: 0x4a8091
-	fmt.Println@pkg/mod/golang.org/toolchain@*/src/fmt/print.go:314
+	fmt.Println@/usr/local/go/src/fmt/print.go:314
```
ajwerner: hmm, this seems perhaps suspect

grantseltzer: Ugh yes we need to fix this - not sure if I need to change my environment or a larger issue. We talked briefly about it yesterday.

piob-io: I'd suggest keeping the local environment that suits you and expand the logic in the test itself that tries to strip env-specific path prefix.

grantseltzer: I'm going to revert the change to this test file for now to keep tests passing and make the redaction part of a separate PR before this gets too big.

grantseltzer:
~~I'm investigating the test errors~~

~~Seems I need to possibly update testdata related to the bug fix for data item index check. And also I may not have reverted the snapshot testdata accurately.~~

I've added a fix for the gosym failing tests as well.

pkg/dyninst/gosym/testdata/snapshot/simple.arch=amd64,toolchain=go1.24.3.out
grantseltzer: ~~Ok I need to do it in this PR because the testdata needs to be updated regardless.~~ Done.

pkg/dyninst/decode/decoder.go
```diff
@@ -219,12 +217,23 @@ func (d *Decoder) encodeValue(
...
+
+		var (
+			header         *output.DataItemHeader
+			pointedData    []byte
+			dontParseValue bool
+		)
 		pointedValue, ok := dataItems[key]
 		if !ok {
-			return errors.New("pointer not found in address pass map")
+			header = &output.DataItemHeader{
```
ajwerner, line 228: this allocation is sort of unfortunate. Let's eliminate it in the followup

pkg/dyninst/decode/decoder.go
```diff
@@ -219,12 +217,23 @@ func (d *Decoder) encodeValue(
...
+				Address: key.addr,
+			}
+			dontParseValue = true
+		} else {
+			header = pointedValue.Header()
+			pointedData = pointedValue.Data()
 		}
-		header := pointedValue.Header()
-		if pointedValue.Header().Address == 0 {
+		if header.Address == 0 {
```
ajwerner, line 236: Seems a little silly for `fmt.Sprintf(` when you literally know that this will result in `"0x0"`. Not to mention that that seems like a pretty meh thing to have in the output.

grantseltzer:
Dismissed previous review as I added above mentioned fixes

pkg/dyninst/gosym/symtab_test.go
```diff
@@ -96,12 +95,10 @@ func runTest(
 			require.NotEmpty(t, locations)
 			fmt.Fprintf(&out, "pc: 0x%x\n", pc)
 			for _, location := range locations {
-				// Drop toolchain-architecture-dependent infix
-				re := regexp.MustCompile("golang.org/toolchain[^/]*/")
-				location.File = re.ReplaceAllString(location.File, "golang.org/toolchain@*/")
-				// Hide path prefix that depends on repository, and GOPATH locations.
-				i := strings.Index(location.File, "/pkg/") + 1
-				fmt.Fprintf(&out, "\t%s@%s:%d\n", location.Function, location.File[i:], location.Line)
+				// Hide path prefixes that depends on toolchain,repository, and GOPATH locations.
```
ajwerner, line 98: uhhhh I don't know about eliminating 100% of paths -- let's do something to just trim the parts that are problematic.

ajwerner (approved):
Let's merge this so it can be out of your way, but please to incorporate my comments here into your other change and stay responsible.

pkg/dyninst/gosym/symtab_test.go
grantseltzer: Yea it's a little rough. I expected that it was just a matter of filtering by $GOROOT, $GOPATH, or $GOTOOLCHAIN but I don't have any of those set. It actually begs the question of why you/Piotr's env is setup different from mine. Where do you have Go stdlib src installed?

ajwerner: The answer is that your "local" go version is the requested go version so it doesn't use the one automatically downloaded from the internet. The real answer is going to be to use `trimpath` when building the binaries. After you merge this I'll come through and revert it and fix the build.

```
-trimpath
	remove all file system paths from the resulting executable.
	Instead of absolute file system paths, the recorded file names
	will begin either a module path@version (when using modules),
	or a plain import path (when using the standard library, or GOPATH).
```
https://pkg.go.dev/cmd/go

grantseltzer:
> Let's merge this so it can be out of your way, but please to incorporate my comments here into your other change and stay responsible.

I'll open another PR now.

PR #38264, opened 2025-06-26
[Orch] Check cache for tags for v4 ecs

JLineaweaver:
### What does this PR do?
- Correctly respect the `ecs_collect_resource_tags_ec2` variable when calling the ECS Agent
- Start caching tags to reduce burden on the ECS Agent
- Start logging error responses from the ECS agent.

### Motivation
We are currently making too many requests for tags to the ECS Agent. We want to reduce the burden of total calls made. We also want to respect the config `ecs_collect_resource_tags_ec2` to turn off tagging calls. In addition we want to make sure errors are visible to the user when the ECS Agent returns errors.
I deployed this agent to an ECS cluster. Then I was able to see the calls change from getTaskWithTags to getTask after tags are stored in the cache.


comp/core/workloadmeta/collectors/internal/ecs/v4parser.go
```diff
@@ -81,18 +84,54 @@ func (c *collector) getTaskWithTagsFromV4Endpoint(ctx context.Context, task v1.T
...
+				log.Warnf("failed to get task from metadata v4 API: %s", err)
+			}
+			return v1TaskToV4Task(task), err
+		}
+		// Add tags if they exist
+		if len(rt.tags) == 0 {
+			taskWithTags.TaskTags = rt.tags
+		}
+		if len(rt.containerInstanceTags) == 0 {
+			taskWithTags.ContainerInstanceTags = rt.containerInstanceTags
```
stzou, line 128: were we previously not adding container instance tags with the v4 parser?

JLineaweaver: Pretty sure both were coming from GetTaskWithTags

comp/core/workloadmeta/collectors/internal/ecs/v4parser.go
```diff
@@ -81,18 +84,54 @@ func (c *collector) getTaskWithTagsFromV4Endpoint(ctx context.Context, task v1.T
...
-	if err != nil {
-		// If it's a timeout error, log it as debug to avoid spamming the logs as the data can be fetched in next run
-		if errors.Is(err, context.DeadlineExceeded) {
-			log.Debugf("timeout while getting task with tags from metadata v4 API: %s", err)
-		} else {
-			log.Warnf("failed to get task with tags from metadata v4 API: %s", err)
+	var returnTask v3or4.Task
+	// If we have tags, are collecting tags and have zero tags present in the cache
+	// Call getTaskWithTags
+	if c.hasResourceTags && c.collectResourceTags && len(rt.tags) == 0 && len(rt.containerInstanceTags) == 0 {
```
kangyili, line 90: With `c.collectResourceTags` , users now need to turn on two feature flags for ECS collection

- ecs_collect_resource_tags_ec2 (default to false)
- ecs_task_collection_enabled

comp/core/workloadmeta/collectors/internal/ecs/v4parser.go
```diff
@@ -81,18 +84,54 @@ func (c *collector) getTaskWithTagsFromV4Endpoint(ctx context.Context, task v1.T
...
+			// If it's a timeout error, log it as debug to avoid spamming the logs as the data can be fetched in next run
+			if errors.Is(err, context.DeadlineExceeded) {
+				log.Debugf("timeout while getting task from metadata v4 API: %s", err)
+			} else {
+				log.Warnf("failed to get task from metadata v4 API: %s", err)
+			}
+			return v1TaskToV4Task(task), err
+		}
+		// Add tags if they exist
+		if len(rt.tags) == 0 {
```
kangyili: should be `if len(rt.tags) != 0 {` ?

comp/core/workloadmeta/collectors/internal/ecs/v4parser.go
```diff
@@ -81,18 +84,54 @@ func (c *collector) getTaskWithTagsFromV4Endpoint(ctx context.Context, task v1.T
...
+			taskWithTags.TaskTags = rt.tags
+		}
+		if len(rt.containerInstanceTags) == 0 {
```
kangyili: same here

comp/core/workloadmeta/collectors/internal/ecs/v4parser.go
kangyili: All tests passed, but it looks like this section isn’t covered.  We can add a new test case for it.

kangyili (approved):
Overall logic makes sense to me. Hitting the rate limit during the agent's first run seems unavoidable when the tag cache is still empty.

comp/core/workloadmeta/collectors/internal/ecs/v4parser.go
JLineaweaver: Good catch on this

comp/core/workloadmeta/collectors/internal/ecs/v4parser.go
JLineaweaver: So I believe the two flags do something different. `ecs_task_collection_enabled` allows users to collect tasks and only if that is true does `ecs_collect_resource_tags_ec2`. I think this could allow us to turn `ecs_collect_resource_tags_ec2` on by default since it shouldn't matter if `ecs_task_collection_enabled` if false. What do you think?

comp/core/workloadmeta/collectors/internal/ecs/v4parser.go
JLineaweaver: Added missing test and corrected test data

comp/core/workloadmeta/collectors/internal/ecs/v4parser.go
kangyili: if `ecs_collect_resource_tags_ec2` is true and `ecs_task_collection_enabled ` is false, then this section will be executed and we will still query v4 endpoint 

https://github.com/DataDog/datadog-agent/blob/main/comp/core/workloadmeta/collectors/internal/ecs/v1parser.go#L65-L70

JLineaweaver: I don't think we ever reach the parser code if task collection is not enabled

https://github.com/datadog/datadog-agent/blob/jlineaweaver%2Fv4-tag-cache/comp/core/workloadmeta/collectors/internal/ecs/ecs.go#L134-L139

rtrieu (approved):

sblumenthal (approved):

comp/core/workloadmeta/collectors/internal/ecs/v4parser_test.go
```diff
@@ -97,19 +109,96 @@ func TestPullWithTaskCollectionEnabledWithV4Parser(t *testing.T) {
...
+	require.Equal(t, "tag_value", rt.containerInstanceTags["tag_key"])
+
+	// Clear the task cache to force re-fetching
+	collector.taskCache.Flush()
+
+	// Reset store notifications to track new events
+	store.notifiedEvents = store.notifiedEvents[:0]
+
+	// Second pull - should fetch from API again since cache was cleared
+	fmt.Println("Josh second run")
```
gabedos: nit

comp/core/workloadmeta/collectors/internal/ecs/v4parser_test.go
```diff
@@ -97,19 +109,96 @@ func TestPullWithTaskCollectionEnabledWithV4Parser(t *testing.T) {
...
+
+// TestPullWithTaskCollectionEnabledWithV4ParserCacheClearing tests the Pull method with taskCollectionEnabled enabled
+// and taskCollectionParser set to parseTasksFromV4Endpoint to parse the tasks from the v4 metadata endpoint
+// This test runs collect twice with cache clearing in between to test tag caching behavior
+func TestPullWithTaskCollectionEnabledWithV4ParserCacheClearing(t *testing.T) {
+	_, store, collector, cleanup := setupV4ParserTest(t, true)
+	defer cleanup()
+
+	// First pull - should fetch tags from API
+	fmt.Println("Josh first run")
```
gabedos: nit

JLineaweaver: Whoops. I thought those were all cleaned up. Much appreciated.

comp/core/workloadmeta/collectors/internal/ecs/v4parser.go
```diff
@@ -81,18 +84,61 @@ func (c *collector) getTaskWithTagsFromV4Endpoint(ctx context.Context, task v1.T
...
+			return v1TaskToV4Task(task), err
+		}
+
+		// Add tags to the cache
+		rt = resourceTags{
+			tags:                  taskWithTags.TaskTags,
+			containerInstanceTags: taskWithTags.ContainerInstanceTags,
+		}
+
+		c.resourceTags[task.Arn] = rt
```
gabedos, line 109: Is the resourceTags cache ever being flushed for tasks that are no longer running? In this PR I'm only seeing resourceTags being added into the cache but never removed.

On ECS EC2 the Agent can run for long periods of time on an EC2 instance that may have a variety of different tasks that run over time and their tags would be saved in this cache even old tasks aren't running.

Should we instead use the `cache` library from `"github.com/patrickmn/go-cache"` used for the `taskCache`?

comp/core/workloadmeta/collectors/internal/ecs/v4parser_test.go
JLineaweaver: Forgot to push. Pushed up now

comp/core/workloadmeta/collectors/internal/ecs/v4parser.go
JLineaweaver: https://github.com/datadog/datadog-agent/blob/jlineaweaver%2Fv4-tag-cache/comp/core/workloadmeta/collectors/internal/ecs/ecs.go#L185-L188

I believe this is the code that cleans it up

gabedos: Sweet! looks good to me then

gabedos (approved):

PR #38270, opened 2025-06-26
Add additional e2e test cases for diskv2 configurations

jeremy-hanna:
### What does this PR do?

This PR adds additional test cases for different configuration settings for the diskv2 check to confirm behavior with the python check as we work through the migration.

### DIFF
--- a/test/new-e2e/tests/agent-runtimes/checks/disk/common.go
+++ b/test/new-e2e/tests/agent-runtimes/checks/disk/common.go
@@ -99,11 +99,11 @@ instances:
 	}
 	p := math.Pow10(v.metricCompareDecimals)
 	for _, testCase := range testCases {
-		v.Run(testCase.name, func() {
-			if testCase.onlyLinux && v.descriptor.Family() == e2eos.WindowsFamily {
-				continue
-			}
+		if testCase.onlyLinux && v.descriptor.Family() == e2eos.WindowsFamily {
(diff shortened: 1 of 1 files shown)


test/new-e2e/tests/agent-runtimes/checks/disk/common.go
```diff
@@ -53,10 +54,55 @@ instances:
...
+  - use_mount: false
+    use_lsblk: true
+`,
+			``,
+			true,
 		},
 	}
 	p := math.Pow10(v.metricCompareDecimals)
 	for _, testCase := range testCases {
+		if testCase.onlyLinux && v.descriptor.Family() == e2eos.WindowsFamily {
```
pgimalac: **🥜 nitpick:** 
```suggestion
		if testCase.onlyLinux && v.descriptor.Family() != e2eos.LinuxFamily {
```

pgimalac (approved):
LGTM, thanks for adding those !

test/new-e2e/tests/agent-runtimes/checks/disk/common.go
jeremy-hanna: fixed in https://github.com/DataDog/datadog-agent/pull/38270/commits/a4a9a17271d15e8bf74e0d8f1fd775dbf7aee0b0

PR #38275, opened 2025-06-27
Increase 'metricCompareFraction' for Windows e2e tests

jose-manuel-almaza:
### What does this PR do?
Increase 'metricCompareFraction' for Windows e2e tests

### Motivation
Windows e2e test of the disk check has started being flaky

### DIFF
--- a/test/new-e2e/tests/agent-runtimes/checks/disk/disk_win_test.go
+++ b/test/new-e2e/tests/agent-runtimes/checks/disk/disk_win_test.go
@@ -23,7 +23,7 @@ func TestWindowsDiskSuite(t *testing.T) {
 	suite := &windowsDiskCheckSuite{
 		diskCheckSuite{
 			descriptor:            e2eos.WindowsDefault,
-			metricCompareFraction: 0.02,
+			metricCompareFraction: 0.05,
 			metricCompareDecimals: 1,
 		},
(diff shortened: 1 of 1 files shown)


test/new-e2e/tests/agent-runtimes/checks/disk/disk_win_test.go
```diff
@@ -23,7 +23,7 @@ func TestWindowsDiskSuite(t *testing.T) {
 	suite := &windowsDiskCheckSuite{
 		diskCheckSuite{
 			descriptor:            e2eos.WindowsDefault,
-			metricCompareFraction: 0.02,
+			metricCompareFraction: 0.05,
```
pgimalac, line 26: Not blocking, but do we want to increase for all metrics, or could we just increase for the one flaky metric ?

pgimalac (approved):

test/new-e2e/tests/agent-runtimes/checks/disk/disk_win_test.go
jose-manuel-almaza: I think this will be enough for now. The solution you propose is very good, but I think it would add complexity to the tests, which I'm not sure we need at this point. In any case, a very good observation.

PR #38285, opened 2025-06-27
Add CRL monitoring for Windows Certificate Store integration

mrafi97:
### What does this PR do?
This PR adds Certificate Revocation List (CRL) expiration monitoring for the Windows Certificate Store. It checks the NextUpdate field of the CRLs of a specified store.

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1565

Add the following configuration in `%PROGRAMDATA%\conf.d\windows_certificate.d\conf.yaml` and restart the Agent. It should return a metric, `windows_certificate.crl_days_remaining` and service check. `windows_certificate.crl_expiration`:
```
init_config:

instances:
  - certificate_store: CA
    enable_crl_monitoring: true
```

### DIFF
--- a/cmd/agent/dist/conf.d/windows_certificate.d/conf.yaml.example
+++ b/cmd/agent/dist/conf.d/windows_certificate.d/conf.yaml.example
@@ -59,6 +59,23 @@ instances:
     #
     # days_critical: 7
 
+    ## @param enable_crl_monitoring - boolean - optional - default: false
+    ## Enables monitoring of expiration of all Certificate Revocation Lists (CRLs) in `certificate_store`.
+    #
+    # enable_crl_monitoring: false
+
+    ## @param crl_days_warning - integer - optional - default: 14
+    ## Number of days before CRL expiration from which the service check
+    ## `windows_certificate.crl_expiration` begins emitting WARNING.
+    #
+    # crl_days_warning: 14
+
+    ## @param crl_days_critical - integer - optional - default: 7
+    ## Number of days before crl expiration from which the service check
+    ## `windows_certificate.crl_expiration` begins emitting CRITICAL.
+    #
+    # crl_days_critical: 7
+
     ## @param tags - list of strings - optional
     ## A list of tags to attach to every metric and service check emitted by this instance.
     ##
--- a/cmd/agent/dist/conf.d/windows_certificate.d/conf.yaml.example
+++ b/cmd/agent/dist/conf.d/windows_certificate.d/conf.yaml.example
@@ -67,6 +67,8 @@ instances:
     ## @param crl_days_warning - integer - optional - default: 0
     ## Number of days before CRL expiration from which the service check
     ## `windows_certificate.crl_expiration` begins emitting WARNING.
+    ## If the CRLs in the store are updated automatically, set to 0 to avoid getting WARNING alerts.
+    ## If the CRLs are updated manually, set to the number days beforehand that you would like to recieve a WARNING.
     #
     # crl_days_warning: 0
 
--- a/cmd/agent/dist/conf.d/windows_certificate.d/conf.yaml.example
+++ b/cmd/agent/dist/conf.d/windows_certificate.d/conf.yaml.example
@@ -64,17 +64,11 @@ instances:
     #
     # enable_crl_monitoring: false
 
-    ## @param crl_days_warning - integer - optional - default: 14
+    ## @param crl_days_warning - integer - optional - default: 0
     ## Number of days before CRL expiration from which the service check
     ## `windows_certificate.crl_expiration` begins emitting WARNING.
     #
-    # crl_days_warning: 14
-
-    ## @param crl_days_critical - integer - optional - default: 7
-    ## Number of days before crl expiration from which the service check
-    ## `windows_certificate.crl_expiration` begins emitting CRITICAL.
-    #
-    # crl_days_critical: 7
+    # crl_days_warning: 0
 
     ## @param tags - list of strings - optional
     ## A list of tags to attach to every metric and service check emitted by this instance.
--- a/pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
+++ b/pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
@@ -529,6 +529,7 @@ func getCrlInfo(storeHandle windows.Handle) ([]crlInfoCopy, error) {
 	var crlInfo []crlInfoCopy
 	defer func() {
 		err = winutil.CertFreeCRLContext(crlContext)
+		log.Debugf("Freeing CRL context")
 		if err != nil {
 			log.Errorf("Error freeing CRL context: %v", err)
 		}
@@ -618,6 +619,7 @@ func getCrlIssuerTags(issuer string) []string {
 			issuerTags = append(issuerTags, component)
 		}
 	}
+	log.Debugf("CRL issuer tags: %v", issuerTags)
 
 	return issuerTags
 }
--- a/pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
+++ b/pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
@@ -41,12 +41,14 @@ const (
 	defaultMinCollectionInterval = 300
 	defaultDaysCritical          = 7
 	defaultDaysWarning           = 14
+	defaultCrlDaysWarning        = 0
 
 	certStoreOpenFlags = windows.CERT_SYSTEM_STORE_LOCAL_MACHINE | windows.CERT_STORE_READONLY_FLAG | windows.CERT_STORE_OPEN_EXISTING_FLAG
 	certEncoding       = windows.X509_ASN_ENCODING | windows.PKCS_7_ASN_ENCODING
 	cryptENotFound     = windows.Errno(windows.CRYPT_E_NOT_FOUND)
 	eInvalidArg        = windows.Errno(windows.E_INVALIDARG)
 	certX500NameStr    = 3
+	certNameStrCRLF    = 0x08000000
 )
 
 // Config is the configuration options for this check
@@ -60,7 +62,6 @@ type Config struct {
 	DaysCritical        int      `yaml:"days_critical" json:"days_critical" minimum:"0"`
 	DaysWarning         int      `yaml:"days_warning" json:"days_warning" minimum:"0"`
 	EnableCRLMonitoring bool     `yaml:"enable_crl_monitoring" json:"enable_crl_monitoring" default:"false"`
-	CrlDaysCritical     int      `yaml:"crl_days_critical" json:"crl_days_critical" minimum:"0"`
 	CrlDaysWarning      int      `yaml:"crl_days_warning" json:"crl_days_warning" minimum:"0"`
 }
 
@@ -134,10 +135,9 @@ func (w *WinCertChk) Configure(senderManager sender.SenderManager, integrationCo
 	}
 
 	config := Config{
-		DaysCritical:    defaultDaysCritical,
-		DaysWarning:     defaultDaysWarning,
-		CrlDaysCritical: defaultDaysCritical,
-		CrlDaysWarning:  defaultDaysWarning,
+		DaysCritical:   defaultDaysCritical,
+		DaysWarning:    defaultDaysWarning,
+		CrlDaysWarning: defaultCrlDaysWarning,
 	}
 
 	if err := yaml.Unmarshal(data, &config); err != nil {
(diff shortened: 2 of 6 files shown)


michaelcretzman (approved):

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
```diff
@@ -437,6 +523,45 @@ func parseCertificate(encodedCert []byte) (*x509.Certificate, error) {
 	return cert, nil
 }
 
+func getCrlInfo(storeHandle windows.Handle) ([]crlInfoCopy, error) {
+	var err error
+	var crlContext *winutil.CRLContext
+	var crlInfo []crlInfoCopy
+	defer func() {
+		err = winutil.CertFreeCRLContext(crlContext)
```
iglendd: Will `winutil.CertFreeCRLContext(crlContext)` work (meaning not crash, if the context is `nil`?

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
```diff
@@ -437,6 +523,45 @@ func parseCertificate(encodedCert []byte) (*x509.Certificate, error) {
...
+		log.Debugf("Freeing CRL context")
```
iglendd: Not being pedantic but `log.Debugf("Freeing CRL context")`  probably should have word `Freed` or move before actual freeing, which I think is better especially if a crash is possible.

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
```diff
@@ -437,6 +523,45 @@ func parseCertificate(encodedCert []byte) (*x509.Certificate, error) {
...
+
+		if crlContext.PCrlInfo == nil {
+			log.Errorf("CRL info pointer is nil")
+			continue
+		}
+
+		pCrlInfo := (*winutil.CRLInfo)(unsafe.Pointer(crlContext.PCrlInfo))
+		crl := crlInfoCopy{
+			Issuer:     convertCertNameBlobToString(&pCrlInfo.Issuer),
+			NextUpdate: time.Unix(0, pCrlInfo.NextUpdate.Nanoseconds()),
```
iglendd, line 555: You probably need both `ThisUpdate` and `NexUpdate` (see the comments above)?

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
```diff
@@ -464,6 +589,93 @@ func getSubjectTags(cert *x509.Certificate) []string {
...
 }
 
+func getCrlIssuerTags(issuer string) []string {
+	issuerTags := []string{}
+
+	if issuer == "" {
+		return issuerTags
+	}
+
+	issuerAttributes := splitByComma(issuer)
```
iglendd: Probably RDNs is better terminology than attributes.

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
```diff
@@ -464,6 +589,93 @@ func getSubjectTags(cert *x509.Certificate) []string {
...
+	}
+	log.Debugf("CRL issuer tags: %v", issuerTags)
+
+	return issuerTags
+}
+
+// splitByComma splits a string by commas, handling quoted parts correctly
+// For example L=Internet, O="VeriSign, Inc.", OU=VeriSign Commercial Software Publishers CA
+// will be split into [L=Internet, O=VeriSign, Inc., OU=VeriSign Commercial Software Publishers CA]
+func splitByComma(s string) []string {
```
iglendd: This looks potentially complicated to handle formatting or split. What if you use CERT_NAME_STR_CRLF_FLAG flag to simplify the split?

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
```diff
@@ -464,6 +589,93 @@ func getSubjectTags(cert *x509.Certificate) []string {
...
+			// Remove quotes if present
+			if strings.HasPrefix(value, "\"") && strings.HasSuffix(value, "\"") {
+				value = strings.Trim(value, "\"")
+			}
+
+			// Format as requested: "crl_issuer_<key>:<value>"
+			component := fmt.Sprintf("crl_issuer_%s:%s", key, value)
+			issuerTags = append(issuerTags, component)
+		}
+	}
```
iglendd: There are lot of thing which are escaped 
Comma (,)
Plus sign (+)
Equal sign (=)
Inch mark (")
Backslash followed by the letter n (\n)
Less than sign (<)
Greater than sign (>)
Number sign (#)
Semicolon (;)
I wonder if more sophisticated/general procedure is required by replacing all of them with underscore. The tags code would do this anyway probably but it is good to do it explicitly I think

cmd/agent/dist/conf.d/windows_certificate.d/conf.yaml.example
```diff
@@ -59,6 +59,23 @@ instances:
...
+    ## Number of days before CRL expiration from which the service check
+    ## `windows_certificate.crl_expiration` begins emitting WARNING.
+    #
+    # crl_days_warning: 14
+
+    ## @param crl_days_critical - integer - optional - default: 7
+    ## Number of days before crl expiration from which the service check
+    ## `windows_certificate.crl_expiration` begins emitting CRITICAL.
+    #
+    # crl_days_critical: 7
```
iglendd: I actually not sure how it can be calculated. There are `ThisUpdate` and `NextUpdate`. I am not sure that it makes sense in this approach. Consider this, This could makes sense if the CRL updates only manually that a human knows when to update it but if it setup to update CRL automatically it should happen on `NextUpdate` and setup for CRL warning/critical would give false alert to totally functional system. On the other hand it would be good to immedeatly give a warning if after a delta (1 hour?) CRL had not been updated. The problem that I do not know if CRL are setup to be updated automatically most of the time. It is hard to imagine otherwise that a human needs to be involved every time. Unless if it is needs to be done infrequently, which makes little sense for this kind of data. Default frequency for automatic update is 7 days (according to Cisco document - https://www.cisco.com/c/en/us/support/docs/security/identity-services-engine/216528-configure-microsoft-ca-server-to-publish.html).

I feel that more thoughts on the subject is needed.

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
mrafi97: @iglendd That does simplify things, I didn't think of that. I'll try it out thanks!

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
mrafi97: @iglendd when I explicitly passed a nil to `winutil.CertFreeCRLContext(nil)` it ran with no errors returned. However to be safe I can check for nil before freeing the context.

cmd/agent/dist/conf.d/windows_certificate.d/conf.yaml.example
mrafi97: As discussed yesterday, I'll only keep `crl_days_warning` and keep it at a default value of 0. This way users have the capability of adding a warning if they want but it will not be enabled by default.

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
mrafi97: From yesterday's conversation, `ThisUpdate` won't be needed since `NextUpdate` is all we'll need to determine the CRLs expiry

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
mrafi97: @iglendd  I can add this but the tags code has been doing well on its own when [checking the metrics in staging](https://dd.datad0g.com/metric/explorer?fromUser=false&fullscreen_end_ts=1751379478618&fullscreen_paused=true&fullscreen_refresh_mode=paused&fullscreen_start_ts=1751375878618&graph_layout=multi&state=H4sIAAAAAAAACl2RXYrDMAyE76LHxZSm3bTUV1mKMbGcCvyTWk4hDbn7oqTtwr5Jw-fRSJ7hgYUpJ9Bw2B--TdOYfQMK-mKHG4P-mcGhp0R1hWaoVAOCBlBbaZie0jenj2ID9WIY0FcRp0GAShEZCyGDgoL3EbluAwrykBOj8blEW_-z93GrBE02ipdIk8R0tlrDeSydyBFroe79ZgIN9tFrnrhi3HXDuBsZy_y1wHJVIMPGYDffV_NnvVyX6yIYD4GqcRQxyZ0E_6hdTp761SBQpAq62SvgXKpcKheHBTQ45A5WL1nLl3WFGbha4Zpz2xzP7eW0Px1bBZjcW7u0L62gL8g3E7OTLTmQo9SDgsGOjA60t4FxUeApVCyviJ_I5kFP8_6EPATiCssvkaYcEvkBAAA&start=1751375960635&end=1751379560635&paused=false#N4Ig7glgJg5gpgFxALlAGwIYE8D2BXJVEADxQEYAaELcqyKBAC1pEbghkcLIF8qo4AMwgA7CAgg4RKUAiwAHOChASAtnADOcAE4RNIKtrgBHPJoQaUAbVBGN8qVoD6gnNtUZCKiOq279VKY6epbINiAiGOrKQdpYZAYgUJ4YThr42gDGSsgg6gi6mZaBZnHKGABuMMiQIlA4YBpO2doSwpmecAB0yVhNRh6iojDAAFQ8AAQARlgTwBp4UwBWcJkITvgULW0QHQhwaQhucDwgfKCR0bmxWABMickIqel4WTl5iIXFIDflVTWieqNZo6HZ7bqZbRoJy9fpwQZiEQjcbTWbASHQiAaBY6DZbKFOLE47TNET4zHY0obPCnHgAXSornceEwoXCTNULIwMVK8TOFFAHK5PJ0dzODJAGjkaByoHkGBlCH2ynqMGarI0u0SaFEBzkimU6R1UG1uqc9CYyhEbg8aH5SSx8kwWCc+veOpESnpPD4kqd4gAwlJhDAUCIWWgeEA)

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
```diff
@@ -464,6 +582,65 @@ func getSubjectTags(cert *x509.Certificate) []string {
...
+	log.Debugf("CRL issuer tags: %v", issuerTags)
+
+	return issuerTags
+}
+
+func convertCertNameBlobToString(nameBlob *windows.CertNameBlob) (string, error) {
+	if nameBlob == nil || nameBlob.Size == 0 {
+		return "", nil
+	}
+	dwStrType := uint32(certX500NameStr | certNameStrCRLF)
```
iglendd, line 633: Is CERT_NAME_STR_CRLF_FLAG and CERT_X500_NAME_STR not accessible in Go? In that case you can define it above as these symbols (and even add a comment with link to the MS help page).

cmd/agent/dist/conf.d/windows_certificate.d/conf.yaml.example
```diff
@@ -59,6 +59,17 @@ instances:
...
     # days_critical: 7
 
+    ## @param enable_crl_monitoring - boolean - optional - default: false
+    ## Enables monitoring of expiration of all Certificate Revocation Lists (CRLs) in `certificate_store`.
+    #
+    # enable_crl_monitoring: false
+
+    ## @param crl_days_warning - integer - optional - default: 0
+    ## Number of days before CRL expiration from which the service check
+    ## `windows_certificate.crl_expiration` begins emitting WARNING.
```
iglendd, line 69: I feel it could be good to provide guidance here stating something along the lines 
---
If specified Certificate Store CRL is updated automatically and on the day of the CRL's NextUpdate then there is less need to specify explicit days to get warning before expirations. If the CRL needs to be updated manually or a few days before CRL;s NextUpdate then providing explicit `windows_certificate.crl_expiration`  could be useful to get reminding to update CRL manually or fix automatic CRL update problem

iglendd (commented):
Looks good, left a few minor comments

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
mrafi97: @iglendd one of the issues is that the linter does not like it if we set variables in all caps with underscores:
```
Linter failures:
pkg\collector\corechecks\system\windowscertificate\windows_certificate.go:54:2: var-naming: don't use ALL_CAPS in Go names; use CamelCase (revive)
        CERT_X500_NAME_STR = 3
        ^
pkg\collector\corechecks\system\windowscertificate\windows_certificate.go:57:2: var-naming: don't use ALL_CAPS in Go names; use CamelCase (revive)
        CERT_NAME_STR_CRLF_FLAG = 0x08000000
        ^
```

cmd/agent/dist/conf.d/windows_certificate.d/conf.yaml.example
mrafi97: @iglendd thinking of adding the following:

```
    ## @param crl_days_warning - integer - optional - default: 0
    ## Number of days before CRL expiration from which the service check
    ## `windows_certificate.crl_expiration` begins emitting WARNING.
    ## If the CRLs in the store are updated automatically, set to 0 to avoid getting WARNING alerts.
    ## If the CRLs are updated manually, set to the number days beforehand that you would like to recieve a WARNING.
    #
    # crl_days_warning: 0
```

iglendd: lgtm

pkg/collector/corechecks/system/windowscertificate/windows_certificate.go
iglendd: wow, you can add then as a comment above otherwise it is very hard to make a connection to the documentation

mrafi97: Added the following:

```
	// https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-certnametostrw
	//
	// CERT_X500_NAME_STR converts OIDs to their X.500 key names
	certX500NameStr = 3

	// CERT_NAME_STR_CRLF_FLAG replaces commas with a \r\n separator
	certNameStrCRLF = 0x08000000
```

iglendd (approved):

PR #38286, opened 2025-06-27
[CXP-2378][agent] add linux specific process checks

gengnamstyle:
### What does this PR do?
- adds duplicate process check files for linux specifically in preparation for process check refactor
- 1 line default formatter change

### Motivation
- adding the same process check files for a linux build specifically will help with future reviewers as they can see a diff instead of just new code

- built agent locally and everything was successful
- used diff command to verify there are no diffs except build flags

`process.go` and `process_linux.go`
```
❯ diff process.go process_linux.go
6c6
< //go:build !linux
---
> //go:build linux
```

`process_test.go` and `process_linux_test.go`
```
❯ diff process_test.go process_linux_test.go
6c6
< //go:build !linux
---
> //go:build linux
```

### DIFF
--- a/pkg/process/checks/process_linux.go
+++ b/pkg/process/checks/process_linux.go
@@ -3,7 +3,7 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2016-present Datadog, Inc.
 
-//go:build !linux
+//go:build linux
 
 package checks
 
--- a/pkg/process/checks/process_linux_test.go
+++ b/pkg/process/checks/process_linux_test.go
@@ -3,6 +3,8 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2016-present Datadog, Inc.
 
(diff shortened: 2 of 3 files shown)


Stephanie0829 (commented):
Just curious, (I might be lacking some context here), but here it seems like we intend to keep the functionality largely the same between the checks, so would moving some of reusable components into a common file be better (for instance, the IsEnabled() function seems pretty generic)? 

Still keeping the separate files, but only with unique code, so you are still able to get diffs.

daniel-taf (commented):
I think we shouldn't merge this kind of change in. It essentially is just adding a build tag + duplicating code to mantain functionality. This approach could get messy pretty easily imo since it's harder to test and modify in isolation. It seems like a better approach to use a new check type for this situation based on a placeholder config, like we discussed before. WDYT?

gengnamstyle:
Got it. I was planning to add the placeholder config in the process linux file, but sure I can make a new check type with a config.

PR #38287, opened 2025-06-27
feat(update.buildimages): Exclude windows by default

chouetz:
### What does this PR do?
- Add a new option to exclude the update of the windows image by default. As we generate the windows image only when it's modified, in the default case we don't need to update its value.
- Add a warning when the image is not generated so that we are aware

### Motivation
Since we merged https://github.com/DataDog/datadog-agent-buildimages/pull/867 and https://github.com/DataDog/datadog-agent/pull/37976 the update of the windows image is done less often, so we have to adapt this script.
A follow-up PR will be done on `buildimages` to generate the correct command in the slack message

Updated unit tests

### Additional Notes

I did not use the test on the dockerhub API because:
- we don't publish in dockerhub for a `test_only` image
- we don't publish all images (eg deb_armhf) This is maybe a problem, and this will be solved when we will decommission these images 😄

### DIFF
--- a/tasks/buildimages.py
+++ b/tasks/buildimages.py
@@ -4,7 +4,7 @@ import yaml
 from invoke import Context, Exit, task
 
 from tasks.libs.ciproviders.gitlab_api import ReferenceTag, update_gitlab_config, update_test_infra_def
-from tasks.libs.common.color import color_message
+from tasks.libs.common.color import Color, color_message
 
 
 @task(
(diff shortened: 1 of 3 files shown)


tasks/buildimages.py
```diff
@@ -15,19 +15,26 @@
...
     }
 )
-def update(_: Context, tag: str = "", images: str = "", test: bool = True, list_images: bool = False):
+def update(
+    _: Context, tag: str = "", images: str = "", test: bool = True, list_images: bool = False, windows: bool = False
+):
     """
     Update local files to use a new version of dedicated images from agent-buildimages
     Use --no-test to commit without the _test_only suffixes
     Use --list-images to list all available images
```
KevinFairise2, line 24:
```suggestion
    Use --list-images to list all available images
    Use --windows to update windows images
```

KevinFairise2 (approved):

PR #38289, opened 2025-06-27
[DEBUG-4110] dyninst/uploader: add code to upload batches of data

ajwerner:
There could be more testing of the actual calls, and certainly there could be more sophistication, but it's a good start.

### DIFF
--- a/pkg/dyninst/uploader/batcher_state.go
+++ b/pkg/dyninst/uploader/batcher_state.go
@@ -1,0 +1,116 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+//go:build linux_bpf
+
+package uploader
+
+import (
+	"encoding/json"
+	"fmt"
+	"time"
+)
+
+// TODO: Implement limits on the number of batches allowed to be in flight.
+// also we'll want to have some rate limits that eventually lead to dropping
+// data if we queue too much in order to bound memory usage.
+
+// TODO: Implement some sort of retry logic for failed batches.
+
+type batcherState struct {
+	cfg         batcherConfig
+	buffer      []json.RawMessage
+	bufferBytes int
+	timerSet    bool
+	metrics     *Metrics
+	nextBatchID batchID
+	inFlight    map[batchID]batchStat
+}
+
+// batchStat holds the size information for a batch that has been sent but is
+// still awaiting an outcome.
+type batchStat struct {
+	items int
+	bytes int
+}
+
+func newBatcherState(cfg batcherConfig) *batcherState {
+	return &batcherState{
+		cfg:      cfg,
+		metrics:  &Metrics{},
+		inFlight: make(map[batchID]batchStat),
+	}
+}
+
+func (s *batcherState) handleEnqueueEvent(data json.RawMessage, now time.Time, eff effects) {
+	s.buffer = append(s.buffer, data)
+	s.bufferBytes += len(data)
+
+	shouldFlushDueToSize := len(s.buffer) >= s.cfg.maxBatchItems ||
+		s.bufferBytes >= s.cfg.maxBatchSizeBytes
+
+	if shouldFlushDueToSize {
+		s.flush(eff)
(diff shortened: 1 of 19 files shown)


ajwerner:
There's a variety of ways this could be better:

* Bounded concurrency sending batches -- some degree of queueing
* The storing of []json.RawMessage is perhaps a little odd -- we could probably encode straight into some sort of batch buffer, but that's an optimization we can leave for later
* Generally bounding the memory usage for the in-flight data

Also the testing could be more complete. I think it's a good enough start to merge.

pkg/dyninst/uploader/batcher_state.go
```diff
@@ -0,0 +1,162 @@
...
+	log.Debugf("uploader received enqueue event of %d", len(data))
+	isFirstItem := len(s.batch) == 0
+	s.batch = append(s.batch, data)
+	s.currentSize += len(data)
+
+	if isFirstItem && s.cfg.maxBufferDuration > 0 {
+		s.nextMaxBufferFlush = now.Add(s.cfg.maxBufferDuration)
+	}
+
+	if s.cfg.idleFlushDuration > 0 {
```
piob-io: Discussed offline, let's simplify by having only one timer, for max buffer duration, with potential TODO for future improvement.

pkg/dyninst/uploader/batcher_state.go
```diff
@@ -0,0 +1,162 @@
...
+		s.nextIdleFlush = now.Add(s.cfg.idleFlushDuration)
+	}
+
+	// If we will flush immediately, skip arming the timer first to avoid a double
```
piob-io: This comment confused me more than helped - if we flush now, we remove the timer (special case of resetTimer with zero value), otherwise we update it. I'd remove this comment and add comment for the resetTimer below:
```
// Clear the timer
eff.resetTimer(time.Time{})
```
Or maybe even have two different methods clearTimer and (re)setTimer.

pkg/dyninst/uploader/batcher_state.go
```diff
@@ -0,0 +1,162 @@
...
+		eff.resetTimer(earliest(s.nextIdleFlush, s.nextMaxBufferFlush))
+	}
+}
+
+func (s *batcherState) handleTimerFiredEvent(now time.Time, eff effects) {
+	log.Debugf("uploader received timer fired event")
+	if len(s.batch) == 0 {
+		return
+	}
+	shouldFlushDueToTime := !s.nextIdleFlush.After(now) ||
```
piob-io: I think this is not required, given that resetTimer consumers the channel if necessary. We only need to resetTimer always when we clear the deadline. Which is not true right now, but we could make it so by having s.clearDeadlines accept eff as an argument, and have it call eff.resetTimer(0). Also would address my comment above. I think this would make this code cleaner.

piob-io (approved):
Neat stuff, thanks!

ajwerner (commented):
I simplified the state logic and added more and better testing. PTAL

pkg/dyninst/uploader/batcher_state.go
ajwerner: Done.

pkg/dyninst/uploader/batcher_state.go
ajwerner: Done.

pkg/dyninst/uploader/batcher_state.go
ajwerner: No longer present

pkg/dyninst/uploader/batcher_state.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	return nil
+}
+
+func (s *batcherState) handleStopEvent(eff effects) {
+	log.Debugf("uploader received stop event")
+	s.clearDeadlines(eff)
+}
+
+// flush clears the current batch and returns its contents.
```
piob-io: stale comment

piob-io (approved):
Thanks!

pkg/dyninst/uploader/batcher_state.go
ajwerner: Done.

ajwerner (commented):
Thanks for the review!

PR #38295, opened 2025-06-27
Go DI: Followup decoding fixes

grantseltzer:
### What does this PR do?

- Removes an unneeded allocation
- Removes unneeded nil addr checks

### Motivation

Optimizations

Ran and updates integration tests


pkg/dyninst/decode/decoder.go
```diff
@@ -206,35 +206,17 @@ func (d *Decoder) encodeValue(
...
 			irType: uint32(v.Pointee.GetID()),
 			addr:   addr,
 		}
 
-		var (
-			header         *output.DataItemHeader
-			pointedData    []byte
-			dontParseValue bool
-		)
+		var address uint64
```
ajwerner, line 214: We should handle the 0 address case explicitly. It means that the pointer was null. I think that it should not have an address field and it should have `"isNull": true`, it should have the type name and then that's it.

pkg/dyninst/decode/decoder.go
```diff
@@ -206,35 +206,17 @@ func (d *Decoder) encodeValue(
...
 		pointedValue, ok := dataItems[key]
```
ajwerner, line 215: given how far away from here you use this boolean, I'd give it a better name like `found` or `dataFound`?

pkg/dyninst/decode/decoder.go
```diff
@@ -244,12 +226,12 @@ func (d *Decoder) encodeValue(
...
 			return err
 		}
-		addrString := fmt.Sprintf("0x%x", header.Address)
+		addrString := fmt.Sprintf("0x%x", address)
 		err = enc.WriteToken(jsontext.String(addrString))
 		if err != nil {
 			return err
 		}
-		if _, ok := currentlyEncoding[key]; !ok && !dontParseValue {
+		if _, alreadyEncoding := currentlyEncoding[key]; !alreadyEncoding && ok {
```
ajwerner, line 234: I think we should be doing rather different things for the already encoding thing and the case where we didn't find the data item. If we didn't find it, it's probably because we hit the depth limits and then I think we should consider something like `"notCapturedReason": "depth"` -- though we'll want to at some point differentiate between depth limits and failing to dereference a pointer. In side-eye we did some tracking of failures to read memory. For now let's assume it's depth.

If we hit a cycle then we should say that in the `"notCapturedReason"`

pkg/dyninst/testdata/decoded/sample.yaml
```diff
@@ -16,10 +17,10 @@ testInt32Array: '{"captures":{"probe_id":"testInt32Array","x":[1,2]},"stack_fram
 testInt64Array: '{"captures":{"probe_id":"testInt64Array","x":[1,2]},"stack_frames":""}'
 testIntArray: '{"captures":{"probe_id":"testIntArray","x":[1,2]},"stack_frames":""}'
 testInterface: '{"captures":{"b":"unimplemented - non-empty interface","probe_id":"testInterface"},"stack_frames":""}'
-testLinkedList: '{"captures":{"a":{"b":{"Address":"","Value":{"b":{"Address":"","Value":{"b":"nil","val":3}},"val":2}},"val":1},"probe_id":"testLinkedList"},"stack_frames":""}'
+testLinkedList: '{"captures":{"a":{"b":{"Address":"","Value":{"b":{"Address":"","Value":{"b":{"Address":""},"val":3}},"val":2}},"val":1},"probe_id":"testLinkedList"},"stack_frames":""}'
```
ajwerner, line 20: let's definitely not ever output `"Address":""`.

grantseltzer: This is because the address is redacted for test data. The redaction update you suggested is in #38233

PR #38296, opened 2025-06-27
[AGENTRUN-518] move file providers to autodiscovery module function

GustavoCaso:
### What does this PR do?

Move the file providers of the Autodiscovery component to the initialization. 

It removes the uncertainty when accessing Autodiscovery at the `run` function if the integration configurations have been parsed. 

Also, it removes arguments from `common.LoadComponents`


louis-cqrl (approved):
LGTM for ARUN files

rahulkaukuntla (approved):
Looks good for AGENTCFG files

comp/core/autodiscovery/autodiscoveryimpl/autoconfig.go
```diff
@@ -205,6 +207,25 @@ func createNewAutoConfig(schedulerController *scheduler.Controller, secretResolv
...
+		"",
+	}
+
+	if conf.GetString("fleet_policies_dir") != "" {
+		confSearchPaths = append(confSearchPaths, filepath.Join(conf.GetString("fleet_policies_dir"), "conf.d"))
+	}
+
+	providers.InitConfigFilesReader(confSearchPaths)
+
+	ac.AddConfigProvider(
```
davidor, line 223: The problem I see is that some config providers are now added inside the component, while others are added outside.

I think they should all be moved inside the component, so this change is a step in the right direction and we don't need to do everything in the same PR.

Just something to keep in mind.

davidor (approved):

comp/core/autodiscovery/autodiscoveryimpl/autoconfig.go
GustavoCaso: Absolutely, completely agree. We could collaborate  to see if we can use value groups to define the config providers to follow the Fx way, or if we need to look for another approach

dkirov-dd (commented):
✅ Agent Integrations

ddrthall (approved):

PR #38297, opened 2025-06-27
[AGENTCFG-313] Bumping the secret-generic-connector side binary to 0.2.5

rahulkaukuntla:
### What does this PR do?
We needed to bump the go version of the sgc binary to 1.24, which we did in sgc release 0.2.5. As such, we need to bump the omnibus script for sgc so that the agent has the most up to date version of sgc.


rdesgroppes (approved):
LGTM

PR #38298, opened 2025-06-27
Migrate Autodiscovery Filtering Logic to CEL Engine

gabedos:
### What does this PR do?

Migrates the filtering engine for Autodiscovery from raw regex to the new Filtering component using cel-go

### Motivation

Unify Agent filtering behind a common component. Make it easier to expose new filtering configuration options that are more expressive in the future. [CONTP-845]

Deploy Agent with the following environment variable configuration and check that `container.*` metrics are missing for the `Agent` (assuming it is in the `datadog-agent` namespace)

```
datadog:
  kubelet:
    tlsVerify: false
  envDict:
    DD_CONTAINER_EXCLUDE: "kube_namespace:datadog-agent"
```

In the worst case, the CEL filters are 10x slower than regular regex. This is not a major concern because there is not a high magnitude of containers on a single node and the Agent only ever processes the container entity once through AutoDiscovery. 

### Additional Notes

Remaining Work: store the errors while creating each filter and make it easy for autodiscovery to propagate these errors. potentially in the future we could remove this field from being embedded in Autodiscovery and include the Filtering component's own status.

https://github.com/DataDog/datadog-agent/blob/ec05870e20c45a2d8d42d2f266dcd8f6a2096ba0/comp/core/autodiscovery/status/status.go#L20-L24

[CONTP-845]: https://datadoghq.atlassian.net/browse/CONTP-845?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nhulston (approved):
Serverless ✅

tobz (approved):

soberpeach (approved):
lgtm 👍

dustmop (approved):

pgimalac (approved):

comp/core/autodiscovery/listeners/kubelet.go
```diff
@@ -45,11 +50,10 @@ func NewKubeletListener(options ServiceListernerDeps) (ServiceListener, error) {
 		return nil, errors.New("workloadmeta store is not initialized")
 	}
 	var err error
-	l.workloadmetaListener, err = newWorkloadmetaListener(name, filter, l.processPod, wmetaInstance, options.Telemetry)
+	l.workloadmetaListener, err = newWorkloadmetaListener(name, wmetaFilter, l.processPod, wmetaInstance, options.Telemetry)
 	if err != nil {
 		return nil, err
 	}
-	l.tagger = options.Tagger
```
GustavoCaso, line 52: Is this intended?

GustavoCaso:
@gabedos the code looks great 🎉 

If I could suggest a different organization, from the point of view of the components.
It seems the filter component is `comp/core/filter`, we usually place either components used by other components here, or once we consider mandatory for the functionality of the agent. 

Looking at the filter component interface:
```go
type Component interface {
	IsContainerExcluded(container *Container, containerFilters [][]ContainerFilter) bool
	IsPodExcluded(pod *Pod, podFilters [][]PodFilter) bool
	IsServiceExcluded(service *Service, serviceFilters [][]ServiceFilter) bool
	IsEndpointExcluded(endpoint *Endpoint, endpointFilters [][]EndpointFilter) bool
}
```

It looks pretty container-focused. 
Upon reviewing the code, it appears that autodiscovery is the only component that requires it. If that is the case, I believe we could move the element to the autodiscovery component itself. We can even consider if we actually need a  separate component for it, or if it's just a separate package within the `comp/core/autodiscovery` folder.

If that is a valid apprach it would simplify the other commands as they would not need to include the filter component at all 😄 

Of course, all this is predicated on the idea that the filter component is only going to be used by the autodiscovery component. 

Feel free to ping if you want to chat over Slack or Zoom

comp/core/filter/def/proto/filter.proto
```diff
@@ -26,3 +26,15 @@ message FilterECSTask {
...
+  string name = 1;
+  string namespace = 2;
+  map<string, string> annotations = 3;
+}
+
+message FilterKubeEndpoint {
+  string name = 1;
+  string namespace = 2;
+  map<string, string> annotations = 3;
+}
```
GustavoCaso, line 40: Why are we defining the proto definitions here and not in the `pkg/proto` folder?

dkirov-dd (approved):

TCheruy (approved):

comp/core/filter/def/proto/filter.proto
gabedos: These proto definitions are tightly coupled to the filter component and represent internal data structures used specifically for filtering operations. They're not meant for broader inter-process communication or external APIs.

The proto defintions and structs in `pkg/proto` seemed more related to IPC whereas these proto structs are the only struct used in the filtering component because the cel-go rule engine is able to parse them efficiently for syntax errors.

If you still think it's better to main these proto definitions in `pkg/proto` I can migrate them over!

GustavoCaso: Sounds good, we can keep it here

GustavoCaso:
I synced with @gabedos over Slack regarding https://github.com/DataDog/datadog-agent/pull/38298#issuecomment-3023127885

The intention is to use the filter component on other components in the future (autodiscovery, tagger, etc.) For that, it makes sense to keep it as a component then.

We agree that the name `filter` is a bit ambiguous, and we settle on renaming it to `workloadfilter`

clamoriniere (approved):
Nice refactoring 💯

PR #38299, opened 2025-06-27
ir: rework probe definition

ajwerner:
Follow-up from #38239. This moves the concept of a ProbeDefinition into the ir layer. It also simplifies the IR probe to directly include its definition. This way, rcjson can directly implement the ProbeDefinition interface, and so can other packages for their own probes.

Part of https://datadoghq.atlassian.net/browse/DEBUG-4039


piob-io (approved):
Thanks!

PR #38309, opened 2025-06-30
[AGENTRUN-517] create a autodiscovery providers types package

GustavoCaso:
### What does this PR do?

This PR creates a `providers/types` package to define the provider types. It's mostly to keep things separate, as the providers' folders include many third-party libraries. If we want to import the provider types in the future, we would not pull all the providers' dependencies. 

### Motivation

Ensure we keep interface separate from implementations


wdhif (approved):
LGTM for @DataDog/container-platform files

mellon85 (approved):

ddrthall (approved):

iliakur (approved):

PR #38310, opened 2025-06-30
KMT: update readme with skip-ssh-config option

mbertrone:
### What does this PR do?
Update KMT README to reflect a previous change https://github.com/DataDog/datadog-agent/pull/37618

### DIFF
--- a/tasks/kernel_matrix_testing/README.md
+++ b/tasks/kernel_matrix_testing/README.md
@@ -331,7 +331,7 @@ dda inv -e kmt.init --remote-setup-only
 
 The `--images` parameter is required unless `--remote-setup-only` or `--all-images` is specified, or you explicitly confirm the interactive prompt. Use `kmt.ls` to see available images.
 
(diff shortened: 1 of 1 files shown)


tasks/kernel_matrix_testing/README.md
```diff
@@ -322,13 +322,16 @@ dda inv -e kmt.init --images=ubuntu_22.04,debian_10
...
+dda inv -e kmt.init --skip-ssh-setup
+
 # If you only intend to manage remote VMs
 dda inv -e kmt.init --remote-setup-only
 ```
 
 The `--images` parameter is required unless `--remote-setup-only` or `--all-images` is specified, or you explicitly confirm the interactive prompt. Use `kmt.ls` to see available images.
 
-This command will also guide you through the [default SSH key configuration](#kmtconfig-ssh-key).
+This command will also guide you through the [default SSH key configuration](#kmtconfig-ssh-key). You can use the `--skip-ssh-setup` flag to bypass this interactive wizard. If you skip it, you will need to provide an SSH key manually using the `--ssh-key` flag for commands that interact with remote VMs.
```
usamasaqib: I think we can specify that the preferred approach is to use the `kmt.config-ssh-key` task to setup the ssh key later. 

Alternative is to use the `--ssh-key` flag explicitly.

mbertrone: Good points. Feedback applied

usamasaqib (approved):

PR #38312, opened 2025-06-30
[EBPF] gpu: handle runtime changes of CUDA_VISIBLE_DEVICES

gjulianm:
### What does this PR do?

This PR adds support for handling runtime updates for the `CUDA_VISIBLE_DEVICES` environment variable. This consists of the following changes:

* Add an uprobe to the `setenv` function in `libc`, that watches for the calls with the `CUDA_VISIBLE_DEVICES` variable name exclusively. The value will be sent to the GPU event ringbuffer.
* Add a new library set to the shared libraries watcher so that we can watch for processes opening `libc`. 
* Add a rule to the GPU monitoring attacher, to attach to all processes that have `setenv`. It's necessary to do for all processes, as a process might change its environment variable before calling any CUDA functions. If we probed only processes that have the CUDA library loaded we would miss those events.
* The GPU consumer now detects the events corresponding to changes in this variable, and will update them in the `systemContext`.
* `systemContext` will prioritize this value now over the one from `/proc/pid/environ` when finding the devices visible to a process.
* In order to select the origin of the `CUDA_VISIBLE_DEVICES`, the code retrieving the environment variable for the process has been moved  from the `pkg/gpu/cuda` package to `systemContext`.

### Motivation

https://datadoghq.atlassian.net/browse/NWL-1275

We have detected an issue in a `ray` cluster. When deployed in a node with multiple GPUs, `ray` will spawn one process per GPU. Each process will then [update](https://github.com/ray-project/ray/blob/c5e45ed42c4d23a1da0cb8ae1ade891d37866e39/python/ray/_private/accelerators/nvidia_gpu.py#L92) its own `CUDA_VISIBLE_DEVICES` environment variable. Our current approach will not detect this change, and will consider that all processes are using the same GPU.

Unit tests provided, tested also in the [GPU load test environment](https://dddev.datadoghq.com/dashboard/ws9-twf-2ry?fromUser=false&refresh_mode=sliding&tpl_var_base_agent-env%5B0%5D=gjsetdev-base&tpl_var_compare_agent-env%5B0%5D=gjsetdev-new&tpl_var_kube_cluster_name%5B0%5D=sp-test-env-gpu-1&from_ts=1751892422424&to_ts=1751894222424&live=true) with no significant changes in resource usage.

Finally, manually tested this in an AWS instance with Ray running on multiple GPUs, replicating the environment with the original error.

### DIFF
--- a/pkg/gpu/context.go
+++ b/pkg/gpu/context.go
@@ -17,6 +17,7 @@ import (
 	"github.com/DataDog/datadog-agent/pkg/gpu/cuda"
 	ddnvml "github.com/DataDog/datadog-agent/pkg/gpu/safenvml"
 	gpuutil "github.com/DataDog/datadog-agent/pkg/util/gpu"
+	"github.com/DataDog/datadog-agent/pkg/util/kernel"
 	"github.com/DataDog/datadog-agent/pkg/util/ktime"
 )
 
@@ -34,6 +35,12 @@ type systemContext struct {
 	// Note that this is the device index as seen by the process itself, which might
 	// be modified by the CUDA_VISIBLE_DEVICES environment variable later
 	selectedDeviceByPIDAndTID map[int]map[int]int32
+
+	// updatedVisibleDevicesEnvVars maps each process ID to the latest visible
+	// devices environment variable that was set by the process. This is used to
+	// keep track of updates during process runtime, whic aren't visible in
+	// /proc/pid/environ.
+	updatedVisibleDevicesEnvVars map[int]string
 
 	// deviceCache is a cache of GPU devices on the system
 	deviceCache ddnvml.DeviceCache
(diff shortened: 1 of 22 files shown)


guyarb (approved):
LGTM on the USM owned files

git-thuerk-done (approved):
good for docs

pkg/gpu/ebpf/c/runtime/gpu.c
```diff
@@ -366,4 +366,44 @@ int BPF_URETPROBE(uretprobe__cudaMemcpy) {
...
+
+    // return value of bpf_probe_read_user_str_with_telemetry is the length of the string read,
+    // including the NULL byte. If the string is not the same length, it's not CUDA_VISIBLE_DEVICES.
+    if (res != sizeof(cuda_visible_devices)) {
+        return 0;
+    }
+
+    // bpf_strncmp is available in kernel 5.17, our minimum kernel version is 5.8.0
+    // so we need to do a manual comparison
+#pragma unroll
```
val06: i think in 5.8.0 we can use bounded loops without #pragma unroll

pkg/gpu/context.go
```diff
@@ -35,6 +36,12 @@ type systemContext struct {
 	// be modified by the CUDA_VISIBLE_DEVICES environment variable later
 	selectedDeviceByPIDAndTID map[int]map[int]int32
 
+	// updatedVisibleDevicesEnvVars maps each process ID to the latest visible
+	// devices environment variable that was set by the process. This is used to
+	// keep track of updates during process runtime, which aren't visible in
+	// /proc/pid/environ.
+	updatedVisibleDevicesEnvVars map[int]string
```
val06: rename to `cudaVisibleDevicesPerProcess`

val06 (commented):
few small comments

val06 (approved):

PR #38316, opened 2025-06-30
(fleet) avoid setting tags when install script parent is injected

arbll:
This PR addresses a weird interaction between APM's SSI and the install script.

APM SSI adds a `DD_TAGS=_dd.injection.mode` env variable to all injected processes. This is usually fine but in a very specific scenario where:
1. A process in python/ruby/etc... is injected
2. This process starts the install script

This can lead to an unexpected `DD_TAGS` being passed to the install script, overriding the previous values.


arbll:
closing since we're moving to deb/rpm first citizen

PR #38322, opened 2025-06-30
[pkg/dyninst/ebpf] Prepare ebpf code for static compilation with dynamic stack machine code

piob-io:

Make changes necessary to successfully statically compile the ebpf program, without breaking the runtime compilation. Prepare for dynamically programming the stack machine - expose types needed to program the stack machine.

Include files and compilation flags are slightly different, this requires some minor type renames and function tags.


ajwerner (approved):
LGTM

pkg/ebpf/printk_patcher_test.go
```diff
@@ -138,6 +138,11 @@ func TestPatchPrintkAllAssets(t *testing.T) {
 
 		progname := strings.TrimSuffix(filepath.Base(path), "-debug.o")
 
+		if progname == "dyninst_event" {
+			// dyninst doesn't use the shared debugging infrastructure yet.
+			return nil
+		}
+
```
gjulianm, line 145: Is this necessary? The test shouldn't fail even if a single program doesn't have `log_debug` calls.

piob-io: The test fails otherwise - afaict it's because we do use bpf_printk in our code (behind our own debug define), so the attempt is made to patch it, but it's not used the way that the code expects, and patching fails.

gjulianm: I see, I'll investigate that in case it's a bug in our patching code.

gjulianm (approved):

piob-io:
Thank you!

PR #38323, opened 2025-06-30
[pkg/dyninst] Switch from runtime compilation to static ebpf program and dynamic stack machine code

piob-io:

Instead of generating C code that programs the stack machine, all relevant data is stored in maps, and injected during loading.
Loading becomes more complicated, compilation phase goes away.

Code structure is preserved as much as possible, I will send another PR that cleans things up. Most notably: runtime compilation setup can be dropped and sm will become the only package of the compiler.


pkg/dyninst/loader/loader.go
```diff
@@ -0,0 +1,417 @@
...
+// WithRingBufSize sets the size of the ring buffer for the Actuator.
+func WithRingBufSize(size uint32) option {
+	return ringBufSizeOption(size)
+}
+
+// WithDebugLevel sets the debug level for the ebpf program.
+// It forces use of a binary compiled in a debug mode
+func WithDebugLevel(level int) option {
+	return debugLevelOption(level)
+}
```
ajwerner, line 52: I can see an argument for wanting this to be per program, but this works for now

pkg/dyninst/loader/serialize.go
```diff
@@ -0,0 +1,129 @@
...
+	MaxOpLen                uint32
+	ChasePointersEntrypoint uint32
+
+	TypeIDs   []uint64
+	TypeInfos []typeInfo
+
+	ThrottlerParams []throttlerParams
+
+	ProbeParams     []probeParams
+	BPFAttachPoints []BPFAttachPoint
```
ajwerner: why did you export these fields but not this type? For json serialization?

pkg/dyninst/loader/loader.go
```diff
@@ -0,0 +1,417 @@
...
+type Program struct {
+	Collection   *ebpf.Collection
+	BpfProgram   *ebpf.Program
+	Attachpoints []BPFAttachPoint
+}
+
+// Close releases the program resources.
+func (p *Program) Close() {
+	if p.Collection != nil {
+		p.Collection.Close() // should already contain the program
```
ajwerner, line 152: should we clear the collection and program here, effectively marking the program as poisoned?

pkg/dyninst/actuator/actuator.go
```diff
@@ -192,144 +170,86 @@ type effects Actuator
...
-			a.compiler, programID, executable, probes,
-			a.configuration.codegenWriter,
-		)
+		ir, err := generateIR(programID, executable, probes)
 		if err != nil {
-			err = fmt.Errorf("failed to compile eBPF program: %w", err)
-			a.reporter.ReportCompilationFailed(programID, err, probes)
-			a.sendEvent(eventProgramCompilationFailed{
-				programID: programID,
+			a.reporter.ReportLoadingFailed(nil, err)
```
ajwerner, line 184: On main the signature for compilation became: ` ReportCompilationFailed(ir.ProgramID, error, []ir.ProbeDefinition)` How about we carve out a separate reporting event for `ReportIRGenerationFailed` and there give it the error and the probe definitions?

pkg/dyninst/compiler/testdata/snapshot/simple.arch=amd64,toolchain=go1.24.3.sm.txt
```diff
@@ -0,0 +1,234 @@
+// Stack machine code
+	Illegal 
+// 0x1: ChasePointers
+	ChasePointers 
+	Return 
+// 0x3: ProcessExpression[Probe[main.intArg]@0x4a806a.expr[0]]
+	ExprPrepare 
+	ExprReadRegister 00 08 00 00 00 00 
+	ExprSave 01 00 00 00 08 00 00 00 00 00 00 00
```
ajwerner, line 9: It'd be cool if the integers were easier to read.

ajwerner (approved):
This is great! If you don't feel like adjusting the reporter here, I can do that where I need it.

pkg/dyninst/loader/loader.go
piob-io: Agreed for the future. Loader will need to keep non-debug and debug collection specs, but that's a minor change.

pkg/dyninst/loader/serialize.go
piob-io: Not even, tinker cruft, initially the type was exported from compiler package; unexported the fields.

pkg/dyninst/actuator/actuator.go
piob-io: Sure, I will do that in a follow-up pr.

pkg/dyninst/compiler/testdata/snapshot/simple.arch=amd64,toolchain=go1.24.3.sm.txt
piob-io: I thought about this, and we could preserve the original ops, so we could just print the actual op parameters. Not a big change, but not trivial either. I planned to do it in some spare time, on lower priority.
No idea how to make integers more readable short-term, it's harder given that most of these are zeros...

pkg/dyninst/loader/loader.go
piob-io: What's the benefit of that? I presume that whatever calls Close is about to drop the pointer to the program too.

piob-io: (discussed offline)

PR #38324, opened 2025-06-30
CXP-2391 [agent] feat: add new wlm query for processes

gengnamstyle:
### What does this PR do?
- adds new query function for WLM to allow to fetch the container associated with a process in the optimal manner (following existing [GetKubernetesPodForContainer](https://github.com/DataDog/datadog-agent/blob/c2c5dfd63cd535d2d1af18686df4c10479d37faf/comp/core/workloadmeta/impl/store.go#L292) structure)
- adds comments for clarification on data model as well

### Motivation
- we previously queried for all containers and filtered out the ones that are related to a process. This was inefficient, and by changing the data model we store a reference to a container on the process entity itself (stored in WLM). 
- part of the refactoring of the process check includes querying WLM for our process data

- unit tests

### DIFF
--- a/comp/core/workloadmeta/def/component.go
+++ b/comp/core/workloadmeta/def/component.go
@@ -96,6 +96,9 @@ type Component interface {
 	// to all entities with kind KindProcess.
 	ListProcesses() []*Process
 
+	// GetContainerForProcess returns the container associated with a process if it exists.
+	// It fetches the entity with kind KindProcess and the given pid and then the entity
+	// with kind KindContainer with the cid from the process entity.
 	GetContainerForProcess(processID string) (*Container, error)
 
 	// GetGPU returns metadata about a GPU device. It fetches the entity
--- a/comp/core/workloadmeta/def/component.go
+++ b/comp/core/workloadmeta/def/component.go
@@ -96,6 +96,8 @@ type Component interface {
 	// to all entities with kind KindProcess.
 	ListProcesses() []*Process
 
+	GetContainerForProcess(processID string) (*Container, error)
+
 	// GetGPU returns metadata about a GPU device. It fetches the entity
 	// with kind KindGPU and the given ID.
 	GetGPU(id string) (*GPU, error)
--- a/comp/core/workloadmeta/def/types.go
+++ b/comp/core/workloadmeta/def/types.go
@@ -1241,7 +1241,7 @@ type Process struct {
 	Name         string   // Name -- /proc/[pid]/status
 	Cwd          string   // Current Working Directory -- /proc/[pid]/cwd
 	Exe          string   // Exceutable Path -- /proc[pid]/exe
-	Comm         string   // Short Command Name -- /proc/[]id]/comm
+	Comm         string   // Short Command Name -- /proc/[pid]/comm
 	Cmdline      []string // Command Line -- /proc/[pid]/cmdline
 	Uids         []int32  // User IDs -- /proc/[pid]/status
 	Gids         []int32  // Group IDs -- /proc/[pid]/status
--- a/comp/core/workloadmeta/def/types.go
+++ b/comp/core/workloadmeta/def/types.go
@@ -1235,18 +1235,18 @@ type Service struct {
 type Process struct {
 	EntityID // EntityID.ID is the PID
 
-	Pid          int32
-	NsPid        int32
-	Ppid         int32
-	Name         string
-	Cwd          string
-	Exe          string
-	Comm         string
-	Cmdline      []string
-	Uids         []int32
-	Gids         []int32
+	Pid          int32    // Process ID -- /proc/[pid]
+	NsPid        int32    // Namespace PID -- /proc/[pid]/status
+	Ppid         int32    // Parent Process ID -- /proc/[pid]/stat
+	Name         string   // Name -- /proc/[pid]/status
+	Cwd          string   // Current Working Directory -- /proc/[pid]/cwd
+	Exe          string   // Exceutable Path -- /proc[pid]/exe
+	Comm         string   // Short Command Name -- /proc/[]id]/comm
+	Cmdline      []string // Command Line -- /proc/[pid]/cmdline
+	Uids         []int32  // User IDs -- /proc/[pid]/status
+	Gids         []int32  // Group IDs -- /proc/[pid]/status
 	ContainerID  string
-	CreationTime time.Time
+	CreationTime time.Time // Process Start Time -- /proc/[pid]/stat
 	Language     *languagemodels.Language
 
 	// Owner will temporarily duplicate the ContainerID field until the new collector is enabled so we can then remove the ContainerID field
--- a/comp/core/workloadmeta/impl/store.go
+++ b/comp/core/workloadmeta/impl/store.go
@@ -288,6 +288,39 @@ func (w *workloadmeta) ListProcessesWithFilter(filter wmdef.EntityFilterFunc[*wm
 	return processes
 }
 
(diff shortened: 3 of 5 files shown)


comp/core/workloadmeta/def/component.go
```diff
@@ -96,6 +96,8 @@ type Component interface {
 	// to all entities with kind KindProcess.
 	ListProcesses() []*Process
 
+	GetContainerForProcess(processID string) (*Container, error)
+
```
daniel-taf, line 103: missing comment here?

comp/core/workloadmeta/def/types.go
```diff
@@ -1235,18 +1235,18 @@ type Service struct {
...
-	Cmdline      []string
-	Uids         []int32
-	Gids         []int32
+	Pid          int32    // Process ID -- /proc/[pid]
+	NsPid        int32    // Namespace PID -- /proc/[pid]/status
+	Ppid         int32    // Parent Process ID -- /proc/[pid]/stat
+	Name         string   // Name -- /proc/[pid]/status
+	Cwd          string   // Current Working Directory -- /proc/[pid]/cwd
+	Exe          string   // Exceutable Path -- /proc[pid]/exe
+	Comm         string   // Short Command Name -- /proc/[]id]/comm
```
daniel-taf:
```suggestion
	Comm         string   // Short Command Name -- /proc/[pid]/comm
```

daniel-taf (approved):

comp/core/workloadmeta/impl/store_test.go
```diff
@@ -781,6 +781,143 @@ func TestGetProcess(t *testing.T) {
...
+				assert.EqualError(t, err, tc.expectedError.Error())
+			} else {
+				assert.NoError(t, err)
+			}
+
+			if tc.expectedContainer == nil {
+				assert.Nil(t, container)
+			} else {
+				assert.Equal(t, *tc.expectedContainer, *container)
+			}
```
wiyu: Why bother checking the container if it's nil? 

I think you can simplify this to the following and also put in the `require.NoError` when you don't expect it to avoid a panic in case an error does return.

```suggestion
			if tc.expectedError != nil {
				assert.EqualError(t, err, tc.expectedError.Error())
			} else {
				require.NoError(t, err)
				assert.Equal(t, *tc.expectedContainer, *container)
			}
```

comp/core/workloadmeta/impl/store_test.go
```diff
@@ -781,6 +781,143 @@ func TestGetProcess(t *testing.T) {
...
+			// store container data into wlm
+			for _, container := range tc.containerData {
+				s.handleEvents([]wmdef.CollectorEvent{
+					{
+						Type:   wmdef.EventTypeSet,
+						Source: fooSource,
+						Entity: container,
+					},
+				})
+			}
```
Stephanie0829, line 904: Super small nit - can add a newline here to separate storing the container data into wlm, and where you start testing the function `GetContainerForProcess`.

comp/core/workloadmeta/impl/store_test.go
gengnamstyle: I had the previous structure because if the expectedContainer is nil, I can't dereference the actual nil container either, so I think I should leave it as is. Thoughts?

wiyu: Sorry, not quite sure what you mean by this. If there are no errors then `expectedContainer` shouldn't be nil right? `nil` is only returned if there is an error.

gengnamstyle: Ah yes I gotcha. Yeah I can simplify to that. Thanks

comp/core/workloadmeta/impl/store.go
```diff
@@ -288,6 +288,39 @@ func (w *workloadmeta) ListProcessesWithFilter(filter wmdef.EntityFilterFunc[*wm
...
+
+	processEntities, ok := w.store[wmdef.KindProcess]
+	if !ok {
+		return nil, errors.NewNotFound(string(wmdef.KindProcess))
+	}
+
+	processEntity, ok := processEntities[processID]
+	if !ok {
+		return nil, errors.NewNotFound(processID)
+	}
```
adel121, line 304: Why not use `GetProcess(pid int32)` directly here?

comp/core/workloadmeta/impl/store.go
```diff
@@ -288,6 +288,39 @@ func (w *workloadmeta) ListProcessesWithFilter(filter wmdef.EntityFilterFunc[*wm
 	return processes
 }
 
+// GetContainerForProcess implements Store#GetContainerForProcess
+func (w *workloadmeta) GetContainerForProcess(processID string) (*wmdef.Container, error) {
```
adel121, line 292: Why is `processID` a string?

comp/core/workloadmeta/impl/store.go
```diff
@@ -288,6 +288,39 @@ func (w *workloadmeta) ListProcessesWithFilter(filter wmdef.EntityFilterFunc[*wm
...
+
+	containerEntities, ok := w.store[wmdef.KindContainer]
+	if !ok {
+		return nil, errors.NewNotFound(process.Owner.ID)
+	}
+
+	container, ok := containerEntities[process.Owner.ID]
+	if !ok {
+		return nil, errors.NewNotFound(process.Owner.ID)
+	}
```
adel121, line 319: Same question, why not use `GetContainer` ?

adel121 (commented):
Left some questions

comp/core/workloadmeta/impl/store.go
gengnamstyle: Yeah I was initially considering that, but after looking at an existing similar function [`GetKubernetesPodForContainer`](https://github.com/DataDog/datadog-agent/blob/5ff505569060cb8fc36e577c902d71e093e7d984/comp/core/workloadmeta/impl/store.go#L325-L325) function and it appears this was done to mitigate an extra lock + unlock of the read mutex.

If i were to use `GetProcess` and `GetContainer` I would introduce 2 locks instead of the one here. However, I'm not as familiar with the performance implication here, so if you think re-using the functions is fine + ideal, I don't mind refactoring!

comp/core/workloadmeta/impl/store.go
gengnamstyle: Consistency with the existing [`GetKubernetesPodForContainer`](https://github.com/DataDog/datadog-agent/blob/5ff505569060cb8fc36e577c902d71e093e7d984/comp/core/workloadmeta/impl/store.go#L325-L325) as it also has a string input

comp/core/workloadmeta/impl/store.go
gengnamstyle: Copied answer like above for ease: 

> Yeah I was initially considering that, but after looking at an existing similar function [GetKubernetesPodForContainer](https://github.com/DataDog/datadog-agent/blob/5ff505569060cb8fc36e577c902d71e093e7d984/comp/core/workloadmeta/impl/store.go#L325-L325) function and it appears this was done to mitigate an extra lock + unlock of the read mutex.
> 
> If i were to use GetProcess and GetContainer I would introduce 2 locks instead of the one here. However, I'm not as familiar with the performance implication here, so if you think re-using the functions is fine + ideal, I don't mind refactoring!

comp/core/workloadmeta/impl/store.go
adel121: Ok this makes sense. 

I think it is fine as is then.

adel121 (approved):

PR #38326, opened 2025-06-30
[AGTMETRICS-306] Add `sequenceID` to config models

rayz:
### What does this PR do?
This pr introduces the `sequenceID` field to the `viper` model and `nodetreemodel` configs. 

This is used to keep track of a `sequenceID`, a `uint64` which is incremented on every `Set` / `Unset` of a configuration setting.

The `NotificationReceivers` have been updated to take in this `sequenceID`

### Motivation

The notion of a `sequenceID` is needed so that we can version configuration changes. This will be used in a future pr for the  implementation of a config stream between the core agent and remote agents.

New and existing unit tests.


louis-cqrl (approved):

pgimalac (approved):
LGTM for Agent Runtimes files

tobz (changes requested):
I think we should be using a single sequence ID for the entire configuration.

One of the original goals was that we could detect when we had received updates for distinct settings out of order (e.g., `[A, B, C]` is observed as `[A, C, B]`, etc) _and_ if updates for a specific setting were out of order.

Per-setting sequence IDs only allow us to detect the latter, but a global sequence ID gives us both the former and the latter _and_ it's simpler to implement.

pkg/config/nodetreemodel/config.go
```diff
@@ -103,6 +103,7 @@ type ntmConfig struct {
 	envTransform   map[string]func(string) interface{}
 
 	notificationReceivers []model.NotificationReceiver
+	sequenceID            uint64
```
rayz, line 106: config already has a mutex so I am not using `sync/atomic` here.

pkg/config/nodetreemodel/config.go
```diff
@@ -332,6 +338,21 @@ func (c *ntmConfig) UnsetForSource(key string, source model.Source) {
...
+
+	newValue := c.leafAtPathFromNode(key, c.root).Get()
+
+	// Value has not changed, do not notify
+	if reflect.DeepEqual(previousValue, newValue) {
+		return
+	}
+
+	c.sequenceID++
+	receivers := slices.Clone(c.notificationReceivers)
```
tobz, line 351: We don't need to clone the receivers any more since we're holding the lock until the end of the function.

pkg/config/nodetreemodel/config.go
```diff
@@ -222,16 +224,17 @@ func (c *ntmConfig) Set(key string, newValue interface{}, source model.Source) {
 	}
 
 	receivers := slices.Clone(c.notificationReceivers)
```
tobz, line 225: We don't need to clone the receivers any more since we're holding the lock until the end of the function.

pkg/config/viperconfig/viper.go
```diff
@@ -82,7 +83,6 @@ func (c *safeConfig) Set(key string, newValue interface{}, source model.Source)
 	// modify the config then release the lock to avoid deadlocks while notifying
 	var receivers []model.NotificationReceiver
 	c.Lock()
```
tobz, line 85: I think we would want to do a deferred unlock here to match the behavior we have in NTM now?

tobz: And then also we'd want to remove the declaration of `receivers`, and just `return` if the value didn't change... basically just make this look more like what we're doing in NTM.

pkg/config/viperconfig/viper.go
```diff
@@ -138,15 +140,16 @@ func (c *safeConfig) UnsetForSource(key string, source model.Source) {
 	c.configSources[source].Set(key, nil)
 	c.mergeViperInstances(key)
 	newValue := c.Viper.Get(key) // Can't use nil, so we get the newly computed value
-	if previousValue != nil {
+	if previousValue != nil && !reflect.DeepEqual(previousValue, newValue) {
 		// if the value has not changed, do not duplicate the slice so that no callback is called
 		receivers = slices.Clone(c.notificationReceivers)
+		c.sequenceID++
```
tobz, line 146: Similar comments here about making things look more like the NTM version.

tobz (changes requested):

pkg/config/viperconfig/viper.go
```diff
@@ -102,17 +101,18 @@ func (c *safeConfig) Set(key string, newValue interface{}, source model.Source)
...
 		log.Debugf("Updating setting '%s' for source '%s' with new value. notifying %d listeners", key, source, len(c.notificationReceivers))
-		// if the value has not changed, do not duplicate the slice so that no callback is called
-		receivers = slices.Clone(c.notificationReceivers)
 	} else {
+		// if the value has not changed, do not notify the receivers
 		log.Debugf("Updating setting '%s' for source '%s' with the same value, skipping notification", key, source)
+		return
 	}
-	c.Unlock()
+	// Increment the sequence ID for the key
```
tobz: Comment is slightly stale now that there's no per-key sequence ID... could probably just drop it since we don't have an equivalent comment for the NTM version.

tobz (approved):
A small non-blocking nit but otherwise LGTM.

soberpeach (approved):

ajgajg1134 (approved):
looks good from APM

pkg/config/nodetreemodel/config.go
rayz: Reverted my commit that updated this ^ since it caused tests to fail. Seems from this comment https://github.com/DataDog/datadog-agent/blob/main/pkg/config/viperconfig/viper.go#L83 we need to release the lock before notifying.

moth-a (approved):

jennchenn (approved):
lgtm for container-integrations owned file

pkg/config/nodetreemodel/config.go
dustmop: Do you recall which tests failed? I originally added this clone here because locking across the notifications being sent would cause deadlocks. I would suspect that to still be the case.

pkg/config/viperconfig/viper_test.go
```diff
@@ -444,3 +446,26 @@ func TestSetWithEnvTransformer(t *testing.T) {
 
 	assert.Equal(t, []string{"runtime"}, cfg.GetStringSlice("setting"))
 }
+
+func TestSequenceID(t *testing.T) {
+	config := NewViperConfig("test", "DD", strings.NewReplacer(".", "_")) // nolint: forbidigo
+
+	assert.Equal(t, uint64(0), config.GetSequenceID())
+
+	config.Set("foo", "bar", model.SourceFile)
```
dustmop: Not a huge deal, but using `model.SourceFile` here isn't correct. This isn't coming from a yaml file.

dustmop (approved):
LGTM, just some small comments

pkg/config/nodetreemodel/config.go
rayz: A couple of them failed dued to timeouts from deadlocks. `TestExternalMetricsProviderEndpointAndRefresh` was one example.

dustmop: Ok great, that makes sense then.

PR #38330, opened 2025-06-30
feat: fingerprint computation logic + default configuration settings

lucastemb:
### What does this PR do?
This pull-request adds the logic required to compute checksum fingerprints to more accurately detect instances of rotation and truncation

### Motivation
The motivation for this change stems from [the following](https://datadoghq.atlassian.net/wiki/spaces/AL/pages/5095229352/RFC-+Log+Rotation+Detection+Improvements+via+Fingerprintings?focusedCommentId=5189697566#Proposed-Solution-Direction) RFC. 

I tested my changes using unit tests that will be part of an additional pull-request for the complete feature. These unit tests encompass standard functionality for creating the fingerprint as well as edge cases (line is too long, not enough lines, not enough bytes, invalid config, etc). They are pictured below:

Screenshot 2025-06-30 at 3 45 42 PM

N/A


rahulkaukuntla (approved):

releasenotes/notes/fingerprinting-logic-a51a9e92474ba9ef.yaml
```diff
@@ -0,0 +1,10 @@
+# Each section from every release note are combined when the
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+features:
+  - |
+    This pull-request adds the logic required to compute checksum fingerprints to more accurately detect instances of rotation and truncation
```
brett0000FF:
```suggestion
    Adds the logic required to compute checksum fingerprints to more accurately detect instances of rotation and truncation.
```

brett0000FF (approved):

pkg/config/setup/config.go
```diff
@@ -1552,6 +1552,16 @@ func logsagent(config pkgconfigmodel.Setup) {
 	config.BindEnvAndSetDefault("logs_config.socks5_proxy_address", "")
 	// disable distributed senders
 	config.BindEnvAndSetDefault("logs_config.disable_distributed_senders", false)
+	// determines fingerprinting strategy to detect rotation and truncation
+	config.BindEnvAndSetDefault("logs_config.fingerprint_strategy", "default")
```
DDuongNguyen: Default strategy is a bit ambiguous.
To keep this more consistent with other current configs, i would add this into the [constant list](https://github.com/DataDog/datadog-agent/blob/lt/fingerprinting-review/pkg/config/setup/config.go#L41-L123) and make sure to put in the comment the available strategies too.

pkg/config/setup/config.go
```diff
@@ -1552,6 +1552,16 @@ func logsagent(config pkgconfigmodel.Setup) {
...
+	// determines maxLines before we create a fingerprint (can also be interpreted as minimum needed to fingerprint)
+	config.BindEnvAndSetDefault("logs_config.fingerprint_config.max_lines", 1)
+	// determines maxBytes before we create a fingerprint (can also be interpreted as minimum needed to fingerprint)
+	config.BindEnvAndSetDefault("logs_config.fingerprint_config.max_bytes", 2048)
```
DDuongNguyen: same logic here for this default max bytes

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,199 @@
...
+	bytesRead, err := io.ReadFull(limitedReader, buffer)
+	if err != nil && err != io.EOF && err != io.ErrUnexpectedEOF {
+		log.Warnf("Failed to read bytes for fingerprint %q: %v", filePath, err)
+		return 0
+	}
+
+	// Trim buffer to actual bytes read
+	buffer = buffer[:bytesRead]
+
+	// Check if we have enough bytes to create a meaningful fingerprint
```
DDuongNguyen, line 103: Do we not assign a fingerprint if a file is too small? There might be edge-case where a log rotation could happen on a time basis and just because there log file is small doesnt mean we shouldn't track it. 

I think its a bit inefficient to read the entire file just to discard a potential fingerprint. I think we could improve this section

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,199 @@
...
+	maxBytes := 0
+	if fingerprintConfig.MaxBytes != nil {
+		maxBytes = *fingerprintConfig.MaxBytes
+	}
+
+	// Create a limited reader for the bytes we want to hash
+	limitedReader := &io.LimitedReader{R: fpFile, N: int64(maxBytes)}
+
+	// Read up to maxBytes for hashing
+	buffer := make([]byte, maxBytes)
```
DDuongNguyen, line 93: I think this can potentially be improved. We are allocating 256 bytes buffer regardless of the actual byte sizes in the file. Additionally, in the case of small file, this buffer seems to be wasted.

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,199 @@
...
+	// Line-based fingerprinting mode
+	fingerprint := computeFingerPrintByLines(fpFile, filePath, fingerprintConfig)
+	if fingerprint == 0 {
+		log.Debugf("Not enough data for line-based fingerprinting of file %q", filePath)
+	}
+	return fingerprint
+}
+
+// computeFileFingerPrintByBytes computes fingerprint using byte-based approach for a given file path
+func computeFingerPrintByBytes(fpFile *os.File, filePath string, fingerprintConfig *logsconfig.FingerprintConfig) uint64 {
```
DDuongNguyen, line 78: Seems like you have some unit test running, but could you also check with the SMP team to see what happens when the agent is under load? 

I think comparing to the original method of just scanning for metadata and comparing data, we are potentially opening, reading, and hashing every single files the agent detect for log collection. Im curious what happens when the agent is under load and if the performance is ok at that point

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,199 @@
...
+			return 0
+		}
+	}
+
+	// Read lines for hashing
+	var buffer []byte
+	linesRead := 0
+	maxLines := 0
+	if fingerprintConfig.MaxLines != nil {
+		maxLines = *fingerprintConfig.MaxLines
```
DDuongNguyen: unlikely to happen but can potentially cause panic if the value is set to negative. So lets make sure we account for that too

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	maxBytes := 0
```
DDuongNguyen: Will definitely cause panic if maxbytes is set to negative number by user

DDuongNguyen: We can probably make a validateFingerprintConfig function to handle all these validation logic

DDuongNguyen: Also we can discuss this monday, but should we have a timeout in case someone misconfigure the maxbyte to be too big?

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,199 @@
...
+	buffer = buffer[:bytesRead]
+
+	// Check if we have enough bytes to create a meaningful fingerprint
+	if bytesRead == 0 || bytesRead < maxBytes {
+		log.Debugf("No bytes available for fingerprinting file %q", filePath)
+		return 0
+	}
+
+	// Compute fingerprint
+	table := crc64.MakeTable(crc64.ISO)
```
DDuongNguyen: Seems like you create the table twice for fingerprintbyline and fingerprintbybytes

I believe that the table created by crc64 is deterministic so you can just make it a package level variable and reuse it inside the functions. Feel free to double check on this also!

DDuongNguyen (commented):
Nice work on this fingerprinting implementation! I really like how you've made it flexible with both byte and line-based strategies, while keeping it efficient with bounded buffers and early termination. The error handling is solid too - good validation and clear error messages. 

Overall, I think we are in a really good position, it just needs a bit adjustment so we can make sure that the log agent can perform well under load.

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,199 @@
...
+		}
+	}
+
+	maxBytes := 0
+	if fingerprintConfig.MaxBytes != nil {
+		maxBytes = *fingerprintConfig.MaxBytes
+	}
+
+	// Create a limited reader for the bytes we want to hash
+	limitedReader := &io.LimitedReader{R: fpFile, N: int64(maxBytes)}
```
DDuongNguyen: love this! 💥

pkg/logs/tailers/file/fingerprint.go
lucastemb: As for the first part of the comment, @ddrthall and I talked this over and after some deliberation and input from @jszwedko, we decided that a tailer isn't spawned until we have a valid fingerprint to avoid any edge cases such as the one you listed. If we can't produce a fingerprint, it implies there is no content since the default config uses the first line to generate the checksum. 

So for now, we return 0 to signify we couldn't produce a fingerprint and thus we wait to spawn the tailer (and there is appropriate logic in place to do so). 

Also, for `computeFingerprintByBytes` bytesRead is at most maxBytes, so we're not reading the whole file, only whatever is configured, which default is 256 bytes.

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,245 @@
...
+// crc64Table is a package-level variable for the CRC64 ISO table
+// to avoid recreating it on every fingerprint computation
+var crc64Table = crc64.MakeTable(crc64.ISO)
+
+// ReturnFingerprintConfig returns the configuration for the fingerprinting algorithm set by user (also used for testing)
+func ReturnFingerprintConfig() *logsconfig.FingerprintConfig {
+	if pkgconfigsetup.Datadog().GetString("logs_config.fingerprint_strategy") != "checksum" {
+		return nil
+	}
+	maxLines := pkgconfigsetup.Datadog().GetInt("logs_config.fingerprint_config.max_lines")
```
ddrthall: Overlapping the line + byte config like this has some unideal connotations. 
For example, when using line based fingerprinting we have a "min lines", "max bytes" and "lines to skip" option set, whereas byte based fingerprinting has "min bytes" and "bytes to skip". Alongside bytesToSkip + linesToSkip being mutually exclusive, maxLines' name doesn't quite make sense in either context and maxBytes looks like it's also operating conditionally as minBytes. Would it make sense to have settable options be tied to a "type" configuration value? We would mimic the [processing rules](https://github.com/DataDog/datadog-agent/blob/7bfdddc3c2d5ef8c4716b25fa705135de6386dc2/comp/logs/agent/config/processing_rules.go) flow for validation, and we could afford to condense fields (e.g have a to_skip field name in the struct rather than both bytes_to_skip and lines_to_skip, without needing to change their names in the config)

comp/logs/agent/config/integration_config.go
```diff
@@ -130,6 +132,14 @@ type SourceAutoMultiLineOptions struct {
 	TagAggregatedJSON *bool `mapstructure:"tag_aggregated_json" json:"tag_aggregated_json" yaml:"tag_aggregated_json"`
 }
 
+// FingerprintConfig defines the options for the fingerprint configuration.
+type FingerprintConfig struct {
```
ddrthall, line 137: This doesn't look like it will be correctly pulled in to replace the global

pkg/config/setup/config.go
```diff
@@ -1558,6 +1568,16 @@ func logsagent(config pkgconfigmodel.Setup) {
 	config.BindEnvAndSetDefault("logs_config.socks5_proxy_address", "")
 	// disable distributed senders
 	config.BindEnvAndSetDefault("logs_config.disable_distributed_senders", false)
+	// determines fingerprinting strategy to detect rotation and truncation
+	config.BindEnvAndSetDefault("logs_config.fingerprint_strategy", DefaultFingerprintingStrategy)
+	// determines maxLines before we create a fingerprint (can also be interpreted as minimum needed to fingerprint)
+	config.BindEnvAndSetDefault("logs_config.fingerprint_config.max_lines", 1)
```
ddrthall: Since fingerprint_config is an internal struct, we can condense these options in a manner similar to the global processing rules

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,245 @@
...
+
+// computeFileFingerPrintByBytes computes fingerprint using byte-based approach for a given file path
+func computeFingerPrintByBytes(fpFile *os.File, filePath string, fingerprintConfig *logsconfig.FingerprintConfig) uint64 {
+	bytesToSkip := 0
+	if fingerprintConfig.BytesToSkip != nil {
+		bytesToSkip = *fingerprintConfig.BytesToSkip
+	}
+	// Skip the configured number of bytes
+	if bytesToSkip > 0 {
+		_, err := io.CopyN(io.Discard, fpFile, int64(bytesToSkip))
```
ddrthall: Seeking is a more efficient alternative to copying bytes to Discard

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+	}
+
+	maxBytes := 0
+	if fingerprintConfig.MaxBytes != nil {
+		maxBytes = *fingerprintConfig.MaxBytes
+	}
+
+	// Create a limited reader for the bytes we want to hash
+	limitedReader := &io.LimitedReader{R: fpFile, N: int64(maxBytes)}
```
ddrthall: We don't actually need to worry about enforcing a "maxBytes" value for this checksum, since the bytes value in question is actually a min - either we fill the maxBytes buffer below or we fail to fingerprint

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,245 @@
...
+	// Compute fingerprint
+	checksum := crc64.Checksum(buffer, crc64Table)
+
+	log.Debugf("Computed byte-based fingerprint 0x%x for file %q (bytes=%d)", checksum, filePath, bytesRead)
+	return checksum
+}
+
+// computeFileFingerPrintByLines computes fingerprint using line-based approach for a given file path
+func computeFingerPrintByLines(fpFile *os.File, filePath string, fingerprintConfig *logsconfig.FingerprintConfig) uint64 {
+	reader := bufio.NewReader(fpFile)
```
ddrthall: A LimitedReader combined with a Scanner will likely be more effective than the ReadString/ReadBytes logic below

comp/logs/agent/config/integration_config.go
```diff
@@ -130,6 +132,14 @@ type SourceAutoMultiLineOptions struct {
...
+	MaxBytes    *int `json:"max_bytes,omitempty" mapstructure:"max_bytes" yaml:"max_bytes"`
```
ddrthall: We likely don't want these values to be pointers - validation can enforce that they're always coherent defaults rather than nillable

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,245 @@
...
+			return fmt.Errorf("maxBytes cannot be zero")
+		}
+	}
+
+	return nil
+}
+
+// ComputeFingerprint computes the fingerprint for the given file path
+func ComputeFingerprint(filePath string, fingerprintConfig *logsconfig.FingerprintConfig) uint64 {
+	fpFile, err := os.Open(filePath)
```
ddrthall, line 48: Open isn't necessarily safe to run on windows, look at our filesystem.OpenShared functionality

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,183 @@
...
+		log.Warnf("fingerprint config is not set for file %q", filePath)
+		return 0
+	}
+
+	maxLines := fingerprintConfig.MaxLines
+
+	// Mode selection:
+	// - If maxLines is 0, it implies byte-mode as line-mode is not viable.
+	// - Otherwise, it's line-mode.
+	if maxLines == 0 {
```
ddrthall, line 65: Having an explicit request for lines vs bytes is more stable than an inferred one. We can either extend the fingerprint_strategy config option to be something like `auto`, `line_checksum`, `byte_checksum`, or shift fingerprint_strategy to be a boolean enabled/disabled value and have the type of fingerprint be added as a field in the FingerprintConfig. I'm leaning the latter, but still need to spend some more brain cycles here.

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+// computeFileFingerPrintByBytes computes fingerprint using byte-based approach for a given file path
+func computeFingerPrintByBytes(fpFile *os.File, filePath string, fingerprintConfig *logsconfig.FingerprintConfig) uint64 {
+	bytesToSkip := fingerprintConfig.ToSkip
+	// Skip the configured number of bytes
+	if bytesToSkip > 0 {
+		_, err := fpFile.Seek(int64(bytesToSkip), io.SeekStart)
+
+		if err != nil && err != io.EOF {
```
ddrthall, line 84: Should we be allowing the execution to continue if we err'ed on an EOF?

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,183 @@
...
+	// Read up to maxBytes for hashing
+	buffer := make([]byte, maxBytes)
+	bytesRead, err := io.ReadFull(fpFile, buffer)
+	if err != nil && err != io.EOF && err != io.ErrUnexpectedEOF {
+		log.Warnf("Failed to read bytes for fingerprint %q: %v", filePath, err)
+		return 0
+	}
+
+	// Trim buffer to actual bytes read
+	buffer = buffer[:bytesRead]
```
ddrthall, line 101: With ReadFull, both of these subsequent bytesRead checks/modifications shouldn't be necessary. From the ReadFull documentation: bytesRead == len(buffer) if and only if err == nil. We might want to move away from ReadFull unless we implement a re-usable buffer, but as long as it's here we don't want the extra logic.

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,183 @@
...
+
+	// Compute fingerprint
+	checksum := crc64.Checksum(buffer, crc64Table)
+
+	log.Debugf("Computed byte-based fingerprint 0x%x for file %q (bytes=%d)", checksum, filePath, bytesRead)
+	return checksum
+}
+
+// computeFileFingerPrintByLines computes fingerprint using line-based approach for a given file path
+func computeFingerPrintByLines(fpFile *os.File, filePath string, fingerprintConfig *logsconfig.FingerprintConfig) uint64 {
```
ddrthall, line 117: Recommend popping in some unit tests for this function + the bytes equivalent

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,183 @@
...
+	linesToSkip := fingerprintConfig.ToSkip
+	maxLines := fingerprintConfig.MaxLines
+	maxBytes := fingerprintConfig.MaxBytes
+
+	// Create a LimitedReader to respect maxBytes constraint
+	var limitedReader io.Reader = fpFile
+	if maxBytes > 0 {
```
ddrthall, line 124: We should always have a non-zero maxBytes value for this function to prevent oom-ing on a badly structured input file. Validation should be utilized to ensure this.

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,183 @@
...
+		limitedReader = io.LimitReader(fpFile, int64(maxBytes))
+	}
+
+	// Create scanner for line-by-line reading
+	scanner := bufio.NewScanner(limitedReader)
+
+	// Skip the configured number of lines
+	for i := 0; i < linesToSkip; i++ {
+		if !scanner.Scan() {
+			if scanner.Err() != nil {
```
ddrthall, line 134: If scanner errors with EOF  and limitedReader.N is 0, we want to fall back to a bytes fingerprint. In this case, we have reached maxBytes and we need to be able to pick the file up for processing with whatever fingerprint we can get. 
The scan loop below has the same edge case (EOF  and limitedReader.N is 0 == shift to a bytes fingerprint), and as such it might be possible to handle the two cases together

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,183 @@
...
+	var buffer []byte
+	linesRead := 0
+	bytesRead := 0
+
+	for linesRead < maxLines && scanner.Scan() {
+		line := scanner.Text()
+		lineWithNewline := line + "\n" // Add newline back for consistency
+
+		// Check if adding this line would exceed maxBytes
+		if maxBytes > 0 && bytesRead+len(lineWithNewline) > maxBytes {
```
ddrthall, line 151: The LimitedReader can be used to bypass a log of this logic - we're guaranteed that Scan will only cover at most maxBytes over its entire lifetime and will error with EOF before it provides more.

ddrthall:
Work completed under the umbrella of https://github.com/DataDog/datadog-agent/pull/39672

PR #38331, opened 2025-06-30
[pkg/dyninst] Cleanup runtime compilation

piob-io:

Remove runtime compilation code and hooks.
We no longer need to unwind loops given clang compiling to explicit ebpf target.
Report ir gen failures separately.


usamasaqib (approved):

pkg/dyninst/compiler/code.go
```diff
@@ -76,6 +77,78 @@ func GenerateCode(program Program, out CodeSerializer) (CodeMetadata, error) {
 	}, nil
 }
 
+// DispatchingSerializer is a CodeSerializer that dispatches to a list of other CodeSerializers.
+type DispatchingSerializer struct {
```
ajwerner: extremely nitpicky suggestion you can ignore if it doesn’t please you: unexport the inner implementations slice, add a constructor (probably with a variadic argument) and then call the constructor MultiCodeSerializer or MultiSerializer. Also probably put the methods on the pointer receiver because when boxing there’ll be a pointer anyway and when you unexport that field the mutability becomes ~moot

ajwerner (approved):
LGTM

pkg/dyninst/compiler/code.go
piob-io: Done, unexported the type too, kept the constructor named aligned - NewDispatchingSerializer.

piob-io: (the constructor returns the interface in this case, so it's not returning the unexported type)

PR #38332, opened 2025-06-30
[OTAGENT-431] Add systemd definitions and init scripts for DDOT packages

liustanley:
### What does this PR do?
Adds systemd service definitions, init scripts, and package scripts for DDOT Linux packages (deb/rpm) added in https://github.com/DataDog/datadog-agent/pull/37647. These are mainly based off of the existing trace agent files.

### Motivation
Necessary to support DDOT on Linux bare-metal (via deb/rpm for preview)

E2E tests


omnibus/config/templates/init-scripts-agent/sysvinit_debian.ddot.erb
```diff
@@ -0,0 +1,163 @@
...
+# Default-Stop: 0 1 6
+### END INIT INFO
+
+. /lib/lsb/init-functions
+
+ETC_DIR="<%= etc_dir %>"
+INSTALL_DIR="<%= install_dir %>"
+AGENTPATH="$INSTALL_DIR/embedded/bin/otel-agent"
+PIDFILE="$INSTALL_DIR/run/otel-agent.pid"
+AGENT_ARGS="--config=$ETC_DIR/datadog.yaml --pidfile=$PIDFILE"
```
truthbk: I think we're missing the args for the otel-specific config here. I also don't think we're actually supporting a pidfile in the OTel Collector, so we might want to add support for that see why [here](https://unix.stackexchange.com/questions/12815/what-are-pid-and-lock-files-for).

omnibus/config/templates/init-scripts-agent/upstart_debian.ddot.conf.erb
```diff
@@ -0,0 +1,24 @@
...
+# Logging to console from the agent is disabled since the agent already logs using file or
+# syslog depending on its configuration. We make upstart log what the process still outputs in order
+# to log panics/crashes to /var/log/upstart/datadog-agent.log
+console log
+env DD_LOG_TO_CONSOLE=false
+
+setuid dd-agent
+
+script
+  exec <%= install_dir %>/embedded/bin/otel-agent --config <%= etc_dir %>/datadog.yaml --pidfile <%= install_dir %>/run/otel-agent.pid
```
truthbk: ditto about the pid file and the otel-specific config file.

Also, I think we need to define the dependency with the core otel-agent.

omnibus/config/templates/init-scripts-agent/upstart_redhat.ddot.conf.erb
```diff
@@ -0,0 +1,33 @@
...
+console output
+
+script
+  # Logging to console from the agent is disabled since the agent already logs using file or
+  # syslog depending on its configuration. We then redirect the stdout/stderr of the agent process
+  # to log panic/crashes.
+  #
+  # setuid is not available in versions of upstart before 1.4. CentOS/RHEL6 use an earlier version of upstart.
+  # This is the best way to set the user in the absence of setuid.
+  exec su -s /bin/sh -c 'DD_LOG_TO_CONSOLE=false exec "$0" "$@"' dd-agent -- <%= install_dir %>/embedded/bin/otel-agent  --config <%= etc_dir %>/datadog.yaml --pidfile <%= install_dir %>/run/otel-agent.pid &>> /var/log/datadog/otel-errors.log
```
truthbk: ditto

omnibus/package-scripts/agent-deb/preinst
```diff
@@ -16,18 +16,21 @@ if [ -f "/lib/systemd/system/$SERVICE_NAME.service" ] || [ -f "/usr/lib/systemd/
         systemctl stop $SERVICE_NAME-sysprobe || true
         systemctl stop $SERVICE_NAME-trace || true
         systemctl stop $SERVICE_NAME-security || true
+        systemctl stop $SERVICE_NAME-ddot || true
```
truthbk, line 19: This would be specific to the DDOT package only, so I think we're going to need to something specific. We'll need to copy this same structure and make something like `omnibus/package-scripts/ddot-deb/preinst` and so forth.

pkg/fleet/installer/packages/datadog_agent_linux.go
```diff
@@ -99,8 +99,8 @@ var (
 	agentService = datadogAgentService{
 		SystemdMainUnitStable: "datadog-agent.service",
 		SystemdMainUnitExp:    "datadog-agent-exp.service",
-		SystemdUnitsStable:    []string{"datadog-agent.service", "datadog-agent-installer.service", "datadog-agent-trace.service", "datadog-agent-process.service", "datadog-agent-sysprobe.service", "datadog-agent-security.service"},
-		SystemdUnitsExp:       []string{"datadog-agent-exp.service", "datadog-agent-installer-exp.service", "datadog-agent-trace-exp.service", "datadog-agent-process-exp.service", "datadog-agent-sysprobe-exp.service", "datadog-agent-security-exp.service"},
+		SystemdUnitsStable:    []string{"datadog-agent.service", "datadog-agent-installer.service", "datadog-agent-trace.service", "datadog-agent-process.service", "datadog-agent-sysprobe.service", "datadog-agent-security.service", "datadog-agent-ddot.service"},
```
truthbk, line 101: This is for the DD Installer, I think, so probably out-of-scope for now. We need their review anyways, so reach out to them too. I think this will come later.

pkg/fleet/installer/packages/embedded/templates/datadog-agent-ddot.service.tmpl
```diff
@@ -0,0 +1,26 @@
+f[Unit]
```
truthbk: Again, these are the installer systemd units, which we don't need to support at the moment. But let's get their review too.

truthbk (commented):
Some feedback. I'm surprised about the systemd units only being present for the `fleet/installer`, it looks like they may have moved those, so I have to review the current process to review that part. If I'm slow, we can get Fleet Automation's feeback in the meantime.

omnibus/config/templates/init-scripts-agent/sysvinit_debian.ddot.erb
BaptisteFoy, line 1: Do we still support OSes running sysvinit?

pkg/fleet/installer/packages/datadog_agent_linux.go
BaptisteFoy: This is for **agent** systemd units (regardless of oci/deb/rpm). Your instructions shouldn't live here, but you have two choices:
1.  If you know the installer will be here (if the agent is installed first); you could create a `datadog_agent_otel_linux.go` file and put your prerm/postinst instructions here. Then add the corresponding installer commands (but that'll require a minimum agent version); and call the installer from your bash prerm/postinst
2.  Write the prerm/postinst instructions in bash directly


I believe the second option to be safer for your use case as it doesn't require pinning the agent version to the DDOT package; but if it's already pinned then option 1 is viable

pkg/fleet/installer/packages/embedded/templates/datadog-agent-ddot.service.tmpl
BaptisteFoy: I think there's an extra `f`?
```suggestion
[Unit]
```

These are not installer-specific units (used in deb/rpm as well) but they should go in your deb package, not the agent's. Again unless we manage to pin agent & DDOT versions somehow.

pkg/fleet/installer/packages/embedded/templates/datadog-agent.service.tmpl
```diff
@@ -3,12 +3,12 @@
 Description=Datadog Agent
 After=network.target
 Conflicts=datadog-agent-exp.service
-Wants=datadog-agent-installer.service datadog-agent-trace.service datadog-agent-process.service datadog-agent-sysprobe.service datadog-agent-security.service
+Wants=datadog-agent-installer.service datadog-agent-trace.service datadog-agent-process.service datadog-agent-sysprobe.service datadog-agent-security.service datadog-agent-ddot.service
```
BaptisteFoy, line 6: I'm not sure this is safe if the unit doesn't exist

omnibus/config/templates/init-scripts-agent/upstart_debian.ddot.conf.erb
liustanley: I might need to check with the FA team on this but I'm wondering if this would define the dependency?
```
start on started datadog-agent
stop on (runlevel [!2345] or stopping datadog-agent)
```

pkg/fleet/installer/packages/embedded/templates/datadog-agent-ddot.service.tmpl
liustanley: Where should these files live if we want them to be in our deb package?

pkg/fleet/installer/packages/datadog_agent_linux.go
liustanley: I believe we always want the agent version to be the same as the DDOT version. But I'm not sure if we can guarantee that the installer will be here since some DDOT customers aren't using the installer.

omnibus/package-scripts/agent-deb/preinst
liustanley: Just added `omnibus/package-scripts/ddot-deb`

pkg/fleet/installer/packages/datadog_agent_linux.go
alopezz: At least on debian, given that ddot has a `Depends` relationship towards datadog-agent, the [Debian docs](https://www.debian.org/doc/debian-policy/ch-relationships.html) indicate that it would be fine for ddot `postinst` / `prerm` scripts to assume that datadog-agent is already installed (though the wording doesn't give a lot of confidence):

> The Depends field should also be used if the postinst or prerm scripts require the depended-on package to be unpacked or configured in order to run.

There's also `Pre-Depends`, which does seem to offer stronger guarantees:

> also forces dpkg to complete installation of the packages named before even starting the installation of the package which declares the pre-dependency

I couldn't find anything on rpm, it would require a bit more digging (unless anyone here happens to know).

BaptisteFoy: To add to @alopezz's comment:

> But I'm not sure if we can guarantee that the installer will be here

The installer binary is embedded in the agent package (since 7.67 at least), so even if it's not running we'd have access to its CLI, we only need the agent to be installed which is already one of your requirements 😄 . This is how we execute the postinst of the agent today

pkg/fleet/installer/packages/embedded/templates/datadog-agent.service.tmpl
liustanley: Should we remove `datadog-agent-ddot.service` from all of the `datadog-agent.service` files?

BaptisteFoy: After discussing offline -- it's fine to ignore this comment if we add a `ConditionPathExists` in the DDOT unit

omnibus/package-scripts/ddot-deb/postinst
```diff
@@ -0,0 +1,21 @@
...
+INSTALL_DIR=/opt/datadog-agent/embedded/bin/otel-agent
+
+# If we are inside the Docker container, do nothing
+if [ -n "$DOCKER_DD_AGENT" ]; then
+    echo "Installation from docker-dd-agent, nothing to do in postinst"
+    exit 0
+fi
+
+# Run the postinst. See pkg/fleet/installer/packages/datadog_agent_linux.go
+${INSTALL_DIR}/embedded/bin/installer postinst datadog-agent-ddot deb || true
```
BaptisteFoy: I believe this is not needed if we do the `ConditionPathExists` trick -- the agent postinst is enough. If you have more instructions though that may still be needed in which case you need to create your own postinst/prerm Go files

omnibus/package-scripts/ddot-deb/postrm
```diff
@@ -0,0 +1,28 @@
...
+        echo "Deleting dd-agent group"
+        (getent group dd-agent >/dev/null && delgroup dd-agent --quiet) || true
+        echo "Force-deleting $INSTALL_DIR"
+        rm -rf $INSTALL_DIR
+        rm -rf $LOG_DIR
+        rm -rf $CONFIG_DIR
+    ;;
+    *)
+    ;;
+esac
```
BaptisteFoy: ⚠️ Except removing the embedded/bin/otel-agent, all of these instructions are destructive to the agent and should be handled by the agent postrm only.

omnibus/package-scripts/ddot-deb/prerm
```diff
@@ -0,0 +1,25 @@
...
+# Run the postinst. See pkg/fleet/installer/packages/datadog_agent_linux.go
+if [ -f ${INSTALLER_DEB} ] && [ "$1" = "remove" ]; then
+    ${INSTALLER_DEB} prerm datadog-agent-ddot deb || true
+elif [ -f ${INSTALLER_DEB} ] && [ "$1" = "upgrade" ]; then
+    ${INSTALLER_DEB} prerm --upgrade datadog-agent-ddot deb || true
+fi
+
+# Remove the agent if it was upgraded using the installer
+if [ -f ${INSTALLER_OCI} ] && [ "$1" = "remove" ]; then
+    ${INSTALLER_OCI} remove datadog-agen-ddott || printf "[ WARNING ]\tFailed to remove datadog-agent-ddot installed by the installer\n"
```
BaptisteFoy: typo

omnibus/package-scripts/ddot-deb/prerm
```diff
@@ -0,0 +1,25 @@
...
+
+INSTALLER_DEB=/opt/datadog-agent/embedded/bin/installer
+INSTALLER_OCI=/opt/datadog-package/datadog-agent/stable/embedded/bin/installer
+
+# Run the postinst. See pkg/fleet/installer/packages/datadog_agent_linux.go
+if [ -f ${INSTALLER_DEB} ] && [ "$1" = "remove" ]; then
+    ${INSTALLER_DEB} prerm datadog-agent-ddot deb || true
+elif [ -f ${INSTALLER_DEB} ] && [ "$1" = "upgrade" ]; then
+    ${INSTALLER_DEB} prerm --upgrade datadog-agent-ddot deb || true
+fi
```
BaptisteFoy: Same as the postinst -- probably not needed?

pkg/fleet/installer/packages/datadog_agent_linux.go
BaptisteFoy: If we use the `ConditionPathExists` trick my first comment here is outdated :)

omnibus/package-scripts/agent-deb/preinst
liustanley: Discussed with @BaptisteFoy and decided that we don't need `package-scripts/ddot-deb` since the only setup/removal we need are the DDOT systemd units (which are being handled by `agent-deb`). Do we need to add this line back though?

omnibus/config/templates/init-scripts-agent/upstart_debian.ddot.conf.erb
liustanley: cc @BaptisteFoy is there anything else needed to define the dependency with the core agent?

omnibus/package-scripts/agent-deb/preinst
liustanley: Added these lines back

omnibus/package-scripts/ddot-deb/postinst
liustanley: Added `ConditionPathExists`

omnibus/config/templates/init-scripts-agent/upstart_debian.ddot.conf.erb
BaptisteFoy: I'm not too well-versed in upstart; but I think your previous comment shows that this should be good

BaptisteFoy (approved):
Overall the changes LGTM, nice work!
Let's just make sure all of the E2E test pass (especially new-e2e-installer); and to trigger them if they're not automatically ran

pkg/fleet/installer/packages/embedded/templates/datadog-agent-ddot.service.tmpl
```diff
@@ -0,0 +1,28 @@
...
+After=datadog-agent.service datadog-agent-exp.service
+BindsTo=datadog-agent.service
+Conflicts=datadog-agent-exp.service datadog-agent-ddot-exp.service
+ConditionPathExists={{.InstallDir}}/embedded/bin/otel-agent
+{{- else}}
+Description=Datadog Distribution of OpenTelemetry Collector Experiment
+After=datadog-agent.service datadog-agent-exp.service
+BindsTo=datadog-agent-exp.service
+Conflicts=datadog-agent.service datadog-agent-ddot.service
+ConditionPathExists={{.InstallDir}}/embedded/bin/otel-agent
```
truthbk, line 13: 🍰

pkg/fleet/installer/packages/embedded/templates/gen/debrpm/datadog-agent-ddot.service
```diff
@@ -0,0 +1,20 @@
...
+ConditionPathExists=/opt/datadog-agent/embedded/bin/otel-agent
+
+[Service]
+Type=simple
+PIDFile=/opt/datadog-agent/run/otel-agent.pid
+User=dd-agent
+Restart=on-failure
+EnvironmentFile=-/etc/datadog-agent/environment
+Environment="DD_FLEET_POLICIES_DIR=/etc/datadog-agent/managed/datadog-agent/stable"
+ExecStart=/opt/datadog-agent/embedded/bin/otel-agent --config /etc/otel-agent/otel-config.yaml --core-config /etc/datadog-agent/datadog.yaml --pidfile /opt/datadog-agent/run/otel-agent.pid
```
truthbk: I _think_ the default location for the `otel-config.yaml` is in `/etc/datadog-agent` too. That's where we put it in the docker image too. You'll have to fix that up in the other `service` units too.

omnibus/config/templates/init-scripts-agent/upstart_debian.ddot.conf.erb
```diff
@@ -0,0 +1,24 @@
...
+# Logging to console from the agent is disabled since the agent already logs using file or
+# syslog depending on its configuration. We make upstart log what the process still outputs in order
+# to log panics/crashes to /var/log/upstart/datadog-agent.log
+console log
+env DD_LOG_TO_CONSOLE=false
+
+setuid dd-agent
+
+script
+  exec <%= install_dir %>/embedded/bin/otel-agent --config /etc/otel-agent/otel-config.yaml --core-config <%= etc_dir %>/datadog.yaml --pidfile <%= install_dir %>/run/otel-agent.pid
```
truthbk: bad location for the `otel-config.yaml`

truthbk (commented):
Looks almost ready to me.

cmd/otel-agent/subcommands/run/command_test.go
```diff
@@ -21,8 +21,10 @@ func TestFxRun_WithDatadogExporter(t *testing.T) {
 	t.Setenv("DD_OTELCOLLECTOR_ENABLED", "true")
 	fxutil.TestRun(t, func() error {
 		ctx := context.Background()
-		params := &subcommands.GlobalParams{
-			ConfPaths: []string{"test_config.yaml"},
+		params := &cliParams{
```
truthbk, line 24: Don't have a super strong opinion, but it might be worth adding a test to check the pid file creation. Not sure if that test would land here or somewhere else, but it might be worth adding.

pkg/fleet/installer/packages/embedded/templates/gen/debrpm/datadog-agent-ddot-exp.service
```diff
@@ -0,0 +1,20 @@
+[Unit]
+Description=Datadog Distribution of OpenTelemetry Collector Experiment
+After=datadog-agent.service datadog-agent-exp.service
```
truthbk, line 3: Does it make sense to have `datadog-agent.service` here if it conflicts with it? Also, if you read my comment below, it's important to determine if `datadog-agent.service` is a real conflict.

pkg/fleet/installer/packages/embedded/templates/gen/debrpm/datadog-agent-ddot.service
```diff
@@ -0,0 +1,20 @@
+[Unit]
+Description=Datadog Distribution of OpenTelemetry Collector
+After=datadog-agent.service datadog-agent-exp.service
```
truthbk, line 3: Similar, comment, does it make sense to have `datadog-agent-exp.service` here if it conflicts with it?

truthbk (commented):
Couple of questions, just trying to be thorough. This looks mostly ready.

pkg/fleet/installer/packages/embedded/templates/datadog-agent.service.tmpl
alopezz: Is it necessary, though? Meaning, does it actually change behaviour meaningfully when compared to not having it.

alopezz (commented):
How have we validated that this works, at least for systemd?

We might want to perhaps add a few extra assertions to the tests on https://github.com/DataDog/datadog-agent/blob/main/test/new-e2e/tests/agent-platform/ddot/ddot_install_test.go to confirm at the very least that the expected services are running as expected.

cmd/otel-agent/subcommands/run/command_test.go
liustanley: Sounds good, added validation for this in the integration test

pkg/fleet/installer/packages/embedded/templates/gen/debrpm/datadog-agent-ddot-exp.service
liustanley: I'm not 100% sure on this, I added this line in order to mirror the corresponding trace-agent file: https://github.com/DataDog/datadog-agent/blob/3010d59475653e9a9e24ffe3e06d3d6a36d10496/pkg/fleet/installer/packages/embedded/templates/gen/debrpm/datadog-agent-trace.service#L5.

@BaptisteFoy would you happen to know the answer?

BaptisteFoy: Yes this is expected! Having this in both stable & exp units enforce all units of a specific group (stable/exp) stop before starting any other from the opposite group; avoiding any eventual "contamination". Here, this means we'll stop ddot-exp before starting the stable core agent.

pkg/fleet/installer/packages/embedded/templates/datadog-agent.service.tmpl
BaptisteFoy: If we don't have that `Wants` it the DDOT won't start when the core agent is (re)started. Our customers usually only restart the main agent unit when updating config

truthbk (approved):

test/new-e2e/tests/agent-platform/ddot/ddot_install_test.go
```diff
@@ -114,9 +126,39 @@ func (is *ddotInstallSuite) TestDDOTInstall() {
...
+		}
+	})
+
+	is.T().Run("check ddot systemd units are running", func(t *testing.T) {
+		ddotUnit := is.host.State().Units["datadog-agent-ddot.service"]
+		require.NotNil(t, ddotUnit)
+		require.Equal(t, "datadog-agent-ddot.service", ddotUnit.Name)
+		require.Equal(t, "active", ddotUnit.Active)
+		require.Equal(t, host.Loaded, ddotUnit.LoadState)
+		require.Equal(t, host.Running, ddotUnit.SubState)
```
alopezz: Suggestion: add assertion messages to these.

test/new-e2e/tests/agent-platform/ddot/ddot_install_test.go
```diff
@@ -114,9 +126,39 @@ func (is *ddotInstallSuite) TestDDOTInstall() {
...
+
+		is.T().Log("--- ConfigureAndRunAgentService ---")
+		for s, u := range is.host.State().Units {
+			is.T().Log(s, u.Name, u.Enabled, u.Active, u.LoadState, u.SubState)
+		}
+	})
+}
```
alopezz: These seem to add a lot of [noise to test output](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1033028557#L210). I can see their use while debugging, but perhaps we no longer need it or we can at least filter for units that may be relevant to the test?

alopezz (approved):

khewonc (approved):

PR #38333, opened 2025-06-30
[AGNTLOG-243] Add Filtering to MRF Feature

soberpeach:
### What does this PR do?

Allows the agent to send a filtered subset of logs to a backup different backup endpoint when MRF is triggered. Allowed logs services are set through the `multi_region_failover.logs_allowlist` variable. 

The change only works when using the `http_sender`, and works in the following manner:
1. If MRF has been triggered, the `processor` will begin using the `multi_region_failover.logs_allowlist` to filter logs by services. Allowed logs will be tagged and sent to the batcher.
2. The batcher will batch normal logs and MRF logs separately then send them each to the `sender/worker`. The MRF payloads will be tagged with an `isMRF` tag.
3. The sender/worker will send all logs to every destination, but will drop non-MRF payloads to non-MRF destinations. 

### Motivation

Needed by MRF

1. Add the following configuration to the `datadog.yaml`
```yaml
multi_region_failover:
  enabled: true
  site: 127.0.0.1:9999
  api_key: api_key
  failover_logs: true
  logs_service_allowlist:
    - mrf-service

logs_enabled: true
log_level: 'debug'
```
2. Have the logs agent tail two log files, one who's logs should be filtered and one who's shouldn't be:
```yaml
logs:
  - type: file
    path: "/var/log/myapp.log"
    service: "mrf-service"
    source: "mrf-source"
```
```yaml
logs:
  - type: file
    path: "/var/log/another_app.log"
    service: "non-mrf-service"
    source: "non-mrf-source"
```
and write logs to both files with the one-liners:
```bash
while true; do echo "$(date) - INFO: Application running" >> /var/log/myapp.log; sleep 1; done
```
```bash
while true; do echo "ANOTHER LOG" >> /var/log/another_app.log; sleep 1; done
```
3. Run the agent and check for the log: `Dropping non-MRF payload to MRF destination: https://agent-http-intake.logs.mrf.127.0.0.1:9999/api/v2/logs`

While it's not possible to tell exactly in which logs payloads have been dropped to the MRF destination while running the agent normally, if you debug the agent, you can see which payloads are making it through (payloads tagged with `IsMRF`).

Slight increase in memory as a result of holding two batches in the batcher as opposed to one.


estherk15 (approved):

rahulkaukuntla (approved):

pkg/logs/processor/processor.go
```diff
@@ -167,6 +169,20 @@ func (p *Processor) run() {
 	}
 }
 
+func (p *Processor) getFailoverAllowlist() (bool, map[string]struct{}) {
+	failoverActive := p.config.GetBool("multi_region_failover.failover_logs")
+	var allowlist map[string]struct{}
+	if failoverActive && p.config.IsConfigured("multi_region_failover.logs_allowlist") {
```
jszwedko:
```suggestion
	if failoverActive && p.config.IsConfigured("multi_region_failover.logs_service_allowlist") {
```

I'd suggest making the option name here more self-describing.

soberpeach: Changed!

pkg/logs/processor/processor.go
```diff
@@ -244,6 +260,12 @@ func (p *Processor) processMessage(msg *message.Message) {
 		// report this message to diagnostic receivers (e.g. `stream-logs` command)
 		p.diagnosticMessageReceiver.HandleMessage(msg, rendered, "")
 
+		failoverActiveForMRF, allowlistForMRF := p.getFailoverAllowlist()
```
remeh: This is reading the configuration for every processed message, maybe part of the hit on performance.

remeh (commented):
I think the approach looks good overall but an early feedback before I complete my review: I'm worried about the impact on CPU performance reported by the SMP quality gates / experiments, it looks like it's a hit for all use-cases, even users not using MRF.
It would be worth rebasing on main and checking the profiles generated by SMP, to validate what's the reason of this performance hit.

pkg/logs/processor/processor.go
soberpeach: Ah that makes sense, I'll try to figure out where the performance hit is coming from, this looks like a good starting point.

soberpeach: I've changed the way this works so it the setting receives updates from the config using the `onUpdate` callback which should mediate the constant polling issue. I'm still seeing that it's not passing the quality gates, but it's not immediately clear what exactly is causing this. I'll have to dig into this some more.

pkg/logs/processor/processor.go
```diff
@@ -58,6 +66,54 @@ func New(inputChan, outputChan chan *message.Message, processingRules []*config.
...
+	}
+}
+
+// updateFailoverConfig updates the cached failover configuration
+func (p *Processor) updateFailoverConfig() {
+	if p.config == nil {
+		return
+	}
+
+	p.failoverMu.Lock()
```
ddrthall: Nit: We can remove the locks by moving the configuration updates in to the run loop - onLogsFailoverSettingChanged can jettison a struct on to a new channel that run can select on alongside the inputChan. While this might make the logic slightly simpler, since we're not exactly expecting a ton of contention on  this lock I'm not sure that this change would actually translate in to a performance boost.

pkg/logs/sender/worker.go
```diff
@@ -118,6 +119,13 @@ func (s *worker) run() {
 			sent := false
 			for !sent {
 				for _, destSender := range reliableDestinations {
+					// Drop non-MRF payloads to MRF destinations
+					if destSender.destination.IsMRF() && !payload.IsMRF() {
```
ddrthall, line 123: Will this be safe for MRF configurations that are legacy or that don't want to have an explicit allow list of services?

pkg/logs/processor/processor.go
```diff
@@ -58,6 +66,54 @@ func New(inputChan, outputChan chan *message.Message, processingRules []*config.
...
+		}
+	}
+	p.failoverAllowlist = allowlist
+}
+
+// getCachedFailoverConfig returns the cached failover configuration
+func (p *Processor) getCachedFailoverConfig() (bool, map[string]struct{}) {
+	p.failoverMu.RLock()
+	defer p.failoverMu.RUnlock()
+	return p.failoverActive, p.failoverAllowlist
```
ddrthall: If you don't refactor according to the above nit, then you'll need to copy the failoverAllowList in order to make it a "cached" config.

pkg/logs/pipeline/pipeline.go
```diff
@@ -102,12 +102,19 @@ func getStrategy(
 	instanceID string,
 ) sender.Strategy {
 	if endpoints.UseHTTP || serverlessMeta.IsEnabled() {
-		var encoder compressioncommon.Compressor
```
ddrthall: Nit: Curious why we're pushing the encoder creation into the batch_strategy - want to make sure I'm not missing a functional component

pkg/logs/sender/batch_strategy.go
```diff
@@ -13,103 +13,138 @@ import (
...
 		utilization:     pipelineMonitor.MakeUtilizationMonitor(metrics.StrategyTlmName, instanceID),
+		maxBatchSize:    maxBatchSize,
+		maxContentSize:  maxContentSize,
+		encoder:         encoder,
 		instanceID:      instanceID,
 	}
-	bs.MakeCompressor()
+
+	bs.mainBatch = bs.MakeBatch()
+	bs.mrfBatch = bs.MakeBatch()
```
ddrthall: Would it make sense to only initialize the mrfBatch when we know we need it? It's not a huge amount of memory usage, but it is memory that we don't expect the vast majority of users to need.

pkg/logs/pipeline/pipeline.go
soberpeach: It's so each batch in `batch_strategy` can have it's own compressor.

pkg/logs/processor/processor.go
soberpeach: This seems like a more elegant solution. I've changed it

pkg/logs/pipeline/pipeline.go
ddrthall: That doesn't look like it's true, we recreate the encoder immediately in the constructor ([here](https://github.com/DataDog/datadog-agent/pull/38333/files#diff-ba94f7aaec863db5b437320428d8778b3381e2e728aae094e8d2dc33973f9b5dR98)). Rather than a unique Compressor per batch we're using a distinct StreamCompressor per batch, which from what I can tell didn't need the encoder restructure. Still just a nit - not a fan of making the batcher endpoint-aware if it doesn't need to be, but will survive just fine in a world where it is.

pkg/logs/sender/batch_strategy.go
soberpeach: Yep, changed it to match this

pkg/logs/pipeline/pipeline.go
soberpeach: Hmm taking another look at this. I think it makes sense to change it back to the old way if possible then

soberpeach: Changed back

pkg/logs/sender/worker.go
soberpeach: I don't think so, I'll have to think about this

soberpeach: Actually, I think it should be alright? What this is doing is just dropping non MRF tagged logs payloads to MRF destinations.

soberpeach: I've changed the way the processor works to only filter logs failover logs when an allowlist is set, otherwise it'll failover everything.

pkg/logs/processor/processor.go
```diff
@@ -23,6 +24,23 @@ import (
 // content for tailers capable of processing both unstructured and structured content.
 const UnstructuredProcessingMetricName = "datadog.logs_agent.tailer.unstructured_processing"
 
+var severities = map[string]int{
```
remeh: WDYT of having a Trace level here?
While I know nobody's using the same levels, it would be silly to not match ours in our product. It's more verbose than Debug, so you'll have to offset all of them here by having `message.StatusTrace: 0`.

pkg/logs/processor/processor.go
```diff
@@ -34,30 +52,88 @@ type Processor struct {
...
+		config.OnUpdate(p.onLogsFailoverSettingChanged)
+	}
+
+	return p
+}
+
+// onLogsFailoverSettingChanged is called when any config value changes
+func (p *Processor) onLogsFailoverSettingChanged(setting string, _ pkgconfigmodel.Source, _, _ any, _ uint64) {
+	// Only update if the changed setting affects failover configuration
+	failoverSettingChanged := setting == "multi_region_failover.failover_logs" || setting == "multi_region_failover.logs_service_allowlist" || setting == "multi_region_failover.logs_min_level"
```
remeh: Please consider using consts strings for the config fields, we have several references of them.

```
const (
  configMRFFailoverLogs = "multi_region_failover.failover_logs"
  configMRFServiceAllowlist = "multi_region_failover.logs_service_allowlist"
```
WDYT?

To push it even further, then you could even have a:

```
var MRFConfigFields = []string{configMRFFailoverLogs, configMRFServiceAllowlist, etc}
```

and test using this slices methods instead of this long `if` with `or`s:
```
failoverSettingChanged := slices.Contains(MRFConfigFields, setting)
```

pkg/logs/processor/processor.go
```diff
@@ -34,30 +52,88 @@ type Processor struct {
...
+	}
+}
+
+// updateFailoverConfig sends the updated config to the processor to update
+func (p *Processor) updateFailoverConfig() {
+	if p.config == nil {
+		return
+	}
+
+	var conf failoverConfig
```
remeh: I know you did it some lines below, but since the default value for `conf.failoverMinLevel` is `-1` and not `0`, consider setting it at the creation time.

```
conf := failoverConfig{
  failoverMinLevel: -1,
  isFailoverActive: p.config.GetBool("multi_region_failover.failover_logs"),
}
```
It will avoid a potential mistake in a future refactoring.

Another way would have to have a constructor attached to `failoverConfig` but I don't think it's necessary here.

pkg/logs/processor/processor.go
```diff
@@ -97,12 +173,19 @@ func (p *Processor) run() {
...
-		p.processMessage(msg)
-		p.mu.Lock() // block here if we're trying to flush synchronously
-		//nolint:staticcheck
-		p.mu.Unlock()
+	for {
+		select {
+		case msg, ok := <-p.inputChan:
+			if !ok {
+				return
+			}
```
remeh, line 166: Any reason to add this validation compared to the original code?
I don't see a problem with it I'm just wondering if I've missed something.

pkg/logs/processor/processor.go
```diff
@@ -138,7 +225,26 @@ func (p *Processor) processMessage(msg *message.Message) {
...
+	// 2. The message service is in the service allowlist.
+	// 3. The message status is at or above the configured minimum log level.
+	noFilters := len(serviceAllowlist) == 0 && minLevel < 0
+	_, serviceMatch := serviceAllowlist[msg.Origin.Service()]
+	msgSeverity, hasSeverity := severities[msg.GetStatus()]
+	statusMatch := minLevel >= 0 && hasSeverity && msgSeverity >= minLevel
+
+	if noFilters || serviceMatch || statusMatch {
+		msg.IsMRFAllow = true
+	}
```
remeh: Because this is a hot path in the Agent, I would consider early returns here. You might win only a few ops, but they will matter, especially since the first if is the less costly and will be the most used branch.
E.g.
```suggestion
	if len(serviceAllowlist) == 0 && minLevel < 0 {
		msg.IsMRFAllow = true
		return
	}
	
	_, serviceMatch := serviceAllowlist[msg.Origin.Service()]
	if serviceMatch {
		msg.IsMRFAllow = true
		return
	}

	msgSeverity, hasSeverity := severities[msg.GetStatus()]
	statusMatch := minLevel >= 0 && hasSeverity && msgSeverity >= minLevel
	if statusMatch {
		msg.IsMRFAllow = true
	}
```

pkg/logs/sender/batch_strategy.go
```diff
@@ -164,30 +190,45 @@ func (s *batchStrategy) addMessage(m *message.Message) (bool, error) {
 	return false, nil
 }
 
+func (s *batchStrategy) chooseBatch(m *message.Message) *batch {
```
remeh, line 193: I'm not sure to understand why you maintain two batches, while in the end you test:
```
if destSender.destination.IsMRF() && !payload.IsMRF() {
```
to see if the message has to be sent to the failover region.

What does having two separate batches provide you?

pkg/logs/processor/processor.go
soberpeach: I think that makes sense. And yeah, it doesn't seem like everyone uses the same log levels.

pkg/logs/processor/processor.go
soberpeach: Good suggestions 👍

pkg/logs/processor/processor.go
soberpeach: Makes sense, intention seems clearer this way.

pkg/logs/processor/processor.go
soberpeach: I actually had trouble with this too. The original code (what's currently running in the processor) uses a `for ... range` loop: 
```go
for msg := range p.inputChan {
```
Since the `for ... range` loop only reads from one channel, it seems like go handles the channel closure implicitly by just closing the loop. This would then terminated the run loop.

I've changed to using a `for ... select` loop to be able to read from the config channel too:
```go
	for {
		select {
		case msg, ok := <-p.inputChan:
			if !ok {
				return
			}
			...
		case conf := <-p.configChan:
		}
	}
```
Originally,  I added the second case and didn't add the check. The result was that the processor would panic. It took a while, but I eventually found that when the `p.inputChan` channel was closed, its `case` will always be selected and sent a zero value (`nil` in this case). Trying to read properties from the `nil` was causing the panics, therefore a check was needed to verify the channel was closed so the `run` loop could terminate. I guess this makes sense: how would the `for ... select` loop know when to terminate if it's reading from two channels? It seems other `run` loops in the agent also have a `done` channel signaling completion of the loop,  but this seemed like the most straightforward case to do it in this instance.

https://gobyexample.com/closing-channels

pkg/logs/processor/processor.go
soberpeach: Ah makes sense. I originally had something similar to that but decided it looked ugly and to just include all the checks in one `if` statement. Early returns makes sense considering the performance implications though.

pkg/logs/sender/batch_strategy.go
soberpeach: For MRF destinations, you want to send payloads composed of **only** MRF messages. Without batching, payloads will be composed of both MRF and non-MRF messages, so it would be impossible to discriminate where each payload should go to.

pkg/logs/processor/processor.go
remeh: I'm learning today about these message.Status*. Do you know what they're used for?

pkg/logs/processor/processor.go
remeh: Got it, indeed!

pkg/logs/sender/batch_strategy.go
remeh: But isn't that filtering actually done L123 of `pkg/logs/sender/worker.go` (the snippet inline above)?

pkg/logs/processor/processor.go
remeh: And shouldn't we introduce a `StatusTrace` instead of relying on the const "trace" you added here?

soberpeach: Your comment prompted me to do some more research about `message.Status`. My conclusion is that I don't think adding level filtering is feasible for the first version of this feature. Explanation below.

I was curious about where these `status` levels came from: apparently they are the syslog standard that many logging mechanisms use (though these levels are not universal, the agent's own internal logging system does not use them for example). My assumption was that the agent _already_ had some sort of parsing mechanism for logs and would then somehow assign these standardized syslog log levels to all logs it receives. As it turns out, this isn't the case. The reality is that logs are almost always just assigned the `info` level [example 1](https://github.com/DataDog/datadog-agent/blob/main/pkg/logs/internal/parsers/kubernetes/kubernetes.go#L93), [example 2](https://github.com/DataDog/datadog-agent/blob/main/pkg/logs/internal/parsers/kubernetes/kubernetes.go#L93), and when they're not assigned any level, querying their status also just returns `info`: [`GetStatus` function](https://github.com/DataDog/datadog-agent/blob/main/pkg/logs/message/message.go#L373). I think if we were going to include filtering by log level then, it would require adding some sort of regex matching every log line or another some other mechanism. I think this would be quite CPU intensive. For these reasons, I think it is better to leave filtering by log out of this iteration of the feature.

pkg/logs/sender/batch_strategy.go
soberpeach: Ah I think I can see where to the confusion is. Basically `payload` = `batch`, payloads and batches are the same thing. For the rest of the comment I will refer to batches and payloads interchangeably. Under the old system with a single batch, the batcher would mix a bunch of mrf and non-mrf messages into one payload that the `worker` would have to do something with. Where would you send this batch of mixed messages if you don't want to send non-mrf payloads to mrf destinations? There's no way of knowing which destinations should get this batch since all the messages in the batch are mixed together.

 The new system with two batches removes this ambiguity. Payloads / batches are comprised purely of mrf messages or purely of non-mrf messages. This way we know that batches with mrf messages should be sent to all destinations, and batches with non-mrf messages should only be sent to non-mrf destinations.

Diagram:
OLD SYSTEM
```
PAYLOAD: {
  {
    "message": "Lorem ipsum"
    "isMRF": no
  },
  {
    "message": "dolor sit amet"
    "isMRF": yes
  },
  {
    "message": "consectetur adipiscing"
    "isMRF": no
  },
}
```
Which destinations do you send this to since it has both MRF and non-MRF messages?

NEW SYSTEM:
```
NON-MRF PAYLOAD: {
  {
    "message": "Lorem ipsum"
    "isMRF": no
  },
  {
    "message": "consectetur adipiscing"
    "isMRF": no
  },
}
MRF PAYLOAD: {
  {
    "message": "dolor sit amet"
    "isMRF": yes
  },
}
```
In this system, the non-mrf payload doesn't get sent to mrf destinations, while the mrf payload gets sent to all destinations. There is no more confusion on the worker's part regarding where to send each payload.

soberpeach: The `chooseBatch` function is perhaps poorly named. It's job is to assign incoming messages to a batch.

remeh: Got it, thank you. Indeed I didn't have the good batch/payload/message level in mind.

remeh (approved):
Great work!
Tested successfully with file tailing from different services.

ddrthall (approved):

PR #38334, opened 2025-06-30
feat: processing rule to exclude truncated logs from being sent

lucastemb:
### What does this PR do?
Adds support for a new log processing rule: exclude_truncated. This rule skips processing for truncated log messages.
### Motivation
To improve log processing efficiency and prevent unnecessary parsing of incomplete (truncated) log entries. Described in further detail [here ](https://datadoghq.atlassian.net/browse/AGNTLOG-232).

A new unit test TestExcludeTruncated verifies:
- Non-truncated logs are processed.
- Truncated logs are correctly skipped and tracked.

N/A


estherk15 (approved):

comp/logs/agent/config/processing_rules.go
```diff
@@ -42,20 +43,22 @@ func ValidateProcessingRules(rules []*ProcessingRule) error {
...
 			return fmt.Errorf("type %s is not supported for processing rule `%s`", rule.Type, rule.Name)
 		}
 
-		if rule.Pattern == "" {
-			return fmt.Errorf("no pattern provided for processing rule: %s", rule.Name)
-		}
-		_, err := regexp.Compile(rule.Pattern)
-		if err != nil {
-			return fmt.Errorf("invalid pattern %s for processing rule: %s", rule.Pattern, rule.Name)
+		if rule.Type != ExcludeTruncated {
```
soberpeach: What's the purpose of this conditional?

lucastemb: I figured that since the ExcludeTruncated rule is concerned more with whether or not the message was truncated, and not the actual content itself, it made sense to not validate or check for any regex patterns for that rule 🤔

soberpeach (approved):

pkg/logs/processor/processor.go
```diff
@@ -267,27 +267,38 @@ func (p *Processor) applyRedactingRules(msg *message.Message) bool {
...
+			case config.ExcludeTruncated:
+				if msg.IsTruncated {
 					msg.RecordProcessingRule(rule.Type, rule.Name)
+					return false
 				}
+
 			}
+		} else {
+			// Log a warning for rules with nil regex - this indicates a configuration issue
+			log.Warnf("Processing rule '%s' of type '%s' has nil regex - rule will be skipped", rule.Name, rule.Type)
```
ddrthall: ExcludeTruncated rules will always hit this instead of being processed

pkg/logs/tailers/windowsevent/tailer.go
```diff
@@ -126,7 +126,14 @@ func (t *Tailer) forwardMessages() {
...
 		if len(decodedMessage.GetContent()) > 0 {
-			t.outputChan <- decodedMessage
+			origin := message.NewOrigin(t.source)
+			// Combine tags from multiple sources: parsing extra tags and source config tags
+			tags := append(decodedMessage.ParsingExtra.Tags, t.source.Config.Tags...)
+			origin.SetTags(tags)
+			msg := message.NewMessage(decodedMessage.GetContent(), origin, decodedMessage.Status, decodedMessage.IngestionTimestamp)
+			// Preserve ParsingExtra information from decoder output (including IsTruncated flag)
+			msg.ParsingExtra = decodedMessage.ParsingExtra
+			t.outputChan <- msg
```
clarkb7: does creating a new message and origin preserve the origin/bookmark information added on lines [279-290](https://github.com/DataDog/datadog-agent/pull/38334/files#diff-cb137568af00553554aacd77ed769ad71b774ab57b44a38687377d16ffb1b4dbL279-L293) ?

pkg/logs/internal/framer/framer.go
```diff
@@ -164,6 +164,11 @@ func (fr *Framer) Process(input *message.Message) {
 			// chop off contentLenLimit raw bytes and output them
 			if len(buf) >= contentLenLimit {
 				content, rawDataLen = buf[:contentLenLimit], contentLenLimit
+				// Only mark as truncated if not already marked by line handler
+				// This prevents double truncation detection for file tailers
+				if !input.ParsingExtra.IsTruncated {
+					input.ParsingExtra.IsTruncated = true
+				}
```
clarkb7: I think you can remove the condition. `IsTruncated` is a `bool` type, it doesn't look to be optional (does not differentiate between set and unset).

```suggestion
				input.ParsingExtra.IsTruncated = true
```

clarkb7 (approved):

pkg/logs/tailers/container/tailer.go
```diff
@@ -388,7 +388,10 @@ func (t *Tailer) forwardMessages() {
 			tags = append(tags, t.tagProvider.GetTags()...)
 			origin.SetTags(tags)
 			// XXX(remy): is it OK recreating a message here?
-			t.outputChan <- message.NewMessage(output.GetContent(), origin, output.Status, output.IngestionTimestamp)
+			msg := message.NewMessage(output.GetContent(), origin, output.Status, output.IngestionTimestamp)
+			// Preserve ParsingExtra information from decoder output (including IsTruncated flag)
+			msg.ParsingExtra = output.ParsingExtra
+			t.outputChan <- msg
```
jszwedko: I'm a bit concerned we'll forget to copy this info sometimes where it is needed. I wonder if (per the `XXX` comment) if we even need to create a new message or if we can just reuse it 🤔

pkg/logs/tailers/file/tailer.go
```diff
@@ -383,6 +383,8 @@ func (t *Tailer) forwardMessages() {
 		}
 
 		msg := message.NewMessage(output.GetContent(), origin, output.Status, output.IngestionTimestamp)
+		// Preserve ParsingExtra information from decoder output (including IsTruncated flag)
+		msg.ParsingExtra = output.ParsingExtra
```
jszwedko: Same here, I'm wondering why we need to create a new message and can't modify/reuse the existing one.

pkg/logs/tailers/socket/tailer.go
```diff
@@ -71,7 +71,10 @@ func (t *Tailer) forwardMessages() {
 		if len(output.GetContent()) > 0 {
 			origin := message.NewOrigin(t.source)
 			origin.SetTags(output.ParsingExtra.Tags)
-			t.outputChan <- message.NewMessage(output.GetContent(), origin, output.Status, output.IngestionTimestamp)
+			msg := message.NewMessage(output.GetContent(), origin, output.Status, output.IngestionTimestamp)
+			// Preserve ParsingExtra information from decoder output (including IsTruncated flag)
+			msg.ParsingExtra = output.ParsingExtra
+			t.outputChan <- msg
```
jszwedko: And here :)

pkg/logs/util/windowsevent/message.go
```diff
@@ -48,20 +49,27 @@ func (m *Message) SetContent(content []byte) {
...
+
 	// old behaviour using an unstructured message with raw data
 	if processRawMessage {
 		jsonEvent, err := m.Json()
 		if err != nil {
 			return nil, err
 		}
-		return message.NewMessageWithSource(jsonEvent, message.StatusInfo, source, time.Now().UnixNano()), nil
+		msg := message.NewMessageWithSource(jsonEvent, message.StatusInfo, source, time.Now().UnixNano())
+		msg.ParsingExtra.IsTruncated = isTruncated
```
jszwedko: The fact that we have to copy this `ParsingExtra` in so many places with this change has me wondering if we should elevate the "truncated" attribute to a top-level attribute on `Message` to make it a "first-class" property of messages.

Granted, this is likely to be a more invasive change. I'm just concerned that the current change is brittle and likely to be accidently broken when someone creates a new message in some place and forgets to copy `ParsingExtra`.

pkg/logs/tailers/journald/tailer.go
```diff
@@ -411,7 +413,10 @@ func (t *Tailer) getOrigin(entry *sdjournal.JournalEntry) *message.Origin {
 	applicationName := t.getApplicationName(entry, tags)
 	origin.SetSource(applicationName)
 	origin.SetService(applicationName)
-	origin.SetTags(append(tags, t.tagProvider.GetTags()...))
+
+	// Combine tags from multiple sources: entry tags, tag provider, and global config
+	allTags := append(tags, t.tagProvider.GetTags()...)
+	origin.SetTags(allTags)
```
jszwedko: I think I'm missing why this change is needed 🤔 Could you elaborate?

comp/logs/agent/config/processing_rules.go
```diff
@@ -64,6 +67,9 @@ func ValidateProcessingRules(rules []*ProcessingRule) error {
 // CompileProcessingRules compiles all processing rule regular expressions.
 func CompileProcessingRules(rules []*ProcessingRule) error {
 	for _, rule := range rules {
+		if rule.Type == ExcludeTruncated {
+			continue
+		}
```
jszwedko, line 70: Suggestion for the future: we could consider decoupling the deserialization from rule initialization so that `ProcessingRule` becomes an `interface` where it can be responsible for its own "compilation" to avoid this increasing complex conditional flow based on type. That is `ProcessingRule` would turn into something like:

```
type ProcessingRule interface {
  Compile() error
  Validate() error
  Match(*message.Message) (bool, error)
}
```

There would need to be a different deserialization struct to read the config into that is the union of all of the possible fields.

Again, no need to change this now, just a suggestion for future refactoring :)

comp/logs/agent/config/processing_rules.go
```diff
@@ -42,20 +43,22 @@ func ValidateProcessingRules(rules []*ProcessingRule) error {
...
-			return fmt.Errorf("invalid pattern %s for processing rule: %s", rule.Pattern, rule.Name)
+		if rule.Type != ExcludeTruncated {
+			if rule.Pattern == "" {
+				return fmt.Errorf("no pattern provided for processing rule: %s", rule.Name)
+			}
+			_, err := regexp.Compile(rule.Pattern)
+			if err != nil {
+				return fmt.Errorf("invalid pattern %s for processing rule: %s", rule.Pattern, rule.Name)
+			}
 		}
```
jszwedko: Suggestion: reuse the `switch` statement above like:

```golang
		switch rule.Type {
		case ExcludeAtMatch, IncludeAtMatch, MaskSequences, MultiLine:
            if rule.Pattern == "" {
				return fmt.Errorf("no pattern provided for processing rule: %s", rule.Name)
			}
			_, err := regexp.Compile(rule.Pattern)
			if err != nil {
				return fmt.Errorf("invalid pattern %s for processing rule: %s", rule.Pattern, rule.Name)
			}
		case ExcludeTruncated:
			break
		case "":
			return fmt.Errorf("type must be set for processing rule `%s`", rule.Name)
		default:
			return fmt.Errorf("type %s is not supported for processing rule `%s`", rule.Type, rule.Name)
		}
```

pkg/logs/processor/processor.go
```diff
@@ -151,27 +151,38 @@ func (p *Processor) applyRedactingRules(msg *message.Message) bool {
 
 	rules := append(p.processingRules, msg.Origin.LogSource.Config.ProcessingRules...)
 	for _, rule := range rules {
-		switch rule.Type {
-		case config.ExcludeAtMatch:
-			// if this message matches, we ignore it
-			if rule.Regex.Match(content) {
+		if rule.Regex != nil || (rule.Regex == nil && rule.Type == config.ExcludeTruncated) {
```
jszwedko: Suggestion: move the `nil` regex check into each processing rule `case` that use regexes rather than this more complex conditional at the top. It introduces a little redundancy, but I think results in more readable code.

_or_ do this `nil` regex check during validation and drop processing rules that should have a regex, but don't, so that here you can be user that they do. This would have the advantage of not executing them each time the rules are applied but just excluding them once up front.

sblumenthal (approved):

pkg/logs/internal/framer/framer.go
```diff
@@ -164,6 +164,7 @@ func (fr *Framer) Process(input *message.Message) {
 			// chop off contentLenLimit raw bytes and output them
 			if len(buf) >= contentLenLimit {
 				content, rawDataLen = buf[:contentLenLimit], contentLenLimit
+				input.ParsingExtra.IsTruncated = true
```
jszwedko, line 167: I believe this is correct, though it does have me wondering if we should also be adding the `...TRUNCATED...` text that the tailers add when they truncate. To me, this seems like it would be more consistent.

Also this gives even more credence to my idea in another comment that truncation should be top-level field rather than part of `ParsingExtra`.

pkg/logs/processor/processor_test.go
```diff
@@ -428,3 +428,50 @@ func BenchmarkMaskSequences(b *testing.B) {
...
+
+	// A truncated message should not be processed
+	msg2 := newMessage([]byte("hello"), &source, "")
+	msg2.IsTruncated = true
+	shouldProcess2 := p.applyRedactingRules(msg2)
+	assert.False(shouldProcess2)
+	assert.Equal(int64(1), msg2.Origin.LogSource.ProcessingInfo.GetCount(ruleType+":"+ruleName))
+}
+
+func TestInvalidProcessingRuleWithNilRegex(t *testing.T) {
```
jszwedko: This test feels unrelated to the change: instead it seems related to a recent change to ignore processing rules with invalid regexes. Did you mean to include it here?

jszwedko (commented):
A few comments and suggestion, but in general this appears to be heading in the right direction! 👍 for the added test.

pkg/logs/tailers/journald/tailer.go
lucastemb: I think while I was experimenting with how I would implement isTruncated tagging for journald messages I added that at some point and forgot to remove it 😅

pkg/logs/processor/processor_test.go
lucastemb: I spoke with Ryan and figured I would knock out [AGNTLOG-112](https://datadoghq.atlassian.net/browse/AGNTLOG-112 ) as well since they both dealt with similar areas of the code base and it was a relatively minor change.

jszwedko: Aha gotcha. I realize the fix for that issue is small, but I'd suggest separating these changes into separate PRs to keep discussion on each change focused.

pkg/logs/processor/processor.go
lucastemb: Opted for the 2nd option. Felt like it was cleaner 😸

pkg/logs/processor/processor_test.go
jszwedko: I see you already have [a PR for the other change](https://github.com/DataDog/datadog-agent/pull/38422) so that would just mean backing it out of this PR.

pkg/logs/util/windowsevent/message.go
remeh: +1

It can be a good time to make sure we need to do the copy here (all these `New*Message*` calls in the different implementations) and, necessary or not, to implement a function directly into `Message` to either do a copy or a shallow copy. That would avoid having the different steps of creating a copy scattered in different places, with possible oversight in the future.

If we're uncertain the copy is necessary or we don't want to look into it now, only having the new function(s) in this PR can make sense to make it to the cutoff date, and to have a follow-up jira card to make sure it's looked into as the next step (e.g. during the QA period).

On moving `IsTruncated` to a first-class property, I think it's in the proper place today (`ParsingExtra` contains all decoding information, is truncated, is multi line, etc.), `ParsingExtra` is just a composed struct into `Message` and its values should always be present alongside the message data. But then I agree, there is a need to have it properly embedded with the message all the time, I think the new copy functions will help with this.

pkg/logs/util/windowsevent/message.go
```diff
@@ -48,20 +49,27 @@ func (m *Message) SetContent(content []byte) {
 
 // MapToMessage packages a Map into either an unstructured message.Message or a structured one.
 func MapToMessage(m *Map, source *sources.LogSource, processRawMessage bool) (*message.Message, error) {
+	// Check if the message was truncated by looking for the truncated flag
+	isTruncated := strings.Contains(m.GetMessage(), truncatedFlag)
```
remeh: To make this slightly faster, instead of `strings.Contains` since we know the truncated flag is either at the start or at the end (IIRC), you can check for the message size and only look at the first & last n bytes (where n is the size of truncatedFlag). Contains is most likely O(n) and even if it's not, because of how many logs/s we're processing, it will matter.

Having it implemented directly on `Message` with a solid implementation and a set of unit tests would be even better, WDYT?

pkg/logs/processor/processor.go
```diff
@@ -150,6 +150,7 @@ func (p *Processor) applyRedactingRules(msg *message.Message) bool {
 	// ---------------------------
 
 	rules := append(p.processingRules, msg.Origin.LogSource.Config.ProcessingRules...)
+	// Filter out rules with invalid regex that require it
```
jszwedko: Did you mean to leave this comment in? It reads like a TODO.

pkg/logs/processor/processor.go
```diff
@@ -172,6 +173,12 @@ func (p *Processor) applyRedactingRules(msg *message.Message) bool {
 					msg.RecordProcessingRule(rule.Type, rule.Name)
 				}
 			}
+		case config.ExcludeTruncated:
```
jszwedko, line 175: Something to consider when refactoring: renaming this method since it applies all processing rules not just redaction.

pkg/logs/internal/framer/framer.go
jszwedko: I suggest we follow up, after this is merged, with the truncation happening here to make sure it is handled consistently with the truncation happening in the line handlers. That is: make sure we are accounting for it in the telemetry we added recently and add the `...TRUNCATED...` text to the message.

pkg/logs/message/message.go
```diff
@@ -226,6 +226,13 @@ func NewMessageWithSource(content []byte, status string, source *sources.LogSour
 	return NewMessage(content, NewOrigin(source), status, ingestionTimestamp)
 }
 
+// NewMessageWithSourceWithParsingExtra adds isTruncated to the parsingExtra tag for a new unstructured message with content, status, source and ingestionTimestamp
+func NewMessageWithSourceWithParsingExtra(content []byte, status string, source *sources.LogSource, ingestionTimestamp int64, isTruncated bool) *Message {
+	msg := NewMessage(content, NewOrigin(source), status, ingestionTimestamp)
+	msg.ParsingExtra.IsTruncated = isTruncated
```
jszwedko: Nit: I'd suggest this could call `NewMessageWithSource` here instead to reduce the duplication a bit and ensure if `NewMessageWithSource` is updated, that this function is also updated.

pkg/logs/tailers/container/tailer.go
```diff
@@ -388,7 +388,9 @@ func (t *Tailer) forwardMessages() {
 			tags = append(tags, t.tagProvider.GetTags()...)
 			origin.SetTags(tags)
 			// XXX(remy): is it OK recreating a message here?
-			t.outputChan <- message.NewMessage(output.GetContent(), origin, output.Status, output.IngestionTimestamp)
+			// Preserve ParsingExtra information from decoder output (including IsTruncated flag)
+			msg := message.NewMessageWithParsingExtra(output.GetContent(), origin, output.Status, output.IngestionTimestamp, output.ParsingExtra)
+			t.outputChan <- msg
```
jszwedko, line 393: I'm still wondering if we need to recreate the message here or if we can just use the `output` from the decoder. Did you look into that?

I'm ok with leaving this for a follow-up refactoring too.

pkg/logs/tailers/file/tailer.go
```diff
@@ -382,7 +382,9 @@ func (t *Tailer) forwardMessages() {
 			continue
 		}
 
-		msg := message.NewMessage(output.GetContent(), origin, output.Status, output.IngestionTimestamp)
+		// Preserve ParsingExtra information from decoder output (including IsTruncated flag)
+		msg := message.NewMessageWithParsingExtra(output.GetContent(), origin, output.Status, output.IngestionTimestamp, output.ParsingExtra)
```
jszwedko, line 386: Same here, I'm wondering if we can just modify `output` rather than create a new message.

pkg/logs/tailers/socket/tailer.go
```diff
@@ -71,7 +71,9 @@ func (t *Tailer) forwardMessages() {
 		if len(output.GetContent()) > 0 {
 			origin := message.NewOrigin(t.source)
 			origin.SetTags(output.ParsingExtra.Tags)
-			t.outputChan <- message.NewMessage(output.GetContent(), origin, output.Status, output.IngestionTimestamp)
+			// Preserve ParsingExtra information from decoder output (including IsTruncated flag)
+			msg := message.NewMessageWithParsingExtra(output.GetContent(), origin, output.Status, output.IngestionTimestamp, output.ParsingExtra)
+			t.outputChan <- msg
```
jszwedko, line 76: And here, could we just use `output` rather than creating a new message. We seem to do that in the `windowsevent` tailer (just reuse the decoded message).

jszwedko (approved):
A few other minor comments and some things we could follow-up on, but in general this looks good to me. I tested it locally with the file tailer and observed that it dropped the truncated logs while still incrementing the telemetry.

pkg/logs/message/message.go
lucastemb: That makes sense I'll update that now ✅

pkg/logs/tailers/container/tailer.go
lucastemb: I didn't, but I'll look into it after the merge. I'll take note of it. 🤔

PR #38365, opened 2025-07-01
system probe: use a higher compression level for BTFS

chouquette:
### What does this PR do?

Use a higher compression level (8) instead of the default one (6)
Also, use multiple threads on the CI to compensate for the increased duration incurred by the higher compression level.

### Motivation

Reduce the size of the minimized-btfs.tar.xz file we ship into most our final artifacts

### Additional Notes

Increasing the compression level has no effect on the decompression, so this is not expected to impact customers besides the size reduction

### DIFF
--- a/tasks/system_probe.py
+++ b/tasks/system_probe.py
@@ -1705,7 +1705,7 @@ def generate_minimized_btfs(ctx, source_dir, output_dir, bpf_programs):
         nw.rule(name="minimize_btf", command="bpftool gen min_core_btf $in $out $input_bpf_programs")
         nw.rule(
             name="compress_minimized_btf",
-            command="tar --mtime=@0 -cJf $out -C $tar_working_directory $rel_in && rm $in",
+            command=f"tar cf $out --mtime=@0 -I 'xz -8 -T{os.environ.get('KUBERNETES_CPU_REQUEST', 1)}' -C $tar_working_directory $rel_in && rm $in",
         )
 
         for root, dirs, files in os.walk(source_dir):


chouquette:

chouquette:

chouquette:

tasks/system_probe.py
```diff
@@ -1705,7 +1705,7 @@ def generate_minimized_btfs(ctx, source_dir, output_dir, bpf_programs):
         nw.rule(name="minimize_btf", command="bpftool gen min_core_btf $in $out $input_bpf_programs")
         nw.rule(
             name="compress_minimized_btf",
-            command="tar --mtime=@0 -cJf $out -C $tar_working_directory $rel_in && rm $in",
+            command=f"tar cf $out --mtime=@0 -I 'xz -8 -T{os.environ.get('KUBERNETES_CPU_REQUEST', 1)}' -C $tar_working_directory $rel_in && rm $in",
```
rdesgroppes, line 1708: For `T{os.environ.get('KUBERNETES_CPU_REQUEST', 1)}`, I just found that:
> [...] a special value **0** makes **xz** use as many threads as there are CPU cores on the system.

chouquette: I think we want to avoid relying on things like `nproc` in case we end up using the physical CPU count rather than the number of CPUs available to the runner, but I agree that it would be much cleaner/easier

rdesgroppes: `nproc` should return the right value in containers should CPU _limits_ be set.

chouquette:
The gains are too marginal to bother with this, closing.

PR #38366, opened 2025-07-01
[AGENTRUN-516] Check dead code elimination enabled in agent binaries in deploy pipelines

pgimalac:
### What does this PR do?
Check whether Go binaries have dead code elimination enabled when building in deploy pipelines.
If the optimization is disabled, the check just logs it (with info on why it's disabled), but doesn't fail the build.

### Motivation
Provides a way to know whether the optimization is enabled in the various binaries that we build, automatically at regular interval (nightlies) and manually if anyone wants to.
This can help debug size increases if you're not familiar with running this locally (or on other platforms).

Only running in deploy pipelines as it increases the duration of the build job by a few minutes.

CI.
Lots of manual testing.


tasks/libs/common/go.py
```diff
@@ -53,9 +58,48 @@ def go_build(
...
+        whydeadcoderes = runner.run("whydeadcode", in_stream=CustomReader(res.stderr), warn=True, hide="out")
+        if whydeadcoderes.stdout:
+            raise Exit(
+                f"dead code elimination is disabled by the following call stack (only the first one is guaranteed to be a true positive):\n{whydeadcoderes.stdout}"
+            )
+
+    return res
+
+
+# reading from stdin in invoke is super slow, see https://github.com/pyinvoke/invoke/issues/818
```
pgimalac:
```suggestion
# reading from stdin in invoke is super slow, see https://github.com/pyinvoke/invoke/issues/819
```

tasks/agent.py
```diff
@@ -230,6 +230,7 @@ def build(
             gcflags=gcflags,
             ldflags=ldflags,
             build_tags=build_tags,
+            check_deadcode_in_ci=sys.platform != 'win32',
```
louis-cqrl: Why for this task specifically we are not doing it for windows but the others we do ?

louis-cqrl (approved):

paulcacheux:
What is the effect of this on CI duration ? when running with `-dumpdep` locally the compilation is measurably slower

vickenty (approved):

tasks/agent.py
pgimalac: Because the optimization is not enabled for windows in the core-agent currently
It's using a windows-specific dependency which imports another dependency which disables it

pgimalac:
@paulcacheux I checked for `datadog-agent-7-x64` and it adds a total of 4 minutes (increased from 7min to 11min), due to running that for each binary sequentially...
That's not great...

gjulianm (approved):
LGTM for the ebpf-platform file

julien-lebot (approved):

internal/tools/go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/DataDog/datadog-agent/internal/tools
 go 1.24.0
 
 require (
+	github.com/aarzilli/whydeadcode v0.0.0-20241226171816-ed86f8ea0a6f
```
JSGette, line 6: Out of curiousity: why do we use this instead of, for instance, [this](golang.org/x/tools/cmd/deadcode)?

pgimalac: That's not the same kind of dead code, basically:
- the [deadcode command](https://pkg.go.dev/golang.org/x/tools/cmd/deadcode) is a static analysis tool to find code in your repo that will not be used when building a binary, see this blog post for details https://go.dev/blog/deadcode
- [github.com/aarzilli/whydeadcode](https://github.com/aarzilli/whydeadcode) is about whether a linker optimization is enabled when building a binary. The optimization determines whether the linker can detect that exported methods are not used or not (if not, every exported method of every reachable type is kept in the binary, making it ~30% bigger on average)

Let me know if you want me to share more details !

pgimalac:

pgimalac:

pgimalac:

pgimalac:

julien-lebot (approved):

tasks/process_agent.py
```diff
@@ -82,6 +82,7 @@ def build(
         build_tags=build_tags,
         bin_path=BIN_PATH,
         env=env,
+        check_deadcode_on_deploy=True,
```
KevinFairise2: That feel weird to have a flag to enable an option when another option is met. Can we just have `check_deadcode` and pass it to true on deploy pipelines? That way it can more easily be used locally also

omnibus/config/software/datadog-agent.rb
```diff
@@ -60,6 +60,10 @@
     major_version_arg = "$MAJOR_VERSION"
   end
 
+  if ENV.has_key?("DEPLOY_AGENT") && ENV["DEPLOY_AGENT"] == "true"
+    env["DEPLOY_AGENT"] = "true"
+  end
```
KevinFairise2: I think it would be easier to just add the variable here, and the environment variable should be passed to the build tasks: https://github.com/DataDog/datadog-agent/blob/main/tasks/libs/common/omnibus.py#L23

ichinaski (approved):

daniel-taf (approved):

omnibus/config/software/datadog-agent.rb
JSGette: I second to this, we should not scatter envvariables

omnibus/config/software/datadog-dogstatsd.rb
```diff
@@ -21,6 +21,10 @@
     'PATH' => "#{gopath.to_path}/bin:#{ENV['PATH']}",
   }
 
+  if ENV.has_key?("DEPLOY_AGENT") && ENV["DEPLOY_AGENT"] == "true"
```
JSGette: same as [above](https://github.com/DataDog/datadog-agent/pull/38366/files#r2337076069)

omnibus/config/software/datadog-iot-agent.rb
```diff
@@ -23,6 +23,10 @@
     'PATH' => "#{gopath.to_path}/bin:#{ENV['PATH']}",
   }
 
+  if ENV.has_key?("DEPLOY_AGENT") && ENV["DEPLOY_AGENT"] == "true"
```
JSGette: same as [above](https://github.com/DataDog/datadog-agent/pull/38366/files#r2337076069)

omnibus/config/software/datadog-otel-agent.rb
```diff
@@ -33,6 +33,10 @@
         "CGO_LDFLAGS" => "-Wl,-rpath,#{install_dir}/embedded/lib -L#{install_dir}/embedded/lib"
     }
 
+    if ENV.has_key?("DEPLOY_AGENT") && ENV["DEPLOY_AGENT"] == "true"
```
JSGette: same as [above](https://github.com/DataDog/datadog-agent/pull/38366/files#r2337076069)

tasks/libs/common/go.py
```diff
@@ -37,12 +37,16 @@ def go_build(
     bin_path: str | Path | None = None,
     verbose: bool = False,
     echo: bool = False,
+    check_deadcode_on_deploy: bool = False,
```
JSGette: I think it beats the purpose of having a boolean parameter. First you set it to `True`, but then have to re-check if another condition is met.

I would actually ask to rename this parameter to `check_deadcode` and move `DEPLOY_AGENT` check higher (to `agent.py`).

tasks/libs/common/go.py
```diff
@@ -60,14 +64,45 @@ def go_build(
         cmd += f" -o {bin_path}"
     if gcflags:
         cmd += f" -gcflags=\"{gcflags}\""
+    if check_deadcode:
```
JSGette, line 67: on the one hand it kind of makes sense to check this here as you are processing flags, on the other you check this flag later again. Ideally, I would avoid checking the same flag and just add necessary ldflag under line 77

tasks/libs/common/go.py
```diff
@@ -60,14 +64,45 @@ def go_build(
...
+        runner = Local(ctx)
+        runner.read_chunk_size = 1024 * 1024 * 10
+        _ = runner.read_chunk_size  # please linters
+        runner.input_sleep = 0
+        _ = runner.input_sleep  # please linters
+
+    # -dumpdep is very verbose so we hide that
+    # any unrecognized log line is shown by whydeadcode anyway
+    result = runner.run(cmd, env=env, hide="stderr" if check_deadcode else None)
+    assert result is not None
```
JSGette: `assert` should never be used in production code, it's for debugging or testing purposes only

tasks/libs/common/go.py
```diff
@@ -60,14 +64,45 @@ def go_build(
...
+    if check_deadcode:
```
JSGette, line 76: hmmm... having the same condition trice. It looks like you should actually split this function into 2 to avoid having the same check several times.

tasks/process_agent.py
JSGette: +1

JSGette (dismissed):
- boolean flag should be unambiguous
- `assert` should never be used in production code
- same check should ideally happen once and not trice within the same function

omnibus/config/software/datadog-agent.rb
pgimalac: I'm not sure I understand the comment, `DEPLOY_AGENT` is already passed to the omnibus build task, but then when calling `inv agent.build` (or other invoke tasks), we list explicitly the environment variables that we want to pass, this piece of code adds the variable to the environment given to the invoke task

tasks/libs/common/go.py
pgimalac: Moving it higher will just duplicate the env var check for every call site so I don't think it would be better, but I'll give it a try

tasks/libs/common/go.py
pgimalac: The `ldflags` are added to the command line 70, this can't be moved down

tasks/libs/common/go.py
pgimalac: I disagree with that, `assert` is a shorthand for checking a condition and raising an exception if it is false, which we do everywhere (and I think raising exception is how error handling is usually done in Python)
In particular in this case the condition will always be true so we don't even care about adding a proper error message

Now to be fair this use of `assert` is just to please the type checker, `runner.run` 's return type is `Result | None`, but actually it's only `None` when a special argument is given, so in this case it is always `Result`
I can just use `typing.cast` if you're not a fan of assertions

tasks/libs/common/go.py
pgimalac: I initially wrote it that way to avoid having `runner.run` twice, but I agree splitting out this logic would make the code simpler, I moved the deadcode logic to a separate function 👍

omnibus/config/software/datadog-agent.rb
pgimalac: The environment might actually be automatically propagated to the invoke task, so this could indeed be useless
I tried removing it and I'll see if it works as expected in CI

pgimalac:

pgimalac:

pgimalac:

pgimalac:

pgimalac:

pgimalac:

pgimalac:

pgimalac:

pgimalac:

pgimalac:
/trigger-ci --variable RUN_ALL_BUILDS=true --variable DEPLOY_AGENT=true --variable RUN_UNIT_TESTS=on

pgimalac:
/trigger-ci --variable RUN_ALL_BUILDS=true --variable DEPLOY_AGENT=true --variable RUN_UNIT_TESTS=on

paulcacheux (approved):

tasks/libs/common/go.py
```diff
@@ -78,3 +88,49 @@ def go_build(
...
     return result
+
+
+def _handle_pipe_to_whydeadcode(ctx: Context, name: str, cmd: str, env: dict[str, str] | None = None) -> Result:
+    # use a custom runner to read stderr in bigger chunks as dumpdep output is huge
+    # and invoke is super slow by default when writing to stdout/stderr
+    # https://github.com/pyinvoke/invoke/issues/774
+    runner = Local(ctx)
+    runner.read_chunk_size = 1024 * 1024 * 10
+    _ = runner.read_chunk_size  # please linters
```
chouetz, line 100: Can you explain how this is helping linters?

tasks/libs/common/go.py
```diff
@@ -78,3 +88,49 @@ def go_build(
...
+    _ = runner.input_sleep  # please linters
+
+    # -dumpdep is very verbose so we hide that
+    # any unrecognized log line is shown by whydeadcode anyway
+    result = cast(Result, runner.run(cmd, env=env, hide="stderr"))
+
+    # worst case it's already installed and nothing happens
+    with ctx.cd("internal/tools"):
+        # pass the env to the command so that it can check GOPATH/GOBIN if provided
+        ctx.run("go install github.com/aarzilli/whydeadcode", env=env)
```
chouetz, line 111: Why do you need to install it like that? Should it be part of the tools installed with `inv install-tools`?

tasks/libs/common/go.py
```diff
@@ -78,3 +88,49 @@ def go_build(
...
+    # it returns non-zero if non-expected input is passed, and 0 otherwise, even if dead code elimination is disabled
+    # so we check whether stdout is empty to know if dead code elimination is disabled
+    whydeadcoderes = cast(
+        Result, runner.run("whydeadcode", in_stream=CustomReader(result.stderr), warn=True, hide="out", env=env)
+    )
+    if whydeadcoderes.stdout:
+        arch = platform.machine()
+        osname = sys.platform
+        print(
+            f"dead code elimination is disabled for {name} on {osname} {arch} by the following call stack (only the first one is guaranteed to be a true positive):\n{whydeadcoderes.stdout}"
```
chouetz, line 124: On line 115 you write `we check whether stdout is empty to know if dead code elimination is disabled` and here you say it's disabled in a block where the `stdout` is _not_ empty. Did I misunderstood something? Can you clarify?

tasks/libs/common/go.py
```diff
@@ -78,3 +88,49 @@ def go_build(
...
+        )
+
+    return result
+
+
+# reading from stdin in invoke is super slow, see https://github.com/pyinvoke/invoke/issues/819
+# so we use a custom reader that always reads 10MiB at a time
+class CustomReader(io.StringIO):
```
chouetz: 🥜 beautifying suggestion
```suggestion
class CustomReader(io.StringIO):
    """
    Custom reader to read 10MiB at a time.
    This is a workaround to increase invoke performance at reading from stdin
    See https://github.com/pyinvoke/invoke/issues/819
    """
```

tasks/libs/common/go.py
```diff
@@ -78,3 +88,49 @@ def go_build(
             os.chown(bin_path, int(uid), int(gid))
 
     return result
+
+
+def _handle_pipe_to_whydeadcode(ctx: Context, name: str, cmd: str, env: dict[str, str] | None = None) -> Result:
+    # use a custom runner to read stderr in bigger chunks as dumpdep output is huge
+    # and invoke is super slow by default when writing to stdout/stderr
+    # https://github.com/pyinvoke/invoke/issues/774
```
chouetz: 🥜 beautifying suggestions
```suggestion
def _handle_pipe_to_whydeadcode(ctx: Context, name: str, cmd: str, env: dict[str, str] | None = None) -> Result:
    """
    - Runs `go build` with the `dumpdep` flag in a custom runner. This runner reads big chunks to improve invoke I/O performance, see https://github.com/pyinvoke/invoke/issues/774
    - Calls `whydeadcode` in the same runner using an associated custom reader.
    """
```

tasks/install_tasks.py
```diff
@@ -23,6 +23,7 @@
     'github.com/vektra/mockery/v2',
     'github.com/wadey/gocovmerge',
     'github.com/uber-go/gopatch',
+    'github.com/aarzilli/whydeadcode',
```
chouetz, line 26: I suppose it answers my question above. Though I still want to be sure if we need the safeguard mechanism.

chouetz (approved):
Can you please clarify the wording in the test on `whydeadcoreres.stdout`? Otherwise only minor comments for me, thanks for the contribution!

BaptisteFoy (approved):

tasks/libs/common/go.py
pgimalac: No that's correct, if there is something on stdout then the optimization is disabled (what is printed is the reason it's disabled)

I'm not sure where the confusion is, would you have preferred `we check whether stdout is not empty to know ...` ?
The comment just means that whether the optimization is enabled depends on whether stdout is empty or not (but doesn't mean it's disabled if output is empty)

tasks/libs/common/go.py
pgimalac: `vulture` complains that `read_chunk_size` and `input_sleep` are unused (so that setting the values is useless)
I think it's because of some dynamic typing which happens under the hood, the values are not accessed in the `Local` class but instead from the `Runner` class

tasks/libs/common/go.py
pgimalac: It is installed with `install-tools`, but actually the build jobs aren't configured to get those
Updating every single build job to reuse the installed tools would have been a much bigger change, and only useful for a small number of pipelines (deploy pipelines)

tasks/install_tasks.py
pgimalac: I explained above why I still install it in the task

PR #38369, opened 2025-07-01
[EBPF] gpu: add priority selection mechanism to NVML metrics

gjulianm:
### What does this PR do?

Adds a priority field to NVML metrics, so that we can de-duplicate in case the same metric is provided by different collectors.

### Motivation

The [GPM collector PR](https://github.com/DataDog/datadog-agent/pull/38254) adds metrics that are already being sent by other collectors, but it only works for Hopper and newer architectures. This change allows us to select which metrics to send and support that case.

Unit tests added.

### DIFF
--- a/pkg/collector/corechecks/gpu/gpu.go
+++ b/pkg/collector/corechecks/gpu/gpu.go
@@ -58,9 +58,10 @@ type Check struct {
 }
 
 type checkTelemetry struct {
-	metricsSent     telemetry.Counter
-	collectorErrors telemetry.Counter
-	activeMetrics   telemetry.Gauge
+	metricsSent      telemetry.Counter
+	duplicateMetrics telemetry.Counter
+	collectorErrors  telemetry.Counter
+	activeMetrics    telemetry.Gauge
 }
 
 // Factory creates a new check factory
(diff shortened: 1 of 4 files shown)


pkg/collector/corechecks/gpu/nvidia/collector.go
```diff
@@ -133,3 +135,16 @@ func GetDeviceTagsMapping(deviceCache ddnvml.DeviceCache, tagger tagger.Componen
...
+func RemoveDuplicateMetrics(metrics []Metric) []Metric {
+	metricsByName := make(map[string]Metric)
+
+	for _, metric := range metrics {
+		if existing, ok := metricsByName[metric.Name]; !ok || existing.Priority < metric.Priority {
+			metricsByName[metric.Name] = metric
+		}
+	}
+
+	return slices.Collect(maps.Values(metricsByName))
```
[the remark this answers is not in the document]
gjulianm: Nope, we don't care about order.

pkg/collector/corechecks/gpu/gpu.go
```diff
@@ -58,9 +58,10 @@ type Check struct {
 }
 
 type checkTelemetry struct {
-	metricsSent     telemetry.Counter
-	collectorErrors telemetry.Counter
-	activeMetrics   telemetry.Gauge
+	metricsSent      telemetry.Counter
+	duplicateMetrics telemetry.Counter
```
val06, line 62: why is that interesting to collect?

val06 (approved):
left 1 small comment

pkg/collector/corechecks/gpu/gpu.go
gjulianm: Mostly for consistency, now the number of metrics we get from each collector might not match with the number of metrics the sender receives because of the duplicate removal. that way we can update the dashboard and ensure the numbers match.

PR #38374, opened 2025-07-01
feat: [SVLS-7036] enable function tags for serverless-init

apiarian-datadog:
Function tags are put in the `_dd.tags.function` tag on the trace payload and the root span metadata for the trace. These are then picked up by our backend to be turned into trace tags and host groups.

### DIFF
--- a/pkg/trace/agent/agent.go
+++ b/pkg/trace/agent/agent.go
@@ -11,6 +11,7 @@ import (
 	"reflect"
 	"runtime"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
@@ -55,6 +56,10 @@ const (
 
 	// tagDecisionMaker specifies the sampling decision maker
 	tagDecisionMaker = "_dd.p.dm"
+
+	// tagFunctionTags is a TracerPayload tag containing the FunctionTags for
+	// all of the traces.
+	tagFunctionTags = "_dd.tags.function"
 )
 
 // TraceWriter provides a way to write trace chunks
@@ -339,6 +344,8 @@ func (a *Agent) Process(p *api.Payload) {
 		}
 	}
 
+	a.ensureFunctionTags(p.TracerPayload)
+
 	gitCommitSha, imageTag := version.GetVersionDataFromContainerTags(p.ContainerTags)
 
 	a.discardSpans(p)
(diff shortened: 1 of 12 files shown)


pkg/trace/agent/agent.go
```diff
@@ -453,6 +461,64 @@ func (a *Agent) Process(p *api.Payload) {
...
+func (a *Agent) ensureFunctionTags(tp *pb.TracerPayload) {
+	if a.conf.FunctionTags == nil {
+		return
+	}
+
+	if tp.Tags == nil {
+		tp.Tags = make(map[string]string)
+	}
+
+	tp.Tags[tagFunctionTags] = a.mergeFunctionTags(tp.Tags[tagFunctionTags], a.conf.FunctionTags)
```
purple4reina: I'm guessing this is where you'll change the function tags from a map to a single string? It would be beneficial to do this conversion only once.

apiarian-datadog: we could, but we might have already had a string-shaped tag on the trace payload already. so in `mergeFunctionTags` we parse that if necessary and then merge in the configured function tags. i suppose we could instead do the conversion at the conf level and then replace any existing function tags on the trace payload with the ones in the conf, but that seems wrong.

purple4reina: Why would the trace payload already have function tags set? That seems completely implausible, or am I missing something?

apiarian-datadog: it is currently impossible, but there's a possibility that this would come up in the future and then we'd be surprised by function tags getting wiped out. in the rewrite of this pr i've instead added a debug log if we are clobbering an existing function tags value

pkg/trace/agent/agent.go
```diff
@@ -138,6 +141,10 @@ type SpanModifier interface {
 	ModifySpan(*pb.TraceChunk, *pb.Span)
 }
 
+type TracerPayloadModifier interface {
+	ModifyTracerPayload(*pb.TracerPayload)
```
ajgajg1134: I think we can simplify the name here, the type and params are sufficiently descriptive:
```suggestion
	Modify(*pb.TracerPayload)
```

pkg/trace/agent/agent.go
```diff
@@ -118,6 +118,9 @@ type Agent struct {
 	// subsequent SpanModifier calls.
 	SpanModifier SpanModifier
 
+	// TracerPayloadModifier will be called on all tracer payloads.
```
ajgajg1134: Let's add a note here that by default it has no functionality and that it acts on tracer payloads _before_ they've undergone normal processing

pkg/trace/agent/agent.go
apiarian-datadog: done!

pkg/trace/agent/agent.go
apiarian-datadog: done!

ichinaski (approved):

joepeeples (approved):

jcstorms1 (approved):
LGTM 👍

astuyve (approved):

apiarian-datadog:
confirmed that this works correctly in both gcp sidecars and aas sidecars.

PR #38379, opened 2025-07-01
dyninst/ebpf: set up stack with link register on arm64

ajwerner:
Before this change we weren't using the value from the link register in the stack trace. This is a straight-up bug on arm64. Also, in a sense, all the frames in arm64 are frameless because of [golang/go#74357] so we don't need to do the adjustments we have for x86 in those cases.

[golang/go#74357]: https://github.com/golang/go/issues/74357

Follow-up bug fix related to https://datadoghq.atlassian.net/browse/DEBUG-3887


grantseltzer (approved):
L G T M

PR #38390, opened 2025-07-02
Final updates for release.json and Go modules for 7.67.1 release + preludes

FlorentClarret:


release.json
```diff
@@ -1,13 +1,12 @@
...
     "base_branch": "7.67.x",
-    "current_milestone": "7.67.1",
+    "current_milestone": "7.67.2",
     "dependencies": {
         "INTEGRATIONS_CORE_VERSION": "7.67.0",
         "OMNIBUS_RUBY_VERSION": "7.67.0",
         "JMXFETCH_VERSION": "0.49.7",
         "JMXFETCH_HASH": "470b1a19933efca4a48157b54ce48ab08a6ec6fb51f0b10337003dc428f50676",
         "SECURITY_AGENT_POLICIES_VERSION": "v0.67.0",
-        "MACOS_BUILD_VERSION": "7.67.0",
```
chouquette: Is this change intended? It looks correct but I'm surprised it's introduced at this time

chouquette (approved):

release.json
FlorentClarret: that's because the command on main got changed to drop this key. I'm going to add it again

FlorentClarret: fixed

PR #38405, opened 2025-07-02
Handle image secrets automatically

KevinFairise2:
### What does this PR do?

Some Kubernetes require to pass a secret to pull the images on private ECR. Let's make it automated when running the task locally, depending on the flags passed by the user

### DIFF
--- a/tasks/new_e2e_tests.py
+++ b/tasks/new_e2e_tests.py
@@ -274,6 +274,23 @@ def run(
         env_vars["E2E_PROFILE"] = profile
 
     parsed_params = {}
+
+    # Outside of CI try to automatically configure the secret to pull agent image
+    if not running_in_ci():
+        # Authentication against agent-qa is required for all kubernetes tests, to use the cache
+        parsed_params["ddagent:imagePullPassword"] = ctx.run(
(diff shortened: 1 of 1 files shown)


tasks/new_e2e_tests.py
```diff
@@ -274,6 +274,23 @@ def run(
         env_vars["E2E_PROFILE"] = profile
 
     parsed_params = {}
+
+    # Outside of CI try to automatically configure the secret to pull agent image
```
chouetz, line 278: I'm wondering if [this](https://github.com/DataDog/datadog-agent/blob/main/.gitlab/e2e/e2e.yml#L13-L49) could be done in the script as well so that the management of credentials is centralised

chouetz (approved):

tasks/new_e2e_tests.py
KevinFairise2: Hmm that part is trickier, because it is setup steps that are only relevant in the CI. I think it would make the new-e2e-tests.run task very complicated if we move all that login inside it.
Also it is convenient to retrieve secrets using the `fetch_secrets.sh` helper we have, less convenient to do that in invoke

PR #38406, opened 2025-07-02
usm: tests: Rework findKeyInMap

guyarb:
### What does this PR do?

Reimplements `findKeyInMap` to use a direct lookup of the key instead of iterating the map, and creating the value in the correct size.

### Motivation

The original implementation had 2 major problems - 
* We iterate the map to find a key
* We can corrupt memory, if the value size in the map is not sizeof(any)

### DIFF
--- a/pkg/network/usm/ebpf_ssl_test.go
+++ b/pkg/network/usm/ebpf_ssl_test.go
@@ -152,7 +152,7 @@ func checkPidExistsInMaps(t *testing.T, manager *manager.Manager, maps []*ebpf.M
 		require.NoError(t, err)
 
 		assert.Eventually(t, func() bool {
-			return findKeyInMap[uint64, any](m, key)
+			return findKeyInMap[uint64](m, key)
 		}, 1*time.Second, 100*time.Millisecond)
 		if t.Failed() {
 			t.Logf("pid '%d' not found in the map %q", pid, mapInfo.Name)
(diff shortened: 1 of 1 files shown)


pkg/network/usm/ebpf_ssl_test.go
```diff
@@ -181,17 +181,9 @@ func checkPidNotFoundInMaps(t *testing.T, manager *manager.Manager, maps []*ebpf
...
-
-	for iter.Next(unsafe.Pointer(&key), unsafe.Pointer(&value)) {
-		if key == theKey {
-			return true
-		}
-	}
-	return false
+func findKeyInMap[K comparable](m *ebpf.Map, theKey K) bool {
+	val := make([]byte, m.ValueSize())
+	return m.Lookup(unsafe.Pointer(&theKey), unsafe.Pointer(&val)) == nil
```
rdesgroppes, line 186:
```suggestion
	return m.Lookup(unsafe.Pointer(&theKey), unsafe.Pointer(&val)) != nil
```
(or?)

guyarb: if there is no error, the key exists

amitslavin (approved):

pkg/network/usm/ebpf_ssl_test.go
rdesgroppes: my bad, thanks!

amitslavin (approved):

PR #38408, opened 2025-07-02
force stop services

clarkb7:
### What does this PR do?
Attempt to force stop the Agent services if they are still running after stopping the Agent via SCM

Restart the Agent services if stopping the services or uninstall the Agent fails

Adds new telemetry spans [[example]](https://app.datadoghq.com/apm/trace/6548138380990018463?graphType=waterfall&shouldShowLegend=true&sort=time&spanID=496705734737405280&timeHint=1751328940978.7317): `stop_all_agent_services`, `terminate_service_process`, `start_agent_service`, `restart_agent_services`

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1552
Make remote agent management operations more resilient to unusual service behavior

adds unit tests

### Additional Notes

Tried adding an E2E test that would make uninstall fail by locking `Global\_MSIExecute`, but taking the mutex seemed to have no effect, even after confirming via winobj and process explorer that it was the same mutex.


git-thuerk-done (approved):
looks good

BaptisteFoy (approved):

pkg/fleet/installer/packages/service/windows/impl.go
```diff
@@ -0,0 +1,207 @@
...
+	span.SetTag("pid", fmt.Sprintf("%d", processID))
+
+	// Open the process with termination rights
+	handle, err := w.api.OpenProcess(windows.SYNCHRONIZE|windows.PROCESS_TERMINATE|windows.PROCESS_QUERY_LIMITED_INFORMATION, false, processID)
+	if err != nil {
+		return fmt.Errorf("could not open process %d for service %s: %w", processID, serviceName, err)
+	}
+	defer w.api.CloseHandle(handle) //nolint:errcheck
+
+	// Verify we still have the correct process by checking the service PID again
```
julien-lebot, line 67: **⚠️ warning:** ‏ If the service exits and restarts quickly enough, Windows may reuse the same PID for the new process.
In that case: 
- `OpenProcess` will succeed, but will return a handle to the new process.
- The second call to `GetServiceProcessID` will also return the same PID.

We falsely assume nothing changed and proceed to terminate the process, which is now the new, live service.
This defeats the safety the re-check we're trying to provide.

If the goal is to ensure we're terminating the same process we originally looked up, it would be more robust to:
- Query the process ID from the handle using [GetProcessId(handle)](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getprocessid) and compare it with the original PID.
- Optionally (for maximum safety), compare the process creation time using [GetProcessTimes](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getprocesstimes) to avoid PID reuse ambiguity.

pkg/fleet/installer/packages/service/windows/impl.go
```diff
@@ -0,0 +1,207 @@
...
+		if err != nil {
+			if errors.Is(err, windows.ERROR_SERVICE_DOES_NOT_EXIST) {
+				continue
+			}
+		} else if !running {
+			continue
+		}
+
+		// Service is running or we failed to check, terminate its process
+		err = w.terminateServiceProcess(ctx, serviceName)
```
julien-lebot, line 155: I think there's a similar issue as above here: We check `IsServiceRunning(serviceName)` after a failed attempt to terminate the service process, to determine whether the service is still running.
But since this call queries the current SCM state, it's possible that:
- The original process exited (maybe even successfully).
- The SCM restarted the service quickly.
- The new service instance is now running under a new or even recycled PID.

In this case, our post-termination check could incorrectly conclude that "termination failed and the service is still running", when in fact the original process did terminate, just not the new one we're now seeing.

Instead of relying on the SCM state again, I think a better strategy would be:
- At the start, resolve and open a handle to all the services processes.
- Use that handle to both terminate the process and observe whether it exited (via [WaitForSingleObject](https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-waitforsingleobject) or [GetExitCodeProcess](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getexitcodeprocess)).

This ensures we're acting on the original process identity, not whatever the SCM is now mapping the service name to.
It also prevents PID reuse from confusing us, since holding the handle guarantees the process object isn't recycled.

julien-lebot (approved):
Approving so that we can meet the 7.69 Agent cut-off, we can address my comments in a follow-up.

pkg/fleet/installer/packages/service/windows/impl.go
clarkb7: linking card
https://datadoghq.atlassian.net/browse/WINA-182

PR #38411, opened 2025-07-02
dyninst/ir,irgen: track probe issues explicitly

ajwerner:
Rather than failing on all issues during ir generation, we want to associate
these issues with specific probes. This will allow us to provide reasonable and
structured diagnostics on a per-probe basis and to survive partial failure.

Relates to https://datadoghq.atlassian.net/browse/DEBUG-3886.


pkg/dyninst/irgen/irgen.go
```diff
@@ -98,14 +111,17 @@ func GenerateIR(
 		typeCatalog:         newTypeCatalog(d, ptrSize),
 		object:              objFile,
 		loclistReader:       loclistReader,
+		issues:              issues,
 	}
 	if err := visitDwarf(d.Reader(), v); err != nil {
 		return nil, err
 	}
-	err = v.instantiateAbstractSubprograms()
+	successes, failures, err := v.instantiateAbstractSubprograms()
```
piob-io: Perhaps "probes, issues, err := ..." to be more specific?

pkg/dyninst/irgen/irgen.go
```diff
@@ -118,107 +134,211 @@ func GenerateIR(
...
 	// variable locations from that instantiation. Sort the entries so that the
 	// output is deterministic.
 	inlinedSubprogramsByUnit := iterMapSorted(v.inlinedSubprograms, cmpEntry)
 	for unit, inlinedSubprograms := range inlinedSubprogramsByUnit {
 		for _, inlinedSubprogram := range inlinedSubprograms {
-			abstractSubprogram := v.abstractSubprograms[inlinedSubprogram.abstractOrigin]
-			if abstractSubprogram == nil {
-				// Not interesting inlined instance.
+			abstractSubprogram, ok := v.abstractSubprograms[inlinedSubprogram.abstractOrigin]
+			// Not interesting inlined instance, or already reported an issue.
```
piob-io, line 217: Shall we proceed even if there is an existing issue, if it is a different issue? And only de-dup issues of the same kind?

pkg/dyninst/ir/issue.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+// Issue is an issue that was encountered while processing a probe.
+type Issue struct {
+	Kind    IssueKind
+	Message string
+}
+
+// IsEmpty returns true if the issue is empty.
+func (i Issue) IsEmpty() bool {
```
piob-io: IsEmpty suggests to me that this is a container that can hold any number of things, not 0 vs 1. How about IsNone or IsUnset?

pkg/dyninst/irgen/irgen.go
```diff
@@ -118,107 +134,211 @@ func GenerateIR(
...
 		for _, probeCfg := range abstractSubprogram.probesCfgs {
-			probe, err := v.newProbe(probeCfg, abstractSubprogram.unit, abstractSubprogram.subprogram)
+			probe, issue, err := v.newProbe(
+				probeCfg, abstractSubprogram.unit, abstractSubprogram.subprogram,
+			)
 			if err != nil {
-				return err
+				return nil, nil, err // propagate global error
+			}
+			if issue != (ir.Issue{}) {
```
piob-io: !issue.IsEmpty()

piob-io (approved):
Thank you!

pkg/dyninst/irgen/irgen.go
ajwerner: Done.

pkg/dyninst/irgen/irgen.go
ajwerner: I thought about that, but at the end of the day, we only get to report one error to the upstream API so I preferred short-circuiting and not modeling the possibility of multiple different issues on a probe. I've added some commentary.

pkg/dyninst/ir/issue.go
ajwerner: Went with `IsNone`

pkg/dyninst/irgen/irgen.go
ajwerner: Done.

PR #38420, opened 2025-07-02
[AGENTRUN-525] Make windows fips job depend on deploy windows job

pgimalac:
### What does this PR do?
Make windows fips job depend on deploy windows job

### Motivation
Fix race in CI.
Currently it can happen that the job starts before the artifacts are pushed.

### DIFF
--- a/.gitlab/e2e/e2e.yml
+++ b/.gitlab/e2e/e2e.yml
@@ -289,7 +289,7 @@ new-e2e-windows-fips-compliance-test:
   needs:
     - !reference [.needs_new_e2e_template]
     - qa_agent_fips
-    - windows_msi_and_bosh_zip_x64-a7-fips
+    - deploy_windows_testing-a7-fips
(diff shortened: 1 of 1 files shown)


KevinFairise2 (approved):

.gitlab/e2e/e2e.yml
```diff
@@ -289,7 +289,7 @@ new-e2e-windows-fips-compliance-test:
   needs:
     - !reference [.needs_new_e2e_template]
     - qa_agent_fips
-    - windows_msi_and_bosh_zip_x64-a7-fips
+    - deploy_windows_testing-a7-fips
```
KevinFairise2, line 293: Hmm actually the `windows_msi_bosh_zip_x64-a7-fips` is required because the test should be retrieving the package from gitlab artifacts.

KevinFairise2 (approved):

PR #38422, opened 2025-07-02
feat: graceful handling of nil regex

lucastemb:
### What does this PR do?
Gracefully handles `nil` regex patterns 
### Motivation
Explained in further deatil [AGNTLOG-112](https://datadoghq.atlassian.net/browse/AGNTLOG-112)

Changes were validated using the `TestInvalidProcessingRuleWithNilRegex` unit test. 

N/A
### Additional Notes

[AGNTLOG-112]: https://datadoghq.atlassian.net/browse/AGNTLOG-112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/logs/processor/processor.go
```diff
@@ -267,27 +267,32 @@ func (p *Processor) applyRedactingRules(msg *message.Message) bool {
...
+					originalContent := content
+					content = rule.Regex.ReplaceAll(content, rule.Placeholder)
+					if !bytes.Equal(originalContent, content) {
+						msg.RecordProcessingRule(rule.Type, rule.Name)
+					}
 				}
 			}
+		} else {
+			// Log a warning for rules with nil regex - this indicates a configuration issue
+			log.Warnf("Processing rule '%s' of type '%s' has nil regex - rule will be skipped", rule.Name, rule.Type)
```
ddrthall: Any objections to popping the warning into the agent status output as well?

pkg/logs/processor/processor.go
```diff
@@ -267,27 +267,32 @@ func (p *Processor) applyRedactingRules(msg *message.Message) bool {
...
-			if !rule.Regex.Match(content) {
-				return false
-			}
-			msg.RecordProcessingRule(rule.Type, rule.Name)
-		case config.MaskSequences:
-			if isMatchingLiteralPrefix(rule.Regex, content) {
-				originalContent := content
-				content = rule.Regex.ReplaceAll(content, rule.Placeholder)
-				if !bytes.Equal(originalContent, content) {
+		if rule.Regex != nil {
```
ddrthall: Question: What was the behavior in the agent before this change?

lucastemb: To my knowledge, it would just panic eventually without safeguards against nil regex.

pkg/logs/processor/processor.go
```diff
@@ -217,3 +220,24 @@ func (p *Processor) GetHostname(msg *message.Message) string {
 	}
 	return hname
 }
+
+func validateAndFilterRules(rules []*config.ProcessingRule) []*config.ProcessingRule {
+	var validRules []*config.ProcessingRule
+
+	for _, rule := range rules {
+		// Rules that require regex
+		if rule.Type == config.ExcludeAtMatch ||
```
soberpeach, line 229: Non-blocking, but for the sake of code clarity, it might be worth it to refactor this `if` statement and the inner `if` statement to something like:
```go
ruleRequiresRegex := rule.Type == config.ExcludeAtMatch || rule.Type == config.IncludeAtMatch || rule.Type == config.MaskSequences
ruleRegexIsNil := rule.Regex == nil
if ruleRequiresRegex && ruleRegexIsNil {
    ...
}
```

You could even move the booleans outside the `for` loop so you don't have to allocate a new variable every run.

pkg/logs/processor/processor.go
```diff
@@ -150,6 +152,7 @@ func (p *Processor) applyRedactingRules(msg *message.Message) bool {
 	// ---------------------------
 
 	rules := append(p.processingRules, msg.Origin.LogSource.Config.ProcessingRules...)
+	rules = validateAndFilterRules(rules)
```
jszwedko, line 155: My suggestion would be to do this even earlier: that is validate and the filter the rules at start-up rather than validating when processing each and every message (which will introduce overhead).

pkg/logs/processor/processor.go
soberpeach: https://grugbrain.dev/#grug-on-expression-complexity

PR #38431, opened 2025-07-03
Improve linter speed on pre-commit

KevinFairise2:
### What does this PR do?

Make sure we do not recursively lint or test everything when `only_modified-packages` or `only_impacted_packages` is activated and should provide the complete list of things to test/lint

### Motivation

Make linter faster

### DIFF
--- a/tasks/gotest.py
+++ b/tasks/gotest.py
@@ -883,8 +883,10 @@ def compute_gotestsum_cli_args(modules: list[GoModule], recursive: bool = True):
             if not target_path.startswith('./'):
                 target_path = f"./{target_path}"
             targets.append(target_path)
-
-    packages = ' '.join(f"{t}/..." if not t.endswith("/...") and recursive else t for t in targets)
+    if recursive:
+        packages = ' '.join(f"{t}/..." if not t.endswith("/...") else t for t in targets)
+    else:
+        packages = ' '.join(targets)
     return packages
 
 
--- a/tasks/gotest.py
+++ b/tasks/gotest.py
@@ -389,9 +389,7 @@ def test(
             result_json=result_json,
             test_profiler=test_profiler,
             coverage=coverage,
-            recursive=not (
-                only_modified_packages or only_impacted_packages
-            ),  # Disable recursive tests when only modified packages or only impacted packages is enabled, to avoid testing a package and all its subpackages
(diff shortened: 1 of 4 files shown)


tasks/gotest.py
```diff
@@ -882,7 +886,7 @@ def compute_gotestsum_cli_args(modules: list[GoModule]):
                 target_path = f"./{target_path}"
             targets.append(target_path)
 
-    packages = ' '.join(f"{t}/..." if not t.endswith("/...") else t for t in targets)
+    packages = ' '.join(f"{t}/..." if not t.endswith("/...") and recursive else t for t in targets)
```
Ishirui: nit: Probably more readable to do this:
```suggestion
    if recursive:
        packages = ' '.join(f"{t}/..." if not t.endswith("/...") else t for t in targets)
    else:
        packages = ' '.join(targets)
```

Or even this:
```suggestion
    to_join = (f"{t}/..." if not t.endswith("/...") else t for t in targets) if recursive else targets
    packages = ' '.join(to_join)
```

Ishirui (approved):
Just one small nit, otherwise LGTM !

Ishirui:
Btw @KevinFairise2 the test failures seems to be because we are now trying to run tests targeting `./cmd`, `./comp` and `./pkg` directly instead of their subdirectories that actually constitute valid targets.

The error message is a bit confusing because my result JSON parsing lib (introduced in #38227) does not handle this properly - I'm opening [a PR to fix this](https://github.com/DataDog/datadog-agent/pull/38432) to reduce noise

KevinFairise2:
> Btw @KevinFairise2 the test failures seems to be because we are now trying to run tests targeting `./cmd`, `./comp` and `./pkg` directly instead of their subdirectories that actually constitute valid targets.
> 
> The error message is a bit confusing because my result JSON parsing lib (introduced in #38227) does not handle this properly - I'm opening [a PR to fix this](https://github.com/DataDog/datadog-agent/pull/38432) to reduce noise

Indeed, that's an oversight, I expected to be able to have the improvement for the `--only-impacted-packages` flag, but that's not the case because we have some logic that clean the number of packages we tests, relying on that recursive test execution. Fixing that

tasks/gotest.py
KevinFairise2: Used the first one!

PR #38434, opened 2025-07-03
Exclude 'system.disk.used' and 'system.disk.utilized' from value comparison in windows e2e tests

jose-manuel-almaza:
### What does this PR do?
Exclude 'system.disk.used' and 'system.disk.utilized' from value comparison in windows e2e tests

### Motivation
Even after increasing the margin of error to 5%, errors are still observed.


test/new-e2e/tests/agent-runtimes/checks/disk/disk_win_test.go
```diff
@@ -25,6 +25,8 @@ func TestWindowsDiskSuite(t *testing.T) {
 			descriptor:            e2eos.WindowsDefault,
 			metricCompareFraction: 0.05,
 			metricCompareDecimals: 1,
```
pgimalac: I guess we could lower back the comparison threshold ?

pgimalac (approved):

louis-cqrl (approved):

PR #29984, opened 2024-10-09
[CONTP-390] Allow user to set the cardinality of a check in the check configuration

zhuminyi:
### What does this PR do?
Add check_tag_cardinality option in the check configuration. And if the user sets it, it overrides the value provided in the agent configuration.
See difference below

The cardinality is passed into this [function](https://github.com/DataDog/datadog-agent/pull/29984/files#diff-12faee531601d8cc7d46d720be825018e508b3d116d25a7051cb4daccf95b9f2R99) in configresolver
```
	if tpl.CheckTagCardinality != "" {
		tags, err = svc.GetTagsWithCardinality(tpl.CheckTagCardinality)
	} 
```

### Motivation
This will allow user to configure the cardinality of check instance (configured with a Pod annotation) directly inside the annotation check config. 

### Describe how to test/QA your changes
deploy dummy app
```
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: redis
  namespace: workload-redis-test
  labels:
    app: redis
spec:
  replicas: 1
  selector:
    matchLabels:
      app: redis
  template:
    metadata:
      labels:
        app: redis
        service: redis
        team: container-integrations
        chart_name: redis_chart
      annotations:
        ad.datadoghq.com/redis.checks: |
          {
            "redisdb": {
              "init_config": {},
              "check_tag_cardinality": "low",
              "instances": [
                {
                  "host": "%%host%%",
                  "port": "6379"
                }
              ]
            }
          }
        ad.datadoghq.com/redis.logs: '[{"type": "docker","image": "redis","service": "redis","source": "redis"}]'
        ad.datadoghq.com/tags: '{redis_port: "REDIS_VERSION_%%env_REDIS_VERSION%%"}'
    spec:
      containers:
        - name: redis
          image: redis:latest
          resources:
            limits:
              cpu: 2000m
              memory: 64Mi
            requests:
              cpu: 1000m
              memory: 64Mi
---
apiVersion: v1
kind: Service
metadata:
  name: redis
  namespace: workload-redis-test
  labels:
    app: redis
spec:
  selector:
    app: redis
  ports:
    - name: redis
      port: 6379
      targetPort: 6379
```
Compare config
"check_tag_cardinality": "low" vs
"check_tag_cardinality": "high",
Screenshot 2024-10-10 at 4 23 13 PM

Container name is high cardinality tag. 

Also run `agent configcheck | grep redis -C 15`

high
```
=== redisdb check ===
Configuration provider: kubernetes-container-allinone
[...]

PR #38440, opened 2025-07-03
 [ AGNTLOG-210 ] Extend the logs agent to support a cardinality level config for the tagger 

ddrthall:
### What does this PR do?
Add support for [tag cardinality](https://datadoghq.atlassian.net/wiki/spaces/AML/pages/3492741848/DD-appended+tags#Tag-Cardinality) configuration in the Logs Agent to mirror the same functionality present in Dogstatsd. This new configuration option will be named DD_LOGS_TAG_CARDINALITY to mirror the checks/dogstatsd equivalents.

In addition to automated testing expansion, all three levels ("high", "low", and "orchestrator") were tested in a kubernetes environment to confirm the appropriate set of tags were applied to each log.

### DIFF
--- a/pkg/config/config_template.yaml
+++ b/pkg/config/config_template.yaml
@@ -206,7 +206,7 @@ api_key:
 ## @env DD_LOGS_TAG_CARDINALITY - string - optional - default: high
 ## Configure the level of granularity of tags to send for logs. Choices are:
 ##   * low: add tags about low-cardinality objects (clusters, hosts, deployments, container images, ...)
-##   * orchestrator: add tags about pod, (in Kubernetes), or task (in ECS or Mesos) -level of cardinality
+##   * orchestrator: add tags about pod (in Kubernetes), or task (in ECS or Mesos) level of cardinality
 ##   * high: add tags about high-cardinality objects (individual containers, user IDs in requests, ...)
 ##
 ## WARNING: sending container tags for logs may create larger payloads.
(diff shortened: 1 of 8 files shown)


pkg/config/config_template.yaml
```diff
@@ -202,6 +202,18 @@ api_key:
 #
 # dogstatsd_tag_cardinality: low
 
+## @param logs_tag_cardinality - string - optional - default: high
+## @env DD_LOGS_TAG_CARDINALITY - string - optional - default: high
+## Configure the level of granularity of tags to send for logs. Choices are:
+##   * low: add tags about low-cardinality objects (clusters, hosts, deployments, container images, ...)
+##   * orchestrator: add tags about pod, (in Kubernetes), or task (in ECS or Mesos) -level of cardinality
```
rtrieu:
```suggestion
##   * orchestrator: add tags about pod (in Kubernetes), or task (in ECS or Mesos) level of cardinality
```

rtrieu (approved):
A minor suggestion but otherwise looks good.

rahulkaukuntla (approved):

gabedos (approved):
Support for a LogCardinality in the tagger looks good to me.

Curious if your team has taken into consideration the `CheckTagCardinality` option that is attached to checks that are autodiscovered. These checks can emit both metrics and logs. This option also allows different checks to have different levels of cardinality. https://github.com/DataDog/datadog-agent/pull/29984. Traditionally it's only been relevant to metrics since we didn't have logs supporting a configurable cardinality. Backward compatibility is likely tricky here since the default values for `CheckTagCardinality` and `LogCardinality` are different. Just something to keep in mind in the future if we want to provide the same experience for log cardinality configuration as metric cardinality configuration

PR #38444, opened 2025-07-03
[NDM] Add first iteration of NCM component SSH implementation

zoedt:
### What does this PR do?

This PR adds the SSH implementation for the NCM (Network Configuration Management) component for NDM. This would be utilized by an agent check and later other aspects of the agent such as the SNMP traps listener - when a trap comes in that signals the configuration on a network device has been recently changed, it will callback to the component methods to retrieve the configuration from said device.

The changes made in this PR are adding implementation for the above functionality. It is not currently actively used by any agent check or listener at the moment (upcoming PR will cover this)

**Relevant documents:**
* [Product brief](https://docs.google.com/document/d/1oj1lKYQWDxGmuL0gQDbfDt8b_ju_JZdKJY-77wVTn8M/edit?tab=t.0#heading=h.ti8ft6mjgkfe)
* [High level RFC of Network Configuration Management](https://datadoghq.atlassian.net/wiki/spaces/II/pages/5182783901/RFC+s+Network+Configuration+Management?focusedCommentId=5187731849)
* [Sub-RFC focused on the trigger-based pulling method vs. agent check scheduled polling](https://datadoghq.atlassian.net/wiki/spaces/II/pages/5249992212/Sub-RFC+Trigger-based+pulls)

**Additional work to fully flesh out the SSH implementation include:**
* Supporting using SSH keys + enable
* Production support for host key callback to validate host keys
* Iterations with other commands, configurable retries, etc.


comp/networkdeviceconfig/impl/networkdeviceconfig.go
```diff
@@ -27,24 +30,72 @@ type Provides struct {
...
-	config config.Component
-	log    log.Component
+	config        *ProcessedNcmConfig
+	log           log.Component
+	clientFactory RemoteClientFactory
 }
 
 // NewComponent creates a new networkdeviceconfig component
 func NewComponent(reqs Requires) (Provides, error) {
+	ncmConfig := newConfig(reqs.Config)
```
GustavoCaso: You are returning `nil` if there is an error when parsing the config. This will panic on line 54. 

I think instead of returning nil if something fails from `newConfig`, I would return `( *ProcessedNcmConfig, error)` and in the component would check for the error, and if there is some, I would `return Provides{}, err`

That way we guarantee that when calling `RetrieveConfiguration`, things will not panic 😄

zoedt: thanks for catching! i'll also add the unit test (probably in another commit after i come back from PTO) in the case the parsing fails / deal with the error gracefully as you mentioned!

comp/networkdeviceconfig/impl/networkdeviceconfig.go
```diff
@@ -27,24 +30,75 @@ type Provides struct {
...
+
+	for _, cmd := range commands {
+		session, err := client.NewSession()
+		if err != nil {
+			return "", n.log.Errorf("Failed to create session: %s", err)
+		}
+		n.log.Debugf("Running command: %s\n", cmd)
+		output, err := session.CombinedOutput(cmd)
+		if err != nil {
+			return "", n.log.Errorf("Command failed: %s\n", err)
```
NouhaManai96:
```suggestion
			return "", n.log.Errorf("Command %s on device %s failed: %s\n", cmd, ipAddress, err)
```

comp/networkdeviceconfig/impl/networkdeviceconfig.go
```diff
@@ -27,24 +30,75 @@ type Provides struct {
...
 	}
 	return provides, nil
 }
 
-// RetrieveConfiguration retrieves the configuration for a given network device ID
-func (n networkDeviceConfigImpl) RetrieveConfiguration(_ string) (string, error) {
-	//TODO implement me
-	return "", nil
+// RetrieveConfiguration retrieves the configuration for a given network device IP
+func (n networkDeviceConfigImpl) RetrieveConfiguration(ipAddress string) (string, error) {
```
NouhaManai96: Not sure what would be the best option, to have two separate functions `RetrieveRunningConfiguration` & `RetrieveStartUpConfiguration` or have the commands list as an argument to this function and have a list for running config commands and another one for startup config commands so that we can distinguish between both ?
```suggestion
func (n networkDeviceConfigImpl) RetrieveRunningConfiguration(ipAddress string) (string, error) {

OR (wherever this is called)

runningConfig, err := RetrieveConfiguration(ipAddress, runningConfigCommands)
startupConfig, err := RetrieveConfiguration(ipAddress, startupConfigCommands)

?
```

comp/networkdeviceconfig/impl/networkdeviceconfig.go
```diff
@@ -27,24 +30,75 @@ type Provides struct {
...
+	for _, cmd := range commands {
+		session, err := client.NewSession()
+		if err != nil {
+			return "", n.log.Errorf("Failed to create session: %s", err)
+		}
+		n.log.Debugf("Running command: %s\n", cmd)
+		output, err := session.CombinedOutput(cmd)
+		if err != nil {
+			return "", n.log.Errorf("Command failed: %s\n", err)
+		}
```
NouhaManai96:
```suggestion
		if err != nil {
		    session.Close()
			return "", n.log.Errorf("Command failed: %s\n", err)
		}
```

comp/networkdeviceconfig/impl/config.go
```diff
@@ -0,0 +1,58 @@
...
+
+// DeviceConfig holds the info to connect to a network device, including its IP address and authentication credentials.
+type DeviceConfig struct {
+	IPAddress string          `mapstructure:"ip_address"` // ip address of the network device, e.g., "10.0.0.1"
+	Auth      AuthCredentials `mapstructure:"auth"`
+}
+
+// RawNcmConfig is the raw config structure for Network Config Management (NCM) taken from the Agent configuration
+type RawNcmConfig struct {
+	Namespace string         `mapstructure:"namespace"` // namespace for the network config management, e.g., "default"
```
NouhaManai96, line 34: do we expect this to be different than the device namespace ? and if this is the case then how to correlate both ? thinking of cases where we have two devices with the same IP but different namespaces, how to attach the config to the appropriate device ?

comp/networkdeviceconfig/impl/config.go
```diff
@@ -0,0 +1,58 @@
...
+
+// ProcessedNcmConfig is the processed config structure for Network Config Management (NCM) to be used by the component
+type ProcessedNcmConfig struct {
+	Namespace string
+	Devices   map[string]DeviceConfig // map of device IP addresses to DeviceConfig
+}
+
+func newConfig(agentConfig config.Component) (*ProcessedNcmConfig, error) {
+	ncm := &RawNcmConfig{}
+	err := structure.UnmarshalKey(agentConfig, "network_config_management", &ncm)
```
NouhaManai96: ❓ should this be under `network_devices` (similarly to `snmp-traps` or `netflow`)?
🥜 to avoid any confusion
```suggestion
	err := structure.UnmarshalKey(agentConfig, "network_devices_config", &ncm)
```

NouhaManai96 (approved):
looks great overall 🎉 
I left a couple comments but nothing blocking

comp/networkdeviceconfig/impl/config.go
zoedt: i think it should match the device namespace; i added it to prevent the case if there were two IP addresses but two different namespaces so we could match easier 🤔 

i was assuming if there's the same ip address, another config would be added to monitor that device with the different namespace and the same would happen here

wasn't sure if each device should have its own namespace instead of directly on the config?

comp/networkdeviceconfig/impl/networkdeviceconfig.go
zoedt: good call out esp w/ the thread re: startup/running config yesterday !

i agree something general and then being able to pass the command to grab it would be useful 👍 

maybe even a version where you can get the running and startup config in the same go if they wanted to, lemme fix something up here

zoedt: for now, i went with separating and keeping the commands internal to that method and making the commands private to the package

it'll probably change a lot as we think about it a bit more; something will be added to those specific methods for determining which commands get the specific "type" of config in the future

but i figure to obscure some of the commands for now to keep it simple

comp/networkdeviceconfig/impl/config.go
zoedt: good question, i wasn't sure where all this config should really live in the end 🤔 

i was thinking the `snmp-traps` and `netflow` configs are pretty small overall in comparison to how big this could possibly get for some customers (more closer to the snmp integration i'd say given the number of devices to define) and could get unwieldy in `datadog.yaml` under `network_devices` or even its own section as i wrote here

(even tho the `snmp-traps` lets you provide a long list of credentials, it's probably smaller than say, the entire number of devices they have in their environment)

i think i'll write something up about the flow since it's been on my mind especially with introduction of an agent check, maybe we can set aside time to brainstorm it out fully as well

comp/networkdeviceconfig/impl/networkdeviceconfig.go
```diff
@@ -27,24 +30,89 @@ type Provides struct {
...
+		if err != nil {
+			return "", n.log.Errorf("Failed to create session to device %s: %s", ipAddress, err)
+		}
+		n.log.Debugf("Running command: %s\n", cmd)
+		output, err := session.CombinedOutput(cmd)
+		if err != nil {
+			session.Close()
+			return "", n.log.Errorf("Command %s on device %s failed: %s\n", cmd, ipAddress, err)
+		}
+		n.log.Debugf("Output: %s\n", output)
```
NouhaManai96: I think we should remove this for now and add it later when we have the config parsing/scrubbing since this output might have credentials/certifcats/etc

zoedt: oops good call out i forgot to remove this after my debugging 🤦 ty for catching!

PR #38445, opened 2025-07-03
[ AGNTLOG-246 ] Force http if a user specifies a logs_dd_url prefixed with http/https

ddrthall:
### What does this PR do?
The logs_config.logs_dd_url configuration option allows users to specify proxies with an http(s):// prefix. However, if the proxy is unavailable at agent start up, the agent will automatically fallback to TCP connectivity for the duration of its runtime. This is obviously invalid for addresses with the http(s) prefix. This PR adjusts behavior to prevent fallback to TCP if HTTP(S) is explicitly requested.

### DIFF
--- a/comp/logs/agent/config/config.go
+++ b/comp/logs/agent/config/config.go
@@ -118,7 +118,13 @@ func BuildEndpointsWithConfig(coreConfig pkgconfigmodel.Reader, logsConfig *Logs
 	}
 
 	mrfEnabled := coreConfig.GetBool("multi_region_failover.enabled")
-	if logsConfig.isForceHTTPUse() || logsConfig.obsPipelineWorkerEnabled() || mrfEnabled || (bool(httpConnectivity) && !(logsConfig.isForceTCPUse() || logsConfig.isSocks5ProxySet() || logsConfig.hasAdditionalEndpoints())) {
+
+	// logs_config.logs_dd_url might specify a HTTP(S) proxy. Never fall back to TCP in this case.
+	haveHTTPProxy := false
+	if logsDDURL, defined := logsConfig.logsDDURL(); defined {
+		haveHTTPProxy = strings.HasPrefix(logsDDURL, "http://") || strings.HasPrefix(logsDDURL, "https://")
+	}
+	if logsConfig.isForceHTTPUse() || haveHTTPProxy || logsConfig.obsPipelineWorkerEnabled() || mrfEnabled || (bool(httpConnectivity) && !(logsConfig.isForceTCPUse() || logsConfig.isSocks5ProxySet() || logsConfig.hasAdditionalEndpoints())) {
 		return BuildHTTPEndpointsWithConfig(coreConfig, logsConfig, endpointPrefix, intakeTrackType, intakeProtocol, intakeOrigin)
 	}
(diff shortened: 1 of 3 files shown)


comp/logs/agent/config/config.go
```diff
@@ -118,7 +118,13 @@ func BuildEndpointsWithConfig(coreConfig pkgconfigmodel.Reader, logsConfig *Logs
 	}
 
 	mrfEnabled := coreConfig.GetBool("multi_region_failover.enabled")
-	if logsConfig.isForceHTTPUse() || logsConfig.obsPipelineWorkerEnabled() || mrfEnabled || (bool(httpConnectivity) && !(logsConfig.isForceTCPUse() || logsConfig.isSocks5ProxySet() || logsConfig.hasAdditionalEndpoints())) {
+
+	// logs_config.logs_dd_url might specify a HTTP(S) proxy. Never fall back to TCP in this case.
+	haveHTTPProxy := false
+	if logsDDURL, defined := logsConfig.logsDDURL(); defined {
+		haveHTTPProxy = strings.HasPrefix(logsDDURL, "http://") || strings.HasPrefix(logsDDURL, "https://")
+	}
```
jszwedko, line 126: Is it necessarily that a proxy has been configured as the endpoint? I'm thinking `logs_dd_url` could be set to some Datadog Logs API-compatible endpoint (like Vector).

releasenotes/notes/logs-dd-url-no-fallback-8e2f0537584c83a0.yaml
```diff
@@ -0,0 +1,11 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+enhancements:
+  - |
+    No longer have the Logs Agent fall back to TCP when configuring logs_config.logs_dd_url with a http(s):// prefix.
```
jszwedko: 👍

comp/logs/agent/config/config.go
ddrthall: Vector specifically has a different setting (`obsPipelineWorkerEnabled`/`getObsPipelineURL`) that is also verified in the same if statement that `haveHTTPProxy` is in. Other code commentary refers to logs_dd_url as intended to be a "proxy", but it's just a url and therefore can actually be anything.

releasenotes/notes/logs-dd-url-no-fallback-8e2f0537584c83a0.yaml
rtrieu:
```suggestion
    No longer have the Logs Agent fall back to TCP when configuring `logs_config.logs_dd_url` with a http(s):// prefix.
```

rtrieu (approved):
Left a tiny suggestion.

comp/logs/agent/config/config.go
jszwedko: Right, yeah, people sometimes also use `logs_dd_url` with Vector too, but it sounds like we usually expect people to use it with reverse proxies.

jszwedko (approved):

PR #38464, opened 2025-07-04
Update lint command to benefits from golanci-lint concurrency

KevinFairise2:
### What does this PR do?

Use a single golangci-lint call with parallel runners to speed up the linting operation.
It reduces linter job duration:
- lint_window-x64: 64min -> 25 min
- lint_linux_arm64: 22min  -> 6 min
- lint_linux_x64: 18min -> 6 min
- lint_macos_gitlab_amd64: 22min -> 4 min

### Motivation

Make linter faster in the CI, and pre-coomit as well. Which nicely improved my local experience


CelianR (approved):
Amazing! 👏

PR #38467, opened 2025-07-04
[NDMII-3540] Collect metrics and options data for Cisco IPsec VPN tunnels

Pierre-L42:
### What does this PR do?

This PR adds collection of metrics and options data for Cisco IPsec VPN tunnels in SNMP.

Installed the Agent on the AWS setup, the metrics can be seen here: [https://dd.datad0g.com/dashboard/tej-uaw-8qe/cisco-ipsec-metrics?fromUser=false&refr[…]e=sliding&from_ts=1752208734720&to_ts=1752223134720&live=true](https://dd.datad0g.com/dashboard/tej-uaw-8qe/cisco-ipsec-metrics?fromUser=false&refresh_mode=sliding&from_ts=1752208734720&to_ts=1752223134720&live=true) and also on the devices with namespace `vpn-s2s`. The new metadata are present when running the `datadog-agent check snmp` command.


releasenotes/notes/ndm-collect-vpn-tunnels-metrics-and-options-7685e040e9580dc1.yaml
```diff
@@ -0,0 +1,11 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+features:
+  - |
+    Adds Cisco IPsec VPN tunnels metrics and options data collection to SNMP. This can be enabled/disabled using the `collect_vpn` config.
```
buraizu:
```suggestion
    Adds Cisco IPsec VPN tunnel metrics and options data collection to SNMP. This can be enabled and disabled using the `collect_vpn` config.
```

buraizu (approved):
Approving with a couple of changes requested

pkg/collector/corechecks/snmp/internal/checkconfig/buildprofile.go
```diff
@@ -71,6 +71,7 @@ func (c *CheckConfig) BuildProfile(sysObjectID string) (profiledefinition.Profil
 		profile.Device.Vendor = rootProfile.Device.Vendor
 	}
 	profile.Metadata = updateMetadataDefinitionWithDefaults(profile.Metadata, c.CollectTopology, c.CollectVPN)
+	profile.Metrics = updateMetricsDefinitionWithDefaults(profile.Metrics, c.CollectVPN)
```
FlorianVeaux, line 74: Both for Metadata and Metrics, this means we'll try to fetch the VPN data on all devices even when it might not be available?
Could we figure out a way to _only_ fetch all these metrics and metadata if we know they are available?

Pierre-L42: Yes that's right, we try to fetch VPN metadata and metrics on all devices if the `collect_vpn` flag is true, this is also what we do for CDP metadata..

I guess we could first try to narrow down by vendor: only fetch Cisco OIDs if the vendor is Cisco
We could also, for every MIB table, try a single OID for this MIB table, if it doesnt exists, then we don't need to try the other OIDs in the same MIB table

But I think this needs more discussion

FlorianVeaux: Comment resolved: this is solved in another PR by storing which OIDs don't have data

FlorianVeaux (approved):

PR #38470, opened 2025-07-04
Allow log files to be opened via system-probe

vitkyrka:
### What does this PR do?

Add a non-default privileged logs system-probe module which allows the core agent to ask it to open log files and send it the open file descriptor via Unix sockets.

See https://datadoghq.atlassian.net/wiki/spaces/DSCVR/pages/5414060329/ and the documents linked from there for more information.

### Motivation

https://datadoghq.atlassian.net/wiki/spaces/DSCVR/pages/5414060329/
https://datadoghq.atlassian.net/browse/DSCVR-186


cmd/system-probe/modules/fdtransfer_linux.go
```diff
@@ -0,0 +1,249 @@
...
+		}
+		responseBytes, _ := json.Marshal(response)
+		if _, _, err := unixConn.WriteMsgUnix(responseBytes, nil, nil); err != nil {
+			log.Errorf("Failed to write error response: %v", err)
+		}
+		return
+	}
+
+	// Open the resolved file with O_NOFOLLOW to prevent symlink attacks
+	file, err := os.OpenFile(resolvedPath, os.O_RDONLY|syscall.O_NOFOLLOW, 0)
```
smeso-ddg: O_NOFOLLOW prevents symlinks resolution in the last part of the file. But it can still happen in the rest of the path. We need a comment regarding why we think that this isn't a problem (e.g. because the only check now is the file name check and we don't check anything at all regarding the path) or address the issue so that even future changes won't cause vulns

pkg/system-probe/api/client/README_fdtransfer.md
```diff
@@ -0,0 +1,100 @@
...
+data := make([]byte, 1024)
+n, err := file.Read(data)
+if err != nil {
+    log.Errorf("Failed to read from file: %v", err)
+    return
+}
+log.Infof("Read %d bytes: %s", n, string(data[:n]))
+```
+
+## Security Considerations
```
smeso-ddg: We need a detailed description of a threat model and all our countermeasures.
Reading the doc it looks like this interface can be used to open any file in the system, but this is not the case.
The description should include things like:

1. who can use this interface (can an attacker outside the agent use it?)
2. what can be accessed and how
3. how we handle symlinks at the end, symlinks along the path, realtive paths, paths that pass thorough sticky directories, and any other unusual scenario

smeso-ddg (commented):
Thanks! Great work

pkg/config/setup/system_probe.go
```diff
@@ -365,6 +365,9 @@ func InitSystemProbeConfig(cfg pkgconfigmodel.Setup) {
 	cfg.BindEnvAndSetDefault(join(discoveryNS, "network_stats.period"), "60s")
 	cfg.BindEnvAndSetDefault(join(discoveryNS, "ignored_command_names"), []string{"chronyd", "cilium-agent", "containerd", "dhclient", "dockerd", "kubelet", "livenessprobe", "local-volume-pr", "sshd", "systemd"})
 
+	// Privileged Logs config
+	cfg.BindEnvAndSetDefault(join("privileged_logs", "enabled"), false)
```
gjulianm: nit: Can we have the `privileged_logs` string defined in a constant, for consistency with the other settings?

pkg/privileged-logs/client/open.go
```diff
@@ -0,0 +1,137 @@
...
+				}
+			}
+		}
+	}
+
+	return nil, fmt.Errorf("no file descriptor received")
+}
+
+// Open opens a file with in case of permission errors, it opens the file with
+// system-probe.
```
gjulianm: I think this comment needs to be edited correctly :D

gjulianm (approved):
LGTM, left two minor nits

L3n41c (approved):
LGTM for the single file owned by @DataDog/container-integrations.

ddrthall (approved):

hush-hush (approved):

smeso-ddg (approved):
Thanks!

jose-manuel-almaza (approved):

PR #38475, opened 2025-07-05
[CWS] use event timestamp in backend event

paulcacheux:
### What does this PR do?

This PR passes the event timestamp down to the backend payload. Ensuring the correct timestamp is used and not just `time.Now` at send time which happens after the retry/tag resolution delay.


pkg/security/module/server.go
```diff
@@ -397,12 +400,18 @@ func (a *APIServer) SendEvent(rule *rules.Rule, event events.Event, extTagsCb fu
 			}
 		}
 
+		timestamp := ev.Timestamp
```
YoannGh: Should we call `ev.ResolveEventTime()` instead of using this field directly?

paulcacheux: Hmm I think it works now because this handler is called before, but just by luck. Thanks !

YoannGh (approved):

PR #31968, opened 2024-12-10
[CWS] Allow triggering an activity dump from CLI

lebauce:
### What does this PR do?

Allow triggering an activity dump from the command line by specifying its cgroup path.

PR #38481, opened 2025-07-07
[CWS] cgroup / containers rework

spikat:
### What does this PR do?

As we are no longer able to determine the cgroup runtimer/manager, all logic around that is now removed:
* The fields `cgroup.manager` and `container.runtime` are no longer exposed.
* The configuration `runtime_security_config.activity_dump.cgroup_managers` to specify which cgroup manager to trace deleted
* Now, the only way to differentiate containers from other cgroups is the regexp looking for a container ID withing the cgroup path.
* Introcution of `runtime_security_config.activity_dump.trace_systemd_cgroups` (disabled by default) to defines if we want to trace cgroups other than container ones.
* All mention to containerID/Cgroup flags in eBPF are removed.
* Add a filter mechanism for cgroup v1, to filter cgroup event (tracing and cgroup_write) only on `systemd` cgroup mount id.
* Now resolves the containerID in userspace from the cgroup path.

### Motivation

The cgroup tracing mechanism (used for activity dumps) based on cgroup mount id was broken on cgroup v1 since https://github.com/DataDog/datadog-agent/pull/31968.

Also, with the new containerd version changing how it manage cgroup paths, it's no possible anymore to determine the container runtime.


michaelcretzman (approved):

buraizu (approved):

dustmop (approved):
Approving for agent-configuration

YoannGh (commented):
Can we also get rid of the part where we read `container_qstr` in `pkg/security/ebpf/c/include/hooks/cgroup.h`?

YoannGh (approved):
:rocket:

paulcacheux (approved):

PR #38488, opened 2025-07-07
[DEBUG-4039] dyninst/rcscrape: preliminaries ahead of scraping RC

ajwerner:
This is a pile of commits that are preliminaries ahead of RC scraping:

#### dyninst/irgen: add options for dynamic data limits

This is necessary to increase the limits for rc scraping
programs.

#### dyninst/procmon: move ProcessID and Executable

The dependency on actuator didn't make a ton of sense, and it
made for surprising interfaces.

#### dyninst/actuator: create Tenant concept

There are multiple users of the actuator that might each want to create
different programs on a per process basis and yet we want them all to
share a single set of resources.

This change feels right insofar as we do want to isolate components that
want to load programs and fit into the program lifecycle independently.
A weird property of this change is that for there to be multiple
programs applied to the same process, they *must* be by different
tenants. This is not different than before -- it is more relaxed than
before, but it still feels a bit off.

#### dyninst/actuator: simplify program->proc state

They have a 1:1 mapping, make that explicit.

#### dyninst/loader,ebpf: set prog_id in header

It turns out it's pretty important. Fixes https://datadoghq.atlassian.net/browse/DEBUG-4017


pkg/dyninst/actuator/testdata/snapshot/multiple_tenants.yaml
```diff
@@ -0,0 +1,50 @@
+# Tests compilation failure. The process should move to CompilationFailed state
```
piob-io: Stale comment?

pkg/dyninst/integration_test.go
```diff
@@ -47,8 +49,9 @@ func TestDyninst(t *testing.T) {
 	cfgs := testprogs.MustGetCommonConfigs(t)
 	programs := testprogs.MustGetPrograms(t)
 	for _, svc := range programs {
```
piob-io: Consider changing this into loop over opt-in list.

pkg/dyninst/procmon/process.go
```diff
@@ -0,0 +1,74 @@
...
+	return fmt.Sprintf("{PID:%d,Svc:%s}", p.PID, p.Service)
+}
+
+// FileHandle identifies a file on a device.
+type FileHandle struct {
+	Dev uint64
+	Ino uint64
+}
+
+// FileCookie identifies a file on a device, and the time it was
```
piob-io: FileCookie seems to only identify time, but not the file? Do we need this extra type? How about just having "LastModified syscall.Timespec" field in the FileKey?

pkg/dyninst/rcscrape/debouncer.go
```diff
@@ -0,0 +1,121 @@
...
+	"slices"
+	"time"
+
+	"github.com/DataDog/datadog-agent/pkg/dyninst/actuator"
+	"github.com/DataDog/datadog-agent/pkg/dyninst/ir"
+	"github.com/DataDog/datadog-agent/pkg/dyninst/rcjson"
+	"github.com/DataDog/datadog-agent/pkg/util/log"
+)
+
+type debouncer struct {
```
piob-io: Let's add a short comment on what this provides.

pkg/dyninst/rcscrape/decoder.go
```diff
@@ -0,0 +1,270 @@
...
+	"github.com/DataDog/datadog-agent/pkg/util/log"
+)
+
+type remoteConfigFile struct {
+	RuntimeID     string
+	ConfigPath    string
+	ConfigContent string
+}
+
+// decoder decodes remote config events into remote config files.
```
piob-io: from remote config files? I might be confusing the terms used here.

piob-io (commented):
If it's not too much trouble I'd split the pr between rcscrape package, and everything else. I need to take another review pass at the rcscrape package, sending current comments.

pkg/dyninst/procmon/process.go
ajwerner: Yes, this is all a bit much given we aren't doing anything fancy with executables right now. This is pure code movement -- nothing new here. That being said, done.

pkg/dyninst/actuator/testdata/snapshot/multiple_tenants.yaml
ajwerner: Done.

pkg/dyninst/integration_test.go
ajwerner: Will do in the follow-up PR.

pkg/dyninst/rcscrape/decoder.go
ajwerner: Moving to the follow-up.

pkg/dyninst/rcscrape/debouncer.go
ajwerner: Moving it into the follow-up.

ajwerner (commented):
I split out the preliminaries at your request. I'll put up the core RC scraping impl momentarily.

piob-io (approved):

piob-io:
Thanks!

PR #38490, opened 2025-07-07
util/strings: optional telemetry on strings blocklist

remeh:
### What does this PR do?

Optional counter telemetry on the strings blocklist.

Unit tests + manual spot check on the telemetry endpoint.

### DIFF
--- a/comp/dogstatsd/server/server.go
+++ b/comp/dogstatsd/server/server.go
@@ -176,6 +176,7 @@ type server struct {
 	tlmProcessed            telemetry.Counter
 	tlmProcessedOk          telemetry.SimpleCounter
 	tlmProcessedError       telemetry.SimpleCounter
+	tlmPointsBlocked        telemetry.SimpleCounter
 	tlmChannel              telemetry.Histogram
 	listernersTelemetry     *listeners.TelemetryStore
 	packetsTelemetry        *packets.TelemetryStore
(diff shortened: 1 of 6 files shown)


pgimalac (approved):

pkg/util/strings/blocklist.go
```diff
@@ -75,10 +80,19 @@ func (b *Blocklist) Test(name string) bool {
 	//
 	// Thus j must be i - 1.
 	if b.matchPrefix && i > 0 && strings.HasPrefix(name, b.data[i-1]) {
+		if b.tlmCounter != nil {
+			b.tlmCounter.Inc()
+		}
```
vickenty, line 85: Would it make sense to make dogstatsd increment the counter itself? There are fewer places where it needs to be done (only one) and that requires much less code changes.

comp/dogstatsd/server/server.go
```diff
@@ -523,12 +525,12 @@ func (s *server) SetBlocklist(metricNames []string, matchPrefix bool) {
...
 	// send the complete blocklist to all workers, the listening part of dogstatsd
 	for _, worker := range s.workers {
-		blocklist := utilstrings.NewBlocklist(metricNames, matchPrefix)
+		blocklist := utilstrings.NewBlocklist(metricNames, matchPrefix, s.tlmPointsBlocked)
 		worker.BlocklistUpdate <- blocklist
 	}
 
 	// send the histogram blocklist used right before flushing to the serializer
-	histoBlocklist := utilstrings.NewBlocklist(histoMetricNames, matchPrefix)
+	histoBlocklist := utilstrings.NewBlocklist(histoMetricNames, matchPrefix, s.tlmPointsBlocked)
```
vickenty, line 533: Would it make sense to create a separate counter for this?

Sharing a counter for two use-cases like this is prone to be confusing, since number of blocked histogram aggregates can be larger than the number of points received. And adding to confusion, this is different to how we structure telemetry normally: by component where things happen, rather than by function.

PR #38493, opened 2025-07-07
disable THP by default in system-probe and the security-agent

paulcacheux:
### What does this PR do?

This PR disables THP usage in the system-probe and the security-agent by default. After running a lot of benchmarks and  internal deployments, we showed that the impact of disabling THPs was neutral or positive, especially in the SMP environment.

Internally we deploy the agent on nodes configured with transparent huge pages configed in madvise mode. This means that only pages that have an additional `madvise` call on them with `MADV_HUGEPAGE` will make use of huge pages. The go runtime does not make this call in the general case, meaning that we have deployed the agent internally with THP disabled since basically forever.

In SMP benchmarks, the impact of disabling THP allows the system-probe and the security agent to not exhibit extra memory usage at the end of the 10 min run:
baseline:
image
comparison:
image
in this particular benchmark the maximum PSS for the total container going from 545MiB to 517MiB with THPs disabled.

Re-enabling THP usage is still possible using the `system_probe_config.disable_thp` or `security_agent.disable_thp` configuration options.

### DIFF
--- a/pkg/config/setup/config.go
+++ b/pkg/config/setup/config.go
@@ -954,7 +954,7 @@ func InitConfig(config pkgconfigmodel.Setup) {
 	config.BindEnvAndSetDefault("security_agent.cmd_port", DefaultSecurityAgentCmdPort)
 	config.BindEnvAndSetDefault("security_agent.expvar_port", 5011)
 	config.BindEnvAndSetDefault("security_agent.log_file", DefaultSecurityAgentLogFile)
-	config.BindEnvAndSetDefault("security_agent.disable_thp", false)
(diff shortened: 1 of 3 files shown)


dustmop (approved):
Approved for agent-configuration owned files

usamasaqib:
Can you share details of the benchmarks and results from deployments?

git-thuerk-done (approved):
good for docs

usamasaqib:
Given that this is an SMP environment variability issue, I am not sure if this is the best place to solve this.
cc: @scottopell

usamasaqib (approved):

PR #38495, opened 2025-07-07
[USMON-1424] Fix go tls map leak

amitslavin:
### What does this PR do?

Adding kernel cleanup for the `conn_tup_by_go_tls_conn` map.

### Motivation

Fixes a leak in the map and improves USM accuracy.

* Deployed the image to the staging cluster and validated the fix.

* Ran a load test with the changes to observe the impact.

* Added a unit test to validate that the cleanup works (confirmed that without the fix, the cleanup does not occur).


pkg/network/usm/ebpf_gotls.go
```diff
@@ -184,6 +186,10 @@ func (p *goTLSProgram) ConfigureOptions(options *manager.Options) {
 		MaxEntries: p.cfg.MaxTrackedConnections,
 		EditorFlag: manager.EditMaxEntries,
 	}
+	options.MapSpecEditors[goTLSConnByTupleMap] = manager.MapSpecEditor{
+		MaxEntries: p.cfg.MaxTrackedConnections,
+		EditorFlag: manager.EditMaxEntries,
+	}
```
guyarb, line 192: What's the map size?

pkg/network/usm/usm_http_monitor_test.go
```diff
@@ -228,3 +228,54 @@ func getHTTPUnixClientArray(size int, unixPath string) []*http.Client {
 
 	return res
 }
+
+func TestGoTLSMapCleanup(t *testing.T) {
```
guyarb, line 232: Perhaps we should document how to reproduce the bug, as it might not be clear to the naive reader

pkg/network/ebpf/c/protocols/tls/go-tls-conn.h
```diff
@@ -164,6 +164,7 @@ static __always_inline conn_tuple_t* conn_tup_from_tls_conn(tls_offsets_data_t*
     }
 
     bpf_map_update_with_telemetry(conn_tup_by_go_tls_conn, &conn, &tuple, BPF_ANY);
+    bpf_map_update_elem(&go_tls_conn_by_tuple, &tuple, &conn, BPF_ANY);
```
guyarb: what about with_telemetry

pkg/network/usm/ebpf_gotls.go
amitslavin: The same as `conn_tup_by_go_tls_conn` - max capacity of 131.1k entries

guyarb (approved):

PR #18173, opened 2023-07-13
Offering: changes required for macOS agent builds on Apple Silicon

timsutton:
### What does this PR do?

This PR is a demonstration of changes I needed to make in order to perform a full agent build on Apple Silicon, including a couple changes in Datadog's forks of omnibus and omnibus-software. Many of the changes are explicit overrides to use arm arch over x86_64, so those should all be considered as hacks and not intended to actually merge. I have been just hacking at trying to get something that builds end to end and installs + runs on a machine without Rosetta.

The high-level summary:

* pull in a current libffi dependency
* replace x86_64 with arm build target on the DD Mac App and openssl3 dependencies
* skip including some extra dependencies that ended up building and shipping pgadmin/pg tools, lxml etc. when it wasn't needed
* tell omnibus to ignore git-fsmonitor stuff from .git (a quick hack, not related to Apple Silicon but just due to newer git tooling – I didn't check omnibus upstream for this)
* hack to include the proper architectures in the Distribution XML ERB used in omnibus resources (best would be if we could directly use https://github.com/chef/omnibus/pull/1005 but it relies on some other methods we didn't have in the DD fork)
* as a bonus, remove a `swift-stdlib-tool` flag that seems to have been removed in Xcode 14.3, so that current Xcode versions can be used to build this

### Motivation

At my dayjob I work for a large customer of DataDog. We've opened a feature req (via a support ticket) but so far it's been hinted that this is probably not coming up around the corner. We run a fairly large Mac build cluster that is very performance-sensitive and currently this agent is the only reason we need to install Rosetta2. 

### Additional Notes

I am _definitely_ not aiming to get these changes landed in exactly their current form. I just wanted to present these changes as something for y'all to use if you haven't done this work internally already. I'm also happy to work with folks on making some of these changes more backwards-compatible. However they do also still require changes in the omnibus/omnibus-software forks as well.

I don't think there's value in trying to compile all these dependencies as Universal, btw. The python stuff gets annoying, go doesn't provide an out-of-the-box way to do it and it would make the package extremely large. I'd assume that when you folks are building Apple Silicon packages that you'd be building packages which aren't dual-architecture.

I also realize that you maybe don't have Apple Silicon machines available for CI. My goal here is to first just get the support landed into the main branches, and leave that part up to you :). In doing this work I'm now confident that I'm able to build these independently out of tagged branches.

### Possible Drawbacks / Trade-offs

This branch can only be used to build an arm64 darwin pkg, x86_64 code won't work.

[...]

Issue #16147, opened 2023-03-16
[Feature Request] Allow installing DataDog agent on M2 Macbooks without Rosetta 2


ryanhugh:
Right now, the datadog-agent for Mac requires Rosetta 2 to be installed. It would be great if you could add support for the arm64 M2 architecture natively, without the need for Rosetta. 

<img width="631" alt="image" src="https://user-images.githubusercontent.com/5777995/225692467-1a827896-3c28-4331-91f0-620af5e2ecc9.png">

Btw, is it possible to compile the agent from source for arm64 for Mac?

Thanks!

vickenty:
This issue tracker is primarily used to track bugs in the Agent codebase to completion. For feature requests, please contact [Datadog support](http://docs.datadoghq.com/help/) and share your ideas with them.

Thanks!

rdesgroppes:
We have good news for you: starting with version 7.70, the Rosetta 2 requirement will be lifted when installing the native version of the Agent for AArch64/ARM64 on macOS, which is in the works. To do this, you'll need to use the latest installation script according to the instructions found at: https://docs.datadoghq.com/agent/basic_agent_usage/osx/

PR #38498, opened 2025-07-07
[ABLD-18] Update macOS installation script for AArch64/ARM64

rdesgroppes:
### What does this PR do?
The present change adapts the macOS installation script to the foreseeable future, by considering the **optional** presence of an **architecture qualifier** in the DMG name.

### Motivation
The script doesn't need to wait for AArch64/ARM64 DMGs to be publicly released - quite the other way around as we're going to add an architecture qualifier to DMGs, see: DataDog/omnibus-ruby#240.

In the absence of automated tests, full credit go to @alopezz:

image

When looking for an architecture-specific DMG, the script sends a HEAD request and simply checks against a HTTP `200` response status (happy path). This means that not only HTTP `404` but also HTTP `50x` as well as `curl` errors will lead to falling back to the "universal" flavor. This is pretty conservative for the time being and might be reconsidered later.

### Additional Notes
See also #16147, #18173.

### DIFF
--- a/cmd/agent/install_mac_os.sh
+++ b/cmd/agent/install_mac_os.sh
@@ -5,7 +5,7 @@
 
 # Datadog Agent install script for macOS.
 set -e
-install_script_version=1.4.0
+install_script_version=1.5.0
 dmg_file=/tmp/datadog-agent.dmg
 dmg_base_url="https://s3.amazonaws.com/dd-agent"
 etc_dir=/opt/datadog-agent/etc
@@ -63,19 +63,13 @@ if [ -n "$DD_AGENT_MINOR_VERSION" ]; then
 fi
 
 arch=$(/usr/bin/arch)
-if [ "$arch" == "arm64" ]; then
-  if ! /usr/bin/pgrep oahd >/dev/null 2>&1; then
-    printf "\033[31mRosetta is needed to run datadog-agent on $arch.\nYou can install it by running the following command :\n/usr/sbin/softwareupdate --install-rosetta --agree-to-license\033[0m\n"
-    exit 1
-  fi
-fi
 
 # Cleanup tmp files used for installation
 rm -f /tmp/install-ddagent/system-wide
 
 function find_latest_patch_version_for() {
     major_minor="$1"
-    patch_versions=$(curl "https://s3.amazonaws.com/dd-agent?prefix=datadog-agent-${major_minor}." 2>/dev/null | grep -o "datadog-agent-${major_minor}.[0-9]*-1.dmg")
+    patch_versions=$(curl "$dmg_base_url?prefix=datadog-agent-${major_minor}." 2>/dev/null | grep -Eo "datadog-agent-${major_minor}.[0-9]*-1(.$arch)?.dmg")
     if [ -z "$patch_versions" ]; then
         echo "-1"
     fi
@@ -218,12 +212,6 @@ if [ -z "$dmg_version" ]; then
     fi
 fi
 
-if [ -z "$agent_dist_channel" ]; then
-    dmg_url="$dmg_base_url/datadog-agent-${dmg_version}.dmg"
-else
-    dmg_url="$dmg_base_url/$agent_dist_channel/datadog-agent-${dmg_version}.dmg"
-fi
-
 if [ "$upgrade" ]; then
     if [ ! -f $etc_dir/datadog.conf ]; then
         printf "\033[31mDD_UPGRADE set but no config was found at $etc_dir/datadog.conf.\033[0m\n"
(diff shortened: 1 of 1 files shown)


cmd/agent/install_mac_os.sh
```diff
@@ -63,19 +63,13 @@ if [ -n "$DD_AGENT_MINOR_VERSION" ]; then
...
-  fi
-fi
 
 # Cleanup tmp files used for installation
 rm -f /tmp/install-ddagent/system-wide
 
 function find_latest_patch_version_for() {
     major_minor="$1"
-    patch_versions=$(curl "https://s3.amazonaws.com/dd-agent?prefix=datadog-agent-${major_minor}." 2>/dev/null | grep -o "datadog-agent-${major_minor}.[0-9]*-1.dmg")
+    patch_versions=$(curl "$dmg_base_url?prefix=datadog-agent-${major_minor}." 2>/dev/null | grep -Eo "datadog-agent-${major_minor}.[0-9]*-1(.$arch)?.dmg")
```
rdesgroppes, line 72: Notes to reviewers:
1. DRY: leveraging existing `dmg_base_url` defined somewhere above (and set to `https://s3.amazonaws.com/dd-agent`)
2. `grep -E`: so as to avoid escaping parenthesis in the regexp (`-E` is portable, contrarily to `-P`)
3. `(.$arch)?`: retained DMGs will be either "universal" (well, let's call it that way) or match the current `$arch` (important is to discard other archs)

cmd/agent/install_mac_os.sh
```diff
@@ -365,6 +353,22 @@ function plist_modify_user_group() {
     $sudo_cmd sh -c "sed $i_cmd -e \"${closing_dict_line},${closing_dict_line}s|</dict>|<key>$user_parameter</key><string>$user_value</string>\n</dict>|\" -e \"${closing_dict_line},${closing_dict_line}s|</dict>|<key>$group_parameter</key><string>$group_value</string>\n</dict>|\" \"$plist_file\""
 }
 
+# Determine agent flavor to install
+if [ -z "$agent_dist_channel" ]; then
+    dmg_url_prefix="$dmg_base_url/datadog-agent-${dmg_version}"
+else
+    dmg_url_prefix="$dmg_base_url/$agent_dist_channel/datadog-agent-${dmg_version}"
+fi
```
rdesgroppes, line 361: Note to reviewers: moved from somewhere above because determining `dmg_url` (now `dmg_url_prefix`) was too distant from its usage.

cmd/agent/install_mac_os.sh
```diff
@@ -365,6 +353,22 @@ function plist_modify_user_group() {
...
+
+dmg_url="$dmg_url_prefix.$arch.dmg"  # favor architecture-specific DMG, if available
+if [ "$(curl --head --output /dev/null --silent --write-out '%{http_code}' "$dmg_url")" = 404 ]; then
+    dmg_url="$dmg_url_prefix.dmg"  # fallback to "universal" DMG
+    if [ "$arch" = arm64 ] && ! /usr/bin/pgrep oahd >/dev/null 2>&1; then
+        printf "\033[31mRosetta is needed to run datadog-agent on $arch.\nYou can install it by running the following command :\n/usr/sbin/softwareupdate --install-rosetta --agree-to-license\033[0m\n"
+        exit 1
+    fi
```
rdesgroppes, line 369: Note to reviewers: snippet moved from somewhere above so that it's executed only when applicable.
In particular, if `$arch` is `arm64` and the retained DMG is _also_ `arm64`, there's no point to prompt for Rosetta 2 installation.

releasenotes/notes/make-macos-install-script-favor-arch-specific-dmgs-b59b3f209c462400.yaml
```diff
@@ -0,0 +1,12 @@
...
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+enhancements:
+  - |
+    Make macOS installation script favor architecture-specific DMGs,
+    when available.
```
alopezz: As these are distributed separately from the channel, and are therefore not tied to Agent releases, I think we should skip the changelog entry in this case.

alopezz (approved):
LGTM (just remember to remove the changelog entry before merging).

releasenotes/notes/make-macos-install-script-favor-arch-specific-dmgs-b59b3f209c462400.yaml
rdesgroppes: :heavy_check_mark: Done removing changelog entry.

khewonc (approved):

PR #38503, opened 2025-07-07
Allow attaching GDB to guest VMs launched with KMT

usamasaqib:
### What does this PR do?
This PR extends the `kmt.launch-stack` task to launch guest VMs with a GDB server. This allows a user to easily debug the running kernel. The PR also adds a helper task `kmt.attach-gdb` which generates the command to attach to the running VM.

### Additional Notes
 

Corresponding PR in test-infra-definitions https://github.com/DataDog/test-infra-definitions/pull/1579


tasks/kernel_matrix_testing/gdb.py
```diff
@@ -0,0 +1,143 @@
...
+from tasks.libs.types.arch import Arch
+
+if TYPE_CHECKING:
+    from tasks.kernel_matrix_testing.types import (
+        Component,  # noqa: F401
+        KMTArchNameOrLocal,
+    )
+
+
+class KernelVersion:
```
gjulianm: I think the `semver` package is already a dependency and has this functionality already (e.g., `version = semver.VersionInfo.parse("5.8.0")`)

tasks/kernel_matrix_testing/gdb.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+class UbuntuGDBProvision:
+    def __init__(self, vm: LibvirtDomain, image_version: str, kernel: str):
+        self.target = vm
+        self.image_version = image_version
+        self.kernel = KernelVersion(kernel.split('-')[0])
+
+    def run(self, ctx: Context, stack: str):
+        self.target.copy(ctx, "tasks/kernel_matrix_testing/provision/ubuntu-dbg.sh", "/tmp/provision.sh")
```
gjulianm:
```suggestion
        self.target.copy(ctx, get_repo_root() / "tasks/kernel_matrix_testing/provision/ubuntu-dbg.sh", "/tmp/provision.sh")
```

Just in case some runs from another path.

gjulianm (approved):

tasks/kernel_matrix_testing/gdb.py
```diff
@@ -0,0 +1,114 @@
...
+        self.target.copy(
+            ctx, get_repo_root() / "tasks/kernel_matrix_testing/provision/ubuntu-dbg.sh", "/tmp/provision.sh"
+        )
+        self.target.run_cmd(ctx, "chmod +x /tmp/provision.sh && /tmp/provision.sh")
+
+        gdb_paths = GDBPaths(self.target.tag, self.image_version, stack, self.target.arch)
+        gdb_paths.vmlinux.parent.mkdir(exist_ok=True, parents=True)
+        self.target.download(ctx, "/usr/lib/debug/boot/vmlinux.dbg", f"{gdb_paths.vmlinux}")
+
+        ctx.run(f"rm -rf {gdb_paths.kernel_source}")
```
chouetz: This sounds "dangerous", should we have something to prevent it from deleting the whole filesystem?

tasks/kernel_matrix_testing/provision/bata24.sh
```diff
@@ -0,0 +1,58 @@
...
+
+echo "[+] apt"
+sudo apt-get update
+sudo apt-get install -y gdb-multiarch binutils gcc file python3-pip ruby-dev git
+
+echo "[+] pip3"
+pip3 install crccheck unicorn capstone ropper keystone-engine tqdm
+
+echo "[+] install seccomp-tools, one_gadget"
+if [ "x$(which seccomp-tools)" = "x" ]; then
```
chouetz:
```suggestion
if [[ -z "$(which seccomp-tools)" ]]; then
```
I think we should follow [these recommendations](https://google.github.io/styleguide/shellguide.html#testing-strings) for testing strings in shell script.

tasks/kernel_matrix_testing/provision/bata24.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+if [[ $UID = 0 ]] ; then
```
chouetz:
```suggestion
if [[ "$UID" == 0 ]] ; then
```
For clarity, we should use == for equality rather than = even though both work. The former encourages the use of [[ and the latter can be confused with an assignment

tasks/kernel_matrix_testing/provision/bata24.sh
```diff
@@ -0,0 +1,58 @@
...
+
+echo "[+] pip3"
+pip3 install crccheck unicorn capstone ropper keystone-engine tqdm
+
+echo "[+] install seccomp-tools, one_gadget"
+if [ "x$(which seccomp-tools)" = "x" ]; then
+    sudo gem install seccomp-tools
+fi
+
+if [ "x$(which one_gadget)" = "x" ]; then
```
chouetz:
```suggestion
if [[ -z "$(which one_gadget)" ]]; then
```

tasks/kernel_matrix_testing/provision/bata24.sh
```diff
@@ -0,0 +1,58 @@
...
+    sudo gem install one_gadget
+fi
+
+echo "[+] install rp++"
+if [ "x$(uname -m)" = "xx86_64" ]; then
+    if [ "x$(which rp-lin)" = "x" ] && [ ! -e /usr/local/bin/rp-lin ]; then
```
chouetz:
```suggestion
if [[ "$(uname -m)" == "x86_64" ]]; then
    if [[ -z "x$(which rp-lin)" ]] && [[ ! -e /usr/local/bin/rp-lin ]]; then
```

tasks/kernel_matrix_testing/provision/bata24.sh
```diff
@@ -0,0 +1,58 @@
...
+        wget -q https://github.com/0vercl0k/rp/releases/download/v2.1.3/rp-lin-clang.zip -P /tmp
+        sudo unzip /tmp/rp-lin-clang.zip -d /usr/local/bin/
+        sudo chmod +x /usr/local/bin/rp-lin
+        rm /tmp/rp-lin-clang.zip
+    fi
+fi
+
+echo "[+] install vmlinux-to-elf"
+if [ "x$(which vmlinux-to-elf)" = "x" ] && [ ! -e /usr/local/bin/vmlinux-to-elf ]; then
```
chouetz:
```suggestion
if [[ -z "$(which vmlinux-to-elf)" ]] && [[ ! -e /usr/local/bin/vmlinux-to-elf ]]; then
```

tasks/kernel_matrix_testing/provision/bata24.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+
+set -euxo pipefail
```
CelianR:
```suggestion
set -euo pipefail
```
We prefer to not use `-x` in this codebase, is it mandatory?

tasks/kernel_matrix_testing/provision/bata24.sh
```diff
@@ -0,0 +1,58 @@
...
+fi
+
+echo "[+] install vmlinux-to-elf"
+if [ "x$(which vmlinux-to-elf)" = "x" ] && [ ! -e /usr/local/bin/vmlinux-to-elf ]; then
+    pip3 install --upgrade lz4 zstandard git+https://github.com/clubby789/python-lzo@b4e39df
+    pip3 install --upgrade git+https://github.com/marin-m/vmlinux-to-elf
+fi
+
+echo "[+] download gef"
+if [ -e ~/.gdbinit-gef.py ]; then
```
chouetz:
```suggestion
if [[ -e ~/.gdbinit-gef.py ]]; then
```

tasks/kernel_matrix_testing/provision/bata24.sh
CelianR:
```suggestion
if [ "$(which seccomp-tools)" = "" ]; then
```
This should work fine, right?

CelianR: (same bellow)

tasks/kernel_matrix_testing/provision/bata24.sh
```diff
@@ -0,0 +1,58 @@
...
+    echo "[-] ~/.gdbinit-gef.py already exists. Please delete or rename."
+    echo "[-] INSTALLATION FAILED"
+    exit 1
+else
+    wget -q https://raw.githubusercontent.com/bata24/gef/dev/gef.py -O ~/.gdbinit-gef.py
+fi
+
+echo "[+] setup gef"
+STARTUP_COMMAND="source ~/.gdbinit-gef.py"
+if [ ! -e ~/.gdbinit ] || [ "x$(grep "$STARTUP_COMMAND" ~/.gdbinit)" = "x" ]; then
```
chouetz:
```suggestion
if [[ ! -e ~/.gdbinit ]] || [[ -z "$(grep "$STARTUP_COMMAND" ~/.gdbinit)" ]]; then
```

tasks/kernel_matrix_testing/provision/ubuntu-dbg.sh
```diff
@@ -0,0 +1,33 @@
...
+set -o nounset
+set -o xtrace
+
+# 1. disable kaslr
+[ -f $(grep -q "GRUB_CMDLINE_LINUX=\".*nokaslr" /etc/default/grub) ] && sed -i 's/^GRUB_CMDLINE_LINUX="/&nokaslr /' /etc/default/grub
+update-grub
+
+# 2. download kernel debug build and kernel sources
+codename=$(lsb_release -c | awk  '{print $2}')
+if [ "${codename}" == "xenial" ]; then
```
chouetz:
```suggestion
if [[ "${codename}" == "xenial" ]]; then
```

tasks/kernel_matrix_testing/provision/ubuntu-dbg.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+set -o errexit
+set -o pipefail
+set -o nounset
+set -o xtrace
```
CelianR:
```suggestion
```
(same question as above)

tasks/kernel_matrix_testing/provision/ubuntu-dbg.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# 1. disable kaslr
+[ -f $(grep -q "GRUB_CMDLINE_LINUX=\".*nokaslr" /etc/default/grub) ] && sed -i 's/^GRUB_CMDLINE_LINUX="/&nokaslr /' /etc/default/grub
```
chouetz:
```suggestion
[[ -f $(grep -q "GRUB_CMDLINE_LINUX=\".*nokaslr" /etc/default/grub) ]] && sed -i 's/^GRUB_CMDLINE_LINUX="/&nokaslr /' /etc/default/grub
```

chouetz (approved):
One question and style suggestion, this is ok for me

chouetz (approved):
One question and style suggestion, this is ok for me

CelianR (approved):

tasks/kernel_matrix_testing/gdb.py
usamasaqib: Yes we definitely should! I will rework it.

tasks/kernel_matrix_testing/provision/bata24.sh
usamasaqib: It is not mandatory, but why is it preferred not to use `-x` or `-xtrace`?

tasks/kernel_matrix_testing/provision/ubuntu-dbg.sh
usamasaqib: Turns out I have to do
`[[ ! $(grep -q "GRUB_CMDLINE_LINUX=\".*nokaslr" /etc/default/grub) ]]` when using `[[`

`-f` does not work

tasks/kernel_matrix_testing/provision/bata24.sh
CelianR: We had some issues related to a token leak some time ago and we prefer to not include these in any script even though the script doesn't contain any secrets as a good practice (to avoid "copy pasting" this into a script that might leak)

PR #38505, opened 2025-07-07
[DEBUG-4039] dyninst/rcscrape: implementation of RC scraping

ajwerner:
### What does this PR do?

This pile of commits implements the logic to scrape the remote config from dd-trace-go processes. The primary implementation is in the last commit. The first two commits adjust the actuator interfaces in important ways. The following two add testing prerequisites:

#### dyninst/testprogs: add rc_tester

This is a program that is useful for testing remote config controlled
probes.

#### dyninst/actuator: fail IR gen when no probes succeed

If we have no probes, then the loading of the program is doomed to fail.
Rather than fail, instead we want to call back into the reporter and tell
it about the issues.

#### dyninst/actuator: add ProcessID to Reporter callbacks

For now, every program is 1:1 with a process. Even in some future where it's
not, we'll always want to associate lifecycle events with processes. In that
future, instead of passing a single process ID, we'll pass a set of them.

### Additional Notes

This was split out of https://github.com/DataDog/datadog-agent/pull/38488.

This implements https://datadoghq.atlassian.net/browse/DEBUG-4039


piob-io (approved):
Thank you!

PR #38506, opened 2025-07-07
[otel] support db.namespace semantic for OTel

jackgopack4:
### What does this PR do?
adds code for OTLP to map `db.namespace` to `db.name`
### Motivation
properly support required DBM semantics
add unit tests

make spans slightly longer (shouldn't be more than the name of the database)
### Additional Notes
if `db.name` and `db.namespace` are included, leaves `db.name` alone

### DIFF
--- a/pkg/trace/transform/transform.go
+++ b/pkg/trace/transform/transform.go
@@ -444,7 +444,7 @@ func OtelSpanToDDSpan(
 
 	// Check for db.namespace and conditionally set db.name
 	if _, ok := ddspan.Meta["db.name"]; !ok {
-		if dbNamespace := traceutil.GetOTelAttrFromEitherMap(otelres.Attributes(), otelspan.Attributes(), false, string(semconv127.DBNamespaceKey)); dbNamespace != "" {
+		if dbNamespace := traceutil.GetOTelAttrValInResAndSpanAttrs(otelspan, otelres, false, string(semconv127.DBNamespaceKey)); dbNamespace != "" {
 			ddspan.Meta["db.name"] = dbNamespace
 		}
 	}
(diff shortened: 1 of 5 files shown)


git-thuerk-done (approved):
good for docs

pkg/trace/transform/transform.go
```diff
@@ -441,6 +442,13 @@ func OtelSpanToDDSpan(
 		ddspan.Meta[k] = v.AsString()
 	}
 
+	// Check for db.namespace and conditionally set db.name
+	if _, ok := ddspan.Meta["db.name"]; !ok {
+		if dbNamespace := traceutil.GetOTelAttrFromEitherMap(otelres.Attributes(), otelspan.Attributes(), false, string(semconv127.DBNamespaceKey)); dbNamespace != "" {
```
liustanley: Is there a reason why we use `GetOTelAttrValInResAndSpanAttrs` above but `GetOTelAttrFromEitherMap` here? It might be better to abstract this condition in these two places if we can use the same function for both.

jackgopack4: I think I agree with you; i can change it to match and see if everything still works

jackgopack4: looks likek this function (GetOTelAttrFromEitherMap) was only created a month ago; it's only used for resource attributes and span attributes; I'd like to separate that into a separate PR to clean up/remove that function

jackgopack4: @IbraheemA added this function to be able to preference span attributes over resource attributes. So whatever this function is doing is usually being used to select from span attrs FIRST and then fall back on resource attributes. Whether that is the correct/intended behavior is not necessarily a trivial discussion. so I think we should leave it as is for now

liustanley (approved):
LGTM

PR #38507, opened 2025-07-07
[AGTMETRICS-307] Use separate channel for configuration changes

rayz:
### What does this PR do?
Two changes:
1. Instead of notifying the receivers in the `Set` / `Unset` , we send the configuration update to a separate channel. In another go routine, we process these updates and notify the receivers. 

It is important to note that `Set` / `Unset` does not wait for the notification to be processed.

2. Introduces `AllSettingsWithSequenceID` method that returns the settings along with the sequence id associated with the settings.

### Motivation
The purpose of the new change is to ensure that notification receiver call backs are processed sequentially.

Existing unit tests

### DIFF
--- a/comp/autoscaling/datadogclient/impl/client.go
+++ b/comp/autoscaling/datadogclient/impl/client.go
@@ -130,3 +130,9 @@ func createDatadogClient(cfg configComponent.Component, logger logComp.Component
 	}
 	return newDatadogSingleClient(cfg, logger)
 }
+
+func (d *datadogClientWrapper) getNumberOfRefreshes() int {
+	d.mux.RLock()
+	defer d.mux.RUnlock()
+	return d.numberOfRefreshes
+}
--- a/pkg/config/model/types.go
+++ b/pkg/config/model/types.go
@@ -132,6 +132,7 @@ type ConfigChangeNotification struct {
 	PreviousValue any
 	NewValue      any
 	SequenceID    uint64
+	Receivers     []NotificationReceiver
 }
 
 // Reader is a subset of Config that only allows reading of configuration
--- a/pkg/config/model/types.go
+++ b/pkg/config/model/types.go
@@ -126,6 +126,14 @@ type Proxy struct {
 // 'NotificationReceiver' should not be blocking.
 type NotificationReceiver func(setting string, oldValue, newValue any, sequenceID uint64)
 
+// ConfigChangeNotification stores the information about a change in the configuration and is sent to the listeners.
+type ConfigChangeNotification struct {
+	Key           string
+	PreviousValue any
+	NewValue      any
+	SequenceID    uint64
+}
+
 // Reader is a subset of Config that only allows reading of configuration
 type Reader interface {
 	Get(key string) interface{}
@@ -157,6 +165,7 @@ type Reader interface {
 	// AllKeysLowercased returns all config keys in the config, no matter how they are set.
 	// Note that it returns the keys lowercased.
 	AllKeysLowercased() []string
+	AllSettingsWithSequenceID() (map[string]interface{}, uint64)
 
 	// SetTestOnlyDynamicSchema is used by tests to disable validation of the config schema
 	// This lets tests use the config is more flexible ways (can add to the schema at any point,
--- a/pkg/config/model/types.go
+++ b/pkg/config/model/types.go
@@ -129,8 +129,8 @@ type NotificationReceiver func(setting string, oldValue, newValue any, sequenceI
 // ConfigChangeNotification stores the information about a change in the configuration and is sent to the listeners.
 type ConfigChangeNotification struct {
 	Key           string
-	PreviousValue any
-	NewValue      any
+	PreviousValue interface{}
+	NewValue      interface{}
 	SequenceID    uint64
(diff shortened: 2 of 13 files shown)


pkg/config/model/types.go
```diff
@@ -126,6 +126,15 @@ type Proxy struct {
 // 'NotificationReceiver' should not be blocking.
 type NotificationReceiver func(setting string, oldValue, newValue any, sequenceID uint64)
 
+// ConfigChangeNotification stores the information about a change in the configuration and is sent to the listeners.
+type ConfigChangeNotification struct {
+	Key           string
+	PreviousValue any
+	NewValue      any
```
rahulkaukuntla: afaik, `any` and `interface{}` have the same functionality, so just to maintain uniformity, could you make these struct values have the `interface{}` type?

rahulkaukuntla (approved):
FYI, some linters seem to be failing. For now, I will give a reluctant LGTM, as you do have a test that adequately checks if the configuration updates are being registered with the new channel by calling `onUpdate`. However, this test isn't using a separate channel to do so, so I feel this test isn't strongly _mimicking the effects of the real code_. 

As the real code uses a channel-based approach, it would be nice if the corresponding test could do the same (I'll message you offline). Here is a rough idea:

```
func TestListenersUnsetForSource(t *testing.T) {
	config := NewViperConfig("test", "DD", strings.NewReplacer(".", "_"))

	// Create a channel to collect notifications
	notifications := make(chan string, 10)
	
	config.OnUpdate(func(_ string, _, next any, _ uint64) {
		nextString := next.(string)
		notifications <- nextString
	})

	config.Set("log_level", "info", model.SourceFile)
	config.Set("log_level", "debug", model.SourceRC)
	config.UnsetForSource("log_level", model.SourceRC)

	// Collect all notifications
	var logLevels []string
	timeout := time.After(100 * time.Millisecond)
	
	for len(logLevels) < 3 {
		select {
		case level := <-notifications:
			logLevels = append(logLevels, level)
		case <-timeout:
			t.Fatal("Timeout waiting for notifications")
		}
	}

	assert.Equal(t, []string{"info", "debug", "info"}, logLevels)
}
```

Let me know what you think.

pkg/config/nodetreemodel/config.go
```diff
@@ -348,12 +360,16 @@ func (c *ntmConfig) UnsetForSource(key string, source model.Source) {
...
-		receiver(key, previousValue, newValue, c.sequenceID)
+	notification := model.ConfigChangeNotification{
+		Key:           key,
+		PreviousValue: previousValue,
+		NewValue:      newValue,
+		SequenceID:    c.sequenceID,
+		Receivers:     slices.Clone(c.notificationReceivers),
 	}
+
+	c.notificationChannel <- notification
```
tobz: One thing that might make this a little cleaner is if we encapsulated this in a method on the type that only bothered sending the notification if `c.notificationReceivers` isn't empty.

tobz (approved):
Overall, LGTM. Left one take-it-or-leave-it suggestion on a way to reduce some of the duplication.

pkg/config/viperconfig/viper.go
```diff
@@ -84,12 +83,6 @@ func (c *safeConfig) sendNotification(key string, oldValue, newValue interface{}
 	c.notificationChannel <- notification
 }
 
-func getCallerLocation(nbStack int) string {
```
rayz, line 70: Linter is complaining about this function being unused. Removing it for now since from the history, it seems like its never been used.

lucastemb (approved):
LGTM 👍

soberpeach (approved):

jennchenn (approved):

comp/autoscaling/datadogclient/impl/client.go
```diff
@@ -130,3 +130,9 @@ func createDatadogClient(cfg configComponent.Component, logger logComp.Component
 	}
 	return newDatadogSingleClient(cfg, logger)
 }
+
+func (d *datadogClientWrapper) getNumberOfRefreshes() int {
```
jennchenn, line 134: Just adding some clarification here for future reference: this is necessary because accessing `numberOfRefreshes` without a lock causes a data race. Since the only actual time `numberOfRefreshes` is [read in the code](https://github.com/DataDog/datadog-agent/blob/fbdd1bbd7ee4d151620c6a88c17fcbf1c632cf71/comp/autoscaling/datadogclient/impl/client.go#L115-L119) is already done with a lock, this function is only used in the unit test right now.

nit: could be added as a comment but if in the future we want to read `numberOfRefreshes` we should probably use this helper anyway (not sure if this is expected to change much though)

mellon85 (approved):

pkg/config/viperconfig/viper.go
```diff
@@ -816,6 +817,8 @@ func NewViperConfig(name string, envPrefix string, envKeyReplacer *strings.Repla
 	config.SetEnvPrefix(envPrefix)
 	config.SetEnvKeyReplacer(envKeyReplacer)
 
+	go config.processNotifications()
```
songy23, line 820: This leads to a goroutine leak http://github.com/DataDog/datadog-agent/issues/40071

PR #38515, opened 2025-07-07
pkg/trace/api: Add container tags hash to info endpoint response

piochelepiotr:
### What does this PR do?

Adds a new header: `Datadog-Container-Tags-Hash` to the info endpoint response.

### Motivation

The tracers can use that hash to tag propagate an "origin" to downstream services.
Services can then be renamed in the Datadog backend, while keeping it possible to track the origin service of requests.

Unit tests are testing that the container tags hash is added as a header to the info endpoint response. Planning to also test that with the Java tracer and make use of the response.


urseberry (approved):

raphaelgavache (approved):

pkg/trace/api/info.go
```diff
@@ -125,7 +145,25 @@ func (r *HTTPReceiver) makeInfoHandler() (hash string, handler http.HandlerFunc)
 		panic(fmt.Errorf("Error making /info handler: %v", err))
 	}
 	h := sha256.Sum256(txt)
-	return fmt.Sprintf("%x", h), func(w http.ResponseWriter, _ *http.Request) {
+	return fmt.Sprintf("%x", h), func(w http.ResponseWriter, req *http.Request) {
+		if containerTags, err := r.conf.ContainerTags(req.Header.Get(containerIDHeader)); err == nil {
```
ichinaski: I'm not sure the header `Datadog-Container-ID` is the standard nowadays.  Check how how the container ID is resolved in traces and stats requests [here](https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/api/api.go#L521), which in turn calls [this function](https://github.com/DataDog/datadog-agent/blob/b2f48fb55dfecb988654951267ec7d579e780805/pkg/trace/api/container_linux.go#L116-L173).

Maybe we can reuse the same logic here?

piochelepiotr: Makes sense, I've updated the code.

ichinaski (approved):

PR #38534, opened 2025-07-07
Promote uds_dogstatsd_to_api_cpu to quality_gate_metrics_logs

blt:
### What does this PR do?

This commit promotes the pre-existing `uds_dogstatsd_to_api_cpu` to a new `quality_gate_metrics_logs` experiment, focused on CPU and asserting CPU load in a non-idle environment. It is understood as of this writing that this experiment represents a high-load scenario but given the long-term nature of this experiment in our CI without comment we understand this scenario to have importance, hence its promotion.


GeorgeHahn (approved):

test/regression/cases/quality_gate_metrics_logs/lading/lading.yaml
```diff
@@ -37,14 +40,35 @@ generator:
             distribution: 0
             set: 0
             histogram: 0
+      # `bytes_per_second` is understood to be a high-end use case,
+      # however Agent's internal telemetry differs from lading's units: UDS
+      # packets per second versus bytes per second.
       bytes_per_second: "100 MiB"
       maximum_prebuild_cache_size_bytes: "500 MiB"
 
+  # Log rotation file system for logs collection (following quality_gate_logs pattern)
```
jszwedko, line 49: I appreciate these comments as, 6-12 months from now, it is likely to be unclear how the parameters were derived. What would you think of standardizing the practice of introducing a README per regression case that documents what it is testing and why? Or we can enforce encapsulating the context in comments like these through PR review.

test/regression/cases/quality_gate_metrics_logs/experiment.yaml
```diff
@@ -0,0 +1,55 @@
...
+      series: "connection.current"
+      upper_bound: 6
+
+  - name: lost_bytes
+    description: "Allowable bytes not processed by Agent"
+    bounds:
+      series: lost_bytes
+      upper_bound: 0KiB
+
+report_links:
```
jszwedko, line 53: Unrelated to this experiment specifically, but the link here seems to be completely templated such that lading could just include it lest we have to update it in N places if we want to make a change.

test/regression/cases/quality_gate_metrics_logs/experiment.yaml
```diff
@@ -0,0 +1,55 @@
+optimization_goal: memory
+erratic: false
+
+target:
+  name: datadog-agent
+  cpu_allotment: 4
+  memory_allotment: 5 GiB
```
jszwedko, line 7: I assume these are the "default" parameters in the Agent? A comment could help.

jszwedko (approved):
Agreed with starting somewhere even if we aren't sure what the most valuable parameters might be.

test/regression/cases/quality_gate_metrics_logs/lading/lading.yaml
blt: I like the idea of a README. I'll add one to this experiment and we can discuss in this context.

test/regression/cases/quality_gate_metrics_logs/experiment.yaml
blt: They actually vary by experiment. We don't have good guidance yet on when to adjust these down -- that's part of my motivation for building additional tooling around threshold style experiments -- but in general we do want them to be "close" to any checks that make assertions on these. 

Today these are loose so we can gain experience with the experiment without immediately tripping an incident.

GeorgeHahn (approved):

jszwedko (approved):
Thanks for adding the README!

PR #38546, opened 2025-07-08
usm: http: tests: Added HTTP GoTLS tests

guyarb:
### What does this PR do?

Adding direct HTTP GoTLS tests, as we don't have a proper tests today

Also, this PR introduces a new file and a new test suite, as HTTP is the only protocol that does not follow this convention. We will be migrating tests to the new file and suite

### Motivation

Adding proper tests for HTTPs with Go-TLS service.

### DIFF
--- a/pkg/network/usm/usm_http_monitor_test.go
+++ b/pkg/network/usm/usm_http_monitor_test.go
@@ -1,0 +1,217 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-present Datadog, Inc.
+
+//go:build linux_bpf
+
+package usm
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"net/http"
+	"strconv"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+
+	"github.com/DataDog/datadog-agent/pkg/ebpf/ebpftest"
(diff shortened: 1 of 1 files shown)


pkg/network/usm/usm_http_monitor_test.go
```diff
@@ -0,0 +1,217 @@
...
+		utils.WaitForProgramsToBeTraced(t, consts.USMModuleName, GoTLSAttacherName, proxyProcess.Process.Pid, utils.ManualTracingFallbackEnabled)
+	}
+
+	tests := []struct {
+		name              string
+		runClients        func(t *testing.T, clientsCount int)
+		expectedEndpoints map[usmhttp.Key]int
+	}{
+		{
+			name: "GET /hello",
```
amitslavin: Maybe we should name the test in a way that better reflects its purpose?

pkg/network/usm/usm_http_monitor_test.go
```diff
@@ -0,0 +1,217 @@
...
+}
+
+func (s *usmHTTPSuite) getSchemeURL(url string) string {
+	if s.isTLS {
+		return "https://" + url
+	}
+	return "http://" + url
+}
+
+func TestHTTPScenarios(t *testing.T) {
```
amitslavin, line 71: Although it's pretty funny to suggest it (since obviously it's Go), let's try to include goTLS in the test names. it would help when searching across the whole project.

amitslavin (approved):

pkg/network/usm/usm_http_monitor_test.go
guyarb: The main issue, the test name is shared with plaintext and TLS
so having GoTLS in the test name will be misleading for the plaintext

PR #38557, opened 2025-07-08
[ACIX-921] Implement code coverage computation for host based tests

KevinFairise2:
### What does this PR do?

Add the base interface to have an environment that compute coverage of the agent under tests.
Remove the WARNING that are emitted by go binaries instrumented with covereage.

Implement coverage computation for Host-based tests

### Motivation

Get Code coverage for e2e tests in order to improve the way we run them

### Additional Notes

The coverage is working properly on Linux, there is an issue with the build of the instrumented binary on Windows. This will be fixed in a dedicated PR


test/new-e2e/pkg/e2e/suite.go
```diff
@@ -327,9 +329,22 @@ func (bs *BaseSuite[Env]) init(options []SuiteOption, self Suite[Env]) {
...
+		bs.coverageOutDir = coverageOutDir
+	} else {
+		bs.coverage = false
+		fmt.Println("WARNING: Coverage pipeline is enabled but coverage out dir is not set, skipping coverage")
+	}
+
 	stackNameSuffix, err := runner.GetProfile().ParamStore().GetWithDefault(parameters.StackNameSuffix, "")
 	if err != nil {
-		bs.T().Fatalf("unable to get stack name suffix: %v", err)
+		fmt.Printf("unable to get stack name suffix: %v\n", err)
```
CelianR: Should this exit?

test/new-e2e/tests/containers/k8s_test.go
```diff
@@ -1606,45 +1604,6 @@ func (suite *k8sSuite) testHPA(namespace, deployment string) {
 	})
 }
 
-type podExecOption func(*corev1.PodExecOptions)
```
CelianR, line 1609: I'm not sure to get it, why is this removed?

CelianR (approved):
LGTM but some parts are blurry to me, feel free to add more context by answering my comments!

test/new-e2e/pkg/e2e/suite.go
KevinFairise2: Yeah indeed, I had to change `bs.T().Fatalf` because at that point `bs.T()` is not initialized yet, which throw a nill pointer error when it fails. I will update it again

test/new-e2e/tests/containers/k8s_test.go
KevinFairise2: I removed it because it was not used at all, I can introduce it back

CelianR: No it's fine like that, it was just to get more context. Dead code elimination is great 👍

sblumenthal (approved):

PR #38558, opened 2025-07-08
[CWS] Expose telemetry on SECL variables

lebauce:
### What does this PR do?

Expose the number of instantiated SECL variables as telemetry

### DIFF
--- a/pkg/security/probe/probe.go
+++ b/pkg/security/probe/probe.go
@@ -60,20 +60,11 @@ type PlatformProbe interface {
 	EnableEnforcement(bool)
 }
 
-const (
-	// subsystem is the subsystem name for the provided telemetry for variables
-	subsystem = "secl"
-)
-
-var (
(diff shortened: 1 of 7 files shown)


spikat (approved):

pkg/security/probe/probe.go
```diff
@@ -59,6 +60,21 @@ type PlatformProbe interface {
...
 }
 
+const (
+	// subsystem is the subsystem name for the provided telemetry for variables
+	subsystem = "secl"
+)
+
+var (
+	// totalVariables tracks the total number of SECL variables
+	totalVariables = telemetry.NewGauge(
```
paulcacheux: Why not using stuff from `pkg/security/metrics`, like the hash resolver ?
This would allow to have `runtime_security` in the metric name (which is not the case today)

lebauce: :+1:

paulcacheux (approved):

PR #38559, opened 2025-07-08
[FUZZ] Support for internal fuzzing infra

edznux-dd:
### What does this PR do?

This adds support to run all the existing Go fuzz test in the internal fuzz infrastructure.
It'll run all Go fuzzer on a daily schedule. For now it'll start them all at roughly the same time, we might want to spread the load later on, but we can iterate on this first.

### Motivation

Improve reliability, and security, by finding bugs, security issues and regression.

We do not run the fuzz test in CI, but in our internal dedicated fuzzing infrastructure. We only need to _build_ the binary, in the CI (and upload them via our API),
Bug report will come through slack / workflows / monitors, but won't block CI.

We probably only want to run this:
- on a daily scheduled run based on main, to avoid starting thousands of fuzzers a day
- on demand, via some manually trigger on specific branches when doing sensitive changes.

### Additional Notes

Some things could be improved in another PR:
- configurable settings. Depending on the fuzz test, the fuzzer may need more or less resources and time. Right now, we don't have data for these fuzzer to assess the needs in resources and time.


.gitlab/JOBOWNERS
```diff
@@ -187,3 +187,6 @@ run_codeql_scan                   @DataDog/sdlc-security
 
 # Experiment systemd units
 validate_experiment_systemd_units @DataDog/fleet
+
+# Fuzz testing
+test_fuzz                         @DataDog/chaos-engineering
```
chouetz: Nit
```suggestion
test_fuzz                         @DataDog/chaos-engineering

```

tasks/fuzz_infra.py
```diff
@@ -0,0 +1,104 @@
...
+from invoke import task
+
+
+@task
+def build_and_upload_fuzz(ctx):
+    """
+    This builds and uploads fuzz targets to the internal fuzzing infrastructure.
+    It needs to be passed the -fuzz flag in order to build the fuzz with efficient coverage guidance.
+    """
+    # TODO: make these configurable
```
chouetz: Should you pass them as argument with default value? `Invoke` handles this for you

tasks/fuzz_infra.py
```diff
@@ -0,0 +1,104 @@
...
+    core_count = 2
+    duration = 3600  # in seconds
+    proc_count = 2
+    fuzz_memory = 4  # in GB
+    team = "chaos-platform"  # TODO: make this use the team name from the codeowners file
+
+    api_url = "https://fuzzing-api.us1.ddbuild.io/api/v1"
+    gitsha = ctx.run('git rev-parse HEAD').stdout.strip()
```
chouetz: nit
```suggestion
    git_sha = ctx.run('git rev-parse HEAD').stdout.strip()
```

tasks/fuzz_infra.py
```diff
@@ -0,0 +1,104 @@
...
+                    f'❌ Build file {rel}/{build_file} does not exist. Skipping... (It is likely that we are missing a tag for this specific fuzz target to be built)'
+                )
+                continue
+
+            # Get presigned URL
+            print(f'Getting presigned URL for {pkgname}...')
+            presigned_url_cmd = (
+                f'curl -X POST "{api_url}/apps/{pkgname}/builds/{gitsha}/url" -H "Authorization: Bearer {auth_header}"'
+            )
+            presigned_response = ctx.run(presigned_url_cmd, hide=True).stdout.strip()
```
chouetz: There is probably a risk to reveal the token here, should we call this using `requests` instead?

tasks/fuzz_infra.py
```diff
@@ -0,0 +1,104 @@
...
+                "duration": duration,
+                "type": "go-native-fuzz",
+                "function": func,
+                "team": team,
+                "process_count": proc_count,
+                "memory": fuzz_memory,
+            }
+
+            run_json_payload = json.dumps(run_payload)
+            start_fuzzer_cmd = f'curl -H "Authorization: Bearer {auth_header}" -H "Content-Type: application/json" "{api_url}/apps/{pkgname}/fuzzers" -d \'{run_json_payload}\''
```
chouetz: Same comment here, I would recommend doing calls with `requests` instead of `curl` through `Context.run`

tasks/fuzz_infra.py
```diff
@@ -0,0 +1,104 @@
...
+    ).stdout.strip()
+
+    max_pkg_name_length = 50
+    for directory, func in search_fuzz_tests(os.getcwd()):
+        with ctx.cd(directory):
+            # eg: convert "/path/to/fuzz/target" to "datadog-agent-path-to-fuzz-target".
+            # It's a unique identifier for the fuzz target.
+            rel = directory.removeprefix("/go/src/github.com/DataDog/datadog-agent/")
+            pkgname = "datadog-agent-"
+            pkgname += "-".join(rel.split('/'))[:max_pkg_name_length]
```
chouetz, line 35: This could deserve a unit test

tasks/fuzz_infra.py
```diff
@@ -0,0 +1,104 @@
...
+
+            run_json_payload = json.dumps(run_payload)
+            start_fuzzer_cmd = f'curl -H "Authorization: Bearer {auth_header}" -H "Content-Type: application/json" "{api_url}/apps/{pkgname}/fuzzers" -d \'{run_json_payload}\''
+            response = ctx.run(start_fuzzer_cmd, hide=True).stdout.strip()
+            print(f'✅ Started fuzzer for {pkgname} ({func})...')
+            response_json = json.loads(response)
+            print(response_json)
+
+
+def search_fuzz_tests(directory):
```
chouetz, line 86: This could deserve a unit test.

chouetz (approved):
Thanks for your contribution.
Few remarks concerning style and testing.

tasks/fuzz_infra.py
edznux-dd: Great idea, done.

tasks/fuzz_infra.py
edznux-dd: Great idea, done.

tasks/fuzz_infra.py
edznux-dd: Yep, didn't know how invoke worked, but yes, that sounds good, and with automatic args creation that works fine.
thanks!

tasks/fuzz_infra.py
```diff
@@ -0,0 +1,100 @@
...
+
+@task
+def build_and_upload_fuzz(ctx, team="chaos-platform", core_count=2, duration=3600, proc_count=2, fuzz_memory=4):
+    """
+    This builds and uploads fuzz targets to the internal fuzzing infrastructure.
+    It needs to be passed the -fuzz flag in order to build the fuzz with efficient coverage guidance.
+    """
+
+    api_url = "https://fuzzing-api.us1.ddbuild.io/api/v1"
+    git_sha = ctx.run('git rev-parse HEAD').stdout.strip()
```
chouetz: 🥜 we could reuse the defined methods in the [git lib](https://github.com/DataDog/datadog-agent/blob/main/tasks/libs/common/git.py#L177)
```suggestion
    git_sha = get_commit_sha(ctx)
```
and you need `from tasks.libs.common.git import get_commit_sha`

tasks/fuzz_infra.py
```diff
@@ -0,0 +1,100 @@
...
+                "core_count": core_count,
+                "duration": duration,
+                "type": "go-native-fuzz",
+                "function": func,
+                "team": team,
+                "process_count": proc_count,
+                "memory": fuzz_memory,
+            }
+
+            headers = {"Authorization": f"Bearer {auth_header}", "Content-Type": "application/json"}
```
chouetz, line 78: 🥜 You could define just once this at line 48 and reuse it, as I think you call for `json` in both cases

chouetz (approved):

PR #38563, opened 2025-07-08
chore(fleet): Best effort to keep comments when updating YAMLs

BaptisteFoy:
### What does this PR do?

Fleet Automation now uses YAML replacement to update configuration, which is much safer than string replaces. The caveat of this is that it removed comments, ours and customers'. This PR fixes that as a best effort by serializing the YAML as `yaml.Node`. It supports the following cases:
- Commented out keys (like in datadog.yaml.example `# site: datadoghq.com`)
- Comments before & after keys
- Inline comments (`key: value # comment`)

Due to YAML Marshal, indented comments will lose their indentation so I updated the `<name>.yaml.example` templates to avoid this. E.g.:
```yaml
     ## this is a comment
```
will become
```yaml
## this is a comment
```

Manual QA, unit tests

### Additional Notes
Most of the changes are to fix the major caveat specified above, to move the comment indentation to __after__ the comment char (`#`).


pkg/fleet/installer/setup/config/write.go
```diff
@@ -0,0 +1,254 @@
...
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+)
+
+var (
+	activeKeyCommentRegex  = regexp.MustCompile(`^(\s*)([\w\-]+)\s*:\s*([^\n#]*?)(\s*)(#.*)?$`)
```
san-jos: **❓ question:** ‏What happens for an entry ?
`key: "#value"`

pkg/fleet/installer/setup/config/write.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+	updatedYamlText := strings.Join(updatedLines, "\n")
+
+	// Step 3: Load original YAML into a node tree
+	var root yaml.Node
+	if err := yaml.Unmarshal([]byte(updatedYamlText), &root); err != nil {
+		return err
+	}
+
+	// Step 5: Merge the updated `config` node tree into the original YAML
```
san-jos: **🥜 nitpick:** No step 4 😄 ‏

pkg/fleet/installer/setup/config/write.go
```diff
@@ -0,0 +1,254 @@
...
+
+	"gopkg.in/yaml.v3"
+)
+
+var (
+	activeKeyCommentRegex  = regexp.MustCompile(`^(\s*)([\w\-]+)\s*:\s*([^\n#]*?)(\s*)(#.*)?$`)
+	commentedKeyValueRegex = regexp.MustCompile(`^(\s*)#\s*([\w\-]+)\s*:\s*(.*?)(\s*)(#.*)?$`)
+)
+
+func writeConfig(path string, config any, perms os.FileMode, merge bool) error {
```
san-jos, line 24: **👏 praise:** ‏!!

pkg/fleet/installer/setup/config/write.go
BaptisteFoy: Oops

pkg/fleet/installer/setup/config/write.go
BaptisteFoy: It returns something weird, good catch; fixed the regex and added a test 👍

san-jos (approved):

PR #38565, opened 2025-07-08
[incident-40404] Pass `ddtrace`'s `crashtracker` to `codesign`

rdesgroppes:
### What does this PR do?
Add `ddtrace`'s `crashtracker` tool to the list of files to process by Apple's code signing.

### Motivation
Notarization requests started failing silently seemingly due to an error code being swallowed, so we didn't notice the binary was not `codesign`ed.
Current macOS deliverables (as of Agent release 7.68.0) are therefore affected, hence the need for a short-term countermeasure that will buy us some time for a proper fix.

### Describe how you validated your changes
Notarization request [succeeds](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1018504482):
```
Sending notarization request.
Results: Conducting pre-submission checks for datadog-agent-7.69.0-devel.git.340.7e6d3e4.pipeline.69918768-1.dmg and initiating connection to the Apple notary service...
Submission ID received
  id: feece83d-242a-4130-9bde-059acb6e6a77
Successfully uploaded file
  id: feece83d-242a-4130-9bde-059acb6e6a77
  path: /Users/ec2-user/builds/S5mXwxTmW/0/DataDog/datadog-agent/omnibus/pkg/datadog-agent-7.69.0-devel.git.340.7e6d3e4.pipeline.69918768-1.dmg
Waiting for processing to complete. Wait timeout is set to 900.0 second(s).
Current status: Accepted....................................Processing complete
  id: feece83d-242a-4130-9bde-059acb6e6a77
  status: Accepted
Submission ID: feece83d-242a-4130-9bde-059acb6e6a77
```

### Possible Drawbacks / Trade-offs
This is obviously a short-term countermeasure: `crashtracker`'s path is hardcoded, notarization failures still pass silently (fixed by #38567), etc.

### Additional Notes
Join the #[incident-40404](https://dd.enterprise.slack.com/archives/C094FAPT2DV) channel.

PR #38567, opened 2025-07-08
[ABLD-94][BARX-847][incident-40404] Verify `notarytool`'s resulting status

rdesgroppes:
### What does this PR do?
First, The fallback expression `|| EXIT_CODE=$?` was evaluated in a subshell, which caused it to be swallowed from the perspective of the parent shell. The fix simply consists in moving it to the parent shell.

In addition, it turns out that `notarytool submit [...] --wait` doesn't return a `nonzero` exit code on `Invalid` statuses, which might be by design (or might be fixed in a later `Xcode` version).

The present change therefore changes the approach to fail fast(er) by:
- first enabling [recommended shell flags](https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail),
- and then verifying `notarytool log`'s resulting status which should be exactly `Accepted` as per https://keith.github.io/xcode-man-pages/notarytool.1.html#wait.

This used to fail as intended until #38565 got merged and became an ancestor to the present change.

For instance, should `notarytool log`'s resulting status be `Invalid`, the `agent_dmg-x64-a7` job would then terminate as follows:
```
Submission was not accepted, got: Invalid
Attempt #3 failed with error code 1
Running after_script
Cleaning up project directory and file based variables
ERROR: Job failed: exit status 1
```
Conversely, for an `Accepted` resulting status:
```
\e[0K
Built signed package
$ $S3_CP_CMD $OMNIBUS_PACKAGE_DIR/version-manifest.json $S3_SBOM_STORAGE_URI/$CI_JOB_NAME/version-manifest.json
upload: omnibus/pkg/version-manifest.json to s3://sbom-root-us1-ddbuild-io/datadog-agent/99999999/agent_dmg-x64-a7/version-manifest.json
Running after_script
Saving cache for successful job
Uploading artifacts for successful job
Cleaning up project directory and file based variables
Job succeeded
```

### Motivation
Swallowed errors led to #[incident-40404](https://dd.enterprise.slack.com/archives/C094FAPT2DV), which we don't want to happen again.

### DIFF
--- a/.gitlab/package_build/build_agent_dmg.sh
+++ b/.gitlab/package_build/build_agent_dmg.sh
@@ -114,8 +114,9 @@ if [ "$SIGN" = true ]; then
     export LATEST_DMG
     # shellcheck disable=SC2016
     ./tools/ci/retry.sh -n "$NOTARIZATION_ATTEMPTS" bash -c '
+        set -euo pipefail
         EXIT_CODE=0
-        RESULT=$(xcrun notarytool submit --timeout "$NOTARIZATION_TIMEOUT" --apple-id "$APPLE_ACCOUNT" --team-id "$TEAM_ID" --password "$NOTARIZATION_PWD" "$LATEST_DMG" --wait || EXIT_CODE=$?)
+        RESULT=$(xcrun notarytool submit --timeout "$NOTARIZATION_TIMEOUT" --apple-id "$APPLE_ACCOUNT" --team-id "$TEAM_ID" --password "$NOTARIZATION_PWD" "$LATEST_DMG" --wait) || EXIT_CODE=$?
         echo "Results: $RESULT"
         SUBMISSION_ID="$(echo "$RESULT" | awk "\$1 == \"id:\"{print \$2; exit}")"
         echo "Submission ID: $SUBMISSION_ID"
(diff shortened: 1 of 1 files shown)


alopezz (approved):

alopezz (approved):

rdesgroppes:
@alopezz That was alas not enough: `notarytool` indeed happens to return a `zero` error code on `Invalid` statuses. I therefore reworked the PR accordingly. Thanks for reviewing it again :pray:

CelianR (approved):
Great, thanks for fixing this 👍

.gitlab/package_build/build_agent_dmg.sh
```diff
@@ -114,16 +114,19 @@ if [ "$SIGN" = true ]; then
...
         echo "Results: $RESULT"
         SUBMISSION_ID="$(echo "$RESULT" | awk "\$1 == \"id:\"{print \$2; exit}")"
         echo "Submission ID: $SUBMISSION_ID"
         # Wait for logs to be available
         sleep 1
         echo "Submission logs:"
         # Always show logs even if notarization fails to have more context
-        xcrun notarytool log --apple-id "$APPLE_ACCOUNT" --team-id "$TEAM_ID" --password "$NOTARIZATION_PWD" "$SUBMISSION_ID"
+        STATUS=$(xcrun notarytool log --apple-id "$APPLE_ACCOUNT" --team-id "$TEAM_ID" --password "$NOTARIZATION_PWD" "$SUBMISSION_ID" | tee /dev/stderr | jq --raw-output .status)
+        set -x
```
alopezz: Is this intentionally left here or a debug leftover?

.gitlab/package_build/build_agent_dmg.sh
```diff
@@ -114,16 +114,19 @@ if [ "$SIGN" = true ]; then
...
+        [ "$STATUS" = Accepted ]  # Fail on any other status like Expired, Invalid, Rejected, Unknown, etc.
```
alopezz: Somewhat related to the other comment: if the idea of `set -x` is to log something about this, it might me worth doing something like printing `$STATUS` explicitly or something along the lines of:

```bash
if [ "$STATUS" != Accepted ]; then
    echo "Submission was not accepted, got: ${STATUS}"
    exit 1
fi
```

alopezz (approved):
I left non-blocking comments.

.gitlab/package_build/build_agent_dmg.sh
rdesgroppes: you guessed it right in your next comment:
> [...] the idea of set -x is to log something about this [...]

.gitlab/package_build/build_agent_dmg.sh
rdesgroppes: Here you go: 72755e99494a0336cba7fb0696e2f437bd03a9d4

alopezz (approved):
👍 Thanks!

PR #38568, opened 2025-07-08
dyninst/decoder: values are always strings

grantseltzer:
### What does this PR do?

- According to the capture value schema the "value" field in uploaded json should always be a string. This PR fixes the issue
- Updates test data
- Also fixes an issue with the json encoding strings that have a not captured reason set.

### Motivation

We need to conform to the captured value schema.

Ran integration tests


ajwerner (approved):
LGTM 

We should add a test that has a very big string that we’ll fail to catch (separate PR).

PR #38569, opened 2025-07-08
CXP-2378 [agent] add new config for process check refactor

gengnamstyle:
### What does this PR do?
- adds a new non-public facing configuration `process_config.process_collection.use_wlm` with build flag os separation for unix and disabled for windows
- adds the new check `runWLM` which duplicates the existing `run` code such that future PRs can easily see the difference
- duplicates an e2e test with this configuration
    - unit tests will be added for the new check once new code will be written as copying the existing unit tests would be significantly changed in the next pr (work that would reverted soon)
- adds configuration unit test

### Motivation
- as part of the processes + service discovery consolidation, we need to refactor the process check to use the new WLM process collector. This config will help streamline + automate testing while we finish the refactor. Eventually this config will be replaced by a build flag that only runs this for linux
    - RFC - https://datadoghq.atlassian.net/wiki/spaces/cxg/pages/5013111768/RFC+Processes+and+Service+Discovery+Agent+Consolidation
    - WLM process collector work currently hardcoded to be disabled https://github.com/DataDog/datadog-agent/blob/fbdd1bbd7ee4d151620c6a88c17fcbf1c632cf71/comp/core/workloadmeta/collectors/internal/process/process_collector.go#L139

- run agent locally with debug mode on to see function process collection 
```
2025-07-08 16:10:05 UTC | CORE | DEBUG | (pkg/process/checks/process.go:228 in runWLM) | Running new WLM process check
2025-07-08 16:10:05 UTC | CORE | DEBUG | (pkg/process/checks/process.go:346 in runWLM) | collected processes in 6.766542ms
```
- added e2e test

- config can potentially be enabled for all non-windows platforms 
- however, if an attentive customer decides to enable the config, behavior should stay the same
    - when the refactor is finished, there should be a linux specific build flag file with the new check

### DIFF
--- a/pkg/process/checks/process.go
+++ b/pkg/process/checks/process.go
@@ -87,9 +87,9 @@ type ProcessCheck struct {
 	// determine if zombies process will be collected
 	ignoreZombieProcesses bool
 
-	// TODO: process_config.process_collection.linux_use_wlm is a temporary configuration for refactoring purposes
+	// TODO: process_config.process_collection.use_wlm is a temporary configuration for refactoring purposes
 	// that determines if linux process collection will use WLM
-	useLinuxWLMProcessCollection bool
+	useWLMProcessCollection bool
 
 	hostInfo                   *HostInfo
 	lastCPUTime                cpu.TimesStat
@@ -196,8 +196,8 @@ func (p *ProcessCheck) Init(syscfg *SysProbeConfig, info *HostInfo, oneShot bool
 		p.extractors = append(p.extractors, p.workloadMetaExtractor)
 	}
 
-	// TODO: process_config.process_collection.linux_use_wlm is a temporary configuration for refactoring purposes
-	p.useLinuxWLMProcessCollection = p.config.GetBool("process_config.process_collection.linux_use_wlm")
+	// TODO: process_config.process_collection.use_wlm is a temporary configuration for refactoring purposes
+	p.useWLMProcessCollection = p.useWLMCollection()
 
 	return nil
 }
@@ -357,7 +357,7 @@ func (p *ProcessCheck) runWLM(groupID int32, collectRealTime bool) (RunResult, e
 }
 
 func (p *ProcessCheck) run(groupID int32, collectRealTime bool) (RunResult, error) {
-	if p.useLinuxWLMProcessCollection {
+	if p.useWLMProcessCollection {
 		return p.runWLM(groupID, collectRealTime)
 	}
 
--- a/pkg/process/checks/process.go
+++ b/pkg/process/checks/process.go
@@ -87,9 +87,9 @@ type ProcessCheck struct {
 	// determine if zombies process will be collected
 	ignoreZombieProcesses bool
 
-	// TODO: process_config.process_collection.use_wlm is a temporary configuration for refactoring purposes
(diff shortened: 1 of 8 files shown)


dustmop (approved):

test/new-e2e/tests/process/linux_test.go
```diff
@@ -268,6 +268,39 @@ func (s *linuxTestSuite) TestProcessChecksInCoreAgent() {
 	requireProcessNotCollected(t, payloads, "process-agent")
 }
 
+func (s *linuxTestSuite) TestProcessChecksInCoreAgentWLMProcessCollector() {
```
daniel-taf: These test names are getting a bit long nowadays 😓 Maybe just `TestProcessCheckWLM` is enough since the other stuff is implied

daniel-taf (approved):

pkg/process/checks/process.go
```diff
@@ -87,6 +87,10 @@ type ProcessCheck struct {
 	// determine if zombies process will be collected
 	ignoreZombieProcesses bool
 
+	// TODO: process_config.process_collection.linux_use_wlm is a temporary configuration for refactoring purposes
+	// that determines if linux process collection will use WLM
+	useLinuxWLMProcessCollection bool
```
wiyu: What happens if this is turned on in windows or mac?

gengnamstyle: It should never be turned on for windows and mac since it's just a configuration we will use for testing. In the future, macos could be supported but it's not the priority. Windows will never be supported for this refactor since there is a lack of permissions on the core agent process. 

Eventually this config will be replaced by a linux build-tag file. However, if it is accidentally turned on it will function normally for mac but potentially lead to some errors on windows.

wiyu: Yes, ideally no one turns this on, but it is a possibility. It would be cleaner to just use build tags and have specific OS implementations. 

We did this for `process_config.run_in_core_agent.enabled`. If we go with a simlar approach we wouldn't need to have an env in the config name either. Just call it `process_config.process_collection.use_wlm`. ([agent_linux.go](https://github.com/DataDog/datadog-agent/blob/44bbe071b977bead7b67fee734ea2f54bdbba0d6/comp/process/agent/agent_linux.go#L86-L91) and [agent_fallback.go](https://github.com/DataDog/datadog-agent/blob/44bbe071b977bead7b67fee734ea2f54bdbba0d6/comp/process/agent/agent_fallback.go#L17-L21))

You can achieve the same thing by having a function maybe called `useWLM()` and specify it in process_nix.go or process_windows.go files and have the specific behavior you're looking for in there instead..

gengnamstyle: Just to clarify, did you mean to use build flags to isolate the configuration per OS or remove the configuration and just use build flags? 

I'm interpreting this as use build flags instead of the config, but I believe we would benefit from the config as a safety measure to gate the new check while it is still be tested? 

I think we could do a hybrid approach where I use build flags to separate the specific implementations and use a config in the linux specific file.

gengnamstyle: For now, I used build flags to create a helper for the config check and renamed the configuration

PR #38570, opened 2025-07-08
[USMON-1328] usm: go-tls: Introduce standalone method to extract conn_tuple from TLSConn

guyarb:
### What does this PR do?

Replaces the method to extract conn_tuple from a TLSConn pointer in our GoTLS monitoring.

The old method extracted the system `fd` belongs with the `TLSConn`, and relied on a preliminary mapping to the actual socket, which was done via a kprobe on `sockfd_lookup_light` method.

However, from kernel 6.11, the method was removed from the kernel and our TLs monitoring fails on such a kernel.

The new method extracts the conn_tuple directly via reading the IP (local and remote) and ports from the TLSConn struct.

Instead of extrating the system `fd` from `netFD` struct, and relying on external mapping, we simply analyze and process the `laddr`, `raddr` and `family` fields, extracting the relevant information.

### Motivation

Supporting kernels 6.11 and later

Simplifying the GoTLS monitoring making it standalone

### Additional Notes

The structs we're dealing with

```
type TCPConn struct {
	conn
}

type conn struct {
	fd *netFD
}

type netFD struct {
	pfd poll.FD

	// immutable until Close
	family      int
	sotype      int
	isConnected bool // handshake completed or use of association with peer
	net         string
	laddr       Addr
	raddr       Addr
}

type Addr interface {
	Network() string 
	String() string
}

type TCPAddr struct {
	IP   IP
	Port int
	Zone string // IPv6 scoped addressing zone
}

type IP []byte
```


pkg/network/ebpf/c/protocols/tls/go-tls-conn.h
```diff
@@ -9,74 +9,151 @@
...
+    if (address_ptr->len == IPV6_ADDR_LEN && family == AF_INET6) {
+        if (bpf_probe_read_user(ip, IPV6_ADDR_LEN, (void*)address_ptr->ptr)) {
+            log_debug("[go-tls-conn] failed to read IPv6 address at %p", (void*)address_ptr->ptr);
+            return false;
+        }
+        *out_address_h = *((__u64*)ip);
+        *out_address_l = *((__u64*)(ip + 8));
+        return true;
+    }
+    log_debug("[go-tls-conn] invalid address length: %llu", address_ptr->len);
```
DanielLavie: Invalid family is also an option here

pkg/network/ebpf/c/protocols/tls/go-tls-conn.h
```diff
@@ -9,74 +9,151 @@
...
     void* conn_fd_ptr;
     if (bpf_probe_read_user(&conn_fd_ptr, sizeof(conn_fd_ptr), conn_fd_ptr_ptr)) {
-        return NULL;
+        return false;
     }
 
-    void* net_fd_pfd_ptr = conn_fd_ptr + cl->net_fd_pfd_offset;
-    // the internal/poll.FD struct is embedded in net.netFD, so just add the offset again
-    void* fd_sysfd_ptr = net_fd_pfd_ptr + cl->fd_sysfd_offset;
+    // Change here
```
DanielLavie: What does this comment means?

DanielLavie (changes requested):

DanielLavie:
Can you explain how you verified that these fields are populated across all supported kernel versions?

guyarb:
> Can you explain how you verified that these fields are populated across all supported kernel versions?

Added dedicated tests that run with and without TLS, on IPv4 and IPv6 - https://github.com/DataDog/datadog-agent/pull/38546
Those tests are running on various kernels to guarantee backward compatibility

Deployed on `oddish-c` for dogfooding and gathering the final results

pkg/network/ebpf/c/protocols/tls/go-tls-conn.h
guyarb: It was a reminder for me to read the `family` field

removed

pkg/network/ebpf/c/protocols/tls/go-tls-conn.h
guyarb: Added a clarification

DanielLavie (approved):

PR #38573, opened 2025-07-08
Expose Autodiscovery filter errors in status

gabedos:
### What does this PR do?

Exposes parsing and creation errors for Autodiscovery related filters from the workloadFilter component

### Motivation

Migrate Autodiscovery to use workloadfilter component instead of pkg/util/containers filters

[CONTP-845]

Deploy the agent locally with a few configuration errors:

```
datadog:
  kubelet:
    tlsVerify: false
  envDict:
    DD_CONTAINER_EXCLUDE_METRICS: "kube_namespace:datadog-agent invalid_key_gabe:meow"
    DD_CONTAINER_INCLUDE_LOGS: "name:cluster-agent bad_key_gabe:woof"
```

```
k exec -it datadog-agent-linux-xxxxx -n datadog-agent -- agent status
...
=============
Autodiscovery
=============

  Enabled Features
  ================
    docker
    kube_orchestratorexplorer
    kubernetes

  Container Inclusion/Exclusion Errors
  ====================================
    container filter bad_key_gabe:woof is unknown, ignoring it. The supported filters are 'image', 'name' and 'kube_namespace'
    container filter invalid_key_gabe:meow is unknown, ignoring it. The supported filters are 'image', 'name' and 'kube_namespace'
```

N/A

### Additional Notes

[CONTP-845]: https://datadoghq.atlassian.net/browse/CONTP-845?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/comp/core/workloadfilter/catalog/utils.go
+++ b/comp/core/workloadfilter/catalog/utils.go
@@ -22,12 +22,12 @@ import (
 func createProgramFromOldFilters(oldFilters []string, objectType workloadfilter.ResourceType) (cel.Program, error) {
 	filterString, err := convertOldToNewFilter(oldFilters, objectType)
 	if err != nil {
-		return nil, fmt.Errorf("error converting filters: %w", err)
+		return nil, err
 	}
 
-	program, progErr := createCELProgram(filterString, objectType)
-	if progErr != nil {
-		return nil, fmt.Errorf("error creating CEL filtering program: %w", progErr)
+	program, err := createCELProgram(filterString, objectType)
+	if err != nil {
+		return nil, err
 	}
 
 	return program, nil
@@ -58,7 +58,6 @@ func createCELProgram(rules string, objectType workloadfilter.ResourceType) (cel
 // getFieldMapping creates a map to associate old filter prefixes with new filter fields
 func getFieldMapping(objectType workloadfilter.ResourceType) map[string]string {
 	return map[string]string{
-		"id":    fmt.Sprintf("%s.id.matches", objectType),
 		"name":  fmt.Sprintf("%s.name.matches", objectType),
 		"image": fmt.Sprintf("%s.image.matches", objectType),
 		"kube_namespace": func() string {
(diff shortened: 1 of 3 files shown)


comp/core/workloadfilter/catalog/utils.go
```diff
@@ -22,12 +22,12 @@ import (
...
 	filterString, err := convertOldToNewFilter(oldFilters, objectType)
 	if err != nil {
-		return nil, fmt.Errorf("error converting filters: %w", err)
+		return nil, err
 	}
 
-	program, progErr := createCELProgram(filterString, objectType)
-	if progErr != nil {
-		return nil, fmt.Errorf("error creating CEL filtering program: %w", progErr)
+	program, err := createCELProgram(filterString, objectType)
```
clamoriniere, line 28: here you are redefining the `err error` 

what you can do is to create the program variable before
 ```suggestion
    var program cel.Program
	if program, err = createCELProgram(filterString, objectType); err != nil {
	  return nil, err
	}
```

gabedos: It should be fine to have the error variable re-defined considering that we don't care about keeping the error value. It's immediately returned if it's a real error. We do it in various locations throughout the codebase and it makes everything cleaner. Lmk if you still think this is a bad pattern.

For example:

https://github.com/DataDog/datadog-agent/blob/1cfd9581020dda6e667249a19a87eaea35215b84/comp/core/autodiscovery/providers/consul.go#L218-L243

https://github.com/DataDog/datadog-agent/blob/1cfd9581020dda6e667249a19a87eaea35215b84/comp/core/workloadmeta/collectors/internal/containerd/container_builder.go#L40-L110

clamoriniere: No, it looks good as-is for this part of the code, and I agree that in this specific case it doesn’t impact anything.
I thought we had a linter that would catch this, but maybe it’s now smart enough to only report actual issues.

My comment was more about making the code future-proof, in case someone modifies the error-handling logic in this function later on.

clamoriniere (approved):

PR #38578, opened 2025-07-08
[AGTMETRICS-243] Stream config updates to `RemoteAgents` in `RemoteAgentRegistry`

rayz:
### What does this PR do?
This PR introduces the implementation for streaming configuration updates to remote agents.

### Motivation
This will allow us to implement RC based changes on remote-agents like ADP.

1. Testing wiith ADP ([PR](https://github.com/DataDog/saluki/pull/797))
2. Testing with the `internal/remote-agent` example


comp/core/remoteagentregistry/impl/remoteagentregistry.go
```diff
@@ -473,6 +521,63 @@ collect:
...
+		return err
+	}
+
+	return nil
+}
+
+func (ra *remoteAgentRegistry) createConfigSnapshot() (*pb.ConfigEvent, error) {
+	allSettings, sequenceID := ra.conf.AllSettingsWithSequenceID()
+
+	// Note: AllSettings returns a map[string]interface{}. The inner values may be a type that structpb.NewValue is not able to handle (ex: map[string]string), so we perform a hacky operation by marshalling the data into a JSON string first.
```
rayz: Without doing this, we get errors while using `structpb.NewValue`

```
Failed to convert setting 'network_devices' to structpb.Value: proto: invalid type: []string
2025-07-09 12:27:36 EDT | CORE | WARN | (comp/core/remoteagentregistry/impl/remoteagentregistry.go:558 in createConfigSnapshot) | Failed to convert setting 'check_system_probe_startup_time' to structpb.Value: proto: invalid type: time.Duration
...
```

comp/core/remoteagentregistry/impl/remoteagentregistry.go
```diff
@@ -473,6 +510,82 @@ collect:
...
+	ra.agentMapMu.Lock()
+	defer ra.agentMapMu.Unlock()
+
+	pbValue, err := structpb.NewValue(update.newValue)
+	if err != nil {
+		log.Warnf("Failed to convert setting '%s' to structpb.Value: %v", update.setting, err)
+		return
+	}
+
+	source := ra.conf.GetSource(update.setting).String()
```
misteriaud: **💬 suggestion:** ‏We maybe should update the configuration component `OnUpdate()` method to provide this the value origin directly in the callback rather than having to call `GetSource()` every time.
What do you think @hush-hush @dustmop ?

comp/core/remoteagentregistry/impl/detail.go
```diff
@@ -0,0 +1,72 @@
...
+	client, err := newRemoteAgentClient(registration)
+	if err != nil {
+		return nil, err
+	}
+
+	return &remoteAgentDetails{
+		displayName:  registration.DisplayName,
+		apiEndpoint:  registration.APIEndpoint,
+		client:       client,
+		configStream: nil,
```
misteriaud: **❓ question:** ‏Do we really want to decouple remoteAgent creation and configStream creation. It seems that in any case if `startConfigStream()` fail, `RegisterRemoteAgent()` will dismiss the registration of the new remoteAgent.
Avoiding having `nil` value for `configStream` will avoid checking in every place if the configStream have been initialized (such as in `handleConfigUpdate()` function).

comp/core/remoteagentregistry/impl/remoteagentregistry.go
```diff
@@ -473,6 +510,82 @@ collect:
...
+		if err != nil {
+			log.Warnf("Failed to convert setting '%s' to structpb.Value: %v", setting, err)
+			continue
+		}
+		settings = append(settings, &pb.ConfigSetting{
+			Source: source,
+			Key:    setting,
+			Value:  pbValue,
+		})
+	}
```
misteriaud, line 688: **❓ question:** ‏Are `setting` each layer of the configuration ?

comp/core/remoteagentregistry/impl/remoteagentregistry.go
```diff
@@ -496,23 +609,83 @@ func newRemoteAgentClient(registration *remoteagentregistry.RegistrationData) (p
...
+
+		// Start by sending an initial snapshot of the current configuration.
+		//
+		// We do this to ensure that when we restart this outer loop, we always resynchronize the remote agent by
+		// providing a complete snapshot of the current configuration. This lets us handle any errors during send
+		// by just restarting the outer loop.
+		initialSnapshot, sequenceID, err := createConfigSnapshot(config)
+		if err != nil {
+			log.Errorf("Failed to create initial config snapshot: %v", err)
+			time.Sleep(1 * time.Second)
```
misteriaud: **💬 suggestion:** ‏We maybe should make the wait until retry duration configurable.

comp/core/remoteagentregistry/impl/remoteagentregistry.go
```diff
@@ -496,23 +609,83 @@ func newRemoteAgentClient(registration *remoteagentregistry.RegistrationData) (p
...
+outer:
+	for {
+		lastEventSequenceID := uint64(0)
+
+		// Start by sending an initial snapshot of the current configuration.
+		//
+		// We do this to ensure that when we restart this outer loop, we always resynchronize the remote agent by
+		// providing a complete snapshot of the current configuration. This lets us handle any errors during send
+		// by just restarting the outer loop.
+		initialSnapshot, sequenceID, err := createConfigSnapshot(config)
```
misteriaud: **💭 thought:** ‏generating the initial snapshot on every loop can be costly for the coreAgent in the case the remoteAgent does not reply. We should maybe cache it and recreate it only if modification have been made to the configuration (I didn't check if  `conf.AllSettingsWithSequenceID()` have a caching mechanism).

misteriaud (commented):
Nice work ! Overall LGTM, I left some comments.

comp/core/remoteagentregistry/impl/remoteagentregistry.go
rayz: `setting` is the full key path of a key.  (this is constructed in the `AllSettings` implementation i believe)

comp/core/remoteagentregistry/impl/remoteagentregistry.go
rayz: Will do 👍

comp/core/remoteagentregistry/impl/remoteagentregistry.go
rayz: True, I can add a check to see if the `sequenceID` has changed or something similar.

comp/core/remoteagentregistry/impl/remoteagentregistry.go
GustavoCaso: I agree with this suggestion 😄

comp/core/remoteagentregistry/impl/detail.go
```diff
@@ -0,0 +1,72 @@
...
+	}, nil
+}
+
+func (rad *remoteAgentDetails) startConfigStream(config config.Component) error {
+	if rad.configStream != nil {
+		return errors.New("config stream already started")
+	}
+
+	ctx, ctxCancel := context.WithCancel(context.Background())
+	configUpdates := make(chan *pb.ConfigUpdate, 8)
```
GustavoCaso, line 55: How did we arrive at the value eight as the channel size?

comp/core/remoteagentregistry/impl/detail.go
GustavoCaso: I agree with this suggestion 😄

pkg/proto/datadog/remoteagent/remoteagent.proto
```diff
@@ -84,3 +87,29 @@ message GetTelemetryResponse {
...
+
+message ConfigSetting {
+  string source = 1;
+  string key = 2;
+  google.protobuf.Value value = 3;
+}
+
+message ConfigSnapshot {
+  string origin = 1;
+  string source = 2;
```
GustavoCaso: We are not using this value, and it might be redundant as the source is at the setting level. Most probably it was an oversight from the RFC, but I believe we can remove it

comp/core/remoteagentregistry/impl/remoteagentregistry.go
```diff
@@ -496,23 +609,83 @@ func newRemoteAgentClient(registration *remoteagentregistry.RegistrationData) (p
...
+}
+
+func (cs *configStream) TrySendUpdate(update *pb.ConfigUpdate) bool {
+	select {
+	case cs.configUpdates <- update:
+		return true
+	default:
+		return false
 	}
+}
```
GustavoCaso: I'm wondering if we should put some timer here in case the `configUpdates` channel is full, instead of dropping the update right away. I say this because if we drop the update, it would cause the `runConfigStream` loop to recalculate the snapshot. 
```go
if currentSequenceID > lastEventSequenceID+1 {
	continue outer
}
```

GustavoCaso (commented):
Left a few comments, but overall this looks great. 
One nitpick I have, is that we have split the remoteagentregistry.go into two file now. I like the idea of having multiple files as it makes skimming easier, but I find some of the organization a bit confusing. For example we could define the `configStream` strcut in the `detail.go` and their functions. This suggestion is purely cosmetic 😄

comp/core/remoteagentregistry/impl/detail.go
rayz: Updated the code to create `configStream` when `RemoteAgentDetails` is created.

comp/core/remoteagentregistry/impl/remoteagentregistry.go
rayz: I've added a timer of `250 ms`. Open to suggestions for a more appropriate value but, something to consider is that the `agentMap` lock is held during this time.

comp/core/remoteagentregistry/impl/detail.go
rayz: It was just a placeholder value i put 😅 . If we expect configuration updates to occur in large bursts I can update this be a larger value.

comp/core/remoteagentregistry/impl/remoteagentregistry.go
rayz: PR https://github.com/DataDog/datadog-agent/pull/39170

comp/core/remoteagentregistry/impl/detail.go
misteriaud: I wonder how it works when a user updates its whole configuration from fleet automation for example, does the diff is calculated in the coreAgent, and only the updated fields are transmitted to remoteAgents ? If this is the case, we could see potential burst if the user changes a big portion of its config.

GustavoCaso (approved):
Awesome work 🎉 

Regarding your comment: 
> I've added a timer of 250 ms. Open to suggestions for a more appropriate value but, something to consider is that the agentMap lock is held during this time.

We might benefit from emitting some internal telemetry data using the telemetry component, to get a sense of how long does it take process config events among the different remote agent, or other useful telemetry data, such as failure etc. 

This could be done on a separate PR 😄

dustmop (approved):

PR #38587, opened 2025-07-09
Fix regex to disable OTel test on fips pipeline

KevinFairise2:
### What does this PR do?

The auto retries of e2e tests changed a bit the expected regex. We should now use a real regex to skip test for OTel in fips pipelines


KevinFairise2:

KevinFairise2:

songy23 (approved):

chouetz (approved):

PR #38590, opened 2025-07-09
Fix wrong logic in `filter_only_leaf_tests` function used in e2e test retries

Ishirui:
### What does this PR do?

Fixes two bugs:
- The `sorted_tests` list in `filter_only_leaf_tests` was sorted in ascending order instead of the expected descending, causing the function to return unexpected results. This in turn caused more tests to be selected for retrial than needed, causing lost time in CI. See [this job run](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1016592569) for example
- We pass the `attempt_number` argument to the call to `test_flavor` - without this parameter being passed, the default value of `0` is always used. In turn, this makes it so that the filename for the JUnit file containing all test results is the same across all retries, so it gets overwritten. By passing the attempt number we avoid this overwrite and so correctly report all test runs to DataDog.

Additionally add a couple more unit tests and refactor the arguments to the `_is_child` function.

### Motivation

Make sure #37862 works properly !

- Unit test for `filter_only_leaf_tests`
- Force failure of a test (fc0664e2e5) and run CI, check that everything happens as expected (only this test is retried + we get the right junit files in the artifact, and we see failing jobs in Datadog UI)

N/A

### DIFF
--- a/tasks/libs/testing/e2e.py
+++ b/tasks/libs/testing/e2e.py
@@ -26,12 +26,14 @@ def filter_only_leaf_tests(tests: Iterable[tuple[str, str]]) -> set[tuple[str, s
     Given some (package, test_name) tuples, return only the leaf tests.
     A test is a leaf if it is not a parent of any other test in the list (within the same package).
     """
-    # Sort tests by descending depth (number of '/' in test name)
-    tests_sorted = sorted(tests, key=lambda t: len(t[1].split('/')))
+    # Sort tests by depth (number of '/' in test name) - deepest tests first
+    tests_sorted = sorted(tests, key=lambda t: len(t[1].split('/')), reverse=True)
     leaf_tests: set[tuple[str, str]] = set()
     for candidate_test in tests_sorted:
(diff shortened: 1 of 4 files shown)


Ishirui:
Test result after [job run](https://gitlab.ddbuild.io/datadog/datadog-agent/builds/1020340803):
- Problem 1 fixed, only the falsely failing job was retried
- Problem 2 not entirely fixed: the junit file *was* named appropriately, but the [job artifact]() still only contains one junit file instead of the 3 we expected. Accordingly, [only a small number of test runs were reported to Datadog](https://app.datadoghq.com/ci/test/runs?query=test_level%3Atest%20%40ci.job.url%3A%22https%3A%2F%2Fgitlab.ddbuild.io%2FDataDog%2Fdatadog-agent%2F-%2Fjobs%2F1020340803%22%20%40test.service%3Adatadog-agent&agg_m=count&agg_m_source=base&agg_t=count&fromUser=false&index=citest&start=1751463406591&end=1752068206591&paused=false) -> **More investigation required**

Ishirui:
Split into #38701 and #38704

PR #38592, opened 2025-07-09
Unify noop stream compressor

vickenty:
### What does this PR do?

Make the noop compressor implementation return an instance of the already existing `NoopStreamCompressor` instead of nil, and remove corresponding nil checks.

This changes metrics serializer to use streaming json serializer when used on serverless (the only flavor without any compression built in) and disabled v2 api.

### Motivation

Remove obstacles for removing non-streaming v1 json serialization from the metrics serializer.

Tests.

### DIFF
--- a/pkg/util/compression/impl-noop/no_strategy.go
+++ b/pkg/util/compression/impl-noop/no_strategy.go
@@ -40,7 +40,7 @@ func (s *NoopStrategy) ContentEncoding() string {
 	return "identity"
 }
 
-// NewStreamCompressor returns a nil when there is no compression implementation.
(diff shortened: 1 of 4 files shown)


pkg/util/compression/impl-noop/no_strategy.go
```diff
@@ -41,6 +41,21 @@ func (s *NoopStrategy) ContentEncoding() string {
 }
 
 // NewStreamCompressor returns a nil when there is no compression implementation.
```
pgimalac: Comment should be updated I guess

pgimalac (approved):

rayz (approved):

jszwedko (approved):
👍 this seems cleaner to me too

PR #38595, opened 2025-07-09
[CWS] Fix CI job not running any host-based e2e tests

YoannGh:
### What does this PR do?

Prior to https://github.com/DataDog/datadog-agent/pull/37862, the `--run TestAgentSuite` parameter was directly passed to gotestsum which matched both the `TestAgentSuiteEC2` and `TestAgentSuiteGCP` go test suites.
However this PR added extra logic that turns this parameter into the following regexp `"^TestAgentSuite$"`, which doesn't match any go test.

This PR fixes this by explicitly matching each testsuite name that the job should run.

### DIFF
--- a/.gitlab/e2e/e2e.yml
+++ b/.gitlab/e2e/e2e.yml
@@ -480,7 +480,7 @@ new-e2e-cws:
     # ON_NIGHTLY_FIPS: "true"
   parallel:
     matrix:
-      - EXTRA_PARAMS: --run TestAgentSuite
+      - EXTRA_PARAMS: --run TestAgentSuite(EC2|GCP)
       - EXTRA_PARAMS: --run TestECSFargate
       - EXTRA_PARAMS: --run TestKindSuite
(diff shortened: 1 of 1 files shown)


Ishirui (approved):
Good catch ! Thanks for being so attentive.
Do you know of any other instances where there might be similar problems so I can try to pre-emptively fix them ?

YoannGh:
> Good catch ! Thanks for being so attentive. Do you know of any other instances where there might be similar problems so I can try to pre-emptively fix them ?

I quickly went through the other job definitions of the e2e stage and I don't think there is (it might be worth to double check though)

Ishirui:
> > Good catch ! Thanks for being so attentive. Do you know of any other instances where there might be similar problems so I can try to pre-emptively fix them ?
> 
> I quickly went through the other job definitions of the e2e stage and I don't think there is (it might be worth to double check though)

Got it, thanks for having a look. I'll double check as you suggested 👍

PR #38599, opened 2025-07-09
[FA] Add config.source file path to integrations

coignetp:
### What does this PR do?

Add config.source file path in the integrations payload
Example in the checks.json payload
```json
"memory": [
            {
                "config.hash": "memory",
                "config.provider": "file",
                "config.source": "/opt/datadog-agent/etc/conf.d/memory.d/conf.yaml.default",
                "init_config": "",
                "instance_config": "{}"
            }
        ],
```

### Motivation

Being able to display which integration config comes from which file in Fleet Automation UI.
Potentially use it to override integrations.


Stanislas167 (approved):

pkg/collector/check/metadata.go
```diff
@@ -19,7 +19,14 @@ func GetMetadata(c Info, includeConfig bool) map[string]interface{} {
 
 	instanceID := string(c.ID())
 	instance["config.hash"] = instanceID
-	instance["config.provider"] = strings.Split(c.ConfigSource(), ":")[0]
+
+	splitSource := strings.Split(c.ConfigSource(), ":")
+	instance["config.provider"] = splitSource[0]
```
Hitsuji-M: ❓ Question : Whats the value like if it does not have a file path (provider not `file`) ?

BaptisteFoy (approved):

pkg/collector/check/metadata.go
coignetp: It can be `container` or `kube_endpoints` for instance

pkg/jmxfetch/state.go
```diff
@@ -126,6 +127,12 @@ func GetIntegrations() (map[string]interface{}, error) {
 			instances = append(instances, GetJSONSerializableMap(rawInstanceConfig).(integration.JSONMap))
 		}
 
+		splitSource := strings.SplitN(config.Source, ":", 2)
+		if len(splitSource) > 1 {
+			c["config.source"] = splitSource[1]
+		} else {
+			c["config.source"] = "unknown"
```
remeh: It is sure that `if len(splitSource) == 1` splitSource doesn't contain anything useful to use instead of defaulting to "unknown"?
I guess you also want to avoid cardinality issues 🤔

remeh (approved):

pkg/jmxfetch/state.go
coignetp: When `len(splitSource) == 1`, the full string is already being sent to the backend, and I haven’t identified any additional delimiter to split on - especially for file-based integrations. If we eventually decide there’s something worth extracting in this case, we can always do the parsing backend-side because the complete data is already there

vickenty: It seems this code is repeated in the collector. Would it make sense to abstract it as a method on `integration.Config`?

Also, I think @remeh is right, and you want the full thing, not just the part after the colon. `config.Source` looks like an urn and can be many other things than a file path.

coignetp: Sure I can abstract it

This part is for JMX integrations only, and we're already setting the `config.provider` as `file` [here](https://github.com/DataDog/datadog-agent/pull/38599/files#diff-257ff1757b8bd8c5271bdd8db247f2d83224795cf6dd165a91ac049a1c9aaa04L55).
We can definitely change it and use `splitSource[0]` instead. What do you think?

vickenty: > Sure I can abstract it

If that's not too much work, please do.

> This part is for JMX integrations only, and we're already setting the config.provider as file [here](https://github.com/DataDog/datadog-agent/pull/38599/files#diff-257ff1757b8bd8c5271bdd8db247f2d83224795cf6dd165a91ac049a1c9aaa04L55).
We can definitely change it and use splitSource[0] instead. What do you think?

I think that `config.provider` isn't correct if the JMX integration is configured via container tags, which they can be. You may end up with `config.provider=file` and `config.source=containerd://sha1`. It's entirely up to you though, I don't have a strong opinion here.

coignetp: After taking a look at it, it seems I would need to add the method to the `check.Info` and `check.Check` interface, meaning that the code will be duplicated in more places ; or add a function in another package that would need to be imported in both these packages ; and I don't know if it's a path we want to take

> I think that config.provider isn't correct if the JMX integration is configured via container tags, which they can be. You may end up with config.provider=file and config.source=containerd://sha1. It's entirely up to you though, I don't have a strong opinion here.

I've updated to code to get the provider correctly

rahulkaukuntla (approved):

jeremy-hanna (approved):

pkg/jmxfetch/state.go
vickenty: > After taking a look at it, it seems I would need to add the method to the check.Info and check.Check interface

Just this, tucked somewhere in `integration/util`,  in  is enough already: 

```go
func ConfigSourceToMetadataMap(source string, instance map[any]any) {
	splitSource := strings.SplitN(source, ":", 2)
	instance["config.provider"] = splitSource[0]
	if len(splitSource) > 1 {
		instance["config.source"] = splitSource[1]
	} else {
		instance["config.source"] = "unknown"
	}
}
```

Or, viewed from another angle, maybe this splitting can live in the metadata component, since it's the consumer of this data. Then you can trust that `config.provider` is a string and make your code a bit easier.

vickenty (approved):

wdhif (commented):
Approved for @DataDog/container-platform file

wdhif (approved):

PR #38612, opened 2025-07-09
[EBPF-666] gpu: support cgroups v2 for device permissions

gjulianm:
### What does this PR do?

Adds support for cgroup v2 in our permission patching module of system-probe, used to ensure access to GPU devices.

### Motivation

Allow deploying the GPUM solution in environments where k8s does not correctly support GPU monitoring apps.

* Added unit tests for both cgroupv1 and cgroupv2. Our KMT tests cover systems with both.
* Deployed in staging for validation
* Manually tested in GKE environments

### Additional Notes

Benchmark results: 

``` 
goos: linux
goarch: arm64
pkg: github.com/DataDog/datadog-agent/pkg/gpu
=== RUN   BenchmarkGetAbsoluteCgroupForProcess
BenchmarkGetAbsoluteCgroupForProcess
    /home/vagrant/dd/datadog-agent/pkg/gpu/cgroups_test.go:440: Created 111110 directories
BenchmarkGetAbsoluteCgroupForProcess-8                 2         623920872 ns/op            111110 dirs/op      99765684 B/op   1166753 allocs/op
PASS
ok      github.com/DataDog/datadog-agent/pkg/gpu        5.519
```
 
~600ms for the complete scan of a directory sturcture with ~100k cgroups.


pkg/gpu/cgroups.go
```diff
@@ -8,97 +8,344 @@
...
 		return fmt.Errorf("failed to write to %s: %w", deviceAllowPath, err)
 	}
 
 	return nil
 }
 
+// detachAllDeviceCgroupPrograms finds and detaches all device cgroup BPF programs from a cgroup
+// cgroupName is the name of the cgroup, e.g. "/kubepods.slice"
+// rootfs is the rootfs where /sys/fs/cgroup is mounted
+func detachAllDeviceCgroupPrograms(cgroupName, rootfs string) error {
```
val06: nit: for consistency, use the same order of params as in `configureCgroupV1DeviceAllow` (rootfs, cgroupName string)

pkg/gpu/cgroups.go
```diff
@@ -8,97 +8,344 @@
...
+		}
+
+		// Detach the program
+		err = link.RawDetachProgram(link.RawDetachProgramOptions{
+			Target:  int(cgroup.Fd()),
+			Program: program,
+			Attach:  ebpf.AttachCGroupDevice,
+		})
+		if err != nil {
+			program.Close()
```
val06: instead, use `defer program.Close` after successful call to `NewProgramFromID`, you close it in both error and success flow anyway

pkg/gpu/cgroups.go
```diff
@@ -8,97 +8,344 @@
...
+
+	containerCgroupInode := stat.Ino
+
+	// Now, walk through the host cgroup tree, looking for a cgroup with the same inode
+	// as the container cgroup.
+	// TODO: We could use the pkg/util/cgroup package but it doesn't detect correctly the host
+	// cgroup mountpoint, once that's fixed we can use it instead of walking the cgroup tree.
+	var hostCgroupPath string
+	rootSysFsCgroup := filepath.Join(rootfs, "/sys/fs/cgroup")
+	err = filepath.WalkDir(rootSysFsCgroup, func(path string, d fs.DirEntry, err error) error {
```
val06, line 147: are we concerned with some edge cases where certain hosts will contain A LOT of cgroups?

val06: is this a recursive traversal? if yes, perhaps you want to add the path to the error inside the handler function instead of just returning the error

pkg/gpu/cgroups.go
```diff
@@ -8,97 +8,344 @@
...
+	_, err = deviceAllowFile.WriteString(cgroupAllowLine)
 	if err != nil {
 		return fmt.Errorf("failed to write to %s: %w", deviceAllowPath, err)
 	}
 
 	return nil
 }
 
+// detachAllDeviceCgroupPrograms finds and detaches all device cgroup BPF programs from a cgroup
+// cgroupName is the name of the cgroup, e.g. "/kubepods.slice"
```
val06, line 303: nit: is that a real name of a cgroup?

pkg/gpu/cgroups.go
```diff
@@ -8,97 +8,344 @@
...
+
+	// Insert the nvidiaSystemdDeviceAllow line after [Scope]
+	newLines, err := insertAfterSection(lines, "[Scope]", nvidiaSystemdDeviceAllow)
+	if err != nil {
+		return fmt.Errorf("failed to insert device allow line in %s: %w", configFilePath, err)
+	}
+
+	// Write the modified content back to the file
+	newContent := strings.Join(newLines, "\n")
+	err = os.WriteFile(configFilePath, []byte(newContent), 0644)
```
val06: shouldn't we use the same permissions as the original file?

pkg/gpu/cgroups.go
```diff
@@ -8,97 +8,344 @@
...
+			return fmt.Errorf("failed to build path for systemd service file: %w", err)
+		}
+
+		if _, err := os.Stat(configFilePath); errors.Is(err, os.ErrNotExist) {
+			return fmt.Errorf("failed to find either a SystemD configuration directory or configuration file for container %s", containerID)
+		}
+	}
+
+	// Read the entire file
+	content, err := os.ReadFile(configFilePath)
```
val06: are we concerned of memory spikes? can this file be very large ?

pkg/gpu/cgroups.go
```diff
@@ -8,97 +8,344 @@
...
+	// Add leading slash, as that's removed by filepath.Rel and cgroup names
+	// should always have that
+	return "/" + hostCgroupPath, nil
+}
 
-func configureDeviceAllow(containerID, rootfs string, devType deviceType) error {
-	var deviceAllowPath string
-	var err error
-	var allowString string
+// insertAfterSection finds a section header in the lines and inserts the new line after it
```
val06: can you elaborate more? what are those sections?

pkg/gpu/cgroups.go
```diff
@@ -8,97 +8,344 @@
...
+			return "", fmt.Errorf("impossible situation: got a relative path for our own cgroup %s for pid %d", cgroupPath, pid)
+		}
+
+		currentCgroup, err := getAbsoluteCgroupForProcess(rootfs, currentPid)
+		if err != nil {
+			return "", fmt.Errorf("failed to get current process (pid=%d) cgroup: %w", currentPid, err)
+		}
+
+		return filepath.Join(currentCgroup, cgroupPath), nil
+	}
```
val06, line 129: should we sanitize the path to avoid path traversal, (`cgroupPath = "../../../etc/passwd"`)

pkg/gpu/cgroups_test.go
```diff
@@ -0,0 +1,337 @@
...
+	require.NoError(t, configureCgroupV1DeviceAllow("", testCgroupName, 1))
+
+	// Test that /dev/null is now accessible
+	f, err = os.Open("/dev/null")
+	require.NoError(t, err)
+	f.Close()
+
+}
+
+func TestGetAbsoluteCgroupForProcess(t *testing.T) {
```
val06, line 274: the function `getAbsoluteCgroupForProcess` is very long with recursion, perhaps we should test it more thoroughly

val06 (commented):
reviewed

pkg/gpu/cgroups.go
gjulianm: This is used just as a "name" for the cgroup, the safety checks are present later when we use that name to build a path.

pkg/gpu/cgroups.go
gjulianm: A little bit, but it's a one-time operation so at most it will delay the startup of system-probe.

gjulianm: Also added the error with a better description.

pkg/gpu/cgroups.go
gjulianm: Added the comment to explain.

pkg/gpu/cgroups.go
gjulianm: Not really, SystemD config files have a handful of lines only, and these specific ones are created by containerd so we don't expect arbitrary content here.

pkg/gpu/cgroups.go
gjulianm: Changed the code to only open the file once, that way we always retain permissions

pkg/gpu/cgroups.go
gjulianm: I think so, I've seen it in a cgroupn hierarchy.

pkg/gpu/cgroups.go
gjulianm: Done, also moved that code to a separate function to make it cleaner

pkg/gpu/cgroups.go
gjulianm: Fixed, there was actually a typo in the argument order for cgroupv1

pkg/gpu/cgroups.go
val06: can we add a benchmark test with a very large dir?
i am not sure about all the consequences but the delay in start can cascade into other issues (like sp socket access error, or perhaps some probes deployment)

gjulianm: I added a benchmark with 100k directory entries, it takes ~2 seconds to scan all of that.

pkg/gpu/cgroups_test.go
gjulianm: I tried to write a test that simulated better the container situation, but setting all that in a way that the test made sense was complicated and seemed too fragile. Let me see if I can mock this in another way

gjulianm: Added a better test with a filesystem based mock

val06: i still think this tests just the happy flow, perhaps we want to add some edge cases?

val06 (approved):

PR #38613, opened 2025-07-09
[AGENTRUN-536] Fix FIPS status for DCA

misteriaud:
### What does this PR do?
This PR adds a DCA version of the `pkg/fips` package.

### Motivation
This change will return the correct value of `fips.Enabled()` and `fips.Status()` when used with FIPS ClusterAgent. 

Verified on a local Kind cluster using both standard and FIPS-compliant images, ensuring that for both the Agent and the Cluster Agent:
- FIPS images report `FIPS Mode: enabled` in the status
- Standard images report `FIPS Mode: not available` in the status


pgimalac (approved):

pkg/fips/fips_nix.go
```diff
@@ -3,7 +3,7 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2016-present Datadog, Inc.
 
-//go:build goexperiment.systemcrypto && !windows && !serverlessfips
+//go:build (goexperiment.systemcrypto || goexperiment.boringcrypto) && !windows && !serverlessfips
```
jeremy-hanna: [serverless FIPS build](https://github.com/DataDog/datadog-lambda-extension/blob/5be732d275652382919645017d0f964259924057/images/Dockerfile.go_agent.compile#L45) also will have `goexperiment.boringcrypto` set

I believe they use the separate pkg/fips/fips_serverless.go for [the "crypto/tls/fipsonly" import](https://github.com/DataDog/datadog-agent/blob/8ee30378ffe345ec0faca3cbd8d1467397af071d/pkg/fips/fips_serverless.go#L14) which if DCA is not doing, that is a good place for it

jeremy-hanna: possibly we could repurpose [pkg/fips/fips_serverless](https://github.com/DataDog/datadog-agent/blob/8ee30378ffe345ec0faca3cbd8d1467397af071d/pkg/fips/fips_serverless.go#L14) to be a  `pkg/fips/boringssl` and change that build tag there?

michaelcretzman (approved):

jeremy-hanna (approved):

PR #38621, opened 2025-07-09
Fix gates artifacts being overwritten by the ancestor's and smooth exception process

Pythyu:
### What does this PR do?
This PR prevents `static_gate_report.json` artifacts from being overwriten by the one from the ancestor when generating relative values.
It also saves the computed relative size inside of the gitlab artifacts to allow a smoother exception process that only require one file.

### Motivation
Current exception process was broken


tasks/quality_gates.py
```diff
@@ -372,21 +375,24 @@ def debug_specific_quality_gate(ctx, gate_name):
 
 
 @task()
-def exception_threshold_bump(ctx):
+def exception_threshold_bump(ctx, pipelineId):
```
rdesgroppes:
```suggestion
def exception_threshold_bump(ctx, pipeline_id):
```
as per PEP-8, whose digest might be found at: https://realpython.com/python-pep8/#naming-styles

tasks/quality_gates.py
```diff
@@ -372,21 +375,24 @@ def debug_specific_quality_gate(ctx, gate_name):
...
     """
     When a PR is exempt of static quality gates, they have to use this invoke task to adjust the quality gates thresholds accordingly to the exempted added size.
 
     Note: This invoke task must be run on a pipeline that has finished running static quality gates
     :param ctx:
+    :param pipelineId: pipeline ID we want to fetch the artifact from to bump gates
```
rdesgroppes:
```suggestion
    :param pipeline_id: pipeline ID we want to fetch the artifact from to bump gates
```

tasks/quality_gates.py
```diff
@@ -372,21 +375,24 @@ def debug_specific_quality_gate(ctx, gate_name):
...
     :return:
     """
     current_branch_name = get_current_branch(ctx)
-    ancestor_commit = get_common_ancestor(ctx, "HEAD")
     repo = get_gitlab_repo()
     with tempfile.TemporaryDirectory() as extract_dir, ctx.cd(extract_dir):
+        curPipeline = repo.pipelines.get(pipelineId)
+        gateJobId = next(job.id for job in curPipeline.jobs.list() if job.name == "static_quality_gates")
+        gateJob = repo.jobs.get(id=gateJobId)
```
rdesgroppes:
```suggestion
        cur_pipeline = repo.pipelines.get(pipeline_id)
        gate_job_id = next(job.id for job in cur_pipeline.jobs.list() if job.name == "static_quality_gates")
        gate_job = repo.jobs.get(id=gate_job_id)
```

tasks/quality_gates.py
```diff
@@ -372,21 +375,24 @@ def debug_specific_quality_gate(ctx, gate_name):
...
         with open(f"{extract_dir}/gate_archive.zip", "wb") as f:
             try:
-                f.write(repo.artifacts.download(ref_name=current_branch_name, job="static_quality_gates"))
+                f.write(gateJob.artifacts())
```
rdesgroppes:
```suggestion
                f.write(gate_job.artifacts())
```

rdesgroppes (approved):
LGTM, but I guess we could stick to the naming convention (PEP-8, otherwise observed in this file)

Pythyu:
Agreed ! Thanks for the review 👍

tasks/quality_gates.py
```diff
@@ -372,21 +375,24 @@ def debug_specific_quality_gate(ctx, gate_name):
 
 
 @task()
-def exception_threshold_bump(ctx):
+def exception_threshold_bump(ctx, pipeline_id):
```
alopezz, line 378: Question: knowing that there's already logic elsewhere to find pipelines for a given branch, should we perhaps try to leverage something like that rather than requiring manual input here?

tasks/quality_gates.py
```diff
@@ -210,19 +210,22 @@ def parse_and_trigger_gates(ctx, config_path=GATE_CONFIG_PATH):
...
-
     # We don't need a PR notification nor gate failures on release branches
     if not is_a_release_branch(ctx, branch):
         github = GithubAPI()
         if github.get_pr_for_branch(branch).totalCount > 0:
             ancestor = get_common_ancestor(ctx, "HEAD")
-            metric_handler.generate_relative_size(ctx, ancestor=ancestor)
+            metric_handler.generate_relative_size(
+                ctx, ancestor=ancestor, report_path="ancestor_static_gate_report.json"
+            )
```
alopezz, line 220: Question: Given that we're generating the reports regardless of the `is_a_release_branch` condition, shouldn't we be doing this bit also in all cases? Though I don't have the full picture of what these "metric reports" might be used for.

alopezz (commented):
I'm somewhat concerned by the lack of tests here, it's hard to tell at a glance whether the bumping task is working correctly in at least some representative examples.

tasks/quality_gates.py
Pythyu: We had some issues with the gitlab API which sometimes didn't return the latest pipeline ID or jobs run. Since the process is manual anyway, I prefer relying on providing the pipeline ID ourselves.

tasks/quality_gates.py
Pythyu: We don't want to generate it outside of PRs, simply because in some cases they are the ancestor themselves like on main. Also `generate_relative_size` currently use some functions that expect to not be on main otherwise they will start overwriting on the current s3 paths.

Although I agree with header comment and I'm adding some unit test to it

alopezz:
I ran the command locally `dda inv quality-gates.exception-threshold-bump 70102905` and I got

```
/Users/alex.lopez/go/src/github.com/DataDog/datadog-agent/tasks/quality_gates.py:391: UserWarning: Calling a `list()` method without specifying `get_all=True` or `iterator=True` will return a maximum of 20 items. Your query returned 20 of 661 items. See https://python-gitlab.readthedocs.io/en/v4.4.0/api-usage.html#pagination for more details. If this was done intentionally, then this warning can be supressed by adding the argument `get_all=False` to the `list()` call. (python-gitlab: /Users/alex.lopez/Library/Application Support/dda/venvs/legacy/lib/python3.12/site-packages/gitlab/client.py:956)
  gate_job_id = next(job.id for job in cur_pipeline.jobs.list() if job.name == "static_quality_gates")
[SUCCESS] Static Quality gate have been updated ! Total gate threshold impact : 487.35 KiB
```

But the updated values seem fine at a glance. I'lm going to give that warning a closer look before merging.

tasks/quality_gates.py
```diff
@@ -372,21 +375,24 @@ def debug_specific_quality_gate(ctx, gate_name):
...
     :param ctx:
+    :param pipeline_id: pipeline ID we want to fetch the artifact from to bump gates
     :return:
     """
     current_branch_name = get_current_branch(ctx)
-    ancestor_commit = get_common_ancestor(ctx, "HEAD")
     repo = get_gitlab_repo()
     with tempfile.TemporaryDirectory() as extract_dir, ctx.cd(extract_dir):
+        cur_pipeline = repo.pipelines.get(pipeline_id)
+        gate_job_id = next(job.id for job in cur_pipeline.jobs.list() if job.name == "static_quality_gates")
```
rdesgroppes: To fix the warning @alopezz spotted [above](https://github.com/DataDog/datadog-agent/pull/38621#issuecomment-3078900230):
```suggestion
        gate_job_id = next(job.id for job in cur_pipeline.jobs.list(iterator=True) if job.name == "static_quality_gates")
```
Note that `next` would throw a `StopIteration` exception should no `static_quality_gates` job be found, but I guess it's hypothetical.

alopezz: I've rebased and fixed this.

> Note that next would throw a StopIteration exception should no static_quality_gates job be found, but I guess it's hypothetical.

That does sound like an unlikely occurrence given the expected usage and I guess having it fail with a traceback is, while not great, more or less acceptable.

This makes me think that, once we're asking for a pipeline id, we may as well just ask for the job id instead and make this simpler (though neither are particularly good from the user's perspective). Though I'm also still not fully convinced by what [this comment](https://github.com/DataDog/datadog-agent/pull/38621#discussion_r2197517759) said and would much prefer this to not require that kind of input.

I'd also rather this had tests, but I think that given the circumstances we can merge this as is (as it does seem to solve the problems we had for the most part) and look at that later.

rdesgroppes: :100:

rdesgroppes (approved):

PR #38622, opened 2025-07-09
service discovery: Collect log files

vitkyrka:
### What does this PR do?

Collect information about log files opened by processes.  Will be used
later for automatic log tailing.

### Motivation

https://datadoghq.atlassian.net/wiki/spaces/DSCVR/pages/5212537551/

Tests are added.


Yumasi (approved):

pkg/collector/corechecks/servicediscovery/module/logs.go
```diff
@@ -0,0 +1,73 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/DataDog/datadog-agent/pkg/util/kernel"
+)
+
+// isLogFile is used from the getOpenFilesInfo function to filter paths which
+// look like log files.
+func isLogFile(path string) bool {
+	if !strings.HasSuffix(path, ".log") {
```
smeso-ddg: Shouldn't we collect all files ending in .log + all files in /var/log even if they don't end with .log?

pkg/collector/corechecks/servicediscovery/module/logs.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+		fields := strings.Fields(lines[1])
+		if len(fields) < 2 {
+			continue
+		}
+		flags, err := strconv.ParseUint(fields[1], 8, 32)
+		if err != nil {
+			continue
+		}
+		// O_WRONLY is 1
```
smeso-ddg: Shouldn't we check for append-only specifically instead of just write-only?

pkg/collector/corechecks/servicediscovery/module/logs.go
```diff
@@ -0,0 +1,73 @@
...
+	if !strings.HasSuffix(path, ".log") {
+		return false
+	}
+
+	// Skip log files from Docker containers and Kubernetes pods since these
+	// are collected by other means.
+	if strings.HasPrefix(path, "/var/lib/docker/containers") ||
+		strings.HasPrefix(path, "/var/log/pods") {
+		return false
+	}
```
smeso-ddg, line 33: Should we verify that the file is actually text and not binary?

vitkyrka: That will be done in the core agent before deciding to use the log file. I think it's better to do it there and not to spend time opening and reading files here in SP since:
- the gathered information will not necessarily be actively used if the autodiscovery provider is not enabled
- the gathered information for specific processes may not be used since the core agent may ignore container-based processes depending on the config
- the core agent may end up not even being to access the file due to permissions
- state may need to be introduced to this module which is stateless to avoid doing the operation repeatedly for the same file on different invocations while the core-agent already has state

smeso-ddg: Yes, you are right

pkg/collector/corechecks/servicediscovery/module/logs.go
vitkyrka: Done.

pkg/collector/corechecks/servicediscovery/module/logs.go
vitkyrka: Done.

smeso-ddg (approved):
Thanks!

PR #38579, opened 2025-07-08
Add fuzz tests for docker stream parsing, fix truncation panic

blt:
### What does this PR do?

This commit adds fuzz tests for the Docker stream parsing in pkg/logs. These tests demonstrated a panic in the event of a truncated message on the stream, owing to malformed Docker response (unlikely) or a severed connection. This panic was re-confirmed with a unit test and corrected.

### Motivation

Improve our fuzz testing coverage of this project, see also [Agent Conclave](https://datadoghq.atlassian.net/wiki/spaces/agent/pages/5245960569/Conclave+Roadmap#Build-additional-quantitative-tooling-for-measuring-if-Agent-is-fit-for-purpose) material.

### Possible Drawbacks / Trade-offs
None known.

PR #38628, opened 2025-07-09
Introduce fuzz tests for our kubernetes logs parser

blt:
### What does this PR do?

This commit introduces a fuzz test for our kubernetes logs parsing. We previously had a FIXME in the tests that indicated timestamp handling was not strict and on review if k8s kicks an invalid timestamp to use we will then potentially use that in the tailer mechanism as a restart time. This FIXME is now removed and corrected.

The fuzz tests found no other odd behavior here.

We previously did not validate that the content in the timestamp was a valid timestamp, meaning that if we have systems today that return invalid content in this field we accept it silently. After this change an error will be thrown. 

### Additional Notes

Peer to #38579


pkg/logs/internal/parsers/kubernetes/kubernetes.go
```diff
@@ -54,7 +55,17 @@ func parseKubernetes(msg *message.Message) (*message.Message, error) {
...
 	timestamp = string(components[0])
+	_, err := time.Parse(time.RFC3339Nano, timestamp)
+	if err != nil {
+		// Invalid timestamp format - return error to prevent downstream issues
+		// where the timestamp is used as an offset for log tailing
+		return message.NewMessage(msg.GetContent(), nil, status, 0), errors.New("invalid timestamp format")
+	}
+
 	status = getStatus(components[1])
 	flag = string(components[2])
```
jszwedko:
```suggestion
	status = getStatus(components[1])
	flag = string(components[2])
	
	// Validate timestamp format. K8s API uses either RFC3339 or RFC3339Nano
	// but RFC3339Nano is a superset that can parse both formats.
	timestamp = string(components[0])
	_, err := time.Parse(time.RFC3339Nano, timestamp)
	if err != nil {
	    msg.SetContent(content)
	    msg.Status = status
	    msg.ParsingExtra = message.ParsingExtra{
	    	IsPartial: isPartial(flag),
	    }
		// Invalid timestamp format - return error to prevent downstream issues
		// where the timestamp is used as an offset for log tailing
		return msg, errors.New("invalid timestamp format")
	}
```

Here we do have `components` and `content` available so we can pull out the parts of the message we _can_ parse. Note that I think we could restructure the code to reduce duplication here but just commenting a diff to show roughly what I mean.

This also has me wondering:

- Why `ParsingExtra.Timestamp` is a string and not a `time.Time` to force it to be parsed at the time it is set. Instead, here, I believe we are going to be double parsing: once to validate it here, once later when we want to use it as a timestamp.
- If we should be doing this same check in the dockerstream and docker_file parsers (they currently also doesn't validate the timestamp)

jszwedko (commented):
One comment, otherwise another win for fuzzing. These parsers certainly make for a good fuzzing target.

pkg/logs/internal/parsers/kubernetes/kubernetes.go
blt: Both good thoughts. Let me tinker with this.

blt: > If we should be doing this same check in the dockerstream and docker_file parsers (they currently also doesn't validate the timestamp)

Also, on this thought I will open a follow-on PR rather than change #38579.

blt: [9d57522](https://github.com/DataDog/datadog-agent/pull/38628/commits/9d57522aaac195b4ba6652568faaa849191065d3)

blt: Fiddling with it, I think the `ParsingExtra.Timestamp` idea is interesting and worth pursuing, outside scope here for sure. But compelling.

releasenotes/notes/Avoid-passing-invalid-timestamps-from-kubernetes-log-parsing-3ed6269788fc99bd.yaml
```diff
@@ -0,0 +1,11 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+fixes:
+  - |
+    Timestamps from the kubernetes logs source are now validated, we will not pass invalid ones.
```
michaelcretzman:
```suggestion
    Kubernetes logs source timestamps are now validated. Invalid timestamps will not be passed.
```

michaelcretzman (approved):
Approved with minor suggestion.

blt:
`docker_containers_cpu` failure unrelated, see https://github.com/DataDog/datadog-agent/pull/38633

pkg/logs/internal/parsers/kubernetes/kubernetes.go
```diff
@@ -54,10 +55,25 @@ func parseKubernetes(msg *message.Message) (*message.Message, error) {
...
+			IsPartial: isPartial(flag),
+		}
+		// Invalid timestamp format - return error to prevent downstream issues
+		// where the timestamp is used as an offset for log tailing
+		return msg, errors.New("invalid timestamp format")
+	}
+
 	msg.SetContent(content)
 	msg.Status = status
 	msg.ParsingExtra = message.ParsingExtra{
```
jszwedko: Nit: I think you could move lines 77-79 up and then you could delete lines 67-69. Like:

```golang
	status = getStatus(components[1])
	flag = string(components[2])

		msg.SetContent(content)
		msg.Status = status
		msg.ParsingExtra = message.ParsingExtra{
			IsPartial: isPartial(flag),
		}

	// Validate timestamp format. K8s API uses either RFC3339 or RFC3339Nano
	// but RFC3339Nano is a superset that can parse both formats.
	timestamp = string(components[0])
	_, err := time.Parse(time.RFC3339Nano, timestamp)
	if err != nil {
		// Invalid timestamp format - return error to prevent downstream issues
		// where the timestamp is used as an offset for log tailing
		return msg, errors.New("invalid timestamp format")
	}

	msg.ParsingExtra.Timestamp = timestamp
  }

```

blt: Incorporated in [ebc4872](https://github.com/DataDog/datadog-agent/pull/38628/commits/ebc48729e003d3219af53fe3a8e701be5148a7f8).

jszwedko (approved):

blt:
* **#38628** <a href="https://app.graphite.dev/github/pr/DataDog/datadog-agent/38628?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/datadog-agent/38628?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #38630, opened 2025-07-09
pkg/dyninst: Check presence bitset, decoding fixes

grantseltzer:
### What does this PR do?

- Adds a check for the presence bitset in root data entries
- Adds population of "type" field in array elements
- Adds "size" field to decoded slices/arrays, and notCaptureReason in case of truncation (only slices)
- Fixes integration test to not skip the entire test suite in cases of mismatching architecture

### Motivation

Correctness in uploaded snapshots

Updated and ran integration tests

### DIFF
--- a/pkg/dyninst/decode/marshal.go
+++ b/pkg/dyninst/decode/marshal.go
@@ -337,7 +337,7 @@ func (d *Decoder) encodeValueFields(
 				currentlyEncoding,
 				v.Element,
 				elementData,
-				"",
+				elementType.GetName(),
 			); err != nil {
 				return err
 			}
--- a/pkg/dyninst/decode/marshal.go
+++ b/pkg/dyninst/decode/marshal.go
@@ -321,15 +321,17 @@ func (d *Decoder) encodeValueFields(
 		return nil
 	case *ir.ArrayType:
 		var err error
+		elementType := v.Element
+		elementSize := int(elementType.GetByteSize())
+		numElements := int(v.Count)
 		if err = writeTokens(enc,
+			jsontext.String("size"),
+			jsontext.String(strconv.Itoa(numElements)),
 			jsontext.String("elements"),
 			jsontext.BeginArray); err != nil {
 			return err
 		}
 
-		elementType := v.Element
-		elementSize := int(elementType.GetByteSize())
-		numElements := int(v.Count)
 		for i := range numElements {
 			elementData := data[i*elementSize : (i+1)*elementSize]
 			if err := d.encodeValue(enc,
@@ -411,6 +413,12 @@ func (d *Decoder) encodeValueFields(
 		}
 		if err = writeTokens(enc, jsontext.EndArray); err != nil {
 			return err
+		}
+		if length > uint64(sliceLength) {
+			return writeTokens(enc,
+				jsontext.String("notCapturedReason"),
+				jsontext.String("collectionSize"),
+			)
 		}
 		return nil
 	case *ir.GoChannelType:
--- a/pkg/dyninst/decode/marshal.go
+++ b/pkg/dyninst/decode/marshal.go
@@ -85,7 +85,8 @@ type argumentsData struct {
 // bytes, and presence bits in the bitset correspond with index of
 // the expression in the root ir.
 func expressionIsPresent(bitset []byte, expressionIndex int) bool {
-	return (bitset[expressionIndex/8] & (1 << byte(expressionIndex%8))) != 0
+	idx, bit := expressionIndex/8, expressionIndex%8
+	return idx < len(bitset) && bitset[idx]&(1<<byte(bit)) != 0
 }
 
 func (ad *argumentsData) MarshalJSONTo(enc *jsontext.Encoder) error {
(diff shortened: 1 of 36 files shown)


pkg/dyninst/integration_test.go
```diff
@@ -87,9 +87,15 @@ func TestDyninst(t *testing.T) {
 			t.Logf("%s is not used in integration tests", svc)
 			continue
 		}
-		t.Run(svc, func(t *testing.T) {
-			runIntegrationTestSuite(t, svc, cfgs, rewrite)
-		})
+		for _, cfg := range cfgs {
+			if cfg.GOARCH != runtime.GOARCH {
```
piob-io: Now that this is pulled out of the code below again, we could use the Skip, as long as we do the arch check from within t.Run. I think that gives us nicer output.

pkg/dyninst/integration_test.go
```diff
@@ -281,6 +287,8 @@ func testDyninst(
 		var decodeOut bytes.Buffer
 		probe, err := decoder.Decode(event, cachingSymbolicator, &decodeOut)
 		require.NoError(t, err)
+
+		t.Log("Output:", decodeOut.String())
```
piob-io: This feels a bit too verbose, what's the value? We could write this out into a file under the temp dir we create for every test, and contains other larger debugging artifacts.

piob-io (approved):
Thanks!

pkg/dyninst/integration_test.go
```diff
@@ -408,42 +416,34 @@ func runIntegrationTestSuite(
...
-			t.Skipf(
-				"cross-execution is not supported, running on %s, skipping %s",
-				runtime.GOARCH, cfg.GOARCH,
-			)
-			continue
+
+	bin := testprogs.MustGetBinary(t, service, cfg)
+	for _, debug := range []bool{false, true} {
+		if testing.Short() && debug {
+			t.Skip("skipping debug mode in short mode")
```
ajwerner: I think this skip call should get pushed into `runTest` or removed.

pkg/dyninst/integration_test.go
ajwerner: How would you feel about gating this on an environment variable? Or using `log.Tracef` so that `DD_LOG_LEVEL` will control it? I find it to be very noisy when just running the tests.

pkg/dyninst/decode/marshal.go
```diff
@@ -79,6 +79,15 @@ type argumentsData struct {
 	decoder  *Decoder
 }
 
+// In the root data item, before the expressions, there is a bitset
+// which conveys if expression values are present in the data.
+// The rootType.PresenceBitsetSize conveys the size of the bitset in
+// bytes, and presence bits in the bitset correspond with index of
+// the expression in the root ir.
+func expressionIsPresent(bitset []byte, expressionIndex int) bool {
+	return (bitset[expressionIndex/8] & (1 << byte(expressionIndex%8))) != 0
```
ajwerner: for safety how about we include a bounds check?

```go
	idx, bit := expressionIndex/8, expressionIndex%8
	return idx < len(bitset) && bitset[idx]&(1<<byte(bit)) != 0
```

pkg/dyninst/testdata/decoded/sample/testEmptyStruct.json
```diff
@@ -26,7 +26,7 @@
             "arguments": {
               "e": {
                 "type": "main.emptyStruct",
-                "fields": {}
+                "notCapturedReason": "unavailable"
```
ajwerner: I think the output before was better. We want to do something different for zero-byte expressions. There's literally nothing to capture. What we want to do, I think, is consult the type's byte size even before looking at the presence bitset. If it's zero-bytes, we always say we have it.

ajwerner (commented):
Other than the zero-sized type thing, LGTM

pkg/dyninst/integration_test.go
grantseltzer: Sure - I like the sanity check and use it during development for decoding. I'll move it to a log.Tracef (Need DD_LOG_LEVEL=trace) as Andrew suggested.

grantseltzer: ~I'll change it to log.Tracef!~ Actually need to use our own env variable.

pkg/dyninst/testdata/decoded/sample/testEmptyStruct.json
grantseltzer: Ah - yea. Good catch, thanks.

ajwerner (approved):
LGTM

ajwerner:
You need to regenerate irgen. Sorry, noticed on the last round and thought you might have too. Generally a good practice to always run `pkg/dyninst/...` before pushing

ajwerner:
The bad news is that I'm going to merge some things before you and you're going to have to rebase...

grantseltzer:
> You need to regenerate irgen. Sorry, noticed on the last round and thought you might have too. Generally a good practice to always run `pkg/dyninst/...` before pushing

Ah... I did notice but misread which test it was. And all good, i'll keep an eye on your PR merging.

grantseltzer:
Single machine performance ... rebased.

PR #38631, opened 2025-07-09
[AGENTCFG-310] Adding flare retry logic, if the error is transitory

rahulkaukuntla:
### What does this PR do?
We want to retry sending flares if the returned error code is transitory (5xx).

Unit tests.


comp/core/flare/helpers/send_flare.go
```diff
@@ -254,13 +260,63 @@ func SendTo(cfg pkgconfigmodel.Reader, archivePath, caseID, email, apiKey, url s
...
+			// Exponential backoff
+			delay := time.Duration(attempt) * defaultRetryConfig.BaseDelay
+			time.Sleep(delay)
+		}
+
+		r, err := readAndPostFlareFile(archivePath, caseID, email, hostname, url, source, client, apiKey)
+		if err != nil {
+			if isRetryableFlareError(err) {
+				lastErr = err
+				continue // Retry
```
hush-hush: We don't call `r.Body.Close` here, is that expected ?

comp/core/flare/helpers/send_flare.go
```diff
@@ -254,13 +260,63 @@ func SendTo(cfg pkgconfigmodel.Reader, archivePath, caseID, email, apiKey, url s
...
+			}
+			// Non-retryable error - return immediately
+			return "", err
+		}
+
+		if r.StatusCode >= 200 && r.StatusCode < 500 {
+			defer r.Body.Close()
+			return analyzeResponse(r, apiKey)
+		} else if r.StatusCode >= 500 && r.StatusCode < 600 {
```
hush-hush: This should be in the `isRetryableFlareError` so all the retry logic is in one place.

comp/core/flare/helpers/send_flare.go
```diff
@@ -254,13 +260,63 @@ func SendTo(cfg pkgconfigmodel.Reader, archivePath, caseID, email, apiKey, url s
...
+			// 5xx error - retry
+			r.Body.Close()
+			lastErr = fmt.Errorf("server error: %s", r.Status)
+			continue
+		}
+		// Other unexpected status code - don't retry
+		defer r.Body.Close()
+		return analyzeResponse(r, apiKey)
```
hush-hush, line 295: This is duplicate with line 288-289, could we have a single case for all the non-retryable errors ?

comp/core/flare/helpers/send_flare.go
```diff
@@ -254,13 +260,63 @@ func SendTo(cfg pkgconfigmodel.Reader, archivePath, caseID, email, apiKey, url s
...
 	}
 
-	r, err := readAndPostFlareFile(archivePath, caseID, email, hostname, url, source, client, apiKey)
-	if err != nil {
-		return "", err
+	// Retry logic for the actual flare file posting
+	var lastErr error
+	var defaultRetryConfig = RetryConfig{
+		MaxRetries: 3,
+		BaseDelay:  100 * time.Millisecond,
```
hush-hush: Why do we need a struct here instead of 2 vars ?

hush-hush (changes requested):

comp/core/flare/helpers/send_flare.go
rahulkaukuntla: No, it is not, I have refactored to ensure that the response body is always closed if there is an error (and is deferred to analyzeResponse function if there is no error).

comp/core/flare/helpers/send_flare.go
rahulkaukuntla: Done.

comp/core/flare/helpers/send_flare.go
rahulkaukuntla: Sounds good, I have refactored to do this.

comp/core/flare/helpers/send_flare.go
rahulkaukuntla: We do not 😅. I think that this is result of a refactor, and so, I have replaced this struct with 2 variables.

rahulkaukuntla:
Thank you for taking a look at this! Your review helped me find an issue in the code, where after calling `client.Do(request)`, if a 5xx code is returned without an error, then the retry logic wasn't being reached. I'm pretty sure that an error should always be returned in that case, but just in case it isn't, I added [some logic](https://github.com/DataDog/datadog-agent/pull/38631/files#diff-d0a8d437666af642c3f9e2c474570dd93c6e1bc710daea6e868b4aefc1e6e0f9R149-R161) to handle this situation.

comp/core/flare/helpers/send_flare.go
```diff
@@ -45,6 +45,12 @@ type FlareSource struct {
 	rcTaskUUID string
 }
 
+// RetryConfig holds configuration for retry behavior
+type RetryConfig struct {
+	MaxRetries int
+	BaseDelay  time.Duration
+}
+
```
hush-hush: Is this used anywhere ?

comp/core/flare/helpers/send_flare.go
```diff
@@ -231,7 +249,7 @@ func mkURL(baseURL string, caseID string) string {
 }
 
 // SendTo sends a flare file to the backend. This is part of the "helpers" package while all the code is moved to
-// components. When possible use the "Send" method of the "flare" component instead.
+// components. When possible use the "Send" method of the "flare" component instead. Configurable retry logic is used.
```
hush-hush: Is it configurable ?

comp/core/flare/helpers/send_flare.go
```diff
@@ -254,13 +272,65 @@ func SendTo(cfg pkgconfigmodel.Reader, archivePath, caseID, email, apiKey, url s
...
 
-	r, err := readAndPostFlareFile(archivePath, caseID, email, hostname, url, source, client, apiKey)
-	if err != nil {
-		return "", err
+	// Retry logic for the actual flare file posting
+	var lastErr error
+	var maxTries = 3
+	var baseDelay = 100 * time.Millisecond
+
+	for attempt := 1; attempt <= maxTries; attempt++ {
```
hush-hush: We don't need 2 variables to track retries.

```suggestion
	var lastErr error
	var baseDelay = 100 * time.Millisecond

	for attempt := 3; attempt > 0; attempt-- {
```

comp/core/flare/helpers/send_flare.go
```diff
@@ -254,13 +272,65 @@ func SendTo(cfg pkgconfigmodel.Reader, archivePath, caseID, email, apiKey, url s
...
+		if attempt > 1 {
+			// Exponential backoff
+			delay := time.Duration(attempt-1) * baseDelay
+			time.Sleep(delay)
+		}
```
hush-hush: This should be moved after the call to `readAndPostFlareFile`, in the error handling case. It usually produce more readable code to have `core logic -> error hangling -> loop back to the top`.

comp/core/flare/helpers/send_flare.go
```diff
@@ -254,13 +272,65 @@ func SendTo(cfg pkgconfigmodel.Reader, archivePath, caseID, email, apiKey, url s
...
-	if err != nil {
-		return "", err
+	// Retry logic for the actual flare file posting
+	var lastErr error
+	var maxTries = 3
+	var baseDelay = 100 * time.Millisecond
+
+	for attempt := 1; attempt <= maxTries; attempt++ {
+		if attempt > 1 {
+			// Exponential backoff
```
hush-hush: I'm not sure we need an exponential backoff since the delay will range from 100ms to 300ms.

comp/core/flare/helpers/send_flare.go
```diff
@@ -254,13 +272,65 @@ func SendTo(cfg pkgconfigmodel.Reader, archivePath, caseID, email, apiKey, url s
 		return "", err
 	}
 
-	r, err := readAndPostFlareFile(archivePath, caseID, email, hostname, url, source, client, apiKey)
-	if err != nil {
-		return "", err
+	// Retry logic for the actual flare file posting
+	var lastErr error
+	var maxTries = 3
+	var baseDelay = 100 * time.Millisecond
```
hush-hush: Let's increase the delay a bit, I'd say 1 second would be better.

comp/core/flare/helpers/send_flare.go
```diff
@@ -254,13 +272,65 @@ func SendTo(cfg pkgconfigmodel.Reader, archivePath, caseID, email, apiKey, url s
...
+				statusCode = r.StatusCode
+			}
+			lastErr = err
+			if attempt < maxTries && isRetryableFlareError(err, statusCode) {
+				continue
+			}
+			// If it's not retryable, return immediately
+			if !isRetryableFlareError(err, statusCode) {
+				return "", err
+			}
```
hush-hush: Here we call the same function with the same parameter twice.

We could avoid this:
```suggestion
			if isRetryableFlareError(err, statusCode) {
				if attempt > 0 {
					log.Warn("Failed to send flare, retrying in 1 second")
					time.Sleep(1*time.Second)
				}
				return "", err
			} else {
				return "", err
			}
```

comp/core/flare/helpers/send_flare.go
```diff
@@ -254,13 +272,65 @@ func SendTo(cfg pkgconfigmodel.Reader, archivePath, caseID, email, apiKey, url s
...
+
+			// If we've exhausted retries for a retryable error, fall through to return retry exhaustion error
+			break
```
hush-hush: If we break here we would clode the body twice in case the error is retryable but we exceed the maximum number of attempts (we would close it once line 291 and line 312).

comp/core/flare/helpers/send_flare.go
```diff
@@ -254,13 +272,65 @@ func SendTo(cfg pkgconfigmodel.Reader, archivePath, caseID, email, apiKey, url s
...
+		if attempt > 1 {
+			// Exponential backoff
+			delay := time.Duration(attempt-1) * baseDelay
+			time.Sleep(delay)
+		}
+
+		r, err := readAndPostFlareFile(archivePath, caseID, email, hostname, url, source, client, apiKey)
+		if err != nil {
+			// Always close the response body if it exists
+			if r != nil && r.Body != nil {
```
hush-hush: In which case will `r.Body` be nil ?

hush-hush (changes requested):

comp/core/flare/helpers/send_flare.go
rahulkaukuntla: Nope, I've removed it.

comp/core/flare/helpers/send_flare.go
rahulkaukuntla: No, I think that was from an earlier version, so I have removed it.

comp/core/flare/helpers/send_flare.go
rahulkaukuntla: 🤦 I've implemented the suggestion.

comp/core/flare/helpers/send_flare.go
rahulkaukuntla: I moved the delay logic inside the error handling block, after the retryability check. Refer [here](https://github.com/DataDog/datadog-agent/pull/38631#discussion_r2226431131)

comp/core/flare/helpers/send_flare.go
rahulkaukuntla: Done.

comp/core/flare/helpers/send_flare.go
rahulkaukuntla: I think that I've come up with a cleaner version--let me know what you think:
```
for attempt := 3; attempt > 0; attempt-- {
	r, err := readAndPostFlareFile(archivePath, caseID, email, hostname, url, source, client, apiKey)
	if err != nil {
		// Always close the response body if it exists
		if r != nil {
			r.Body.Close()
		}

		statusCode := 0
		if r != nil {
			statusCode = r.StatusCode
		}
		lastErr = err

		if !isRetryableFlareError(err, statusCode) {
			return "", err
		}
		log.Warn("Failed to send flare, retrying in 1 second")
		time.Sleep(baseDelay)
		continue
	}

	// Success case - analyze the response
	defer r.Body.Close()
	return analyzeResponse(r, apiKey)
}
```

comp/core/flare/helpers/send_flare.go
rahulkaukuntla: Makes sense, I just make it 1 second every time.

comp/core/flare/helpers/send_flare.go
rahulkaukuntla: Nice, I didn't think of that. Refer to [this comment](https://github.com/DataDog/datadog-agent/pull/38631#discussion_r2226431131) to see how I refactored--now there is no break statement, so this issue shouldn't arise.

comp/core/flare/helpers/send_flare.go
rahulkaukuntla: Technically speaking, if the HTTP client contract were to be violated by some sort of custom HTTP transport implementation, a non-nil response with a nil Body field could be returned (though definitely unlikely). However, since other Datadog employees will be calling this function, I think that we can trust them to follow conventions. As such, I've removed the `r.Body != nil` check.

comp/core/flare/helpers/send_flare.go
```diff
@@ -254,13 +267,55 @@ func SendTo(cfg pkgconfigmodel.Reader, archivePath, caseID, email, apiKey, url s
...
+		if err != nil {
+			// Always close the response body if it exists
+			if r != nil {
+				r.Body.Close()
+			}
+
+			statusCode := 0
+			if r != nil {
+				statusCode = r.StatusCode
+			}
```
hush-hush:
```suggestion
			// Always close the response body if it exists
			statusCode := 0
			if r != nil {
				statusCode = r.StatusCode
				r.Body.Close()
			}
```

hush-hush (approved):
Added one last nit-pick. Looks good otherwise. Feel free to merge after addressing.

PR #38643, opened 2025-07-10
Push back "chore: Datadog URL connectivity" (#38629)"

san-jos:
This reverts commit 053f87bd6728861c3e5795066e007846fe8b4822.

### What does this PR do?

Remove a failing test causing #incident-40491

### DIFF
--- a/pkg/diagnose/connectivity/inventoryendpoint_test.go
+++ b/pkg/diagnose/connectivity/inventoryendpoint_test.go
@@ -1,0 +1,498 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+package connectivity
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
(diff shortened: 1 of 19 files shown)


coignetp (approved):

louis-cqrl (approved):

pkg/diagnose/connectivity/inventoryendpoint_test.go
```diff
@@ -397,27 +396,6 @@ func TestCheckGet(t *testing.T) {
 	assert.Contains(t, err.Error(), "invalid status code: 401")
 }
 
-func TestCheckGetConnectionFailure(t *testing.T) {
-	// Create a resolvedEndpoint with an invalid URL
-	endpoint := resolvedEndpoint{
-		url:    "http://invalid-url-that-does-not-exist.com",
```
dustmop: Not necessary for this PR, but if you want to add test for this in the future, you could use `invalid.com` as recommended by [RFC 2606](https://www.rfc-editor.org/rfc/rfc2606.html#section-2). Sorry for not catching this in the first review.

dustmop (approved):
LGTM

pkg/diagnose/connectivity/inventoryendpoint_test.go
san-jos: You can't see it all :)

PR #38647, opened 2025-07-10
[AMCC-13] dogstatsd: report telemetry on blocked points/hist aggregates.

remeh:
### What does this PR do?

Report extra telemetry on blocked points & hist aggregates.

Validated the values appear on the telemetry endpoint.


pkg/aggregator/aggregator.go
```diff
@@ -167,6 +167,7 @@ var (
 		[]string{"shard", "metric_type"}, "Count the number of checks contexts in the check aggregator, by metric type")
 	tlmChecksContextsBytesByMtype = telemetry.NewGauge("aggregator", "checks_contexts_bytes_by_mtype",
 		[]string{"shard", "metric_type", tags.BytesKindTelemetryKey}, "Estimated count of bytes taken by contexts in the check aggregator, by metric type")
+	tlmBlockedHistoAggregates = telemetry.NewSimpleCounter("aggregator", "blocked_hist_aggregates", "How many histogram aggregates were blocked")
```
vickenty: Would it make sense to give this a bit more generic name (like `blocked_metrics`), since from the aggregator point of view it isn't limited to histograms? Do we need to distinguish it from eventual blocked metrics from checks?

Or at least spell histogram in full:

```suggestion
	tlmBlockedHistoAggregates = telemetry.NewSimpleCounter("aggregator", "blocked_histogram_aggregates", "How many histogram aggregates were blocked")
```

vickenty (approved):

pkg/aggregator/aggregator.go
remeh: Very good point, will change this.

PR #38658, opened 2025-07-10
Create a release branch in datadog-agent-buildimages after the cut-off on the right commit

FlorentClarret:
### What does this PR do?

Create a release branch in datadog-agent-buildimages after the cut-off on the right commit

### Motivation

We currently create the branch on top of main, which is not always technically correct. We need to create it on top of the commit we're currently using


tasks/libs/common/gomodules.py
```diff
@@ -74,7 +74,7 @@ def to_dict(self) -> dict[str, object]:
         modules_config.update(
             {name: module.to_dict(remove_path=True) or 'default' for name, module in self.modules.items()}
         )
-        modules_config.update({module: 'ignored' for module in self.ignored_modules})
+        modules_config.update(dict.fromkeys(self.ignored_modules, 'ignored'))
```
FlorentClarret:
```suggestion
        modules_config.update({module: 'ignored' for module in self.ignored_modules})
```

chouetz (approved):

spencergilbert (approved):

PR #38662, opened 2025-07-10
[NDINT-226] Versa Interface (and Link) Metrics and Metadata

ken-schneider:
### What does this PR do?
The largest changes in this PR are around support for interfaces in Versa. 4 new endpoints can be called depending on config flags passed to the agent. These endpoints gather interface metadata, interface metrics (throughput/bandwidth), link availability, link usage metrics (throughput/bandwidth).

Additionally, this PR cleans up some of configuration options and re-organizes the main check to consolidate the loops that exist for tenant based calls.

### Motivation
Continue to build out Versa integration

Ran locally against our own Versa cluster

GetInterfaces is a single call which currently has no pagination whatsoever. For smaller organizations this call is probably fine, but for larger ones this is not. Unfortunately, through the docs and UI I haven't been able to find an API call that would give us the same data in a paginated way for the entire tenant.


pkg/collector/corechecks/network-devices/versa/client/client.go
```diff
@@ -251,6 +265,163 @@ func (client *Client) GetChildAppliancesDetail(tenant string) ([]Appliance, erro
...
+func (client *Client) closeQuery(queryID string) {
+	closeEndpoint := "/vnms/dashboard/appliance/close_query"
+	params := map[string]string{
+		"queryId": queryID,
+	}
+
+	// We don't expect any meaningful response from the close endpoint
+	_, err := client.get(closeEndpoint, params, false)
+	if err != nil {
+		log.Warnf("failed to close query %s: %v", queryID, err)
```
ken-schneider: Debating the level of this, maybe it makes sense to be an info or even a debug?

pkg/collector/corechecks/network-devices/versa/client/client.go
```diff
@@ -65,9 +67,21 @@ func NewClient(directorEndpoint, analyticsEndpoint, username, password string, u
 		return nil, err
 	}
 
+	// TODO: remove this in favor of allowing certs to be
```
jmw51798: Are this and other TODOs intentionally left in?

pkg/collector/corechecks/network-devices/versa/client/client.go
```diff
@@ -47,14 +49,14 @@ type Client struct {
...
+	lookback            string
 }
 
 // ClientOptions are the functional options for the Versa client
 type ClientOptions func(*Client)
 
 // NewClient creates a new Versa HTTP client.
-func NewClient(directorEndpoint, analyticsEndpoint, username, password string, useHTTP bool, options ...ClientOptions) (*Client, error) {
+func NewClient(directorEndpoint string, directorPort int, analyticsEndpoint string, username string, password string, useHTTP bool, skipCertVerification bool, options ...ClientOptions) (*Client, error) {
 	err := validateParams(directorEndpoint, analyticsEndpoint, username, password)
```
jmw51798: Should directorPort be validated?

pkg/collector/corechecks/network-devices/versa/client/client.go
jmw51798: I would agree that a lower level should be fine as GetInterfaceMetrics() will return an appropriate error if there's an actual issue.

pkg/collector/corechecks/network-devices/versa/client/client.go
```diff
@@ -318,18 +489,109 @@ func (client *Client) GetSLAMetrics() ([]SLAMetrics, error) {
...
+			}
+		}
+		rows = append(rows, m)
+	}
+	return rows, nil
+}
+
+// GetLinkUsageMetrics gets link metrics for a Versa tenant
+func (client *Client) GetLinkUsageMetrics(tenant string) ([]LinkUsageMetrics, error) {
+	analyticsURL := client.buildAnalyticsPath(tenant, "SDWAN", "linkusage(site,accckt,accckt.uplinkBW,accckt.downlinkBW,accckt.type,accckt.media,accckt.ip,accckt.isp)", "tableData", []string{
```
jmw51798, line 596: IDK much about Versa or SDWAN, so just checking is "SDWAN" correct here?

pkg/collector/corechecks/network-devices/versa/client/client.go
ken-schneider: Yes, good catch let me add that.

pkg/collector/corechecks/network-devices/versa/client/client.go
ken-schneider: Some of them are still left in intentionally, but this one can go.

pkg/collector/corechecks/network-devices/versa/client/client.go
ken-schneider: It is, it's a path param that's part of certain API calls for metrics about SDWAN related metrics.

pkg/collector/corechecks/network-devices/versa/client/client.go
ken-schneider: I'm moving it down to debug

jmw51798 (approved):

PR #38663, opened 2025-07-10
[CWS] Some edit on setsockopt tracking

theop-dd:
### What does this PR do?
Some code modifications, refactoring, and the addition of a check flag were made
### Motivation
We had to make some modifications to improve the tracking of the setsockopt syscalls.

### DIFF
--- a/pkg/security/probe/field_handlers_ebpf.go
+++ b/pkg/security/probe/field_handlers_ebpf.go
@@ -951,6 +951,7 @@ func (fh *EBPFFieldHandlers) ResolveSetSockOptFilterInstructions(_ *model.Event,
 		instructions, allDecoded := bpf.Disassemble(raw)
 		if !allDecoded {
 			seclog.Warnf("failed to decode setsockopt filter instructions: %s", e.FilterHash)
+			return ""
 		}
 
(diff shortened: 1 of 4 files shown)


pkg/security/probe/field_handlers_ebpf.go
```diff
@@ -948,7 +948,10 @@ func (fh *EBPFFieldHandlers) ResolveSetSockOptFilterInstructions(_ *model.Event,
 			})
 		}
 
-		instructions, _ := bpf.Disassemble(raw)
+		instructions, allDecoded := bpf.Disassemble(raw)
+		if !allDecoded {
+			seclog.Warnf("failed to decode setsockopt filter instructions: %s", e.FilterHash)
```
spikat, line 953: should we return here to not resolve incomplete/garbage instructions?

theop-dd: That's what I was thinking about right now yes, I don't really know because maybe we want to see if we can get the first instructions if the algo stop in the middle for example. It's probably safer to return "" but we can lose some info sometimes

spikat (approved):

PR #38677, opened 2025-07-10
[AGENTRUN-540] Use a integer distance to compare check values for networkv2 e2e test…

jeremy-hanna:
… output

### What does this PR do?

Using 20% margin to compare small integer counts of connections fails like [here](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1023668030) where we are comparing 3 vs 4 which would need a 25% margin.

Instead we can compare with a integer distance to avoid this issue.


pgimalac (approved):

jose-manuel-almaza:
Are all metrics now compared this way? Or should only some be like the one you mention in the description?

jeremy-hanna:
> Are all metrics now compared this way? Or should only some be like the one you mention in the description?

just the network check metrics will be compared this way

PR #38679, opened 2025-07-10
dyninst: add a more efficient way to resolve PCs

andreimatei:
This patch adds a facility for resolving multiple increasing program
counters within a single function more efficiently than repeated calls
to GoSymbolTable.LocatePC(), by amortizing the lookups for the function
information. When exploring DWARF for SymDB we're generally in the
context of a function and we want to resolve many program counters
within it before moving on to the next function.

The new PCIterator basically exports the internal `stepper`, with a more
tailored interface.

### DIFF
--- a/pkg/dyninst/gosym/pc_iterator.go
+++ b/pkg/dyninst/gosym/pc_iterator.go
@@ -1,0 +1,150 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+//go:build linux_bpf
+
+package gosym
+
+import "fmt"
+
+// PCIterator iterates over program counter ranges within a single function,
+// mapping them to source line numbers.
+type PCIterator struct {
+	stepper stepper
+
+	// inlinedPcRanges is a list of program counter ranges that correspond to
+	// other functions inlined within the current function. The ranges are
+	// inclusive on the left, exclusive on the right, disjoint and ordered.
+	inlinedPcRanges [][2]uint64
+	// A cursor into inlinedPcRanges positioned corresponding to the stepper.pc:
+	// inlinedPcRanges[inlinedPcRangesIdx][1] <= stepper.pcStart < inlinedPcRanges[inlinedPcRangesIdx][1]
+	inlinedPcRangesIdx int
+}
+
+// makeFuncPCIterator builds a PCIterator for the function described by fi.
+func makeFuncPCIterator(fi *funcInfo, inlinedPcRanges [][2]uint64) (PCIterator, error) {
+	pcLine, found := fi.pcln()
+	if !found {
+		return PCIterator{}, fmt.Errorf("failed to find line table entry for function")
+	}
+
+	offset := fi.lt.pcTab[0] + int(pcLine)
+	if offset >= len(fi.lt.data) {
+		return PCIterator{}, fmt.Errorf("function line table offset out of bounds")
+	}
+	cursor := fi.lt.data[offset:]
+	entryPC, ok := fi.entryPC()
+	if !ok {
+		return PCIterator{}, fmt.Errorf("failed to get function entry PC")
+	}
+	stepper := makeStepper(cursor, entryPC, fi.lt.quantum)
+
+	return PCIterator{
+		stepper:            stepper,
+		inlinedPcRanges:    inlinedPcRanges,
+		inlinedPcRangesIdx: 0,
+	}, nil
+}
+
+func (f *PCIterator) Reset() {
+	f.stepper.reset()
+	f.inlinedPcRangesIdx = 0
+}
+
+// PCToLine returns the line number corresponding to the given program counter.
+// Only line numbers that correspond to the iterator's own function are
+// returned; if the PC corresponds to an inlined function, returns (0, false).
+//
+// Upon return, the iterator is positioned on the PC range containing the given
+// PC.
+//
+// Any PC inside the function is a valid input, but the PCIterator is more
+// efficient if PCToLine is called with monotonically increasing PCs.
+//
+// Returns false if the PC could not be resolved to a line number.
+func (f *PCIterator) PCToLine(pc uint64) (uint32, bool) {
+	if f.stepper.pcEnd > pc {
+		// We cannot step backwards, so re-initialize the iterator and
+		// start the iteration from the beginning of the function.
+		f.Reset()
+	}
+
+	// Advance the iterator to the right position.
+	var mapping PCRangeMapping
+	for f.stepper.first || pc >= f.stepper.pcEnd {
+		var ok bool
+		mapping, ok = f.Next()
+		if !ok {
+			return 0, false
+		}
+	}
+	return mapping.Line, true
+}
+
+// PCRangeMapping represents a mapping from a range of program counters to a
+// line of source code.
+type PCRangeMapping struct {
+	// The start of the range; inclusive.
+	PCStart uint64
+	// The end of the range; exclusive.
+	PCEnd uint64
+	// The line that the range [PCStart, PCEnd) maps to.
+	Line uint32
+}
+
+// Next advances the iterator to the next PC range, skipping over ranges
+// corresponding to inlined code. Returns information about the current PC
+// range. Returns PCRangeMapping{}, false if the iterator has been exhausted.
+//
+// Calling Next() again after it has returned false is a no-op that will
+// continue returning false.
+func (f *PCIterator) Next() (PCRangeMapping, bool) {
+	// Advance the stepper possibly multiple times to skip over ranges
+	// corresponding to inlined functions.
+	for {
+		if !f.stepper.step() {
+			return PCRangeMapping{}, false
+		}
+		f.advanceInlinedRanges()
+		if !f.pcInInlinedFunc() {
+			break
+		}
+	}
+
+	return PCRangeMapping{
+		PCStart: f.stepper.pcStart,
+		PCEnd:   f.stepper.pcEnd,
+		Line:    uint32(f.stepper.val),
+	}, true
+}
+
+// advanceInlinedRanges moves up f.inlinedPcRangesIdx such that it points after all the ranges
+// that end before the current PC range (i.e. before f.stepper.pcStart).
+func (f *PCIterator) advanceInlinedRanges() {
+	for f.inlinedPcRangesIdx < len(f.inlinedPcRanges) {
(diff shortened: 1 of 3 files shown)


pkg/dyninst/gosym/symtab.go
```diff
@@ -1180,3 +1193,86 @@ func stringFromOffsetPtr(data []byte, offset int) *string {
...
 	return &result
 }
+
+// FuncPCResolver resolves program counters within a single function to source
+// line numbers. It is optimized for resolving multiple PCs in increasing order.
+type FuncPCResolver struct {
+	funcPC uint64
+	lt     *lineTable
+
+	stepper *stepper
```
ajwerner: nit: no need to have this be a pointer, just make it part of the struct and change `newStepper` to `makeStepper`?

pkg/dyninst/gosym/symtab.go
```diff
@@ -1180,3 +1193,86 @@ func stringFromOffsetPtr(data []byte, offset int) *string {
...
+//
+// Any PC inside the function is a valid input, but the FuncPCResolver is more
+// efficient if PCToLine is called with monotonically increasing PCs.
+//
+// Returns false if the PC could not be resolved to a line number.
+func (f *FuncPCResolver) PCToLine(pc uint64) (uint32, bool) {
+	if f.stepper.pc > pc {
+		// We cannot step backwards, so re-initialize the FuncPCResolver and
+		// start the iteration from the beginning of the function.
+		newResolver := newFuncPCResolver(f.lt, f.funcPC)
```
ajwerner: nit: make an `init` method so that you don't reallocate here?
```

File: pkg/dyninst/gosym/symtab.go [hunk 3]
```diff
@@ -1180,3 +1193,86 @@ func stringFromOffsetPtr(data []byte, offset int) *string {
...
+		if newResolver == nil {
+			// If we managed to create the current FuncPCResolver, we should be
+			// able to re-create it.
+			panic("failed to re-initialize FuncPCResolver")
+		}
```
ajwerner: this feels like a wild panic given this is a public method. The caller could have passed a pc that's outside of the range of the function. Just return false.

pkg/dyninst/gosym/symtab.go
```diff
@@ -89,6 +89,16 @@ func (gst *GoSymbolTable) LocatePC(pc uint64) []GoLocation {
 	return gst.pclntab.locatePC(pc, gst.gofunc)
 }
 
+// FunctionPCResolver builds a FuncPCResolver for the function containing the
+// given program counter (any pc inside the function is valid). Returns nil if
+// the PC could not be resolved to a function.
+//
+// The FuncPCResolver can be used as an alternative to GoSymbolTable.LocatePC()
+// for more efficiently resolving multiple PCs within the same function.
+func (gst *GoSymbolTable) FunctionPCResolver(pc uint64) *FuncPCResolver {
```
ajwerner: I think that functions that are fallible but rely on nil checking is a trap. Why not just return errors?

grantseltzer:
Is there an application with this for the sake of decoding?

piob-io:
> Is there an application with this for the sake of decoding?

The only place I see where it could be used are recursive functions, but I don't think it would be practically useful.

pkg/dyninst/gosym/symtab.go
andreimatei: done

pkg/dyninst/gosym/symtab.go
andreimatei: done

pkg/dyninst/gosym/symtab.go
andreimatei: This resetting does not depend on the function argument (note  that we use `f.funcPC` for the reset, not `pc`), so returning `false` would be confusing. The resetting is guaranteed to succeed; the panic is an assertion.

pkg/dyninst/gosym/symtab.go
andreimatei: done

pkg/dyninst/gosym/symtab.go
```diff
@@ -1180,3 +1196,157 @@ func stringFromOffsetPtr(data []byte, offset int) *string {
 	result := string(data[offset:end])
 	return &result
 }
+
+// FuncPCResolver resolves program counters within a single function to source
+// line numbers. It is optimized for resolving multiple PCs in increasing order.
+type FuncPCResolver struct {
```
piob-io: I would keep exported things together at the top of the file, in particular this could go right above the FunctionPCResolver function, along with FuncPCResolver::PCToLine (which to keep things exported things short, could just call a private counterpart).

pkg/dyninst/gosym/symtab.go
```diff
@@ -1180,3 +1196,157 @@ func stringFromOffsetPtr(data []byte, offset int) *string {
...
+// EndLine returns the line number corresponding to the end of the function's
+// code.
+//
+// Note that figuring out the source line for the end of the function is not as
+// simple as PCToLine(<function's highPC>) because the last instructions in Go
+// functions relate to the stack growth and are actually mapped to the first
+// source line for the function. Instead, we iterate over the pclinetab to the
+// end of the function and keep track of the highest line number encountered,
+// while also ignoring inlined functions.
+func (f *FuncPCResolver) EndLine() uint32 {
```
piob-io: Current implementation of EndLine adds custom responsibilities to the rest of the class, making it more complicated and brittle (tracking the maxLineNumber). Could we export some more general functionality, say, to iterate over pcs or lines of a function, that could be used to implement EndLine at reasonably similar efficiency, with more isolated code responsibilities?

pkg/dyninst/gosym/symtab.go
piob-io, line 1: I think we should refactor existing functionality rather than just adding another one. I have several questions:
 - could existing implementation rely on new implementation - or vice versa? e.g. LocatePc looking up a function, and then call the new resolver to do the rest? 
 - could GoFunction be used to implement the new methods, rather than second struct that represent function (FunctionPcResolver)? or - if there is too much of shared stuff to put there, should the new functionality be exposed via methods on the linetable, that take function as an additional parameter?
 - should stepper be expanded to include currentPCRangeStart, inlinedPcRanges and InlinedPcRangesIdx to avoid having pieces of state that have strict invariants between them spread between different structures?

piob-io (commented):
Sorry for late, non-trivial comments. Usually when there are multiple reviewers I have suboptimal, sequential approach to the review process.

pkg/dyninst/gosym/symtab.go
andreimatei: mmm I don't know about that; I wouldn't split a type's exported methods from its unexported methods.
But I moved all the `PCIterator` code to a new file; `symtab.go` is big enough as it is.

pkg/dyninst/gosym/symtab.go
andreimatei: yes, this method is now gone

pkg/dyninst/gosym/symtab.go
andreimatei: I've rewritten and simplified the new struct as `PCIterator`; now, it's a pretty thin, exported, layer on top of the `stepper` that deals with skipping inlined functions. It no longer maintains state about the highest line it sees; that's left to the user. Also, I've pushed down some functionality into the `stepper` -- the ability to reset it, and keeping track of the start and end of the current pc range (as opposed to just the end).
I've considered teaching the stepper about inlined functions, but I don't think that's a good idea; I think it's better to keep it dumb and close to the disk data structures that it deals with.
I've made `GoFunction` produce `PCIterator`.

As such, there's not much left to integrate into the rest of the `lineTable` code, although:

> could existing implementation rely on new implementation - or vice versa? e.g. LocatePc looking up a function, and then call the new resolver to do the rest?

I've made `lineTable.pcToLine` (which is indirectly used by `GoSymbolTable.LocatePC`) use the new `PCIterator`.  I have *not* made the more generic `lineTable.pcValue()` use the `PCIterator` -- `pcValue()` can deal with any table, not just the line table, and `PCIterator` cannot at the moment...

piob-io: Great, thank you!

pkg/dyninst/gosym/pc_iterator.go
```diff
@@ -0,0 +1,153 @@
...
+
+func (f *PCIterator) Reset() {
+	f.stepper.reset()
+	f.inlinedPcRangesIdx = 0
+}
+
+// PCToLine returns the line number corresponding to the given program counter.
+// Only line numbers that correspond to the iterator's own function are
+// returned; if the PC corresponds to an inlined function, the returned line
+// number might not be meaningful.
```
piob-io: Far from being english expert - but I found this last sentence a bit confusing in regards to whether we actually return the inlined lines that aren't meaningful, or we don't because they aren't meaningful... The if conditions on a pc corresponding to an inline function, by that part is not the unreal part, it's that the returned line number is not actually returned...

How about:

"if the PC corresponds to an inlined function, 0, false is returned, as the line number might not be meaningful."

piob-io: Actually - if I read the code right, if there is a higher pc value, that isn't inlined, and isn't past the end of the searched range - we return corresponding line. But that line doesn't correspond to the argument pc. Is that intended? Should we bail out of the loop when pc < f.stepper.pcStart or separate the pcInInlinedFunc() check?

piob-io (approved):

pkg/dyninst/gosym/pc_iterator.go
andreimatei: > But that line doesn't correspond to the argument pc. Is that intended?

It was intended in the sense that I didn't think more handling is worth it, but I think you're right; I've now made it return `0, false`.

piob-io: Sg, yeah, I don't think we care what is returned, but returning 0 is safer.

pkg/dyninst/gosym/symtab.go
```diff
@@ -30,6 +30,12 @@ type GoFunction struct {
 	DeferReturn uint32
 	// The index of the function in the symbol table
 	idx uint32
+	// The internal function info
+	funcInfo *funcInfo
+}
+
+func (f *GoFunction) PCIterator(inlinedPcRanges [][2]uint64) (PCIterator, error) {
```
ajwerner, line 47: Given this is a public API, can you comment on what `inlinedPcRanges` is and how it is supposed to be computed?

ajwerner (approved):

pkg/dyninst/gosym/symtab.go
andreimatei: done

andreimatei:
Closing in favor of #38892

PR #26711, opened 2024-06-13
[Windows] Adding property in installer to support driver rollback

mrafi97:
### What does this PR do?
Adds the properties `DDDRIVERROLLBACK_NPM` and `DDDRIVERROLLBACK_PROCMON` to allow for rollback of `ddnpm` and `ddprocmon` drivers after a failed install/upgrade. 

To account for prior agent versions not having the uninstall rollback action, the installer will view previous and current versions and only perform rollbacks for newer agent versions.

### Motivation
https://datadoghq.atlassian.net/browse/WINA-726

### Describe how to test/QA your changes
1. Run the Datadog Agent MSI with the property `WIXFAILWHENDEFERRED=1` to induce an failed install/upgrade:
```
Start-Process -Wait msiexec -ArgumentList '/qn /i datadog-agent*.msi APIKEY="APIKEY" WIXFAILWHENDEFERRED=1'
```
2. Check running services:
```
Get-Service "datadog*" 
Get-Service "ddnpm" 
Get-Service "ddprocmon"
```
Expected results for failed install/upgrade:
- Fresh install: All services are removed
- Upgrade from 7.52, 7.54, 7.55: All services are still installed
- Upgrade from < 7.52: Only `ddprocmon` is removed

PR #38682, opened 2025-07-11
ignore errors trying to fetch MSI version

clarkb7:
### What does this PR do?
ignores errors from `MsiGetProductInfo` when trying to adjust rollback behavior to compensate for old MSI versions.

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1611
Don't fail install for the sake of rollback compatibility with old MSI versions

Adds unit test

Have existing E2E tests that validate MSI driver rollback behavior (see https://github.com/DataDog/datadog-agent/pull/26711)

Manual testing:
Can create similar Windows Installer registry corruption by hard power off the host or hard killing the `msiserver` service in the middle of an upgrade.
log snippets from this scenario
```
Action 14:30:44: FindRelatedProducts. Searching for related applications
Action start 14:30:44: FindRelatedProducts.
Action ended 14:30:44: FindRelatedProducts. Return value 1.
...
CA: 14:30:45: SetDDDriverRollback. WIX_UPGRADE_DETECTED: {876508B5-4B92-4A2E-AFDC-C501FF7B2069}
CA: 14:30:45: SetDDDriverRollback. Error setting DDDriverRollback, values may not be set correctly for the version being upgraded from: System.Exception: Failed to get version string for product {876508B5-4B92-4A2E-AFDC-C501FF7B2069} ---> System.ComponentModel.Win32Exception: This action is only valid for products that are currently installed
```

Tradeoff between behavior on error when upgrading from an old version vs a new version.
On error we default to assuming we are upgrading from an MSI version that supports driver rollback. If we're actually upgrading from an old MSI version (<7.56) and we fail/rollback, then the drivers will be missing and the install must be repaired. If we assume the opposite then the drivers would always be present on rollback, which reintroduces the issue this code is originally meant to fix https://github.com/DataDog/datadog-agent/pull/26711.

### Additional Notes

These errors "shouldn't" happen, but they do. They seem to be caused by a corrupt Windows Installer registry.

Screenshot from customer showing multiple Agent versions are "installed".
image

MSI log showing multiple Agent versions are "installed"
```
Action start 9:58:14: FindRelatedProducts.
FindRelatedProducts: Found application: {18CE1F1C-379B-4A20-A5B5-188BC533D2B9}
FindRelatedProducts: Found application: {C83199D2-EE9A-404D-B9C5-D1881CE611F2}
Action ended 9:58:14: FindRelatedProducts. Return value 1.
```

Attempting to uninstall these product codes fails because they are not really installed, they're just listed in the Windows Installer registry.

The [Microsoft Program Install and Uninstall troubleshooter](https://support.microsoft.com/en-us/topic/fix-problems-that-block-programs-from-being-installed-or-removed-cca7d1b6-65a9-3d98-426b-e9f927e1eb4d) can be used to fix these issues but it is being deprecated in 2025. Microsoft documents that some MSDT tools are being replaced, others are being removed, but this tool is not mentioned at all.
- [Microsoft Support Diagnostic Tool resources](https://learn.microsoft.com/en-us/windows/whats-new/deprecated-features-resources#microsoft-support-diagnostic-tool-resources)
- [Deprecation of Microsoft Support Diagnostic Tool (MSDT) and MSDT Troubleshooters](https://support.microsoft.com/en-us/windows/deprecation-of-microsoft-support-diagnostic-tool-msdt-and-msdt-troubleshooters-0c5ac9a2-1600-4539-b9d0-069e71f9040a)

We've contacted Microsoft support about a replacement and they've told us there is none. If this a concern to you as well, please contact Microsoft support and reference our tickets: 2402260010004360, 2409270040012495.

A workaround for fixing the Windows Installer registry that we've had success with is, for each extra product code/version, reinstall the MSI corresponding to that version and then uninstall it again.


releasenotes/notes/ignore-msi-upgrade-version-errors-7eb4b8a78704596b.yaml
```diff
@@ -0,0 +1,13 @@
+---
+fixes:
+  - |
+    The Agent MSI no longer fails when it is unable to read the
+    version information from [MsiGetProductInfo](https://learn.microsoft.com/en-us/windows/win32/api/msi/nf-msi-msigetproductinfow).
+
+    When upgrading from an Agent version lower than 7.56,
+    if ``MsiGetProductInfo`` fails and the MSI fails and rolls back then
```
brett0000FF:
```suggestion
    if ``MsiGetProductInfo`` fails and the MSI fails and rolls back, then
```

brett0000FF (approved):

tools/windows/DatadogAgentInstaller/AgentCustomActions/InstallStateCustomActions.cs
```diff
@@ -72,34 +72,51 @@ private void SetDDDriverRollback()
...
+                        }
                     }
                 }
             }
+            catch (Exception e)
+            {
+                _session.Log($"Error setting DDDriverRollback, values may not be set correctly for the version being upgraded from: {e}");
+                // Do not fail the installer here.
+                // We should only end up here in the Windows installer bug cases described above.
+                // In this case we assume that the version being upgraded from is a recent version that implements driver rollback.
```
alexn-dd, line 116: Is there a way to check for this assumption, since the exception handler seems broad?

alexn-dd (approved):

tools/windows/DatadogAgentInstaller/AgentCustomActions/InstallStateCustomActions.cs
clarkb7: It's meant to broadly cover any errors caused by `GetVersionString()` failing or returning unexpected output.
We have unit and e2e tests that ensure the function operates correctly in normal circumstances, so we shouldn't have to worry about it masking a loss in functionality.

PR #38696, opened 2025-07-11
[CHAOSPLT-874] Add return early for v0.5 traces

edznux-dd:
### What does this PR do?

Add an exit early when the msgpack payload is invalid, instead of further processing it.

### Motivation

This causes a limited, but high, and unnecessary allocation.

Added a "end to end" test, that parses the trace payload from the api handler.

It will start returning "nil" instead of a `pb.TracerPayload` when providing an invalid msgp payload.


pkg/trace/api/api.go
```diff
@@ -494,6 +494,9 @@ func decodeTracerPayload(v Version, req *http.Request, cIDProvider IDProvider, l
 		}
 		var traces pb.Traces
 		err = traces.UnmarshalMsgDictionary(buf.Bytes())
+		if err != nil {
+			return nil, err
+		}
```
ichinaski, line 498: nit: consider inlining the conditional with
```go
if err = traces.UnmarshalMsgDictionary(buf.Bytes()); err != nil {
    return nil, err
}
```

ichinaski (approved):

pkg/trace/api/api.go
edznux-dd: Done, thanks

PR #38713, opened 2025-07-11
Disable Windows Docker static quality gates

alopezz:
### What does this PR do?

Disables the static quality gates (size limits) for Windows Docker images, with the intention of reinstating them once the measurements are more reliable.

### Motivation

After having readjusted gate limits, we've seen the `static_quality_gate_docker_agent_windows1809_core_jmx` gate [fail on main](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1024875847), even when merging PR's that don't add any size to any other artifact.

### Additional Notes

- [ABLD-81](https://datadoghq.atlassian.net/browse/ABLD-81) already reported problems with the way we're measuring on-disk size for Windows Docker images.
- The rest of the gates provide fairly good guarantees that the size is not generally increasing, so we don't expect significant impact from this temporary deactivation. Increasing size on Docker artifacts while not increasing size elsewhere is unlikely to happen unless for very specific work.

[ABLD-81]: https://datadoghq.atlassian.net/browse/ABLD-81?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/static/static_quality_gates.yml
+++ b/test/static/static_quality_gates.yml
@@ -46,30 +46,30 @@ static_quality_gate_docker_agent_jmx_amd64:
 static_quality_gate_docker_agent_jmx_arm64:
   max_on_disk_size: 980.54 MiB
   max_on_wire_size: 324.2 MiB
-static_quality_gate_docker_agent_windows1809:
-  max_on_disk_size: 1190.61 MiB
-  max_on_wire_size: 420.48 MiB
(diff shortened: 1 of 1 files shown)


ofek:
Merging since the eligible reviewers are away currently.

PR #38731, opened 2025-07-11
dyninst/procmon: cache executable analysis

ajwerner:
### What does this PR do?
This PR adds caching to make the checking cheap in the case that we know the binary is not interesting. This code runs on every process exec, so it deserves to be optimized.

### Motivation

Analyzing an executable is not free. For a big binary it can be a few
ms. There are cases where executables fork children frequently or get
caught in loops and to incur milliseconds of CPU time per process launch
could be rather painful.

### DIFF
--- a/pkg/dyninst/procmon/analyze.go
+++ b/pkg/dyninst/procmon/analyze.go
@@ -16,8 +16,10 @@ import (
 	"path"
 	"strconv"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
+	"unsafe"
 
 	lru "github.com/hashicorp/golang-lru/v2"
 	"golang.org/x/time/rate"
@@ -48,15 +50,19 @@ func analyzeProcess(
 	resolver ContainerResolver,
 	executableAnalyzer *executableAnalyzer,
 ) (processAnalysis, error) {
+	maybeWrapErr := func(msg string, err error) error {
+		if err == nil || errors.Is(err, os.ErrNotExist) {
+			return nil
+		}
+		pid, msg, err := pid, msg, err
+		return fmt.Errorf("analyzeProcess: pid %d: %s: %w", pid, msg, err)
+	}
 
 	exeLinkPath := path.Join(procfsRoot, strconv.Itoa(int(pid)), "exe")
 	exePath, err := os.Readlink(exeLinkPath)
-	if errors.Is(err, os.ErrNotExist) {
-		return processAnalysis{}, nil
-	}
 	if err != nil {
-		return processAnalysis{}, fmt.Errorf(
-			"failed to open exe link for pid %d: %w", pid, err,
+		return processAnalysis{}, maybeWrapErr(
+			"failed to open exe link", err,
 		)
 	}
 
@@ -68,31 +74,29 @@ func analyzeProcess(
 			procfsRoot, strconv.Itoa(int(pid)), "root",
 			strings.TrimPrefix(exePath, "/"),
 		)
-		log.Debugf(
-			"exe for pid %d does not exist in root fs, trying under proc_root: %s",
-			pid, exePath,
-		)
+		if log.ShouldLog(log.TraceLvl) {
+			exePath := exePath
+			log.Tracef(
+				"exe for pid %d does not exist in root fs, trying under proc_root: %s",
+				pid, exePath,
+			)
+		}
 		exeFile, err = os.Open(exePath)
 	}
 	if err != nil {
-		if !errors.Is(err, os.ErrNotExist) {
-			err = fmt.Errorf("failed to open exe: %w", err)
-		} else {
-			err = nil
-		}
-		return processAnalysis{}, err
+		return processAnalysis{}, maybeWrapErr("failed to open exe", err)
 	}
 	defer exeFile.Close()
 
 	statI, err := exeFile.Stat()
 	if err != nil {
-		return processAnalysis{}, fmt.Errorf("process %d: failed to stat exe: %w", pid, err)
+		return processAnalysis{}, maybeWrapErr("failed to stat exe", err)
 	}
 	statSys := statI.Sys()
 	st, ok := statSys.(*syscall.Stat_t)
 	if !ok {
-		return processAnalysis{}, fmt.Errorf(
-			"process %d: failed to cast stat: %T", pid, statSys,
+		return processAnalysis{}, maybeWrapErr(
+			"failed to cast stat", fmt.Errorf("got %T", statSys),
 		)
 	}
 	fileKey := FileKey{
@@ -110,8 +114,8 @@ func analyzeProcess(
 
 	ddEnv, err := analyzeEnviron(int32(pid), procfsRoot)
 	if err != nil {
-		return processAnalysis{}, fmt.Errorf(
-			"failed to check if process %d is interesting: %w", pid, err,
+		return processAnalysis{}, maybeWrapErr(
+			"failed to analyze environ", err,
 		)
 	}
 	if ddEnv.serviceName == "" || !ddEnv.diEnabled {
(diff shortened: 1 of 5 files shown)


pkg/dyninst/procmon/process_monitor.go
```diff
@@ -94,11 +95,12 @@ func (pm *ProcessMonitor) NotifyExit(pid uint32) {
...
-		procfsRoot: procFS,
-		resolver:   resolver,
-		eventsCh:   make(chan event, 1024),
-		doneCh:     make(chan struct{}),
+		handler:            h,
+		procfsRoot:         procFS,
+		resolver:           resolver,
+		eventsCh:           make(chan event, 1024),
+		doneCh:             make(chan struct{}),
+		executableAnalyzer: makeExecutableAnalyzer(128),
```
tylfin: Curious why you picked 128 for the LRU cache size? You might also consider defining these values as consts to document the rationale

ajwerner: Thanks, I had some reasons but it was useful to type them back out. FWIW my back of the envelop of ~100bytes per entry was a bit off because the cache is sort of wasteful and the FileKey is bigger than I had guessed. I changed it to 64 and added words.

pkg/dyninst/procmon/analyze.go
```diff
@@ -296,3 +326,121 @@ func isGoElfBinaryWithDDTraceGo(f *os.File) (bool, error) {
...
+		return &baseExecutableAnalyzer{}
+	}
+	return newFileKeyCacheExecutableAnalyzer(
+		cacheSize,
+		newHtlHashCacheExecutableAnalyzer(cacheSize, &baseExecutableAnalyzer{}),
+	)
+}
+
+// mustNewLruCache panics if the cache creation fails which will only happen
+// if the size is non-positive.
```
grantseltzer, line 451: What's the advantage of having a panic-y constructor like this? Why not possibly return an error, or take a uint instead?

pkg/dyninst/procmon/analyze.go
```diff
@@ -296,3 +326,121 @@ func isGoElfBinaryWithDDTraceGo(f *os.File) (bool, error) {
...
+	isInteresting(f *os.File, key FileKey) (bool, error)
+}
+
+// baseExecutableAnalyzer implements executableAnalyzer without any caching.
+type baseExecutableAnalyzer struct{}
+
+func (a *baseExecutableAnalyzer) checkFileKeyCache(
+	_ FileKey,
+) (interesting bool, known bool) {
+	return false, false
```
grantseltzer, line 351: If there's no cache, I think it'd make more sense to default to all processes being interesting.

pkg/dyninst/procmon/htlhash.go
```diff
@@ -0,0 +1,72 @@
...
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+)
+
+// computeHtlHash computes a hash of the first 4096 bytes of the file and the
+// last 4096 bytes of the file, and the length of the file. This is a cheap to
+// compute hash that is extremely likely to be unique for the executable.
```
piob-io: Extremely likely is relative, I'd state what we need here, like "hash that is practically unique for the executable."

pkg/dyninst/procmon/htlhash.go
```diff
@@ -0,0 +1,72 @@
...
+//
+// If the function returns an error, the file will be left in an unknown state.
+// Otherwise, the file will have been seeked to the original position.
```
piob-io: Is this a significant optimization? Perhaps we shouldn't say anything, and callers should re-seek if they need to? It's a low level assumption, it's less brittle if it's not there.

pkg/dyninst/procmon/htlhash.go
```diff
@@ -0,0 +1,72 @@
...
+		}
+	}()
+	h := sha256.New()
+	if _, err := f.Seek(0, io.SeekStart); err != nil {
+		return "", maybeWrapErr("failed to seek to original position", err)
+	}
+	if _, err := io.Copy(h, io.LimitReader(f, 4096)); err != nil {
+		return "", maybeWrapErr("failed to copy file to hash", err)
+	}
+	size, err := f.Seek(0, io.SeekEnd)
```
piob-io, line 42: Huh, seeking to get a size... Should we be concerned about some files that don't have size? Can seeking to end ever get stuck, or Go has a notion of some "current end, even if file is allowed to grow further".

pkg/dyninst/procmon/process_monitor.go
```diff
@@ -91,14 +92,34 @@ func (pm *ProcessMonitor) NotifyExit(pid uint32) {
...
+// This value balances performance and memory usage:
+//   - Avoids redundant analysis when processes are created/destroyed frequently
+//   - Limits memory usage to approximately 32KiB for both caches combined
+//
+// Memory calculation (approximate):
+//
+//	FileKey: 32 bytes (16 for FileHandle + 16 for LastModified)
+//	HashCacheEntry: 32 bytes (16 for string header + 16 for string data)
+//	LRU overhead: ~48 bytes per entry
+//	Map overhead: ~16 bytes per entry
```
piob-io: Should there be a scaling factor due to expected map utilization?

pkg/dyninst/procmon/analyze.go
```diff
@@ -74,52 +74,70 @@ func analyzeProcess(
...
-		}
+	if ddEnv.serviceName == "" || !ddEnv.diEnabled {
+		log.Tracef(
+			"process %d is not interesting: service name is %q, %s=%t",
+			pid, ddEnv.serviceName, ddDynInstEnabledEnvVar, ddEnv.diEnabled,
+		)
+		return processAnalysis{interesting: false}, nil
+	}
+
+	isGo, err := executableAnalyzer.isInteresting(exeFile, fileKey)
```
piob-io, line 130: Should we skip analysis if executable is known and interesting?

piob-io (approved):
Neat!

pkg/dyninst/procmon/analyze.go
ajwerner: I check the size above the calls to this function. A uint isn't good enough, 0 will panic.

pkg/dyninst/procmon/analyze.go
ajwerner: Well no, `interesting` is only relevant if `known` is true. I've added commentary on the interface method.

pkg/dyninst/procmon/analyze.go
ajwerner: 🤷 the implementation of the analyzer checks the cache again

pkg/dyninst/procmon/htlhash.go
ajwerner: Done.

pkg/dyninst/procmon/htlhash.go
ajwerner: Fair, done.

pkg/dyninst/procmon/htlhash.go
ajwerner: I don't think this has anything to do with Go, this is about the linux syscalls. Seek is just a thin wrapper around https://linux.die.net/man/2/lseek. Generally I agree that it's suspect if the file grows or changes during this call. Elsewhere we should handle that by looking at mtime but I don't think this function is on the hook for that. I definitely don't think this can get stuck

pkg/dyninst/procmon/process_monitor.go
ajwerner: Sure. I was throwing that into "overhead" but fair enough. I added some more words to the calculation

pkg/dyninst/procmon/analyze.go
piob-io: I see, it's a bit surprising then that cache is checked directly for the non-interesting case, but is checked indirectly for the interesting case. Not a big deal, if that makes code simpler in some other aspect.

PR #38735, opened 2025-07-11
[AGENTCFG-319] fixing test Test_Configs

rahulkaukuntla:
### What does this PR do?
The test `Test_Configs` in the `pkg/snmp` package is failing in ntm as ntm was merging previously generated configs instead of starting from scratch as viper does. This PR ensures that ntm acts like viper in this regard, and that warnings are actually configured for viper.


pkg/config/nodetreemodel/read_config_file.go
```diff
@@ -130,20 +133,16 @@ func loadYamlInto(dest InnerNode, source model.Source, inData map[string]interfa
 
 		// check if the key is defined in the schema
 		schemaChild, err := schema.GetChild(key)
-		if err != nil {
+		// schemaChild might be nil if not found or error
+		unknownKey := err != nil
+		if unknownKey {
```
hush-hush:
```suggestion
		if err != nil {
```

hush-hush (approved):

pkg/config/nodetreemodel/read_config_file.go
```diff
@@ -64,6 +64,9 @@ func (c *ntmConfig) ReadConfig(in io.Reader) error {
 	c.Lock()
 	defer c.Unlock()
 
+	// Reset the file tree like Viper does, so previous config is cleared
+	c.file = newInnerNode(nil)
+
```
hush-hush, line 72: Could you also update the docstring for `ReadConfig` and `ReadInConfig` to remove the mention to viper and highlight that `ReadConfig` reset the data for the file layer while `ReadInConfig` does not.

pkg/config/nodetreemodel/read_config_file.go
```diff
@@ -64,6 +64,9 @@ func (c *ntmConfig) ReadConfig(in io.Reader) error {
 	c.Lock()
 	defer c.Unlock()
 
+	// Reset the file tree like Viper does, so previous config is cleared
+	c.file = newInnerNode(nil)
```
hush-hush, line 71: Should we call `mergeAllLayers` for this method too ?

rahulkaukuntla: If you scroll down to the bottom of this method, you can see that we return a call to `mergeAllLayers` already 😅

pkg/config/nodetreemodel/read_config_file.go
rahulkaukuntla: Done.

pkg/config/nodetreemodel/read_config_file.go
rahulkaukuntla: Docstrings have been updated.

pkg/config/nodetreemodel/read_config_file.go
```diff
@@ -31,7 +31,7 @@ func (c *ntmConfig) findConfigFile() {
 	}
 }
 
-// ReadInConfig wraps Viper for concurrent access
+// ReadInConfig reads the configuration from the file system, without resetting the file tree.
```
dustmop: Turns out, both ReadConfig and ReadInConfig will reset the file layer in viper. You've correctly identified that nodetreemodel has a bug where it doesn't reset the file in ReadConfig, but the only difference between ReadConfig and ReadInConfig should be how they are receiving the file.

Could you add tests to [this file](https://github.com/DataDog/datadog-agent/blob/729c1eafea6b0953dd34e27d41ab3be0f47889c6/pkg/config/nodetreemodel/compatibility_test.go) that demonstrates ReadConfig and ReadInConfig work the same (in both viper and ntm) with regards to resetting the file?

pkg/config/nodetreemodel/read_config_file.go
```diff
@@ -130,20 +133,15 @@ func loadYamlInto(dest InnerNode, source model.Source, inData map[string]interfa
 
 		// check if the key is defined in the schema
 		schemaChild, err := schema.GetChild(key)
+		// schemaChild might be nil if not found or error
```
dustmop: No need for this comment: this is a standard idiom in golang. For any function that returns a value and an error, that value is expected to be invalid if the error is non-nil.

pkg/config/nodetreemodel/read_config_file_test.go
```diff
@@ -212,7 +212,8 @@ c:
 			"a": &leafNodeImpl{val: "orange", source: model.SourceFile},
 			"c": &innerNode{
 				children: map[string]Node{
-					"d": &leafNodeImpl{val: 1234, source: model.SourceFile},
+					"d":       &leafNodeImpl{val: 1234, source: model.SourceFile},
+					"unknown": &leafNodeImpl{val: "key", source: model.SourceFile},
```
dustmop: This test is using a strict schema (which is what gets run in prod / non-test code) so it should not have unknown nodes showing up in the file layer. The cause of this change is due to what's being modified in `ReadConfig`.

pkg/config/nodetreemodel/read_config_file.go
```diff
@@ -130,20 +133,15 @@ func loadYamlInto(dest InnerNode, source model.Source, inData map[string]interfa
...
 		if err != nil {
 			warnings = append(warnings, fmt.Errorf("unknown key from YAML: %s", currPath))
 			if !allowDynamicSchema {
-				continue
```
dustmop: This change shouldn't be here. Its modifying how unknown content in the file layer is handled. While that feature needs fixing, its unrelated to the change in this PR so it's best to save it for a later more comprehensive fix.

Specifically, skipping this check for `allowDynamicSchema` is the reason you need to add "unknown" to the test `TestReadSingleFile`

pkg/config/nodetreemodel/read_config_file_test.go
```diff
@@ -70,11 +70,11 @@ func TestReadConfig(t *testing.T) {
 	err = cfg.ReadConfig(strings.NewReader(confYaml2))
 	require.NoError(t, err)
 
-	assert.Equal(t, true, cfg.GetBool("network_devices.snmp_traps.enabled"))
+	assert.Equal(t, false, cfg.GetBool("network_devices.snmp_traps.enabled")) // no longer set
```
dustmop: This is good and expected now that you've fixed how `ReadConfig` resets the file layer. However, instead of leaving the trailing comment after each line, how about putting a comment before this block of checks explaining what these assertions demonstrate. Having trailing comments looks okay in this diff because the context is clear, but won't work as well for someone looking at this code on a later date.

pkg/snmp/snmp_test.go
```diff
@@ -526,6 +527,8 @@ network_devices:
 	assert.Equal(t, "legacyCommunityString", legacyConfig.Community)
 	assert.Equal(t, "legacySnmpVersion", legacyConfig.Version)
 	assert.Equal(t, "127.2.0.0/30", legacyConfig.Network)
+	warnings := pkgconfigsetup.Datadog().Warnings()
```
dustmop: Instead of using the global `pkgconfigsetup.Datadog()` you can change line 505 to this:

```
corecfg := configmock.NewFromYAML(t, `
```

Then here use:

```
warnings := corecfg.Warnings()
```

pkg/snmp/snmp_test.go
```diff
@@ -526,6 +527,8 @@ network_devices:
...
+	assert.NotNil(t, warnings)
```
dustmop: Could you add a method `Count()`to `pkg/config/model/warnings.go`. It should return the numbers of `errors` that the `Warnings` object has. Compare that count here instead of checking that the object is non-nil. I believe it should be zero since my initial assumption (that there was an error being swallowed by the code) seems to be wrong.

pkg/config/structure/compatibility_test.go
```diff
@@ -91,7 +91,7 @@ network_devices:
 	})
 
 	warnings := viperConf.Warnings()
-	assert.Nil(t, warnings)
+	assert.Equal(t, &model.Warnings{TraceMallocEnabledWithPy2: false, Errors: []error(nil)}, warnings)
```
dustmop: Could you add a function `NewWarnings` to `pkg/config/model/warnings.go`, it should take an `[]error` to build the Warnings object. That way we can avoid exposing details about it here.

pkg/config/structure/unmarshal.go
```diff
@@ -201,7 +202,9 @@ func copyStruct(target reflect.Value, input nodetreemodel.Node, currPath []strin
 
 			err := copyAny(target.FieldByName(f.Name), input, nextPath, fs)
 			if err != nil {
-				return err
+				warning := fmt.Sprintf("Failed to parse key %q at %v: %v", fieldKey, strings.Join(nextPath, "."), err)
+				log.Warn(warning)
+				continue // don't fail, just warn
```
dustmop: Is this change required, or otherwise covered by any of the tests in this change? It seems at first glance to be unrelated, but maybe I'm missing it.

dustmop (changes requested):
Sorry for taking a bit to review this. Started looking into how exactly `ReadConfig` and `ReadInConfig` work, and while doing so I got lost in details. I think there's going to be some follow-up work, especially around error handling, in order to improve compatibility with Viper.

pkg/config/nodetreemodel/read_config_file.go
rahulkaukuntla: I've changed the function string for `ReadInConfig` to say that the file tree is being reset as well. I have added two tests to the bottom of `compatibility_test.go`--one to check that ReadConfig is the same between ntm and viper, and one for ReadInConfig.

pkg/config/nodetreemodel/read_config_file.go
rahulkaukuntla: Removed.

pkg/config/nodetreemodel/read_config_file_test.go
rahulkaukuntla: Reverted this change.

pkg/config/nodetreemodel/read_config_file.go
rahulkaukuntla: Reverted this change as well.

pkg/config/nodetreemodel/read_config_file_test.go
rahulkaukuntla: More context is always good--i've done as you suggested, and explained why those options weren't being set anymore.

pkg/snmp/snmp_test.go
rahulkaukuntla: Done.

pkg/snmp/snmp_test.go
rahulkaukuntla: Added the Count() method and used it here.

pkg/config/structure/compatibility_test.go
rahulkaukuntla: Added the NewWarnings() method as suggested, and used it here.

pkg/config/structure/unmarshal.go
rahulkaukuntla: I don't remember 😂. I've removed it, anyways.

pkg/config/model/warnings.go
```diff
@@ -10,3 +10,16 @@ type Warnings struct {
...
 	Errors                    []error
 }
+
+// NewWarnings creates a new Warnings instance
+func (w *Warnings) NewWarnings(errors []error) *Warnings {
+	return &Warnings{
+		TraceMallocEnabledWithPy2: w.TraceMallocEnabledWithPy2,
+		Errors:                    append(w.Errors, errors...),
+	}
+}
```
dustmop: It would be more idiomatic to make this constructor a stand-alone function instead of a member function:

```suggestion
func NewWarnings(errors []error) *Warnings {
	return &Warnings{
		TraceMallocEnabledWithPy2: false,
		Errors:                    errors,
	}
}
```

pkg/config/structure/compatibility_test.go
```diff
@@ -91,7 +91,9 @@ network_devices:
 	})
 
 	warnings := viperConf.Warnings()
-	assert.Nil(t, warnings)
+	warnings.NewWarnings([]error{nil})
+	assert.Equal(t, warnings.NewWarnings(nil), warnings)
```
dustmop:
```suggestion
	assert.Equal(t, model.NewWarnings(nil), warnings)
```

dustmop (approved):
LGTM, with a few small suggestions. After fixing them, this is good to merge!

FlorianVeaux (approved):

PR #38737, opened 2025-07-11
[Serverless-Init] Support Cloud Run Jobs with serverless-init

nhulston:
### What does this PR do?

Support Cloud Run Jobs in serverless-init. So we get a variety of features:
- stdout & stderr piping for in-process instrumentation
- setting logs source to `cloudrun`
- add tags to spans: job name, execution name, task index, task attempt #, task count, and resource name. Looks like we already get container_id (which is some long hex string) out of the box by the agent
Screenshot 2025-07-11 at 4 43 13 PM
<br/>
Screenshot 2025-07-11 at 4 43 20 PM

Added logic to skip cold start & shutdown metrics, since they're irrelevant for Cloud Run jobs (every execution will have a "cold start", every execution will have a shutdown).

I follow the GCP convention and prefix logs with `gcp.run.job` instead of `gcp.run`.

Tested manually and unit tests

Screenshot 2025-07-11 at 3 56 21 PM

### Additional Notes

Jobs kind of work with the sidecar approach, but the flushing logic is sketchy. When the main app is done executing, all other containers immediately receive a `SIGKILL`, so any pending traces or logs will not be sent.

Therefore, we will be recommending customers use the in-process approach in our documentation.


brett0000FF (approved):

cmd/serverless-init/cloudservice/cloudrun_jobs.go
```diff
@@ -0,0 +1,91 @@
...
+	resourceNameTag      = "resource_name"
+)
+
+// CloudRunJobs has helper functions for getting Google Cloud Run data
+type CloudRunJobs struct{}
+
+// GetTags returns a map of gcp-related tags for Cloud Run Jobs.
+func (c *CloudRunJobs) GetTags() map[string]string {
+	tags := metadataHelperFunc(GetDefaultConfig(), false)
+	tags["origin"] = CloudRunJobsOrigin
```
nina9753: I think this can be removed. We wanted to make origin a hidden tag, but in case some customers were using it, we needed to be backward compatible. Since this is a new product, we can exclude it here

nhulston: good call, thanks for catching this!

cmd/serverless-init/main.go
```diff
@@ -109,7 +109,10 @@ func run(_ secrets.Component, _ autodiscovery.Component, _ healthprobeDef.Compon
 	prefix := cloudService.GetPrefix()
 	origin := cloudService.GetOrigin()
 
-	metric.AddShutdownMetric(prefix, origin, metricAgent.GetExtraTags(), time.Now(), metricAgent.Demux)
+	// Don't emit shutdown metric for Cloud Run Jobs
+	if _, ok := cloudService.(*cloudservice.CloudRunJobs); !ok {
```
apiarian-datadog: nit: this sort of thing is usually a smell, adding service-specific logic gated by an if statement checking the service's struct type.

can we move this logic into the cloudService class, with the cloud run job implementation as a noop? or at the very least, can we add a `shouldEmitShutdownMetric` method to the cloud run services so that we consolidate the service-specific logic in with the cloud service implementation?

(i know we have other things like this already. i think i've seen some azure-specific code sprinkled around, but let's not propagate unfortunate patterns if we don't have to.)

apiarian-datadog: actually, not a nit. let's do this differently. this and the cold start metric as well.

nina9753: We emit this enhanced metric for Azure and GCP. I agree we can move this into cloud service, but we were also talking about adding cloud run job specific enhanced metrics relating to the jobs and tasks, like AWS step functions. I'll leave it up to Nick if he wants to do this in a separate PR or in this one

nhulston: Yes, since we're going to emit an enhanced metric, we can rename `AddColdStartMetric` to `AddStartMetric`, which will be implemented by each service. Same with shutdown. wdyt?

cmd/serverless-init/cloudservice/service.go
```diff
@@ -57,6 +67,10 @@ func GetCloudServiceType() CloudService {
 		return &CloudRun{spanNamespace: cloudRunService}
 	}
 
+	if isCloudRunJob() {
```
apiarian-datadog, line 73: probably not in this pr, but this whole if ladder is a bit funky since it leaks the individual service details into this overall interface. might be nice to refactor this. but if you're not going to be making any more changes, maybe add a note for us to do it in the future?

apiarian-datadog (approved):
looks fine to me!

cmd/serverless-init/cloudservice/service.go
nhulston: i agree, that's wonky. left a comment

PR #38738, opened 2025-07-11
Corecheck Filter Migration

gabedos:
### What does this PR do?

Migrates container corecheck filtering mechanism to use filtering component

### Motivation

Unify filtering for corechecks. [CONTP-844]

Deploy the Agent and Cluster Agent with a filtering option to exclude the DCA itself from collection.
```
datadog:
  kubelet:
    tlsVerify: false
  clusterName: <INSERT_NAME>
  envDict:
    DD_CONTAINER_EXCLUDE: "name:cluster-agent"
```

In both scenarios, we should expect the `containers.running` metric to not filter out the `cluster-agent`. However, for `kubernetes.liveness_probe.success.total` and `kubernetes.kubelet.container.log_filesystem.used_bytes` you should expect no container tag for `cluster-agent`.

Legacy:
image

New:
image

Note: during the migration process, we noticed that the kubelet summary provider was not filtering it's collection based on autodiscovery pod annotations whereas all the other providers did. This unifies the experiences for users but may lead to a subset of missing metrics that should be present in the first place. The summary provider does give a lot of data on pod/container memory/cpu/network usage.

This has been discussed with container-integrations and agree that annotation filter should be used in this provider. If we think we want to maintain backwards compatibility, we can manually remove the AD annotation filters from being used in that particular client call.

### Additional Notes

To reduce redundancies with generated filters, at least for now we can collapse the image filters in the container filters. Users will still be able to define filters for images, but behind the scenes, it will just be treated as a container with the only attribute being the image. This works because the scope of image filtering is a subset of the container filters.

[CONTP-844]: https://datadoghq.atlassian.net/browse/CONTP-844?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


comp/core/workloadfilter/catalog/pod.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// PodADAnnotationsMetricsProgram creates a program for filtering pods metrics based on AD annotations
+func PodADAnnotationsMetricsProgram(_ config.Component, logger log.Component) program.CELProgram {
+	programName := "PodADAnnotationsMetricsProgram"
+
+	var initErrors []error
+	// Use 'in' operator to safely check if annotation exists before accessing it
+	excludeFilter := `(("ad.datadoghq.com/metrics_exclude") in pod.annotations && 
+		 pod.annotations["ad.datadoghq.com/metrics_exclude"] in ["1", "t", "T", "true", "TRUE", "True"])`
```
davidor: The array with the truthy values is used a few times in this component.
I think it makes sense to extract it into a variable and reuse it.

pkg/collector/corechecks/containers/docker/events.go
```diff
@@ -44,7 +44,8 @@ func (d *DockerCheck) retrieveEvents(du docker.Client) ([]*docker.ContainerEvent
 	if d.lastEventTime.IsZero() {
 		d.lastEventTime = time.Now().Add(-60 * time.Second)
 	}
-	events, latest, err := du.LatestContainerEvents(context.TODO(), d.lastEventTime, d.containerFilter)
+	// TODO: update the docker util to use the workloadfilter component
```
davidor, line 47: Should we add a card in our board to remember about this?

davidor (approved):

paulcacheux (approved):

dustmop (approved):

jeremy-hanna (approved):
👍 for agent-runtime files

sarah-witt (approved):

buraizu (approved):

louis-cqrl (approved):

wdhif (approved):

pkg/collector/corechecks/containers/docker/check.go
```diff
@@ -234,15 +237,31 @@ func (d *DockerCheck) runDockerCustom(sender sender.Sender, du docker.Client, ra
...
+				},
+				Runtime: workloadmeta.ContainerRuntimeDocker,
+				State: workloadmeta.ContainerState{
+					Running: rawContainer.State == string(workloadmeta.ContainerStatusRunning),
+				},
+			}
 		}
+		pod, _ := d.store.GetKubernetesPodForContainer(rawContainer.ID)
+
+		isContainerExcluded := d.filterStore.IsContainerExcluded(workloadmetafilter.CreateContainer(wmetaContainer, workloadmetafilter.CreatePod(pod)), workloadfilter.GetContainerSharedMetricFilters())
```
sblumenthal: I have some concerns about the change in the way we are interacting with the container filter in this PR. 

This is one cherry-picked example showing the, apparent, "correct" way to check if a container is excluded, albeit an extreme one.

pkg/collector/corechecks/sbom/processor.go
```diff
@@ -149,7 +152,7 @@ func (p *processor) processContainerImagesEvents(evBundle workloadmeta.EventBund
 			container := event.Entity.(*workloadmeta.Container)
 			p.registerContainer(container)
 
-			if containerFilter.IsExcluded(nil, container.Name, container.Image.Name, "") {
+			if p.filterStore.IsContainerExcluded(workloadmetafilter.CreateContainer(container, nil), [][]workloadfilter.ContainerFilter{{workloadfilter.ContainerFilter(workloadfilter.LegacyContainerSBOM)}}) {
```
sblumenthal: This is another, less extreme, example of the change resulting in a setup that isn't necessarily easy to read and potentially contains a lot of hidden complexity which could introduce risk when trying to correctly query this function

sblumenthal (commented):
I have some concerns about the user experience of developers trying to interact with the container filter after this change

I worry that it is not exactly clear the number of steps needed to take ahead of time before being able to check the filter, and the verbosity of the steps involved here makes me worry that we are inviting user error when trying to build checks which rely on the container filter in any way. Couple that with the fact that I do not see any clear instructions in either docstrings or in a package level readme on how to properly use this package, it makes me wonder if there is a cleaner middleground which could preserve not just the old behavior, but the spirit of the simplicity in how it was used.

pkg/collector/corechecks/sbom/processor.go
gabedos: Hey Steve, totally fair point — the one-liners are pretty verbose. I’ve definitely gotten a bit too close to them and may have developed some tunnel vision.

That said, I do think the interface itself communicates intent pretty well but it does become verbose and blurry in practice:
```
// IsContainerExcluded returns true if the container is excluded by the selected container filters.
IsContainerExcluded(container *Container, containerFilters [][]ContainerFilter) bool
...
p.filterStore.IsContainerExcluded(workloadmetafilter.CreateContainer(container, nil), [][]workloadfilter.ContainerFilter{{workloadfilter.ContainerFilter(workloadfilter.LegacyContainerSBOM)}})
```

The verbosity partially comes from the fact that the Container being filtered isn’t the same as the workloadmeta one. We wanted to avoid circular dependencies, limit the full set of exposed fields while also making it easier to expose new ones, and optimize the validation path for filter inputs.

One way to reduce the visual complexity here could be to extract the filterable container ahead of time:
```
filterableContainer := workloadmetafilter.CreateContainer(container, nil)
```

As for `[][]ContainerFilter` I agree in practice it doesn't look lightweight either. We can reduce that by exposing constants or typed getters where it makes sense. We already do that for core filters like GetContainerSharedMetricFilters, GetPodSharedMetricFilters, and GetAutodiscoveryFilters.

The hope is that with a bit of restructuring, we land on something closer to:

```
filterableContainer := workloadmetafilter.CreateContainer(container, nil)
selectedFilters := workloadfilter.GetContainerSBOMFilters()
if p.filterStore.IsContainerExcluded(filterableContainer, selectedFilters) {
    ...
}
```
This version reads more naturally and makes the filtering logic easier to follow at the call site.

In terms of usability, one upside of this newer interface is that it makes the source of the filter config more transparent. With the old `d.containerFilter`, you had to trace back through several layers to where it was initilalized to know which filtering config input was being used. Now, you can see it right where the decision is made which I think helps reduce hidden complexity in a different way.

I’m also actively working on documentation to help bridge the gap and make the usage patterns clearer! Happy to talk tomorrow further about this

sblumenthal (approved):
Synced offline about the change, going to approve what is here

PR #38741, opened 2025-07-13
USM: Add invalid HTTP status code check and telemetry

DanielLavie:
### What does this PR do?
Introduces validation for HTTP status codes and adds telemetry within USM HTTP monitoring.

### Motivation
Ensure early failure on invalid status codes and improve observability around such occurrences.

### DIFF
--- a/pkg/network/protocols/http/statkeeper.go
+++ b/pkg/network/protocols/http/statkeeper.go
@@ -160,6 +160,16 @@ func (h *StatKeeper) add(tx Transaction) {
 		return
 	}
 
+	// Validate HTTP status code
+	statusCode := tx.StatusCode()
+	if statusCode < 100 || statusCode > 599 {
+		h.telemetry.invalidStatusCode.Add(1)
+		if h.oversizedLogLimit.ShouldLog() {
+			log.Warnf("invalid status code: %s", tx.String())
+		}
+		return
+	}
+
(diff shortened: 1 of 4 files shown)


pkg/network/protocols/http/statkeeper.go
```diff
@@ -160,6 +160,16 @@ func (h *StatKeeper) add(tx Transaction) {
...
 
+	// Validate HTTP status code
+	statusCode := tx.StatusCode()
+	if statusCode < 100 || statusCode > 599 {
+		h.telemetry.invalidStatusCode.Add(1)
+		if h.oversizedLogLimit.ShouldLog() {
+			log.Warnf("invalid status code: %s", tx.String())
+		}
+		return
+	}
```
guyarb: That already exists in AddRequest - https://github.com/DataDog/datadog-agent/blob/abb5cfcf3a7e368e1313ba1a48f9f3e35813e2ae/pkg/network/protocols/http/stats.go#L208-L210

Do we need to move the original condition to here? Is there a good reason for moving it?

DanielLavie: We can leave the validation in StatKeeper (telemetry + early exit) and keep the original guard inside AddRequest as a cheap last-defence

guyarb (approved):

PR #38747, opened 2025-07-14
Skip debian 10 on ddot install tests

alopezz:
### What does this PR do?

It skips the Debian 10 platform on tests for ddot installation.

### Motivation

#incident-40625.

Debian Buster repositories stopped being available in the default mirrors last Saturday, breaking every test that tries to `apt update` or similar on Debian 10 (buster).

This remediation will minimize disruption for contributors while we fix the underlying cause (the mirror setup).


rdesgroppes (approved):

alopezz:
Force-merging after the relevant tests have passed to unblock.

PR #38785, opened 2025-07-15
Also `codesign` non Mach-O executables (ABLD-94 follow-up)

rdesgroppes:
### What does this PR do?
Because there's no point in `codesign`ing non-executable scripts (those meant to be launched through `bash script.sh`, `python script.py`, etc.), the present change simply consists in `codesign`ing all files having some executable bits set in addition to formely considered Mach-O binaries.

### Motivation
While looking for good practices wrt what to `codesign` or not (see Datadog/datadog-agent#38755), I stumbled upon the following archived piece of [documentation](https://developer.apple.com/library/archive/technotes/tn2206/_index.html#//apple_ref/doc/uid/DTS40007919-CH1-TNTAG201):
> Store Python, Perl, shell, and other script files and other non-Mach-O executables in your app's `Contents/Resources` directory.
> [...]
> Put another way, a properly-signed app that has all of its files in the correct places will not contain any signatures stored as extended attributes.

Since we don't happen to store our non Mach-O executables under `Contents/Resources` but instead under `Contents/Payload`, we're better off `codesign`ing them.

### Additional Notes
The speedup (from ~14s down to ~11s) comes from the fact there are now less files being scanned to infer their MIME type, plus many Mach-O binaries also have some executable bits set.

Contrarily to PNGs being returned in the notarization log response despite not being explicitly `codesign`ed, non Mach-O executables may not be returned in that response despite being `codesign`ed.

### DIFF
--- a/omnibus/config/software/datadog-agent-finalize.rb
+++ b/omnibus/config/software/datadog-agent-finalize.rb
@@ -180,13 +180,17 @@ build do
                 # Sometimes the timestamp service is not available, so we retry
                 codesign = "../tools/ci/retry.sh codesign"
 
-                # Codesign all Mach-O binaries leveraging parallelism (~280 binaries out of ~28000 files)
+                # Codesign ~480 files (out of ~28000)
                 command <<-SH.gsub(/^ {20}/, ""), cwd: Dir.pwd
                     set -euo pipefail
-                    find #{install_dir} -type f -print0 |
-                        xargs -0 -n1000 -P#{workers} file -n --mime-type |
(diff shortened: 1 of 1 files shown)


alopezz (approved):

rdesgroppes:
Minor: I just rebased on top of `main` & [tweaked comments](https://github.com/DataDog/datadog-agent/compare/d5ee163dd0b53589df307dadd89d6855718c7480..7142f6be39a0a90b11027008bf29c3922a3f8065#diff-95bc93ec240754688a825e84ffb75c6b11ebd1619fab70f2dfee148f95213831) (no change in logic)

rdesgroppes:
I had to re-rebase on top of `main`.

rdesgroppes:
Hopefully, that was the last needed rebase...

PR #38787, opened 2025-07-15
Remove system probe dependency in heroku build job

pgimalac:
### What does this PR do?
Remove system probe dependency in heroku build job.

### Motivation
Heroku build doesn't use system-probe.


alopezz (approved):
Good catch, thanks!

PR #38789, opened 2025-07-15
[SBOM] Add /usr/share/rpm/* to the list of allowed folders

lebauce:
On RHEL Edge, /usr/lib/rpm is a symlink and is therefore ignored by the walker used in SBOM generation.


safchain (approved):

0intro:
`/usr/lib/rpm` or `/var/lib/rpm` is a symlink?

I suppose this requires this change in Trivy:

https://github.com/DataDog/trivy/commit/0ecdc33e288279fc226a9941ea27b6a084cba0b3

0intro:
Look good to me after the Trivy bump.

0intro (approved):

PR #38793, opened 2025-07-15
[FA] Support removing files through fleet

coignetp:
### What does this PR do?

Support removing files with fleet-automation

### Motivation

We'll extend the config feature to apply multiple configs
With this `remove` part, we can override instances


pkg/fleet/installer/installer.go
```diff
@@ -852,6 +868,7 @@ func (i *installerImpl) writeConfig(dir string, rawConfig []byte) error {
 		if err != nil {
 			return fmt.Errorf("could not write config file: %w", err)
 		}
+		fmt.Println("wrote config file", filepath.Join(dir, file.Path))
```
BaptisteFoy: nit: leftover

pkg/fleet/installer/installer.go
```diff
@@ -945,3 +962,32 @@ func ensureRepositoriesExist() error {
 
 	return nil
 }
+
+// mergeConfigs merges multiple config files into a single map of config files.
```
BaptisteFoy, line 965: q: rawConfigs contains Remote Config configs, and we output what we'd write on disk, right?

coignetp: yes exactly, I'll add a comment to make it clearer

BaptisteFoy (approved):

clarkb7 (approved):

PR #38799, opened 2025-07-15
[EBPF] gpu: support GKE device plugin IDs

gjulianm:
### What does this PR do?

This PR adds support for GKE device plugin way of reporting GPU device IDs, which returns `nvidiaX` instead of the device UUID.

### Motivation

Support GKE. https://datadoghq.atlassian.net/browse/NWL-1334

Added unit tests to cover these cases.

### Additional Notes

In order to make the code more easily testable, and also in preparation for Docker support where this logic might need to be changed, the functions have been moved to standalone functions in a `pkg/gpu/containers.go` file.

MIG will not be supported in this PR, if the GKE cluster has MIG devices there might be wrong assignments or errors when matching devices to containers.

### DIFF
--- a/pkg/gpu/containers.go
+++ b/pkg/gpu/containers.go
@@ -1,0 +1,93 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-present Datadog, Inc.
+
+// this file contains utilities to work with GPU assignment to containers
+
+//go:build linux_bpf && nvml
+
+package gpu
+
+import (
+	"errors"
+	"fmt"
+	"regexp"
+	"strconv"
+
+	workloadmeta "github.com/DataDog/datadog-agent/comp/core/workloadmeta/def"
+	ddnvml "github.com/DataDog/datadog-agent/pkg/gpu/safenvml"
+	gpuutil "github.com/DataDog/datadog-agent/pkg/util/gpu"
+)
+
+var errCannotMatchDevice = errors.New("cannot find matching device")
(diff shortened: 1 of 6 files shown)


pkg/gpu/containers.go
```diff
@@ -0,0 +1,93 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-present Datadog, Inc.
+
+// this file contains utilities to work with GPU assignment to containers
+
+//go:build linux_bpf && nvml
```
val06, line 8: why do we need `linux_bpf` build tag here?

val06 (approved):
left 1 comment

pkg/gpu/containers.go
gjulianm: The functions are not exported, and only used from code that has the `linux_bpf` tag, without these we get linter errors.

releasenotes/notes/gpu-system-probe-gke-0c444e01966d6913.yaml
```diff
@@ -0,0 +1,11 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+fixes:
+  - |
+    GPUM: fix kubernetes device allocation detection in Google Kubernetes Engine
```
janine-c:
```suggestion
    GPUM: fix Kubernetes device allocation detection in Google Kubernetes Engine
```

janine-c (approved):

PR #38593, opened 2025-07-09
[EBPF] Split GPU core-check and GPU Probe

val06:
### What does this PR do?

Decouples GPU core-check in the core-agent from the System-Probe GPUM module.

Config changes:

**Agent Platform capabilities:**
- wmeta and autodiscovery will be **enabled by default** (today under disabled feature flag `enable_nvml_detection`)
- still could be disabled by adding "nvml" under the existing `autoconfig_exclude_features` config flag of the agent

**System-Probe - GPU Probe**
(ALREADY EXISTS) will have its own config flag (gpu_monitoring.enabled)

**Core-Agent - GPU Core-Check**
(NEW FLAG) will be enabled via a dedicated config flag: `gpu.enabled`, **disabled by default**

### Motivation

simplify deployment of GPUM product, by providing basic version without the need for System-Probe component

### Describe how you validated your changes

- [x]  added UTs, updated existing UTs, following refactor
- [x]  manual tests on single and multi GPU hosts
- [x]  will deploy a custom image to staging

PR #38801, opened 2025-07-15
[EBPF] gpu: disable autodiscovery provider for the GPU check

gjulianm:
### What does this PR do?

Remove the GPU autodiscovery provider.

### Motivation

With https://github.com/DataDog/datadog-agent/pull/38593 we move to remove the `enable_nvml_detection` feature flag and rely on a specific flag to enable GPUM in the agent. This means that we don't want to enable the GPU check via autodiscovery or the sysprobe-check script anymore.

Feature completely removed. E2E tests ensure we still run the check properly, and also tested manually that the check runs when it needs to.


rdesgroppes (approved):
LGTM

dustmop (approved):

ofek (approved):

adel121 (approved):

PR #38805, opened 2025-07-15
[OTAGENT-455] Create ddot OCI packages

tim-liakh-datadog:
### What does this PR do?
Adding OCI images for DDOT. This PR creates a system that automatically generates the newest OCI image through leveraging the datadog-installer and then deploys it to local registries. Once there, it is possible to use datadog-installer to download DDOT on a linux bare-metal machine.
### Motivation
This is the first step in adding support for Fleet Automation Installer to allow remote management of DDOT installations on linux bare-metal. 

I manually tested the installation process on a linux vm and added an e2e test. 

### Additional Notes

- This PR is dependent on @liustanley [PR](https://github.com/DataDog/datadog-agent/pull/39215) that system files attempting to start before installation is complete
- [JIRA ticket](https://datadoghq.atlassian.net/browse/OTAGENT-455?atlOrigin=eyJpIjoiNjJjYWY0ZmNiNTA0NDM0YjkwZjg0NTMxZTdjNjY0ZDUiLCJwIjoiaiJ9)
-->


test/new-e2e/tests/installer/unix/package_agent_test.go
```diff
@@ -461,3 +463,73 @@ func (s *packageAgentSuite) installDebRPMAgent() {
...
+		DD_OTELCOLLECTOR_ENABLED=true
+		`, apiKey)
+
+	s.host.Run(fmt.Sprintf(`printf "%s" | sudo tee /etc/datadog-agent/environment > /dev/null`, envContent))
+
+	// Install ddot
+	pipelineID := os.Getenv("E2E_PIPELINE_ID")
+	ddotOCIURL := fmt.Sprintf("oci://installtesting.datad0g.com/ddot-package:pipeline-%s", pipelineID)
+
+	s.host.Run(fmt.Sprintf("sudo datadog-installer install %q", ddotOCIURL))
```
BaptisteFoy: **🥜 Nitpick:** ‏Could be regrouped in a single line without losing clarity

test/new-e2e/tests/installer/unix/package_agent_test.go
```diff
@@ -461,3 +463,73 @@ func (s *packageAgentSuite) installDebRPMAgent() {
...
+	if apiKey == "" {
+		apiKey = "deadbeefdeadbeefdeadbeefdeadbeef"
+	}
+
+	envContent := fmt.Sprintf(`DD_API_KEY=%s
+		DD_SITE=datadoghq.com
+		DD_OTELCOLLECTOR_ENABLED=true
+		`, apiKey)
+
+	s.host.Run(fmt.Sprintf(`printf "%s" | sudo tee /etc/datadog-agent/environment > /dev/null`, envContent))
```
BaptisteFoy: **❓ Question:** ‏I'm curious why we need to put environment variables here, and to see if we can avoid it

test/new-e2e/tests/installer/unix/package_agent_test.go
```diff
@@ -461,3 +463,73 @@ func (s *packageAgentSuite) installDebRPMAgent() {
...
+	pipelineID := os.Getenv("E2E_PIPELINE_ID")
+	ddotOCIURL := fmt.Sprintf("oci://installtesting.datad0g.com/ddot-package:pipeline-%s", pipelineID)
+
+	s.host.Run(fmt.Sprintf("sudo datadog-installer install %q", ddotOCIURL))
+	s.host.AssertPackageInstalledByInstaller("datadog-agent-ddot")
+
+	s.host.Run("sudo cp /etc/datadog-agent/otel-config.yaml.example /etc/datadog-agent/otel-config.yaml")
+
+	// Start ddot service
+	s.host.Run("sudo systemctl start datadog-agent-ddot.service")
```
BaptisteFoy: **💬 Suggestion:** Both of these should be in the postinst‏

test/new-e2e/tests/installer/unix/package_agent_test.go
tim-liakh-datadog: I mainly went with the env variable approach because some e2e installer tests were using it ([example](https://github.com/DataDog/datadog-agent/blob/d90e09479b8b8f0e79525a493ae773d4575d99df/test/new-e2e/tests/installer/unix/package_definitions.go#L130-L132))
To work without them we would need to substitute in the API and site into otel-config.yaml and set otel collector enabled in datadog.yaml. This is always an option if you think it will be better.

test/new-e2e/tests/installer/unix/package_agent_test.go
tim-liakh-datadog: @liustanley is working on this in his [PR](https://github.com/DataDog/datadog-agent/pull/39215). Will pull changes down and fix test as needed.

test/new-e2e/tests/installer/unix/package_agent_test.go
```diff
@@ -461,3 +463,71 @@ func (s *packageAgentSuite) installDebRPMAgent() {
...
+
+	s.host.WaitForUnitActive(s.T(), ddotUnit)
+
+	// Verify running
+	state := s.host.State()
+	s.assertUnits(state, false)
+	s.assertDdotUnits(state, false)
+
+	// Verify files exist
+	// state.AssertFileExists("/etc/datadog-agent/otel-config.yaml", 0644, "dd-agent", "dd-agent")
```
tim-liakh-datadog: This line will be uncommented once Stanley's PR is merged. Currently failing due to otel-config being assigned to root. Once the file creation process is moved to post-install script, the permissions will be fixed.

pkg/fleet/installer/installer.go
```diff
@@ -305,7 +305,7 @@ func (i *installerImpl) doInstall(ctx context.Context, url string, args []string
 	if err != nil {
 		return fmt.Errorf("could not remove package installation in db: %w", err)
 	}
-	configDir := filepath.Join(i.userConfigsDir, pkg.Name)
+	configDir := filepath.Join(i.userConfigsDir, "datadog-agent")
```
tim-liakh-datadog, line 310: This is something I wanted to double check. As far as @BaptisteFoy and I know no other installation should happen outside of datadog-agent. This change was necessary because ddot config should live inside /etc/datadog-agent. Without this, it would get created in /etc/datadog-agent-ddot

clarkb7: for the future: we'll need to change this code for Windows as well. the config dir doesn't match the package name.

git-thuerk-done (approved):

ofek (approved):

spencergilbert (approved):
Approved ADEL owned files

pkg/fleet/installer/installer.go
BaptisteFoy: Good point @clarkb7, we can maybe update `userConfigsDir` directly? IIRC it's set differently depending on the OS

@tim-liakh-datadog consider this not relevant for your PR, we'll open a new one with these changes when we need to

releasenotes/notes/create-ddot-oci-678f8e0c29ccd48c.yaml
```diff
@@ -0,0 +1,12 @@
...
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+features:
+  - |
+    Added support for building, packaging, and installing ddot as an OCI package. The project can now
+    be installed using the image.
```
BaptisteFoy: Not sure about using "DDOT" -- is this customer-facing?

```suggestion
    Added support for building, packaging, and installing the Datadog Distribution of Open Telemetry Collector (DDOT) as an OCI package. The project can now be installed using the Datadog Installer.
```

test/new-e2e/tests/installer/unix/package_agent_test.go
BaptisteFoy: I think it's better to use the config file; my worry with using env vars is that they have to be declared in systemd units to actually be used

.gitlab/deploy_packages/oci.yml
```diff
@@ -46,3 +46,10 @@ deploy_installer_oci:
   variables:
     OCI_PRODUCT: "datadog-installer"
     S3_PACKAGE: "installer-package"
+
+deploy_ddot_oci:
+  extends: ".deploy_packages_oci"
+  needs: [ "ddot_oci", "go_tools_deps" ]
+  variables:
+    OCI_PRODUCT: "datadog-agent-ddot"
```
clarkb7, line 54: since `ddot` stands for `datadog opentelemetry` (i think?) this is pretty much `datadog-agent-datadog-opentelemetry` 
should we choose a name like `datadog-agent-otel` or `datadog-otel-collector` etc. ?

pkg/fleet/installer/packages/datadog_agent_ddot_linux.go
```diff
@@ -0,0 +1,73 @@
...
+
+var (
+	ddotConfigPermissions = file.Permissions{
+		{Path: "otel-config.yaml.example", Owner: "dd-agent", Group: "dd-agent", Mode: 0644},
+	}
+
+	// ddotPackagePermissions are the ownerships and modes that are enforced on the DDOT package files
+	ddotPackagePermissions = file.Permissions{
+		{Path: ".", Owner: "dd-agent", Group: "dd-agent", Recursive: true},
+	}
```
clarkb7: what's the default mode? should we be explicit?

pkg/fleet/installer/packages/datadog_agent_ddot_linux.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+var datadogAgentDdotPackage = hooks{
+	postInstall: postInstallDatadogAgentDdot,
+	preRemove:   preRemoveDatadogAgentDdot,
+}
+
+var (
+	ddotConfigPermissions = file.Permissions{
+		{Path: "otel-config.yaml.example", Owner: "dd-agent", Group: "dd-agent", Mode: 0644},
```
clarkb7, line 35: who is expected to copy this `.example` to `otel-config.yaml` ? the user or this package?

pkg/fleet/installer/packages/datadog_agent_ddot_linux.go
```diff
@@ -0,0 +1,73 @@
...
+
+// preRemoveDatadogAgentDdot performs pre-removal steps for the DDOT package
+// All the steps are allowed to fail
+func preRemoveDatadogAgentDdot(ctx HookContext) error {
+	if !ctx.Upgrade {
+		// Only remove config files during actual uninstall, not during upgrades
+		err := ddotConfigUninstallPaths.EnsureAbsent("/etc/datadog-agent")
+		if err != nil {
+			log.Warnf("failed to remove DDOT config files: %s", err)
+		}
```
clarkb7: the Agent package does this already, so a full uninstall has this covered. I assume we don't want uninstalling only ddot to delete the Agent's config, too? we should try to specifically delete ddot's files, or just leave it up tot he Agent package.

clarkb7 (commented):
I don't think you need explicit approval from the Windows team, since this is just a Linux PR.

releasenotes/notes/create-ddot-oci-678f8e0c29ccd48c.yaml
tim-liakh-datadog: I updated it to include the fall name for clarity.

pkg/fleet/installer/packages/datadog_agent_ddot_linux.go
tim-liakh-datadog: The package will copy it. Stanley's PR does this currently, just waiting for it to get merged to add this functionality here.

pkg/fleet/installer/packages/datadog_agent_ddot_linux.go
tim-liakh-datadog: `ddotConfigUninstallPaths = file.Paths{
		"otel-config.yaml.example",
	}`
	
This code can be a little confusing. `ddotConfigUninstallPaths` holds the name of the files we want uninstalled and `"/etc/datadog-agent"` the directory where we should be looking. So only otel-config.yaml.example should get uninstalled, leaving the other files in that folder

pkg/fleet/installer/packages/datadog_agent_ddot_linux.go
tim-liakh-datadog: I just checked, the structure inside has different modes. 0755 for directories & executable files, 0644 for regular files, 0777 for symbolic links. The core agent doesn't specify the default mode either, so I think it should be fine

```
agentConfigPermissions = file.Permissions{
		{Path: ".", Owner: "dd-agent", Group: "dd-agent", Recursive: true},
```

releasenotes/notes/create-ddot-oci-678f8e0c29ccd48c.yaml
liustanley: I think there's still more work that needs to be done to install DDOT alongside the agent, and we're not ready to announce support for OCI/bare-metal yet. @BaptisteFoy do you think we should do a release note in a future PR instead of this one?

.gitlab/deploy_packages/oci.yml
tim-liakh-datadog: I named it `datadog-agent-ddot` because existing ddot workflows such as `ddot.rb` reference it as such. I agree it is confusing, so it might be worth looking into changing what it is called. I will look into it.

omnibus/package-scripts/ddot-deb/preinst
```diff
@@ -0,0 +1,30 @@
...
+    # Stop an already running agent
+    # supports systemd, upstart and sysvinit
+    if command -v systemctl >/dev/null 2>&1; then
+        systemctl stop $SERVICE_NAME || true
+    elif command -v initctl >/dev/null 2>&1; then
+        initctl stop $SERVICE_NAME || true
+    elif command -v service >/dev/null 2>&1; then
+        service $SERVICE_NAME stop || true
+    else
+        echo "[ WARNING ]\tCannot detect a supported init system. The datadog-agent-ddot package only provides service files for systemd, upstart and sysvinit."
```
rdesgroppes: Without `-e`, a literal `\t` would be printed. Better use `printf` anyway, and write unexpected cases to the standard error.
```suggestion
        printf "[ WARNING ]\tCannot detect a supported init system. The datadog-agent-ddot package only provides service files for systemd, upstart and sysvinit." >&2
```

omnibus/package-scripts/ddot-deb/preinst
```diff
@@ -0,0 +1,30 @@
...
+    fi
+fi
+
+# Uninstall the agent if it was installed by the installer
+if command -v datadog-installer >/dev/null 2>&1 && datadog-installer is-installed datadog-agent-ddot; then
+    datadog-installer remove datadog-agent-ddot || printf "[ WARNING ]\tFailed to remove datadog-agent-ddot installed by the installer\n"
```
rdesgroppes:
```suggestion
    datadog-installer remove datadog-agent-ddot || printf "[ WARNING ]\tFailed to remove datadog-agent-ddot installed by the installer\n" >&2
```
etc. (several occurrences)

omnibus/package-scripts/ddot-deb/postinst
```diff
@@ -0,0 +1,19 @@
...
+# If we are inside the Docker container, do nothing
+if [ -n "$DOCKER_DD_AGENT" ]; then
+    echo "Installation from docker-dd-agent, nothing to do in postinst"
+    exit 0
+fi
+
+# Run the postinst. See pkg/fleet/installer/packages/datadog_agent_ddot_linux.go
+datadog-installer postinst datadog-agent-ddot deb || true
+
+exit 0
```
rdesgroppes: minor: redundant (many occurrences). Or is it by convention?

rdesgroppes (approved):
Only minor comments

CelianR (approved):

releasenotes/notes/create-ddot-oci-678f8e0c29ccd48c.yaml
BaptisteFoy: Yeah that makes sense

.gitlab/deploy_packages/oci.yml
liustanley: We ran `datadog-agent-ddot` and `datadog-agent-otel` by product as options for the package name and decided on `datadog-agent-ddot`. I think the consensus was to use ddot as the public facing name whenever possible.

omnibus/package-scripts/ddot-deb/postinst
liustanley: This is here because we mirrored the files in `agent-deb` and `agent-rpm`

omnibus/package-scripts/ddot-deb/preinst
liustanley: Sounds good, I can make these changes in my PR. These lines are also mirrored from the files in `agent-deb` and `agent-rpm` in case you want to edit those as well

pkg/fleet/installer/packages/embedded/templates/main.go
```diff
@@ -114,27 +115,47 @@ var (
...
+	ddotExpDataOCI = systemdTemplateData{
+		InstallDir:       "/opt/datadog-packages/datadog-agent-ddot/experiment",
+		EtcDir:           "/etc/datadog-agent",
+		FleetPoliciesDir: "/etc/datadog-agent/managed/datadog-agent/experiment",
+		PIDDir:           "/opt/datadog-packages/datadog-agent/experiment",
+		Stable:           false,
+	}
+
+	systemdUnitsOCI    = systemdUnits(stableDataOCI, expDataOCI, ddotStableDataOCI, ddotExpDataOCI, true)
+	systemdUnitsDebRpm = systemdUnits(stableDataDebRpm, expDataDebRpm, stableDataDebRpm, expDataDebRpm, false)
```
liustanley, line 160: I think there's an extra `stableDataDebRpm` here?

pkg/fleet/installer/packages/embedded/templates/main.go
```diff
@@ -114,27 +115,47 @@ var (
...
 		Stable:           false,
 	}
-	systemdUnitsOCI    = systemdUnits(stableDataOCI, expDataOCI, true)
-	systemdUnitsDebRpm = systemdUnits(stableDataDebRpm, expDataDebRpm, false)
+
+	ddotStableDataOCI = systemdTemplateData{
+		InstallDir:       "/opt/datadog-packages/datadog-agent-ddot/stable",
+		EtcDir:           "/etc/datadog-agent",
+		FleetPoliciesDir: "/etc/datadog-agent/managed/datadog-agent/stable",
+		PIDDir:           "/opt/datadog-packages/datadog-agent/stable",
```
liustanley, line 148: Just curious, is there a reason why we need to use `datadog-agent` instead of `datadog-agent-ddot` for the PID file?

pkg/fleet/installer/packages/datadog_agent_ddot_linux.go
```diff
@@ -0,0 +1,240 @@
...
+		return fmt.Errorf("failed to set DDOT package ownerships: %v", err)
+	}
+
+	// Set DDOT config permissions
+	if err = ddotConfigPermissions.Ensure("/etc/datadog-agent"); err != nil {
+		return fmt.Errorf("failed to set DDOT config ownerships: %v", err)
+	}
+
+	// Enable otelcollector in datadog.yaml
+	if err = enableOtelCollectorConfig(); err != nil {
```
liustanley, line 107: This should add `otelcollector.enabled` for the DDOT core config, but I'm not sure if the core agent also needs to be restarted if it needs this config option to be set? cc @truthbk

pkg/fleet/installer/packages/datadog_agent_ddot_linux.go
```diff
@@ -0,0 +1,240 @@
...
+	ddotPackagePermissions = file.Permissions{
+		{Path: ".", Owner: "dd-agent", Group: "dd-agent", Recursive: true},
+	}
+
+	// ddotConfigUninstallPaths are the files that are deleted during an uninstall
+	ddotConfigUninstallPaths = file.Paths{
+		"otel-config.yaml.example",
+	}
+
+	datadogYamlPath = "/etc/datadog-agent/datadog.yaml"
```
liustanley: I think it makes more sense for this to be a `const` since we're not modifying it

pkg/fleet/installer/packages/datadog_agent_ddot_linux.go
```diff
@@ -0,0 +1,226 @@
...
+func enableOtelCollectorConfig() error {
+	// Read existing config
+	var existingConfig map[string]interface{}
+	if data, err := os.ReadFile(datadogYamlPath); err == nil {
+		if err := yaml.Unmarshal(data, &existingConfig); err != nil {
+			return fmt.Errorf("failed to parse existing datadog.yaml: %w", err)
+		}
+	} else if !os.IsNotExist(err) {
+		return fmt.Errorf("failed to read datadog.yaml: %w", err)
+	}
```
liustanley: `datadog.yaml` is required to run DDOT and is generally expected here, so I don't think we need to discern the `os.IsNotExist` error. Also, it's best to avoid nested if statements and `else if` when possible in Go.
```suggestion
        data, err := os.ReadFile(datadogYamlPath)
	if err != nil {
		return fmt.Errorf("failed to read datadog.yaml: %w", err)
	}
	if err := yaml.Unmarshal(data, &existingConfig); err != nil {
		return fmt.Errorf("failed to parse existing datadog.yaml: %w", err)
	}
```

pkg/fleet/installer/packages/datadog_agent_ddot_linux.go
```diff
@@ -0,0 +1,226 @@
...
+
+	// Config is empty
+	if existingConfig == nil {
+		existingConfig = make(map[string]interface{})
+	}
+
+	existingConfig["otelcollector"] = map[string]interface{}{"enabled": true}
```
liustanley, line 187: @truthbk do you think we need to make sure `existingConfig["otelcollector"]` isn't already set before adding `enabled: true`? If a user has their own `otelcollector:` config section defined before installing DDOT, this would override it.

pkg/fleet/installer/packages/datadog_agent_ddot_linux.go
```diff
@@ -0,0 +1,226 @@
...
+func disableOtelCollectorConfig() error {
+	// Read existing config
+	data, err := os.ReadFile(datadogYamlPath)
+	if err != nil {
+		// Nothing to delete if the file doesn't exist
+		if os.IsNotExist(err) {
+			return nil
+		}
+		return fmt.Errorf("failed to read datadog.yaml: %w", err)
+	}
```
liustanley:
```suggestion
	// Nothing to delete if the file doesn't exist
	if err != nil && os.IsNotExist(err) {
		return nil
	}
	if err != nil {
		return fmt.Errorf("failed to read datadog.yaml: %w", err)
	}
```

pkg/fleet/installer/packages/embedded/templates/main.go
tim-liakh-datadog: When I was looking at the structure all of the PID files were being stored inside of datadog-agent. I wanted to keep the structure

pkg/fleet/installer/packages/embedded/templates/main.go
tim-liakh-datadog: The stableData for Deb and Rpm is the same for both regular and ddot. However, for OCI the ddot and regular version slightly differ. I agree it is confusing, I just didn't want to duplicate code.

liustanley: Oh that makes sense, I misread the function params here so this looks good

fanny-jiang (approved):

pkg/fleet/installer/packages/datadog_agent_ddot_linux.go
```diff
@@ -25,10 +28,15 @@ const (
 	agentDDOTPackage = "datadog-agent-ddot"
 )
 
+const (
+	datadogYamlPath = "/etc/datadog-agent/datadog.yaml"
```
liustanley: nit: Could you group this constant in the same block as `agentDDOTPackage`?

pkg/fleet/installer/packages/datadog_agent_ddot_linux.go
```diff
@@ -140,6 +154,87 @@ func preRemoveDatadogAgentDdot(ctx HookContext) error {
...
+
+	return nil
+}
+
+// enableOtelCollectorConfig adds otelcollector.enabled: true to datadog.yaml
+func enableOtelCollectorConfig() error {
+	// Read existing config
+	var existingConfig map[string]interface{}
+	data, err := os.ReadFile(datadogYamlPath)
+	if err != nil && !os.IsNotExist(err) {
```
liustanley: Is there a reason why `!os.IsNotExist` is necessary here? I was under the impression that we should always have a `datadog.yaml` present when installing.

test/new-e2e/tests/agent-platform/ddot/ddot_install_test.go
```diff
@@ -142,6 +142,10 @@ func (is *ddotInstallSuite) ConfigureAndRunAgentService(VMclient *common.TestCli
 		} else {
 			ExecuteWithoutError(t, VMclient, "sudo systemctl restart datadog-agent.service")
 		}
+
+		// Reset systemd failure state and start ddot service
+		ExecuteWithoutError(t, VMclient, "sudo systemctl reset-failed datadog-agent-ddot.service")
+		ExecuteWithoutError(t, VMclient, "sudo systemctl start datadog-agent-ddot.service")
```
liustanley, line 149: This test should be passing without these two lines, the expected behavior is that DDOT will start successfully when `datadog-agent.service` is restarted. What errors were you observing before adding these lines?

Also, you should remove this line since you added support for it in the postinst:
```
ExecuteWithoutError(t, VMclient, "sudo sh -c \"printf 'otelcollector:\\n enabled: true\\n' >> /etc/datadog-agent/datadog.yaml\"")
```

test/new-e2e/tests/installer/unix/package_agent_test.go
```diff
@@ -461,3 +463,69 @@ func (s *packageAgentSuite) installDebRPMAgent() {
...
+	// Substitute API & site into otel-config.yaml
+	apiKey := os.Getenv("DD_API_KEY")
+	if apiKey == "" {
+		apiKey = "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
+	}
+	site := os.Getenv("DD_SITE")
+	if site == "" {
+		site = "datadoghq.com"
+	}
+	s.host.Run(fmt.Sprintf("sudo sh -c \"sed -i -e 's/\\${env:DD_API_KEY}/%s/' -e 's/\\${env:DD_SITE}/%s/' /etc/datadog-agent/otel-config.yaml\"", apiKey, site))
```
liustanley: Does `os.Getenv("DD_API_KEY/DD_SITE")` ever need to be used? If not, I think we can replace this block with the command used in `ddot_install_test`:
```
ExecuteWithoutError(t, VMclient, "sudo sh -c \"sed -i -e 's/\\${env:DD_API_KEY}/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/' -e 's/\\${env:DD_SITE}/datadoghq.com/' /etc/datadog-agent/otel-config.yaml\"")
```

I think we were trying to use env variables at first but after reading Baptiste's [comment](https://github.com/DataDog/datadog-agent/pull/38805#discussion_r2230857894) it seems like we don't need to.

test/new-e2e/tests/installer/unix/package_agent_test.go
```diff
@@ -461,3 +463,69 @@ func (s *packageAgentSuite) installDebRPMAgent() {
...
+	defer s.Purge()
+	s.host.AssertPackageInstalledByInstaller("datadog-agent")
+	s.host.WaitForUnitActive(s.T(), agentUnit, traceUnit)
+
+	// Install ddot
+	s.host.Run(fmt.Sprintf("sudo datadog-installer install oci://installtesting.datad0g.com/ddot-package:pipeline-%s", os.Getenv("E2E_PIPELINE_ID")))
+	s.host.AssertPackageInstalledByInstaller("datadog-agent-ddot")
+
+	// Check if datadog.yaml exists, if not return an error
+	s.host.Run("sudo test -f /etc/datadog-agent/datadog.yaml || { echo 'Error: datadog.yaml does not exist'; exit 1; }")
```
liustanley, line 491: Should we be using `AssertFileExists` here instead?

test/new-e2e/tests/installer/unix/package_agent_test.go
```diff
@@ -461,3 +463,69 @@ func (s *packageAgentSuite) installDebRPMAgent() {
...
+	state.AssertFileExists("/etc/datadog-agent/otel-config.yaml", 0644, "dd-agent", "dd-agent")
+
+	state.AssertDirExists("/opt/datadog-packages/datadog-agent-ddot/stable", 0755, "dd-agent", "dd-agent")
+	state.AssertFileExists("/opt/datadog-packages/datadog-agent-ddot/stable/embedded/bin/otel-agent", 0755, "dd-agent", "dd-agent")
+
+	s.host.Run("sudo grep -q 'otelcollector:' /etc/datadog-agent/datadog.yaml")
+}
+
+// Verify ddot service running
+func (s *packageAgentSuite) assertDdotUnits(state host.State, oldUnits bool) {
```
liustanley: nit: Would prefer this function to be `assertDDOTUnits` (all same casing for DDOT)

pkg/fleet/installer/packages/datadog_agent_ddot_linux.go
tim-liakh-datadog: My thought process was just in case datadog-agent is pre-installed and someone removes datadog.yaml on accident. Instead of failing the entire process, I would just create a new datadog.yaml. I agree in theory it should never happen, so I am fine removing it if you think it is best.

liustanley: Yes I think removing it and simplifying the block below would be best:
```
	if err := yaml.Unmarshal(data, &existingConfig); err != nil {
		return fmt.Errorf("failed to parse existing datadog.yaml: %w", err)
	}
```

The datadog-agent package also fails the process if the config file is not present: https://github.com/DataDog/datadog-agent/blob/46ef69f5e294c47a50bbf6d4d16394afd3af5765/pkg/fleet/installer/packages/datadog_agent_linux.go#L460

test/new-e2e/tests/agent-platform/ddot/ddot_install_test.go
tim-liakh-datadog: The issue in this case is datadog.yaml doesn't have its API key substituted until after we try to start the ddot service, causing the initial attempt at starting it to fail. I added the start command to have it try again and the reset-failed to bypass the "attempt to start made too quickly"

pkg/fleet/installer/packages/datadog_agent_ddot_linux.go
BaptisteFoy: If we change the agent config and expect the agent to pick it up it must be restarted

BaptisteFoy (approved):
lgtm, probably just requires a core agent restart

test/new-e2e/tests/agent-platform/ddot/ddot_install_test.go
liustanley: Discussed offline to leave these two lines here. The reason why the DDOT service fails at first in this test is because `api_key` doesn't get set when installing with apt/yum directly, and `enableOtelCollectorConfig` doesn't preserve `api_key:` in the datadog.yaml. This should work as expected in the install script since `api_key` gets set there.

Also you have to remove the otelcollector enabled line again in this function due to the revert.

test/new-e2e/tests/agent-platform/ddot/ddot_install_test.go
```diff
@@ -142,6 +142,10 @@ func (is *ddotInstallSuite) ConfigureAndRunAgentService(VMclient *common.TestCli
 		} else {
 			ExecuteWithoutError(t, VMclient, "sudo systemctl restart datadog-agent.service")
 		}
+
```
liustanley, line 146:
```
ExecuteWithoutError(t, VMclient, "sudo sh -c \"sed 's/api_key:.*/api_key: XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/' /etc/datadog-agent/datadog.yaml.example > /etc/datadog-agent/datadog.yaml\"")
```
This line will actually override the existing `datadog.yaml` (with the otel collector enabled changes) with the config template. Since we know that `datadog.yaml` only contains `otelcollector.enabled` in this test, we can change it to:
```
ExecuteWithoutError(t, VMclient, "sudo sh -c \"printf 'api_key: XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX' >> /etc/datadog-agent/datadog.yaml\"")
```

liustanley (approved):
Two small test comments, other than that LGTM

pkg/fleet/installer/packages/datadog_agent_ddot_linux.go
liustanley: I think the agent-platform-ddot tests might be failing now because `api_key:` is removed in this function and the agent service is failing to restart on this new config

PR #38807, opened 2025-07-15
[EBPF] kmt: update error message when a stack does not exist

gjulianm:
### What does this PR do?

This PR changes the error message generated when a user selects a KMT stack that does not exist, and also changes the code to emit the same message from a single function.

### Motivation

Fit the workflow changed in https://github.com/DataDog/datadog-agent/pull/38178/files

Manually tested, linters passing.


tasks/kmt.py
```diff
@@ -136,7 +136,7 @@ def gen_config(
     init_stack=True,
     vcpu: str | None = None,
     memory: str | None = None,
-    new=False,
+    new=True,
```
mbertrone, line 139: nit: just to make sure this is intentional. I think the new default makes the UX much simpler for the user.

mbertrone (approved):
LGTM!

tasks/kmt.py
gjulianm: Yes, it's intentional. In the future I think it would be even better to merge `init_stack` and `new` but I think for that we need to give more thought to the UX of creating stacks.

PR #38812, opened 2025-07-15
[CONTP-791] Add cluster check runner count to cluster agent metadata

zhuminyi:
### What does this PR do?
Report cluster check runner count in cluster-agent metadata payload

### Motivation

Cluster Agent Metadata: Metadata reporting has been enabled in the Cluster Agent since version 7.67. This payload enhances visibility into cluster information and aids in debugging.

Lack of Awareness in Cluster Agent: Historically, the Cluster Agent has had limited visibility into cluster check providers. As a result, both the node agent and the cluster check runner could end up executing cluster checks, depending on the user’s configuration.

Purpose of This PR: This pull request introduces reporting of the cluster check runner count. The goal is to enable better optimization of cluster check usage. For instance, the Kube State Metrics (KSM) check should not be run solely on a single node agent. If the cluster check runner count is 0 while feature_cluster_checks_enabled is true, it indicates a misconfiguration, and a cluster check runner should be enabled.

Data Visibility: The reported information can be monitored in DDSQL, pending a separate enabling PR.

Implementation:
clc runner post status with node type -> cluster agent -> node store -> metadata 

Go to leader cluster agent pod, and run
`root@datadog-agent-linux-cluster-agent-5dc58987bc-4brlv:/# agent flare
`
Download flare and find cluster-agent-metadata.json
The metadata should be like
```
    "datadog_cluster_agent_metadata": {
        "cluster_check_runner_count": 3, # number of clc runners
        .....
```

Turn off clc runner and get flare again, the count should be zero now

### DIFF
--- a/comp/core/autodiscovery/providers/clusterchecks.go
+++ b/comp/core/autodiscovery/providers/clusterchecks.go
@@ -118,6 +118,7 @@ func (c *ClusterChecksConfigProvider) IsUpToDate(ctx context.Context) (bool, err
 
 	status := types.NodeStatus{
 		LastChange: c.lastChange,
+		NodeType:   "clc-runner",
 	}
 
 	reply, err := c.dcaClient.PostClusterCheckStatus(ctx, c.identifier, status)
@@ -216,6 +217,7 @@ func (c *ClusterChecksConfigProvider) postHeartbeat(ctx context.Context) error {
 
 	status := types.NodeStatus{
 		LastChange: types.ExtraHeartbeatLastChangeValue,
+		NodeType:   "clc-runner",
 	}
 
 	_, err := c.dcaClient.PostClusterCheckStatus(ctx, c.identifier, status)
--- a/comp/core/autodiscovery/providers/clusterchecks.go
+++ b/comp/core/autodiscovery/providers/clusterchecks.go
@@ -54,7 +54,7 @@ func NewClusterChecksConfigProvider(providerConfig *pkgconfigsetup.Configuration
 		graceDuration:    defaultGraceDuration,
 		degradedDuration: defaultDegradedDeadline,
 		heartbeat:        atomic.NewTime(time.Now()),
-		nodeType:         types.NodeTypeUnknown,
+		nodeType:         types.NodeTypeCLCRunner,
 	}
 
 	c.identifier = pkgconfigsetup.Datadog().GetString("clc_runner_id")
@@ -71,7 +71,7 @@ func NewClusterChecksConfigProvider(providerConfig *pkgconfigsetup.Configuration
 	}
 
 	// Set nodeType based on config
-	if pkgconfigsetup.Datadog().GetBool("clc_runner_enabled") {
+	if pkgconfigsetup.IsCLCRunner(pkgconfigsetup.Datadog()) {
 		c.nodeType = types.NodeTypeCLCRunner
 	} else {
 		c.nodeType = types.NodeTypeNodeAgent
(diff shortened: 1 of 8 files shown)


comp/core/autodiscovery/providers/clusterchecks.go
```diff
@@ -68,6 +70,13 @@ func NewClusterChecksConfigProvider(providerConfig *pkgconfigsetup.Configuration
 		}
 	}
 
+	// Set nodeType based on config
+	if pkgconfigsetup.Datadog().GetBool("clc_runner_enabled") {
```
davidor: There's a helper for this:
https://github.com/DataDog/datadog-agent/blob/bc9b4d65a237a123d41de58b4d49e838f5472f9c/pkg/config/setup/config.go#L2633

pkg/clusteragent/clusterchecks/types/types.go
```diff
@@ -17,9 +17,25 @@ const (
...
 
+/*
+NodeType represents the type of node reporting status to the cluster agent.
+It is used to distinguish between CLC runners, node agents, and unknown types.
+*/
+type NodeType uint8
+
+const (
+	// NodeTypeUnknown is used when the node type is not set or cannot be determined.
+	NodeTypeUnknown NodeType = 0
```
davidor: I think this is never used in the code, because by looking at the function I suggested above, we should always be able to tell whether it’s a CLC or a node agent.

zhuminyi: removed

pkg/clusteragent/clusterchecks/stats.go
```diff
@@ -33,6 +33,31 @@ func GetStats() (*types.Stats, error) {
...
 
+// GetCLCRunnerCount returns the number of nodes with NodeType == "clc-runner"
+func GetCLCRunnerCount() (int, error) {
+	key := cache.BuildAgentKey(handlerCacheKey)
+	x, found := cache.Cache.Get(key)
+	if !found {
+		return 0, errors.New("Clusterchecks not running")
+	}
+
+	handler, ok := x.(*Handler)
```
davidor, line 44: I find this pattern of storing the handler in a cache a bit strange, but I see that it’s used in other functions in the file, so we can leave it like this and address it later.

pkg/clusteragent/clusterchecks/stats.go
```diff
@@ -33,6 +33,31 @@ func GetStats() (*types.Stats, error) {
...
+
+	handler.dispatcher.store.RLock()
+	defer handler.dispatcher.store.RUnlock()
+
+	count := 0
+	for _, node := range handler.dispatcher.store.nodes {
+		if node != nil && node.nodetype == types.NodeTypeCLCRunner {
+			count++
+		}
+	}
```
davidor: I think we should move this to the `nodeStore` struct in `pkg/clusteragent/clusterchecks/stores.go`. We can create a function called `CountCLCRunners` or something similar there, to avoid having this logic that takes the internal mutex and everything here.

comp/metadata/clusteragent/impl/cluster_agent.go
```diff
@@ -211,6 +212,13 @@ func (dca *datadogclusteragent) getMetadata() map[string]interface{} {
 			dca.metadata["is_leader"] = leaderEngine.IsLeader()
 		}
 	}
+
+	// Add cluster check runner count
+	// Import "github.com/DataDog/datadog-agent/pkg/clusteragent/clusterchecks"
+	if count, err := clusterchecks.GetCLCRunnerCount(); err == nil {
+		dca.metadata["cluster_check_runner_count"] = count
```
davidor: Is this enough, or do we also need the count of node agents running cluster checks?
I’m thinking that if this count is 0, we won’t be able to distinguish between a setup that’s running the cluster checks on the DCA and one that’s running them on the node agents.

davidor (approved):

PR #38817, opened 2025-07-15
[CXP-2186] [agent] process check refactor

gengnamstyle:
### What does this PR do?
- Removes main `procfs` call of `ProcessesByPid` in process check run when `process_config.process_collection.use_wlm` is enabled
- Changes use of `time.Now()` to use an interface and concrete implementation for mocking purposes

### What does this PR NOT do?
- refactor container collection (work for a followup pr)
    - if added, I believe the PR would be too big and allow more room for bugs, this additional work is tracked here https://datadoghq.atlassian.net/jira/software/c/projects/CXP/boards/5501?selectedIssue=CXP-2436

### Motivation
as part of the processes + service discovery consolidation, we need to refactor the process check to use the new WLM process collector. RFC  - https://datadoghq.atlassian.net/wiki/spaces/cxg/pages/5013111768/RFC+Processes+and+Service+Discovery+Agent+Consolidation

- adds unit test for `run` with wlm process collection enabled
- built agent locally and enabled the new process collection (`process_config.process_collection.use_wlm`), note state is unknown because it exited 
image

- matches old agent with new process collection disabled:
image

### Additional Notes

- there is still work to do to update the `GetContainers` call to use `GetContainerForProcess` function in WLM which was introduced in the last [PR](https://datadoghq.atlassian.net/jira/software/c/projects/CXP/boards/5501?selectedIssue=CXP-2391) 
    - this was not done in this PR as there would be too many changes


pkg/process/checks/process.go
```diff
@@ -243,11 +256,36 @@ func (p *ProcessCheck) runWLM(groupID int32, collectRealTime bool) (RunResult, e
...
+			Name:    wlmProc.Name,
+			Cwd:     wlmProc.Cwd,
+			Exe:     wlmProc.Exe,
+			Comm:    wlmProc.Comm,
+			Cmdline: wlmProc.Cmdline,
+			Uids:    wlmProc.Uids,
+			Gids:    wlmProc.Gids,
+			Stats:   statsForProcess[wlmProc.Pid],
+		}
+	}
```
daniel-taf: Let's put this in a helper function or within some sort WLM flavor of the probe to keep the function from getting too large. Should help with mocking as well, like we talked about offline.

pkg/process/checks/process.go
```diff
@@ -271,6 +309,7 @@ func (p *ProcessCheck) runWLM(groupID int32, collectRealTime bool) (RunResult, e
 		cacheValidity = cacheValidityRT
 	}
 
+	// TODO: this will be replaced by the GetContainerForProcess function in WLM
```
daniel-taf: Can you create a card to track this work and link it here?

pkg/process/procutil/process_model.go
```diff
@@ -77,9 +77,14 @@ func (p *Process) DeepCopy() *Process {
...
-	// R: Running S: Sleep T: Stop I: Idle
-	// Z: Zombie W: Wait L: Lock
+	// Supported return values:
+	// U: unknown state
+	// D: uninterruptible sleep
+	// R: running
+	// S: interruptible sleep
+	// T: stopped
+	// W: waiting (todo: potentially removable, W = paging only before Linux 2.6.0, waking Linux 2.6.33 to 3.13 only)
+	// Z: zombie
```
daniel-taf, line 87: What lead to this change?

pkg/process/checks/process.go
```diff
@@ -45,6 +45,16 @@ const (
...
 
+type processClock interface {
+	Now() time.Time
+}
+
+type realClock struct{}
+
+func (r realClock) Now() time.Time {
+	return time.Now()
+}
```
daniel-taf: Maybe we should switch to using an actual clock interface like `github.com/benbjohnson/clock` instead of having this indirection.

pkg/process/procutil/process_model.go
gengnamstyle: the previous comments were incorrect and outdated. There is no I (idle) nor L (lock) state and it was missing the values defined in the agent payload status

pkg/process/checks/process.go
gengnamstyle: https://datadoghq.atlassian.net/jira/software/c/projects/CXP/boards/5501?selectedIssue=CXP-2436

pkg/process/checks/process.go
gengnamstyle: updated to that interface!

pkg/process/checks/process.go
gengnamstyle: made a helper function since I don't think the probe should handle the mapping responsibility. https://github.com/DataDog/datadog-agent/blob/1441bc04732667ebe1c4966496e95ef75f4d7727/pkg/process/checks/process.go#L238-L252

pkg/process/checks/process.go
daniel-taf: I think the ticket needs further context to:
1. Provide better explanation of the intent behind the task (i.e. what is the benefit?). Currently it assumes the reader has the full context which may not be always true. Other people than you or me may read this. Since this technically isn't a priority for functionality/feature parity, keep in mind some details can be lost if it takes a while until we tackle this.
2. explain what you foresee the solution looking like for this specific task or even better provide concrete steps.

pkg/process/procutil/process_model.go
daniel-taf: Not saying you're wrong, I don't know myself, but what is the source here?

gengnamstyle: gotcha. Yeah the source was from the man page for procfs https://man7.org/linux/man-pages/man5/proc_pid_stat.5.html and the agent payload code https://github.com/DataDog/agent-payload/blob/master/proto/process/agent.proto#L633-L642

pkg/process/checks/process_nix_test.go
```diff
@@ -427,3 +442,705 @@ func TestProcessGPUTagging(t *testing.T) {
...
+			// Test second check runs without error and has correct number of chunks
+			processCheck.Run(testGroupID(0), getChunkingOption(tc.noChunking))
+			actual, err := processCheck.Run(testGroupID(0), getChunkingOption(tc.noChunking))
+			require.NoError(t, err)
+			assert.Len(t, actual.Payloads(), tc.expectedPayloadLength)
+		})
+	}
+}
+
+func TestProcessCheckZombieConfig(t *testing.T) {
```
daniel-taf: Why did we add this in this PR? Your changes shouldn't alter this behavior since the config is used at the probe level. We have an existing test: https://github.com/DataDog/datadog-agent/blob/main/pkg/process/checks/process_test.go#L544

pkg/process/checks/process_nix_test.go
```diff
@@ -427,3 +442,705 @@ func TestProcessGPUTagging(t *testing.T) {
...
+			actual, err := processCheck.runWLM(0, tc.realtimeCollection)
+
+			// The first run returns nothing because processes must be observed on two consecutive runs
+			expected1 := CombinedRunResult{}
+			require.NoError(t, err)
+			assert.Equal(t, expected1, actual)
+			// creation times should be the same
+			for _, wlmProc := range wlmProcs {
+				assert.Equal(t, wlmProc.CreationTime.Unix(), statsByPid1[wlmProc.Pid].CreateTime)
+			}
```
daniel-taf: I'm confused about why this needs to be checked. Doesn't seem to be able to change since the return is mocked and static.

pkg/process/checks/process_nix_test.go
```diff
@@ -427,3 +442,705 @@ func TestProcessGPUTagging(t *testing.T) {
...
+	}
+}
+
+// randRange returns a random number between min and max inclusive [min, max]
+func randRange(min, max int) int {
+	return rand.IntN(max+1-min) + min
+}
+
+// randomUnprivilegedProcessStats returns process stats with reasonable randomized data
+func randomProcessStats(createTime int64, withPriviledgedData bool) *procutil.Stats {
```
daniel-taf: Why are static values not sufficient? I think we only need to test the stats are there, so this make the test for complex than necessary.

pkg/process/checks/process.go
gengnamstyle: Added some more context!

pkg/process/checks/process_nix_test.go
```diff
@@ -427,3 +442,705 @@ func TestProcessGPUTagging(t *testing.T) {
...
+				require.Len(t, actual.RealtimePayloads(), 1)
+				rt := actual.RealtimePayloads()[0].(*model.CollectorRealTime)
+				assert.ElementsMatch(t, expectedStats, rt.Stats)
+				assert.Equal(t, int32(1), rt.GroupSize)
+				assert.Equal(t, int32(len(processCheck.hostInfo.SystemInfo.Cpus)), rt.NumCpus)
+			} else {
+				assert.Nil(t, actual.RealtimePayloads())
+			}
+
+			// THIRD RUN
```
daniel-taf: Why the third run?

pkg/process/checks/process_nix_test.go
```diff
@@ -427,3 +442,705 @@ func TestProcessGPUTagging(t *testing.T) {
...
+					Info:      processCheck.hostInfo.SystemInfo,
+					Hints:     &model.CollectorProc_HintMask{HintMask: 0b0},
+				},
+				&model.CollectorProc{
+					Processes: []*model.Process{modelProcFromWLMProc(proc6, statsByPid3[proc6.Pid], statsByPid2[proc6.Pid], []string{"process_context:cilium-agent-bpf-map-metrics"}, constantClock3.Now(), constantClock2.Now())},
+					GroupSize: int32(len(wlmProcs)),
+					Info:      processCheck.hostInfo.SystemInfo,
+					Hints:     &model.CollectorProc_HintMask{HintMask: 0b0},
+				},
+			}
```
daniel-taf: Is there real value in having more than >2-3 processes for the test? It seems to just add bloat.

pkg/process/checks/process_nix_test.go
```diff
@@ -427,3 +442,705 @@ func TestProcessGPUTagging(t *testing.T) {
...
+	for i, wlmProc := range wlmProcs {
+		pids[i] = wlmProc.Pid
+	}
+	mockProbe.On("StatsForPIDs", pids, mock.Anything).Return(statsByPid, nil)
+
+	for n := 0; n < b.N; n++ {
+		_, err := processCheck.runWLM(0, false)
+		require.NoError(b, err)
+	}
+}
```
daniel-taf: I don't think we need to include this

pkg/process/checks/process.go
```diff
@@ -243,11 +263,24 @@ func (p *ProcessCheck) runWLM(groupID int32, collectRealTime bool) (RunResult, e
...
+		pids[i] = wlmProc.Pid
+	}
+
+	statsForProcess, err := p.probe.StatsForPIDs(pids, p.timer.Now())
 	if err != nil {
 		return nil, err
 	}
 
+	// maintaining pre-existing variable types for ease of testing (don't have to change much following code)
+	//procs, err := p.probe.ProcessesByPID(time.Now(), true)
```
daniel-taf: mistake?

pkg/process/checks/process.go
```diff
@@ -243,11 +263,24 @@ func (p *ProcessCheck) runWLM(groupID int32, collectRealTime bool) (RunResult, e
...
+	for i, wlmProc := range wlmProcList {
+		pids[i] = wlmProc.Pid
+	}
+
+	statsForProcess, err := p.probe.StatsForPIDs(pids, p.timer.Now())
 	if err != nil {
 		return nil, err
 	}
 
+	// maintaining pre-existing variable types for ease of testing (don't have to change much following code)
```
daniel-taf:
```suggestion
	// map to common process type used by other versions of the check
```
I think the rest is unnecessary.

pkg/process/checks/process_nix_test.go
gengnamstyle: There is also a mirroring benchmark function for the existing run function. Should I remove that too?

pkg/process/checks/process_nix_test.go
gengnamstyle: I maintained the same amount of processes as the current run tests and added a one more process to remove the need for a [separate process_context test](https://github.com/DataDog/datadog-agent/blob/e4c205cc4795def9926a0e6bfd81bc318e232511/pkg/process/checks/process_test.go#L584) (using the last process with context `process_context:cilium-agent-bpf-map-metrics`) since it has a prefixed command line.

The main value added is the confirmation that everything is working as normal since we have the same number of tests and processes as the existing tests to ensure feature parity. However, other than that I think we could reduce the number of processes. The only bloat I see is creating the expected results?

pkg/process/checks/process_nix_test.go
gengnamstyle: I wanted to ensure that we tests that the normal and most common run check for customers is done because the first run which returns nothing and the second run will be the first time data is processed.

So to ensure that data continues to be processed like a check running multiple times, I want to do a third run.

pkg/process/checks/process_nix_test.go
gengnamstyle: It doesn't appear to be at the probe level as I see that the zombie processes are dropped within the check run and the config is stored on the check struct itself or am I missing something? https://github.com/DataDog/datadog-agent/blob/1441bc04732667ebe1c4966496e95ef75f4d7727/pkg/process/checks/process.go#L811

I believe there is value since the existing test doesn't test the usage of it in the `runWLM` function.

pkg/process/checks/process_nix_test.go
gengnamstyle: Yeah, I'll remove it. I wanted to ensure that the start time stored in wlm is the same as what was gathered in the stats but since they are both mocked this provides no value.

pkg/process/checks/process_nix_test.go
gengnamstyle: With the same static values on each process gives me less confidence that the rates and calculations are correct during the check. However, if they are randomized it gives me more confidence, and if I were to set different static values on each process, it seems like more bloat than the random data. Why do you think it is unnecessary?

pkg/process/checks/process.go
gengnamstyle: No, I added it for context as to why we do the mapping still, but as you noticed it's not super clear, so I reworded the comment and removed the function comment :)

zhuminyi (approved):

pkg/process/checks/process_nix_test.go
daniel-taf: No, I don't know why that one was added. But this benchmark isn't valuable since the WLM stuff is mocked.

pkg/process/checks/process_nix_test.go
daniel-taf: I find it unnecessary since that functionality is already [tested](https://github.com/DataDog/datadog-agent/blob/0b6796a62c2869c63b8ebcef0ee4cd37e17c4ca5/pkg/process/checks/process_common_test.go#L344). AFAICT, the WLM version isn't doing anything unique in that aspect, it uses the very same functions. As it stands now, the unit test is attempting to test too many things which makes it confusing to understand its purpose.

pkg/process/checks/process.go
```diff
@@ -243,11 +263,23 @@ func (p *ProcessCheck) runWLM(groupID int32, collectRealTime bool) (RunResult, e
...
 	if err != nil {
 		return nil, err
 	}
 
+	// map to common process type used by other versions of the check
+	procs := make(map[int32]*procutil.Process, len(wlmProcList))
+	for _, wlmProc := range wlmProcList {
+		procs[wlmProc.Pid] = mapWLMProcToProc(wlmProc, statsForProcess[wlmProc.Pid])
+	}
+
```
daniel-taf: Just thought of this: I think we can fix a lot of the code duplication between `runWLM` and `run` by having a helper function called `GetProcesses`. That way we don't need a separate run function. Some pseudocode:

```go
func run(...) {
	...
	procs, err := p.GetProcesses(...)
}

func GetProcesses(...) {
 	if p.useWLMCollection() {
		wlmProcList := p.wmeta.ListProcesses()
		pids := make([]int32, len(wlmProcList))
		for i, wlmProc := range wlmProcList {
			pids[i] = wlmProc.Pid
		}
	
		statsForProcess, err := p.probe.StatsForPIDs(pids, p.timer.Now())
		if err != nil {
			return nil, err
		}
	
		// map to common process type used by other versions of the check
		procs := make(map[int32]*procutil.Process, len(wlmProcList))
		for _, wlmProc := range wlmProcList {
			procs[wlmProc.Pid] = mapWLMProcToProc(wlmProc, 	statsForProcess[wlmProc.Pid])
		}
	}
	procs, err := p.probe.ProcessesByPID(time.Now(), true)
	if err != nil {
		return nil, err
	}
	return procs
}
```
This way we can leverage our existing unit tests as much as possible and don't need to add tests for anything other than the process collection. Overall, the implementation is much simpler imo. WDYT?

pkg/process/checks/process_nix_test.go
daniel-taf: Similar to the case with the real time stats, this feels unnecessary to me since so little has changed compared to the original implementation of `run`. But, I understand that technically speaking the codepath is different so we should have coverage. I think [this approach](https://github.com/DataDog/datadog-agent/pull/38817#discussion_r2219614996) can fix this since the original codepath will still be used and thus the original unit tests still apply.

daniel-taf: Forgot to add, this is where the config is used in the probe:
https://github.com/DataDog/datadog-agent/blob/main/pkg/process/procutil/process_linux.go#L265

This was the main place we were worried about to prevent OOMs.

pkg/process/checks/process.go
```diff
@@ -59,6 +60,7 @@ func NewProcessCheck(config pkgconfigmodel.Reader, sysprobeYamlConfig pkgconfigm
 		gpuSubscriber:       gpuSubscriber,
 		statsd:              statsd,
 		grpcServerTLSConfig: grpcServerTLSConfig,
+		timer:               clock.New(),
```
robertjli: I would just call it `clock`. It does shadow the package name unfortunately, but IMO it makes more semantic sense than `timer` and also it's the general standard in the repo ([code search](https://github.com/search?q=repo%3ADataDog%2Fdatadog-agent%20clock.Clock&type=code))

pkg/process/procutil/process_model.go
robertjli: Should we just delete the comment?

pkg/process/checks/process_nix_test.go
```diff
@@ -427,3 +442,693 @@ func TestProcessGPUTagging(t *testing.T) {
...
+		//ContainerId:            "", // TODO: container data not yet queried from wlm
+		//ContainerKey:           0, // TODO: container data not yet queried from wlm
+		VoluntaryCtxSwitches:   uint64(stats.CtxSwitches.Voluntary),
+		InvoluntaryCtxSwitches: uint64(stats.CtxSwitches.Involuntary),
+		ByteKey:                nil, // not used
+		ContainerByteKey:       nil, // not used
+		Networks:               nil, // not used
+		ProcessContext:         processContext,
+		Tags:                   nil,
+	}
```
wiyu: Some nits about style
- Unnecessary newline. 
- Also why set values to nil when the default value is already implicitly nil as that's a default value, you can just skip setting it and save on the lines. 
- Also generally un-exported functions or helpers are at the bottom of a file. Check out https://github.com/uber-go/guide/blob/master/style.md#function-grouping-and-ordering

pkg/process/checks/process_nix_test.go
```diff
@@ -427,3 +442,693 @@ func TestProcessGPUTagging(t *testing.T) {
 		})
 	}
 }
+
+func processCheckWithMockProbeWLM(t *testing.T, elevatedSystemProbePermissions bool) (*ProcessCheck, *probeMocks.Probe, *wmimpl.MockWLM) {
+	t.Helper()
```
wiyu: This doesn't seem necessary as it doesn't look like the test can fail here.

pkg/process/procutil/process_model.go
wiyu: Let's add the link to the comment so we can reference that in the future.

pkg/process/checks/process.go
gengnamstyle: done!

pkg/process/checks/process_nix_test.go
gengnamstyle: totally agreed. Removed both

pkg/process/checks/process_nix_test.go
gengnamstyle: removed!

pkg/process/checks/process_nix_test.go
gengnamstyle: removed the third run. Did not add additional tests here for now

pkg/process/checks/process_nix_test.go
gengnamstyle: removed!

pkg/process/checks/process_nix_test.go
gengnamstyle: removed

pkg/process/checks/process_nix_test.go
gengnamstyle: removed thanks!

pkg/process/checks/process_nix_test.go
gengnamstyle: I think there is value in keeping this function for now since it provides data that actually isn't currently tested on the complete `run` codepath such as `IOStat`, `CPUTime`, `CtxSwitches` etc

Current process creation:
https://github.com/DataDog/datadog-agent/blob/150e7c6f7ad5544e530151e941811e0a8f7ecb6a/pkg/process/checks/process_common_test.go#L45-L63

There is remaining work to ensure that these untested fields are incorporated in the main unit test `run` codepath ^
and that the current code with a subset of stats is also randomized. Thoughts?

pkg/process/checks/process_fallback.go
```diff
@@ -0,0 +1,21 @@
...
+
+//go:build !linux
+
+package checks
+
+import (
+	"github.com/DataDog/datadog-agent/pkg/process/procutil"
+)
+
+// processesByPID returns the processes by pid from the system probe
```
daniel-taf: Its not from the system probe, process probe is just the name of the interface. The system probe is only used to collect IO stats, if enabled. Confusing name, I suppose.
```suggestion
// processesByPID returns the processes by pid from the process probe
```

pkg/process/checks/process_fallback_test.go
```diff
@@ -0,0 +1,68 @@
...
+			description:      "wlm collection DISABLED, with stats ENABLED",
+			useWLMCollection: false,
+			collectStats:     true,
+		},
+		{
+			description:      "wlm collection DISABLED, with stats DISABLED",
+			useWLMCollection: false,
+			collectStats:     false,
+		},
+	} {
```
daniel-taf: I think this test can be simplified to remove cases based on `collectStats` it doesn't change anything about the test. It's just a value passed into the function.

pkg/process/checks/process_linux.go
```diff
@@ -0,0 +1,61 @@
...
+//go:build linux
+
+package checks
+
+import (
+	workloadmetacomp "github.com/DataDog/datadog-agent/comp/core/workloadmeta/def"
+	"github.com/DataDog/datadog-agent/pkg/process/procutil"
+)
+
+func mapWLMProcToProc(wlmProc *workloadmetacomp.Process, stats *procutil.Stats) *procutil.Process {
```
daniel-taf: helpers go at the bottom per Will's comment:
https://github.com/DataDog/datadog-agent/pull/38817#discussion_r2222991656

pkg/process/checks/process_linux_test.go
```diff
@@ -0,0 +1,206 @@
...
+	"time"
+
+	wmdef "github.com/DataDog/datadog-agent/comp/core/workloadmeta/def"
+	wmimpl "github.com/DataDog/datadog-agent/comp/core/workloadmeta/impl"
+	"github.com/DataDog/datadog-agent/pkg/process/procutil"
+	probemocks "github.com/DataDog/datadog-agent/pkg/process/procutil/mocks"
+	"github.com/stretchr/testify/assert"
+)
+
+func wlmProcessWithCreateTime(pid int32, spaceSeparatedCmdline string, creationTime int64) *wmdef.Process {
```
daniel-taf, line 93: helpers at the bottom

pkg/process/checks/process_linux_test.go
```diff
@@ -0,0 +1,206 @@
...
+			// TESTING
+			actual, err := processCheck.processesByPID(tc.collectStats)
+			assert.NoError(t, err)
+			assert.Equal(t, expected, actual)
+		})
+	}
+}
+
+// TestWLMProcessesByPID tests processesByPID normal probe usage when WLM collection is OFF
+func TestWLMCollectedProcessesByPIDOff(t *testing.T) {
```
daniel-taf: I don't think we need this test. This case is covered by this one:
https://github.com/DataDog/datadog-agent/blob/5813fd2f9903965f373148054d93b8a2facbefd0/pkg/process/checks/process_test.go#L102
For clarity, I think you should you edit that existing test to set the WLM config explicitly and remove the one below.

pkg/process/checks/process_linux_test.go
```diff
@@ -0,0 +1,206 @@
...
+func createTestWLMProcessStats(wlmProcs []*wmdef.Process, elevatedPermissions bool) map[int32]*procutil.Stats {
+	statsByPid := make(map[int32]*procutil.Stats, len(wlmProcs))
+	for _, wlmProc := range wlmProcs {
+		statsByPid[wlmProc.Pid] = randomProcessStats(wlmProc.CreationTime.Unix(), elevatedPermissions)
+	}
+	return statsByPid
+}
+
+// TestWLMProcessesByPID tests processesByPID map creation when WLM collection is ON
+func TestWLMCollectedProcessesByPIDOn(t *testing.T) {
```
daniel-taf:
```suggestion
func TestProcessCheckWLM(t *testing.T) {
```

pkg/process/checks/process_test.go
```diff
@@ -471,29 +474,11 @@ func TestProcessWithNoCommandline(t *testing.T) {
 	assert.Empty(t, proc.Command.Args)
 }
 
-func BenchmarkProcessCheck(b *testing.B) {
```
daniel-taf: I don't think we need to remove this one. Not 100% sure why it was added but I think it's to test the aspects of the check outside of process collection itself (i.e. payload normalizing), so it may have value in that context. I didn't think we needed the WLM version since that part is the same.

pkg/process/checks/process_fallback.go
daniel-taf: Also, some more context would be good on why we need different versions per build.

pkg/process/checks/process_linux.go
```diff
@@ -0,0 +1,61 @@
...
+		Exe:     wlmProc.Exe,
+		Comm:    wlmProc.Comm,
+		Cmdline: wlmProc.Cmdline,
+		Uids:    wlmProc.Uids,
+		Gids:    wlmProc.Gids,
+		Stats:   stats,
+	}
+}
+
+// processesByPID returns the processes by pid from different sources depending on the configuration (system probe or workloadmeta)
```
daniel-taf:
```suggestion
// processesByPID returns the processes by pid from different sources depending on the configuration (process probe or workloadmeta)
```

pkg/process/checks/process_fallback.go
gengnamstyle: Changed the wording. 

I have noticed the differing names but I find it confusing because when processes are fetched since there is no separate component when we do or do not collect elevated permission stats. It's only a bool check of whether to include the privileged data.

https://github.com/DataDog/datadog-agent/blob/deb796fa8fd53a5673c852ee233666afdac1b4cb/pkg/process/procutil/process_linux.go#L303-L305

Where is the distinction here?

pkg/process/checks/process_fallback_test.go
gengnamstyle: Removed

pkg/process/checks/process_linux_test.go
gengnamstyle: Changed. Though this turns this specific case into an implicit test. Do you think the loss of the readability is worth it? There is also the https://github.com/DataDog/datadog-agent/blob/5813fd2f9903965f373148054d93b8a2facbefd0/pkg/process/checks/process_test.go#L124

Which also redundantly tests the first run as well. 

As a side, it seems potentially confusing that the existing tests only mock the probe and the not parent function which seems to imply we are lacking tests for the codepath when wlm collection is ON.

pkg/process/checks/process_test.go
gengnamstyle: added back.

pkg/process/checks/process_linux_test.go
gengnamstyle: renamed to `TestProcessesByPIDWLM` since I'm not testing the process check's main function `run`

pkg/process/checks/process_fallback.go
daniel-taf: That is odd that the probe has the ability to to try to get those stats since we call for it from the system probe [here](https://github.com/DataDog/datadog-agent/blob/main/pkg/process/checks/process.go#L258) which collects it [here](https://github.com/DataDog/datadog-agent/blob/main/cmd/system-probe/modules/process.go#L80). Not sure in which case `p.elevatedPermissions` is true and the stats can actually be collected.

pkg/process/checks/process_linux_test.go
daniel-taf: For now, rename the existing test: `TestProcessCheckWithProbe`. Whether the test you removed existed or not, the implicitness would still exist as you say. I think some re-org of these tests might be needed when we do the cutover however. It would likely be cleaner to have these tests in build-specific files (fallback and linux) since there is an actual difference between builds now which was not the case when this test was created.

pkg/process/checks/process_linux_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+			// MOCKING
+			nowSeconds := mockConstantClock.Now().Unix()
+			proc1 := wlmProcessWithCreateTime(1, "git clone google.com", nowSeconds)
+			proc2 := wlmProcessWithCreateTime(2, "mine-bitcoins -all -x", nowSeconds-1)
+			proc3 := wlmProcessWithCreateTime(3, "datadog-agent --cfgpath datadog.conf", nowSeconds+2)
+			proc4 := wlmProcessWithCreateTime(4, "/bin/bash/usr/local/bin/cilium-agent-bpf-map-metrics.sh", nowSeconds-3)
+			procs := []*wmdef.Process{proc1, proc2, proc3, proc4}
+			statsByPid := make(map[int32]*procutil.Stats)
+			mockWLM.EXPECT().ListProcesses().Return(procs).Once()
```
davidor: I think this could be replaced with `Set` calls.
If I’m not mistaken, the other use of the mock is in another file and just checks that ListProcesses wasn’t called.
If that’s the case, I think adding a new mock based on mockery isn’t worth the added complexity.

pkg/process/checks/process_linux_test.go
gengnamstyle: renamed!

pkg/process/checks/process_linux_test.go
gengnamstyle: Removed the generated mock and changed to set calls. I agree that the current usage does not really warrant the generated mock.

However, it does feel "off" to have to insert data into the store and effectively test WLM when I just want to use specific functions, but the test setup could be moved to a helper function 👍

daniel-taf (approved):

davidor (approved):
:+1: for container-platform

There's only one file that we own (`comp/core/workloadmeta/collectors/internal/process/process_collector.go`) and it's a small change.

pkg/process/checks/process_fallback.go
gengnamstyle: `p.elevatedPermissions` is true when the process module is enabled for the system probe: 

https://github.com/DataDog/datadog-agent/blob/a747ad4ce42c8087210a5998379f76e26da5a2b2/pkg/process/checks/process.go#L149-L150

so it appears that we are getting privileged data twice. Made a loose ends ticket to fix this bug: https://datadoghq.atlassian.net/browse/CXP-2488

daniel-taf: Right, except it doesn't actually mean the agent has the privileges though, its just a flag to allow attempts to retrieve privileged data. It likely doesn't actually work and is a remnant of something from the past. I think the card is better framed as "Investigate use of elevated privileges in process probe".

gengnamstyle: I changed the title, but unfortunately, the process probe does work after testing locally and enabling the system probe.

daniel-taf: Hm, testing locally how? If the agent was running as root then that would make sense.

pkg/process/checks/process_linux.go
```diff
@@ -0,0 +1,68 @@
...
+		for _, wlmProc := range wlmProcList {
+			procs[wlmProc.Pid] = mapWLMProcToProc(wlmProc, statsForProcess[wlmProc.Pid])
+		}
+		return procs, nil
+	}
+	procs, err := p.probe.ProcessesByPID(p.clock.Now(), collectStats)
+	if err != nil {
+		return nil, err
+	}
+	return procs, nil
```
wiyu, line 51: I know you merged this already, but in a follow-up you could just be simplify this to:

`return p.probe.ProcessesByPID(p.clock.Now(), collectStats)`

pkg/process/checks/process_fallback.go
gengnamstyle: ah yes that makes sense

PR #38833, opened 2025-07-16
Add new events serializer to replace three old ones

vickenty:
### What does this PR do?

Add a new events marshaler to eventually  replace three older ones.

### Motivation

Simplify the events marshaler implementation, as well as remove last use of `split.Payloads` ([draft removal PR](https://github.com/DataDog/datadog-agent/pull/38834/files))

Basic smoke test that sending events still works, as well as unit tests.

We need to live with even more code for a while


dustmop (approved):

pkg/config/setup/config.go
```diff
@@ -1379,6 +1379,7 @@ func serializer(config pkgconfigmodel.Setup) {
 	config.BindEnvAndSetDefault("serializer_max_series_uncompressed_payload_size", 5242880)
 	config.BindEnvAndSetDefault("serializer_compressor_kind", DefaultCompressorKind)
 	config.BindEnvAndSetDefault("serializer_zstd_compressor_level", DefaultZstdCompressionLevel)
+	config.BindEnvAndSetDefault("serializer_use_events_marshaler_v2", true)
```
remeh, line 1382: Confidence check: we switch to the new one in next release?

pkg/serializer/internal/metrics/events.go
```diff
@@ -280,3 +292,240 @@ func (events Events) CreateMarshalersBySourceType() []marshaler.StreamJSONMarsha
...
+			if ok {
+				break
+			}
+		}
+	}
+	if err := e.closePayload(); err != nil {
+		return nil, err
+	}
+
+	return e.payloads, nil
```
remeh, line 530: Can't you
```
return e.payloads, e.closePayload()
```

You really don't want to return `e.payloads` in case of err?

pkg/serializer/internal/metrics/events.go
```diff
@@ -280,3 +292,240 @@ func (events Events) CreateMarshalersBySourceType() []marshaler.StreamJSONMarsha
...
+		return events[i].SourceTypeName < events[j].SourceTypeName
+	})
+
+	stream := jsoniter.NewStream(jsonConfig, nil, 4096)
+
+	writeEventsHeader(stream)
+	header := slices.Clone(stream.Buffer())
+	stream.Reset(nil)
+
+	stream.WriteArrayEnd()
```
remeh, line 353: What's this for?

pkg/serializer/internal/metrics/events.go
```diff
@@ -280,3 +292,240 @@ func (events Events) CreateMarshalersBySourceType() []marshaler.StreamJSONMarsha
...
+		maxPayloadSize:      config.GetInt("serializer_max_payload_size"),
+		maxUncompressedSize: config.GetInt("serializer_max_uncompressed_payload_size"),
+	}
+}
+
+func (e *eventsMarshaler2) flushPayload() error {
+	err := e.closePayload()
+	if err != nil {
+		return nil
+	}
```
remeh: What would be the reason of this failing? Should we log something?
Also, returning `nil` looks suspicious, is it a mistake?

remeh (approved):
Made a few comments but it looks mostly good to me!

pkg/config/setup/config.go
vickenty: Yes.

pkg/serializer/internal/metrics/events.go
vickenty: > `return e.payloads, e.closePayload()`

Go evaluates left to right, so this would loose last payload if `closePayload` needs to re-allocate `e.payloads`.

pkg/serializer/internal/metrics/events.go
vickenty: It's a closing bracket for the opening one we write on [line 484](https://github.com/DataDog/datadog-agent/pull/38833/files#diff-11bbd2851c6d284ebe29884cd690539a8c030bde3f11c3c29c4cd0523d615534R484).

It needs to be in the footer for the compressor to reserve space for it.

PR #38838, opened 2025-07-16
CAP-2612 fix noisy error logs when scrubbing large commands.

kangyili:
### What does this PR do?

It was pointed out that the pod check has been consistently reporting this error:
```
2025-06-17 09:47:40 UTC | CORE | ERROR | (pkg/orchestrator/redact/pod.go:97 in func1) | Failed to parse cmd from pod, obscuring whole command
```

It turns out there is a slice bounds out of range panic in `scrubContainer` when processing containers with complex multi-line scripts.

```
panic: runtime error: slice bounds out of range [:X] with capacity Y
```

- The function calculated word count by splitting command strings on spaces
- However, ScrubSimpleCommand tokenizes differently and returns fewer elements
- When trying to slice the scrubbed result using the original word count, it exceeded the array bounds

This PR adds bounds checking and makes error logs more descriptive.

### QA
No QA needed as unit test can cover it


hestonhoffman (approved):

pkg/orchestrator/redact/pod_test.go
```diff
@@ -714,6 +714,32 @@ func getScrubCases() map[string]struct {
...
+if [ "$CLOUDPROVIDER" == "azure" ]; then
+  source /etc/datadog/azureSASEnv;
+
+  export SPARK_HISTORY_OPTS="$SPARK_HISTORY_OPTS \
+    -Dspark.hadoop.fs.azure.local.sas.key.mode=true;
+fi;
+/opt/spark/bin/spark-class org.apache.spark.deploy.history.HistoryServer;
+	`},
+			},
+			expected: v1.Container{
```
stzou, line 740: Can we add a test case where we have Command + Args and validate that args is moved in this scenario.

stzou (approved):

pkg/orchestrator/redact/pod_test.go
kangyili: [e5a3b16](https://github.com/DataDog/datadog-agent/pull/38838/commits/e5a3b16aa4bccc6ba5c9c453a4a5f97b71324640)

flmmartins:
Hello @kangyili, 
We are seeing this issue when updating Datadog. When is version 7.70.0 is going to be released? Is there a workaround meanwhile?

flmmartins:
@kangyili Is there a launch date for 7.7.0 to be released? We are facing this issue. Thanks!

PR #38840, opened 2025-07-16
[ACIX-926]: Enable retries for all end-to-end tests

Ishirui:
### What does this PR do?

Enables the retry mechanism introduced in #37862 for all e2e tests. It was previously only enabled on jobs in `new-e2e-installer` and `new-e2e-installer-script` stages for verification purposes.

### Motivation

Improve reliability of CI and reduce impact of e2e test flakes.

N/A

- Increased CI time when tests are "reliably" broken

### Additional Notes

I set the `MAX_RETRIES_FLAG` variable directly in the root `.new-e2e-template` instead of setting it manually in all (~100) leaf jobs. This might not be the way we want to do it as it could make it less clear that tests have this retry functionality built-in.

Note however that it is still possible to disable/change the number of retries for an individual job by overriding `MAX_RETRIES_FLAG` (making it `""` for example) at the job level.


Ishirui:
Decided to do this more incrementally, see ticket desc for details

PR #38846, opened 2025-07-16
[AGENTCFG-320] properly handling time.Duration types in nodetreemodel

rahulkaukuntla:
### What does this PR do?
By definition, the time.Duration type represents the elapsed time between two instants as an int64 _nanosecond_ count. However, we were just assuming that the value passed into a time.Duration value represented seconds and not nanoseconds, causing [this test](https://github.com/DataDog/datadog-agent/blob/884793f5cd8e745a019b56d303270579e739a864/comp/metadata/host/hostimpl/host_test.go#L46) to fail. We properly handle time.Duration values with this PR.


pkg/config/nodetreemodel/leaf_node.go
```diff
@@ -41,6 +41,10 @@ func isSlice(v interface{}) bool {
 }
 
 func newLeafNode(v interface{}, source model.Source) Node {
+	if timeObject, ok := v.(time.Duration); ok {
+		// The time.Duration type represents the elapsed time between two instants as an int64 nanosecond count
+		return &leafNodeImpl{val: timeObject.Nanoseconds(), source: source}
+	}
```
dustmop: I don't think this is the right approach. Viper stores data as-in, and then relies on calling code to use the accessor for the type they want to get. Modifying data as it is being inserted into the config breaks this behavior. Practically speaking, this means that the call `config.Get("setting")` will no longer return the right value because the original value doesn't exist. Similarly, `AllSettings()` is also affected.

I think the actual issue is that our conversion code doesn't handle type conversion correctly.

dustmop (changes requested):

dustmop:
I pushed a commit https://github.com/DataDog/datadog-agent/commit/37160c406f3249dc3805215f3b0787df5b697ee6 on top of your branch, it fixes `UnmarshalKey` by directly copies values of the same type from source to destination. This fixes the issue with `time.Duration` that was breaking the [unit test](https://github.com/DataDog/datadog-agent/blob/884793f5cd8e745a019b56d303270579e739a864/comp/metadata/host/hostimpl/host_test.go#L46) in question. You can reset your branch to this commit to get the fix.

Could you add some more unit tests that exercise the code handling `int` and `int64` values? For example, what if I have a struct with a field of type `int32` and I try to UnmarshalKey input data that has `int64` numbers? And vice versa.

rahulkaukuntla:
Added a unit test that checks every combination of `int`, `int32`, and `int64`.

dustmop (approved):
LGTM, thanks for the extra tests!

PR #38861, opened 2025-07-16
pkg/dyninst: Fix unsafe.Pointer capture/decoding

grantseltzer:
### What does this PR do?

- Defines a new `ir.VoidPointerType` which handles the case of unsafe.Pointers for Go
- Adds a test case for unsafe.Pointer

### Motivation

unsafe.Pointer arguments were breaking decoding because they lacked a reflect.Kind.

Added a test case in integration testing.

### DIFF
--- a/pkg/dyninst/decode/marshal.go
+++ b/pkg/dyninst/decode/marshal.go
@@ -201,7 +201,8 @@ func (d *Decoder) encodeValue(
 	); err != nil {
 		return err
 	}
-	if err := d.encodeValueFields(enc,
+	if err := d.encodeValueFields(
+		enc,
 		dataItems,
 		currentlyEncoding,
 		irType,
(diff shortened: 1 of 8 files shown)


pkg/dyninst/ir/types.go
```diff
@@ -108,6 +109,14 @@ type BaseType struct {
 
 func (t *BaseType) irType() {}
 
+// VoidPointerType is a type that represents an unsafe pointer.
```
ajwerner: nit: I might say `VoidPointerType is a type that represents a pointer to a value of an unknown type. In Go, the unsafe.Pointer type is such a type.` or something like that.

pkg/dyninst/decode/marshal.go
```diff
@@ -233,6 +234,15 @@ func (d *Decoder) encodeValueFields(
 			return err
 		}
 		return encodeBaseTypeValue(enc, v, data)
+	case *ir.VoidPointerType:
+
```
ajwerner: nit: stray newline

ajwerner (approved):
Thank you!

PR #38876, opened 2025-07-17
Detect uretprobe seccomp bug

usamasaqib:
### What does this PR do?
This PR builds a runtime detection for a seccomp bug which may cause a probed application to segfault. 

https://lore.kernel.org/lkml/CAHsH6Gs3Eh8DFU0wq58c_LF8A4_+o6z456J7BidmcVY2AqOnHQ@mail.gmail.com/

Changes covered by existing testing on Ubuntu 24.10 with kernel `6.11.0-8-generic`

### Additional Notes

#incident-40539


YoannGh (approved):

pkg/ebpf/kernelbugs/kernel_bugs_amd64.go
```diff
@@ -0,0 +1,98 @@
...
+
+// TriggerProgram holds the bytecode for a userspace program used for detecting a bug in the kernel
+//
+//go:embed c/detect-seccomp-bug
+var TriggerProgram []byte
+
+// HasUretprobeSyscallSeccompBug returns true if the running kernel blocks the uretprobe syscall in seccomp
+// This can cause a probed application running within a seccomp context to segfault.
+// https://lore.kernel.org/lkml/CAHsH6Gs3Eh8DFU0wq58c_LF8A4_+o6z456J7BidmcVY2AqOnHQ@mail.gmail.com/
+func HasUretprobeSyscallSeccompBug() (bool, error) {
```
gjulianm: Can we memoize this result? Specially with the sleep, if multiple callers try to get the value it might delay the startup by a lot.

pkg/ebpf/kernelbugs/kernel_bugs_amd64.go
```diff
@@ -0,0 +1,98 @@
...
+			log.Print(err)
+		}
+	}()
+
+	if err := m.Start(); err != nil {
+		return false, err
+	}
+
+	// wait for uprobe to be attached
+	time.Sleep(3 * time.Second)
```
gjulianm: Maybe a bit excessive, do we have a better way to check if a uprobe is correctly attached? If not, is it possible to reduce the wait time a bit?

usamasaqib: Looking at it again, this sleep should not be here but in the `c` code. 

We can check the address space of the C program to know if the uprobe is attached for sure, but that seems unnecessary when a sleep will suffice.

Reducing the time adds the risk that we incorrectly determine the bug is not present.

tasks/system_probe.py
```diff
@@ -343,6 +350,35 @@ def ninja_network_ebpf_programs(nw: NinjaWriter, build_dir, co_re_build_dir):
         ninja_network_ebpf_co_re_program(nw, infile, outfile, network_co_re_flags)
 
 
+def ninja_kernel_bug_binaries(nw: NinjaWriter, arch: str | Arch):
+    arch = Arch.from_str(arch)
+
+    # do not build for arm64
+    if arch.ci_arch == "arm64":
```
gjulianm:
```suggestion
    if arch == ARCH_ARM64
```

(import from `tasks.libs.types.arch`)

tasks/system_probe.py
```diff
@@ -343,6 +350,35 @@ def ninja_network_ebpf_programs(nw: NinjaWriter, build_dir, co_re_build_dir):
...
+    for binary in embedded_bins:
+        infile = os.path.join(ebpf_c_dir, f"{binary}.c")
+        outfile = os.path.join(ebpf_c_dir, binary)
+        cc = "gcc"
+
+        nw.build(
+            inputs=[infile],
+            outputs=[outfile],
+            rule="cbin",
+            variables={"cc": cc, "cflags": "-static", "ldflags": "-lseccomp"},
```
gjulianm, line 372: Looks like the size quality gate is failing. Maybe we can add `-Os -s` to reduce the size of this binary.

gjulianm (approved):

hmahmood (approved):

chouetz (approved):

DanielLavie (approved):

akarpz (approved):

PR #38877, opened 2025-07-17
[RC] WebSocket data gathering

domodwyer:
### What does this PR do?

Cause Agents that have Remote Config enabled to connect to the RC backend using a WebSocket connection every 24h (at `/api/v0.2/echo-test` - server side not yet implemented).

This connectivity test can be disabled by setting `remote_configuration.no_websocket_echo=true` in config. The test is opt-out due to the need to gather connectivity data from a representative sample of deployed Agents to evaluate the feasibility of switching to this new protocol in the future, or if it would cause significant breakage to existing users.

Every 24h a WebSocket connection is established, and the server sends a series of small data frames to the client, which immediately echos them back. It is not possible to use the WebSocket to induce any other behaviour in the Agent.

The implementation is careful to enforce safety limits (timeouts + payload size limits), avoids logging anything above DEBUG, and implements a panic recovery handler to minimise bug risk that might affect the production workload.

### Motivation

Gather connectivity / deliverability / latency / traffic management metrics, aiding development of a new RC / Agent protocol.

Tests are included for typical and edge case functionality.

The impact of this test should be negligible, and is designed to fail without causing any further effects.

### Additional Notes

Part of RC-2602.

This is my first PR to the Agent repo and I'm not really familiar with the codebase.

---

* feat(rc): WebSocket connectivity (9cd1980e6e)
      
      Add functionality to the api.HTTPClient to enable it to establish
      connections to a WebSocket service in the RC backend.
      
      Dialling a new WebSocket connection takes care to replicate the auth
      credential transmission mechanism used by the HTTP client, as well as
      the user-provided TLS & proxy config.

* feat(rc): implement WebSocket echo test (62657991e1)
      
      This commit implements a simple WebSocket echo test, connecting to the
      RC backend (at "/api/v0.2/echo-test") and exchanging a series of data
      frames.
      
      This test allows RC to evaluate the connectivity and performance of
      WebSockets in the wild, as a potential transport for a new protocol
      between the Agent and RC.
      
      Part of RC-2602.

* docs(rc): CoreAgentService.Stop() multiple calls (2a939cdcfb)
      
      Document that the Stop() function may be invoked multiple times: it is
      called >1 by at least some tests, even if not in the "prod" path.

* feat(rc): WebSocket connectivity tests every 24h (bd9f392e96)
      
      Make Remote Config enabled Agents perform periodic WebSocket
      connectivity tests every 24h.
      
      This connectivity test is required to evaluate real-world connectivity,
      through HTTP(S) proxies and egress filters, and gather latency and
      deliverability measurements that guide development of a new RC / Agent
      protocol.
      
      These tests are best effort and lightweight, exchanging a handful of
      frames with little memory or CPU overhead. They do not log any output
      above DEBUG level, except in the case of a panic which is caught to
      prevent the test affecting production workloads and logged at WARN.
      
      Once data gathering is complete, RC may stop servicing requests made to
      the test path ("/api/v0.2/echo-test") without causing any effects or
      logs on the Agent (tested).
      
      Tests can be explicitly disabled by configuring:
      
          remote_configuration.no_websocket_echo=true
      
      Part of RC-2602.


pkg/config/remote/api/http.go
```diff
@@ -256,6 +260,59 @@ func (c *HTTPClient) UpdateAPIKey(apiKey string) {
...
+	transport, ok := c.client.Transport.(*http.Transport)
+	if !ok {
+		return nil, errors.New("failed to extract transport config from HTTP client")
+	}
+
+	dialer := &websocket.Dialer{
+		HandshakeTimeout: 30 * time.Second,
+		TLSClientConfig:  transport.TLSClientConfig,
+		Proxy:            transport.Proxy,
+	}
```
domodwyer, line 279: WebSocket connections use the same TLS & proxy configuration as the RC HTTP client, which can be (and are) configured by the user in some deployments.

pkg/config/remote/api/http.go
```diff
@@ -256,6 +260,59 @@ func (c *HTTPClient) UpdateAPIKey(apiKey string) {
...
+
+	// The WebSocket request MUST include the same auth credentials as the plain
+	// HTTP requests.
+	c.headerLock.RLock()
+	headers := maps.Clone(c.header)
+	c.headerLock.RUnlock()
```
domodwyer, line 285: The WebSocket connections are authenticated in the same way as the RC HTTP requests.

pkg/config/remote/api/http_test.go
```diff
@@ -0,0 +1,166 @@
...
+			path:        "/",
+			serverHTTP2: true,
+			serverTLS:   false,
+		},
+		{
+			desc:        "path",
+			path:        "/api/v0.2/echo",
+			serverHTTP2: false,
+			serverTLS:   false,
+		},
```
domodwyer, line 68: Varying transport configurations are tested:

* TLS on/off
* HTTP vs HTTP2

This is probably going to be relevant in deployments that use non-transparent HTTP egress proxies irrespective of what RC uses in the backend.

pkg/config/remote/service/service.go
```diff
@@ -471,6 +476,14 @@ func NewService(cfg model.Reader, rcType, baseRawURL, hostname string, tagsGette
 
 	clock := clock.New()
 
+	// WebSocket test actor - must call Start() to spawn the background task.
+	var websocketTest startstop.StartStoppable
+	if cfg.GetBool("remote_configuration.no_websocket_echo") {
+		websocketTest = &noOpRunnable{}
+	} else {
+		websocketTest = NewWebSocketTestActor(http)
+	}
```
domodwyer, line 489: My intention is to make it possible to disable this test, but I'm not familiar with how the config system works. 

Is this call sufficient to enable a user to disable the test or is there more config plumbing I need to do?

pkg/config/remote/service/websocket_echo.go
```diff
@@ -0,0 +1,98 @@
...
+	if s.stopCh != nil {
+		panic("attempt to start WebSocketTestActor more than once")
+	}
+
+	s.stopCh = make(chan struct{})
+
+	go func() {
+		s.run()
+	}()
+}
```
domodwyer, line 51: I make the assumption that [the `CoreAgentService.Start()` method](https://github.com/DataDog/datadog-agent/blob/5624e49c09e450b27333fbb0b6969a763d951ac1/pkg/config/remote/service/service.go/#L542) is called once (and therefore never concurrently).

I also assume that [the `CoreAgentService.Stop()` method](https://github.com/DataDog/datadog-agent/blob/5624e49c09e450b27333fbb0b6969a763d951ac1/pkg/config/remote/service/service.go/#L639) _can_ be called more than once (at least one test does) but never concurrently.

Are these assumptions correct?

domodwyer (commented):
I've left some review notes and questions :+1:

domodwyer:
(rebased to resolve merge conflict)

comp/remote-config/rcwebsocket/echo.go
```diff
@@ -0,0 +1,126 @@
...
+	cb := conn.PingHandler()
+	conn.SetPingHandler(
+		func(message string) error {
+			bumpReadDeadline(conn)
+			return cb(message)
+		},
+	)
+
+	// Perform the frame echo test routine.
+	n := uint(0)
```
mellon85:
```suggestion
	attemptCount := uint(0)
```

pkg/config/remote/service/service.go
mellon85: Yeah, the only other alternative is adding a DD signed configurations to sync it to every org

pkg/config/remote/service/websocket_echo.go
```diff
@@ -0,0 +1,98 @@
...
+func (s *WebSocketTestActor) run() {
+	// This test loop is best effort - it SHOULD never kill the Agent process,
+	// even if it encounters a bug or protocol error.
+	//
+	// If the test panics, this run() call exits gracefully and the test is not
+	// performed again until the agent is restarted. Stop() may be called safely
+	// after recovery has occurred.
+	defer func() {
+		if err := recover(); err != nil {
+			log.Warnf("unexpected websocket connectivity test failure: %s", err)
```
mellon85, line 72: only Debug, to avoid causing concerns around the test

domodwyer: We chatted and we'll leave this as-is. It's the only non-DEBUG log, and it was intentional - if this code panics, we want to know before we start using it for production workloads.

mellon85 (approved):

releasenotes/notes/RC-WebSocket-connectivity-tests-9f04946537a732ad.yaml
```diff
@@ -0,0 +1,40 @@
+# Each section from every release note are combined when the
```
michaelcretzman: looks like this is still in Draft as there is only placeholder text here.

domodwyer: Sorry @michaelcretzman I must have added these by accident trying to get the python tool working :D 

After internal discussion we don't intend to add release notes for this (if you think we should that's fine - just let me know :+1:)

PR #38892, opened 2025-07-17
dyninst: add a more efficient way to resolve PCs

andreimatei:
This patch adds a facility for resolving multiple increasing program counters within a single function more efficiently than repeated calls to GoSymbolTable.LocatePC(), by amortizing the lookups for the function information. When exploring DWARF for SymDB we're generally in the context of a function and we want to resolve many program counters within it before moving on to the next function.

The new PCIterator basically exports the internal `stepper`, with a more tailored interface.


andreimatei:
This PR was reviewed in https://github.com/DataDog/datadog-agent/pull/38679 ,  but I had to recreate because the original PR had my forked repo as the source branch.

piob-io (approved):

PR #38895, opened 2025-07-17
[DEBUG-3883] dyninst/object: handle binaries with mixed dwarf

ajwerner:
### What does this PR do?

We ran into an issue in the wild where a binary had some compile units at dwarf 5 from a C compiler and the go was using dwarf 4. Our logic to decide which section to use to read locations lists was too brittle -- it just preferred the loclists section if it existed. This fixes that bug and adds testign that would have caught it.

Added testing.

None.

### Additional Notes

This bug makes Go DI unusable against binaries that link C libraries that have dwarf. We see this with librdkafka unforunately.

### DIFF
--- a/pkg/dyninst/dwarf/loclist/reader.go
+++ b/pkg/dyninst/dwarf/loclist/reader.go
@@ -19,24 +19,27 @@ import (
 
 // Reader reads DWARF loclists.
 type Reader struct {
-	data              []byte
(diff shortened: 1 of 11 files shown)


grantseltzer (approved):

piob-io (approved):
Thank you for debugging and fixing!

PR #38897, opened 2025-07-17
upgrade to go-tuf version with the 5.3.11 TUF spec fix and fix transa…

moth-a:
### What does this PR do?

Fixes an issue when handling a threshold number of TUF meta keys being revoked.

The intended behaviors is that when the threshold number of signing keys for a given TUF meta are revoked (rotated), the TUF client will delete the corresponding locally stored meta, so that said meta's last-known-good version is not used to validate against the version of the same meta in the incoming remote Update we're processing. This enforces 5.3.11 in the TUF spec.

There were two issues causing this behavior to not work as intended:
1. A `go-tuf` bug regarding how the last-known-good version of a meta isn't updated, specifically when the meta is deleted from localStorage. This is fixed in [this `go-tuf` PR](https://github.com/DataDog/go-tuf/pull/70), which is being pulled into this `datadog-agent` PR through the [pkg/config/remote/go.mod submodule](https://github.com/DataDog/datadog-agent/pull/38897/files). 
2. Our `transactional_store` implementation of the `localStore` still returns a given meta even after it has been deleted during a given transaction. Specifically, the `delete()` function deletes the meta from _memory but not disk_, and the `getAll()` function was [reading-through memory](https://github.com/DataDog/datadog-agent/pull/38897/files#diff-624bc118439a3d844b6705a5abdd0ae74c38019a91eca254c665114668fdf634R68-R76) and loading the meta from disk.

Thorough unit tests regarding this sequence of interactions have been added both in datadog-agent in this PR as well as in the go-tuf repo to ensure both components are exhibiting the correct behavior.

### DIFF
--- a/pkg/config/remote/go.mod
+++ b/pkg/config/remote/go.mod
@@ -13,6 +13,7 @@ require (
 	github.com/DataDog/datadog-agent/pkg/util/log v0.64.1
 	github.com/DataDog/datadog-agent/pkg/util/startstop v0.0.0-00010101000000-000000000000
 	github.com/DataDog/datadog-agent/pkg/util/uuid v0.59.0
+	github.com/DataDog/go-tuf v1.1.1-0.5.2
 	github.com/Masterminds/semver v1.5.0
 	github.com/benbjohnson/clock v1.3.5
 	github.com/coreos/go-semver v0.3.1
@@ -52,7 +53,6 @@ require (
 	github.com/DataDog/datadog-agent/pkg/util/system/socket v0.61.0 // indirect
 	github.com/DataDog/datadog-agent/pkg/util/winutil v0.61.0 // indirect
 	github.com/DataDog/datadog-agent/pkg/version v0.64.1
-	github.com/DataDog/go-tuf v1.1.0-0.5.2
 	github.com/DataDog/viper v1.14.1-0.20250612143030-1b15c8822ed4 // indirect
 	github.com/Microsoft/go-winio v0.6.2 // indirect
 	github.com/cihub/seelog v0.0.0-20170130134532-f561c5e57575 // indirect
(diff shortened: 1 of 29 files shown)


pkg/config/remote/go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/DataDog/datadog-agent/pkg/util/http v0.61.0
 	github.com/DataDog/datadog-agent/pkg/util/log v0.64.1
 	github.com/DataDog/datadog-agent/pkg/util/uuid v0.59.0
+	github.com/DataDog/go-tuf v1.1.0-0.5.2.0.20250717041116-d6a89bbc1095 // TODO use properly tagged version once PR is merged.
```
vsavlidis-ddog: What do we mean by the TODO here? Are we pulling in the new version of `go-tuf` and then swapping over to using it once we are fully deployed (i.e. a two-phase deployment?)

moth-a: This TODO will be resolved before I open+merge the PR. Right now, the dependency is "pinned" to my branch in `go-tuf` (PR: https://github.com/DataDog/go-tuf/pull/70).. specifically commit `d6a89bbc1095`.

This let's me temporarily use the bugfixed version of go-tuf to write tests and confirm the go-tuf fix works for our usage in the agent.

After getting more reviews and confidence in the go-tuf change, I will merge that PR first, and update this PR to properly use a tagged version of the `master` branch of `go-tuf`

pkg/config/remote/go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/DataDog/datadog-agent/pkg/util/http v0.61.0
 	github.com/DataDog/datadog-agent/pkg/util/log v0.64.1
 	github.com/DataDog/datadog-agent/pkg/util/uuid v0.59.0
+	github.com/DataDog/go-tuf v1.1.1-0.5.2
```
moth-a, line 16: This pulls in this change: https://github.com/DataDog/go-tuf/pull/70

mellon85 (approved):

PR #39009, opened 2025-07-17
Fixes unreported eviction metric in kubelet check

frank-spano:
### What does this PR do?
Adds check to only flush first value on first run of the kubelet core check.

### Motivation
Business request

First I deployed a default datadog agent, and had a pod evicted on a cluster - verified that the datadog agent did NOT report the eviction. Then I repeated the steps with my fix built and verified that the agent did indeed report the eviction. Here are the steps

#### Step 1: Deploy cluster configured with resource limits.
```
k3d cluster create \
  --agents 2 \
  --agents-memory 3g \
  --k3s-arg "--disable=traefik@server:0" \
  --k3s-arg "--kubelet-arg=eviction-hard=memory.available<1Gi@agent:0" \
  --k3s-arg "--kubelet-arg=eviction-hard=memory.available<1Gi@agent:1" \
  --k3s-node-label "role=worker@agent:0,1" 
  ```
  Set the correct context
```kubectl config use-context k3d-k3s-default```

#### Step 2: Install datadog operator
`helm install datadog-operator datadog/datadog-operator`

#### Step 3: Update agent CRD to include kubelet check, and limit agent resources so it's not evicted
Config map
```
#datadog-agent.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: kubelet-config
  namespace: default
data:
  kubelet.yaml: |
    init_config:
    instances:
      - kubelet_url: "https://%%host%%:10250"
        use_histogram: true
        telemetry: true
---
```

Include in node agent conf and set resource limits
```
spec:
  override:
    nodeAgent:
      containers: # same level as extraConfd
        agent:
          resources:
            requests:
              memory: "256Mi"
              cpu: "200m"
            limits:
              memory: "512Mi"
              cpu: "500m"
        trace-agent:
          resources:
            requests:
              memory: "128Mi"
              cpu: "100m"
            limits:
              memory: "256Mi"
              cpu: "200m"
      extraConfd: # same level as containers
        configMap:
          name: kubelet-config
          items:
            - key: kubelet.yaml
              path: kubelet.d/kubelet.yaml
```

#### Step 4: Deploy agent
We should see datadog agents running, as well as the operator
```
❯ kubectl apply -f datadog-agent.yaml
❯ kubectl get pods
NAME                                    READY   STATUS    RESTARTS   AGE
datadog-agent-2c2gc                     2/2     Running   0          4h56m
datadog-agent-h75s5                     2/2     Running   0          4h56m
datadog-agent-z2hff                     2/2     Running   0          4h56m
datadog-cluster-agent-987b67bb9-bf854   1/1     Running   0          4h56m
datadog-operator-6c9b79787f-jnzk8       1/1     Running   0          4h57m
```

#### Step 5: Verify there are no evictions
Given that we want to test that we report the first eviction, we should verify that our cluster is clean and has no evictions
```
❯ kubectl get --raw "/api/v1/nodes/k3d-k3s-default-agent-0/proxy/metrics" 2>/dev/null | grep "kubelet_evictions{" ||
       echo "No evictions metric found"
No evictions metric found
```

#### Step 6: Deploy a memory hog pod to cause an eviction.
We're going to deploy a pod that consumes enough resources that it will be evicted, while monitoring for the metric.
First create memory-hog.yaml
```
apiVersion: v1
kind: Pod
metadata:
  name: memory-hog-single-test
spec:
  nodeSelector:
    kubernetes.io/hostname: k3d-k3s-default-agent-0
  containers:
  - name: memory-hog
    image: polinux/stress
    resources:
      requests:
        memory: "100Mi"
      limits:
        memory: "2.5Gi"
    command: ["stress"]
    args: ["--vm", "1", "--vm-bytes", "2200M", "--vm-hang", "1"]
```
Now deploy the pod
```
kubectl apply -f memory-hog.yaml
```

Simultaneously in another terminal window we'll monitor for evictions
```
❯ kubectl get events --field-selector reason=Evicted --watch
LAST SEEN   TYPE      REASON    OBJECT                       MESSAGE
0s          Warning   Evicted   pod/memory-hog-single-test   The node was low on resource: memory. Threshold quantity: 1Gi, available: 438753Ki. Container memory-hog was using 924896Ki, request is 100Mi, has larger consumption of memory.
```

Once we see an eviction we remove the pod
```
❯ kubectl delete pod memory-hog-single-test
```

We can verify that a pod was evicted and the metric was incremented 
```
❯ kubectl get --raw "/api/v1/nodes/k3d-k3s-default-agent-0/proxy/metrics" 2>/dev/null | grep "kubelet_evictions{" ||
       echo "No evictions metric found"
kubelet_evictions{eviction_signal="memory.available"} 1
```

In this example, we can verify on the datadog dashboard that NO evictions were reported by the agent.

#### Step 7: Repeat with fix implemented
After repeating all the above steps, except with the fix integrated with the agent, we can see an eviction reported the next time the kubelet check runs. 

Screenshot 2025-07-17 at 8 26 36 AM

### CODE REVIEW

**Review by michaelcretzman — APPROVED** [review]

File: releasenotes/notes/fix-kubelet-check-flush-first-value-f90aadc1f37c7962.yaml [hunk 1]
```diff
@@ -0,0 +1,12 @@
...
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+fixes:
+  - |
+    Adds check to only flush first value on first run of the
+    kubelet check.
```
sblumenthal: So, this fix is more narrow in focus than this though, it is only a subset of metrics that are affected by it, and it is fixing an issue where the first datapoint would not be flushed on newly detected metrics.

Maybe something like
```suggestion
    Fixes a bug in the kubelet check where the first seen eviction would not result in the `kubernetes.kubelet.evictions` metric being emitted.
```

Or something else more narrowly focused?

pkg/collector/corechecks/containers/kubelet/provider/kubelet/provider.go
```diff
@@ -146,13 +149,23 @@ func NewProvider(filter *containers.Filter, config *common.KubeletConfig, store
...
+}
+
 func (p *Provider) sendAlwaysCounter(metricFam *prom.MetricFamily, sender sender.Sender) {
 	metricName := metricFam.Name
 	nameWithNamespace := common.KubeletMetricsPrefix + counterMetrics[metricName]
 
 	for _, metric := range metricFam.Samples {
 		tags := p.MetricTags(metric)
-		sender.MonotonicCount(nameWithNamespace, float64(metric.Value), "", tags)
+		sender.MonotonicCountWithFlushFirstValue(nameWithNamespace, float64(metric.Value), "", tags, !p.firstRun)
```
vboulineau, line 169: Is this an accurate fix? I think that on every restart it's going to produce value `x`, `x` being the total number of evictions that happened _so far_, so a spike on every restart? Is that not the case?

I am not thinking that it's not necessarily related to first run or not. I think the `kubelet_evictions` value will only appear in Prom endpoint output when at least once happens, meaning that we never submit the 0 value, though `p.firstRun` will be false while it should be true.

frank-spano: This now matches the python implementation here: https://github.com/DataDog/integrations-core/blob/25c7fbea4e3bd01a1112553602c478bd04a4800a/kubelet/datadog_checks/kubelet/kubelet.py#L718

The flow should be as follows:

 1. Agent starts - kubelet has 5 evictions total
    - First run: self.first_run=True, so flush_first_value=False
    - Agent stores value 5 but doesn't emit it (waits for next value to calculate diff)
  2. Agent second check - kubelet has 7 evictions total
    - self.first_run=False, so flush_first_value=True
    - Agent calculates diff: 7-5=2 evictions, emits 2
  3. Agent crashes and restarts - kubelet still has 7 evictions total
    - First run after restart: self.first_run=True, so flush_first_value=False
    - Agent stores value 7 but doesn't emit it (preventing false spike)
  4. Agent second check after restart - kubelet has 9 evictions total
    - self.first_run=False, so flush_first_value=True
    - Agent calculates diff: 9-7=2 evictions, emits 2

vboulineau: Right my bad, it works on restart given that no change happened in between. However I'm still wondering if it solves the first case or not.

If
```
# HELP kubelet_evictions [ALPHA] Cumulative number of pod evictions by eviction signal
# TYPE kubelet_evictions counter
kubelet_evictions{eviction_signal="<whatever>"} 0
```

Is not present, I think we would never call this and the first 0 -> 1 is lost?

frank-spano: @vboulineau Sorry what's the first case? If you provide me the scenario I can replicate it.

vboulineau: The first time there's an eviction on the node, Agent already running.

frank-spano: That's the scenario I described testing in the description, I had a running agent and it reported the first (and only) eviction.

vboulineau (approved):

PR #39018, opened 2025-07-17
[orch] Add config for max crs to collect

JLineaweaver:
### What does this PR do?
We want to allow users to set the maximum number of CRs they want to collect, but we still want to have a hard cap that users cannot exceed. So we've set the old maximum as the default, but provided a way for that to be increased"

### Motivation

We want to maintain our max custom resource value by default, but we want to add some wiggle room for anyone hitting that limit.


releasenotes-dca/notes/CRMaxQuota-38f82cbcbea35793.yaml
```diff
@@ -0,0 +1,10 @@
+# Each section from every release note are combined when the
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+enhancements:
+  - |
+    Created a new configuration option `orchestrator_explorer.custom_resource.max_count` that will allow customers to increase the number of custom resource they can collect in the Cluster Agent.
```
michaelcretzman:
```suggestion
    Created a new configuration option `orchestrator_explorer.custom_resource.max_count` that will allow customers to increase the number of custom resources they can collect in the Cluster Agent.
```

michaelcretzman (approved):
Approved with one suggestion

rahulkaukuntla (approved):

stzou (approved):

PR #39019, opened 2025-07-17
[CNM] add plaintext protocol in classifier

akarpz:
### What does this PR do?

Explicitly adds an encryption layer protocol tag for plaintext traffic if TLS classification fails. This will allow the CNM backend to be certain that traffic is not TLS.

### DIFF
--- a/pkg/network/ebpf/c/protocols/classification/protocol-classification.h
+++ b/pkg/network/ebpf/c/protocols/classification/protocol-classification.h
@@ -193,6 +193,13 @@ __maybe_unused static __always_inline void protocol_classifier_entrypoint(struct
             }
         }
         return;
+    } else {
+        // Add a plaintext encryption layer tag if TLS classification returns false
+        protocol_stack = get_or_create_protocol_stack(&classification_ctx->tuple);
+        if (!protocol_stack) {
+            return;
+        }
+        update_protocol_information(classification_ctx, protocol_stack, PROTOCOL_PLAINTEXT);
     }
 
     // If we have already classified the encryption layer, we can skip the rest of the classification
--- a/pkg/network/tracer/tracer_linux_test.go
+++ b/pkg/network/tracer/tracer_linux_test.go
@@ -3107,3 +3107,182 @@ func (s *TracerSuite) TestTCPSynRst() {
 	assert.Equal(t, uint32(1), conn.TCPFailures[uint16(unix.ECONNREFUSED)])
 	assert.Equal(t, uint16(1), conn.Monotonic.TCPClosed)
 }
+
+func (s *TracerSuite) TestPlaintextClassification() {
+	t := s.T()
+	cfg := testConfig()
+
+	if !kprobe.ClassificationSupported(cfg) {
+		t.Skip("protocol classification not supported")
+	}
+
+	tr := setupTracer(t, cfg)
+
+	// Test plaintext HTTP traffic
+	t.Run("PlaintextHTTP", func(t *testing.T) {
+		if ebpftest.GetBuildMode() == ebpftest.Fentry {
+			t.Skip("protocol classification not supported for fentry tracer")
+		}
+		t.Cleanup(func() {
+			tr.RemoveClient(clientID)
+			_ = tr.Pause()
+		})
+
+		// Set up a simple HTTP server
+		srv := usmtestutil.NewTCPServer("localhost:0", func(conn net.Conn) {
+			defer conn.Close()
+			tracertestutil.SetTestDeadline(conn)
+			buf := make([]byte, 1024)
+			_, err := conn.Read(buf)
+			if err != nil {
+				if err != io.EOF {
+					t.Logf("Failed to read data: %v\n", err)
+				}
+				return
+			}
+			// Send a simple HTTP response
+			response := "HTTP/1.1 200 OK\r\nContent-Length: 5\r\n\r\nHello"
+			_, err = conn.Write([]byte(response))
+			if err != nil {
+				t.Logf("Failed to write response: %v\n", err)
+			}
+		}, false)
+		done := make(chan struct{})
+		require.NoError(t, srv.Run(done))
+		t.Cleanup(func() { close(done) })
+
+		waitForTracer(t, tr, srv.Address())
+
+		tr.RemoveClient(clientID)
+		require.NoError(t, tr.RegisterClient(clientID))
+		require.NoError(t, tr.Resume(), "enable probes - before post tracer")
+
+		// Get the port assigned to the server
+		addr := srv.Address()
+		_, portStr, err := net.SplitHostPort(addr)
+		require.NoError(t, err)
+		portInt, err := strconv.Atoi(portStr)
+		require.NoError(t, err)
+		port := uint16(portInt)
+
+		// Make a plaintext HTTP request
+		conn, err := tracertestutil.DialTCP("tcp", addr)
+		require.NoError(t, err)
+		defer conn.Close()
+		tracertestutil.SetTestDeadline(conn)
+
+		// Send HTTP request
+		request := "GET / HTTP/1.1\r\nHost: localhost\r\n\r\n"
+		_, err = conn.Write([]byte(request))
+		require.NoError(t, err)
+
+		// Read response
+		buf := make([]byte, 1024)
+		_, err = conn.Read(buf)
+		require.NoError(t, err)
+
+		require.NoError(t, tr.Pause(), "disable probes - after post tracer")
+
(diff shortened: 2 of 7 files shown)


pkg/network/ebpf/c/protocols/classification/protocol-classification.h
```diff
@@ -193,6 +193,13 @@ __maybe_unused static __always_inline void protocol_classifier_entrypoint(struct
...
         }
         return;
+    } else {
+        // Add a plaintext encryption layer tag if TLS classification returns false
+        protocol_stack = get_or_create_protocol_stack(&classification_ctx->tuple);
+        if (!protocol_stack) {
+            return;
+        }
+        update_protocol_information(classification_ctx, protocol_stack, PROTOCOL_PLAINTEXT);
     }
```
guyarb, line 203: There's an indentation problem, and most likely the code is invalid (else after return)

pkg/network/tracer/tracer_linux_test.go
```diff
@@ -3107,3 +3107,182 @@ func (s *TracerSuite) TestTCPSynRst() {
...
+		require.EventuallyWithT(t, func(ct *assert.CollectT) {
+			payload, cleanup := getConnections(ct, tr)
+			defer cleanup()
+			var found bool
+			for _, c := range payload.Conns {
+				if c.DPort == port {
+					// Should have plaintext encryption layer
+					assert.True(ct, c.ProtocolStack.Encryption == protocols.Plaintext, "expected plaintext encryption layer, got %v", c.ProtocolStack.Encryption)
+					// Should not have any application layer protocol (since it's not HTTP/etc)
+					assert.Equal(ct, protocols.Unknown, c.ProtocolStack.Application, "expected no application layer protocol, got %v", c.ProtocolStack.Application)
```
pimlu, line 3278: Am I understanding correctly that prior to this PR, if it didn't find any protocol at all, the backend would determine `tls_encrypted:unknown` and that's what this PR fixes by adding plaintext?

pkg/network/ebpf/c/protocols/classification/protocol-classification.h
```diff
@@ -193,6 +193,13 @@ __maybe_unused static __always_inline void protocol_classifier_entrypoint(struct
             }
         }
         return;
+    } else {
+        // Add a plaintext encryption layer tag if TLS classification returns false
+        protocol_stack = get_or_create_protocol_stack(&classification_ctx->tuple);
```
guyarb, line 198: that's a repetition of lines 168-169
Let's re-use the protocol stack if it is in `if-else` clause

pkg/network/ebpf/c/protocols/classification/protocol-classification.h
```diff
@@ -193,6 +193,13 @@ __maybe_unused static __always_inline void protocol_classifier_entrypoint(struct
...
+        if (!protocol_stack) {
+            return;
+        }
+        update_protocol_information(classification_ctx, protocol_stack, PROTOCOL_PLAINTEXT);
```
guyarb, line 202: Can't you set it to plaintext, and just override if that's TLS?

pimlu, line 202: I'm not super familiar with this code, but any chance this would mark pre-existing TLS connections where we missed the handshake as PLAINTEXT instead of tls unknown?

pkg/network/ebpf/c/protocols/classification/protocol-classification.h
```diff
@@ -193,6 +193,13 @@ __maybe_unused static __always_inline void protocol_classifier_entrypoint(struct
             }
         }
         return;
+    } else {
```
guyarb, line 196: also, now it seems that we're going to set PLAINTEXT tag for every connection, even if it does not have supported protocols.
I'm concerned about the impact of the map utilization and it can lead into stuffing the map with "irrelevant" entries and then we won't have enough space for the relevant entries

pkg/network/tracer/tracer_linux_test.go
```diff
@@ -3107,3 +3107,182 @@ func (s *TracerSuite) TestTCPSynRst() {
 	assert.Equal(t, uint32(1), conn.TCPFailures[uint16(unix.ECONNREFUSED)])
 	assert.Equal(t, uint16(1), conn.Monotonic.TCPClosed)
 }
+
+func (s *TracerSuite) TestPlaintextClassification() {
```
guyarb, line 3111: TBH, that's a redundant test
We already have TLSClassification and Plaintext (non TLS) classification tests
you should be just adding to those tests

pkg/network/ebpf/c/protocols/classification/protocol-classification.h
akarpz: this is a good point. The purpose of this PR is to ensure that we somehow mark non-TLS connections as such even if we don't support the protocol. I will have to test this a bit to make sure we aren't drastically increasing the map usage

pimlu (approved):

pkg/network/ebpf/c/protocols/classification/protocol-classification.h
akarpz: we could? I thought it would be clearer to put it after the TLS block returns false

pkg/network/tracer/tracer_linux_test.go
akarpz: yes if you look at the CNM backend code you'll find that if there are no protocol tags, we will always tag `tls_encrypted:unknown`. So for protocols we don't support that are plaintext, we will say unknown even if TLS classification returns false

pkg/network/ebpf/c/protocols/classification/protocol-classification.h
akarpz: the TLS classifier above also inspects TLS app data packets

pkg/network/tracer/tracer_linux_test.go
akarpz: I'll look into deduping these

guyarb (changes requested):
Have you deployed it to a staging cluster? Have you checked the impact on the map utilization and map errors (E2BIG)?

What alternatives do we have?

Why can't we deduce in the backend, the connection is plaintext if it is not TLS?

akarpz:
@guyarb I will deploy to a staging cluster shortly. The issue with trying to deduce in the backend is that if protocol classification is turned off, if the agent is too old, or if the eBPF map is full we won't have protocol tags either, which is the legitimate use case for `tls_encrypted:unknown`

pkg/network/ebpf/c/protocols/classification/protocol-classification.h
guyarb: That's redundant

see the if-clause https://github.com/DataDog/datadog-agent/blob/bf46cda6ffb5b48efe7325eff0f1e922fc82a87d/pkg/network/ebpf/c/protocols/classification/protocol-classification.h#L167-L173

If the connection matches the condition, it will have TLS tag. if it does not it is plaintext.

guyarb: lacking the TLS tag means it is plaintext, adding explicitly plaintext tag does not give any guarantee that you're looking for.

guyarb:
> @guyarb I will deploy to a staging cluster shortly. The issue with trying to deduce in the backend is that if protocol classification is turned off or if the agent is too old, we won't have protocol tags either, which is the legitimate use case for `tls_encrypted:unknown`

The agent version scenario you can solve by getting that information (I'm sure it is available in the backend, and if not, the fix should be adding the agent-version to the payload)

The configuration is also being reported back using fleet automation, and again, you can add the information to the payload (classification enabled/disabled)

akarpz:
@guyarb this is a good point, but what if the map fills up

guyarb:
Also, this change seems to cause undesired side effects, as now https://github.com/DataDog/datadog-agent/blob/bf46cda6ffb5b48efe7325eff0f1e922fc82a87d/pkg/network/ebpf/c/protocols/classification/protocol-classification.h#L156 will return true to `encryption_layer_known` and then we will stop trying classifying the actual protocols (https://github.com/DataDog/datadog-agent/blob/bf46cda6ffb5b48efe7325eff0f1e922fc82a87d/pkg/network/ebpf/c/protocols/classification/protocol-classification.h#L205-L208)


so any TCP handshake will be marked as plaintext, and we will loose actual classification

guyarb:
> @guyarb this is a good point, but what if the map fills up

Then your solution breaks as well :)

PR #39020, opened 2025-07-17
dyninst/symdb: extract symbols from Go binaries for SymDB

andreimatei:
Walk through the DWARF debug info for a Go binary and extract the
symbols data that we'll export to SymDB.
    
The code is packaged as `SymDBBuilder`, a utility to be hooked up to
remote config, and a CLI utility to test the builder on a given binary.

https://datadoghq.atlassian.net/browse/DEBUG-3890


pkg/dyninst/dwarf/dwarfutil/util.go
```diff
@@ -0,0 +1,150 @@
...
+
+	return locations, nil
+}
+
+// FilterIncompleteLocationLists filters out location lists that have missing pieces.
+func FilterIncompleteLocationLists(lists []ir.Location) []ir.Location {
+	var filtered []ir.Location
+	for _, loc := range lists {
+		if len(loc.Pieces) == 0 {
+			continue // Skip empty locations
```
piob-io: Drop empty locations? ("skip" doesn't carry information what happens in this case)

pkg/dyninst/symdb/symdb.go
```diff
@@ -0,0 +1,965 @@
...
+	// compile unit.
+	res.Types = make([]Type, len(b.types.packages[res.Name]))
+	for i, t := range b.types.packages[res.Name] {
+		res.Types[i] = *t
+	}
+	return res, nil
+}
+
+// DW_INL_inlined is the value for AttrInline marking the function as inlined.
+const DW_INL_inlined = 1 //nolint:revive
```
piob-io: If you prefer the uppercase spelling, how about we put all dwarf constants into some dwarf-specific package, so that we don't leak random crap from a more general package? I prefer smart autocompletion to not suggest me dwarf constants when I am using something from symdb package.

pkg/dyninst/symdb/symdb.go
```diff
@@ -0,0 +1,965 @@
...
+			"subprogram %s has invalid file index %d, expected in range [0, %d)",
+			funcQualifiedName, fileIdx, len(b.filesInCurrentCompileUnit),
+		)
+	}
+	fileName := b.filesInCurrentCompileUnit[fileIdx]
+	// There are some funky auto-generated functions that we can't deal with
+	// because we can't parse their names (e.g.
+	// type:.eq.sync/atomic.Pointer[os.dirInfo]). Skip everything
+	// auto-generated.
+	if fileName == "<autogenerated>" {
```
piob-io, line 620: Move this check prior to parseFuncName, so when we will report warnings about failing to parse, we don't report autogenerated ones?

pkg/dyninst/symdb/symdb.go
```diff
@@ -0,0 +1,965 @@
...
+		return Function{}, fmt.Errorf("subprogram without lowpc: %s", funcQualifiedName)
+	}
+	highpc, ok := subprogEntry.Val(dwarf.AttrHighpc).(uint64)
+	if !ok {
+		return Function{}, errors.New("subprogram without highpc")
+	}
+
+	// Do a pass through the subprogram's DWARF to find all the inlined
+	// subroutines. We need to know the PC ranges of inlined subroutines to figure
+	// out the source lines for the function.
```
piob-io: "of this function"? (as opposed to needing these for the inlined functions)

pkg/dyninst/symdb/symdb.go
```diff
@@ -0,0 +1,965 @@
...
+		File:          fileName,
+		Scope: Scope{
+			StartLine: int(firstLine),
+			EndLine:   int(maxLine),
+			Variables: inner.vars,
+			Scopes:    inner.scopes,
+		},
+	}
+
+	if funcName.GenericFunction {
```
piob-io: Why not check this right after name parsing?

piob-io (approved):
Nice!

pkg/dyninst/gosym/pc_iterator.go
```diff
@@ -63,7 +63,7 @@ func (f *PCIterator) Reset() {
 // PC.
 //
 // Any PC inside the function is a valid input, but the PCIterator is more
```
ajwerner: I think this now is spelled wrong... 

https://dictionary.cambridge.org/us/dictionary/english/monotonically

pkg/dyninst/symdb/symdb_test.go
```diff
@@ -0,0 +1,76 @@
...
+		fn, ok := findFunction(pkg, "stringArg")
+		require.Truef(t, ok, "function 'stringArg' not found in package 'main' in %s", binaryPath)
+		require.NotZero(t, fn.StartLine)
+		require.NotZero(t, fn.EndLine)
+		require.Less(t, fn.StartLine, fn.EndLine)
+		v, ok := findVariable(fn.Scope, "s")
+		require.Truef(t, ok, "variable 's' not found in function 'stringArg' in package 'main' in %s", binaryPath)
+		require.True(t, v.FunctionArgument)
+		require.NotZero(t, v.DeclLine)
+		require.NotEmpty(t, v.AvailableLineRanges)
```
ajwerner: nit: use subtests per config

pkg/dyninst/symdb/symdb.go
```diff
@@ -0,0 +1,965 @@
...
+	}
+	res.Name = name
+
+	cuLineReader, err := b.dwarfData.LineReader(entry)
+	if err != nil {
+		return Package{}, fmt.Errorf("could not get file line reader for compile unit %s: %w", name, err)
+	}
+	var files []string
+	if cuLineReader != nil {
+		for i, file := range cuLineReader.Files() {
```
ajwerner, line 493: allocate the slice? as in:

```go
readerFiles := cuLineReader.Files()
files = make([]string, 0, len(readerFiles))
```

pkg/dyninst/symdb/symdb.go
```diff
@@ -0,0 +1,965 @@
...
+				// used as a sentinel by file references to indicate that the
+				// file is not known.
+				if i != 0 {
+					return Package{}, fmt.Errorf(
+						"compile unit %s has invalid nil file entry at index %d", name, i)
+				}
+				files = append(files, "")
+				continue
+			}
+			files = append(files, strings.Clone(file.Name))
```
ajwerner: I don't think this strings.Clone is needed. It looks to me like the dwarf library is already allocating fresh strings for every file name...

pkg/dyninst/symdb/symdb.go
```diff
@@ -0,0 +1,965 @@
...
+}
+
+// DW_INL_inlined is the value for AttrInline marking the function as inlined.
+const DW_INL_inlined = 1 //nolint:revive
+
+// exploreSubprogram processes a subprogram entry, corresponding to a Go
+// function (entry's tag is TagSubprogram). It accumulates the function's
+// lexical blocks and variables. If the function is free-standing (not a
+// method), returns the accumulated info. If the function is a method (i.e. has
+// a receiver), it is appended to the types methods in the types collection.
```
ajwerner: say whether it's also returned (it's not, but that surprised me)

pkg/dyninst/dwarf/dwarfutil/util.go
andreimatei: went with "ignore" :P

pkg/dyninst/symdb/symdb.go
```diff
@@ -0,0 +1,965 @@
...
+		}
+	}
+
+	funcName, err := parseFuncName(funcQualifiedName)
+	if err != nil {
+		// Note that we do not return an error here; we simply ignore the
+		// function so that we don't choke on weird function names that we
+		// haven't anticipated.
+		// TODO: log the error, but don't spam the logs.
+		return Function{}, nil
```
ajwerner: do we not want to skip children here?

pkg/dyninst/symdb/symdb.go
```diff
@@ -0,0 +1,965 @@
...
+	}
+	if funcName.Empty() {
+		return Function{}, nil
```
ajwerner: do we not want to skip children here?

pkg/dyninst/symdb/symdb.go
andreimatei: ok, I created a `constants.go` in our `dwarf` package

pkg/dyninst/symdb/symdb.go
```diff
@@ -0,0 +1,965 @@
...
+
+	// Do a pass through the subprogram's DWARF to find all the inlined
+	// subroutines. We need to know the PC ranges of inlined subroutines to figure
+	// out the source lines for the function.
+	inlinedPcRanges, err := dwarfutil.ExploreInlinedPcRangesInSubprogram(reader, b.dwarfData)
+	if err != nil {
+		return Function{}, fmt.Errorf("error exploring inlined subroutines in function %s: %w", funcQualifiedName, err)
+	}
+	// Reset the reader to the start of the subprogram.
+	reader.Seek(subprogEntry.Offset)
```
ajwerner, line 660: it's annoying that the reader doesn't expose its current offset, right now you've got exploreCode silently ignoring the subprogram entry but re-parsing it

pkg/dyninst/symdb/symdb.go
andreimatei: done

pkg/dyninst/symdb/symdb.go
```diff
@@ -0,0 +1,965 @@
...
+	}
+	highpc, ok := subprogEntry.Val(dwarf.AttrHighpc).(uint64)
+	if !ok {
+		return Function{}, errors.New("subprogram without highpc")
+	}
+
+	// Do a pass through the subprogram's DWARF to find all the inlined
+	// subroutines. We need to know the PC ranges of inlined subroutines to figure
+	// out the source lines for the function.
+	inlinedPcRanges, err := dwarfutil.ExploreInlinedPcRangesInSubprogram(reader, b.dwarfData)
```
ajwerner, line 655: aside: it's obnoxious how expensive parsing these DIEs is in this dwarf library. It's clearly very nice to separate the iteration for finding these inlined ranges from the below exploration of code, but it's really inefficient... it's like 4 allocations and several us per DIE.

pkg/dyninst/symdb/symdb.go
```diff
@@ -0,0 +1,965 @@
...
+//
+// pcIt is a PC iterator for the function containing this code.
+func (b *SymDBBuilder) exploreCode(reader *dwarf.Reader, pcIt *gosym.PCIterator) (exploreCodeResult, error) {
+	var res exploreCodeResult
+	for child, err := reader.Next(); child != nil; child, err = reader.Next() {
+		if err != nil {
+			return exploreCodeResult{}, err
+		}
+		if dwarfutil.IsEntryNull(child) {
+			break // End of children for this subprogram.
```
ajwerner: nit: this isn't just exploring subprograms, it can also be exploring lexical blocks etc 
nit^2: sentences for full comments, not for inline comments

pkg/dyninst/symdb/symdb.go
```diff
@@ -0,0 +1,965 @@
...
+	}
+	d.pcRanges = ranges
+	d.pcRangesResolved = true
+	return d.pcRanges, nil
+}
+
+// resolveLines computes the start and end lines of the lexical block by
+// combining PC ranges from DWARF with pclinetab data. Returns [0,0] if the
+// block has no address ranges, and thus no lines.
+func (d *dwarfBlock) resolveLines(dwarfData *dwarf.Data, pcIt *gosym.PCIterator) ([2]int, error) {
```
ajwerner: nit: define a type called `type lineRange [2]int` just for readability.

pkg/dyninst/symdb/func_name.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+// parseFuncName parses a Go qualified function name. For a qualifiedName name
+// like:
+// github.com/cockroachdb/cockroach/pkg/kv/kvserver.(*raftSchedulerShard).worker
+// the package is: github.com/cockroachdb/cockroach/pkg/kv/kvserver
+// the type is: raftSchedulerShard (note that it doesn't include the '*' signifying a pointer receiver).
+// the name is: worker
+//
+// Returns (zero value, nil) if the function should be ignored.
```
ajwerner, line 87: It's sometimes surprising that it returns things like `funcName{GenericFunction: true}, nil`.

ajwerner, line 87: Maybe instead of "should be ignored" it should say something like aren't supported and then the caller can determine that it wants to ignore more things and also definitely will be ignored because this implementation isn't really giving it much choice.

ajwerner (approved):

pkg/dyninst/gosym/pc_iterator.go
andreimatei: I've got no idea what triggered this change... reverted

pkg/dyninst/symdb/symdb_test.go
andreimatei: done

pkg/dyninst/symdb/symdb.go
andreimatei: done

pkg/dyninst/symdb/symdb.go
andreimatei: done

pkg/dyninst/symdb/symdb.go
andreimatei: done

pkg/dyninst/symdb/symdb.go
andreimatei: done

pkg/dyninst/symdb/symdb.go
andreimatei: yes, good catch. It doesn't matter all that much cause the caller is not very sensitive, but yeah.

pkg/dyninst/symdb/symdb.go
andreimatei: done

pkg/dyninst/symdb/symdb.go
andreimatei: Indeed.
I've now moved the parsing of the subprogram entry here, to make it explicit.

pkg/dyninst/symdb/symdb.go
andreimatei: yeah... added a TODO

pkg/dyninst/symdb/symdb.go
andreimatei: done

pkg/dyninst/symdb/symdb.go
andreimatei: done

pkg/dyninst/symdb/symdb.go
andreimatei: done

pkg/dyninst/symdb/func_name.go
andreimatei: well from the perspective of this function, the cases where it returns the zero value should be ignored. I'm not sure how to make the comment better.

andreimatei: yeah... But I like that for the tests. Maybe I should return a "reason" for ignoring the function for all cases. I think I'll do that in a followup.

PR #39022, opened 2025-07-17
[clusteragent/autoscaling]  Make it possible to apply vertical recommendations to init sidecar containers as well as to regular containers

ADKosm:
### What does this PR do?

This PR extends the Datadog autoscaler's vertical scaling recommendations to apply to sidecar init containers within a Pod.

### Motivation

Kubernetes implements sidecar containers as a special case of init containers. This feature is active by default since Kubernetes v1.29. 
Since these sidecar init containers run alongside the main application containers, they are also suitable targets for vertical autoscaling.

The primary use cases are covered by unit tests. I am also in the process of manually validating this agent version on a live cluster.

This new feature is backward-compatible - it only introduces additional logic after the main patch loop. [From the official k8s documentation](https://github.com/kubernetes-client/go/blob/92040c8d5731095abc7378b48b96f3cec2b4090b/kubernetes/docs/V1PodSpec.md?plain=1#L19): `The name for an init container or normal container must be unique among all containers.`. This uniqueness ensures that there are no conflicts when dispatching recommendations to either regular or init containers based on their names.

UPD: I have tested manually the behaviour on test cluster. You can find it [here it the cluster](https://dddev.datadoghq.com/orchestration/scaling/workload?inspect_group=kube_deployment%3Astress-test-deployment%2Ckube_namespace%3Aalexey-test%2Ckube_cluster_name%3Aalexey-kosmachev-aws-eks&workload_scaling_parent_page=%2Forchestration%2Fscaling%2Fworkload%3Fquery%3Dkube_cluster_name%253Aalexey-kosmachev-aws-eks%26ws-cols%3Ddeployment_name%252Ccluster%252Cdeployment_pod_count%252Ccpu_waste_v2%252Cmemory_waste_v2%252Ccost%252Cidle_cost%252Cdeployment_autoscaling%252Cnamespace) is not autoremoved yet.


ADKosm:

jbartosik (commented):
Mosty looks good but test fails:

```
Test failures:
- github.com/DataDog/datadog-agent/pkg/fleet/installer/packages/embedded/templates TestGenerationIsUpToDate
```

also this PR needs qa label

ADKosm:
@jbartosik well, in order to set qa label I have to do the QA :) I [asked in backroom](https://dd.slack.com/archives/C08FUDH30EA/p1752840403066209) about how to do some qa tests. If you have some information on that - please share 🙏

pkg/clusteragent/autoscaling/workload/pod_patcher.go
```diff
@@ -106,31 +106,53 @@ func (pa podPatcher) ApplyRecommendations(pod *corev1.Pod) (bool, error) {
...
+			// sidecar container by definition is an init container with `restartPolicy: Always`
+			if cont.RestartPolicy == nil || *cont.RestartPolicy != corev1.ContainerRestartPolicyAlways {
+				continue
 			}
-			for resource, request := range reco.Requests {
-				if request != cont.Resources.Requests[resource] {
-					cont.Resources.Requests[resource] = request
-					patched = true
-				}
+			if cont.Name != reco.Name {
```
clamoriniere: question about possible edge case: an init-container and a container can have the same name. How do we handle this edge case with the recommandations that we receive?

ADKosm: I mentioned this in description but I think it is a good idea to left comment here as well 😅 
[From the official k8s documentation](https://github.com/kubernetes-client/go/blob/92040c8d5731095abc7378b48b96f3cec2b4090b/kubernetes/docs/V1PodSpec.md?plain=1#L19): `The name for an init container or normal container must be unique among all containers.`

So fortunately kubernetes just doesn't allow this situation to happen and we don't need to do anything in this case

urseberry (approved):

ADKosm:

pkg/clusteragent/autoscaling/workload/pod_patcher.go
```diff
@@ -100,37 +100,61 @@ func (pa podPatcher) ApplyRecommendations(pod *corev1.Pod) (bool, error) {
 	}
 
 	// Even if annotation matches, we still verify the resources are correct, in case the POD was modified.
+	// K8s guarantees that the name for an init container or normal container are unique among all containers.
+	// It means that dispatching recommendations just by container names is sufficient
 	for _, reco := range autoscaler.ScalingValues().Vertical.ContainerResources {
 		for i := range pod.Spec.Containers {
 			cont := &pod.Spec.Containers[i]
 			if cont.Name != reco.Name {
```
vboulineau: nit: the `continue` was useful to avoid indenting the large block. As the code moved to a dedicated function, it's easier to do positive `if` instead

pkg/clusteragent/autoscaling/workload/pod_patcher.go
```diff
@@ -100,37 +100,61 @@ func (pa podPatcher) ApplyRecommendations(pod *corev1.Pod) (bool, error) {
...
 			}
+			patched = pa.patchContainerByRecommendation(reco, cont) || patched
 			break
 		}
 	}
 
 	return patched, nil
 }
 
+func (pa podPatcher) patchContainerByRecommendation(reco datadoghqcommon.DatadogPodAutoscalerContainerResources, cont *corev1.Container) (patched bool) {
```
vboulineau: This function does not need to be a member function, it does not use any parameter.

vboulineau:
```suggestion
func (pa podPatcher) patchContainerResources(reco datadoghqcommon.DatadogPodAutoscalerContainerResources, cont *corev1.Container) (patched bool) {
```

pkg/clusteragent/autoscaling/workload/pod_patcher.go
ADKosm: Done

pkg/clusteragent/autoscaling/workload/pod_patcher.go
```diff
@@ -100,31 +100,27 @@ func (pa podPatcher) ApplyRecommendations(pod *corev1.Pod) (bool, error) {
...
-					cont.Resources.Limits[resource] = limit
-					patched = true
-				}
-			}
-			for resource, request := range reco.Requests {
-				if request != cont.Resources.Requests[resource] {
-					cont.Resources.Requests[resource] = request
-					patched = true
-				}
+		}
```
jbartosik: The function is getting long, what do you think of extracting this as a helper?

pkg/clusteragent/autoscaling/workload/pod_patcher_test.go
```diff
@@ -298,6 +344,334 @@ func TestPatcherApplyRecommendations(t *testing.T) {
...
+						Resources: corev1.ResourceRequirements{
+							Limits:   corev1.ResourceList{"cpu": resource.MustParse("200m")},
+							Requests: corev1.ResourceList{"memory": resource.MustParse("128Mi")},
+						},
+					}},
+				},
+			},
+		},
+		{
+			name: "update init sidecar resources when there are none",
```
jbartosik, line 418: I think it would be nice to add tests for patchContainerResources

jbartosik (commented):
Looks good, a couple suggestions

pkg/clusteragent/autoscaling/workload/pod_patcher.go
ADKosm: good idea, done

pkg/clusteragent/autoscaling/workload/pod_patcher_test.go
ADKosm: yeap, and for new patchPod as well

pkg/clusteragent/autoscaling/workload/pod_patcher.go
```diff
@@ -208,3 +186,49 @@ func (pa podPatcher) observedPodCallback(ctx context.Context, pod *workloadmeta.
 	}
 	log.Debugf("Event sent and POD %s/%s patched with event annotation", pod.Namespace, pod.Name)
 }
+
+func patchContainerResources(reco datadoghqcommon.DatadogPodAutoscalerContainerResources, cont *corev1.Container) (patched bool) {
```
vboulineau, line 212: nit: in call order, `patchContainerResources` is deeper than `patchPod`, ideally order in the file follows that

vboulineau (approved):

jennchenn (approved):

PR #39027, opened 2025-07-17
[NDINT-290] Versa Tunnel Metrics

ken-schneider:
### What does this PR do?
Adds tunnel metrics to the Versa integration via an analytics call.

### Motivation
Continue to build out the Versa integration

Tested locally against our own Versa cluster, but since we don't have a tunnel in our environment, the response is empty. The structure is shared with the other analytics calls we have so we should be fine once we have data for these calls.


ken-schneider:
Since we don't have any tunnels defined in our environment today, I've put this in as an RC required PR so we can test with preview customers as well as try to get something configured in our own environment.

pkg/collector/corechecks/network-devices/versa/client/client.go
```diff
@@ -612,6 +612,73 @@ func (client *Client) GetLinkUsageMetrics(tenant string) ([]LinkUsageMetrics, er
 	return metrics, nil
 }
 
+// parseTunnelMetrics parses the raw AaData response into TunnelMetrics structs
+func parseTunnelMetrics(data [][]interface{}) ([]TunnelMetrics, error) {
+	var rows []TunnelMetrics
+	for _, row := range data {
+		m := TunnelMetrics{}
+		// Based on the new structure, we expect 6 columns
```
jmw51798: 7 columns?

pkg/collector/corechecks/network-devices/versa/client/client.go
```diff
@@ -612,6 +612,73 @@ func (client *Client) GetLinkUsageMetrics(tenant string) ([]LinkUsageMetrics, er
...
+			return nil, fmt.Errorf("expected string for LocalIP")
+		}
+		if m.RemoteIP, ok = row[3].(string); !ok {
+			return nil, fmt.Errorf("expected string for RemoteIP")
+		}
+		if m.VpnProfName, ok = row[4].(string); !ok {
+			return nil, fmt.Errorf("expected string for VpnProfName")
+		}
+
+		// Handle float metrics from indices 4-5
```
jmw51798: 5-6?

pkg/collector/corechecks/network-devices/versa/client/client.go
```diff
@@ -612,6 +612,73 @@ func (client *Client) GetLinkUsageMetrics(tenant string) ([]LinkUsageMetrics, er
...
+		if val, ok := row[5].(float64); ok {
+			m.VolumeRx = val
+		} else {
+			return nil, fmt.Errorf("expected float64 for VolumeRx at index 4")
```
jmw51798: 5?

pkg/collector/corechecks/network-devices/versa/client/client.go
```diff
@@ -612,6 +612,73 @@ func (client *Client) GetLinkUsageMetrics(tenant string) ([]LinkUsageMetrics, er
...
+		}
+		if val, ok := row[6].(float64); ok {
+			m.VolumeTx = val
+		} else {
+			return nil, fmt.Errorf("expected float64 for VolumeTx at index 5")
```
jmw51798: 6?

jmw51798 (approved):
Mostly looks good, left a few minor comments.

pkg/collector/corechecks/network-devices/versa/client/client.go
zoedt: not sure it's worth it at this point / forgot if it's even possible to pass in the "fields" to check from a struct but wondering if we can generalize the recurring logic to check the expected columns + the type assertions for the fields and then the float metrics afterwards (since it follows this pattern)

the type assertions i think is the hardest part and def not a requirement but i am a golang noob so i'm not sure if this is truly a lost cause in terms of value to effort

zoedt (approved):
overall lgtm, just a comment about the recurring parsing logic (that i hadn't solved either lol) nothing blocking and if there's not many more analytics things to add, maybe don't need to focus that in another pr later at all

pkg/collector/corechecks/network-devices/versa/client/client.go
ken-schneider: I've gone back and forth on whether or not it's a good idea to try and use reflection to generalize some of the logic. I just don't know that it's worth the overhead.

I do plan to refactor things out to have all the parsing for these structs in a separate file at least, but since I've got a few PRs out here I wanted to wait until those are in and I can handle that separately.

ken-schneider: Ah, yes. I added a field quickly and forgot to update these. Thanks for pointing it out!

PR #39028, opened 2025-07-17
New bookmark system with initial bookmark seeding

davidioin:
### What does this PR do?
 Implements initial bookmark seeding for Windows Event Logs to fix the amnesia bug where events occurring while the Datadog Agent is down are lost if no bookmark was saved before shutdown. This fix applies to both:
  - The Event Log Check integration (comp/checks/windowseventlog/)
  - The Logs tailer (pkg/logs/tailers/windowsevent/)

  When no bookmark exists, the agent now creates and immediately persists an initial bookmark from the most recent event (or an empty bookmark for empty logs), ensuring no events are lost during agent restarts.
### Motivation
(https://datadoghq.atlassian.net/browse/WINA-596) Solves the issue described in this ticket where bookmark would in the following scenario, the agent will see no events:
Agent configured to monitor an event log
Agent is stopped
Event records written to event log
Agent is started


.claude/settings.local.json
```diff
@@ -0,0 +1,31 @@
+{
```
ddrthall: This file looks like it snuck in.

ddrthall (approved):

releasenotes/notes/windows-event-log-bookmark-fix-d8ae5cfd846ab0e2.yaml
```diff
@@ -0,0 +1,12 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+fixes:
+  - |
+    Fixed an issue where the Windows Event Log tailer could lose historical events (the 'amnesia bug') when it started with no prior bookmark and no events were processed immediately. The agent now correctly seeds an initial bookmark from the most recent event upon startup, ensuring no events are missed on subsequent runs.
```
urseberry:
```suggestion
    Fixed an issue where the Windows Event Log tailer could lose historical events (the 'amnesia bug') when it started with no prior bookmark and no events were processed immediately. The Agent now correctly seeds an initial bookmark from the most recent event upon startup, ensuring no events are missed on subsequent runs.
```

urseberry (approved):

pkg/logs/tailers/windowsevent/tailer.go
```diff
@@ -312,22 +327,74 @@ func (t *Tailer) enrichEvent(m *windowsevent.Map, event evtapi.EventRecordHandle
...
+	err = bookmark.Update(returnedHandles[0])
+	if err != nil {
+		bookmark.Close()
+		evtapi.EvtCloseRecord(t.evtapi, returnedHandles[0])
+		return nil, fmt.Errorf("failed to update initial bookmark: %w", err)
+	}
+
+	// Clean up the event handle
+	evtapi.EvtCloseRecord(t.evtapi, returnedHandles[0])
+
```
clarkb7: we must always close the event handle, so we can use defer to make closing easier to manage.

```suggestion
        defer evtapi.EvtCloseRecord(t.evtapi, returnedHandles[0])

	// Create bookmark from the most recent event
	bookmark, err := evtbookmark.New(evtbookmark.WithWindowsEventLogAPI(t.evtapi))
	if err != nil {
		return nil, fmt.Errorf("failed to create initial bookmark: %w", err)
	}

	// Update bookmark with the most recent event
	err = bookmark.Update(returnedHandles[0])
	if err != nil {
		bookmark.Close()
		return nil, fmt.Errorf("failed to update initial bookmark: %w", err)
	}

	// Clean up the event handle
```

pkg/logs/tailers/windowsevent/tailer.go
```diff
@@ -312,22 +327,74 @@ func (t *Tailer) enrichEvent(m *windowsevent.Map, event evtapi.EventRecordHandle
...
 	defer evtapi.EvtClosePublisherMetadata(t.evtapi, pm)
 
 	windowsevent.AddRenderedInfoToMap(m, t.evtapi, pm, event)
 
 	return nil
 }
+
+// createInitialBookmark creates a bookmark from the most recent event in the log
+// This ensures we have a valid starting position even if no events are processed
+func (t *Tailer) createInitialBookmark() (evtbookmark.Bookmark, error) {
```
clarkb7: Check and Tailer both have this function that does the same thing. Since it's a common/shared task to initialize the bookmark, lets move it to the shared library in `pkg\util\winutil\eventlog\bookmark\`.
something like
```go
// FromLatestEvent creates a bookmark that points to the most recent event from channel/query
//
// A multi-channel query will return a bookmark with information from all channels in the query
func FromLatestEvent(channel string, query string) (Bookmark, error)
```

comp/checks/windowseventlog/windowseventlogimpl/check/subscription.go
```diff
@@ -148,6 +179,87 @@ func (c *Check) initSubscription() error {
...
+
+	if isMultiChannel {
+		// For multi-channel queries, path is ignored and query contains the XML QueryList
+		path = ""
+		flags = evtapi.EvtQueryFilePath | evtapi.EvtQueryReverseDirection
+	} else {
+		// Single channel query
+		path = channelPath
+		flags = evtapi.EvtQueryChannelPath | evtapi.EvtQueryReverseDirection
+	}
```
clarkb7: lets leave this out when we move the function. We don't support evtx file paths, and EvtQuery handles the single/multi channel query selection itself, so we don't need this case here.

comp/checks/windowseventlog/windowseventlogimpl/check/check_test.go
```diff
@@ -230,6 +231,154 @@ start: oldest
...
+	require.NoError(s.T(), err)
+
+	// Create new check instance (simulating restart)
+	check2, err := s.newCheck(instanceConfig)
+	require.NoError(s.T(), err)
+	defer check2.Cancel()
+
+	// Should collect only the missed events (not starting from "latest" again)
+	s.assertCountEvents(check2, missedEvents)
+}
```
clarkb7: `assertCountEvents` just asserts the check saw three events, it doesn't differentiate between the first 5 generated events and the three events generated later. If we add `assertNoEvents` then we'll know that the check had no more events to read, so it must have read the most recent three.

```suggestion
	// Should collect only the missed events (not starting from "latest" again)
	s.assertCountEvents(check2, missedEvents)
	s.assertNoEvents(check2)
}
```

comp/checks/windowseventlog/windowseventlogimpl/check/check_test.go
```diff
@@ -230,6 +231,154 @@ start: oldest
...
+
+// Test initial bookmark creation with empty event log
+func (s *GetEventsTestSuite) TestInitialBookmarkWithEmptyLog() {
+	// Ensure log is empty
+	err := s.ti.API().EvtClearLog(s.channelPath)
+	require.NoError(s.T(), err)
```
clarkb7: this is done before each test via `SetupTest`, we don't need to do it again here.

comp/logs/auditor/mock/auditor_mock.go
```diff
@@ -131,7 +131,7 @@ func (r *Registry) SetTailingMode(tailingMode string) {
 	r.tailingMode = tailingMode
 }
 
-// SetTailed stores the tailed status of the identifier.
+// SetTailed stores the tailed status of the identifiers
```
clarkb7, line 134: i think singular is correct? lets revert this change

pkg/logs/tailers/windowsevent/tailer_test.go
```diff
@@ -40,13 +40,15 @@ type ReadEventsSuite struct {
 }
 
 func TestReadEventsSuite(t *testing.T) {
-	testerNames := eventlog_test.GetEnabledAPITesters()
+	for _, tiName := range eventlog_test.GetEnabledAPITesters() {
+		// create a copy so the closure captures the right value
+		tiName := tiName
```
clarkb7: this is no longer needed since go 1.22
https://go.dev/blog/loopvar-preview
```suggestion
```

pkg/logs/tailers/windowsevent/tailer_test.go
```diff
@@ -40,13 +40,15 @@ type ReadEventsSuite struct {
...
 
-	for _, tiName := range testerNames {
-		if tiName != "Windows" {
-			t.Skip("test interface not implemented")
-		}
 		t.Run(fmt.Sprintf("%sAPI", tiName), func(t *testing.T) {
+			if tiName != "Windows" {
```
clarkb7, line 45: good catch

pkg/logs/tailers/windowsevent/tailer_test.go
```diff
@@ -267,6 +270,227 @@ func (s *ReadEventsSuite) TestBookmarkNewTailer() {
...
+	// are processed before shutdown.
+
+	// Setup test environment
+	channelPath := "dd-test-channel-initial-bookmark"
+	eventSource := "dd-test-source-initial-bookmark"
+
+	ti := eventlog_test.GetAPITesterByName("Windows", t)
+	if ti == nil {
+		t.Skip("Windows API tester not available")
+	}
```
clarkb7: add the new tests to `ReadEventsSuite` so they can take advantage of the common setup and teardown done by the suite

pkg/logs/tailers/windowsevent/tailer_test.go
```diff
@@ -267,6 +270,227 @@ func (s *ReadEventsSuite) TestBookmarkNewTailer() {
 	// if tailer started from bookmark correctly, there should only be s.numEvents
 }
 
+func TestInitialBookmarkSeeding(t *testing.T) {
+	// Test that verifies the fix for the amnesia bug:
+	// When a tailer starts with no bookmark, it creates an initial bookmark
+	// from the most recent event and saves it immediately, even if no events
+	// are processed before shutdown.
```
clarkb7: go convention puts comments above the function

```suggestion
// TestInitialBookmarkSeeding tests ....
func TestInitialBookmarkSeeding(t *testing.T) {
```

pkg/logs/tailers/windowsevent/tailer_test.go
```diff
@@ -267,6 +267,177 @@ func (s *ReadEventsSuite) TestBookmarkNewTailer() {
...
+	// The implementation now uses direct SetOffset instead of synthetic messages
+	var initialBookmark string
+
+	// Wait a moment for the tailer to initialize and set the bookmark
+	time.Sleep(100 * time.Millisecond)
+
+	// Get the bookmark directly from the mock registry
+	mockRegistry := tailer.registry.(*auditormock.Auditor)
+	initialBookmark = mockRegistry.GetOffset(tailer.Identifier())
+	s.Require().NotEmpty(initialBookmark, "Initial bookmark must not be empty")
```
clarkb7, line 304: avoid sleeps, different environments can have different delays and cause flaky failures. Instead, use a wait loop

```suggestion
        initialBookmark := ""
    	err := backoff.Retry(func() error {
            // Get the bookmark directly from the mock registry
            mockRegistry := tailer.registry.(*auditormock.Auditor)
            initialBookmark = mockRegistry.GetOffset(tailer.Identifier())
            if initialBookmark == "" {
                return fmt.Errorf("bookmark is empty")
            }
            return nil
	}, backoff.NewConstantBackOff(100*time.Millisecond))
	s.Require().NoError(err, "initial bookmark must not be empty")
```

pkg/logs/tailers/windowsevent/tailer_test.go
```diff
@@ -267,6 +267,177 @@ func (s *ReadEventsSuite) TestBookmarkNewTailer() {
...
+drainLoop:
+	for {
+		select {
+		case msg := <-msgChan:
+			// Update bookmark if we process real events
+			initialBookmark = msg.Origin.Offset
+		case <-drainTimeout:
+			break drainLoop
+		}
+	}
```
clarkb7, line 327: the tailer should start at latest event so it shouldn't read any of the 10 original events, so we should not need this loop. Did you see otherwise?

pkg/logs/tailers/windowsevent/tailer_test.go
```diff
@@ -267,6 +267,177 @@ func (s *ReadEventsSuite) TestBookmarkNewTailer() {
...
+		case <-eventTimeout:
+			break collectLoop
+		}
+	}
+
+	// The exact count might vary slightly due to timing and which events
+	// were processed before shutdown, but we should receive approximately
+	// the missed events (not 0, and not all 15)
+	s.Require().Greater(receivedEvents, uint(0), "Should receive at least some events (not 0 due to amnesia)")
+	s.Require().LessOrEqual(receivedEvents, missedEvents+uint(2), "Should not receive significantly more than missed events")
```
clarkb7, line 371: it shouldn't vary, we should see tham all before stopping the tailer. If we're failing here maybe we should increase or remove the `eventTimeout`

clarkb7 (approved):

PR #39045, opened 2025-07-18
[Closed] Do not add default metadata to profile when device does not report it

Pierre-L42:
### What does this PR do?

This PR adds the logic to check if a device does report a MIB table from the default metadata by trying one OID from the table. If the device does not have a value for this one OID, then we don't add any of the OIDs from the MIB table to the device profile, if yes, we add it.
The device profile is refreshed regularly (time interval defined in the `profileRefreshDelay` variable) through the `BuildProfile` function to detect when an OID in the device appear.

Moved `buildprofile.go` into a new package and moved `BuildProfile` tests in `config_test.go` to `buildprofile_test.go` because we would get cyclic imports between `checkconfig`, `session`, and `buildprofile` packages otherwise.

### Motivation

Decrease the number of SNMP queries we do to the device every checks.

The changes have been tried on CiscoCML with the VPN tunnels metadata:

First with the `datadog-agent check snmp` command. The device has a VPN set up, thus the VPN OIDs were present in the list of OIDs we query to the device. Then, after remove VPN from the device, the VPN OIDs were indeed removed from the list after redoing a `check snmp`.

Second with running the Agent. The device has a VPN set up, we run the Agent, we see the VPN tunnel OID in the queried list. We remove the VPN from the device, we wait for `profileRefreshDelay`, and the VPN OIDs have been removed from the queried list. We readd the VPN to the device, we wait for `profileRefreshDelay`, and we do see again the VPN OIDs in the queried list.

Third, we check if the queried list from an Agent without these changed are the same as the ones with this change (except the ones that are filtered now).


pkg/collector/corechecks/snmp/internal/devicecheck/devicecheck.go
```diff
@@ -156,9 +159,30 @@ func NewDeviceCheck(config *checkconfig.CheckConfig, ipAddress string, sessionFa
...
+			defer func() {
+				err = d.session.Close()
+				if err != nil {
+					log.Warnf("failed to close session: %v", err)
+				}
+			}()
+		}
+	}
+
+	_, err = d.profileCache.Update("", time.Now(), d.session, validConnection, d.config)
```
AlexandreYang, line 185: Is that change related to the main goal ("Do not add default metadata to profile when device does not report it") of this PR?

pkg/collector/corechecks/snmp/internal/buildprofile/buildprofile.go
```diff
@@ -3,14 +3,18 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
```
AlexandreYang, line 3: Sorry, I might have missed that, but where do you implement "This PR adds the logic to check if a device does report a MIB table from the default metadata by trying one OID from the table." ?

AlexandreYang: Never mind, I see here: `checkOidIfConnectedOrSkip`

pkg/collector/corechecks/snmp/internal/devicecheck/devicecheck.go
AlexandreYang: I mean, why do we need to refresh the session here?

AlexandreYang: Ok, I think I see the logic here, we want to only skip some OIDs if we have a valid connection.

But I'm wondering if we need that "validConnection" verification at all.

I mean, if the session is not valid, we won't be able to retrieve any OIDs anyway right? So, I'm wondering if we can just not have the code above here.

pkg/collector/corechecks/snmp/internal/devicecheck/devicecheck.go
```diff
@@ -105,9 +107,10 @@ func (pc *profileCache) IsOutdated(sysObjectID string, profileName string, lastU
 		return true
 	}
 	// If we get here then either we're auto-detecting but the sysobjectid hasn't
-	// changed, or we have a static name; either way we're out of date if and only
-	// if the profile provider has updated.
-	return pc.timestamp.Before(lastUpdate)
+	// changed, or we have a static name; either way we're out of date if the profile
+	// refresh delay has been exceeded or if the profile provider has updated.
+	return now.Sub(pc.timestamp) > profileRefreshDelay*time.Second ||
+		pc.timestamp.Before(lastUpdate)
```
AlexandreYang: Maybe we can have a separate condition, to keep things more readable and have standalone clearer comment.
```suggestion
	if now.Sub(pc.timestamp) > profileRefreshDelay*time.Second {
		// TODO: COMMENT
		return true
	}

	// changed, or we have a static name; either way we're out of date if and only
	// if the profile provider has updated.
	return pc.timestamp.Before(lastUpdate)
```

pkg/collector/corechecks/snmp/internal/buildprofile/default_metadata.go
```diff
@@ -0,0 +1,471 @@
...
+func checkOidIfConnectedOrSkip(sess session.Session, validConnection bool, oid string) bool {
+	if !validConnection {
+		return true
+	}
+	return checkOid(sess, oid)
+}
+
+// checkOid checks whether a given OID is present on the device
+func checkOid(sess session.Session, oid string) bool {
+	result, err := sess.GetNext([]string{oid})
```
AlexandreYang: Ideally we might want to batch the GetNext oids to reduce number of call, but maybe not a big issue if we only run that once every 10min.

pkg/collector/corechecks/snmp/internal/devicecheck/devicecheck.go
Pierre-L42: But if the session is not valid, then `checkOid` would return `false` because an error would be returned by `GetNext`, while we want to return `true` when session is not valid because we still want the metadata merged in this case.

Actually I'm wondering if we can just remove the call to `profileCache.Update` in `NewDeviceCheck` which called during the Configure of the SNMP check. Because we also call `profileCache.Update` during the Run of the check, so if we don't call the update in `NewDeviceCheck`, then it would be called in the first Run anyway. And I don't see the profileCache being used between the Configure and the first Run. That would remove everything about creating the session in `NewDeviceCheck` and `validConnection` thing.
Do you see any problem that could happen with that?

pkg/collector/corechecks/snmp/internal/devicecheck/devicecheck.go
Pierre-L42: Done in [4384418](https://github.com/DataDog/datadog-agent/pull/39045/commits/4384418e8fb82b1e33630107dbb6a3396082aae4) and [57de8c5](https://github.com/DataDog/datadog-agent/pull/39045/commits/57de8c54a4a94bdcbec9bd6cf9907320bad9aa0e)

pkg/collector/corechecks/snmp/internal/devicecheck/devicecheck.go
Pierre-L42: It is also fine to me if you think we should return `false` and not merge the metadata in the case we don't have a valid session.

pkg/collector/corechecks/snmp/internal/buildprofile/default_metadata.go
Pierre-L42: Changed that in [d37eea6](https://github.com/DataDog/datadog-agent/pull/39045/commits/d37eea683f69752d0b43916ed2a2b62254bf13c8)

PR #39048, opened 2025-07-18
Print traceback in case of InfraError in Static Quality Gates

JSGette:
### What does this PR do?
It simply adds `traceback.print_exception(e)` in case of `InfraError` to ensure
that we see the original error message and where it occurred.
### Motivation
Sometimes static quality gates are failing, but we are unable to see
the root cause as original InfraError exception is suppressed and
the error message is replaced with something generic. This PR
improves visibility of errors that should enable developers in
debugging.

New unit test is added to ensure we see the original InfraError message.

None

### DIFF
--- a/tasks/quality_gates.py
+++ b/tasks/quality_gates.py
@@ -196,7 +196,7 @@ def parse_and_trigger_gates(ctx, config_path=GATE_CONFIG_PATH):
             gate_states.append({"name": gate, "state": False, "error_type": "AssertionError", "message": str(e)})
         except InfraError as e:
             print(f"Gate {gate} flaked ! (InfraError)\n Restarting the job...")
-            traceback.print_exc()
+            traceback.print_exception(e)
             ctx.run("datadog-ci tag --level job --tags static_quality_gates:\"restart\"")
             raise Exit(code=42) from e
         except Exception:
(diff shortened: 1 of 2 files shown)


tasks/quality_gates.py
```diff
@@ -196,6 +196,7 @@ def parse_and_trigger_gates(ctx, config_path=GATE_CONFIG_PATH):
             gate_states.append({"name": gate, "state": False, "error_type": "AssertionError", "message": str(e)})
         except InfraError as e:
             print(f"Gate {gate} flaked ! (InfraError)\n Restarting the job...")
+            traceback.print_exc()
```
rdesgroppes: sad we're not using `logging`, otherwise we would have leveraged sth like:
```suggestion
            logging.exception(f"Gate {gate} flaked ! (InfraError)\n Restarting the job...")
```

rdesgroppes (approved):
Thanks!

tasks/quality_gates.py
alopezz: Can we instead use 
```suggestion
            traceback.print_exception(e)
```
Doesn't change much in practice perhaps, but it looks safer to not rely on implicit context.

JSGette: I was going to come up with the plan on how to improve SQG but would park this discussion at least until end of next week if it works for you

JSGette: Done

alopezz (approved):
Thanks!

CelianR (approved):

PR #39050, opened 2025-07-18
Create custom MacOS temporary directory

CelianR:
### What does this PR do?

This creates a custom temporary directory for each job. The goal is to avoid collisions within `/tmp` we can have.
Everything is exported under `$TMPDIR` and can be reused. (also applies to processes spawned from there)

### DIFF
--- a/.gitlab/common/macos.yml
+++ b/.gitlab/common/macos.yml
@@ -16,6 +16,8 @@
       dda inv -- -e macos.remove-inactive-versions -l python -t "$PYTHON_VERSION" || true
       dda inv -- -e macos.remove-inactive-versions -l go -t "$(cat .go-version)" || true
     fi
+  # Flush temporary folder
+  - rm -rf /tmp/gitlabci
 
 .select_python_env_commands:
   # Select the virtualenv using the current Python version. Create it if it doesn't exist.
@@ -105,6 +107,11 @@
     - !reference [.install_python_dependencies]
     # See there is no virtualization, we need to clean the environment regularly
     - !reference [.macos_runner_maintenance]
+    # Create temporary folder
+    - |
+      export MACOS_TMPDIR="/tmp/gitlabci/$CI_JOB_ID"
+      mkdir -p "$MACOS_TMPDIR"
+      echo "Temporary folder created at $MACOS_TMPDIR"
     - dda inv -- -e rtloader.make
     - dda inv -- -e rtloader.install
     - dda inv -- -e install-tools
--- a/.gitlab/common/macos.yml
+++ b/.gitlab/common/macos.yml
@@ -107,7 +107,7 @@
     - !reference [.install_python_dependencies]
     # See there is no virtualization, we need to clean the environment regularly
     - !reference [.macos_runner_maintenance]
-    # Create temporary folder
+    # Create custom temporary folder to isolate jobs from each other
     - |
       export TMPDIR="/tmp/gitlabci/$CI_JOB_ID"
       mkdir -p "$TMPDIR"
--- a/.gitlab/common/macos.yml
+++ b/.gitlab/common/macos.yml
@@ -17,7 +17,7 @@
       dda inv -- -e macos.remove-inactive-versions -l go -t "$(cat .go-version)" || true
     fi
   # Flush temporary folder
-  - rm -rf /tmp/gitlabci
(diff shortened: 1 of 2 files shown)


.gitlab/common/macos.yml
```diff
@@ -105,6 +107,11 @@
     - !reference [.install_python_dependencies]
     # See there is no virtualization, we need to clean the environment regularly
     - !reference [.macos_runner_maintenance]
+    # Create temporary folder
```
alopezz: Nit: I would probably expand this comment a bit to mention that the goal is to provide a job-specific default tmpdir to improve isolation relying on the fact that TMPDIR is a standard variable that is taken into account by most tools.

.gitlab/common/macos.yml
```diff
@@ -16,6 +16,8 @@
       dda inv -- -e macos.remove-inactive-versions -l python -t "$PYTHON_VERSION" || true
       dda inv -- -e macos.remove-inactive-versions -l go -t "$(cat .go-version)" || true
     fi
+  # Flush temporary folder
+  - rm -rf /tmp/gitlabci
```
alopezz: Shouldn't this rather be `rm -rf $TMPDIR`? In case one day we get concurrent jobs on the same runner or something like that.

alopezz (approved):
Looks good, I left a couple of suggestions and wait to confirm with CI that everything works 👍 . Thanks!

.gitlab/common/macos.yml
```diff
@@ -105,6 +107,11 @@
     - !reference [.install_python_dependencies]
     # See there is no virtualization, we need to clean the environment regularly
     - !reference [.macos_runner_maintenance]
+    # Create temporary folder
+    - |
+      export TMPDIR="/tmp/gitlabci/$CI_JOB_ID"
```
rdesgroppes, line 112: Even better: using the _standard_ `TMPDIR` environment variable would make sure any temp file or directory created through `mktemp` (and variants: `mkstemp`, `mkdtemp`, etc.) honors the job id-based split.

Some refs:
- https://apple.stackexchange.com/questions/94964/where-is-the-temp-folder
- https://osxdaily.com/2018/08/17/where-temp-folder-mac-access

💡 a frequent symptom for processes "gardening" in the same temp file is that they may get killed with signals 11 (SIGSEGV) and 7 (SIGBUS) esp. `java` for the latter.

.gitlab/common/macos.yml
CelianR: This is executed at the beginning of the script so the goal is to remove past folders and avoid collisions (that's why I added $CI_JOB_ID and I'm not only using /tmp/gitlabci). I don't think that we will have concurrency on those runners without virtualization

rdesgroppes (approved):
Thank you!

.gitlab/common/macos.yml
chouetz: Is there a risk to move things here and make them deleted? I'm always scared by `rm -rf` (so my remark is similar to Alex's)

chouetz (approved):
I 👍🏽 on @rdesgroppes's comment about `TMPDIR`

.gitlab/common/macos.yml
alopezz: If we're really sure we'll never have concurrency here, then just doing `rm -rf $TMPDIR` could be enough (without setting up the TMPDIR). Basically, if we wipe the whole common tmpdir at the beginning of every run it then the pipeline-ID based TMPDIR differentiation doesn't add anything.

I think we should try to be consistent; either we completely buy into the assumption that there will never be concurrency and then just wipe the common default $TMPDIR at the start of every job, or we don't and then we only ever touch the pipeline-id specific folder.

CelianR: @alopezz 
> TMPDIR differentiation doesn't add anything
Not really in my opinion, there are always edge cases where files cannot be deleted ("temporary" files that the os delete after some time but you cannot delete them even with `sudo`). Using this would avoid extra edge cases and I don't see the benefits of having only a single directory.

@chouetz 
This is wiped after the job ends since it's in the `before_script` (the job after wipes the previous job) so there shouldn't be anything important here for the job / the runner (the runner is not affected since the variable is changed only within the job). It's possible that some binaries are using this like a cache but it will only slow down the job in the worst case scenario

PR #39056, opened 2025-07-18
[CWS-4923] Add Protocol in flow_pid map and consistently get it

theop-dd:
### What does this PR do?
Adds the protocol to the key of the flow_pid map and uses the recently introduced helper to consistently retrieve the protocol for all socket types.

### Motivation
This PR leverages a newly added helper to reliably determine the protocol of any socket. This enables us to distinguish entries in the flow_pid map when multiple sockets are bound to the same port but use different protocols (e.g., TCP vs UDP).

Two functional tests were added to verify that entries in the map can be differentiated based solely on the protocol, and to confirm that the associated PIDs are correctly reported for each case.

### DIFF
--- a/pkg/security/ebpf/c/include/hooks/network/flow.h
+++ b/pkg/security/ebpf/c/include/hooks/network/flow.h
@@ -1,6 +1,5 @@
 #ifndef _HOOKS_NETWORK_FLOW_H_
 #define _HOOKS_NETWORK_FLOW_H_
-
 #include "constants/offsets/network.h"
 #include "constants/offsets/netns.h"
 #include "helpers/network/pid_resolver.h"
@@ -53,9 +52,9 @@ int hook_security_sk_classify_flow(ctx_t *ctx) {
             // this is an AF_INET6 flow
             bpf_probe_read(&key.addr, sizeof(u64) * 2, (void *)fl + get_flowi6_saddr_offset());
             bpf_probe_read(&key.l4_protocol, 1, (void *)fl + get_flowi6_proto_offset());
-            if (key.l4_protocol == 0) {
-                key.l4_protocol = 3; 
-            }
+            char msg[] = "flow-if (flow_family == AF_INET6): l4_protocol: %u\n";
+            bpf_trace_printk(msg, sizeof(msg), key.l4_protocol);
+
 
         }
     }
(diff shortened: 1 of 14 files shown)


urseberry (approved):

pkg/security/tests/flow_pid_bind_test.go
```diff
@@ -797,5 +834,283 @@ func TestFlowPidBindLeak(t *testing.T) {
...
+						}
+					}
+					if strings.HasPrefix(line, "Listening on port") {
+						tcplistenReady <- struct{}{} // Synchro on listen ready
+					}
+					if strings.HasPrefix(line, "Closing socket...") {
+						tcpCloseReady <- struct{}{} // Synchro on close ready
+					}
+				}
+				_ = cmd.Wait()
```
paulcacheux: I would suggest to put this in a defer, this way even a `t.Fatal` will still wait the process

pkg/security/tests/flow_pid_bind_test.go
```diff
@@ -797,5 +834,283 @@ func TestFlowPidBindLeak(t *testing.T) {
...
+						}
+					}
+					if strings.HasPrefix(line, "Waiting on port") {
+						udpwaitReady <- struct{}{} // Synchro on wait ready
+					}
+					if strings.HasPrefix(line, "Closing socket...") {
+						udpCloseReady <- struct{}{} // Synchro on close ready
+					}
+				}
+				_ = cmd.Wait()
```
paulcacheux: same here

pkg/security/ebpf/c/include/hooks/network/flow.h
```diff
@@ -287,6 +286,8 @@ __attribute__((always_inline)) int handle_sk_release(struct sock *sk) {
     if (route.netns == 0) {
         return 0;
     }
+    // extraact protocol
```
safchain: nit: extraact -> extract

pkg/security/ebpf/c/include/hooks/network/flow.h
```diff
@@ -565,10 +566,11 @@ __attribute__((always_inline)) int handle_inet_bind_ret(int ret) {
 
     // add netns information
     route.netns = get_netns_from_sock(sk);
+    route.l4_protocol = get_protocol_from_sock(sk);
     if (route.netns != 0) {
         bpf_map_update_elem(&netns_cache, &tid, &route.netns, BPF_ANY);
     }
-
+
```
safchain: unless change

pkg/security/probe/constantfetch/fallback.go
```diff
@@ -1004,3 +1004,14 @@ func getSockStructSKProtocolOffset(kv *kernel.Version) uint64 {
 		return ErrorSentinel
 	}
 }
+
+func getFlowiProtoOffset(kv *kernel.Version) uint64 {
+	switch {
+	case kv.IsRockyKernel() && kv.IsInRangeCloseOpen(kernel.Kernel5_14, kernel.Kernel5_15):
+		return 18
```
YoannGh, line 1011: nit: Do we need this extra case statement if the returned value is the same as the returned one?

theop-dd: This is tricky but if I remove this case, one of the rocky kernel will have the offset set at 14 because the version of the kernel will match the case. However, this specific version needs to have the offset at 18.

theop-dd: In theory, all the kernel before 5.18 are supposed to have the offset at 14, but for some reason, rocky 5.14 has the "new" offset at 18

theop-dd: I can add a check in the case 14 to skip this kernel, so it will drop in the default case.

safchain (approved):

pkg/security/probe/constantfetch/fallback.go
YoannGh: My bad I didn't notice the overlap between the kernel version ranges

YoannGh (approved):

PR #39063, opened 2025-07-18
Add a retry to deploy_windows_testing-a7* jobs

FlorentClarret:
### What does this PR do?

Add a retry to deploy_windows_testing-a7* jobs

### Motivation

https://datadoghq.atlassian.net/browse/BARX-769


chouetz (approved):
I'm not sure if we should try to ensure we do the retry only on the `core dump` case.

PR #39064, opened 2025-07-18
[ AGNTLOG-249 ] Reduce the Logs Agent HTTP timeout back to 5 seconds on agent startup

ddrthall:
### What does this PR do?
HTTP timeout was intentionally set to a smaller value during agent startup to prevent delays. This functionality was overridden when the timeout was made to be configurable, and has been re-added here.

### DIFF
--- a/pkg/logs/client/http/destination.go
+++ b/pkg/logs/client/http/destination.go
@@ -39,6 +39,12 @@ const (
 	ProtobufContentType = "application/x-protobuf"
 )
 
+// NoTimeoutOverride is a special value that tells the httpClientFactory to use the logs_config.http_timeout setting.
+// This should generally be used for all HTTP destinations barring special cases like the HTTP connectivity check.
+const (
+	NoTimeoutOverride = -1
+)
+
 // HTTP errors.
 var (
 	errClient  = errors.New("client error")
@@ -112,7 +118,7 @@ func NewDestination(endpoint config.Endpoint,
 	return newDestination(endpoint,
 		contentType,
 		destinationsContext,
-		0,
+		NoTimeoutOverride,
 		shouldRetry,
 		destMeta,
 		cfg,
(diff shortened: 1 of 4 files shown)


pkg/logs/client/http/destination.go
```diff
@@ -112,7 +112,7 @@ func NewDestination(endpoint config.Endpoint,
 	return newDestination(endpoint,
 		contentType,
 		destinationsContext,
-		time.Second*10,
+		0,
```
jszwedko: Nit: I wish Go had union types. It isn't readily apparent that `0` means "no override" though there is some precedent with `0` and negative values meaning "no timeout" in the Go networking libraries.

I suppose to model this in a more readable way we could introduce an `interface` and pass in implementers like `NoTimeoutOverride` and `TimeoutOverride`.

jszwedko (commented):
I'm not immediately seeing how this sets the timeout at start-up to 5 seconds. Is there code not modified by this PR that would be relevant to look at?

Also, is there a test we could add that would have caught this?

ddrthall:
> I'm not immediately seeing how this sets the timeout at start-up to 5 seconds. Is there code not modified by this PR that would be relevant to look at?
> 
> Also, is there a test we could add that would have caught this?

[Here](https://github.com/DataDog/datadog-agent/blob/7a5bf686873546703e56a930d1ff7fd13cd81808/pkg/logs/client/http/destination.go#L491) is the code that limits to 5 seconds

jszwedko:
> > I'm not immediately seeing how this sets the timeout at start-up to 5 seconds. Is there code not modified by this PR that would be relevant to look at?
> > Also, is there a test we could add that would have caught this?
> 
> [Here](https://github.com/DataDog/datadog-agent/blob/7a5bf686873546703e56a930d1ff7fd13cd81808/pkg/logs/client/http/destination.go#L491) is the code that limits to 5 seconds

Gotcha, thanks! That was the missing piece for me.

jszwedko (approved):

pkg/logs/client/http/destination.go
ddrthall: Dropped in a constant to add clarity - if you prefer more type safety than that offers, rather than an timeout interface I would prefer to move reading configuration values to a higher level function and specifying all of the configurable values explicitly in the httpClientFactory signature.

jszwedko: I'm happy with the readability a constant adds. Thanks!

urseberry (approved):

PR #39066, opened 2025-07-18
[CWS] introduce network filter action

safchain:
### What does this PR do?

The goal is to define network filter action to drop packet for a process or for a cgroup

```
    expression: exec.file.path == "/usr/bin/free"
    actions:
      - network_filter:
          filter: dst 172.67.133.176 or dst 104.21.5.178
          policy: drop
          scope: cgroup

```

### DIFF
--- a/pkg/security/ebpf/c/include/constants/custom.h
+++ b/pkg/security/ebpf/c/include/constants/custom.h
@@ -66,12 +66,12 @@ enum TC_TAIL_CALL_KEYS {
 };
 
 // see probes/rawpacket/pcap.go
-#define RAW_PACKET_FILTER_MAX_TAIL_CALL 5
+#define RAW_PACKET_MAX_TAIL_CALL 5
 
 enum TC_RAWPACKET_KEYS {
     RAW_PACKET_FILTER,
     // reserved keys for raw packet filter tail calls
-    RAW_PACKET_SHOOTER = RAW_PACKET_FILTER + RAW_PACKET_FILTER_MAX_TAIL_CALL,
+    RAW_PACKET_DROP_ACTION = RAW_PACKET_FILTER + RAW_PACKET_MAX_TAIL_CALL + 1, // + 1 for the sender program
 };
 
 #define DNS_MAX_LENGTH 256
--- a/pkg/security/ebpf/c/include/constants/custom.h
+++ b/pkg/security/ebpf/c/include/constants/custom.h
@@ -65,9 +65,13 @@ enum TC_TAIL_CALL_KEYS {
     DNS_RESPONSE
 };
 
+// see probes/rawpacket/pcap.go
+#define RAW_PACKET_FILTER_MAX_TAIL_CALL 5
+
 enum TC_RAWPACKET_KEYS {
     RAW_PACKET_FILTER,
     // reserved keys for raw packet filter tail calls
+    RAW_PACKET_SHOOTER = RAW_PACKET_FILTER + RAW_PACKET_FILTER_MAX_TAIL_CALL,
 };
 
 #define DNS_MAX_LENGTH 256
--- a/pkg/security/ebpf/c/include/constants/custom.h
+++ b/pkg/security/ebpf/c/include/constants/custom.h
@@ -263,7 +263,8 @@ enum link_target_dentry_origin {
 };
 
 enum global_rate_limiter_type {
-    RAW_PACKET_LIMITER = 0,
+    RAW_PACKET_FILTER_LIMITER = 0,
+    RAW_PACKET_ACTION_LIMITER,
 };
 
 #define TAIL_CALL_FNC_NAME(name, ...) tail_call_##name(__VA_ARGS__)
--- a/pkg/security/ebpf/c/include/hooks/network/tc.h
+++ b/pkg/security/ebpf/c/include/hooks/network/tc.h
@@ -36,7 +36,7 @@ __attribute__((always_inline)) int prepare_raw_packet_event(struct __sk_buff *sk
     }
 
     evt->process.pid = pkt->pid;
-    evt->container.cgroup_context.cgroup_file.ino = pkt->cgroup_id;
+    evt->cgroup.cgroup_file.ino = pkt->cgroup_id;
 
     bpf_skb_pull_data(skb, 0);
 
--- a/pkg/security/ebpf/c/include/hooks/network/tc.h
+++ b/pkg/security/ebpf/c/include/hooks/network/tc.h
@@ -28,12 +28,14 @@ int classifier_egress(struct __sk_buff *skb) {
     return route_pkt(skb, pkt, EGRESS);
 };
 
-__attribute__((always_inline)) int prepare_raw_packet_event(struct __sk_buff *skb) {
+__attribute__((always_inline)) int prepare_raw_packet_event(struct __sk_buff *skb, struct packet_t *pkt) {
     struct raw_packet_event_t *evt = get_raw_packet_event();
     if (evt == NULL) {
         // should never happen
         return TC_ACT_UNSPEC;
     }
+
+    evt->process.pid = pkt->pid;
 
     bpf_skb_pull_data(skb, 0);
 
@@ -90,7 +92,7 @@ int classifier_raw_packet_ingress(struct __sk_buff *skb) {
         return TC_ACT_UNSPEC;
     }
 
-    if (prepare_raw_packet_event(skb) != TC_ACT_UNSPEC) {
+    if (prepare_raw_packet_event(skb, pkt) != TC_ACT_UNSPEC) {
         return TC_ACT_UNSPEC;
     }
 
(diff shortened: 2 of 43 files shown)


git-thuerk-done (approved):
lgtm

pkg/security/probe/probe_ebpf.go
```diff
@@ -2998,6 +3116,37 @@ func (p *EBPFProbe) HandleActions(ctx *eval.Context, rule *rules.Rule) {
 			}
 		case action.Def.Hash != nil:
 			if p.fileHasher.HashAndReport(rule, action.Def.Hash, ev) {
+				p.probe.onRuleActionPerformed(rule, action.Def)
+			}
+		case action.Def.NetworkFilter != nil:
+			var policy rawpacket.Policy
```
paulcacheux, line 3126: It would probably be nice to move this raw packet code to some other file, similar to how it's done for the file hasher. This probe_ebpf file is already quite crowded

pkg/security/ebpf/c/include/constants/custom.h
```diff
@@ -259,7 +263,8 @@ enum link_target_dentry_origin {
 };
 
 enum global_rate_limiter_type {
-    RAW_PACKET_LIMITER = 0,
+    RAW_PACKET_FILTER_LIMITER = 0,
+    RAW_PACKET_ACTION_LIMITER,
```
YoannGh, line 267: The size of the `global_rate_limiters` map needs to be increased to handle this new key

pkg/security/ebpf/c/include/constants/custom.h
```diff
@@ -65,9 +65,13 @@ enum TC_TAIL_CALL_KEYS {
     DNS_RESPONSE
 };
 
+// see probes/rawpacket/pcap.go
+#define RAW_PACKET_MAX_TAIL_CALL 5
+
 enum TC_RAWPACKET_KEYS {
     RAW_PACKET_FILTER,
     // reserved keys for raw packet filter tail calls
+    RAW_PACKET_DROP_ACTION = RAW_PACKET_FILTER + RAW_PACKET_MAX_TAIL_CALL + 1, // + 1 for the sender program
```
YoannGh, line 74: Shouldn't it be `RAW_PACKET_DROP_ACTION = RAW_PACKET_FILTER + (RAW_PACKET_MAX_TAIL_CALL * 2) + 1`? to match  https://github.com/DataDog/datadog-agent/pull/39066/files#diff-af052d7be9b6d2db6307fa36cd9d7d5936cb77790b02a5d951836ad6f25eac17R102

pkg/security/ebpf/probes/rawpacket/pcap.go
```diff
@@ -110,40 +124,75 @@ func BPFFilterToInsts(index int, filter string, opts ProgOpts) (asm.Instructions
...
+			// == 0, no match
+			asm.JEq.Imm(cbpfcOpts.Result, 0, mismatchLabel).WithSymbol(resultLabel),
+
+			// check the pid
+			// load the pid from the packet
+			asm.LoadMem(asm.R7, opts.eventPtrReg, structRawPacketEventPidOffset, asm.Word),
+			asm.JEq.Imm(asm.R7, int32(filter.Pid), opts.onMatchLabel),
+			asm.Mov.Imm(asm.R4, 0).WithSymbol(mismatchLabel), // nop instruction, just hold the symbol
+		)
+	} else if !filter.CGroupPathKey.IsNull() {
```
YoannGh, line 147: Should we also check if `hasGetCurrentCgroupId` is true here?

pkg/security/secl/model/model.go
```diff
@@ -101,6 +101,8 @@ type NetworkContext struct {
 	Destination      IPPortContext `field:"destination"`       // destination of the network packet
 	NetworkDirection uint32        `field:"network_direction"` // SECLDoc[network_direction] Definition:`Network direction of the network packet` Constants:`Network directions`
 	Size             uint32        `field:"size"`              // SECLDoc[size] Definition:`Size in bytes of the network packet`
+
+	Action uint32 `field:"-"`
```
YoannGh: This field is not used

pkg/security/probe/probe_ebpf.go
```diff
@@ -682,6 +665,89 @@ func (p *EBPFProbe) setupRawPacketProgs(rs *rules.RuleSet) error {
...
+	opts := rawpacket.DefaultProgOpts()
+	opts.WithProgPrefix("raw_packet_drop_action_")
+	opts.WithGetCurrentCgroupID(p.kernelVersion.HasBpfGetCurrentPidTgidForSchedCLS())
+
+	// adapt max instruction limits depending of the kernel version
+	if p.kernelVersion.Code >= kernel.Kernel5_2 {
+		opts.MaxProgSize = 1_000_000
+	}
+
+	seclog.Errorf("generate rawpacket filter programs with a limit of %d max instructions", opts.MaxProgSize)
```
YoannGh: Should this log be removed or changed to an info level?

pkg/security/ebpf/c/include/hooks/network/tc.h
```diff
@@ -111,14 +100,29 @@ int classifier_raw_packet_egress(struct __sk_buff *skb) {
...
+
+    if (prepare_raw_packet_event(skb, pkt) != TC_ACT_UNSPEC) {
         return TC_ACT_UNSPEC;
     }
 
-    if (prepare_raw_packet_event(skb) != TC_ACT_UNSPEC) {
+    // call the drop action
+    if (pkt->pid > 0 || pkt->cgroup_id > 0) {
+        bpf_tail_call_compat(skb, &raw_packet_classifier_router, RAW_PACKET_DROP_ACTION);
+    }
```
YoannGh, line 118: Won't this tail call break raw_packet filters on the egress path?

Maybe we should use a new TC hook specifically for drop actions with a lower priority than the one used for raw_packet filters so that the action is performed only after the packets are matched against the raw_packet filters

pkg/security/ebpf/c/include/constants/custom.h
safchain: I'm not sure, RAW_PACKET_DROP_ACTION correspond to https://github.com/DataDog/datadog-agent/pull/39066/files#diff-af052d7be9b6d2db6307fa36cd9d7d5936cb77790b02a5d951836ad6f25eac17R100

pkg/security/ebpf/c/include/constants/custom.h
safchain: Done

pkg/security/ebpf/probes/rawpacket/pcap.go
safchain: I don't think this is required as we use a fallback kernel side when this helper is not present. https://github.com/DataDog/datadog-agent/pull/39066/files#diff-e33df0516f8f10c6eb987ae0ef565ff88425b53931ec6a9c4d2da35e7e3ceb42R105-R108

pkg/security/probe/probe_ebpf.go
safchain: Done

pkg/security/secl/model/model.go
safchain: Removed

pkg/security/ebpf/c/include/hooks/network/tc.h
safchain: I would like to no have the rawpacket filter program called and having packet captured being reported. Here if the packet if dropped, it will be reported as dropped by the action program, if the packet doesn't match the raw packet filter program will be called here : https://github.com/DataDog/datadog-agent/pull/39066/files#diff-81853b5953ffa7c874e254ecb233a0eb262dd713e2d266c6c0b143277bc04b69R352-R354

pkg/security/probe/probe_ebpf.go
safchain: I'll try to move all the raw packet code (not only the action) in a follow-up PR. This PR is already pretty complex, I would prefer to not move already existing piece of code in this PR.

pkg/security/ebpf/c/include/constants/custom.h
YoannGh: Indeed

pkg/security/probe/probe_ebpf.go
```diff
@@ -682,6 +665,89 @@ func (p *EBPFProbe) setupRawPacketProgs(rs *rules.RuleSet) error {
...
+
+	// add or close if none
+	return p.setupRawPacketProgs(progSpecs, probes.TCRawPacketDropActionKey, probes.RawPacketMaxTailCall, &p.rawPacketActionCollection)
+}
+
+func (p *EBPFProbe) addRawPacketActionFilter(actionFilter rawpacket.Filter) error {
+	seclog.Infof("add raw packet action filter: %+v\n", actionFilter)
+
+	if slices.ContainsFunc(p.rawPacketActionFilters, func(af rawpacket.Filter) bool {
+		return actionFilter.RuleID == af.RuleID
```
YoannGh: Won't this prevent from adding a new blocking filter in case the same rule is matched by another process/cgroup?

safchain: You're right, updated to add a key allowing more than one filter

YoannGh (approved):

PR #39069, opened 2025-07-18
[CWS] fix cgroup ns

safchain:
### What does this PR do?

This fixes the cgrouv2 & cgroupv1 snapshot in case of cgroup ns. It adds unit tests that are snapshot of /proc /sys for cgroupv1 & cgroupv2.

The cgroup root path is used to resolve cgroup mounts as well.


pkg/security/utils/cgroup.go
```diff
@@ -296,3 +317,64 @@ func GetCgroup2MountPoint() (string, error) {
 
 	return "", nil
 }
+
+func isInCGroupNamespace(pid uint32) (bool, error) {
+	cgroups, err := GetProcControlGroups(pid, pid)
+	if err != nil {
+		return false, err
+	}
+	for _, cgroup := range cgroups {
```
kovagsm: I think you can make the checks below for just one entry in the list to know if you're in the namespace and there's no need to iterate over all of them

lebauce (approved):

kovagsm (approved):

PR #39072, opened 2025-07-18
Adding `secret_refresh_interval` and `secret_refresh_scatter` to the config template

rahulkaukuntla:
### What does this PR do?

Even though API Key refresh has been GA since 7.67, the corresponding config options haven't been documented in the config template.


michaelcretzman (approved):

pkg/config/config_template.yaml
```diff
@@ -864,6 +864,24 @@ api_key:
...
+
+## @param secret_refresh_scatter - boolean - optional
+## @env DD_SECRET_REFRESH_SCATTER - boolean - optional
+##
+## `secret_refresh_scatter`, if set to true, will randomize the first secret refresh. `secret_refresh_interval` needs to be set
+## for this to take effect. This is useful to avoid all agents refreshing secrets at the same time.
+## For more information see: https://docs.datadoghq.com/agent/configuration/secrets-management/#refreshing-apiapp-keys-at-runtime
+##
+# secret_refresh_scatter: <REFRESH_SCATTER>
+
```
hush-hush: Let's move this after `secret_backend_config` so the order makes sens. We could also add a header for the secrets section now that we have a larger set of settings (similar to the header for logs line 968)

pkg/config/config_template.yaml
```diff
@@ -864,6 +864,24 @@ api_key:
...
 # windows_use_pythonpath: false
 {{ end }}
+## @param secret_refresh_interval - integer - optional
+## @env DD_SECRET_REFRESH_INTERVAL - integer - optional
+##
+## `secret_refresh_interval` is the interval (in seconds) at which api/app key secrets are refreshed.
+## For more information see: https://docs.datadoghq.com/agent/configuration/secrets-management/#refreshing-apiapp-keys-at-runtime
+##
+#
+# secret_refresh_interval: <REFRESH_INTERVAL>
```
hush-hush: Let's give a valid value as example:

```suggestion
# secret_refresh_interval: 3600
```

pkg/config/config_template.yaml
```diff
@@ -864,6 +864,24 @@ api_key:
...
+
+## @param secret_refresh_scatter - boolean - optional
+## @env DD_SECRET_REFRESH_SCATTER - boolean - optional
+##
+## `secret_refresh_scatter`, if set to true, will randomize the first secret refresh. `secret_refresh_interval` needs to be set
+## for this to take effect. This is useful to avoid all agents refreshing secrets at the same time.
+## For more information see: https://docs.datadoghq.com/agent/configuration/secrets-management/#refreshing-apiapp-keys-at-runtime
+##
+# secret_refresh_scatter: <REFRESH_SCATTER>
```
hush-hush: We want to use the default value in the config as example, that way if the line is uncommented the behavior of the agent doesn't change.

```suggestion
# secret_refresh_scatter: true
```

pkg/config/config_template.yaml
```diff
@@ -864,6 +864,24 @@ api_key:
...
+## For more information see: https://docs.datadoghq.com/agent/configuration/secrets-management/#refreshing-apiapp-keys-at-runtime
+##
+#
+# secret_refresh_interval: <REFRESH_INTERVAL>
+
+## @param secret_refresh_scatter - boolean - optional
+## @env DD_SECRET_REFRESH_SCATTER - boolean - optional
+##
+## `secret_refresh_scatter`, if set to true, will randomize the first secret refresh. `secret_refresh_interval` needs to be set
+## for this to take effect. This is useful to avoid all agents refreshing secrets at the same time.
```
hush-hush:
```suggestion
## for this to take effect. This prevent a fleet of Agents from a single deployment to refresh their secrets at the same time.
```

hush-hush (changes requested):

pkg/config/config_template.yaml
rahulkaukuntla: Done.

pkg/config/config_template.yaml
rahulkaukuntla: Understood, I'll keep that in mind for next time. Done.

pkg/config/config_template.yaml
rahulkaukuntla: Done.

pkg/config/config_template.yaml
rahulkaukuntla: Moved the options to after `secret_backend_config`. I think that adding a header would be a nice change to include once we add a dedicated `secret_backend` section (we could have, for example for `secret_backend`, `secrets.secret_backend.config`), so I will not add the header for now.

pkg/config/config_template.yaml
```diff
@@ -888,6 +888,24 @@ api_key:
...
 #   <KEY_1>: <VALUE_1>
 
+## @param secret_refresh_interval - integer - optional
+## @env DD_SECRET_REFRESH_INTERVAL - integer - optional
+##
+## `secret_refresh_interval` is the interval (in seconds) at which api/app key secrets are refreshed.
+## For more information see: https://docs.datadoghq.com/agent/configuration/secrets-management/#refreshing-apiapp-keys-at-runtime
+##
+#
+# secret_refresh_interval: 3600
```
hush-hush: We want the template to reflect the default value from the configuration. That way if a customer uncomment the line it doesn't change the behavior of the agent. The default from the config is `0`.

```suggestion
# secret_refresh_interval: 0
```

pkg/config/config_template.yaml
```diff
@@ -888,6 +888,24 @@ api_key:
 # secret_backend_config:
 #   <KEY_1>: <VALUE_1>
 
+## @param secret_refresh_interval - integer - optional
+## @env DD_SECRET_REFRESH_INTERVAL - integer - optional
+##
+## `secret_refresh_interval` is the interval (in seconds) at which api/app key secrets are refreshed.
+## For more information see: https://docs.datadoghq.com/agent/configuration/secrets-management/#refreshing-apiapp-keys-at-runtime
+##
```
hush-hush:
```suggestion
## @param secret_refresh_interval - integer - optional - default 0
## @env DD_SECRET_REFRESH_INTERVAL - integer - optional - default 0
##
## `secret_refresh_interval` is the interval (in seconds) at which api/app key secrets are refreshed. A 0 value means the feature is disabled.
## For more information see: https://docs.datadoghq.com/agent/configuration/secrets-management/#refreshing-apiapp-keys-at-runtime
##
```

pkg/config/config_template.yaml
```diff
@@ -888,6 +888,24 @@ api_key:
...
+#
+# secret_refresh_interval: 3600
+
+## @param secret_refresh_scatter - boolean - optional
+## @env DD_SECRET_REFRESH_SCATTER - boolean - optional
```
hush-hush:
```suggestion
## @param secret_refresh_scatter - boolean - optional - default true
## @env DD_SECRET_REFRESH_SCATTER - boolean - optional - default true
```

pkg/config/config_template.yaml
```diff
@@ -888,6 +888,24 @@ api_key:
...
+##
+## `secret_refresh_scatter`, if set to true, will randomize the first secret refresh. `secret_refresh_interval` needs to be set
+## for this to take effect. This prevents a fleet of Agents from having a deployment where they refresh their secrets at the same time.
```
hush-hush:
```suggestion
## for this to take effect. This prevents a fleet of Agents from refreshing their secrets at the same time.
```

hush-hush (changes requested):

hush-hush (approved):

PR #39073, opened 2025-07-18
Adding more context to `secret_backend_config` in the config template

rahulkaukuntla:
### What does this PR do?

We want to make it very clear that the purpose of `secret_backend_config` is to provide the information required by `secret_backend_type` to resolve secrets.

### DIFF
--- a/pkg/config/config_template.yaml
+++ b/pkg/config/config_template.yaml
@@ -881,7 +881,8 @@ api_key:
 ## @env DD_SECRET_BACKEND_CONFIG - map - optional
 ## This feature is in preview.
 ##
-## The section contains information required by `secret_backend_type` to resolve secrets.
+## The section contains configuration required by `secret_backend_type` to resolve secrets.
+## The necessary configuration depends upon which type is used.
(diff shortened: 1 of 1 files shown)


pkg/config/config_template.yaml
```diff
@@ -881,7 +881,7 @@ api_key:
 ## @env DD_SECRET_BACKEND_CONFIG - map - optional
 ## This feature is in preview.
 ##
-## The section contains information required to resolved secrets.
+## The section contains information required by `secret_backend_type` to resolve secrets.
```
dustmop: May I suggest a bit more?

```suggestion
## The section contains configuration required by `secret_backend_type` to resolve secrets. The necessary configuration depends upon which type is used.
```

dustmop (approved):
Suggesting a bit more text, and approving as well.

pkg/config/config_template.yaml
rahulkaukuntla: done.

michaelcretzman (approved):

PR #39091, opened 2025-07-21
Reapply "Create custom MacOS temporary directory"

CelianR:
Reapplies https://github.com/DataDog/datadog-agent/pull/39050 fixing macos tests.

The goal is to create an isolated `$TMPDIR`.

We need to create a symlink and set the proper permissions for this to avoid breaking any test. The symlink is within `/tmp` to avoid path length issues.


.gitlab/common/macos.yml
```diff
@@ -16,6 +16,20 @@
       dda inv -- -e macos.remove-inactive-versions -l python -t "$PYTHON_VERSION" || true
       dda inv -- -e macos.remove-inactive-versions -l go -t "$(cat .go-version)" || true
     fi
+  # Create custom temporary folder to isolate jobs from each other
+  # We have to symlink it to /tmp/gitlabci to avoid some path length issues (sockets should be <= 104 characters on MacOS)
+  - |
+    # Remove the trailing slash
+    export OLDTMPDIR="${TMPDIR%/}"
```
rdesgroppes: This relies on `TMPDIR` being set - if it's not you'd get an empty string.

rdesgroppes: To remove any trailing slash (such as `/tmp/` -> `/tmp`), I'd use either:
1. GNU, doesn't follow symlinks:
    - `readlink -m`
    - `readlink --canonicalize-missing`
2. GNU/BSD, follows symlinks:
    - `realpath -m`
    - `realpath --canonicalize-missing`

rdesgroppes: Altogether:
```suggestion
    export OLDTMPDIR="$(realpath -m ${TMPDIR-/tmp})"
```

CelianR: If this isn't set, I exit 1 with an error now. This should always be present in the runners we use. Applied suggestions 👍

.gitlab/common/macos.yml
```diff
@@ -16,6 +16,24 @@
...
+    if [ -z "$TMPDIR" ]; then
+      echo "TMPDIR must be set" >& 2
+      exit 1
+    fi
+    # Remove the trailing slashes etc.
+    export OLDTMPDIR="$(readlink -m "$TMPDIR")"
+    sudo rm -rf "$OLDTMPDIR/gitlabci"
+    NEWTMPDIR="$OLDTMPDIR/gitlabci/$CI_JOB_ID"
+    mkdir -p "$NEWTMPDIR"
+    sudo chmod 777 "$NEWTMPDIR"
```
rdesgroppes: no need to change anything here but, for completeness, `mkdir` supports creation flags such as:

```suggestion
    mkdir -m 777 -p "$NEWTMPDIR"
```

.gitlab/common/macos.yml
```diff
@@ -16,6 +16,24 @@
...
+    sudo chown "$USER":staff "$NEWTMPDIR"
+    sudo unlink /tmp/gitlabci 2> /dev/null || true
+    sudo ln -s "$NEWTMPDIR" /tmp/gitlabci
+    export TMPDIR="/tmp/gitlabci/"
```
rdesgroppes:
```suggestion
    export TMPDIR="/tmp/gitlabci"
```
(the semantics of the directory is conveyed by the variable name, plus derived paths such as below `$TMPDIR/datadog-agent-build/bin` would expand to `/tmp/gitlabci//datadog-agent-build/bin`)

rdesgroppes (approved):
OK

.gitlab/common/macos.yml
CelianR: This ends with a slash by default, removed slash to see if jobs are still properly behaving

.gitlab/common/macos.yml
rdesgroppes: @CelianR sorry to gave misled you as it appears macOS still ships with an old BSD `readlink`:
```
readlink: illegal option -- m
```

CelianR: I used realpath instead

rdesgroppes: I guess you'll face the same problem as `-m` started as a GNU extension that BSD didn't have when Darwin got forked out of it. Let's see.

CelianR: Removed the `-m` option since the temporary directory should exist already

CelianR: It seems to work fine like this 👍

.gitlab/source_test/macos.yml
```diff
@@ -10,6 +10,11 @@ include:
   rules:
     - !reference [.except_disable_unit_tests]
     - !reference [.fast_on_dev_branch_only]
+    - changes:
+        paths:
+          - .gitlab/common/macos.yml
+        compare_to: $COMPARE_TO_BRANCH
+      when: on_success
```
KevinFairise2: What do you want to achieve with that?
The `.fast_on_dev_branch_only` should already make sure that the tests are executed by default. So adding this should not change anything in the behavior of the job

CelianR: It was to run this on macos yaml file change but I can remove it

KevinFairise2: Yeah without it it is already running always except when unit tests are explicitly disabled

KevinFairise2 (approved):

PR #39093, opened 2025-07-21
fix(compare_to): Improve cleaning of created pipelines

chouetz:
### What does this PR do?
- Improve the cleanup of `compare_to` branches created during the `compare_to` test
- Fix insights raised by datadog static analysis

### Motivation
Technical debt

Unit tests and checked the pipeline executions created by this PR


spencergilbert (approved):

tasks/pipeline.py
```diff
@@ -856,24 +856,26 @@ def compare_to_itself(ctx):
     for attempt in range(max_attempts):
         print(f"[{datetime.now()}] Waiting 30s for the pipelines to be created {attempt + 1}/{max_attempts}")
         time.sleep(30)
-        pipelines = agent.pipelines.list(ref=new_branch, get_all=True)
-        for pipeline in pipelines:
+        for pipeline in agent.pipelines.list(ref=new_branch, get_all=True):
```
rdesgroppes, line 859: Unrelated to the change, but I'd however recommend `iterator=True` over `get_all=True` esp. when there's no need to have the full list of pipelines in memory (potentially large + I/O time), which is the case when used as an input iterator within a for loop:
```suggestion
        for pipeline in agent.pipelines.list(ref=new_branch, iterator=True):
```
(there are several occurrences in the present file)

ofek (approved):

tasks/unit_tests/pipeline_tests.py
```diff
@@ -29,8 +29,7 @@ class TestCompareToItself(unittest.TestCase):
     def side(x):
         if x == "c0mm1t":
             return MagicMock(author_name=pipeline.BOT_NAME, title="Commit to compare to itself")
-        else:
-            return MagicMock(author_name="Aimee Jaquet")
+        return MagicMock(author_name="Aimee Jaquet")
```
rdesgroppes, line 32: [😆](https://en.wikipedia.org/wiki/Aim%C3%A9_Jacquet)

rdesgroppes (approved):

tasks/pipeline.py
chouetz: I created a dedicated card for this https://datadoghq.atlassian.net/browse/ACIX-970

PR #39097, opened 2025-07-21
[CHAOSPLT-820] Add fuzz to slack channel selection

edznux-dd:
### What does this PR do?

This adds support for automatic selecting the correct slack channel to report the fuzz finding in.

### Motivation

The current reports of the fuzzing infra are currently sent to Error Tracking, but it's out the workflow of most team working on the agent.  Sending the crash report + enriched data to the engineers directly via slack will help getting notified of bugs.

I created a new slack channel for a "fallback" mechanism.


tasks/fuzz_infra.py
```diff
@@ -8,6 +8,43 @@
...
+
+    Args:
+        directory_path: The directory path to find owners for
+
+    Returns:
+        The Slack channel for the first owner, or default channel if no owner found
+    """
+    try:
+        # Remove the leading datadog-agent prefix if it exists
+        rel_path = directory_path.removeprefix("/go/src/github.com/DataDog/datadog-agent/")
```
CelianR, line 37: Should we assert that the path is either relative or has this prefix?

tasks/fuzz_infra.py
```diff
@@ -8,6 +8,43 @@
...
+
+        # Take the first owner, we assume the first one is enough.
+        # The api currently only supports one slack channel per fuzz target. If need be we could change this.
+        first_owner = owners[0].lower()
+
+        # Map the owner to a slack channel
+        return GITHUB_SLACK_MAP.get(first_owner, DEFAULT_FUZZING_SLACK_CHANNEL)
+
+    except Exception as e:
+        print(f"Warning: Could not determine slack channel for {directory_path}: {e}")
```
CelianR: Could you replace this and add the proper imports?

```suggestion
        print(f"{color_message('Warning', Color.YELLOW)}: Could not determine slack channel for {directory_path}: {e}", file=sys.stderr)
```

CelianR (approved):

tasks/fuzz_infra.py
edznux-dd: Cool thanks!

tasks/fuzz_infra.py
edznux-dd: Added 
```
        # Assert that the path is either relative or had the expected prefix
        assert (
            not directory_path.startswith('/') or
            directory_path.startswith("/go/src/github.com/DataDog/datadog-agent/")
        ), f"Expected relative path or path starting with '/go/src/github.com/DataDog/datadog-agent/', got: {directory_path}"
```

Why not indeed! 
thanks

PR #39099, opened 2025-07-21
Fix curl and wget retries for installer install-script

KevinFairise2:
### What does this PR do?

Update the way we retry curl to fetch the installer, to rely on `--output` or `-O` flags, otherwise the output of the first failure is included inside the saved file, which makes it invalid and fails the integrity check.

### Motivation

Make the retry feature working

Manually tested on a VM that the script was working even if upstream fails on the first run
+ E2E tests


CelianR (approved):

BaptisteFoy (approved):
Let's make sure it works in the tests but LGTM, thanks 🙏

PR #39101, opened 2025-07-21
feat(e2e coverage): Pass the E2E_COVERAGE_PIPELINE env variable to windows build job

KevinFairise2:
### What does this PR do?

Update Windows build job to pass the `E2E_COVERAGE_PIPELINE` environment variable to the build job.
Make sure that Windows jobs also build the coverage instrumented version of the agent.

### DIFF
--- a/.gitlab/package_build/windows.yml
+++ b/.gitlab/package_build/windows.yml
@@ -38,6 +38,7 @@
       -e AGENT_FLAVOR=${AGENT_FLAVOR}
       -e OMNIBUS_RUBY_VERSION="${OMNIBUS_RUBY_VERSION}"
       -e PYTHONUTF8=1
+      -e E2E_COVERAGE_PIPELINE=${E2E_COVERAGE_PIPELINE}
       ${WINBUILDIMAGE}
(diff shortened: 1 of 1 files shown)


rdesgroppes (approved):

.gitlab/package_build/windows.yml
```diff
@@ -38,6 +38,7 @@
       -e AGENT_FLAVOR=${AGENT_FLAVOR}
       -e OMNIBUS_RUBY_VERSION="${OMNIBUS_RUBY_VERSION}"
       -e PYTHONUTF8=1
+      -e E2E_COVERAGE_PIPELINE=${E2E_COVERAGE_PIPELINE}
```
julien-lebot, line 41: **❓ question:** ‏ Would that increase the size of the Windows binaries? Does this flag affect all builds or can it be turned on or off?

julien-lebot (approved):

.gitlab/package_build/windows.yml
KevinFairise2: It will probably make the binary a bit bigger. But this is meant to only be used on a nightly pipelines dedicated to computing the code coverage of our e2e tests. So will not impact at all the real binary that we ship/deploy

PR #39115, opened 2025-07-21
dyninst/dymdb: add function name parse failure reason

andreimatei:
Make it more explicit why parsing a function name failed, for testing.

https://datadoghq.atlassian.net/browse/DEBUG-3890

### DIFF
--- a/pkg/dyninst/symdb/func_name.go
+++ b/pkg/dyninst/symdb/func_name.go
@@ -55,6 +55,20 @@ var (
 	parseAnonFuncInsideInlinedFuncRE = regexp.MustCompile(`^((?P<pkg>(.*/)?.*?)\.)\w+(\.\w+)+\.(func)?(\d)+$`)
 )
 
+type parseFuncNameFailureReason int
+
+const (
+	parseFuncNameFailureReasonUndefined parseFuncNameFailureReason = iota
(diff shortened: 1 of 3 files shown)


pkg/dyninst/symdb/func_name.go
```diff
@@ -67,14 +81,16 @@ type funcName struct {
...
-	// GenericFunction is set if the function takes type arguments. We don't
-	// support parsing these functions at the moment, so no other fields except
-	// QualifiedName are set.
-	GenericFunction bool
+	// ParseErrorReason is set if the function name was not be parsed because
+	// the function is not supported. Such functions should be ignored.
+	//
+	// If ParseErrorReason != 0, then only QualifiedName is set; other fields
+	// are left empty.
+	ParseErrorReason parseFuncNameErrorReason
```
ajwerner: I find it odd that this is called `ParseErrorReason` but it's only populated when no error is returned. What if you instead created an error type (and a set of constant instances for all the reasons) and returned those errors. Then the caller can filter the failures accordingly. I find this middle ground to be bizarre. 

I also really dislike this unexported type on a public field.

andreimatei: moved to a different result type

ajwerner (approved):
LGTM

PR #39119, opened 2025-07-21
pkg/dyninst: swiss map support, decoder specific types, and a bug fix

grantseltzer:
### What does this PR do?

- Adds support for swiss maps in Go DI
- Adds a notion of decoder-specific ir.Type equivalents.
```
// decoderType is a decoder-specific representation of an ir.Type
// It is used to so that specific types can implement their own
// encoding methods. We can track these types in the decoder as a
// way of caching type-specific information such as map key and
// value types.
type decoderType interface {
...
```
- A bug fix in codegen for an issue with offset calculations in array/slices.

### Motivation

We need map support in Go DI. The decoder specific types are an optimization.

Integration tests


pkg/dyninst/json_redaction_test.go
```diff
@@ -115,3 +116,268 @@ func redactJSON(t *testing.T, input []byte, redactors []jsonRedactor) (redacted
 	}
 	return bytes.TrimSpace(buf.Bytes())
 }
+
+type sorter interface {
```
ajwerner: I feel like the `replacer` interface already covers this completely. Also, this log all feels too complicated. If you really care about the notCapturedReason coming last I handled that

Consider:

```go
type entriesSorter struct{}

func boolToInt(b bool) int {
    if b {
        return 1
    }
    return 0
}

var notCapturedReason = []byte(`"notCapturedReason"`)

func compareNotCapturedReason(a, b []byte) int {
    return cmp.Compare(
        boolToInt(bytes.Contains(a, notCapturedReason)),
        boolToInt(bytes.Contains(b, notCapturedReason)),
     )
}

func sortEntries(input []byte) ([]byte, error) {
    var entries [][2]json.RawValue
    if err := json.Unmarshal([]byte(v), &entries); err != nil {
        return nil, err
    }
    slices.SortFunc(entries, func(a, b [2]json.RawValue) int {
        return cmp.Or(
            compareNotCapturedReason(a[0], b[0]),
            bytes.Compare(a[0], b[0])
         )
    })
    return json.Marshal(entries)
}


func (entriesSorter) replace(v jsontext.Value) jsontext.Value {
   replaced, err := sortEntries(v)
   if err != nil {
       replaced, _ = json.Marshal(err.Error())
   }
   return jsontext.Value(replaced)
}
```

pkg/dyninst/json_redaction_test.go
```diff
@@ -115,3 +116,268 @@ func redactJSON(t *testing.T, input []byte, redactors []jsonRedactor) (redacted
...
+	sort(jsontext.Value) jsontext.Value
+}
+
+type entriesSorter struct {
+	ascending bool
```
ajwerner: why does this need to be configurable?

pkg/dyninst/ebpf/stack_machine.h
```diff
@@ -982,53 +984,57 @@ static long sm_loop(__maybe_unused unsigned long i, void* _ctx) {
...
-  //   }
-  //   target_ptr_t data = *(target_ptr_t*)&((*buf)[sm->buf_offset_0]);
+    sm->buf_offset_0 = sm->offset + sm_read_program_uint8(sm);
 
-  //   sm->buf_offset_0 = sm->offset + sm_read_program_uint8(sm);
-  //   if (!scratch_buf_bounds_check(&sm->buf_offset_0, sizeof(int64_t))) {
-  //     return 1;
-  //   }
-  //   uint64_t length_mask = *(uint64_t*)&((*buf)[sm->buf_offset_0]);
+    LOG(1, "Offset diff: %d", sm->buf_offset_0-sm->offset);
```
ajwerner: this feels like it should be higher than level 1. At least 3, maybe 4 or 5?

pkg/dyninst/ebpf/stack_machine.h
```diff
@@ -982,53 +984,57 @@ static long sm_loop(__maybe_unused unsigned long i, void* _ctx) {
...
-  //                          group_byte_len * (length_mask + 1))) {
-  //     LOG(3, "enqueue: failed swiss map groups record");
-  //   }
-  // } break;
+    sm->buf_offset_0 = sm->offset + sm_read_program_uint8(sm);
+    if (!scratch_buf_bounds_check(&sm->buf_offset_0, sizeof(int64_t))) {
+      return 1;
+    }
+    uint64_t length_mask = *(uint64_t*)&((*buf)[sm->buf_offset_0]);
+    LOG(1, "group_slice_type: %d, data: 0x%llx, length_mask: %llu", group_slice_type, data, length_mask);
```
ajwerner: Also definitely at least 3

pkg/dyninst/integration_test.go
```diff
@@ -294,13 +296,14 @@ func testDyninst(
...
 		probeID := probe.GetID()
 		probeRet := retMap[probeID]
 		expIdx := len(probeRet)
-		retMap[probeID] = append(retMap[probeID], json.RawMessage(redacted))
+		retMap[probeID] = append(retMap[probeID], json.RawMessage(sorted))
 		if expIdx < len(expOut[probeID]) {
 			outputToCompare := expOut[probeID][expIdx]
-			assert.JSONEq(t, string(outputToCompare), string(redacted))
+			assert.JSONEq(t, string(outputToCompare), string(sorted))
 		}
```
ajwerner: I feel like I meant to remove these `JSONEq` calls. The below branch checks them in a way that produces a better diff. I can do it in a separate PR if you don't want to.

pkg/dyninst/decode/marshal.go
```diff
@@ -486,9 +491,46 @@ func (d *Decoder) encodeValueFields(
 		}
 		return nil
 	case *ir.GoMapType:
-		return writeTokens(enc,
-			jsontext.String("notCapturedReason"),
-			jsontext.String("unimplemented"),
+		if len(data) < int(v.GetByteSize()) {
+			return errors.New("passed data not long enough for pointer")
```
ajwerner: nit: use a different message for the map?

pkg/dyninst/decode/marshal.go
```diff
@@ -486,9 +491,46 @@ func (d *Decoder) encodeValueFields(
...
+				return err
+			}
+			return nil
+		}
+		var address uint64
+		pointedValue, dataItemExists := dataItems[key]
+		if !dataItemExists {
+			return writeTokens(enc,
+				jsontext.String("notCapturedReason"),
+				jsontext.String("depth"), //TODO: can we distinguish if it's depth or ran out of buffer space?
```
ajwerner: nit: this is a general problem everywhere we use this message, right? Maybe instead of copying this to another place, let's centralize this TODO at the top of the file?

pkg/dyninst/decode/marshal.go
```diff
@@ -486,9 +491,46 @@ func (d *Decoder) encodeValueFields(
...
+			)
+		}
+		address = pointedValue.Header().Address
+		headerDataItem := dataItems[typeAndAddr{
+			irType: uint32(v.HeaderType.GetID()),
+			addr:   address,
+		}]
```
ajwerner: Err, isn't `pointedValue` the same thing as `headerDataItem`? Seems like you're doing the same map lookup twice.

pkg/dyninst/decode/marshal.go
```diff
@@ -501,10 +543,99 @@ func (d *Decoder) encodeValueFields(
...
+		for _, f := range v.Fields {
+			switch f.Name {
+			case "dirLen":
+				dirLen = int64(binary.NativeEndian.Uint64(data[f.Offset : f.Offset+f.Type.GetByteSize()]))
+			case "dirPtr":
+				dirPtr = binary.NativeEndian.Uint64(data[f.Offset : f.Offset+f.Type.GetByteSize()])
+			}
+		}
+
+		slotsField := v.GroupType.Fields[1]
```
ajwerner: The mix of searching for field name and then using hard-coded field indexes is pretty confusing.

pkg/dyninst/decode/marshal.go
```diff
@@ -501,10 +543,99 @@ func (d *Decoder) encodeValueFields(
...
+				tablePtrSliceDataItem,
+				keyType,
+				valueType,
+			)
+			if err != nil {
+				return err
+			}
+		}
+		if err := writeTokens(enc,
+			jsontext.EndArray); err != nil {
```
ajwerner: nit: this is an odd wrapping. Also this can all just be `return writeTokens(env, jsontext.EndArray)`

pkg/dyninst/decode/marshal.go
```diff
@@ -501,10 +543,99 @@ func (d *Decoder) encodeValueFields(
...
+		keyType := d.program.Types[keyField.Type.GetID()]
+
+		valueField := noalgstructType.Fields[1]
+		valueType := d.program.Types[valueField.Type.GetID()]
+
+		if err := writeTokens(enc,
+			jsontext.String("entries"),
+			jsontext.BeginArray); err != nil {
+			return err
+		}
```
ajwerner: extreme nit: I think the following wrapping looks nicer:
```go
if err := writeTokens(
    enc, jsontext.String("entries"), jsontext.BeginArray,
); err != nil {
    return err
}
```

pkg/dyninst/decode/marshal.go
```diff
@@ -639,3 +770,149 @@ func writeTokens(enc *jsontext.Encoder, tokens ...jsontext.Token) error {
 	}
 	return nil
 }
+
+// collectSwissMapTables collects the data items for the swiss map tables.
+// It is called for each table pointer in the table pointer slice.
```
ajwerner: > It is called for each table pointer in the table pointer slice

Is it really?

pkg/dyninst/decode/marshal.go
```diff
@@ -639,3 +770,149 @@ func writeTokens(enc *jsontext.Encoder, tokens ...jsontext.Token) error {
 	}
 	return nil
 }
+
+// collectSwissMapTables collects the data items for the swiss map tables.
```
ajwerner: I don't think calling this method `collect*` makes sense given it doesn't actually collect anything, it just encodes directly.

pkg/dyninst/decode/marshal.go
```diff
@@ -639,3 +770,149 @@ func writeTokens(enc *jsontext.Encoder, tokens ...jsontext.Token) error {
...
+	// Deduplicate addrs by sorting and then removing duplicates.
+	// Go swiss maps may have multiple table pointers for the same group.
+	slices.Sort(addrs)
+	dedupedAddrs := addrs[:0]
+	for i, addr := range addrs {
+		if i == 0 || addr != addrs[i-1] {
+			dedupedAddrs = append(dedupedAddrs, addr)
+		}
+	}
+	addrs = dedupedAddrs
```
ajwerner: This is just:
```go
addrs = slices.Compact(addrs)
```

See https://pkg.go.dev/slices#Compact

pkg/dyninst/ebpf/stack_machine.h
```diff
@@ -295,6 +296,7 @@ sm_record_pointer(global_ctx_t* ctx, type_t type, target_ptr_t addr,
     item = pointers_queue_push_front(&ctx->stack_machine->pointers_queue);
   }
   if (item == NULL) {
+    LOG(1, "sm_record_pointer: pointers queue push failed");
```
ajwerner: maybe 2 or 3?

pkg/dyninst/ebpf/stack_machine.h
```diff
@@ -52,6 +52,7 @@ static bool chased_pointers_push(chased_pointers_t* chased, target_ptr_t ptr,
   }
   uint32_t i = chased->n;
   if (i >= MAX_CHASED_POINTERS) { // to please the verifier
+    LOG(1, "chased_pointers_push: pointers buffer exhausted");
```
ajwerner: maybe 2 or 3?

pkg/dyninst/ebpf/stack_machine.h
```diff
@@ -982,53 +984,57 @@ static long sm_loop(__maybe_unused unsigned long i, void* _ctx) {
...
-  //   int64_t dir_len = *(int64_t*)&((*buf)[sm->buf_offset_1]);
+    if (!scratch_buf_bounds_check(&sm->buf_offset_0, sizeof(target_ptr_t))) {
+      return 1;
+    }
+    target_ptr_t dir_ptr = *(target_ptr_t*)&((*buf)[sm->buf_offset_0]);
+    if (!scratch_buf_bounds_check(&sm->buf_offset_1, sizeof(int64_t))) {
+      return 1;
+    }
+    int64_t dir_len = *(int64_t*)&((*buf)[sm->buf_offset_1]);
+    LOG(1, "type: %d, dir_ptr: 0x%llx, dir_len: %lld", group_type, dir_ptr, dir_len)
```
ajwerner: 4 or 5?

pkg/dyninst/ebpf/stack_machine.h
```diff
@@ -982,53 +984,57 @@ static long sm_loop(__maybe_unused unsigned long i, void* _ctx) {
...
-  //   type_t table_ptr_slice_type = (type_t)sm_read_program_uint32(sm);
-  //   type_t group_type = (type_t)sm_read_program_uint32(sm);
-  //   sm->buf_offset_0 = sm->offset + sm_read_program_uint8(sm);
-  //   sm->buf_offset_1 = sm->offset + sm_read_program_uint8(sm);
+  case SM_OP_PROCESS_GO_SWISS_MAP: {
+    type_t table_ptr_slice_type = (type_t)sm_read_program_uint32(sm);
+    type_t group_type = (type_t)sm_read_program_uint32(sm);
+    sm->buf_offset_0 = sm->offset + sm_read_program_uint8(sm);
+    sm->buf_offset_1 = sm->offset + sm_read_program_uint8(sm);
+    LOG(1, "offset: %d", sm->buf_offset_1-sm->offset);
```
ajwerner: 4 or 5?

pkg/dyninst/compiler/generate.go
```diff
@@ -262,7 +262,7 @@ func (g *generator) addTypeHandler(t ir.Type) (FunctionID, bool, error) {
 			CallOp{
 				FunctionID: elemFunc,
 			},
-			ProcessSliceDataRepeatOp{ElemByteLen: t.Element.GetByteSize()},
+			ProcessSliceDataRepeatOp{ElemByteLen: t.Element.GetByteSize() - g.typeFuncMetadata[t.Element.GetID()].offsetShift},
```
ajwerner, line 266: let's add slice and array tests for this bug too

pkg/dyninst/decode/types.go
```diff
@@ -0,0 +1,801 @@
...
+	_ decoderType = (*goHMapBucketType)(nil)
+	_ decoderType = (*goSwissMapGroupsType)(nil)
+	_ decoderType = (*goChannelType)(nil)
+	_ decoderType = (*goEmptyInterfaceType)(nil)
+	_ decoderType = (*goInterfaceType)(nil)
+	_ decoderType = (*goSubroutineType)(nil)
+	_ decoderType = (*eventRootType)(nil)
+)
+
+func (d *Decoder) getDecoderType(irType ir.Type) (decoderType, error) {
```
ajwerner: I think by storing the decoder types directly you've potentially created some problems for yourself. For one, you can end up allocating more than one decoderType for a given type. You can defeat that problem by memoizing and updating the map internally here, but that runs into problems due to cyclic types. Consider:

`type t []*t`

then if you are passed the input type as `*t` then you'll infinitely recur. Consider https://go.dev/play/p/mCVLcTpTnHh

ajwerner: The answer I'd go with would be to just store type ids in your decoder types and do the hashmap lookup

pkg/dyninst/decode/types.go
```diff
@@ -0,0 +1,801 @@
...
+	var err error
+	if err = writeTokens(enc,
+		jsontext.String("elements"),
+		jsontext.BeginArray); err != nil {
+		return err
+	}
+	sliceLength := int(sliceDataItem.Header().Length) / elementSize
+	sliceData := sliceDataItem.Data()
+	for i := range int(sliceLength) {
+		elementData := sliceData[i*int(s.elementDecoderType.irType().GetByteSize()) : (i+1)*int(s.elementDecoderType.irType().GetByteSize())]
```
ajwerner: pull these calls out of the loop -- there's no reason to keep calling them in the loop

pkg/dyninst/decode/types.go
```diff
@@ -0,0 +1,801 @@
...
+	enc *jsontext.Encoder,
+	dataItems map[typeAndAddr]output.DataItem,
+	currentlyEncoding map[typeAndAddr]struct{},
+	data []byte,
+	valueType string) error {
+
+	if len(data) < int(s.ByteSize) {
+		return writeTokens(enc,
+			jsontext.String("notCapturedReason"),
+			jsontext.String("no buffer space"),
```
ajwerner: why are some of these reasons multiple words and some camelCase? Doesn't have to be right here right now but can we standardize on something and define that something symbolically?

pkg/dyninst/decode/types.go
```diff
@@ -0,0 +1,801 @@
...
+// value types.
+type decoderType interface {
+	irType() ir.Type
+	encodeValueFields(
+		decoder *Decoder,
+		enc *jsontext.Encoder,
+		dataItems map[typeAndAddr]output.DataItem,
+		currentlyEncoding map[typeAndAddr]struct{},
+		data []byte,
+		valueType string) error
```
ajwerner: as far as I can tell, nothing uses this `valueType` string.

pkg/dyninst/decode/types.go
```diff
@@ -0,0 +1,801 @@
...
+}
+
+func (b *baseType) irType() ir.Type { return (*ir.BaseType)(b) }
+func (b *baseType) encodeValueFields(
+	decoder *Decoder,
+	enc *jsontext.Encoder,
+	dataItems map[typeAndAddr]output.DataItem,
+	currentlyEncoding map[typeAndAddr]struct{},
+	data []byte,
+	valueType string) error {
```
ajwerner: nit: I rather dislike this wrapping. I much prefer:
```go
func (b *baseType) encodeValueFields(
	decoder *Decoder,
	enc *jsontext.Encoder,
	dataItems map[typeAndAddr]output.DataItem,
	currentlyEncoding map[typeAndAddr]struct{},
	data []byte,
	valueType string,
) error {
```

pkg/dyninst/decode/types.go
```diff
@@ -0,0 +1,801 @@
...
+type goChannelType ir.GoChannelType
+type goEmptyInterfaceType ir.GoEmptyInterfaceType
+type goInterfaceType ir.GoInterfaceType
+type goSubroutineType ir.GoSubroutineType
+type eventRootType ir.EventRootType
+
+// Compile-time type assertions
+var (
+	_ decoderType = (*baseType)(nil)
+	_ decoderType = (*baseType)(nil)
```
ajwerner: duplicated

pkg/dyninst/decode/marshal.go
```diff
@@ -219,456 +223,10 @@ func (d *Decoder) encodeValue(
...
-			jsontext.String("notCapturedReason"),
-			jsontext.String("unimplemented"),
-		)
-	}
-	return fmt.Errorf("invalid type %s (ID: %d)", irType.GetName(), irType.GetID())
-}
-
-func encodeBaseTypeValue(enc *jsontext.Encoder, irType *ir.BaseType, data []byte) error {
-	kind, ok := irType.GetGoKind()
+func encodeBaseTypeValue(enc *jsontext.Encoder, decodeType *baseType, data []byte) error {
```
ajwerner: this could just be `t *ir.BaseType` or `t *baseType` (fun that the methods work either way due to embedding...). `decodeType` is not adding value here.

pkg/dyninst/decode/types.go
```diff
@@ -0,0 +1,801 @@
...
+		)
+	}
+	length := binary.NativeEndian.Uint64(data[8:16])
+	if length == 0 {
+		return writeTokens(enc,
+			jsontext.String("elements"),
+			jsontext.BeginArray,
+			jsontext.EndArray)
+	}
+	elementSize := int(s.elementDecoderType.irType().GetByteSize())
```
ajwerner: no reason to call `irType()` more than once per call

pkg/dyninst/decode/types.go
```diff
@@ -0,0 +1,801 @@
...
+		return nil, fmt.Errorf("unknown type %s", irType.GetName())
+	}
+}
+
+func (b *baseType) irType() ir.Type { return (*ir.BaseType)(b) }
+func (b *baseType) encodeValueFields(
+	decoder *Decoder,
+	enc *jsontext.Encoder,
+	dataItems map[typeAndAddr]output.DataItem,
+	currentlyEncoding map[typeAndAddr]struct{},
```
ajwerner: It seems to me like all 4 of these fields are always the same for every call. I think we should do:

```
type decoderContext struct {
    d *Decoder
    enc *jsontext.Encoder
    dataItems map[typeAndAddr]output.DataItem
    currentlyEncoding map[typeAndAddr]struct{}
}
type decoderType interface {
	irType() ir.Type
	encodeValueFields(*decoderContext, data []byte) error
}

func encodeValue(
	ctx *decoderContext,
	decoderType decoderType,
	data []byte,
	valueType string,
) error {
```

We should be able to store this in the decoder, or, alternatively, just unify the two things

pkg/dyninst/decode/marshal.go
```diff
@@ -430,6 +435,11 @@ func (d *Decoder) collectSwissMapGroup(
 		slotsData   []byte
 	)
 	for _, groupField := range s.GroupType.Fields {
+		fieldEnd := groupField.Offset + groupField.Type.GetByteSize()
+		if fieldEnd > uint32(len(groupData)) {
+			return fmt.Errorf("group field %s extends beyond data bounds: need %d bytes, have %d", groupField.Name, fieldEnd, len(groupData))
+		}
```
ajwerner: these sorts of offsets feel like the things you should be pre-computing

pkg/dyninst/decode/types.go
```diff
@@ -566,7 +581,15 @@ func (a *arrayType) encodeValueFields(
 	}
 
 	for i := range numElements {
-		elementData := data[i*elementSize : (i+1)*elementSize]
+		offset := i * elementSize
+		endIdx := offset + elementSize
+		if endIdx < offset {
+			return fmt.Errorf("overflow in array element %d offset calculation", i)
+		}
```
ajwerner: this one seems a bit much

ajwerner (commented):
Here's a first pass.

pkg/dyninst/decode/marshal.go
grantseltzer: I'll just make a JIRA ticket if it's possible to distinguish (I believe it is based on the size of data)

pkg/dyninst/decode/types.go
grantseltzer: The ones that are camel cased are the values the front end looks for in order to display them a certain way. Otherwise they display them as strings. Agreed I'll move this to consts or something.

pkg/dyninst/decode/marshal.go
```diff
@@ -205,14 +205,14 @@ func (d *Decoder) encodeValue(
...
-		d,
-		enc,
-		dataItems,
-		currentlyEncoding,
+		&decoderContext{
+			decoder:           d,
+			enc:               enc,
+			dataItems:         dataItems,
+			currentlyEncoding: currentlyEncoding,
+		},
```
ajwerner: This misses the point of `decoderContext` as I suggested it -- it was to make it so we don't have to pass so many arguments to `encodeValue` and elsewhere, and to avoid putting so many arguments on the stack for each recursive call. Let's just stop passing `dataItems` and `currentlyEncoding` anywhere you have a `*Decoder` and use the values already in the decoder (you'll need to add a reusable map for `currentlyEncoding` and while you're at it, rename the confusingly named `addressReferenceCount` map. 

I think I'd also put the output encoder into the `Decoder` to simplify all the signatures. Maybe it makes sense to put all these fields that get reset per call into a sub-struct so that the resetting logic is isolated.

pkg/dyninst/decode/types.go
```diff
@@ -58,8 +58,16 @@ type goHMapHeaderType ir.GoHMapHeaderType
...
+	keyTypeID    uint32
+	valueTypeID  uint32
+	dirLenOffset uint32
+	dirLenSize   uint32
+	dirPtrOffset uint32
+	dirPtrSize   uint32
+	ctrlOffset   uint32
+	ctrlSize     uint32
+	slotsOffset  uint32
+	slotsSize    uint32
```
ajwerner: uber-nit: these sizes can all be uint8 and can be grouped to make this smaller

pkg/dyninst/decode/types.go
```diff
@@ -206,203 +223,160 @@ func (d *Decoder) getDecoderType(irType ir.Type) (decoderType, error) {
...
 			return fmt.Errorf("table ptr slice data item pointer not found for addr %x", dirPtr)
 		}
-		tablePtrSliceDataItem, ok := dataItems[typeAndAddr{
+		tablePtrSliceDataItem, ok := ctx.dataItems[typeAndAddr{
 			irType: tablePtrSliceDataItemPtr.Header().Type,
 			addr:   tablePtrSliceDataItemPtr.Header().Address,
 		}]
 		if !ok {
 			return fmt.Errorf("table ptr slice data item not found for addr %x", tablePtrSliceDataItemPtr.Header().Address)
 		}
```
ajwerner: In general, I don't think it's sound to fail out if we're missing a pointer -- we might have run out of space.

pkg/dyninst/decode/types.go
```diff
@@ -206,203 +223,160 @@ func (d *Decoder) getDecoderType(irType ir.Type) (decoderType, error) {
...
 		return err
 	}
 	if dirLen == 0 {
 		// This is a 'small' swiss map where there's only one group.
 		// We can collect the data item for the group directly.
-		groupDataItem := dataItems[typeAndAddr{
+		groupDataItem := ctx.dataItems[typeAndAddr{
 			irType: uint32(s.GroupType.GetID()),
 			addr:   dirPtr,
 		}]
```
ajwerner: I think you need to check whether this succeeded (and not fail if it didn't)

pkg/dyninst/decode/types.go
```diff
@@ -206,203 +223,160 @@ func (d *Decoder) getDecoderType(irType ir.Type) (decoderType, error) {
...
 	} else {
 		// This is a 'large' swiss map where there are multiple groups of data/control words
 		// We need to collect the data items for the table pointers first.
-		tablePtrSliceDataItemPtr, ok := dataItems[typeAndAddr{
+		tablePtrSliceDataItemPtr, ok := ctx.dataItems[typeAndAddr{
 			irType: uint32(s.TablePtrSliceType.GetID()),
 			addr:   dirPtr,
 		}]
 		if !ok {
 			return fmt.Errorf("table ptr slice data item pointer not found for addr %x", dirPtr)
```
ajwerner: In general, I don't think it's sound to fail out if we're missing a pointer -- we might have run out of space.

pkg/dyninst/decode/types.go
```diff
@@ -206,203 +223,160 @@ func (d *Decoder) getDecoderType(irType ir.Type) (decoderType, error) {
...
 			s,
 			tablePtrSliceDataItem,
 		)
 		if err != nil {
 			return err
 		}
 	}
-	if err := writeTokens(enc,
+	if err := writeTokens(ctx.enc,
 		jsontext.EndArray); err != nil {
```
ajwerner: nit: this wrapping. Also it's the last statement

pkg/dyninst/decode/types.go
```diff
@@ -513,66 +478,57 @@ func (p *pointerType) encodeValueFields(
...
+			ctx.currentlyEncoding,
 			fieldType,
 			data[field.Offset:field.Offset+field.Type.GetByteSize()],
 			field.Type.GetName(),
 		); err != nil {
 			return err
 		}
 	}
-	if err = writeTokens(enc, jsontext.EndObject); err != nil {
+	if err = writeTokens(ctx.enc, jsontext.EndObject); err != nil {
```
ajwerner: nit: last statement

pkg/dyninst/decode/types.go
```diff
@@ -583,101 +539,99 @@ func (a *arrayType) encodeValueFields(
...
 			elementDecoderType.irType().GetName(),
 		); err != nil {
 			return err
 		}
 	}
-	if err = writeTokens(enc, jsontext.EndArray); err != nil {
+	if err = writeTokens(ctx.enc, jsontext.EndArray); err != nil {
 		return err
 	}
 	return nil
```
ajwerner: nit: last statement

pkg/dyninst/decode/types.go
```diff
@@ -583,101 +539,99 @@ func (a *arrayType) encodeValueFields(
...
-		return writeTokens(enc,
+		return writeTokens(ctx.enc,
 			jsontext.String("elements"),
 			jsontext.BeginArray,
 			jsontext.EndArray)
 	}
-	elementSize := int(s.elementDecoderType.irType().GetByteSize())
+	var elementDecoderType decoderType
+	var err error
+	elementDecoderType, err = ctx.decoder.getDecoderType(s.Data.Element)
```
ajwerner: useless predeclaration

pkg/dyninst/decode/types.go
```diff
@@ -687,148 +641,136 @@ func (s *goSliceHeaderType) encodeValueFields(
...
+	if err := writeTokens(ctx.enc,
 		jsontext.String("value"),
 	); err != nil {
 		return err
 	}
 	str := unsafe.String(unsafe.SliceData(stringData), int(length))
-	if err := writeTokens(enc, jsontext.String(str)); err != nil {
+	if err := writeTokens(ctx.enc, jsontext.String(str)); err != nil {
 		return err
 	}
```
ajwerner: nit: last statement

pkg/dyninst/decode/types.go
```diff
@@ -687,148 +641,136 @@ func (s *goSliceHeaderType) encodeValueFields(
...
 	stringData := stringValue.Data()
 	length := stringValue.Header().Length
 	if len(stringData) < int(length) {
-		return writeTokens(enc,
+		return writeTokens(ctx.enc,
 			jsontext.String("notCapturedReason"),
 			jsontext.String("no buffer space"),
 		)
 	}
```
ajwerner: I don't think this is the right thing to do when we have a truncated string (can be a separate PR). I think this will happen on literally every long string. We should return the prefix but also make sure we communicate that it was truncated.

pkg/dyninst/decode/types.go
ajwerner: 🤔 seems like you decided instead to always construct and allocate a decoderType on the fly -- and you kept it recursive. You sidestepped the slice-ptr bug, but there's still infinite recursion bugs here: `type t map[int]t`. 

I don't understand why you'd ever called `getDecoderType` in this function particularly given in all the cases where you do that you end up just accessing the ir type...

To make it concrete, I did expect there to be a hash map of decoder types and I expected the constructor to not need to access it. To make that explicitly clear, consider making the constructor:

```
func newDecoderType(ir ir.Type, types map[ir.TypeID]ir.Type) (decoderType, error)
```

I think it'd be ok to make that lazy -- but I wouldn't. I'd eagerly construct it when you construct the decoder.

pkg/dyninst/decode/marshal.go
```diff
@@ -223,10 +223,10 @@ func (d *Decoder) encodeValue(
 	return nil
 }
 
-func encodeBaseTypeValue(enc *jsontext.Encoder, decodeType *baseType, data []byte) error {
-	kind, ok := decodeType.GetGoKind()
+func encodeBaseTypeValue(enc *jsontext.Encoder, t *baseType, data []byte) error {
```
ajwerner: on some level I am unsure as to why this is here and not just in the method on `*baseType`

pkg/dyninst/decode/marshal.go
```diff
@@ -352,13 +352,7 @@ func (d *Decoder) collectSwissMapTables(
...
 	slices.Sort(addrs)
-	dedupedAddrs := addrs[:0]
-	for i, addr := range addrs {
-		if i == 0 || addr != addrs[i-1] {
-			dedupedAddrs = append(dedupedAddrs, addr)
-		}
-	}
-	addrs = dedupedAddrs
+	addrs = slices.Compact(addrs)
 	tableType, ok := s.TablePtrSliceType.Element.(*ir.PointerType)
```
ajwerner: you could and perhaps should precompute this assertion and all the other type assertions. I get how that leads use back to the problem of cycle decoderType references, but I think in practice it's trivial to side-step.

pkg/dyninst/decode/types.go
```diff
@@ -486,24 +451,24 @@ func (p *pointerType) encodeValueFields(
 		return fmt.Errorf("no go kind for type %s (ID: %d)", pointedType.GetName(), pointedType.GetID())
 	}
 	if goKind != reflect.Pointer {
```
ajwerner, line 635: this is the void pointers? can you provide commentary on when this happens? Maybe better to make a different `decoderType` for this case

ajwerner (changes requested):

pkg/dyninst/decode/marshal.go
grantseltzer: I've mostly addressed this. I'm not moving encoder as a decoder field as it'd need to change the signature of the decode method which will need to be updated in various places and it's making this PR too big. I can handle in follow up.

pkg/dyninst/decode/types.go
grantseltzer: Yea I can see why we'd want to just end the array and return nil. This case (and there are others) should have partial capture, but also a notCapturedReason set. I need to confirm that the UI will display this in a meaningful way. Will do so in the morning.

pkg/dyninst/decode/types.go
grantseltzer: Yea leftover from before adding void pointer. Removing the code.

pkg/dyninst/decode/types.go
grantseltzer: length is populated with the truncated string length from bpf. I'm not actually sure what would cause this to occur?

Either way if there's partial strings we should surface it and include a notCapturedReason as already mentioned.

grantseltzer:
I added a change to how pointers are encoded which I believe corrects its format to use "value". I remember having a conversation about this, this seems correct.

pkg/dyninst/actuator/errors.go
```diff
@@ -22,6 +22,10 @@ type NoSuccessfulProbesError struct {
 
 // Error implements the error interface.
 func (e *NoSuccessfulProbesError) Error() string {
+
+	for i := range e.Issues {
+		fmt.Println(e.Issues[i].Message)
+	}
```
ajwerner: ?

pkg/dyninst/decode/decoder.go
```diff
@@ -29,22 +29,24 @@ type probeEvent struct {
...
-	program               *ir.Program
-	stackFrames           map[uint64][]symbol.StackFrame
-	probeEvents           map[ir.TypeID]probeEvent
-	snapshotMessage       snapshotMessage
-	addressReferenceCount map[typeAndAddr]output.DataItem
+	program            *ir.Program
+	stackFrames        map[uint64][]symbol.StackFrame
+	probeEvents        map[ir.TypeID]probeEvent
+	snapshotMessage    snapshotMessage
+	dataItemReferences map[typeAndAddr]output.DataItem
```
ajwerner: nit: can you separate out the fields that are persistent from those that are reset per Decode call?

ajwerner: what about just `dataItems` instead of `dataItemReferences`?

pkg/dyninst/decode/marshal.go
ajwerner: huh still seems like these aren't being precomputed

pkg/dyninst/decode/marshal.go
```diff
@@ -201,441 +195,168 @@ func (d *Decoder) encodeValue(
...
-			return errors.New("passed data not long enough for uint32")
-		}
-		return writeTokens(enc, jsontext.String(strconv.FormatUint(uint64(binary.NativeEndian.Uint32(data)), 10)))
-	case reflect.Uint64:
-		if len(data) != 8 {
-			return errors.New("passed data not long enough for uint64")
+	valueType, ok := d.program.Types[valueTypeID]
+	if !ok {
+		return fmt.Errorf("value type not found for value type ID %d", valueTypeID)
+	}
```
ajwerner: may as well just precompute the byte sizes of these key and value types

pkg/dyninst/decode/marshal.go
ajwerner: In what way did you mostly address this? You still allocate a ton of these decoderContext structure on the fly. It seems wild. I think it'll make the whole PR smaller and relative to the size of the PR as a whole, it's not big. If you want to "handle in a follow up" you could build trust on that by having already implemented it. As it stands this implementation is less clear to read and less efficient, so I'm not inclined to approve it.

pkg/dyninst/decode/marshal.go
```diff
@@ -201,441 +195,168 @@ func (d *Decoder) encodeValue(
...
-		var address uint64
-		pointedValue, dataItemExists := dataItems[key]
-		if !dataItemExists {
-			return writeTokens(enc,
-				jsontext.String("notCapturedReason"),
-				jsontext.String("depth"), //TODO: can we distinguish if it's depth or ran out of buffer space?
-			)
-		}
-		address = pointedValue.Header().Address
+// encodeSwissMapTables collects the data items for the swiss map tables.
```
ajwerner: does it collect? I find this comment confusing

pkg/dyninst/decode/marshal.go
```diff
@@ -201,441 +195,168 @@ func (d *Decoder) encodeValue(
...
-			}
-		}
-		return nil
-	case *ir.StructureType:
-		var err error
-		if err = writeTokens(enc,
-			jsontext.String("fields"),
-			jsontext.BeginObject); err != nil {
-			return err
+			return fmt.Errorf("table data item not found for addr %x", addr)
```
ajwerner: Again, we really cannot bail out on all decoding when a pointer is missing

pkg/dyninst/decode/marshal.go
```diff
@@ -201,441 +195,168 @@ func (d *Decoder) encodeValue(
...
-		elementType := v.Element
-		elementSize := int(elementType.GetByteSize())
-		numElements := int(v.Count)
-		if err = writeTokens(enc,
-			jsontext.String("size"),
-			jsontext.String(strconv.Itoa(numElements)),
-			jsontext.String("elements"),
-			jsontext.BeginArray); err != nil {
+		tableStructType := s.tableStructType
+		groupField, err := getFieldByName(tableStructType.Fields, "groups")
```
ajwerner: Elsewhere you said you addressed pre-computing the fields. I don't see a lot of evidence of that.

pkg/dyninst/decode/types.go
ajwerner: I am not inclined to accept this as it is. Before this patch if there was no map we'd succeed and now if there's a big map we'll completely fail. That's a bad tradeoff. It's not so hard to fix this, is it?

ajwerner (commented):
We should add testing for big maps with lots of element and maps with big elements that we can't capture etc. I am not going to buy the idea that because a PR has been taking a long time or is big is a reason to merge it.

If you want to rework a change in a major way and find that incremental commits are getting to be too much -- fine, feel free to close the PR and put up a new one. I'm happy to review.

pkg/dyninst/actuator/errors.go
grantseltzer: Leftover debugging

ajwerner: ?

pkg/dyninst/compiler/generate.go
```diff
@@ -341,9 +348,47 @@ func (g *generator) addTypeHandler(t ir.Type) (FunctionID, bool, error) {
...
+		lengthMaskOffset, err := offsetOf(t.RawFields, "lengthMask")
+		if err != nil {
+			return nil, false, err
+		}
+		needed = true
+		offsetShift = 0
+		ops = []Op{
+			ProcessGoSwissMapGroupsOp{
+				DataOffset:       uint8(dataOffset), //TODO: check if dataOffset fits in a uint8
+				LengthMaskOffset: uint8(lengthMaskOffset),
```
ajwerner: These TODOs also apply to the `GoSwissMapHeaderType` and are trivial. Why not just do them?

pkg/dyninst/compiler/instructions.go
```diff
@@ -140,18 +140,25 @@ func makeInstruction(op Op) codeFragment {
 		}
 
 	case ProcessGoSwissMapOp:
-		bytes := make([]byte, 0, 8)
+		bytes := make([]byte, 0, 16)
 		bytes = binary.LittleEndian.AppendUint32(bytes, uint32(op.TablePtrSlice.GetID()))
 		bytes = binary.LittleEndian.AppendUint32(bytes, uint32(op.Group.GetID()))
+		bytes = append(bytes, byte(op.DirPtrOffset))
```
ajwerner: does anybody ask for these `byte` casts? I thought `uint8` and `byte` are literally the same type in go. Assuming that's true, having them is actually noise that causes confusion -0- why are we doing the cast?

pkg/dyninst/compiler/instructions.go
```diff
@@ -140,18 +140,25 @@ func makeInstruction(op Op) codeFragment {
...
+		bytes = append(bytes, byte(op.DirLenOffset))
 		return staticInstruction{
 			opcode: OpcodeProcessGoSwissMap,
 			bytes:  bytes,
 		}
 
 	case ProcessGoSwissMapGroupsOp:
+		bytes := make([]byte, 0, 16)
+		bytes = binary.LittleEndian.AppendUint32(bytes, uint32(op.GroupSlice.GetID()))
+		bytes = binary.LittleEndian.AppendUint32(bytes, uint32(op.Group.GetByteSize()))
```
ajwerner: why this uint32 cast?

pkg/dyninst/decode/decoder.go
```diff
@@ -29,22 +29,27 @@ type probeEvent struct {
...
-	snapshotMessage       snapshotMessage
-	addressReferenceCount map[typeAndAddr]output.DataItem
+	program         *ir.Program
+	snapshotMessage snapshotMessage
+
+	decoderTypes      map[ir.TypeID]decoderType
+	probeEvents       map[ir.TypeID]probeEvent
+	stackFrames       map[uint64][]symbol.StackFrame
+	dataItems         map[typeAndAddr]output.DataItem
+	currentlyEncoding map[typeAndAddr]struct{}
```
ajwerner: please separate these fields into what is initialized at construction time and what is reset per message (and use comments to make that clear).

pkg/dyninst/decode/marshal.go
```diff
@@ -201,441 +193,139 @@ func (d *Decoder) encodeValue(
...
+	}
+	if err := decoderType.encodeValueFields(
+		d,
 		enc,
-		dataItems,
-		currentlyEncoding,
-		irType,
 		data,
-		valueType,
 	); err != nil {
```
ajwerner: nit; this wrapping is silly now

pkg/dyninst/decode/marshal.go
```diff
@@ -201,441 +193,139 @@ func (d *Decoder) encodeValue(
...
 		return err
 	}
 	if err := writeTokens(enc, jsontext.EndObject); err != nil {
 		return err
 	}
+	return nil
+}
 
+func writeTokens(enc *jsontext.Encoder, tokens ...jsontext.Token) error {
+	var err error
```
ajwerner: why is this predeclared? I think it's more confusing this way

pkg/dyninst/decode/marshal.go
```diff
@@ -201,441 +193,139 @@ func (d *Decoder) encodeValue(
...
+// encodeSwissMapTables traverses the table pointer slice and encodes the data items for each table.
+func (d *Decoder) encodeSwissMapTables(
 	enc *jsontext.Encoder,
-	dataItems map[typeAndAddr]output.DataItem,
-	currentlyEncoding map[typeAndAddr]struct{},
-	irType ir.Type,
-	data []byte,
-	valueType string,
+	s *goSwissMapHeaderType,
+	tablePtrSliceDataItem output.DataItem,
```
ajwerner: nit: I think this would be cleaner as a method on `*goSwissMapHeaderType` than on `Decoder` (the way we now invoke all the other methods on `decoderType`s and pass in a `Decoder`.

pkg/dyninst/decode/marshal.go
```diff
@@ -201,441 +193,139 @@ func (d *Decoder) encodeValue(
...
+			return fmt.Errorf("table pointer %d extends beyond data bounds: need %d bytes, have %d", i, endIdx, len(tablePointers))
+		}
+		addrs = append(addrs, binary.NativeEndian.Uint64(tablePointers[startIdx:endIdx]))
+	}
+	// Deduplicate addrs by sorting and then removing duplicates.
+	// Go swiss maps may have multiple table pointers for the same group.
+	slices.Sort(addrs)
+	addrs = slices.Compact(addrs)
+	tableType := s.tableType
+	someDataNotCapture := false
```
ajwerner: nit: may as well call it `someDataNotCaptured` to make it read better in english

pkg/dyninst/decode/marshal.go
```diff
@@ -201,441 +193,139 @@ func (d *Decoder) encodeValue(
...
-			irType: uint32(v.Pointee.GetID()),
+	tablePointers := tablePtrSliceDataItem.Data()
+	addrs := []uint64{}
+	for i := range tablePtrSliceDataItem.Header().Length / 8 {
+		startIdx := i * 8
+		endIdx := startIdx + 8
+		if endIdx > uint32(len(tablePointers)) {
+			return fmt.Errorf("table pointer %d extends beyond data bounds: need %d bytes, have %d", i, endIdx, len(tablePointers))
+		}
+		addrs = append(addrs, binary.NativeEndian.Uint64(tablePointers[startIdx:endIdx]))
```
ajwerner, line 228: I think I'd expect there to be some nils here, or at least, there could be. Can you filter them here?

pkg/dyninst/decode/marshal.go
```diff
@@ -201,441 +193,139 @@ func (d *Decoder) encodeValue(
...
+	}
+	// Deduplicate addrs by sorting and then removing duplicates.
+	// Go swiss maps may have multiple table pointers for the same group.
+	slices.Sort(addrs)
+	addrs = slices.Compact(addrs)
+	tableType := s.tableType
+	someDataNotCapture := false
+	for _, addr := range addrs {
+		tableDataItem, ok := d.dataItems[typeAndAddr{
+			irType: uint32(tableType.Pointee.GetID()),
```
ajwerner: nit: pull the `tableTypeID` into a variable above the loop, or alternatively just have tableTypeID in `s`

pkg/dyninst/decode/marshal.go
```diff
@@ -201,441 +193,139 @@ func (d *Decoder) encodeValue(
...
-			)
-		}
-		length := binary.NativeEndian.Uint64(data[8:16])
-		if length == 0 {
-			return writeTokens(enc,
-				jsontext.String("elements"),
-				jsontext.BeginArray,
-				jsontext.EndArray)
+			someDataNotCapture = true
+			log.Tracef("table data item not found for addr %x", addr)
```
ajwerner: I don't know if this is worth leaving in

pkg/dyninst/decode/marshal.go
```diff
@@ -201,441 +193,139 @@ func (d *Decoder) encodeValue(
...
 		}
-		elementType := v.Data.ID
-		elementSize := int(v.Data.Element.GetByteSize())
-		sliceDataItem, ok := dataItems[typeAndAddr{
-			addr:   address,
-			irType: uint32(elementType),
+		groupData := tableDataItem.Data()[s.groupFieldOffset : s.groupFieldOffset+uint32(s.groupFieldSize)]
+		groupAddress := groupData[s.dataFieldOffset : s.dataFieldOffset+uint32(s.dataFieldSize)]
+		groupDataItem, ok := d.dataItems[typeAndAddr{
+			irType: uint32(s.groupType.GroupSliceType.GetID()),
```
ajwerner: probably just pull `groupTypeID` into `s`

pkg/dyninst/decode/marshal.go
```diff
@@ -201,441 +193,139 @@ func (d *Decoder) encodeValue(
...
-			fieldByteSize uint32
-		)
-		for _, field := range v.RawFields {
-			if field.Name == "str" {
-				fieldByteSize = field.Type.GetByteSize()
-				if fieldByteSize != 8 || len(data) < int(field.Offset+fieldByteSize) {
-					return fmt.Errorf("malformed string field 'str': field size %d != 8 or data length %d < required %d", fieldByteSize, len(data), field.Offset+fieldByteSize)
-				}
-				address = binary.NativeEndian.Uint64(data[field.Offset : field.Offset+fieldByteSize])
+		elementType := s.groupType.GroupSliceType.Element
```
ajwerner: nit: probably pull `groupSliceTypeID` and `groupByteSize` into `s` and stop calling all these methods / lookups

pkg/dyninst/decode/marshal.go
```diff
@@ -201,441 +193,139 @@ func (d *Decoder) encodeValue(
...
 		}
-		if address == 0 {
-			return writeTokens(enc,
-				jsontext.String("isNull"),
-				jsontext.Bool(true),
-			)
-		}
+	}
+	if someDataNotCapture {
+		return errors.New("some data not captured")
```
ajwerner: why would this be an error? It's like if you were going to return an error all along, why track this and not just early return?

pkg/dyninst/decode/marshal.go
```diff
@@ -201,441 +193,139 @@ func (d *Decoder) encodeValue(
...
-		return writeTokens(enc,
-			jsontext.String("notCapturedReason"),
-			jsontext.String("unimplemented"),
-		)
-	case *ir.EventRootType:
-		return writeTokens(enc,
-			jsontext.String("notCapturedReason"),
-			jsontext.String("unimplemented"),
+		err = d.encodeValue(enc,
+			s.valueTypeID, valueData, s.valueTypeName,
```
ajwerner: nit: lots of unneeded wrapping (and yet, also lots of very long lines above)

pkg/dyninst/ebpf/stack_machine.h
```diff
@@ -982,53 +983,57 @@ static long sm_loop(__maybe_unused unsigned long i, void* _ctx) {
...
-  //                          group_byte_len * (length_mask + 1))) {
-  //     LOG(3, "enqueue: failed swiss map groups record");
-  //   }
-  // } break;
+    sm->buf_offset_0 = sm->offset + sm_read_program_uint8(sm);
+    if (!scratch_buf_bounds_check(&sm->buf_offset_0, sizeof(int64_t))) {
+      return 1;
+    }
+    uint64_t length_mask = *(uint64_t*)&((*buf)[sm->buf_offset_0]);
+    LOG(3, "group_slice_type: %d, data: 0x%llx, length_mask: %llu", group_slice_type, data, length_mask);
```
ajwerner: this feels more 4 than 3

pkg/dyninst/testdata/decoded/sample/testMapArrayToArray.json
```diff
@@ -0,0 +1,78 @@
...
+                        },
+                        {
+                          "type": "int",
+                          "value": "2"
+                        }
+                      ]
+                    }
+                  ]
+                ],
+                "notCapturedReason": "depth"
```
ajwerner: why is this here? I think I know that it's because there's missing data but it's glaringly obvious that what we're missing here is `"size"`

pkg/dyninst/testdata/decoded/sample/testSingleFloat32.json
```diff
@@ -26,7 +26,7 @@
             "arguments": {
               "x": {
                 "type": "float32",
-                "notCapturedReason": "unavailable"
+                "notCapturedReason": "pruned"
```
ajwerner: Isn't pruned misleading here? It wasn't pruned.

pkg/dyninst/decode/marshal.go
```diff
@@ -116,18 +112,16 @@ func (ad *argumentsData) MarshalJSONTo(enc *jsontext.Encoder) error {
 				jsontext.BeginObject,
 				jsontext.String("type"),
 				jsontext.String(parameterType.GetName()),
-				jsontext.String("notCapturedReason"),
-				jsontext.String("unavailable"),
+				notCapturedReason,
+				notCapturedReasonPruned,
```
ajwerner: Is pruned reasonable here? pruned has a specific meaning and it's not unavailable...

ajwerner (commented):
I think the big missing there here is printing the size of the collection. The rest are mostly nitpicky, but there's some that are legit

pkg/dyninst/decode/types.go
```diff
@@ -0,0 +1,955 @@
...
+			return nil, fmt.Errorf("group field type is not a swiss map groups type: %s", groupField.Type.GetName())
+		}
+		groupFieldOffset := groupField.Offset
+		groupFieldSize := groupField.Type.GetByteSize()
+		dataField, err := getFieldByName(groupType.RawFields, "data")
+		if err != nil {
+			return nil, fmt.Errorf("malformed swiss map header type: %w", err)
+		}
+		dataFieldOffset := dataField.Offset
+		dataFieldSize := dataField.Type.GetByteSize()
```
ajwerner, line 195: you have unchecked size casts to uint8 here, and it's inconsistent why you use some uint32 for offset and some uint8 for sizes. I feel like the overall motivation here was that for some of these we know that they are constant small values and can fit in uint8. It's all fine because they do in fact fit, but it's odd to me that you check sometimes but not others.

pkg/dyninst/decode/types.go
```diff
@@ -0,0 +1,955 @@
...
+	slotsSize        uint8
+	groupFieldOffset uint32
+	groupFieldSize   uint8
+	dataFieldOffset  uint32
+	dataFieldSize    uint8
+	tableTypeID      ir.TypeID
+	groupTypeID      ir.TypeID
+	elementTypeSize  uint32
+	usedOffset       uint32
+	usedSize         uint8
```
ajwerner: nit: this is pretty hard to understand from a distance. It'd be nice to have commentary explaining what these fields are (probably with permalinks to go source code). Imagine coming back to this later!

pkg/dyninst/decode/types.go
```diff
@@ -0,0 +1,955 @@
...
+		}
+		if keyField == nil {
+			return nil, errors.New("type map entry array element has no key field: " + entryArray.Element.GetName())
+		}
+		elem, err := getFieldByName(noalgstructType.RawFields, "elem")
+		if err != nil {
+			return nil, fmt.Errorf("malformed swiss map header type: %w", err)
+		}
+
+		tableType := s.TablePtrSliceType.Element.(*ir.PointerType)
```
ajwerner: should we call this `tablePtrType`? I got quite confused below when `tableTypeID` is not the same as `tableType.GetID()`...

pkg/dyninst/decode/types.go
```diff
@@ -0,0 +1,955 @@
...
+			dirLenOffset:         dirLenOffset,
+			dirLenSize:           uint8(dirLenSize),
+			dirPtrOffset:         dirPtrOffset,
+			dirPtrSize:           uint8(dirPtrSize),
+			ctrlOffset:           ctrlOffset,
+			ctrlSize:             uint8(ctrlSize),
+			slotsOffset:          slotsField.Offset,
+			slotsSize:            uint8(slotsFieldType.GetByteSize()),
+			tableType:            tableType,
+			tableStructType:      tableStructType,
```
ajwerner: yeah, I said this above, but isn't it better to call it `tablePtrType` and `tableType`?

pkg/dyninst/decode/types.go
```diff
@@ -0,0 +1,955 @@
...
+	dataFieldOffset  uint32
+	dataFieldSize    uint8
+	tableTypeID      ir.TypeID
+	groupTypeID      ir.TypeID
+	elementTypeSize  uint32
+	usedOffset       uint32
+	usedSize         uint8
+	tableType        *ir.PointerType
+	tableStructType  *ir.StructureType
+	groupType        *ir.GoSwissMapGroupsType
```
ajwerner: why are these still here? As far as I can tell, nothing uses them

pkg/dyninst/decode/types.go
```diff
@@ -0,0 +1,955 @@
...
+	if err := writeTokens(
+		enc, jsontext.String("entries"), jsontext.BeginArray,
+	); err != nil {
+		return err
+	}
+	if dirLen == 0 {
+		// This is a 'small' swiss map where there's only one group.
+		// We can collect the data item for the group directly.
+		groupDataItem, ok := d.dataItems[typeAndAddr{
+			irType: uint32(s.GroupType.GetID()),
```
ajwerner: you have `groupTypeID`

pkg/dyninst/decode/types.go
```diff
@@ -0,0 +1,955 @@
...
+			notCapturedReason,
+			notCapturedReasonDepth,
+		)
+	}
+	address = pointedValue.Header().Address
+	goKind, ok := p.Pointee.GetGoKind()
+	if !ok {
+		return fmt.Errorf("no go kind for type %s (ID: %d)", p.Pointee.GetName(), p.Pointee.GetID())
+	}
+	if goKind != reflect.Pointer {
```
ajwerner, line 635: If you're taking another pass, maybe worth adding commentary about why this is here

pkg/dyninst/testdata/decoded/sample/testMapArrayToArray.json
```diff
@@ -0,0 +1,84 @@
...
+                        },
+                        {
+                          "type": "int",
+                          "value": "2"
+                        }
+                      ]
+                    }
+                  ],
+                  "notCapturedReason",
+                  "pruned"
```
ajwerner: I think this is in the wrong place. Shouldn't this be parallel to entries, not inside of entries?

pkg/dyninst/testdata/decoded/sample/testMassiveString.json
```diff
@@ -0,0 +1,40 @@
...
+            "method": "main.testMassiveString"
+          }
+        },
+        "captures": {
+          "entry": {
+            "arguments": {
+              "x": {
+                "type": "string",
+                "size": 126417,
+                "notCapturedReason": "pruned",
```
ajwerner: I think the protocol for strings is `"truncated": true` not `"notCapturedReason"`, right?

pkg/dyninst/decode/types.go
```diff
@@ -0,0 +1,955 @@
...
+			notCapturedReasonLength,
+		)
+	}
+	address := binary.NativeEndian.Uint64(data[s.strFieldOffset : s.strFieldOffset+uint32(s.strFieldSize)])
+	if address == 0 {
+		return writeTokens(enc,
+			jsontext.String("isNull"),
+			jsontext.Bool(true),
+		)
+	}
```
ajwerner: I don't think this makes sense. In go an empty string can have a nil address but it's never `nil`.

pkg/dyninst/decode/types.go
```diff
@@ -0,0 +1,955 @@
...
+	d *Decoder,
+	enc *jsontext.Encoder,
+	data []byte,
+) error {
+	realLength := binary.NativeEndian.Uint64(data[s.lenFieldOffset : s.lenFieldOffset+uint32(s.lenFieldSize)])
+	if err := writeTokens(enc,
+		jsontext.String("size"),
+		jsontext.Uint(realLength)); err != nil {
+		return err
+	}
```
ajwerner: This should only be written if the value is truncated. The spec says: 
> When values are capped, an additional field is added to the Snapshot to indicate there is more data. The field truncated indicates that data were truncated because too long:
"truncated": true.  The original length of the string is also indicated by the size field only if truncated

pkg/dyninst/decode/types.go
```diff
@@ -0,0 +1,955 @@
...
+	stringValue, ok := d.dataItems[typeAndAddr{
+		irType: uint32(s.Data.GetID()),
+		addr:   address,
+	}]
+	if !ok {
+		return writeTokens(enc,
+			notCapturedReason,
+			notCapturedReasonDepth,
+		)
+	}
```
ajwerner, line 869: Just so I'm extra clear, if indeed we don't have any data for the string, then yes using notCapturedReason is correct. Only use `"truncated"` if we have part of the string, and in that case, white the size. I think it'd be okay to write the size also in this case.

ajwerner (commented):
Getting close. There's still some issues with where you're outputting the `"notCapturedReason"` for the maps, and some things to fix up with the strings when they are truncated.

pkg/dyninst/testdata/decoded/sample/testMapIntToInt.json
```diff
@@ -0,0 +1,42 @@
...
+          "entry": {
+            "arguments": {
+              "m": {
+                "type": "map[int]int",
+                "size": 10,
+                "entries": [
+                  "notCapturedReason",
+                  "pruned"
+                ]
+              }
```
ajwerner: what's up with this one?

ajwerner (commented):
Can we make sure to have a test with a map where we definitely do not collect some things?

pkg/dyninst/testdata/decoded/sample/testMapIntToInt.json
grantseltzer: It appear the group data isn't present in the event's data items. Investigating.

pkg/dyninst/decode/types.go
```diff
@@ -0,0 +1,949 @@
...
+			// Fields related to user defined key and value types
+			keyTypeID:     keyField.Type.GetID(),
+			valueTypeID:   elem.Type.GetID(),
+			keyTypeSize:   keyField.Type.GetByteSize(),
+			valueTypeSize: elem.Type.GetByteSize(),
+			keyTypeName:   keyField.Type.GetName(),
+			valueTypeName: elem.Type.GetName(),
+
+			// Fields in go swiss map internal representation
+			// Seehttps://github.com/golang/go/blob/cd3655a8243b5f52b6a274a0aba5e01d998906c0/src/internal/runtime/maps/map.go#L195
```
ajwerner, line 214: nit: add a space between `See` and the url. Also 8 chars is enough on permalinks and is more pleasant in terms of line length

pkg/dyninst/decode/marshal.go
```diff
@@ -201,441 +187,111 @@ func (d *Decoder) encodeValue(
...
-		if len(data) != 8 {
-			return errors.New("passed data not long enough for uintptr")
+func (s *goSwissMapHeaderType) encodeSwissMapGroup(
+	d *Decoder,
+	enc *jsontext.Encoder,
+	groupData []byte,
+) (valuesEncoded int, err error) {
+	slotsData := groupData[s.slotsOffset : s.slotsOffset+s.slotsSize]
+	controlWord := binary.LittleEndian.Uint64(groupData[s.ctrlOffset : s.ctrlOffset+uint32(s.ctrlSize)])
+	entrySize := (s.valueFieldOffset - s.keyFieldOffset) + s.valueTypeSize
```
ajwerner, line 272: this isn't right sadly. Consider a `bool` value. You'll run into the same problem. You need to record the slot size explicitly too. Add a test with bool values while you're at it and put a few entries in that map

pkg/dyninst/decode/types.go
```diff
@@ -0,0 +1,960 @@
...
+		usedOffset := usedField.Offset
+		usedSize := usedField.Type.GetByteSize()
+
+		slotsField, err := getFieldByName(s.GroupType.RawFields, "slots")
+		if err != nil {
+			return nil, fmt.Errorf("malformed swiss map header type: %w", err)
+		}
+		slotsFieldType, ok := types[slotsField.Type.GetID()]
+		if !ok {
+			return nil, errors.New("type map slot field not found in types: " + s.GroupType.Name)
```
ajwerner, line 151: why not just `fmt.Errorf` here and elsewhere? I think of errors.New as really for static strings and don't see benefit for things like this but it's not a blocking comment

pkg/dyninst/decode/types.go
```diff
@@ -0,0 +1,960 @@
...
+
+func newDecoderType(
+	irType ir.Type,
+	types map[ir.TypeID]ir.Type,
+) (decoderType, error) {
+	switch s := irType.(type) {
+	case *ir.GoSwissMapHeaderType:
+		dirPtrField, err := getFieldByName(s.RawFields, "dirPtr")
+		if err != nil {
+			return nil, fmt.Errorf("malformed swiss map header type: %w", err)
```
ajwerner, line 127: the repetition of the string `"malformed swiss map header type: "` makes me think you should extract a helper that returns a concrete type and an error and wrap it if it's non-nil...

ajwerner (commented):
This should be the last round. Just gotta make sure you're moving through the slots array correctly.

ajwerner (approved):
I've changed my mind. You can merge it and fast-follow with the last round of issues.

PR #39131, opened 2025-07-21
check agent user password before remote update

clarkb7:
### What does this PR do?
during a remote update, checks that the Agent user password is available (if required) and stops the update before it uninstalls the Agent.

This is only applicable to domain joined hosts using a custom Agent username and password. See [Agent 7.66.0 upgrade notes](https://github.com/DataDog/datadog-agent/releases/tag/7.66.0) for more information.

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1649

Using the MSI to upgrade from a version before remote updates was introduced (7.66.0) without providing the active directory account password will cause remote updates to fail because the password is not available. This failure leaves the host without an Agent installed because the same error prevents us from reinstalling the previous Agent version.

Example MSI log error
```
MSI (s) (40!48) [22:39:04:336]: Product: Datadog Agent -- A password was not provided. Passwords are required for domain accounts.
```

Add unit tests that exercise some limited functionality
E2E tests
- `TestUpgradeWithMissingPassword` - new, ensures the error case is caught
- `TestAgentUpgradesOnDCWithGMSA$/TestUpgradeAgentPackage$` - tests upgrade with gMSA accounts
- `TestAgentUpgradesOnDC$/TestUpgradeAgentPackage$` - tests upgrade with domain account
- `TestAgentUpgrades$/TestUpgradeAgentPackage$` - tests upgrade with local account

Add manual test script `remote-tests.ps1` to aid in devloops. Runs the new code/manual-tests on a host via SSH so it can be quickly tested in the different environments (local host, domain account, gMSA account).

On our laptops where the Agent is installed with a local user, the manual tests 
(run as admin)
```powershell
cd pkg/fleet/installer/packages/user/windows
go test . -v -tags manualtest,test
```
example (`NetIsServiceAccount` fails on non-domain hosts)
```
=== RUN   TestAgentUser
user:  LAPTOP-J4KNN9ON\ddagentuser
domain:  LAPTOP-J4KNN9ON
sid:  S-1-5-21-1313994865-2680462364-1677150189-1004
password in LSA: true
is local account: true
    user_manual_test.go:61:
                Error Trace:    C:/Users/branden.clark/projects/datadog-agent/pkg/fleet/installer/packages/user/windows/user_manual_test.go:61
                Error:          Received unexpected error:
                                failed to check if account 'LAPTOP-J4KNN9ON\ddagentuser' is a service account: The RPC server is unavailable. Please ensure the netlogon service is running and the domain controller is available
                Test:           TestAgentUser
is service account: false
```

### Additional Notes

The MSI has historically allowed users to update (major upgrade) without re-providing the Agent user password. The MSI does this by not deleting the services ([overwrites ](https://github.com/DataDog/datadog-agent/blob/a2349f366113fe74a613d8a9fca7e8127e15a88d/tools/windows/DatadogAgentInstaller/WixSetup/Datadog%20Agent/AgentInstaller.cs#L334-L338)the built-in action `DeleteService` condition) between upgrades. Though this can cause errors if an upgrade introduces a new service but doesn't have the password. Now that the password can be fetched from LSA this should be less of an issue.


git-thuerk-done (approved):
lgtm

BaptisteFoy (approved):

pkg/fleet/installer/packages/user/windows/lsa.c
```diff
@@ -0,0 +1,118 @@
...
+
+    // Open LSA policy
+    status = LsaOpenPolicy(NULL, &object_attributes, POLICY_GET_PRIVATE_INFORMATION, &lsa_handle);
+    if (status != STATUS_SUCCESS) {
+        goto done;
+    }
+
+    // Create LSA unicode string for the key
+    lsa_key_name.Buffer = key_copy;
+    lsa_key_name.Length = wcslen(key_copy) * sizeof(WCHAR);
```
jack0x2, line 58:
```suggestion
    // Specifies the length, in bytes, of the string pointed to by the Buffer member, not including the terminating null character, if any
    // https://learn.microsoft.com/en-us/windows/win32/api/lsalookup/ns-lsalookup-lsa_unicode_string
    lsa_key_name.Length = wcslen(key_copy) * sizeof(WCHAR);
```

jack0x2: Might not be 100% necessary, but these always get screwed up

jack0x2 (approved):

CelianR (approved):

L3n41c (approved):
👍🏻 LGTM for the single file owned by @DataDog/container-integrations.

PR #39142, opened 2025-07-22
[EBPF] gpu: add extra test for cgroupv2 code

gjulianm:
### What does this PR do?

Adds an extra test for the cgroupv2 patching code in GPU monitoring.

### Motivation

Extra test coverage.

The change is a test.


pkg/gpu/cgroups_test.go
```diff
@@ -276,18 +279,105 @@ func TestGetAbsoluteCgroupForProcess(t *testing.T) {
...
-	currentCgroup, err = getAbsoluteCgroupForProcess("", uint32(os.Getpid()))
+	currentCgroup, err = getAbsoluteCgroupForProcess("", "/", uint32(os.Getpid()), uint32(os.Getpid()), containerdcgroups.Mode())
 	require.NoError(t, err)
 	require.Equal(t, "/"+testCgroupName, currentCgroup)
 }
 
+func TestGetAbsoluteCgroupV2ForProcessInsideContainer(t *testing.T) {
+	if os.Geteuid() != 0 {
+		t.Skip("Test requires root privileges for the bind mount")
+	}
```
val06, line 340: does it mean that it won't be executed in the CI de-facto? can we introduce some mocks to overcome the root privileges requirement?

pkg/gpu/cgroups_test.go
```diff
@@ -276,18 +279,105 @@ func TestGetAbsoluteCgroupForProcess(t *testing.T) {
...
+
+	// Our target cgroup is a child cgroup too, so create a nested hierarchy
+	parentCgroupName := fmt.Sprintf("test-parent-cgroup-%s", utils.RandString(10))
+
+	// Create our child cgroup, using a bind mount so the inode is the same as the cgroup directory for the parent
+	childCgroupName := fmt.Sprintf("test-child-cgroup-%s", utils.RandString(10))
+	childCgroupPath := filepath.Join(parentCgroupName, childCgroupName)
+	childCgroupFullPath := filepath.Join(hostCgroupFs, childCgroupPath)
+	require.NoError(t, os.MkdirAll(childCgroupFullPath, 0755))
+	err := unix.Mount(containerCgroupFs, childCgroupFullPath, "bind", unix.MS_BIND, "")
```
val06: can you move that into a helper function that "creates" the fake env?

val06 (commented):
left comments

pkg/gpu/cgroups_test.go
gjulianm: It runs in KMT only, not in the `test_{deb,rpm,...}` jobs.

val06: and we run it with root privileges in KMT?

gjulianm: Yes, always, and without containers

pkg/gpu/cgroups_test.go
gjulianm: I'm not sure if in this case it's very helpful, as I'm using some of the internal variables for validation and I don't think this code can be easily reused

val06: I've made an attempt [here](https://github.com/DataDog/datadog-agent/pull/39496)

gjulianm: But I don't see what we gain there. It's a separate function that cannot be reused directly because it's made for this specific test in mind, and if we need changes we now have to add things to another struct to share them. I was thinking of doing a more generic fake-cgroup-fs system, but only once we need it for more tests and we have a better idea of how to approach it.

val06 (approved):

pkg/gpu/cgroups_test.go
gjulianm: Discussed offline, ended up moving the setup code to use a generic fake cgroupfs

PR #39156, opened 2025-07-22
Re-introduce `quality_gate_metrics_logs` with limits set to stops

blt:
### What does this PR do?

This commit re-introduces the `quality_gate_metrics_logs` Quality Gate experiment with the limits set to allotments. We have previously set limits on new experiments "loose" by observation with the previous release. If runtime behavior has shifted since the most recent release we may find that the limit set is too tight, as happened with this experiment. With checks set to allotments we are mechanically guaranteed that this experiment will not signal failure.

### Additional Notes

Last night's run suggests that the proper limit on CPU is 3.2 cores, see [checks board](https://app.datadoghq.com/dashboard/vz3-jd5-bdi?fromUser=false&refresh_mode=paused&tpl_var_experiment%5B0%5D=quality_gate_metrics_logs&tpl_var_job_id%5B0%5D=9bb90645-472b-4fdd-872b-9d3e3b544671&view=spans&from_ts=1753171727000&to_ts=1753172327000&live=false).

REF PR #39140
REF PR #38534


test/regression/cases/quality_gate_metrics_logs/experiment.yaml
```diff
@@ -0,0 +1,55 @@
...
+    description: "Memory usage quality gate. This puts a bound on the total collector memory usage."
+    bounds:
+      series: total_pss_bytes
+      upper_bound: 1.625 GiB
+
+  - name: cpu_usage
+    description: "CPU usage quality gate. This puts a bound on the total average collector millicore usage."
+    bounds:
+      series: avg(total_cpu_usage_millicores)
+      upper_bound: 2500
```
preinlein: is this supposed to be 2500 or 3200?

scottopell (approved):

test/regression/cases/quality_gate_metrics_logs/experiment.yaml
blt: Bad append on my part. Good looking out.

PR #39164, opened 2025-07-22
dyninst/symdb: add snapshot testing

andreimatei:
This patch adds a simple text serialization format for symbols and
creates golden files for one of the test programs.

The symdb cli tool now outputs the data.

### DIFF
--- a/pkg/dyninst/symdb/cli/main.go
+++ b/pkg/dyninst/symdb/cli/main.go
@@ -15,20 +15,28 @@ import (
 	"net/http"
 	_ "net/http/pprof"
 	"os"
+	"strings"
+	"time"
 
-	"github.com/DataDog/datadog-agent/pkg/dyninst/object"
 	"github.com/DataDog/datadog-agent/pkg/dyninst/symdb"
+	"github.com/DataDog/datadog-agent/pkg/dyninst/symdb/symdbutil"
 )
 
 var (
-	pprofPort  = flag.Int("pprof-port", 8081, "Port for pprof server.")
-	binaryPath = flag.String("binary-path", "", "Path to the binary to analyze.")
+	pprofPort      = flag.Int("pprof-port", 8081, "Port for pprof server.")
+	binaryPath     = flag.String("binary-path", "", "Path to the binary to analyze.")
+	mainModuleOnly = flag.Bool("main-module-only", false, "If set, only symbols from the module containing the \"main\" function are printed.")
+	silent         = flag.Bool("silent", false, "If set, the collected symbols are not printed.")
 )
 
 func main() {
 	flag.Parse()
 	if *binaryPath == "" {
-		fmt.Println("Usage: symdbcli --binary-path <path-to-binary> [--pprof-port <port>]")
+		fmt.Print(`Usage: symdbcli --binary-path <path-to-binary> [--main-module-only] [--silent]
+
+The symbols from the specified binary will be extracted and printed to stdout
+(unless --silent is specified).
+`)
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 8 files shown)


pkg/dyninst/symdb/cli/main.go
```diff
@@ -88,5 +90,18 @@ func statsFromSymbols(s symdb.Symbols) symbolStats {
 			}
 		}
 	}
+
+	if *outputFile != "" {
```
piob-io: Wouldn't it be useful to just print symbols on stdout? Do we need this option? Couldn't the user pipe the output however they want, including redirecting to /dev/null to silence?

pkg/dyninst/symdb/symdb_test.go
```diff
@@ -50,6 +52,51 @@ func TestSymDB(t *testing.T) {
 	}
 }
 
+var rewrite = flag.Bool("rewrite", false, "rewrite the snapshot files")
```
piob-io: Let's be consistent with all other snapshots tests we have - we use "REWRITE" env var instead.

pkg/dyninst/symdb/testdata/snapshot/sample.arch=amd64,toolchain=go1.24.3.out
```diff
@@ -0,0 +1,531 @@
...
+		Arg: p: [3]uint32 (declared at line 90, available: )
+		Arg: q: [3]uint32 (declared at line 90, available: )
+		Arg: r: [3]uint32 (declared at line 90, available: )
+		Arg: s: [3]uint32 (declared at line 90, available: )
+		Arg: t: [3]uint32 (declared at line 90, available: )
+		Arg: u: [3]uint32 (declared at line 90, available: )
+	Function: testVeryLargeArray (main.testVeryLargeArray) in /home/andrei/go/src/github.com/DataDog/datadog-agent/pkg/dyninst/testprogs/progs/sample/arrays.go [95:95]
+		Arg: a: [100]uint (declared at line 95, available: )
+	Function: executeArrayFuncs (main.executeArrayFuncs) in /home/andrei/go/src/github.com/DataDog/datadog-agent/pkg/dyninst/testprogs/progs/sample/arrays.go [98:123]
+	Function: testSingleByte (main.testSingleByte) in /home/andrei/go/src/github.com/DataDog/datadog-agent/pkg/dyninst/testprogs/progs/sample/basics.go [11:11]
```
piob-io: To make snapshots test work, we need to strip the machine-specific prefix.

piob-io (commented):
Neat!

pkg/dyninst/symdb/symdb_test.go
andreimatei: done. But who wants the env var?

pkg/dyninst/symdb/cli/main.go
andreimatei: I considered that, but I feel like this tool will be used more for benchmarking, where you don't want the output (but you maybe want some line about those stats that it currently prints).
Also this tool will also learn to produce the SymDB JSON soon, so we'll need to decide on a way to opt for one type of output or another.
I'd say let's start with this --out flag...

pkg/dyninst/symdb/testdata/snapshot/sample.arch=amd64,toolchain=go1.24.3.out
andreimatei: done

pkg/dyninst/symdb/symdb_test.go
piob-io: No strong opinion, except that all snapshot tests should have the same way to control rewrite mode :shrug:

pkg/dyninst/symdb/cli/main.go
piob-io: Then I'd suggest to use a flag that more directly configures what you want - "-s" "--silence" to enable/disable producing the output, so that it doesn't mess with being able to redirect output.

andreimatei: done

piob-io (approved):

PR #39168, opened 2025-07-22
CAP-2711 Add arn tags to ecs task

stzou:
### What does this PR do?
Adds `cluster_arn`, `service_arn`, and `task_definition_arn` tags to ECS tasks.

### Motivation
There is an ongoing effort to to improve the tagging experience with ECS. The addition on these `arn` tags will allow the ecs explorer to better correlate with other products when using an agent for task collection.

1. Spin up an ECS cluster
2. Deploy the image `datadog/agent-dev:stzou-cap-2711-full` using both fargate and ec2 launch types
3. Confirm that metrics, logs, and tasks on ecs explorer sent from both fargate and ec2 are properly tagged with `cluster_arn`, `service_arn`, `task_definition_arn`.


buraizu (approved):

comp/core/tagger/tags/tags.go
```diff
@@ -142,8 +142,12 @@ const (
 	EcsContainerName = "ecs_container_name"
 	// EcsClusterName is the tag for the ECS cluster name
 	EcsClusterName = "ecs_cluster_name"
+	// EcsClusterARN is the tag for the ECS cluster ARN (Amazon Resource Name)
+	EcsClusterARN = "cluster_arn"
```
gabedos, line 146: Nit: could we name these 2 new tags `ecs_cluster_arn` and `ecs_service_arn` for standardization between all the other `ecs` prefixed tags. We want to avoid the use of tags generally prefixed `cluster_` in favor of more context specific tags like `kube_cluster_`, `ecs_cluster_`, `kafka_cluster_`, etc.

comp/core/tagger/tags/tags.go
```diff
@@ -142,8 +142,12 @@ const (
...
 	// EcsServiceName is the tag for the ECS service name
 	EcsServiceName = "ecs_service"
+	// EcsServiceARN is the tag for the ECS service ARN (Amazon Resource Name)
```
gabedos, line 149: Nit reminder: could you track a follow up task to expose the tags in the public documentation. Thanks! https://docs.datadoghq.com/containers/amazon_ecs/tags/

comp/core/workloadmeta/collectors/util/ecs_util.go
```diff
@@ -46,12 +47,18 @@ func ParseV4Task(task v3or4.Task, seen map[workloadmeta.EntityID]struct{}) []wor
 	taskContainers, containerEvents := ParseV4TaskContainers(task, seen)
 	region, awsAccountID := ParseRegionAndAWSAccountID(task.TaskARN)
 
+	clusterName := parseClusterName(task.ClusterName)
+	clusterARN := BuildClusterARN(clusterName, awsAccountID, region)
+	serviceARN := BuildServiceARN(clusterName, task.ServiceName, awsAccountID, region)
+	taskDefinitionARN := BuildTaskDefinitionARN(awsAccountID, task.Family, region, task.Version)
+
```
gabedos, line 54: Curious: in the PR description and during biweekly discussions, we talked about how the ARNs would be useful for cross product features in the ECS explorer. However, these new tags appear to be a composition of our existing old tags. Since all the individual sub-tags exist on the telemetry, it is not possible to make a multi arg query to locate related entities? 

For example, if we want other telemetry for a specific service, you can use `ecs_cluster_name:x AND ecs_service_name:y AND awsAccountID:z AND region:a AND task_version:b` to find any other data like logs, metrics, traces, etc for that specific workload.

comp/core/tagger/tags/tags.go
stzou: Yep, created PR here: https://github.com/DataDog/documentation/pull/30663

comp/core/workloadmeta/collectors/util/ecs_util.go
stzou: Yes, we could use the sub-tags to build out a query and that is what we're currently doing, but we found that this approach can get messy due to the different tagging methods. For example, the logs from the agent include `ecs_service` whereas fluenbit uses `service`. Instead, we wanted to rely on a single tag for correlation.

comp/core/tagger/tags/tags.go
stzou: The reason we want to use `cluster_arn` and `service_arn` is because this is the current naming convention being used for ecs arn tags in our product. While I agree that tags prefixed with `cluster_` may not be specific enough, I believe `arn` are specific enough where we don't need to add the `ecs_` prefix.

comp/core/workloadmeta/collectors/util/ecs_util.go
gabedos: Gotcha, understand that it can simplify / optimize the ECS explorer side. Should be fine adding these extra tags.

Side note, fluentbit doesn't get the tags from the Agent correct? so on ECS Fargate you won't have `cluster_arn` and `service_arn` and will still need to have some legacy compatibility with the `service` tag right? I think it could be possible to request agent-logs-pipelines to make an update to the Datadog fluentbit forwarder if you want to standardize further as well.

comp/core/tagger/tags/tags.go
gabedos: Agree that the ARN does implicitly suggest that we are working on an AWS product. The only potential conflict would be with EKS and their clusterARN but we currently have no product specifically focused on that. We can keep the tag already used in the ECS product

gabedos (approved):

comp/core/workloadmeta/collectors/util/ecs_util.go
stzou: That is correct, we've reached out to the agent logs team and put in a request to get the fluentbit output plugin updated with these tags as well

sblumenthal (approved):

PR #39173, opened 2025-07-22
Add missing aws ena metric collection via ethtool

jeremy-hanna:
### What does this PR do?

The python network check collects [AWS Enhanced Network](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/monitoring-network-performance-ena.html#network-performance-metrics) metrics (https://github.com/DataDog/integrations-core/pull/8331) from the ethtool interface output, this was missing in the original migration PR so we're adding it here along with some e2e tests of this configuration


pkg/collector/corechecks/net/networkv2/network.go
```diff
@@ -866,6 +907,10 @@ func (c *NetworkCheck) Configure(senderManager sender.SenderManager, _ uint64, r
 		}
 	}
 
+	if c.config.instance.CollectEthtoolMetrics || c.config.instance.CollectEnaMetrics {
```
jeremy-hanna, line 910: https://github.com/DataDog/integrations-core/blob/master/network/datadog_checks/network/network.py#L66

collect_ethtool_stats is really the union of these two configuration keys

pkg/collector/corechecks/net/networkv2/network.go
```diff
@@ -295,24 +305,55 @@ func handleEthtoolStats(sender sender.Sender, interfaceIO net.IOCountersStat) er
...
+			"device:" + interfaceIO.Name,
 			"driver_name:" + driverName,
 			"driver_version:" + driverVersion,
-			extraTag,
 		}
 
-		for metricName, metricValue := range keyValuePairing {
+		count := 0
+		for metricName, metricValue := range enaMetrics {
 			metricName := fmt.Sprintf("system.net.%s", metricName)
```
jeremy-hanna, line 318: the full prefix here is eg. `system.net.aws.ec2.bw_in_allowance_exceeded` for the ENA metrics

pkg/collector/corechecks/net/networkv2/network.go
```diff
@@ -295,24 +305,55 @@ func handleEthtoolStats(sender sender.Sender, interfaceIO net.IOCountersStat) er
 		}
 	}
 
-	processedMap := getEthtoolMetrics(driverName, statsMap)
-	for extraTag, keyValuePairing := range processedMap {
+	if collectEnaMetrics {
+		enaMetrics := getEnaMetrics(driverName, statsMap)
 		tags := []string{
-			"interface:" + interfaceIO.Name,
+			"device:" + interfaceIO.Name,
```
jeremy-hanna, line 311: correcting this tag: https://github.com/DataDog/integrations-core/blob/master/network/datadog_checks/network/check_linux.py#L462

louis-cqrl (approved):

PR #39184, opened 2025-07-23
[FA] Populate fleet_layers

coignetp:
### What does this PR do?

Populate `fleet_layers` and `config_id` from the mergeConfigs
This field existed before, so we're re-using the pipeline

### Additional Notes

Manual QA


pkg/fleet/installer/installer.go
```diff
@@ -995,5 +999,41 @@ func mergeConfigs(rawConfigs [][]byte) (map[string]configFile, error) {
...
+		datadogFile = configFile{
+			Path:     "/datadog.yaml",
+			Action:   configFileActionAdd,
+			Contents: json.RawMessage(`{"fleet_layers": []]}`),
+		}
+	}
+
+	// Add fleet_layers configuration
+	datadogConfig["fleet_layers"] = configOrder
+	datadogConfig["config_id"] = configOrder[len(configOrder)-1]
```
BaptisteFoy: nit: i'm a bit worried that we're losing info here, maybe a simple join would be better?

BaptisteFoy (approved):

BaptisteFoy (approved):

clarkb7 (approved):

PR #39185, opened 2025-07-23
[AGENTRUN-585] Add new hostname metrics

louis-cqrl:
### What does this PR do?

- Introduces two new Agent telemetry metrics under the `hostname` namespace:
  1. A **counter** `drift_detected` with labels `state`, `provider`, and `cloud_provider`, which increments whenever the Agent detects a change in hostname or provider.
  2. A **histogram** `drift_resolution_time_ms` (same labels) to record the time taken (in milliseconds) to resolve a hostname during each check.
- Implements a background scheduler that:
  - Waits an initial delay (default 20 minutes) after Agent startup, then performs the first hostname drift check.
  - Continues to perform recurring checks at a configurable interval (default 12 hours) thereafter.
- Hooks the scheduler into the main `getHostname` flow so that, once the Agent has determined and saved its initial hostname, it will continue to monitor and emit telemetry on any subsequent drift.

### Motivation

Hostname “drift”, when the Agent’s notion of its own hostname or cloud provider changes over time, can lead to data fragmentation, duplicate metrics, and troubleshooting complexity. By adding these telemetry metrics and a periodic drift checker, we can:

- Proactively surface drift events in our monitoring dashboards.
- Measure how long it takes the Agent to re-resolve its hostname after each drift.

Tested on a VM

- Added 2 globals for the metrics which is not the best state, we should at some point remove them to use telemetry component, however it adds a lot of overhead, thus this task is delayed

### Additional Notes

- The new metrics use the existing telemetry framework, so no external dependencies are added.  
- The scheduler is excluded from serverless builds (`//go:build !serverless`).  
- Configuration knobs `DefaultInitialDelay` and `DefaultRecurringInterval` can be overridden in future Agent config if different timing is desired.


releasenotes/notes/add-hostname-metric-COAT-6ca8603db0466a34.yaml
```diff
@@ -0,0 +1,12 @@
...
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+other:
+  - |
+    Add a new metric counter to the Agent telemetry for hostname.drift_detected.
+    This will help to observe if there is any issue with the hostname resolution during the Agent lifetime.
```
estherk15:
```suggestion
    Add a metric counter to the Agent telemetry for hostname.drift_detected.
    This helps track whether hostname resolution issues occur during the Agent's lifetime.
```

estherk15 (approved):

pkg/util/hostname/drift.go
```diff
@@ -0,0 +1,132 @@
...
+	DefaultRecurringInterval = 6 * time.Hour
+)
+
+// Telemetry metrics
+var (
+	tlmDriftDetected = telemetry.NewCounter("hostname", "drift_detected",
+		[]string{"state", "provider"}, "Hostname drift detection status")
+	tlmDriftResolutionTime = telemetry.NewHistogram("hostname", "drift_resolution_time_ms",
+		[]string{"state", "provider"}, "Hostname drift resolution time in milliseconds", []float64{.5, 1, 2.5, 5, 10, 60})
+)
```
GustavoCaso: Could we avoid using globals for telemetry metrics and instead use the telemetry component? 😄

GustavoCaso: If the investment on removing the use of globals is too big for this PR, let's note in the PR description and see if we can work on it on a follow-up PR

pkg/util/hostname/providers.go
```diff
@@ -138,8 +140,8 @@ var (
 // * container (kube_apiserver, Docker, kubelet)
 // * OS hostname
 // * EC2
-func getProviderCatalog(legacyHostnameResolution bool) []provider {
-	providerCatalog := []provider{
+func GetProviderCatalog(legacyHostnameResolution bool) []Provider {
```
GustavoCaso: Why are we making the function and the struct public?

pkg/util/hostname/providers.go
```diff
@@ -227,7 +229,9 @@ func getHostname(ctx context.Context, keyCache string, legacyHostnameResolution
 	warnAboutFQDN(ctx, hostname)
 
 	if hostname != "" {
-		return saveHostname(cacheHostnameKey, hostname, providerName, legacyHostnameResolution), nil
+		hostnameData := saveHostname(cacheHostnameKey, hostname, providerName, legacyHostnameResolution)
+		scheduleHostnameDriftChecks(ctx, hostnameData)
```
GustavoCaso: The `scheduleHostnameDriftChecks` is new logic that is added to a crucial piece of functionality of the Agent, and there is some involved logic. I would add tests ensuring that the metrics are being emitted

GustavoCaso (changes requested):
@louis-cqrl great work. To feel more comfortable that things work and that we ensure no changes to the new COAT metrics I encourage to write test for the new added functionality 😄

pkg/util/hostname/providers.go
louis-cqrl: Added unit tests, I was trying to do an e2e test however due to initial delay before first run of `20 min` it might add overhead to the CI for nothing, thing that compromised the possibility of efficient e2e tests, however what we can do is to add a config option that could override this to have a shorter timing in test, nonetheless I don't know if this is worth for a test ?

pkg/util/hostname/providers.go
louis-cqrl: fixed

pkg/util/hostname/drift.go
louis-cqrl: Gave it a try and it involves 24 updates of callers for the module, might want to split the work on 2 PRs for this

louis-cqrl: I also updated the PR description

pkg/util/hostname/drift.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/DataDog/datadog-agent/pkg/telemetry"
+	"github.com/DataDog/datadog-agent/pkg/util/cache"
+)
+
+// Default timing values that can be modified for testing
+var (
+	// DefaultInitialDelay is the default delay before the first check (20 minutes)
+	DefaultInitialDelay = 20 * time.Minute
+	// DefaultRecurringInterval is the default interval for recurring checks (6 hours)
+	DefaultRecurringInterval = 6 * time.Hour
```
GustavoCaso: I would make these constant private. We avoid issues in the future if we want to change them, but other part of the code use it

pkg/util/hostname/drift.go
```diff
@@ -0,0 +1,157 @@
...
+	// timingMutex protects access to the timing variables
+	timingMutex sync.RWMutex
+)
+
+// Telemetry metrics
+var (
+	tlmDriftDetected = telemetry.NewCounter("hostname", "drift_detected",
+		[]string{"state", "provider"}, "Hostname drift detection status")
+	tlmDriftResolutionTime = telemetry.NewHistogram("hostname", "drift_resolution_time_ms",
+		[]string{"state", "provider"}, "Hostname drift resolution time in milliseconds", []float64{.5, 1, 2.5, 5, 10, 60})
```
GustavoCaso: @louis-cqrl how did you arrived to this values `[]float64{.5, 1, 2.5, 5, 10, 60}`?

louis-cqrl: mainly tested in local with VM on azure and aws, if we see in the future this does not fit with customer data we can still change the bucket I think

pkg/util/hostname/drift_test.go
```diff
@@ -0,0 +1,286 @@
...
+	require.True(t, found, "Expected hostname data to be cached")
+
+	cachedHostnameData, ok := cachedData.(Data)
+	require.True(t, ok, "Expected cached data to be of type Data")
+	assert.Equal(t, hostnameData.Hostname, cachedHostnameData.Hostname)
+	assert.Equal(t, hostnameData.Provider, cachedHostnameData.Provider)
+
+	// Cancel the context to stop the goroutine
+	cancel()
+
```
GustavoCaso: The test is not asserting on the different telemetry data we are emitting, I think that part is important to make sure we are emitting the right metrics. 

You are going to have to use `telemetry.GetRegistry().Gather()` to access the telemetry data, and then iterate over it and find the right metric and check their value.

You can take inspiration from: https://github.com/DataDog/datadog-agent/blob/b7eefbcbe4fd1c9d2dc99c1bc820480a7acc9e0e/comp/core/telemetry/telemetryimpl/telemetry_mock.go#L107-L108

pkg/util/hostname/drift_test.go
```diff
@@ -0,0 +1,286 @@
...
+// Benchmark tests for performance
+func BenchmarkDetermineDriftState(b *testing.B) {
+	oldData := Data{Hostname: "old-host", Provider: "old-provider"}
+	newData := Data{Hostname: "new-host", Provider: "new-provider"}
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		determineDriftState(oldData, newData)
+	}
+}
```
GustavoCaso: Why are we benchmarking this? I don't think this part of the code is important to benchmark

pkg/util/hostname/drift.go
```diff
@@ -0,0 +1,157 @@
...
+	timingMutex.Lock()
+	defaultInitialDelay = delay
+	timingMutex.Unlock()
+}
+
+func setDefaultRecurringInterval(interval time.Duration) {
+	timingMutex.Lock()
+	defaultRecurringInterval = interval
+	timingMutex.Unlock()
+}
```
GustavoCaso: I think we can simplify things a bit: 

We do not need a mutex to configure the intervals. We have this because the way we designed the code. We can have a new struct that store the durations and attach the function `scheduleHostnameDriftChecks` to it. Then in the test we do not have to call this methods because we created an instance of the struct with the different values.

```go
type driftService struct {
	initialDelay      time.Duration
	recurringInterval time.Duration
}

...
```

In the `pkg/util/hostname/providers.go` file, we can do:
```go
var (
	hostnameExpvars  = expvar.NewMap("hostname")
	hostnameProvider = expvar.String{}
	hostnameErrors   = expvar.Map{}
	driftCalculator  = driftService{
		initialDelay:      DefaultInitialDelay,
		recurringInterval: DefaultRecurringInterval,
	}
)
```

And then we can call the `scheduleHostnameDriftChecks` on the new variable `driftCalculator` variable. 

This would simplify things for the tests and in the code

GustavoCaso (commented):
Great work. I believe we can still improve the code a bit so it is easier to test, and we should test that we emit the correct metrics 😄

iglendd (approved):
Looks good for telemetry configuration

pkg/util/hostname/drift_test.go
louis-cqrl: Removed 👍🏻

pkg/util/hostname/drift_test.go
louis-cqrl: Added e2e test to check these metrics

pkg/util/hostname/drift.go
louis-cqrl: added

pkg/config/config_template.yaml
```diff
@@ -115,6 +115,19 @@ api_key:
...
+## The initial delay before the first hostname drift check. This allows the agent
+## to settle before starting drift detection.
+#
+# hostname_drift_initial_delay: 20m
+
+## @param hostname_drift_recurring_interval - duration - optional - default: 6h
+## @env DD_HOSTNAME_DRIFT_RECURRING_INTERVAL - duration - optional - default: 6h
+## The interval between recurring hostname drift checks after the initial check.
+#
+# hostname_drift_recurring_interval: 6h
```
GustavoCaso: Is this something we want to expose to customers? @louis-cqrl, could you explain your reasons for adding a configuration option for this? 

Is this so you can tweak the values in the e2e test? If so, instead of advertising them publicly, you can define them in https://github.com/DataDog/datadog-agent/blob/b1d11fe065ade85c42274cf9ffa5d9dc013c64da/pkg/config/setup/config.go#L274 and remove them from this file. The config template is public for our customers, but you want an internal configuration key that is not exposed to customers (at lest not for now)

pkg/util/hostname/drift.go
```diff
@@ -0,0 +1,161 @@
...
+// driftService contains configuration for hostname drift detection
+type driftService struct {
+	initialDelay      time.Duration
+	recurringInterval time.Duration
+}
+
+// Default timing values
+const (
+	DefaultInitialDelay      = 20 * time.Minute
+	DefaultRecurringInterval = 6 * time.Hour
```
GustavoCaso: I would make these constants private, as they are only meant to be use in this package 😄

test/new-e2e/tests/agent-runtimes/hostname/drift_linux_test.go
```diff
@@ -0,0 +1,68 @@
...
+		if !assert.NoError(c, err, "Should be able to execute diagnose command") {
+			return
+		}
+
+		// Check that we have drift resolution time metrics in the output
+		assert.Contains(c, output, "drift_resolution_time_ms", "Should have drift_resolution_time_ms metrics in telemetry")
+
+		// Check for specific metric components
+		assert.Contains(c, output, "provider=", "Should have provider tag in metrics")
+		assert.Contains(c, output, "state=", "Should have state tag in metrics")
```
GustavoCaso: Can we assert the actual provider and state?

test/new-e2e/tests/agent-runtimes/hostname/drift_linux_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+	// Wait for the agent to start and perform initial hostname detection
+	v.EventuallyWithT(func(c *assert.CollectT) {
+		// Check that agent is running
+		status := v.Env().Agent.Client.Status(agentclient.WithArgs([]string{"--json"}))
+		assert.NotEmpty(c, status.Content, "Agent status should not be empty")
+	}, 2*time.Minute, 10*time.Second, "Agent should be running")
+
+	// Wait for the initial drift check to complete (initial delay + some buffer)
+	time.Sleep(45 * time.Second)
```
GustavoCaso: 45 seconds is a long time 😄 
Can we reduce the amount of time for the `hostname_drift_initial_delay`?

pkg/config/config_template.yaml
louis-cqrl: My idea on this was to give customer the ability to troubleshoot their hop issue on aws as well, for that a small timing window to compute the hostname may needed (~60 sec), however this issue might not be very recurrent as well, any thought on this ?

test/new-e2e/tests/agent-runtimes/hostname/drift_linux_test.go
louis-cqrl: i think we can let me take a look to it

test/new-e2e/tests/agent-runtimes/hostname/drift_linux_test.go
louis-cqrl: i can give it a try, the time value i put here where a bit random to ensure agent was ready

pkg/config/config_template.yaml
GustavoCaso: I would start by not exposing a new configuration nob to the customer and have it defined in https://github.com/DataDog/datadog-agent/blob/b1d11fe065ade85c42274cf9ffa5d9dc013c64da/pkg/config/setup/config.go#L274

If customers experience an issue, we could always tell them to use the non-public configuration. If more and more customers experience issues, we can consider making it public

test/new-e2e/tests/agent-runtimes/hostname/drift_common_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+func (v *baseHostnameDriftSuite) TestHostnameDriftMetricsEmission() {
+	// Wait for the agent to start and perform initial hostname detection
+	v.EventuallyWithT(func(c *assert.CollectT) {
+		// Check that agent is ready
+		assert.True(c, v.Env().Agent.Client.IsReady(), "Agent should be ready")
+	}, 2*time.Minute, 10*time.Second, "Agent should be ready")
+
+	// Wait for the initial drift check to complete (initial delay + some buffer)
+	time.Sleep(20 * time.Second)
```
GustavoCaso: I'm wondering if we even need the sleep. The test function `EventuallyWithT` seems to accept a waitFor and tick.

What do you think?

louis-cqrl: done

GustavoCaso (approved):
Awesome work. Thank you so much for taking into consideration my comments

dustmop (approved):

PR #39187, opened 2025-07-23
Use cloned manager.Probes slice when iterating

usamasaqib:
### What does this PR do?
The `manager.Probes` slice is mutable and may change from underneath us when iterating over it. Use the `manager.GetProbes` function to get a copy of the slice before accessing its elements.


spikat (approved):

pkg/ebpf/mappings.go
```diff
@@ -296,7 +296,10 @@ func AddProbeFDMappings(mgr *manager.Manager) {
 	mappingLock.Lock()
 	defer mappingLock.Unlock()
 
-	for _, p := range mgr.Probes {
+	// GetProbes returns a copy of the probes. We use this because mgr.Probes is mutable
+	// and so may change from underneath us.
+	probes := mgr.GetProbes()
```
val06, line 301: did we consider using an rwlock on the probes in all places we need access and copy them to avoid the race?

val06 (approved):

pkg/ebpf/mappings.go
usamasaqib: Yes this approach was suggested, but it would require identifying all places where the Probes slice is modified and locking it, or replacing it with an appropriate function call. This approach seemed simpler, so I went with this.

val06: it is more costly, in terms of memory, but perhaps this is not significant given the average size of the array and the frequency of those methods being called.

since you mentioned that you considered the other approach and decided to choose the copy solution, i'm closing this comment

PR #37084, opened 2025-05-16
(fleet)  migrate the prerm to the installer 2/2

arbll:
This PR finishes the migration started in https://github.com/DataDog/datadog-agent/pull/37078.

- Unfortunately everything put in the RPM `preinst` to hack a `prerm`-like behavior must be kept for older versions. This is not new but means we'll have to leave some legacy logic in the `preinst`.

PR #39190, opened 2025-07-23
fix(rpm): Stop the agent in legacy preinst

BaptisteFoy:
### What does this PR do?
https://github.com/DataDog/datadog-agent/pull/37078 removed the agent stop command from the RPM `preinst` to migrate it in the installer, while https://github.com/DataDog/datadog-agent/pull/37084 preserved some `preinst` commands in bash for upgrades from older agents. The combination of these two prevents us from stopping the agent before upgrading from an older version, leaving compiled Python files on disk after the upgrade that are not removed by an uninstallation of the agent.

Manual QA

### DIFF
--- a/omnibus/package-scripts/agent-rpm/preinst
+++ b/omnibus/package-scripts/agent-rpm/preinst
@@ -33,6 +33,49 @@ fi
 #                   DO NOT EDIT THIS SECTION                             #
 ##########################################################################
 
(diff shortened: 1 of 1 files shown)


raphaelgavache (approved):
nice catch

coignetp (approved):

CelianR (approved):

PR #39191, opened 2025-07-23
Cleanup WMI database inside install.ps1

JSGette:
### What does this PR do?
We have seen in the past that one particular windows docker image's size varies in a range
of 25 Mb. Layer by layer analysis shows that it is related to OBJECTS.DATA file that is
created and used by WMI as a database. Sometimes cleanup kicks in before RUN
of Dockerfiles/agent/install.ps1 finishes, sometimes it doesn't. That is the reason why
the size varies. We are adding an unconditional WMI database cleanup step. It is allowed to fail, 
because non-core images do not have WMI, therefore, script fails.

Considering, however, non-deterministic nature of docker build process and overall Windows images bulkiness 
it does not actually look very viable and/or sustainable to reactive quality gates for those images.

### Motivation
We want to achieve somewhat consistent results when we build the same thing. This step brings us closer to that goal. 

I triggered the pipeline several times and measured produced image's size to ensure it does not fluctuate anymore.

For instance, I noticed that all servercore images became 20MiB bigger with no apparent reason. Thus, it is hard to enforce the rules in case of Windows docker containers


Dockerfiles/agent/install.ps1
```diff
@@ -92,3 +92,18 @@ install_method:
...
   installer_version: docker-win-$env:INSTALL_INFO
 "@ > C:/ProgramData/Datadog/install_info
+
+# After this script is executed sometimes
+# the WMI database is approximately 25 MB
+# bigger than otherwise which leads to
+# a failing static quality gate.
+# This is a workaround to clean up the
+# WMI database and reduce the image size.
+# It is ignored for non-core base images.
```
rdesgroppes: I guess you may safely use more than 41 [characters per line](https://poshcode.gitbook.io/powershell-practice-and-style/style-guide/code-layout-and-formatting#maximum-line-length), ha ha

rdesgroppes (approved):
Thanks!

Dockerfiles/agent/install.ps1
JSGette: Reformatted

PR #39194, opened 2025-07-23
[CWS] Get all immediates used in bpf filters from setsockopt

theop-dd:
### What does this PR do?
Get all the immediates used in the BPF filter of a setsockopt event.
### Motivation
Allow to create rules that match on specific immediates. This can be used to detect the use of magic values in malwares for example.
Example : `<magic_value> in setsockopt.used_immediates` can be used as a rule.
Add the field setsockopt.used_immediates in the rules tested.

### DIFF
--- a/pkg/security/probe/field_handlers_ebpf.go
+++ b/pkg/security/probe/field_handlers_ebpf.go
@@ -974,13 +974,25 @@ func (fh *EBPFFieldHandlers) ResolveSetSockOptFilterInstructions(_ *model.Event,
 }
 
 func CustomParseBPFFilter(raw []bpf.RawInstruction) []int {
-	magic_values := []uint32{21139, 29269, 960051513}
+	contains := func(slice []int, e int) bool {
+		for _, a := range slice {
+			if a == e {
+				return true
+			}
+		}
+		return false
+	}
+	// For test only
+	magic_values := []uint32{11, 17, 24, 30, 48, 55}
+	// magic_values := []uint32{21139, 29269, 960051513}
 	var magic_values_found []int
 	for _, inst := range raw {
 		// Check if we load or branch on a magic value
 		for _, magic := range magic_values {
 			if inst.K == magic {
-				magic_values_found = append(magic_values_found, int(inst.K))
+				if !contains(magic_values_found, int(inst.K)) {
+					magic_values_found = append(magic_values_found, int(inst.K))
+				}
 			}
 		}
 	}
--- a/pkg/security/probe/field_handlers_ebpf.go
+++ b/pkg/security/probe/field_handlers_ebpf.go
@@ -987,14 +987,14 @@ func (fh *EBPFFieldHandlers) ResolveSetSockOptUsedImmediates(_ *model.Event, e *
 		}
 		return false
 	}
-	var K_values []int
+	var kValues []int
 	for _, inst := range raw {
 		// Check if we load or branch on a magic value
-		if !contains(K_values, int(inst.K)) {
-			K_values = append(K_values, int(inst.K))
+		if !contains(kValues, int(inst.K)) {
+			kValues = append(kValues, int(inst.K))
 		}
 
 	}
-	e.UsedImmediates = K_values
+	e.UsedImmediates = kValues
 	return e.UsedImmediates
 }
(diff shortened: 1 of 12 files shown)


releasenotes/notes/add-detection-of-immediates-used-in-bpf-filter-d314c7e979a154af.yaml
```diff
@@ -0,0 +1,18 @@
+# Each section from every release note are combined when the
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+features:
```
paulcacheux: let's keep the release note out for now, release notes are common to all agent sub-systems for a given release so we usually don't talk about internal changes

theop-dd: Okay! 
So I remove it completely?

pkg/security/probe/field_handlers_ebpf.go
```diff
@@ -971,3 +975,26 @@ func (fh *EBPFFieldHandlers) ResolveSetSockOptFilterInstructions(_ *model.Event,
 	}
 	return e.FilterInstructions
 }
+
+// ResolveSetSockOptUsedImmediates resolves the immediates in the bpf filter of a setsockopt event
+func (fh *EBPFFieldHandlers) ResolveSetSockOptUsedImmediates(_ *model.Event, e *model.SetSockOptEvent) []int {
+	raw := parseFilter(e)
+	contains := func(slice []int, e int) bool {
```
YoannGh: Is it possible to re-use [`slice.Contains`](https://pkg.go.dev/slices#Contains) here?

theop-dd: Oh yes you're right, no need for a custom func here.

releasenotes/notes/add-detection-of-immediates-used-in-bpf-filter-d314c7e979a154af.yaml
```diff
@@ -0,0 +1,18 @@
...
+    - | 
+      Get all the immediates used in the BPF filter of a setsockopt event.
+    - | 
+      Allow to create rules that match on specific immediates.
```
rtrieu:
```suggestion
      Allow users to create rules that match on specific immediates.
```

releasenotes/notes/add-detection-of-immediates-used-in-bpf-filter-d314c7e979a154af.yaml
```diff
@@ -0,0 +1,18 @@
...
+      This can be used to detect the use of magic values in malwares for example.
```
rtrieu:
```suggestion
      This can be used, for example, to detect the use of magic values in malware.
```

rtrieu (approved):
A few minor suggestions, but otherwise looks good!

releasenotes/notes/add-detection-of-immediates-used-in-bpf-filter-d314c7e979a154af.yaml
theop-dd: We decided to remove the release note so it will solve that :)

lebauce (approved):

paulcacheux (approved):

pkg/security/probe/field_handlers_ebpf.go
```diff
@@ -971,3 +976,18 @@ func (fh *EBPFFieldHandlers) ResolveSetSockOptFilterInstructions(_ *model.Event,
 	}
 	return e.FilterInstructions
 }
+
+// ResolveSetSockOptUsedImmediates resolves the immediates in the bpf filter of a setsockopt event
+func (fh *EBPFFieldHandlers) ResolveSetSockOptUsedImmediates(_ *model.Event, e *model.SetSockOptEvent) []int {
+	raw := parseFilter(e)
```
YoannGh, line 985: You could early return `e.UsedImmediates` here in case the value was already computed

YoannGh (approved):

pkg/security/probe/field_handlers_ebpf.go
theop-dd: You are completely right

PR #1369, opened 2018-02-28
Clean up further mentions of "beta" in the docs

nmuesch:
### What does this PR do?

Remove mentions of beta in the github docs

### Motivation

Clean up documentation

### Additional Notes

Anything else we should know when reviewing?

PR #465, opened 2017-08-16
add dsd6 udp daemonset documentation

xvello:
### What does this PR do?

- Update the README with a UDP DaemonSet
- Add back the old `DD_SEND_HOST_METADATA` variable to both DaemonSets as the current beta build uses this one.

PR #39196, opened 2025-07-23
Remove beta-only option

ian28223:
### What does this PR do?

Remove `DD_SEND_HOST_METADATA` which only applied to the dogstatsd:beta build. (Ref: https://github.com/DataDog/datadog-agent/pull/465); stable builds use `DD_ENABLE_METADATA_COLLECTION` instead.
This transitioned out of beta phase in 2018. (Ref: https://github.com/DataDog/datadog-agent/pull/1369)

### Motivation

Caused some confusion as to what `DD_SEND_HOST_METADATA` with no mention that this only applied to beta builds that is no longer used.


rdesgroppes (approved):
Thanks for the cleanup!

PR #39197, opened 2025-07-23
USM: Prevents Segmentation Faults

DanielLavie:
### What does this PR do?
This PR prevents segmentation faults caused by USM by detecting and disabling uretprobe-based functionality on kernels affected by the specific bug. The changes add kernel bug detection to:
  - Native TLS monitoring (ebpf_ssl.go)
  - Istio monitoring (istio.go)
  - NodeJS monitoring (nodejs.go)

Additionally, it adds Ubuntu 24.10 to the Kernel Matrix Testing (KMT) CI pipeline and includes test skips for affected kernel versions.
It also introduces a new test, TestOpenSSLTLSContainer, to cover a previously untested scenario in which a server using OpenSSL runs inside a container, triggering the bug.

### Motivation
A kernel bug exists where uretprobes can cause segmentation faults in applications running with seccomp filters. This affects kernels with the __x64_sys_uretprobe syscall symbol. The bug manifests when USM's uretprobe-based monitoring is active on containerized applications that use seccomp.

- Small runtime overhead from kernel bug detection (mitigated by memoization)


pkg/gpu/probe_test.go
```diff
@@ -235,6 +236,13 @@ func (s *probeTestSuite) TestMultiGPUSupport() {
 func (s *probeTestSuite) TestDetectsContainer() {
 	t := s.T()
 
+	kernelVersion, err := kernel.HostVersion()
+	require.NoError(t, err)
+
+	if kernelVersion >= kernel.VersionCode(6, 11, 0) {
+		t.Skip("skipping test for kernel version 6.11.0 and higher until we correct cgroupv2 device allow")
+	}
+
```
gjulianm: Why not check directly if we have the bug?

DanielLavie: Sorry it was by mistake, deleted.

pkg/network/usm/config/config.go
```diff
@@ -47,6 +48,17 @@ func TLSSupported(c *config.Config) bool {
...
+	hasUretprobeBug, err := kernelbugs.HasUretprobeSyscallSeccompBug()
+	if err != nil {
+		log.Errorf("failed to check for uretprobe syscall seccomp bug: %v", err)
+		return false
+	}
+	if hasUretprobeBug {
+		log.Warn("TLS monitoring disabled due to kernel bug that causes segmentation faults with uretprobes and seccomp filters")
+		return false
+	}
+
```
guyarb: That change is going to disable go-tls as well, although it does not use uretprobes.

Please fix that.

Also, AFAIK, only nodejs made the segfault, but openssl didn't
do we know why?

DanielLavie: Fixed.
I assumed OpenSSL caused the segfault because it uses uretprobe, but I wasn’t aware that wasn’t the case

guyarb: I wasn't able to reproduce it using our own tests.
Perhaps we don't have openssl scenario running within a container
It might be useful to try and verify if we have such a scenario or to add one, based on our current python3 HTTPs servers.

pkg/network/usm/config/config.go
```diff
@@ -54,6 +55,29 @@ func TLSSupported(c *config.Config) bool {
 	return kversion >= MinimumKernelVersion
 }
 
+// UretprobeTLSSupported returns true if TLS monitoring using uretprobes is supported.
+// This includes general TLS platform support plus checking for the kernel bug that
+// causes segfaults with uretprobes and seccomp filters
+func UretprobeTLSSupported(c *config.Config) bool {
```
guyarb: Just for you to consider - the relation between TLSSupported and UretprobeSupported is artificial
Today we only use Uretprobes for TLS, but it might not always be the case.

And from my perspective, I think the coupling does not add much

guyarb (approved):

pkg/network/usm/config/config.go
DanielLavie: Agreed and fixed.

pkg/network/usm/config/config.go
DanielLavie: I didn't find any openssl scenario running within a container, I'll add one

pkg/network/protocols/http/testutil/testdata/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+version: '3'
+name: python
+services:
+  python:
+    image: python:3-alpine
```
guyarb, line 5: you need to open a PR (like [this](https://github.com/DataDog/ami-builder/pull/193/)) to add the image for the CI, otherwise the test might failed with `error 429` from dockerhub

pkg/network/protocols/http/testutil/pythonserver.go
```diff
@@ -120,3 +125,80 @@ func rawConnect(ctx context.Context, t *testing.T, host, port string) {
...
+	return nil
+}
+
+// HTTPPythonServerContainer launches an HTTPs server written in Python inside a container.
+func HTTPPythonServerContainer(t *testing.T, key, cert, serverPort string) error {
+	t.Helper()
+	dir, _ := CurDir()
+
+	// Set up certificates
+	if err := linkFile(t, key, dir+"/testdata/certs/srv.key"); err != nil {
```
guyarb: It fails as `certs` dir does not exist under `testdata`
you can either change it by removing the `certs` dir from the path, or create an empty directoy `certs` and add dummy file `.keep` to it, so git will add this directory

pkg/network/protocols/http/testutil/pythonserver.go
```diff
@@ -120,3 +125,80 @@ func rawConnect(ctx context.Context, t *testing.T, host, port string) {
...
+	}
+
+	env := []string{
+		"ADDR=0.0.0.0",
+		"PORT=" + serverPort,
+		"CERTS_DIR=/v/certs",
+		"TESTDIR=" + dir + "/testdata",
+	}
+
+	scanner, err := globalutils.NewScanner(regexp.MustCompile("Server running at https.*"), globalutils.NoPattern)
```
guyarb, line 183: the script does not generate such a log, the test fails on it

guyarb: you should be adding on line 59 - `    print(f"Server running at https://{server_address[0]}:{server_address[1]}/")`
but also, change the docker compose to use `-u` (unbufferred output) to show the prints in the docker console

pkg/network/protocols/http/testutil/testdata/docker-compose.yml
DanielLavie: I don't see any test failing, are you sure this is required?

gjulianm: Not "required" exactly, specially as the pull_test_dockers jobs should download the image before starting up KMT, but it's recommended to avoid rate limits once we get this into main.

gjulianm (approved):

pkg/network/protocols/http/testutil/testdata/docker-compose.yml
guyarb: If you don't add it, then it will be pulled in the CI, and then you might fail downloading it if our CI exceeded the rate limit of dockerhub

pkg/network/usm/monitor_tls_test.go
```diff
@@ -1097,18 +1097,21 @@ func testOpenSSLSegfaultBehavior(t *testing.T, usmMonitor *Monitor, pythonPID ui
 	t.Log("Kernel bug detected - testing OpenSSL behavior in container (currently still enabled)")
 
 	initialPID := pythonPID
-	client := &nethttp.Client{Timeout: 5 * time.Second}
+	client := &nethttp.Client{
+		Timeout: 1 * time.Second,
```
amitslavin, line 1119: You reduced the timeout from 5 to 1, was that intentional?

pkg/network/usm/monitor_tls_test.go
```diff
@@ -1097,18 +1097,21 @@ func testOpenSSLSegfaultBehavior(t *testing.T, usmMonitor *Monitor, pythonPID ui
...
+		Transport: &nethttp.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		},
+	}
+
+	time.Sleep(60 * time.Second)
```
amitslavin: why have you added this 1 min sleep?

pkg/network/protocols/http/testutil/pythonserver.go
```diff
@@ -154,21 +155,13 @@ func linkFile(t *testing.T, src, dst string) error {
...
-	if err := linkFile(t, cert, dir+"/testdata/certs/srv.crt"); err != nil {
-		return err
-	}
-
-	// Create Python script using existing format
-	certPath := "/v/certs/srv.crt"
-	keyPath := "/v/certs/srv.key"
+	// Create Python script using existing format - reference original certificate files directly
+	certPath := "/v/cert.pem.0"
+	keyPath := "/v/server.key"
```
amitslavin: would create those as consts

pkg/network/protocols/http/testutil/testdata/docker-compose.yml
amitslavin: @DanielLavie have you added the image?

pkg/network/usm/ebpf_ssl_test.go
```diff
@@ -40,7 +40,7 @@ func testArch(t *testing.T, arch string) {
 	cfg := utils.NewUSMEmptyConfig()
 	cfg.EnableNativeTLSMonitoring = true
 
-	if !usmconfig.TLSSupported(cfg) {
+	if !usmconfig.TLSSupported(cfg) || !usmconfig.UretprobeSupported() {
 		t.Skip("shared library tracing not supported for this platform")
```
amitslavin, line 44: The skip message should be updated for all cases covered by this new check, as we may be skipping for a different reason

pkg/network/protocols/http/testutil/testdata/docker-compose.yml
DanielLavie: https://github.com/DataDog/ami-builder/pull/237

pkg/network/usm/monitor_tls_test.go
```diff
@@ -998,3 +1073,148 @@ func (s *tlsSuite) TestNodeJSTLS() {
...
+		require.NoError(t, err)
+		resp.Body.Close()
+
+		requests = append(requests, req)
+	}
+
+	client.CloseIdleConnections()
+
+	// Verify that all requests are captured
+	requestsMap := make(map[*nethttp.Request]bool)
```
amitslavin: not sure I understood this part

pkg/network/usm/monitor_tls_test.go
DanielLavie: All other tests are using 1 as well that was a mistake

pkg/network/usm/monitor_tls_test.go
DanielLavie: By mistake, removed

pkg/network/usm/ebpf_ssl_test.go
DanielLavie: The message is wrong to begin with, we can fix it in another PR

pkg/network/usm/monitor_tls_test.go
DanielLavie: Removed

amitslavin (approved):

PR #39201, opened 2025-07-23
CAP-2684 Add unit tests for Metadata and Manifest chunking and buffering

kangyili:
### What does this PR do?

For chunking, we already have `TestChunkOrchestratorMetadataBySizeAndWeight`. This PR moves the test to the right file.
For buffering, we already have `TestOrchestratorManifestBuffer`. This PR just add a new to for `NewManifestBuffer`


fisherevans (changes requested):
I think we need to add a test for the flushing, which is where we had a regression last time - we were not generating the right tags before the context was not being constructed correctly.

https://github.com/DataDog/datadog-agent/blob/406ee75ade937aad05d0e5a25168fe30238facaf/pkg/collector/corechecks/cluster/orchestrator/manifest_buffer.go#L91-L109

kangyili:
> I think we need to add a test for the flushing, which is where we had a regression last time - we were not generating the right tags before the context was not being constructed correctly.
> 
> https://github.com/DataDog/datadog-agent/blob/406ee75ade937aad05d0e5a25168fe30238facaf/pkg/collector/corechecks/cluster/orchestrator/manifest_buffer.go#L91-L109

I think you added it https://github.com/DataDog/datadog-agent/blob/406ee75ade937aad05d0e5a25168fe30238facaf/pkg/collector/corechecks/cluster/orchestrator/manifest_buffer_test.go#L46 

I can add another one https://github.com/DataDog/datadog-agent/pull/39201/commits/b642023ffdb86ebf4397eeafaa34bc2321a13ece

fisherevans (approved):

PR #39202, opened 2025-07-23
[EBPF-632] gpu: support Docker in container device mapping

gjulianm:
### What does this PR do?

Adds support for reading the `NVIDIA_VISIBLE_DEVICES` environment variable in Docker containers, so that we assign correctly the devices used in that container.

It also sets up the code so that container<->device assignment can be retrieved from other sources.

Finally, it moves the code to generate fake procFS entries to the `pkg/util/kernel` package to avoid import issues.

### Motivation

https://datadoghq.atlassian.net/browse/EBPF-632

Added unit tests, manual validation.


pkg/gpu/containers/containers.go
```diff
@@ -13,22 +13,99 @@ import (
...
+
+// HasGPUs returns true if the container has GPUs assigned to it.
+func HasGPUs(container *workloadmeta.Container) bool {
+	switch container.Runtime {
+	case workloadmeta.ContainerRuntimeDocker:
+		// If we have an error, we assume there are no GPUs for the container, so
+		// ignore it.
+		envVar, _ := getDockerVisibleDevicesEnv(container)
+		return envVar != ""
+	default:
```
val06, line 40: there is no way to distinguish k8s?

pkg/gpu/containers/containers.go
```diff
@@ -13,22 +13,99 @@ import (
...
+	var multiErr error
+
+	visibleDevicesVar, err := getDockerVisibleDevicesEnv(container)
+	if err != nil {
+		return nil, err
+	}
+
+	if visibleDevicesVar == "" {
+		return nil, fmt.Errorf("%s is not set, can't match devices", nvidiaVisibleDevicesEnvVar)
+	}
```
val06, line 86: do you plan in the future to add support matching the devices by parsing the specs using container runtime API?

val06 (approved):
approved, and left a few questions, also the gpu.go in the core-check will change soon to return just one container for each gpu device, that change won't break the logic of this PR?

guyarb (approved):

releasenotes/notes/gpu-nvidia-vis-devices-3981406035f0703f.yaml
```diff
@@ -0,0 +1,11 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+fixes:
+  - |
+    gpu: fix container to device assignment for Docker containers
```
drichards-87:
```suggestion
    gpu: fix container-to-device assignment for Docker containers
```

drichards-87 (approved):
Left a small suggestion on the release note and approved the PR.

pkg/gpu/containers/containers.go
gjulianm: No, the container runtime is separate from the scheduler here. My plan was to use kubernetes as a fallback if the container-specific methods aren't available.

pkg/gpu/containers/containers.go
gjulianm: Yes, but for now looking at the env variable seemed the most reliable/simple method

gjulianm (commented):
/merge

val06 (approved):

guyarb (approved):

PR #39204, opened 2025-07-23
fix(ami): Use new AMI with less storage request

chouetz:
### What does this PR do?
Update the recent AMI we created for debian with a 8GB default storage instead of 200GB


tbavelier (approved):

chouetz:

PR #39206, opened 2025-07-23
[CWS-4908] Report capabilities usage

YoannGh:
### What does this PR do?

Report Linux capabilities that a process attempted and successfully used while running.

A new event `capabilities` event type is introduced.
Capabilities kernels events are generated if new capabilities are used on three cases:
- when a process exits
- when a process executes a new program
- every `event_monitoring_config.capabilities_monitoring.period`

This last case constrains this feature through the availability of the `bpf_for_each_map_elem` helper (introduced in Linux v5.13, and fixed in v5.17 on arm64 kernels).

`capabilities` events are reported in activity dumps and rules can be written to detect whenever a set of capabilities is used by a process: `capabilities.used & (CAP_DAC_OVERRIDE|CAP_DAC_READ_SEARCH) > 0` or attempted `capabilities.attempted & CAP_SYS_ADMIN > 0`.

The same information is also reported for each individual process through the `process.caps_used` and `process.caps_attempted` fields.

This feature is gated by the `event_monitoring_config.capabilities_monitoring.enabled` configuration parameter, which is disabled by default for now.

### Motivation

Being able to detect processes that have access to capabilities they aren't using.


rahulkaukuntla (approved):

docs/cloud-workload-security/linux_expressions.md
```diff
@@ -449,6 +456,17 @@ A BPF command was executed
 | [`bpf.prog.type`](#bpf-prog-type-doc) | Type of the eBPF program |
 | [`bpf.retval`](#common-syscallevent-retval-doc) | Return value of the syscall |
 
+### Event `capabilities`
+
+_This event type is experimental and may change in the future._
```
estherk15, line 510: What does this mean and why is it necessary to include this? Is it in Preview?
Docs best practice is to document what's true now, so I suggest removing mention of what might happen in the future.

```suggestion
_This event type is in Preview._
```

estherk15 (approved):
Left one suggestion to remove temporal language!

paulcacheux (approved):

usamasaqib (approved):

PR #39212, opened 2025-07-23
dyninst/symdb: rework parsing of function names

andreimatei:
This patch changes the parsing of function names such that anonymous
functions defined inside methods are no longer considered methods
themselves. E.g.
`github.com/foo/bar/myPkg/myType.myMethod.func1`
used to be parsed as a method called `myMethod.func1` on type myType.
Now, it's parsed as a standalone function called
`myType.myMethod.func1`.

This patch also stops trying to recognize "anonymous functions defined
inside functions which were inlined". These are functions that look
like:
`runtime.gcMarkDone.forEachP.func5`
Here, `forEachP` is a function that was inlined inside `gcMarkDone`.
`forEachP.func5` is a function that was not inlined, but (I think) a
specialization of `forEachP.func5` was produced for this particular
inlined caller. We'd like to ignore such functions because they're
useless to a user.
The problem with ignoring these functions is that their parsing is
ambiguous. This example could also be parsed as a type called
`gcMarkDone` with a method called `forEachP` with an anonymous function
inside it. We had some logic trying to distinguish the two, but it was
buggy crap. For now, I gave up on trying. Although there are tons of
such functions polluting our output, so I might try to do something
again in the future.

### DIFF
--- a/pkg/dyninst/symdb/func_name.go
+++ b/pkg/dyninst/symdb/func_name.go
@@ -16,43 +16,30 @@ import (
 // Utilities for parsing Go function names.
 
 var (
-	// The parsing goes as follows:
-	// - an optional package name: (((.*/)?.*?)\.)?. Consume (greedily)
-	// everything up to the last slash, and then non-greedily up to the
-	// following dot.
-	// - an optional type name: (\(?\*?(.*?)\)?\.)?. The type name maybe be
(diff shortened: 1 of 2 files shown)


andreimatei:
Only the last commit is to be reviewed here.

piob-io (approved):
Last commit lg

PR #39219, opened 2025-07-24
[ACIX-616] Stack cleaner migration: enable it on EKS jobs

CelianR:
### What does this PR do?

This enables remote stack cleaning on EKS jobs to migrate step by step. This was enabled only on `cws` test for the moment.


chouetz (approved):
Ship it!

PR #39223, opened 2025-07-24
[CWS] Allow hash action on all events

lebauce:
### What does this PR do?

Allow setting a `hash` action on other events that `open` and `exec`.

### Motivation

This allows the hash computation of a process doing a `connect` syscall for instance.


paulcacheux:
I'm wondering if we really want that, IMO what we really want is a new action "hash_process" or something like that to actually hash the action-doing process. Otherwise it's going to get confusing and we won't give customers the ability to do "hash the process, on an open event" which is not great IMO.

lebauce:
@paulcacheux As discussed offline, this is done in 2 steps:
- the first one is to allow hashing on all events with the "default" behavior (for non FIM related events, hashing only on the process context makes sense)
- the second one is to add an optional field on the `hash` action to be able to do both: hashing on the process context OR on the targeted file(s).
Does it make sense ?

paulcacheux (approved):

PR #39225, opened 2025-07-24
[NDMII-3554] Do not re-query an OID that is missing from a device for a certain delay

Pierre-L42:
### What does this PR do?

This PR adds logic to not try to re-get an OID from a device when the device does not report this OID for the next `profileRefreshDelay` seconds in the SNMP check.

### Motivation

Decrease the number of SNMP queries we do to the device every checks.

Changes has been tried in mimic, and we check that the number of submitted metrics is still the same before and after the changes:
image
The agent was installed at around 14:15, and we can see that the number of submitted metrics is still the same (don't pay attention to the peak at 14:15, it happens when we install a new agent).

It has also been tested on CiscoCML by checking the list of queried OIDs: we check that the missing OIDs have been removed from the list of queried OIDs. And that if the device does start to report an OID that was missing, this OID is re-added back to the list of queried OIDs after `profileRefreshDelay`.


AlexandreYang (approved):
💯 Nice!

PR #39226, opened 2025-07-24
[AGENTRUN-583] add internal telemetry to Remote Agent Registry

GustavoCaso:
### What does this PR do?

This PR adds internal metrics for the Remote Agent Registry. 

With the upcoming work and the importance of this component, it is important to be able to observe changes during the increased usage of the component when deploying to staging and production. 

Once we start deploying, we can create an internal dashboard that will facilitate consuming the information

### DIFF
--- a/comp/core/remoteagentregistry/impl/remoteagentregistry.go
+++ b/comp/core/remoteagentregistry/impl/remoteagentregistry.go
@@ -91,62 +91,62 @@ func newRemoteAgent(reqs Requires) *remoteAgentRegistry {
 }
 
 type telemetryStore struct {
-	// remoteAgentRegistry tracks how many remote agents are registered.
-	remoteAgentRegistry telemetry.Counter
-	// remoteAgentRegistryError tracks how many remote agents failed to register.
-	remoteAgentRegistryError telemetry.Counter
-	// remoteAgentRegistryUpdate tracks how many remote agents are updated.
-	remoteAgentRegistryUpdate telemetry.Counter
-	// remoteAgentRegistryUpdateError tracks how many remote agents failed to update.
-	remoteAgentRegistryUpdateError telemetry.Counter
-	// remoteAgentRegistryDeregister tracks how many remote agents are deregistered.
-	remoteAgentRegistryDeregister telemetry.Counter
-	// remoteAgentRegistryActionError tracks the number of errors encountered while performing actions on the remote agent registry.
-	remoteAgentRegistryActionError telemetry.Counter
-	// remoteAgentRegistryActionDuration tracks the duration of actions performed on the remote agent registry.
-	remoteAgentRegistryActionDuration telemetry.Histogram
-	// remoteAgentRegistryActionTimeout tracks the number of times an action on the remote agent registry timed out.
-	remoteAgentRegistryActionTimeout telemetry.Counter
+	// remoteAgentRegistered tracks how many remote agents are registered.
+	remoteAgentRegistered telemetry.Counter
+	// remoteAgentRegisteredError tracks how many remote agents failed to register.
+	remoteAgentRegisteredError telemetry.Counter
+	// remoteAgentUpdated tracks how many remote agents are updated.
+	remoteAgentUpdated telemetry.Counter
+	// remoteAgentUpdatedError tracks how many remote agents failed to update.
+	remoteAgentUpdatedError telemetry.Counter
+	// remoteAgentDeregistered tracks how many remote agents are deregistered.
+	remoteAgentDeregistered telemetry.Counter
+	// remoteAgentActionError tracks the number of errors encountered while performing actions on the remote agent registry.
+	remoteAgentActionError telemetry.Counter
+	// remoteAgentActionDuration tracks the duration of actions performed on the remote agent registry.
+	remoteAgentActionDuration telemetry.Histogram
+	// remoteAgentActionTimeout tracks the number of times an action on the remote agent registry timed out.
+	remoteAgentActionTimeout telemetry.Counter
 }
 
 func newTelemetryStore(telemetryComp telemetry.Component) *telemetryStore {
 	return &telemetryStore{
-		remoteAgentRegistry: telemetryComp.NewCounterWithOpts(
+		remoteAgentRegistered: telemetryComp.NewCounterWithOpts(
 			"remote_agent_registry",
-			"registered_agents",
+			"registered",
 			[]string{"name"},
 			"Number of remote agents registered in the remote agent registry.",
 			telemetry.Options{NoDoubleUnderscoreSep: true},
 		),
-		remoteAgentRegistryError: telemetryComp.NewCounterWithOpts(
+		remoteAgentRegisteredError: telemetryComp.NewCounterWithOpts(
 			"remote_agent_registry",
-			"registered_agents_error",
+			"registered_error",
 			[]string{"name", "error"},
 			"Number of remote agents that failed to register in the remote agent registry.",
 			telemetry.Options{NoDoubleUnderscoreSep: true},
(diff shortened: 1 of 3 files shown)


rayz (approved):
Perhaps we could create a helper function for measuring actions but otherwise LGTM 🙂

comp/core/remoteagentregistry/impl/remoteagentregistry.go
```diff
@@ -126,11 +210,14 @@ func (ra *remoteAgentRegistry) RegisterRemoteAgent(registration *remoteagentregi
 		// connecting when we try to query the remote agent for status or flare data.
 		details, err := newRemoteAgentDetails(registration)
 		if err != nil {
+			ra.telemetryStore.remoteAgentRegisteredError.Inc(registration.DisplayName, err.Error())
```
tobz: I think we'd want to be using `agentID`, since display name will be stuff like "Agent Data Plane", etc... which is hard to use as a tag value for querying purposes.

comp/core/remoteagentregistry/impl/remoteagentregistry.go
```diff
@@ -141,6 +228,7 @@ func (ra *remoteAgentRegistry) RegisterRemoteAgent(registration *remoteagentregi
 
 		client, err := newRemoteAgentClient(registration)
 		if err != nil {
+			ra.telemetryStore.remoteAgentUpdatedError.Inc(registration.DisplayName, err.Error())
```
tobz: The error here is going to be _super_ noisy as a tag value, I think?

I wonder if there's some existing helper method for trying to take a gRPC client error (which is 95% of the errors we're going to get here) and and getting a more programmatic error identifier from it... like `dns_lookup_failed`, etc.. that sort of thing. 🤔

tobz (commented):
Overall, I'm +1 on these changes.

Left some comments around what we're using for tag values, etc, to try and tighten up the telemetry we'll end up emitting here.

comp/core/remoteagentregistry/impl/remoteagentregistry.go
```diff
@@ -89,14 +90,97 @@ func newRemoteAgent(reqs Requires) *remoteAgentRegistry {
...
+	// remoteAgentActionDuration tracks the duration of actions performed on the remote agent registry.
+	remoteAgentActionDuration telemetry.Histogram
+	// remoteAgentActionTimeout tracks the number of times an action on the remote agent registry timed out.
+	remoteAgentActionTimeout telemetry.Counter
+}
+
+func newTelemetryStore(telemetryComp telemetry.Component) *telemetryStore {
+	return &telemetryStore{
+		remoteAgentRegistered: telemetryComp.NewCounterWithOpts(
+			"remote_agent_registry",
```
pgimalac: We might want to store this in a const since it's the same for every metric

comp/core/remoteagentregistry/impl/remoteagentregistry.go
GustavoCaso: I don't have a strong opinion about this, but looking at the RFC for the new Remote Agent Registry, it seems that the agentID might be removed from the registration payload https://datadoghq.atlassian.net/wiki/spaces/ARUN/pages/5262901901/RFC+Future-proofing+the+RemoteAgentRegistry#Removed-items

If the ID is removed, we can sanitize the display name

comp/core/remoteagentregistry/impl/remoteagentregistry.go
GustavoCaso: I can have a look to see if we can get a simple error message

comp/core/remoteagentregistry/impl/remoteagentregistry.go
```diff
@@ -89,14 +90,97 @@ func newRemoteAgent(reqs Requires) *remoteAgentRegistry {
...
+	remoteAgentUpdatedError telemetry.Counter
+	// remoteAgentDeregistered tracks how many remote agents are deregistered.
+	remoteAgentDeregistered telemetry.Counter
+	// remoteAgentActionError tracks the number of errors encountered while performing actions on the remote agent registry.
+	remoteAgentActionError telemetry.Counter
+	// remoteAgentActionDuration tracks the duration of actions performed on the remote agent registry.
+	remoteAgentActionDuration telemetry.Histogram
+	// remoteAgentActionTimeout tracks the number of times an action on the remote agent registry timed out.
+	remoteAgentActionTimeout telemetry.Counter
+}
```
pgimalac: Do we need registered + deregistered, or should we just have a Gauge with the number of live registered agents ?

pgimalac (approved):
LGTM

comp/core/remoteagentregistry/impl/remoteagentregistry.go
GustavoCaso: Fair point. Let's have a single gauge metric

comp/core/remoteagentregistry/impl/remoteagentregistry.go
GustavoCaso: Done https://github.com/DataDog/datadog-agent/pull/39226/commits/dac824855b58e48f7a8561ca14c269144aadddc9

comp/core/remoteagentregistry/impl/remoteagentregistry.go
```diff
@@ -232,9 +325,24 @@ func (c *registryCollector) getRegisteredAgentsTelemetry(ch chan<- prometheus.Me
 
 	for agentID, details := range c.agentMap {
 		go func() {
+			start := time.Now()
+			defer func() {
+				c.telemetryStore.remoteAgentActionDuration.Observe(
+					time.Since(start).Seconds(),
+					details.sanatizedDisplayName,
+					"telemetry",
```
misteriaud: **🥜 nitpick:** ‏It's clearer to use a constant at the top of the file, it will make it easier then to have a list of the available providers IMO.

comp/core/remoteagentregistry/impl/remoteagentregistry.go
```diff
@@ -424,9 +548,25 @@ func (ra *remoteAgentRegistry) fillFlare(builder flarebuilder.FlareBuilder) erro
...
 			// We push any errors into "failure reason" which ends up getting shown in the status details.
 			resp, err := details.client.GetFlareFiles(ctx, &pb.GetFlareFilesRequest{}, grpc.WaitForReady(true))
 			if err != nil {
+				errorString := codes.Unknown.String()
+				status, ok := grpcStatus.FromError(err)
+				if ok {
+					errorString = status.Code().String()
+				}
+
+				ra.telemetryStore.remoteAgentActionError.Inc(details.sanatizedDisplayName, "flare", errorString)
```
misteriaud: **💬 suggestion:** ‏This grpc error conversion logic is repeated in each provider, we should refactor it in one separated function to keep the logic at one place.

misteriaud (approved):
Nice work ! LGTM, I just left some nitpicks and refactoring suggestions.

comp/core/remoteagentregistry/impl/remoteagentregistry.go
GustavoCaso: Done https://github.com/DataDog/datadog-agent/pull/39226/commits/3274f06686b71698366051b3d239b3c2ea641e81

comp/core/remoteagentregistry/impl/remoteagentregistry.go
GustavoCaso: Done https://github.com/DataDog/datadog-agent/pull/39226/commits/3274f06686b71698366051b3d239b3c2ea641e81

PR #39229, opened 2025-07-24
Fix object memory sizes in Python interpreter memory usage summary

Enzu83:
### What does this PR do?

This PR fixes a bug in the Python interpreter memory usage summary. Object memory sizes were incorrectly multiplied by the number of instances, even though the sizes already accounted for that.

### Motivation

The Python interpreter summary wasn't correct, object memory sizes were far too high.

The definition of a `pympler.summary` is defined  [here](https://pympler.readthedocs.io/en/latest/library/summary.html).

The `size` in each entry corresponds to the total size of the objects, not the size of one object.
So there's no need to multiply the size with the number of objects.


dustmop (approved):
LGTM, thanks for the fix!

PR #39232, opened 2025-07-24
[AMCC-85] dogstatsd: on blocklist update, update the runtime config values.

remeh:
### What does this PR do?

When updating the list on remote config polls, update the runtime config values for them to properly appear in `agent config` and flares.

### Motivation

Easier troubleshooting.

Manually tested using the `agent config` command.

### DIFF
--- a/comp/dogstatsd/server/server.go
+++ b/comp/dogstatsd/server/server.go
@@ -111,7 +111,7 @@ type localBlocklistConfig struct {
 // Server represent a Dogstatsd server
 type server struct {
 	log    log.Component
-	config model.Reader
+	config model.ReaderWriter
 	// listeners are the instantiated socket listener (UDS or UDP or both)
 	listeners []listeners.StatsdListener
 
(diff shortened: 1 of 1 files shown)


vickenty (approved):

comp/dogstatsd/server/server.go
```diff
@@ -605,6 +609,12 @@ func (s *server) handleMessages() {
 }
 
 func (s *server) restoreBlocklistFromLocalConfig() {
+	s.log.Debug("Restoring blocklist with local config.")
+
+	// update the runtime config to be consistent
+	// in `agent config` calls.
+	s.config.Set("statsd_metric_blocklist", s.localBlocklistConfig.metricNames, model.SourceAgentRuntime)
+
 	s.SetBlocklist(
```
olivielpeau, line 618: I think this `SetBlocklist` call will immediately set the config source back to `model.SourceRC` no?

comp/dogstatsd/server/server.go
```diff
@@ -605,6 +609,12 @@ func (s *server) handleMessages() {
 }
 
 func (s *server) restoreBlocklistFromLocalConfig() {
+	s.log.Debug("Restoring blocklist with local config.")
+
+	// update the runtime config to be consistent
+	// in `agent config` calls.
+	s.config.Set("statsd_metric_blocklist", s.localBlocklistConfig.metricNames, model.SourceAgentRuntime)
```
olivielpeau, line 616: I think the correct source to set here is what the source of the blocklist was before the blocklist was set via RC (if it was set via RC).

I don't think the behavior inline will lead to bugs today, but I feel it doesn't fit well with how the config package is designed 🤔

comp/dogstatsd/server/server.go
remeh: Correct! Thank you. Addressed in https://github.com/DataDog/datadog-agent/pull/39375

comp/dogstatsd/server/server.go
remeh: Good eye. Thank you! Addressed in https://github.com/DataDog/datadog-agent/pull/39375

PR #39235, opened 2025-07-24
[NDM]enrich ndm interface metrics with redapl tags

jedupau:
### What does this PR do?

This PR udpates snmp, cisco_sdwan and versa integrations to tag metrics by ndm_interface and ndm_device

### DIFF
--- a/pkg/collector/corechecks/network-devices/versa/report/sender.go
+++ b/pkg/collector/corechecks/network-devices/versa/report/sender.go
@@ -374,6 +374,7 @@ func (s *Sender) SendInterfaceMetrics(interfaceMetricsByDevice map[string][]clie
 				"interface:" + metric.Interface,
 				"vrf:" + metric.VRF,
 				"host_inf:" + metric.HostInf,
+				"dd.internal.resource:ndm_interface:" + metric.Interface,
 			}
 
 			tags := append(deviceTags, interfaceTags...)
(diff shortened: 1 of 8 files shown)


zoedt (approved):
overall, lgtm ! just the same comment about recording the table name change for context to reference for rest of team to reference!

pkg/collector/corechecks/network-devices/versa/report/sender.go
```diff
@@ -374,6 +374,7 @@ func (s *Sender) SendInterfaceMetrics(interfaceMetricsByDevice map[string][]clie
 				"interface:" + metric.Interface,
 				"vrf:" + metric.VRF,
 				"host_inf:" + metric.HostInf,
+				"dd.internal.resource:ndm_interface:" + s.namespace + ":" + deviceIP + ":" + metric.Interface,
```
TCheruy: looks like we're using a different ID format:

https://github.com/DataDog/datadog-agent/blob/42a3ff7ba48f7272ad4121a51b18c19da677433d/pkg/collector/corechecks/network-devices/versa/payload/interfaces.go#L77-L83

jedupau: yes good catch, this should work then 86bd2694656575ca1864c261b00b3ce8434a4b9e
I'll ask for a second review from ndm-integrations just to be sure, I am a bit surprised that the device_id is not contained in the interface_id 🤔

TCheruy: We build the interface id in the backend by concatenating `device_id:raw_id_type-raw_id` I believe, we might want to defer this logic to integrations though

TCheruy: So the resource tag should be `dd.internal.resource:ndm_interface:device_id:versa_interface-raw_id `

jedupau: oh yes sorry i was missing this thanks
1ce179c567872c99fdb3a39f5dff00adceb33c4a

TCheruy (approved):

PR #39237, opened 2025-07-24
feat(gitlab): Reduce the cache compression level

chouetz:
### What does this PR do?
Change the cache compression level

### Motivation
CI duration possible improvement. I've seen in the #performance-improvement channel
```
Enabling the flag allows Gitlab to use a faster compression library with basically no downsides. IMO, it should be the default.
```

CI pipeline is ok


Ishirui (approved):
Let's keep an eye on the CI time after this gets merged, to try and gauge if we are actually gaining anything (seems weird to me that using `slowest` compression would _accelerate_ things, unless we make up the time lost by uploading less data to the cache)

PR #39238, opened 2025-07-24
[autodiscovery/providers/prometheus] Migrate to workloadmeta

davidor:
### What does this PR do?

This PR migrates the prometheus autodiscovery provider to workloadmeta.

This provider was querying the Kubelet directly, but there's really no reason for that.

Other autodiscovery providers and parts of the code are using workloadmeta to fetch the pods and in the end, we'd like to have a single part that queries the Kubelet (the workloadmeta collector) and the rest should rely on workloadmeta.


zhuminyi:
The file is still named as prometheus_kubelet but kubelet is not used in the file. We could add some comment in the file to explain this change.

davidor:
@zhuminyi I've renamed the file to avoid confusion

wdhif (approved):

clamoriniere (approved):

PR #39244, opened 2025-07-24
[CWS/CSPM] Allow excluding containers

lebauce:
### What does this PR do?

Allow excluding containers from compliance and runtime security products.

### Motivation

This PR allows excluding containers from compliance and runtime security
monitoring in the same way than infrastructure monitoring.

### DIFF
--- a/pkg/security/module/server.go
+++ b/pkg/security/module/server.go
@@ -237,7 +237,11 @@ func (a *APIServer) dequeue(now time.Time, cb func(msg *pendingMsg) bool) {
 			return true
 		}
 
-		if msg.skip || msg.retry >= maxRetry {
+		if msg.skip {
+			return true
+		}
+
+		if msg.retry >= maxRetry {
 			seclog.Errorf("failed to sent event, max retry reached: %d", msg.retry)
 			return true
 		}
--- a/pkg/security/module/server.go
+++ b/pkg/security/module/server.go
@@ -41,6 +41,7 @@ import (
 	"github.com/DataDog/datadog-agent/pkg/security/seclog"
 	"github.com/DataDog/datadog-agent/pkg/security/serializers"
 	"github.com/DataDog/datadog-agent/pkg/security/utils"
+	"github.com/DataDog/datadog-agent/pkg/util/containers"
 	"github.com/DataDog/datadog-agent/pkg/util/fargate"
 	"github.com/DataDog/datadog-agent/pkg/util/log"
 	"github.com/DataDog/datadog-agent/pkg/util/startstop"
@@ -70,6 +71,7 @@ type pendingMsg struct {
 	extTagsCb       func() ([]string, bool)
 	sendAfter       time.Time
 	retry           int
+	skip            bool
 }
 
 func (p *pendingMsg) isResolved() bool {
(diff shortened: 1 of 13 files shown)


safchain:
The metric part is missing ?

hush-hush (approved):

0intro (approved):

rahulkaukuntla (approved):

pkg/security/module/server.go
```diff
@@ -234,7 +237,7 @@ func (a *APIServer) dequeue(now time.Time, cb func(msg *pendingMsg) bool) {
 			return true
 		}
 
-		if msg.retry >= maxRetry {
+		if msg.skip || msg.retry >= maxRetry {
```
YoannGh: Shouldn't we avoid logging an error when the event is dropped because of filters?
Also, what do you think about adding a metric for the number of events that were dropped because of this?

pkg/util/containers/filter.go
```diff
@@ -215,6 +216,18 @@ func GetPauseContainerFilter() (*Filter, error) {
 	return NewFilter(GlobalFilter, nil, excludeList)
 }
 
+// NewContainerFilter returns a new include/exclude filter for containers
+func NewContainerFilter(cfg model.Config, prefix string) (*Filter, error) {
```
sblumenthal: I am having a hard time at first glance from function signatures and function docstrings understanding the difference between the `NewFilter` function already in place and the `NewContainerFilter` function being added here.

Additionally, given the non-package-local usages of `NewFilter` ([example](https://github.com/DataDog/datadog-agent/blob/5383aaec518b598f3877ec485f39eb69fcd97e75/pkg/clusteragent/admission/mutate/cwsinstrumentation/cws_instrumentation.go#L338), [another example](https://github.com/DataDog/datadog-agent/blob/363dd62acefa0caed191dd0272f41e94a1fa1dbc/pkg/sbom/collectors/collectors.go#L104)) to do essentially what is being done here, I wonder if it makes sense for this function to live in this package?

If it DOES make sense for it to live here, I don't think the name is clear enough regarding what it is supposed to do given the other functions already in place in this file

sblumenthal (changes requested):

pkg/util/containers/filter.go
lebauce: I thought it would make sense to put this helper in the general pkg/utils/containers package but I'll move it somewhere else

sblumenthal (approved):

gabedos:
FYI: we are in the process of migrating the container infrastructure monitoring filtering mechanism to a new model.

You can checkout this [doc here](https://datadoghq.atlassian.net/wiki/spaces/CONTP/pages/5050598441/Introducing+a+Unified+Workload+Filtering+Component) for more context.

There's various PRs merged and in progress to facilitate this transition. If you're interested you could use the `workloadFilter` component and we will work this quarter to expose the more powerful filtering capabilities for users to configure.

https://github.com/DataDog/datadog-agent/pull/38738, https://github.com/DataDog/datadog-agent/pull/38474, https://github.com/DataDog/datadog-agent/pull/38298, https://github.com/DataDog/datadog-agent/pull/36591

pkg/security/module/server.go
lebauce: I fixed the log. I suggest adding the metric in a follow up PR

PR #39245, opened 2025-07-24
[CONTP-898] Use cluster tagger in KSM check to tag metrics with namespace labels and annotations as tags

adel121:
### What does this PR do?

This PR enriches KSM metrics with tags based on namespace labels and annotations as tags in the case where the KSM doesn't have access to namespace metatata by leveraging the remote tagger and the cluster tagger.

Notes:

In follow up PRs, we will:
- Update the operator / helm chart to automatically enable the cluster tagger in the dca and the remote tagger in the clc runner automatically for KSM when namespace labels and/or annotations as tags are configured. 

### Motivation

Be able to tag KSM metrics with namespace tags based on labels and annotations as tags via the cluster tagger without needing to have access to the namespace labels and annotations.

NOTE: 

E2E tests are implemented in the current PR and [in test-infra-definitions.](https://github.com/DataDog/test-infra-definitions/pull/1640)

Manual Testing.

Deploy the agent with the following setup that runs KSM check in a cluster check runner using pod collector only.

Be sure to:
- Activate the cluster tagger.
- Add namespace labels and/or annotations as tags to activate collection of namespace metadata in workloadmeta and cluster tagger.
- Activate usage of remote tagger in the cluster check runner by setting DD_CLC_RUNNER_REMOTE_TAGGER_ENABLED env var to true

```
datadog:

  kubelet:
    tlsVerify: false

  clusterTagger:
    # datadog.clusterTagger.collectKubernetesTags -- Enables Kubernetes resources tags collection.
    collectKubernetesTags: true

  clusterChecksRunner:
    enabled: true

  env:
    - name: DD_HOSTNAME
      valueFrom:
        fieldRef:
          fieldPath: spec.nodeName

  kubernetesResourcesLabelsAsTags:
    namespaces:
      namespace-team: team

  kubeStateMetricsCore:
    useClusterCheckRunners: true
                  
clusterAgent:
  confd:
    kubernetes_state_core.yaml: |-
      cluster_check: true
      init_config:
      instances:
        - collectors:
          - pods
          skip_leader_election: true
      
clusterChecksRunner:
  enabled: true
  env:
    - name: DD_CLC_RUNNER_REMOTE_TAGGER_ENABLED
      value: "true"
```

Create a sample pod or deployment in a namespace having labels or annotations matching the config. 

In this case, we created a namespace named `dev` and added `namespace-team: contp` label to match the config.

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
  namespace: dev
  labels:
    app: nginx
spec:
  replicas: 3  # Number of pods
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
        - name: nginx
          image: nginx:latest
          ports:
            - containerPort: 80

```

When everything is set, run the check:

```
kubectl exec <clc-runner-pod-name> -- agent check kubernetes_state_core --table -r
```

You should be able to see `team:contp` tag on emitted metrics.
Example: 
```
  kubernetes_state.container.restarts                gauge  1753433997  0          kube_cluster_name:adel-test-csi, cluster_name:adel-test-csi, uid:6598dd7b-15b4-450f-b391-e979b60c6945, kube_container_name:nginx, kube_deployment:nginx-deployment, kube_replica_set:nginx-deployment-96b9d695, node:kind-control-plane, orch_cluster_id:ec21f839-04ae-482b-9d9c-9c6a2fdc955b, pod_name:nginx-deployment-96b9d695-7m8sq, pod_phase:Running, team:contp          
```


jeremy-hanna (approved):
👍 for agent-runtime files, looks like some lint failures in the ksm/stub.go

GustavoCaso:
@adel121, any reason not to include the e2e test as part of this PR? I think splitting the logic and the test into separate PR is not the standard way of doing things

GustavoCaso (changes requested):
@adel121, any reason not to include the e2e test as part of this PR? I think splitting the logic and the test into separate PR is not the standard way of doing things. Imagine that we need to revert this changes if we have single PR it would be easier, but if we have logic and test in separate PR we are going t o have to revert multiple PRs. 

Could we add them as part of this PR?

test/new-e2e/tests/containers/k8s_test.go
```diff
@@ -935,6 +929,7 @@ func (suite *k8sSuite) TestKSM() {
 				`^kube_instance_tag:static$`,                            // This is applied via KSM core check instance config
 				`^stackid:` + regexp.QuoteMeta(suite.clusterName) + `$`, // Pulumi applies this via DD_TAGS env var
 			},
+			AcceptUnexpectedTags: true,
```
adel121: This is necessary because when the KSM check runs as a cluster check, the dispatcher attaches `cluster_name` tag in the configuration, so we end up generating `cluster_name` in addition to `kube_cluster_name`

pkg/collector/corechecks/cluster/ksm/kubernetes_state.go
```diff
@@ -736,8 +742,14 @@ func (k *KSMCheck) hostnameAndTags(labels map[string]string, labelJoiner *labelJ
 	// generate a dedicated tags slice
 	tagList := make([]string, 0, len(labels)+len(labelsToAdd))
 
-	ownerKind, ownerName := "", ""
+	ownerKind, ownerName, resourceNamespace := "", "", ""
+
 	for key, value := range labels {
+
+		if key == namespaceKey {
```
lavigne958, line 749: can we have multiple occurrences where `key == "namespace"`  ?
meaning if that exists we'll override the namespace on the next occurence.

adel121: I don't think we can have multiple occurrences (unless if the user decides to set this up to result with a conflict on purpose, not even sure this is possible, cc @sblumenthal )

I assume it is not an issue because we are already following the same pattern for `createdByKindKey` and `ownerKindKey`.

lavigne958: I see, we already have the same pattern for other tags, so that's good for me thanks.

test/new-e2e/tests/containers/k8s_test.go
sblumenthal: Does the test-infra-def change result in any new namespace-related tags from the cluster tagger showing up on the metrics? If so, could we add them to the expected tags?

adel121: No it doesn't result with new tags. 

The same tag is used (it is the `team` tag).

sblumenthal: But I don't see `team` in the expected list, so does it make sense to add it?

sblumenthal: Oh, I see, it is for `kubernetes_state.deployment.replicas_available`

pkg/collector/corechecks/cluster/ksm/kubernetes_state.go
```diff
@@ -780,6 +797,20 @@ func (k *KSMCheck) hostnameAndTags(labels map[string]string, labelJoiner *labelJ
...
+	if resourceNamespace != "" {
+		namespaceTags, tagErr = k.tagger.Tag(types.NewEntityID(types.KubernetesMetadata, string(util.GenerateKubeMetadataEntityID("", "namespaces", "", resourceNamespace))), types.LowCardinality)
+	}
+
+	if tagErr != nil {
+		log.Errorf("failed to get namespace tags for %q from tagger: %v", resourceNamespace, tagErr)
+	} else {
+		log.Debugf("obtained tags for namespace %q from tagger: %v", resourceNamespace, namespaceTags)
+		tagList = append(tagList, namespaceTags...)
+	}
```
GustavoCaso, line 812:
```suggestion
if resourceNamespace != "" {
	namespaceTags, err := k.tagger.Tag(types.NewEntityID(types.KubernetesMetadata, string(util.GenerateKubeMetadataEntityID("", "namespaces", "", resourceNamespace))), types.LowCardinality)
	
	if err != nil {
		log.Errorf("failed to get namespace tags for %q from tagger: %v", resourceNamespace, err)
	} else {
		log.Debugf("obtained tags for namespace %q from tagger: %v", resourceNamespace, namespaceTags)
		tagList = append(tagList, namespaceTags...)
	}
}
```

Nitpick sugggestion 😄

test/new-e2e/tests/containers/k8s_test.go
adel121: Also for this

https://github.com/DataDog/datadog-agent/blob/47f81a040ac360b227d6750425a45e590af77be9/test/new-e2e/tests/containers/k8s_test.go#L906

test/new-e2e/tests/containers/k8s_test.go
```diff
@@ -388,14 +388,6 @@ func (suite *k8sSuite) testClusterAgentCLI() {
...
 		suite.Contains(stdout, "Collector")
 		suite.Contains(stdout, "Running Checks")
-		suite.Contains(stdout, "kubernetes_state_core")
-		if suite.Env().Agent.FIPSEnabled {
-			// Verify FIPS mode is reported as enabled
-			suite.Contains(stdout, "FIPS Mode: enabled", "Cluster agent should report FIPS Mode as enabled")
-			suite.T().Logf("Cluster agent correctly reports FIPS status as enabled")
-		} else {
-			suite.NotContains(stdout, "FIPS Mode: enabled", "Cluster agent should not report FIPS Mode as enabled")
-		}
```
GustavoCaso: Is the removal of these assertions intentional?

GustavoCaso (approved):
Left a few question but otherwise the changes look good

test/new-e2e/tests/containers/k8s_test.go
```diff
@@ -430,24 +429,14 @@ func (suite *k8sSuite) testClusterAgentCLI() {
...
-	suite.Run("cluster-agent checkconfig", func() {
-		stdout, stderr, err := suite.Env().KubernetesCluster.KubernetesClient.PodExec("datadog", leaderDcaPodName, "cluster-agent", []string{"datadog-cluster-agent", "checkconfig"})
-		suite.Require().NoError(err)
-		suite.Empty(stderr, "Standard error of `datadog-cluster-agent checkconfig` should be empty")
-		suite.Contains(stdout, "=== kubernetes_state_core check ===")
-		suite.Contains(stdout, "Config for instance ID: kubernetes_state_core:")
-		if suite.T().Failed() {
-			suite.T().Log(stdout)
-		}
-	})
```
sblumenthal: Instead of removing this check, can you replace `kubernetes_state_core` with `kubernetes_apiserver`?

That check only runs on the cluster agent, is running by default, and we know is running because k8s events show up from the CI clusters. The goal of this test was to test the cli command, not to test kubernetes_state_core

test/new-e2e/tests/containers/k8s_test.go
```diff
@@ -388,7 +388,6 @@ func (suite *k8sSuite) testClusterAgentCLI() {
 		suite.Empty(stderr, "Standard error of `datadog-cluster-agent status` should be empty")
 		suite.Contains(stdout, "Collector")
 		suite.Contains(stdout, "Running Checks")
-		suite.Contains(stdout, "kubernetes_state_core")
```
sblumenthal, line 391: Likewise, can you replace this with `kubernetes_apiserver`?

sblumenthal (approved):

test/new-e2e/tests/containers/k8s_test.go
sblumenthal: Wouldn’t it just be safer to add
```
`^cluster_name:` + regexp.QuoteMeta(suite.clusterName) + `$`,
```
to the 4 tests, instead of just saying we don’t want to know when new tags are added? If only `cluster_name` is the new tag?

wdhif (approved):

ofek (approved):

PR #39246, opened 2025-07-24
[Cloud Run Jobs] Enhanced Metrics

nhulston:
### What does this PR do?

- Fix flushing enhanced metrics at the end of a Cloud Run Job. The container shuts down immediately after execution is done, so we need to force flush dogstatsd
- Rename Cloud Run Jobs start+end metrics to `gcp.run.job.enhanced.task.started` and `gcp.run.job.enhanced.task.ended` to clarify that each metric is per task, not per execution. And better align with the naming convention set by Step Functions
- Add a `gcp.run.job.enhanced.task.duration` metric in unit milliseconds
- Due to cyclic imports, I had to do some refactoring. I removed the `originToMetricSource()` method and instead, each cloudservice implements a `GetSource()` method
- I added default descriptions and units for these metrics: https://github.com/DataDog/integrations-internal-core/pull/1926

Screenshot 2025-07-24 at 12 50 02 PM
Screenshot 2025-07-24 at 12 49 49 PM
Screenshot 2025-07-25 at 12 43 00 PM

### Motivation

Better enhanced metrics for Cloud Run Jobs

Tested manually and verified that all metrics are delivered with the correct values and tags.


cmd/serverless-init/main.go
```diff
@@ -106,10 +106,16 @@ func run(_ secrets.Component, _ autodiscovery.Component, _ healthprobeDef.Compon
...
 	origin := cloudService.GetOrigin()
 
-	metric.AddShutdownMetric(prefix, origin, metricAgent.GetExtraTags(), time.Now(), metricAgent.Demux)
+	metric.Add(cloudService.GetShutdownMetricName(), 1.0, origin, metricAgent.GetExtraTags(), time.Now(), metricAgent.Demux)
+
+	// Add task duration metric for Cloud Run Jobs
+	if crj, ok := cloudService.(*cloudservice.CloudRunJobs); ok {
+		metricName, duration := crj.GetTaskDuration()
+		metric.Add(metricName, duration, origin, metricAgent.GetExtraTags(), time.Now(), metricAgent.Demux)
+	}
```
nhulston: This is messy, but I don't think there's a better way to do this. Open to suggestions!

I tried creating a `Shutdown()` method that each service implements, but we can't import the `metricAgent` due to a circular import.

pkg/aggregator/demultiplexer_serverless.go
```diff
@@ -111,7 +113,8 @@ func (d *ServerlessDemultiplexer) Stop(flush bool) {
 
 // ForceFlushToSerializer flushes all data from the time sampler to the serializer.
 func (d *ServerlessDemultiplexer) ForceFlushToSerializer(start time.Time, waitForSerializer bool) {
-	d.forceFlushToSerializer(start, waitForSerializer, false)
+	_, forceFlushAll := os.LookupEnv(cloudservice.CloudRunJobNameEnvVar)
+	d.forceFlushToSerializer(start, waitForSerializer, forceFlushAll)
 }
```
nhulston: This is not optimal, but I can't think of any other way to do this. Unless we modify `ForceFlushToSerializer` in the `Demultiplexer` interface, which affects the entire agent, not just serverless

Update: I also tried enabling `dogstatsd_flush_incomplete_buckets`, but that did not solve the problem

pkg/aggregator/demultiplexer_serverless.go
```diff
@@ -7,6 +7,8 @@ package aggregator
 
 import (
 	"context"
+	"github.com/DataDog/datadog-agent/cmd/serverless-init/cloudservice"
```
vickenty: 1) please don't import `cmd` from `pkg`, these tend to cause import cycles at unopportunate times.

2) we structure imports in three blocks built-ins, third party, and agent code. You can see example of this pattern in this file. Please follow it.

nhulston: Ah bad timing, but I just removed that cyclic import in the previous commit

cmd/serverless-init/cloudservice/cloudrun_jobs.go
```diff
@@ -79,14 +86,32 @@ func (c *CloudRunJobs) GetPrefix() string {
...
+func (c *CloudRunJobs) Shutdown(metricAgent serverlessMetrics.ServerlessMetricAgent) {
+	metricName := fmt.Sprintf("%s.enhanced.task.duration", c.GetPrefix())
+	duration := float64(time.Since(c.startTime).Milliseconds())
+	metric.Add(metricName, duration, c.GetSource(), metricAgent.GetExtraTags(), time.Now(), metricAgent.Demux)
+}
+
 // GetStartMetricName returns the metric name for container start events
 func (c *CloudRunJobs) GetStartMetricName() string {
-	return fmt.Sprintf("%s.enhanced.start", c.GetPrefix())
+	return fmt.Sprintf("%s.enhanced.task.started", c.GetPrefix())
```
nhulston: renaming to follow step functions convention

cmd/serverless-init/cloudservice/cloudrun_jobs.go
```diff
@@ -79,14 +86,32 @@ func (c *CloudRunJobs) GetPrefix() string {
...
+}
+
+// GetShutdownMetricName returns the metric name for container shutdown events
+func (c *CloudRunJobs) GetShutdownMetricName() string {
+	return fmt.Sprintf("%s.enhanced.task.ended", c.GetPrefix())
```
nhulston: same here

iadjivon (approved):
All set from Docs!

cmd/serverless-init/cloudservice/cloudrun_jobs.go
```diff
@@ -79,14 +86,32 @@ func (c *CloudRunJobs) GetPrefix() string {
...
 
-// Init is empty for CloudRunJobs
+// GetSource returns the metrics source
+func (c *CloudRunJobs) GetSource() metrics.MetricSource {
+	return metrics.MetricSourceGoogleCloudRunEnhanced
+}
+
+// Init records the start time for CloudRunJobs
 func (c *CloudRunJobs) Init() error {
+	c.startTime = time.Now()
```
apiarian-datadog, line 92: nit: should we be using now? does go have a monotonic now that we can use instead? (especially important for long-running things that may cross time savings boundaries and other wall clock shenanigans.)

cmd/serverless-init/cloudservice/cloudrun_jobs.go
```diff
@@ -79,14 +86,32 @@ func (c *CloudRunJobs) GetPrefix() string {
...
 	return nil
 }
 
+// Shutdown submits the task duration metric for CloudRunJobs
+func (c *CloudRunJobs) Shutdown(metricAgent serverlessMetrics.ServerlessMetricAgent) {
+	metricName := fmt.Sprintf("%s.enhanced.task.duration", c.GetPrefix())
+	duration := float64(time.Since(c.startTime).Milliseconds())
+	metric.Add(metricName, duration, c.GetSource(), metricAgent.GetExtraTags(), time.Now(), metricAgent.Demux)
```
apiarian-datadog: side question: what are these extra tags and the demux? that seems like a lower-level metrics agent thing leaking up into the this metric call. i'm concerned since i'm looking into tagging for metrics in general and this looks like one more place to surprise us. if it's not easy to do now, can we look into shifting some of this logic and putting a ticket on the backlog for it?

cmd/serverless-init/cloudservice/cloudrun_jobs.go
apiarian-datadog: i guess since this is a new metric and jobs support is still in flux, it's okay that we are changing the name? we don't need to do any sort of aliasing or such anywhere?

cmd/serverless-init/cloudservice/service.go
```diff
@@ -23,11 +27,20 @@ type CloudService interface {
...
 	Init() error
 
+	// Shutdown cleans up the CloudService
+	Shutdown(agent serverlessMetrics.ServerlessMetricAgent)
+
 	// GetStartMetricName returns the metric name for start events
 	GetStartMetricName() string
+
+	// GetShutdownMetricName returns the metric name for shutdown events
+	GetShutdownMetricName() string
```
apiarian-datadog, line 37: nit: we're starting to get a mix of metrics things and lifecycle things and labeling things in this interface. maybe not now, but would it make sense to separate things out a bit more? do we have, for example, metrics approaches that are the same for different combinations of services?

cmd/serverless-init/main.go
```diff
@@ -106,10 +106,10 @@ func run(_ secrets.Component, _ autodiscovery.Component, _ healthprobeDef.Compon
 	cloudService, logConfig, traceAgent, metricAgent, logsAgent := setup(modeConf, tagger, compression)
 
 	err := modeConf.Runner(logConfig)
-	prefix := cloudService.GetPrefix()
-	origin := cloudService.GetOrigin()
 
-	metric.AddShutdownMetric(prefix, origin, metricAgent.GetExtraTags(), time.Now(), metricAgent.Demux)
+	metric.Add(cloudService.GetShutdownMetricName(), 1.0, cloudService.GetSource(), metricAgent.GetExtraTags(), time.Now(), metricAgent.Demux)
+	cloudService.Shutdown(*metricAgent)
```
apiarian-datadog: nit: should the startup and shutdown metrics stuff be part of the init and shutdown functions?

cmd/serverless-init/cloudservice/cloudrun_jobs.go
```diff
@@ -79,14 +86,32 @@ func (c *CloudRunJobs) GetPrefix() string {
...
+// Init records the start time for CloudRunJobs
 func (c *CloudRunJobs) Init() error {
+	c.startTime = time.Now()
 	return nil
 }
 
+// Shutdown submits the task duration metric for CloudRunJobs
+func (c *CloudRunJobs) Shutdown(metricAgent serverlessMetrics.ServerlessMetricAgent) {
+	metricName := fmt.Sprintf("%s.enhanced.task.duration", c.GetPrefix())
+	duration := float64(time.Since(c.startTime).Milliseconds())
```
apiarian-datadog, line 99: is this a value that our customers could also pull from the trace stats for their code? are we including this in case a customer isn't running a tracer?

cmd/serverless-init/cloudservice/cloudrun_jobs.go
nhulston: Yes, we should be good. I've added the metric descriptions & units based on the new values in this PR

cmd/serverless-init/cloudservice/cloudrun_jobs.go
nhulston: To be extra safe, we can do
```go
c.startTime = time.Now().UTC()

// later in Shutdown()
duration := float64(time.Now().UTC().Sub(c.startTime).Milliseconds())
```

and i tested again, the values are still correct:
Screenshot 2025-07-25 at 3 06 27 PM

cmd/serverless-init/cloudservice/cloudrun_jobs.go
nhulston: I'm not really sure why they're called "extra", but they're a combination of tags [defined by the cloud service implementation](https://github.com/DataDog/datadog-agent/blob/e078b8bb2a51dd11275759ddfe9f8e2d84682e74/cmd/serverless-init/cloudservice/cloudrun_jobs.go#L44-L53) and [any custom tags set by the user](https://github.com/DataDog/datadog-agent/blob/cedfd68a55bc86be6edefb391a124ae9791f1d63/pkg/config/utils/tags.go#L12-L30).

Edit: I think this is easy to refactor here

nhulston: ok i just pushed a commit that really simplifies the `metric.Add()` implementation

cmd/serverless-init/main.go
nhulston: Every service submits a metric at startup and shutdown, so I think it makes sense to leave them as they are now.

But on the other hand, if we did move the metrics to the init and shutdown functions, we could get rid of the `GetStartMetricName()` and `GetShutdownMetricName()`. WDYT?

cmd/serverless-init/cloudservice/cloudrun_jobs.go
nhulston: Yes, these metrics get sent even without the tracer

cmd/serverless-init/cloudservice/service.go
nhulston: All the services except for Cloud Run Jobs submit a `*.enhanced.cold_start` and `*.enhanced.shutdown` metric

vickenty (approved):

cmd/serverless-init/cloudservice/cloudrun_jobs.go
apiarian-datadog: o, actually, looks like this is not necessary in go: https://pkg.go.dev/time#hdr-Monotonic_Clocks (see also https://en.wikipedia.org/wiki/Leap_second re: UTC)

cmd/serverless-init/cloudservice/service.go
apiarian-datadog: sure, but i mean that there are two distinct concepts hiding inside this service interface. metadata about the service and its operations, and related but separate metrics stuff.

cmd/serverless-init/main.go
apiarian-datadog: i can see arguments both ways, so i'll leave it to your judgement re: the resulting code if we have a lot of duplicated code, it probably makes sense to leave it in here.

cmd/serverless-init/cloudservice/service.go
nhulston: Yeah, this should be refactored in the future

cmd/serverless-init/main.go
```diff
@@ -106,10 +106,10 @@ func run(_ secrets.Component, _ autodiscovery.Component, _ healthprobeDef.Compon
 	cloudService, logConfig, traceAgent, metricAgent, logsAgent := setup(modeConf, tagger, compression)
 
 	err := modeConf.Runner(logConfig)
-	prefix := cloudService.GetPrefix()
-	origin := cloudService.GetOrigin()
 
-	metric.AddShutdownMetric(prefix, origin, metricAgent.GetExtraTags(), time.Now(), metricAgent.Demux)
+	metric.Add(cloudService.GetShutdownMetricName(), 1.0, cloudService.GetSource(), *metricAgent)
+	cloudService.Shutdown(*metricAgent)
```
apiarian-datadog: can we defer this call? or does it need to happen before the lastFlush call? actually, can we defer the lastFlush?

pkg/aggregator/demultiplexer_serverless.go
```diff
@@ -111,7 +112,8 @@ func (d *ServerlessDemultiplexer) Stop(flush bool) {
 
 // ForceFlushToSerializer flushes all data from the time sampler to the serializer.
 func (d *ServerlessDemultiplexer) ForceFlushToSerializer(start time.Time, waitForSerializer bool) {
-	d.forceFlushToSerializer(start, waitForSerializer, false)
+	_, forceFlushAll := os.LookupEnv("CLOUD_RUN_JOB")
```
apiarian-datadog, line 115: i'm concerned that we're still doing this. this is a thing that will inevitably be forgotten or missed. since the ServerlessDemultiplexer has serverless in the name, we could, for example, move this into a parameter we pass to `InitAndStartServerlessDemultiplexer`. that way this cloud run check can happen up in our setup method and thread its way down through the serverless metrics package.

cmd/serverless-init/cloudservice/cloudrun_jobs.go
apiarian-datadog, line 99: is milliseconds the standard unit for our durations? (it's a float, so it could theoretically be a fractional second, right?)

cmd/serverless-init/cloudservice/cloudrun.go
```diff
@@ -140,16 +142,29 @@ func (c *CloudRun) GetPrefix() string {
...
+func (c *CloudRun) Shutdown(serverlessMetrics.ServerlessMetricAgent) {}
+
 // GetStartMetricName returns the metric name for container start (coldstart) events
 func (c *CloudRun) GetStartMetricName() string {
 	return fmt.Sprintf("%s.enhanced.cold_start", c.GetPrefix())
 }
 
+// GetShutdownMetricName returns the metric name for container shutdown events
+func (c *CloudRun) GetShutdownMetricName() string {
+	return fmt.Sprintf("%s.enhanced.shutdown", c.GetPrefix())
```
apiarian-datadog: nit: for the implementations that are "standard" it may be worth adding a `func standardShutdownMetricName(s Service) string` that we can call here. but i also see the value in keeping each service implementation local and explicit, so we can leave these as they are, since we're probably not going to be making sweeping changes across the services.

Though, by that same logic, we can probably hard-code the prefix in here since that's a fixed string. Or, rather, turn the prefix into a string constant. Do we call `GetPrefix` outside of the local cloud service packages? If not, we can drop the method in favor of a local string constant.

apiarian-datadog (changes requested):

cmd/serverless-init/main.go
nhulston: It does need to go before `lastFlush`, but we can defer `lastFlush`

(verified by testing manually)

cmd/serverless-init/cloudservice/cloudrun_jobs.go
nhulston: Yes, we could pick any unit -- i just went with ms. Should it be changed?

I had a PR here to set the unit to milliseconds: https://github.com/DataDog/integrations-internal-core/pull/1926
so we'd have to update that repo

pkg/aggregator/demultiplexer_serverless.go
nhulston: `InitAndStartServerlessDemultiplexer` isn't used by serverless-init; it's only used in `pkg/serverless/` (the Lambda extension)

The code path here is: 
```
serverless-init/main.go run() ->
main.go lastFlush() ->
main.go flushAndWait() ->
serverless/metrics/metric.go Flush() ->
comp/dogstatsd/server/server.go ServerlessFlush() ->
pkg/aggregator/demultiplexer_serverless.go ForceFlushToSerializer()  
```

Yes, I think this is messy and needs to be refactored, but I'm starting to think that that is out of the scope of this PR

cmd/serverless-init/cloudservice/cloudrun.go
nhulston: Nice catch! `GetPrefix` is not used elsewhere and can be converted into a string constant

cmd/serverless-init/cloudservice/cloudrun_jobs.go
apiarian-datadog: no, that's fine. if milliseconds are the thing, we can use milliseconds. just wanted to confirm if this was our standard duration unit.

cmd/serverless-init/main.go
apiarian-datadog: (we can defer multiple things and they defer in LIFO order like a stack https://go.dev/blog/defer-panic-and-recover )

apiarian-datadog (approved):
ok. if we're not going to fix the demultiplexer hack now, let's do it in a quick followup. at least to move the "is this cloudrun" boolean logic out of the demultiplexer file and up into the serverless-init main and/or cloudservice.

cmd/serverless-init/main.go
nhulston: Awesome, i'll defer both then

pkg/aggregator/demultiplexer_serverless.go
nhulston: Will be merged in a following PR
https://github.com/DataDog/datadog-agent/commit/e228be7c86a5a7482e2976fd0528710a1b045fdc#diff-863a106a2a0859344dfa5a216f1729a858b03f15ea12d3c506774caa132f792a

PR #39247, opened 2025-07-24
[dyninst] Don't fail hard on esoteric types, and "_" fields 

piob-io:


pkg/dyninst/actuator/errors.go
```diff
@@ -26,3 +26,17 @@ func (e *NoSuccessfulProbesError) Error() string {
...
+	for i := 0; i < min(len(e.Issues), 3); i++ {
+		fmt.Fprintf(f, "%s: %s", e.Issues[i].Kind.String(), e.Issues[i].Message)
+		if i < len(e.Issues)-1 {
+			fmt.Fprintf(f, ", ")
+		}
+	}
+	if len(e.Issues) > 3 {
+		fmt.Fprintf(f, "...")
+	}
+}
```
ajwerner: I think we should say something in this string that there were no successful probes, and the number of issues (at least if it's truncated)

pkg/dyninst/irgen/type_catalog.go
```diff
@@ -394,6 +395,11 @@ structChildren:
 					"failed to get name for member %q: %w", name, err,
 				)
 			}
+			// Special field that we don't currently have use for,
+			// but would need special handling to deal with duplicates.
+			if strings.HasPrefix(name, "_") {
```
ajwerner: this is sadly wrong: only exactly `_` is ignored. `_foo` is totally fine... Let's add that to the test and make sure we capture it.

https://go.dev/play/p/dN20_KSyn9f

ajwerner (changes requested):

pkg/dyninst/actuator/errors.go
piob-io: Done

pkg/dyninst/irgen/type_catalog.go
piob-io: Gtk, done

ajwerner (approved):
LGTM

piob-io:
Unfortunately we need to include these fields for type memory layout calculation. Renamed StructureType.Fields to RawFields to hint at some complexity with accessing them, and provided a Fields() helper to filter out the "_" fields.

ajwerner (approved):
LGTM

PR #39253, opened 2025-07-24
dyninst/symdb: support filtering out 3rd-party symbols

andreimatei:
For purposes of saving on both resources while processing DWARF and the
amount of data produced for SymDB, we'll want to restrict the symbols we
collect to 1st party code (at least optionally). This patch teaches the
extractor to distinguish 1st party vs 3rd party. For binaries with
buildinfo (i.e. not built by Bazel), we look at the URL of the main
module and select the github org. For Bazel binaries, we exclude some
file prefixes.

### DIFF
--- a/pkg/dyninst/symdb/symdb.go
+++ b/pkg/dyninst/symdb/symdb.go
@@ -88,9 +88,6 @@ func (f Function) empty() bool {
 
 // SerializationOptions defines options for serializing symbols.
 type SerializationOptions struct {
-	// OnlyMainModule indicates that only the main module should be serialized.
-	OnlyMainModule bool
-
 	PackageSerializationOptions
 }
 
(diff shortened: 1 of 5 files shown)


pkg/dyninst/symdb/symdb.go
```diff
@@ -582,6 +649,17 @@ func (b *SymDBBuilder) ExtractSymbols() (Symbols, error) {
 			continue
 		}
 
+		if opt != ExtractScopeAllSymbols {
+			name, ok := entry.Val(dwarf.AttrName).(string)
+			if !ok {
+				return Symbols{}, errors.New("compile unit without name")
+			}
+			if !pkgPassesFilter(name) {
```
piob-io: This won't fly for inlined functions. Feel free to keep this as-is, and I can move this around when adding inlining support; this will obviously regress performance somewhat...

The way it needs to be for inlined functions, is that we pass this package filter down to explore subprogram function. Then we use it to determine whether that specific subprogram (regular or abstract), should be captured. Regardless if it is captured or not, we explore children, because we need to visit inlined subroutines descendants (but can skip capturing variables).

piob-io (approved):
Thank you!

pkg/dyninst/symdb/symdb.go
andreimatei: I figured that a 1st party function cannot be inlined into a 3rd party one, so we can skip whole 3rd party compile units, no?

piob-io: Duh, indeed, sorry -.-

PR #39264, opened 2025-07-25
[autodiscovery/prometheus] Return empty configs when prometheus port annotation has no matching container

davidor:
### What does this PR do?

This PR fixes a bug in the Prometheus autodiscovery provider where incorrect configurations were generated when the `prometheus.io/port` annotation specified a port that no container exposed.

When a pod had the `prometheus.io/port` annotation but no container declared that port, the provider would generate openmetrics check configurations for all containers in the pod, using the annotated port in the URL. This resulted in multiple failing checks since none of the containers were actually listening on the specified port.

I added unit tests that cover the bug.

I also tested manually. I enabled prometheus scraping (`datadog.prometheusScrape.enabled` in the Helm chart) and created a Prometheus deployment with the `prometheus.io/port` annotation.

### DIFF
--- a/comp/core/autodiscovery/common/utils/prometheus_kubelet.go
+++ b/comp/core/autodiscovery/common/utils/prometheus_kubelet.go
@@ -29,18 +29,22 @@ func ConfigsForPod(pc *types.PrometheusCheck, pod *kubelet.Pod) []integration.Co
 	instances, found := buildInstances(pc, pod.Metadata.Annotations, namespacedName)
 	if found {
 		// If `prometheus.io/port` annotation has been provided, let’s keep only the container that declares this port.
-		var containerUsingThePort string
-		if portFromAnnotationString, portFromAnnotationFound := pod.Metadata.Annotations[types.PrometheusPortAnnotation]; portFromAnnotationFound {
-			if portFromAnnotation, err := strconv.Atoi(portFromAnnotationString); err == nil {
-			containerFound:
-				for _, containerSpec := range pod.Spec.Containers {
-					for _, port := range containerSpec.Ports {
-						if port.ContainerPort == portFromAnnotation {
-							containerUsingThePort = containerSpec.Name
-							break containerFound
-						}
-					}
-				}
+		portAnnotationString, hasPortAnnotation := pod.Metadata.Annotations[types.PrometheusPortAnnotation]
+		var containerWithPortInAnnotation string
+
+		if hasPortAnnotation {
+			portNumber, err := strconv.Atoi(portAnnotationString)
+			if err != nil {
+				log.Debugf("Port in annotation '%s' is not an integer", portAnnotationString)
+				// Don't return configs with an invalid port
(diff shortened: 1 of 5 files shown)


comp/core/autodiscovery/common/utils/prometheus_kubelet.go
```diff
@@ -49,9 +53,11 @@ func ConfigsForPod(pc *types.PrometheusCheck, pod *kubelet.Pod) []integration.Co
 				log.Debugf("Container '%s' doesn't match the AD configuration 'kubernetes_container_names', ignoring it", containerStatus.Name)
 				continue
 			}
-			if containerUsingThePort != "" && containerStatus.Name != containerUsingThePort {
+
+			if hasPortAnnotation && containerStatus.Name != containerWithPortInAnnotation {
```
clamoriniere, line 56: can we return an error in case of name mismatch so it can be reported in the `agent status` and in the futur in Fleet Automation?

davidor: I added a new commit that collects errors about mismatched or incorrect port.

These errors are now returned in `GetConfigErrors()` (which always returned empty before) and now they're shown in the "Autodiscovery" section of the `agent status` command.

clamoriniere (approved):

releasenotes/notes/contp-903-prometheus-ad-no-matching-port-3be7c25adfddec23.yaml
```diff
@@ -0,0 +1,14 @@
...
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+fixes:
+  - |
+    When the option ``prometheus_scrape.enabled`` is set to true and the pod has
+    a ``prometheus.io/port`` annotation, the agent no longer schedules
+    ``openmetrics`` checks for the containers that don't expose that port in
+    their spec.
```
estherk15: Added a suggestion for rewording to make the sentence clearer.

```suggestion
When the ``prometheus_scrape.enabled`` option is set to true and the pod has a ``prometheus.io/port`` annotation, containers must expose that port in their spec in order for the Agent to schedule ``openmetrics`` checks.
```

estherk15 (approved):

releasenotes/notes/contp-903-prometheus-ad-no-matching-port-3be7c25adfddec23.yaml
davidor: Thanks :+1:
I've changed it

demikl:
Hi, this change is a BREAKING CHANGE for us since the scraped port is not defined in the `ports` section of any container of our pods. With this change, Datadog has stopped from scraping our metrics on this port.

davidor:
@demikl This PR should fix the issue: https://github.com/DataDog/datadog-agent/pull/42619

PR #39273, opened 2025-07-25
dyninst/loader: strip pt_regs relocation information

ajwerner:
This relocation information is not needed for pt_regs offsets -- they are only needed if you want to be architecture independent. By avoiding these relocations, we can avoid loading all the kernel btf which is ~10MiB, and surely not fast.

Fixes [DEBUG-4221](https://datadoghq.atlassian.net/browse/DEBUG-4221).

[DEBUG-4221]: https://datadoghq.atlassian.net/browse/DEBUG-4221?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/dyninst/loader/loader.go
+++ b/pkg/dyninst/loader/loader.go
@@ -11,11 +11,12 @@ package loader
 import (
 	"errors"
 	"fmt"
-	"io"
 	"path/filepath"
 	"unsafe"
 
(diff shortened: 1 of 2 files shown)


piob-io (approved):
Neat!

PR #39277, opened 2025-07-25
[AGENTCFG-351] Avoid passing in the rds.Config struct into SetWithoutSource

rahulkaukuntla:
### What does this PR do?
We are moving our configuration management system from viper to nodetreemodel, which doesn't allow structs to be passed into `SetWithoutSource` (done [here](https://github.com/DataDog/datadog-agent/blob/853fdeea6979e371eacda2a7022ba6291f211eec/comp/core/autodiscovery/listeners/dbm_rds_test.go#L178)). As such, the `config` field in the `DBMRdsListener` struct (which is what is passed into SetWithoutSource) needs to be converted from a `rds.Config` object to a general `map[string]interface{}` object.

Unit testing is sufficient.


comp/core/autodiscovery/listeners/dbm_rds.go
```diff
@@ -29,7 +30,7 @@ type DBMRdsListener struct {
 	delService   chan<- Service
 	stop         chan bool
 	services     map[string]Service
-	config       rds.Config
```
dustmop: The primary goal is to fix `SetWithoutSource`, by not passing struct data to it, instead only giving it plain types (map / slice / scalars). It shouldn't be necessary to modify this type in non-test code, as it loosens the type-safety which is undesirable. Instead, could you keep these non-test files the same, and modify [the constructor here](https://github.com/DataDog/datadog-agent/blob/3f3ad3c10f4d39748fcaef09a58cb09cb7cd31f7/comp/core/autodiscovery/listeners/dbm_rds_test.go#L182) so that the test's plain data gets serialized into `rds.Config`.

dustmop (changes requested):

comp/core/autodiscovery/listeners/dbm_rds.go
rahulkaukuntla: You're right--just changing the test is a lot easier 😅

comp/core/autodiscovery/listeners/dbm_rds_test.go
```diff
@@ -179,7 +180,9 @@ func TestDBMRdsListener(t *testing.T) {
 			mockAWSClient := aws.NewMockRdsClient(ctrl)
 			tc.rdsClientConfigurer(mockAWSClient)
 			ticks := make(chan time.Time, 1)
-			l := newDBMRdsListener(tc.config, mockAWSClient, ticks)
+			var newRdsConfig rds.Config
+			_ = mapstructure.Decode(tc.config, &newRdsConfig)
```
dustmop:
```suggestion
			err := mapstructure.Decode(tc.config, &newRdsConfig)
			assert.NoError(t, err)
```

dustmop (approved):
LGTM, with one small suggestion

comp/core/autodiscovery/listeners/dbm_rds_test.go
```diff
@@ -179,7 +180,10 @@ func TestDBMRdsListener(t *testing.T) {
 			mockAWSClient := aws.NewMockRdsClient(ctrl)
```
gabedos, line 180: Nit: If I'm understanding this correctly, can we remove this from the test case if we are using a different source for the config now?
https://github.com/DataDog/datadog-agent/blob/ab2dce4ced357adb44d37c63675f0626c97ca874/comp/core/autodiscovery/listeners/dbm_rds_test.go#L178-L179

gabedos (approved):

comp/core/autodiscovery/listeners/dbm_rds_test.go
rahulkaukuntla: These two lines aren't for one test case--they are part of the test code that is shared across test cases, so we can't just remove one test case (we would need to remove the entire test). So I will merge this PR as-is.

PR #39279, opened 2025-07-25
[AGENTCFG-264] Secret backend test uses 1 binary instead of 6

dustmop:
### What does this PR do?

Previously, this test would compile multiple executable binaries in order to test the secret backend command. This is both (1) inefficient and (2) flaky. This PR replaces the multiple executables with a single binary while still covering the same functionality. 

~~Additionally, the test files contain prebuilt binaries for a couple of platforms that we support, and the test will use this if it matches the OS and CPU architecture, invoking the go compiler only if no prebuilt binary is found.~~

### Motivation

Fix test flakiness.

`dda inv test -o --module=comp/core/secrets/secretsimpl --targets=. --test-run-name=TestExecCommandError -v`

~~On platforms with a prebuilt binary, there is a log line:
`using prebuilt secret backend binary 'test/prebuilt/linux_amd64/test_command'`~~

~~Storing binaries in the repo is suboptimal, but is an improvement of invoking the go compiler multiple times.~~

### Additional Notes

~~Consider adding an invoke task to build the test binary for a given platform, if there is demand for it.~~


rahulkaukuntla (approved):
Nice!

PR #39280, opened 2025-07-25
dyninst/symdb: add support to the symdb cli for tracing

andreimatei:
Add a flag to the cli to capture an execution trace while extracting
symbols. Useful for seeing how the heap size evolves.


ajwerner (approved):
LGTM

PR #39284, opened 2025-07-28
[AGENTRUN-586] Add grpc metrics

louis-cqrl:
### What does this PR do?

This PR introduces a new set of built-in **gRPC telemetry metrics** for the Datadog Agent.

Specifically:
- Adds server and client-side gRPC interceptors that emit telemetry.
- Provides metrics for request counts, errors, latencies, payload sizes, and active requests.
- Adds utility wrappers in `pkg/util/grpc` to simplify interceptor setup.
- Updates all internal gRPC server instantiations to use the new unified metrics-aware interceptors.
- Introduces `ServerOptionsWithMetricsAndAuth` and related helper functions to combine metrics and authentication logic safely.
- Adds new telemetry definitions in `metrics.go` and logic for collecting them in `interceptors.go`.

The following metrics are now available:
| Metric Name                     | Type      | Description                                     |
| ------------------------------ | --------- | ----------------------------------------------- |
| `grpc.request_count`           | Counter   | Total number of gRPC requests processed         |
| `grpc.error_count`             | Counter   | Total number of gRPC errors encountered         |
| `grpc.request_duration_seconds`| Histogram | Distribution of gRPC request latencies          |
| `grpc.payload_size_bytes`      | Histogram | Distribution of payload sizes for gRPC calls    |

### Motivation

- Improve visibility into internal gRPC communications across Agent components.
- Enable better performance monitoring and debugging of gRPC-based data paths.
- Lay the groundwork for future enhancements like dynamic latency analysis, error tracking, or profiling by exposing low-level RPC metrics.
- Standardize interceptor setup across the codebase, reducing risk of misconfiguration.

- **Unit Tests**: Added coverage for the new interceptors.
- **Manual Validation**:
  - Started various Agent components locally with interceptors enabled.
  - Observed telemetry output from active gRPC streams (via logs and metrics sink).
- **Regression Coverage**: Ensured no breaking changes to existing gRPC auth or TLS configuration logic.

- **Telemetry Overhead**: Introducing metrics collection may introduce slight CPU/memory cost. However, metrics are batched and histogram buckets are optimized for typical gRPC call durations and sizes.

### Additional Notes

- The new `pkg/util/grpc` package is designed to be used consistently across components; this will help future initiatives such as unified observability.
- No behavior is changed in existing gRPC endpoints besides the added telemetry.
- This PR is part of ongoing efforts to improve internal instrumentation under the `COAT` initiative.


guyarb (approved):

paulcacheux (approved):

davidor (approved):
:+1: for container-platform files (just 1 with a minor change)

val06 (approved):
LGTM for files owned by ebpf-platform team

releasenotes/notes/add-grpc-metric-COAT-02e917d92e7ec5c4.yaml
```diff
@@ -0,0 +1,19 @@
...
+other:
+  - |
+    Added the following metrics for gRPC communication:
+    | Metric Name                     | Type      | Description                                     |
+    | ------------------------------- | --------- | ----------------------------------------------- |
+    | `grpc.request_count`            | Counter   | Total number of gRPC requests processed         |
+    | `grpc.error_count`              | Counter   | Total number of gRPC errors encountered         |
+    | `grpc.request_duration_seconds` | Histogram | Distribution of gRPC request latencies          |
+    | `grpc.payload_size_bytes`       | Histogram | Distribution of payload sizes for gRPC calls    |
+    | `grpc.active_requests`          | Gauge     | Number of currently active (in-flight) requests |
```
maycmlee:
```suggestion
    | `grpc.request_count`            | Count   | Total number of gRPC requests processed.         |
    | `grpc.error_count`              | Count   | Total number of gRPC errors encountered.         |
    | `grpc.request_duration_seconds` | Histogram | Distribution of gRPC request latencies.          |
    | `grpc.payload_size_bytes`       | Histogram | Distribution of payload sizes for gRPC calls.    |
    | `grpc.active_requests`          | Gauge     | Number of currently active (in-flight) requests. |
```

maycmlee (approved):
Just a couple of small suggestions, but approving to not block merge

pkg/util/grpc/metrics.go
```diff
@@ -0,0 +1,35 @@
...
+
+	// PayloadSize tracks distribution of payload sizes for gRPC calls
+	PayloadSize = telemetry.NewHistogram("grpc", "payload_size_bytes",
+		[]string{"service_method", "peer", "direction"}, "Distribution of payload sizes for gRPC calls",
+		[]float64{100, 500, 1000, 5000, 10000, 50000, 100000, 500000, 1000000, 5000000, 10000000})
+
+	// ActiveRequests tracks number of currently active (in-flight) requests
+	ActiveRequests = telemetry.NewGauge("grpc", "active_requests",
+		[]string{"service_method", "peer"}, "Number of currently active (in-flight) requests")
+)
```
pgimalac: Do they need to be exported ? I don't believe they're used from outside the package (and I don't think they should)

releasenotes/notes/add-grpc-metric-COAT-02e917d92e7ec5c4.yaml
```diff
@@ -0,0 +1,19 @@
...
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+
+other:
+  - |
+    Added the following metrics for gRPC communication:
```
pgimalac: We might want to make it clear that it's agent telemetry, ie. that it's to debug the agent, it doesn't give information about other software running on the host

pgimalac (approved):
LGTM, just some minor nitpicks

daniel-taf (approved):

pkg/util/grpc/metrics.go
louis-cqrl: Done

releasenotes/notes/add-grpc-metric-COAT-02e917d92e7ec5c4.yaml
louis-cqrl: Good point

pkg/util/grpc/interceptors.go
```diff
@@ -0,0 +1,347 @@
...
+// getStatusCode returns the gRPC status code as a string
+func getStatusCode(err error) string {
+	if err == nil {
+		return "OK"
+	}
+
+	if st, ok := status.FromError(err); ok {
+		return st.Code().String()
+	}
+	return "UNKNOWN"
```
misteriaud: **🥜 nitpick:** ‏`UNKNOWN` is in fact [an existing gRPC status code](https://pkg.go.dev/google.golang.org/grpc@v1.75.0/codes#Code). Maybe you should rename it by `UNDEFINED`. 
```
const (
[...]
	// Unknown error. An example of where this error may be returned is
	// if a Status value received from another address space belongs to
	// an error-space that is not known in this address space. Also
	// errors raised by APIs that do not return enough error information
	// may be converted to this error.
	//
	// The gRPC framework will generate this error code in the above two
	// mentioned cases.
	Unknown Code = 2
[...]
)
```

pkg/util/grpc/interceptors.go
```diff
@@ -0,0 +1,347 @@
...
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		start := time.Now()
+
+		// Extract metadata
+		serviceMethod := extractMethodInfo(info.FullMethod)
+		peer := extractPeerInfo(ctx)
+
+		// Track active requests
+		activeRequests.Add(1, serviceMethod, peer)
+		defer activeRequests.Add(-1, serviceMethod, peer)
```
misteriaud: **💬 suggestion:** ‏Since these request are meant to be short live, They generally use timeouts in the order of seconds. Compared to the frequency of the COAT payload generation (~10minutes), this would be completely transparent.
We probably should only rely on `requestCount` for this type of requests IMO.

pkg/util/grpc/interceptors.go
```diff
@@ -0,0 +1,347 @@
...
+
+		// Track payload size if available
+		if req != nil {
+			if size, ok := req.(interface{ Size() int }); ok {
+				sizeBytes := float64(size.Size())
+				payloadSize.Observe(sizeBytes, serviceMethod, peer, "request")
+			}
+		}
+
+		// Call the handler
```
misteriaud, line 97: **💬 suggestion:** ‏You should probably move the `start := time.Now()` here to not be impacted by the potential time spend to retrieve request metadata before truly handling the request.

pkg/util/grpc/interceptors.go
```diff
@@ -0,0 +1,347 @@
...
+		// Record metrics
+		duration := time.Since(start).Seconds()
+		statusCode := getStatusCode(err)
+		errorCode := getErrorCode(err)
+
+		requestCount.Inc(serviceMethod, peer, statusCode)
+		requestDuration.Observe(duration, serviceMethod, peer)
+
+		if err != nil {
+			errorCount.Inc(serviceMethod, peer, errorCode)
```
misteriaud, line 109: **💬 suggestion:** ‏Maybe you could compute the `errorCode := getErrorCode(err)` here, only `if err != nil`.

pkg/util/grpc/interceptors.go
```diff
@@ -0,0 +1,347 @@
...
+			log.Debugf("gRPC client error: service_method=%s, peer=%s, error=%v", serviceMethod, peer, err)
+		}
```
misteriaud: Same comment here as for UnaryServerInterceptor:
- move the `start := time.Now()` just before the handler
- remove the `activeRequests` counting
- move the `errorCode := getErrorCode(err)` in the if statement below

pkg/util/grpc/interceptors.go
```diff
@@ -0,0 +1,347 @@
...
+	ctx := s.ClientStream.Context()
+
+	// Decrement active requests when the stream ends
+	go func() {
+		<-ctx.Done()
+		activeRequests.Add(-1, s.serviceMethod, s.peer)
+	}()
+
+	return ctx
+}
```
misteriaud: **🔨 warning:** ‏I think that with this implement, if the function `metricsClientStream.Context()` is called multiple times, the `activeRequests` will decrease multiple times instead of once.

pkg/util/grpc/metrics.go
```diff
@@ -0,0 +1,35 @@
...
+		[]float64{.001, .005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10, 30, 60})
+
+	// payloadSize tracks distribution of payload sizes for gRPC calls
+	payloadSize = telemetry.NewHistogram("grpc", "payload_size_bytes",
+		[]string{"service_method", "peer", "direction"}, "Distribution of payload sizes for gRPC calls",
+		[]float64{100, 500, 1000, 5000, 10000, 50000, 100000, 500000, 1000000, 5000000, 10000000})
+
+	// activeRequests tracks number of currently active (in-flight) requests
+	activeRequests = telemetry.NewGauge("grpc", "active_requests",
+		[]string{"service_method", "peer"}, "Number of currently active (in-flight) requests")
```
misteriaud: **💬 suggestion:** As said above, we should probably only track the stream and not the unary requests. So maybe rename that `activeStreams`.‏

internal/remote-agent/main.go
misteriaud, line 1: **💭 thought:** this main is used for testing purpose with the RemoteAgentRegistry only. It is not supposed to send metrics to COAT. So I'm not sure it's relevant instrument it.‏

pkg/trace/api/otlp.go
misteriaud, line 1: **❓ question:** This server is used to receive OTLP traces from customer workloads, it's not meant to exchange data between Agent processes. With that said, is it relevant to monitor it with the same metric collectors (counter. histogram, ...) than those for the IPC servers‏ ?

misteriaud (commented):
Nice work ! 🎉
It looks good overall, there are just some potential improvement that can be made, and some behaviors to clarify IMO.

pkg/util/grpc/interceptors.go
louis-cqrl: Good catch, fixed

pkg/util/grpc/interceptors.go
louis-cqrl: removed this metric

pkg/util/grpc/interceptors.go
louis-cqrl: fixed it

pkg/util/grpc/interceptors.go
louis-cqrl: good catch, done

pkg/util/grpc/interceptors.go
louis-cqrl: I think I did not forget one, if you can check again to be sure ?

pkg/util/grpc/interceptors.go
louis-cqrl: metric removed so it should be okay

pkg/util/grpc/metrics.go
louis-cqrl: metric removed for this one

internal/remote-agent/main.go
louis-cqrl: Good catch, when i did this i was mainly an addition to all places where gRPC was called

pkg/trace/api/otlp.go
louis-cqrl: Good catch to remove it since we already have the otlp metrics

ogaca-dd (approved):

misteriaud (approved):

iglendd (approved):

PR #39289, opened 2025-07-28
Also make `curl` retry in macOS install script (ABLD-18 follow-up)

rdesgroppes:
### What does this PR do?
This aims at making the script slightly more resistant to transient network errors, similarly to what's been done otherwise, for instance:
- https://github.com/DataDog/datadog-agent/blob/877f00e9765b2e6bd9f95a6238cd3438cf684799/cmd/agent/install_script.sh#L493
- https://github.com/DataDog/datadog-agent/blob/877f00e9765b2e6bd9f95a6238cd3438cf684799/pkg/fleet/installer/setup/install.sh#L60

The proposed number of retries is here of `2`, which means at most `3` attempts will be made. (instead of `1`)

Also, the shebang specification was missing; here-added `#!/usr/bin/env bash` is consistent with [instructions](https://app.datadoghq.eu/fleet/install-agent/latest?platform=macos):
```sh
[...] bash -c "$(curl -L https://install.datadoghq.com/scripts/install_mac_os.sh)"
```
### Motivation
- [Everything Curl](https://everything.curl.dev/usingcurl/downloads/retry.html#retry)
- [What are the http codes to automatically retry the request?](https://stackoverflow.com/questions/51770071/what-are-the-http-codes-to-automatically-retry-the-request)

- [x]  Reviewers: could you please check your **original** macOS-provided `curl` (**not** a later `brew`-installed one) indeed supports the `--retry` flag? (yes according to the [manpage from 2012](https://www.unix.com/man_page/osx/1/curl/), but better safe than sorry...)

### Additional Notes
Transient error means either: a timeout, an HTTP 408, 429, 500, 502, 503 or 504 response code.

### DIFF
--- a/cmd/agent/install_mac_os.sh
+++ b/cmd/agent/install_mac_os.sh
@@ -1,3 +1,4 @@
+#!/usr/bin/env bash
 # Unless explicitly stated otherwise all files in this repository are licensed
 # under the Apache License Version 2.0.
 # This product includes software developed at Datadog (https://www.datadoghq.com/).
@@ -5,7 +6,7 @@
 
 # Datadog Agent install script for macOS.
 set -e
-install_script_version=1.5.0
+install_script_version=1.5.1
 dmg_file=/tmp/datadog-agent.dmg
 dmg_base_url="https://s3.amazonaws.com/dd-agent"
 etc_dir=/opt/datadog-agent/etc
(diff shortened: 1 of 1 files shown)


fanny-jiang (approved):

aiuto (approved):
The change you made looks good.

But WOW, the overall script could use some TLC. There are subtle things like `find_latest_patch_verstion_for` which grep's out file names from a blob of XML.
Without some comments or unit tets, I don't have faith it will pick the right version when it sees 
```
datadog-agent-7.16.0-beta.1-1.dmg
datadog-agent-7.16.0-1.dmg
```
It does not matter for this CL, but someone should work on robustness of this, and it does not seem like a job for the Build team. But that's another question.

rdesgroppes:
> But WOW, the overall script could use some TLC. There are subtle things like `find_latest_patch_verstion_for` which grep's out file names from a blob of XML. Without some comments or unit tets, I don't have faith it will pick the right version when it sees
> 
> ```
> datadog-agent-7.16.0-beta.1-1.dmg
> datadog-agent-7.16.0-1.dmg
> ```
> 
> It does not matter for this CL, but someone should work on robustness of this, and it does not seem like a job for the Build team. But that's another question.

I'm with you... I didn't feel at ease so I limited changes I brought there to the bare minimum by running snippet on Alpine's `bash`/POSIX tools without GNU extensions.
But, yes, this deliverable definitely needs to be taken care of.

cmd/agent/install_mac_os.sh
```diff
@@ -63,13 +64,14 @@ if [ -n "$DD_AGENT_MINOR_VERSION" ]; then
 fi
 
 arch=$(/usr/bin/uname -m)
+curl_retries=(--retry 2)
```
aiuto, line 67: Verified that native install curl on recent Macos (curl 8.7.1) supports retry flag

rdesgroppes:
I got the following confirmation: (in addition to the above)
> - `curl 7.12.3` was released in 2004, well before macOS 10.5 (Leopard, 2007).
> - Therefore, all macOS versions since Leopard include a `curl` binary that supports `--retry` and related flags.

PR #39293, opened 2025-07-28
Retry once on any error, except for test lint e2e test

KevinFairise2:
### What does this PR do?

Change our strategy to improve reliability of the CI. We will now retry all the jobs once when they fail. Except the jobs that can be expected to fail like unit tests, linters and e2e tests.
Ideally we'd like tests and linters to be retried on main because they are not expected to fail there. However this is not possible to have such a precise configuration with current gitlab configuration.

What would be possible:
- Retry everything on main + never retry by default on PRs
- Have a separate service that watches the CI and is responsible for retries of the jobs. This would allow us to configure it as we want

### Motivation

Improve reliability and reduce flakiness of the CI

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -41,18 +41,7 @@ include:
 
 default:
   retry:
-    max: 2
-    exit_codes:
-      - 42
-      - 101 # Failed to extract dependencies
-    when:
-      - runner_system_failure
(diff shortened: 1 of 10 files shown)


Ishirui (approved):
LGTM ! Thanks for this

julien-lebot (approved):

PR #39296, opened 2025-07-28
[CWS] Separate each connect test for consistant CI

theop-dd:
### Motivation
Reduce the number of failing tests.

### DIFF
--- a/pkg/security/tests/connect_test.go
+++ b/pkg/security/tests/connect_test.go
@@ -34,18 +34,6 @@ func TestConnectEvent(t *testing.T) {
 			ID:         "test_connect_af_inet",
 			Expression: `connect.addr.family == AF_INET && process.file.name == "syscall_tester"`,
 		},
-		{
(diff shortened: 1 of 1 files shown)


pkg/security/tests/connect_test.go
```diff
@@ -85,6 +73,22 @@ func TestConnectEvent(t *testing.T) {
...
+		},
+	}
+
+	test, err := newTestModule(t, nil, ruleDefs)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer test.Close()
 
 	t.Run("io-uring-connect-af-inet-any-tcp-success", func(t *testing.T) {
```
paulcacheux: we don't need the `t.Run` anymore since you are splitting the tests

theop-dd: Oh yes you're right, copy paste issue ^^

paulcacheux (approved):

PR #39301, opened 2025-07-28
feat(rc): set User-Agent header for RC requests

domodwyer:
This will help identify requests in the RC backend at the transport level, suitable for traffic routing by L7 proxies.

---

* refactor: import tidying (68022a3620)
      
      Normalise imports into std, 3rd party, datadog.

* feat(rc): set User-Agent header for RC requests (7fa2dd8e21)
      
      This causes all Agent HTTP (and WebSocket handshake) requests to the
      remote config backend to present a user agent in the form of:
      
          datadog-agent/7.68.2 (go1.24.5)
      
      While the Agent does declare its version in some request payloads, this
      only works for those specific requests, and only if those payload can be
      successfully parsed.
      
      Setting the user agent on the HTTP request can be used to identify the
      version of the agent making a request to RC without attempting to parse
      the payload, prior to reading the body at all.

### DIFF
--- a/pkg/config/remote/api/http_test.go
+++ b/pkg/config/remote/api/http_test.go
@@ -13,9 +13,10 @@ import (
 	"testing"
 	"time"
 
-	"github.com/DataDog/datadog-agent/pkg/config/mock"
 	"github.com/gorilla/websocket"
 	"github.com/stretchr/testify/assert"
+
+	"github.com/DataDog/datadog-agent/pkg/config/mock"
 )
 
 // Test WebSocket connectivity under varying transport configurations, and
--- a/pkg/config/remote/api/http_test.go
+++ b/pkg/config/remote/api/http_test.go
@@ -10,9 +10,11 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"net/url"
+	"regexp"
 	"testing"
 	"time"
 
+	"github.com/coreos/go-semver/semver"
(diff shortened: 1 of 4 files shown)


pkg/config/remote/api/http_test.go
```diff
@@ -169,3 +170,41 @@ func TestNewWebSocket(t *testing.T) {
...
+
+	_, _ = client.FetchOrgData(ctx)
+
+	select {
+	case ua := <-userAgentCh:
+		// Regex explained:
+		//   * ^datadog-agent\\/ == must start with "datadog-agent/"
+		//   * (unknown|\\d+\\.\\d+\\.\\d+) == either "unknown" or a semver string
+		//   * \\(go\\d+\\.\\d+\\.\\d+\\)$ == ends in " (go1.2.3)" where 1.2.3 is a semver string
+		assert.Regexp("^datadog-agent\\/(unknown|\\d+\\.\\d+\\.\\d+) \\(go\\d+\\.\\d+\\.\\d+\\)$", ua)
```
domodwyer: I used regex to be insensitive to agent & go version changes.

It's a little hard to read because there's no raw strings in go?

domodwyer: Actually go does have raw strings! And this regex will fail for release candidates. Stand by 🫡

mellon85 (approved):

PR #39313, opened 2025-07-28
[dyninst/symdb] Produce function line layout upfront and include inlined functions

piob-io:
https://datadoghq.atlassian.net/browse/DEBUG-4241

This has two immediate effects:
 - we report more variable availability - previously if a function body was fully covered by another, inlined functions, excluding that range resulted in us finding no available lines; current approach digs through inlined pc ranges, to resolve lines for both inlined and inlinee functions.
 - performance benefit - mostly from resolving pc to line mapping once; on my local machine symbol extraction goes down from 23s to 5s; there is more room for improvement

The main purpose is that it will allow us to use the same mapping for resolving line locations of inlined functions, which will come in a separate pr.


pkg/dyninst/gosym/symtab.go
```diff
@@ -627,6 +625,191 @@ func (lt *lineTable) locatePC(pc uint64, goFuncData []byte) []GoLocation {
...
+	return
+}
+
+type fileRange struct {
+	pcLo uint64
+	pcHi uint64
+	file string
+}
+
+func (lt *lineTable) fileRanges(f *GoFunction) ([]fileRange, error) {
```
andreimatei, line 655: I think this function deserves a comment suggesting that there is such a thing as a "files table" (or whatever it's called) in gosym, which is separate from the lines table

pkg/dyninst/gosym/symtab.go
```diff
@@ -627,6 +625,191 @@ func (lt *lineTable) locatePC(pc uint64, goFuncData []byte) []GoLocation {
...
+			pcHi: index.PC,
+			name: name,
+			file: file,
+		})
+		pc = index.PC
+	}
+	return ranges
+}
+
+func (lt *lineTable) functionLines(pc uint64, goFuncData []byte) (map[string]FunctionLines, error) {
```
andreimatei: could this function and others like it take in the `*GoSymbolTable` instead of the `goFuncData`? It's not very clear what `goFuncData` is.

pkg/dyninst/gosym/symtab.go
```diff
@@ -627,6 +625,191 @@ func (lt *lineTable) locatePC(pc uint64, goFuncData []byte) []GoLocation {
...
+
+	res := make(map[string]FunctionLines, len(funcs))
+	upsert := func(name string, file string, lr LineRange) {
+		fl := res[name]
+		fl.File = file
+		fl.Lines = append(fl.Lines, lr)
+		res[name] = fl
+	}
+
+	// Iterate over lines, while tracking which functions the belong to.
```
andreimatei: something missing in this sentence

pkg/dyninst/gosym/symtab.go
```diff
@@ -627,6 +625,191 @@ func (lt *lineTable) locatePC(pc uint64, goFuncData []byte) []GoLocation {
...
+}
+
+type functionRange struct {
+	pcLo uint64
+	pcHi uint64
+	name string
+	file string
+}
+
+func (lt *lineTable) inlinedFunctionsMapping(f *GoFunction, goFuncData []byte) []functionRange {
```
andreimatei: give this function a comment

pkg/dyninst/gosym/symtab.go
```diff
@@ -627,6 +625,191 @@ func (lt *lineTable) locatePC(pc uint64, goFuncData []byte) []GoLocation {
...
+	inlineTree := f.funcInfo.funcData(funcdataInlTree, goFuncData)
+	if inlineTree == nil {
+		return nil
+	}
+
+	offset, found := f.funcInfo.pcDataStart(pcdataInlTreeIndex)
+	if !found {
+		return nil
+	}
+	offset += uint32(lt.pcTab[0])
```
andreimatei, line 707: can you put a comment on the `pcTab` field explaining what are the two values in it? Or perhaps replace that field with two fields with better names.

pkg/dyninst/gosym/symtab.go
```diff
@@ -627,6 +625,191 @@ func (lt *lineTable) locatePC(pc uint64, goFuncData []byte) []GoLocation {
...
+	pc := f.Entry
+	var ranges []functionRange
+	for index := range iterPCValues(lt.data[offset:], pc, lt.quantum) {
+		// [pc, index.PC) maps to inline function data at index.Val.
+		var name string
+		if index.Val < 0 {
+			name = f.Name
+		} else {
+			funcID, nameOff, _, found := readInlTree(inlineTree, uint32(index.Val))
+			if !found || funcID == lt.wrapperFuncID {
```
andreimatei, line 730: mind giving the `wrapperFuncID` field a comment that hints to what a "wrapper function" is?

pkg/dyninst/gosym/symtab.go
```diff
@@ -627,6 +625,191 @@ func (lt *lineTable) locatePC(pc uint64, goFuncData []byte) []GoLocation {
...
+	fileIdx := 0
+
+	names := map[uint32]string{}
+	pc := f.Entry
+	var ranges []functionRange
+	for index := range iterPCValues(lt.data[offset:], pc, lt.quantum) {
+		// [pc, index.PC) maps to inline function data at index.Val.
+		var name string
+		if index.Val < 0 {
+			name = f.Name
```
andreimatei: `index.Val < 0` means... we're back to the outer function? Please add a comment, even if it repeats the code :P

pkg/dyninst/gosym/symtab.go
```diff
@@ -627,6 +625,191 @@ func (lt *lineTable) locatePC(pc uint64, goFuncData []byte) []GoLocation {
...
+		return nil
+	}
+
+	files, err := lt.fileRanges(f)
+	if err != nil {
+		return nil
+	}
+	fileIdx := 0
+
+	names := map[uint32]string{}
```
andreimatei: s/names/functionNames ?
There's a lot of names everywhere and multiple functions with similar structures; it's easy to get lost. Let's make the names explicit where we can.

pkg/dyninst/gosym/symtab.go
andreimatei: also comment that results are sorted by pc

pkg/dyninst/gosym/symtab.go
```diff
@@ -627,6 +625,191 @@ func (lt *lineTable) locatePC(pc uint64, goFuncData []byte) []GoLocation {
...
+	pc := f.Entry
+	var ranges []functionRange
+	for index := range iterPCValues(lt.data[offset:], pc, lt.quantum) {
```
andreimatei: s/index/inlinedFuncEntry ?
It's not that easy to tell what we're iterating over here.

pkg/dyninst/gosym/symtab.go
```diff
@@ -105,6 +94,26 @@ func (gst *GoSymbolTable) LocatePC(pc uint64) []GoLocation {
...
+// LineRange represents a range of PC that map to the same line number.
+type LineRange struct {
+	PCLo uint64
+	PCHi uint64
+	Line uint32
+}
+
+// FunctionLines represents the lines of a function. Line ranges are non-overlapping
+// and sorted by the PCLo. Line number values may repeat across different pc ranges.
+type FunctionLines struct {
```
andreimatei, line 108: say that pcs corresponding to inlined functions are skipped

pkg/dyninst/gosym/symtab.go
```diff
@@ -740,96 +935,53 @@ func (lt *lineTable) pcValue(off uint32, entry uint64, targetPC uint64) (uint32,
...
-		return false
-	}
-	s.cursor = s.cursor[deltaBytes:]
-
-	var vdelta int32
-	if (uvdelta & 1) != 0 {
-		vdelta = int32(^(uvdelta >> 1))
-	} else {
-		vdelta = int32(uvdelta >> 1)
+func iterPCValues(cursor []byte, pc uint64, quantum uint32) iter.Seq[pcValue] {
```
andreimatei: nit: call the function argument `table` (I think that's what whatever little documentation exists on the gosym datastructures calles them?). You can have assign it to a local variable called `cursor` if you want.

andreimatei (approved):
nice!

pkg/dyninst/gosym/symtab.go
piob-io: Done

pkg/dyninst/gosym/symtab.go
piob-io: Done - changed the receiver.

pkg/dyninst/gosym/symtab.go
piob-io: Done

pkg/dyninst/gosym/symtab.go
piob-io: Done

pkg/dyninst/gosym/symtab.go
piob-io: Added comment for all the similiar fields. These are all offset ranges that each sub-table.

pkg/dyninst/gosym/symtab.go
piob-io: Yes, done, also added comment to the whole function, that hints at this special case too.

pkg/dyninst/gosym/symtab.go
piob-io: Fair, done

pkg/dyninst/gosym/symtab.go
piob-io: inlinedFuncIdx

pkg/dyninst/gosym/symtab.go
piob-io: Done

pkg/dyninst/gosym/symtab.go
piob-io: It's not a whole table though... It's a specific "sequence" inside the table (where other things reference beginning of the sequence, and end of the sequences is marked by a zero uvdelta. Renamed to pcValueSeq

piob-io:
Thank you!

PR #39337, opened 2025-07-29
[autodiscovery/scheduler/controller] Add health check

davidor:
### What does this PR do?

This PR adds a health check in the autodiscovery scheduler controller.

The controller is responsible for calling `Schedule` and `Unschedule` on registered schedulers. In rare cases, these operations can block. For example when there's a bug in a check’s `Configure` function. When this happens, the entire autodiscovery controller becomes blocked, preventing other checks from being scheduled or unscheduled.

To prevent this from going unnoticed, this change introduces a health check that will fail if the controller is blocked.

Manually introduced a `Sleep` call in a check’s `Configure` function to simulate the blocking scenario. Verified that the health check failure was triggered and reported in `agent health`.


comp/core/autodiscovery/scheduler/controller.go
```diff
@@ -9,13 +9,18 @@ import (
 	"sync"
 	"time"
 
+	"go.uber.org/atomic"
```
vboulineau: Built-in `atomic` package with structs is available now

comp/core/autodiscovery/scheduler/controller.go
```diff
@@ -36,6 +41,9 @@ type Controller struct {
 
 	started     bool
 	stopChannel chan struct{}
+
+	healthProbe                 *health.Handle
+	schedulerOperationStartTime *atomic.Int64 // Unix timestamp, 0 means no active operation
```
vboulineau: It assumes a single operation, thus a single worker, which is true ATM but should be highlighted.

comp/core/autodiscovery/scheduler/controller.go
davidor: There are docs in the repo that mention we should prefer that version over the built-in one: https://github.com/DataDog/datadog-agent/blob/main/docs/dev/atomics.md

I’m not sure if the reasons given there are still valid.

comp/core/autodiscovery/scheduler/controller.go
davidor: I added a comment to clarify this :+1:

comp/core/autodiscovery/scheduler/controller.go
vboulineau: Yeah, this was before built-in introduced `struct`, it should be removed. The alignement issue pertains to antiquated platforms that we do not support.

wdhif (approved):

comp/core/autodiscovery/scheduler/controller.go
davidor: ok, I changed it :+1:

vboulineau (approved):

releasenotes/notes/ad-scheduler-health-check-63e42f3e2d3d093f.yaml
```diff
@@ -0,0 +1,12 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+fixes:
+  - |
+    When the autodiscovery scheduler gets blocked, it will now trigger a health
```
urseberry:
```suggestion
    When the Autodiscovery scheduler gets blocked, it triggers a health
```

urseberry (approved):
Left a nonblocking grammar/formatting nit.

releasenotes/notes/ad-scheduler-health-check-63e42f3e2d3d093f.yaml
davidor: Thanks. Fixed :+1:

PR #39342, opened 2025-07-29
Restrict environment variables used for omnibus build

alopezz:
### What does this PR do?

It:
- Uses `replace_env=True` to block environment variables not defined in the `env` dictionary from being used on the Omnibus build ([see docs](https://docs.pyinvoke.org/en/stable/api/runners.html#invoke.runners.Runner.run)).
- To make that work, it defines explicit sets of variables that will be allowed to pass through the environment.
- It simplifies exclusion of variables when computing the omnibus git cache key to an explicit set of exclusions, without using prefixes / suffixes which allow too much freedom.

### Motivation

[ABLD-4](https://datadoghq.atlassian.net/browse/ABLD-4)

- We hope to reduce the number of ["cache mutation" events](https://app.datadoghq.com/event/explorer?query=source%3Aomnibus-cache%20%22omnibus%20cache%20mutated%22&cols=&messageDisplay=expanded-lg&options=&refresh_mode=sliding&sort=DESC&view=all&from_ts=1750922173263&start=1753365704533&end=1753366904533&to_ts=1753514173263&live=true&paused=false) that we get (partial cache hits that mess up with how our caching is set up).
- Explicitly listing variables that are allowed within the build gives us better control over what can affect the build, putting us in the path towards tighter control over build inputs.
- More accurate cache computation may result in more cache hits than before.

- Passing CI. Artifact sizes should not change, and build times should be similar to what we were seeing before.
- I made a pass through variables that got blocked as well as by variable usage inside omnibus such that there should hopefully be nothing missing that significantly affects the build.

### Additional Notes

[ABLD-4]: https://datadoghq.atlassian.net/browse/ABLD-4?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tasks/libs/common/omnibus.py
```diff
@@ -16,190 +16,113 @@
 CACHE_VERSION = 2
 
 
+ENV_PASSHTROUGH = {
+    'CI': 'dda may rely on this to be able to tell whether it\'s running on CI and adapt behavior',
```
rdesgroppes:
```suggestion
    'CI': "dda may rely on this to be able to tell whether it's running on CI and adapt behavior",
```

tasks/libs/common/omnibus.py
```diff
@@ -16,190 +16,113 @@
...
+    'DD_CC': 'Points at cc compiler',
```
rdesgroppes:
```suggestion
    'DD_CC': 'Points at c compiler',
```

tasks/libs/common/omnibus.py
```diff
@@ -16,190 +16,113 @@
...
+    'DD_CXX': 'Points at c++ compiler',
+    'DD_CMAKE_TOOLCHAIN': 'Points at cmake toolchain',
+    'DDA_NO_DYNAMIC_DEPS': 'ariable affecting dda behavior',
```
rdesgroppes:
```suggestion
    'DDA_NO_DYNAMIC_DEPS': 'Variable affecting dda behavior',
```

tasks/omnibus.py
```diff
@@ -153,26 +170,27 @@ def get_omnibus_env(
...
 
 
+def _passthrough_env_for_os(starting_env: dict[str, str], platform: str) -> dict[str, str]:
+    expected_env = set(ENV_PASSHTROUGH) | set(OS_SPECIFIC_ENV_PASSTHROUGH[platform])
+
+    missing_env = expected_env - set(starting_env)
+    if missing_env:
+        warnings.warn(
+            f'Missing expected environment variables for Omnibus build: {missing_env}',
+            stacklevel=1,
```
rdesgroppes, line 187: note to other reviewers: `ruff` used to complain without this default (`stacklevel=1`) being explicitly set. (limitation?)

tasks/omnibus.py
```diff
@@ -111,10 +128,10 @@ def get_omnibus_env(
...
         env['MACOSX_DEPLOYMENT_TARGET'] = '11.0' if os.uname().machine == "arm64" else '10.12'
 
-    if skip_sign:
-        env['SKIP_SIGN_MAC'] = 'true'
-    if hardened_runtime:
-        env['HARDENED_RUNTIME_MAC'] = 'true'
+        if skip_sign:
+            env['SKIP_SIGN_MAC'] = 'true'
+        if hardened_runtime:
+            env['HARDENED_RUNTIME_MAC'] = 'true'
```
rdesgroppes, line 134: Hum, either Python isn't strictly checking indentations, or this will eventually fail when checking indentations.

tasks/omnibus.py
```diff
@@ -94,6 +97,20 @@ def get_omnibus_env(
...
+            'WINDOWS_DDNPM_DRIVER',
+            'WINDOWS_DDNPM_VERSION',
+            'WINDOWS_DDNPM_SHASUM',
+            'WINDOWS_DDPROCMON_DRIVER',
+            'WINDOWS_DDPROCMON_VERSION',
+            'WINDOWS_DDPROCMON_SHASUM',
+        ]
+        for var in windows_only_vars:
+            if var in env:
+                del env[var]
```
rdesgroppes, line 112:
```suggestion
            env.pop(var, None)
```

tasks/libs/common/omnibus.py
```diff
@@ -16,190 +16,113 @@
...
+    },
+    'linux': {
+        'DEB_GPG_KEY': 'Used to sign packages',
+        'DEB_GPG_KEY_NAME': 'Used to sign packages',
+        'DEB_SIGNING_PASSPHRASE': 'Used to sign packages',
+        'RPM_GPG_KEY': 'Used to sign packages',
+        'RPM_GPG_KEY_NAME': 'Used to sign packages',
+        'RPM_SIGNING_PASSPHRASE': 'Used to sign packages',
+    },
+    'darwin': {},
```
rdesgroppes, line 82: I guess the macOS build would require, at least:
```py
    {
        "APPLE_ACCOUNT": "Used to sign packages",
        "CODESIGNING_CERT_BASE64": "Used to sign packages",
        "CODESIGNING_CERT_PASSPHRASE": "Used to sign packages",
        "INSTALLER_CERT_BASE64": "Used to sign packages",
        "INSTALLER_CERT_PASSPHRASE": "Used to sign packages",
        "MACOS_APPLE_APPLICATION_SIGNING": "Used to sign packages",
        "MACOS_APPLE_DEVELOPER_ACCOUNT": "Used to sign packages",
        "MACOS_APPLE_INSTALLER_SIGNING": "Used to sign packages",
        "MACOS_KEYCHAIN_PWD": "Used to sign packages",
        "NOTARIZATION_ATTEMPTS": "Used to sign packages",
        "NOTARIZATION_PWD": "Used to sign packages",
        "NOTARIZATION_TIMEOUT": "Used to sign packages",
        "NOTARIZATION_WAIT_TIME": "Used to sign packages",
        "SIGN": "Used to sign packages",
        "USER": "",
    }
```
(esp. the `SIGN` flag, but maybe some don't need to pass through `omnibus` because they're only referenced by the `build_agent_dmg.sh` script)

tasks/omnibus.py
alopezz: 🤷 I considered that, but it felt off to me to use `pop` and not using the value, but I'm fine changing it.

tasks/libs/common/omnibus.py
alopezz: I'm not sure we need those, at least not the signing / notarization related ones, given that:
- They seem to only be used in the [enclosing script](https://github.com/DataDog/datadog-agent/blob/main/.gitlab/package_build/build_agent_dmg.sh), not during the Omnibus.
- Notarization was [successful](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1054978152#L1400).

Not sure about `"USER"` either, I guess we could add it for both Linux and macOS though seeing that the build completes correctly I'm not sure it's really needed.

tasks/omnibus.py
alopezz: Why?

I indented this because it didn't make sense to me to set these variables when `sys.platform` is not `darwin` (as they refer to `MAC`).

tasks/omnibus.py
rdesgroppes: linters may indeed complain we're not using the value - was only a suggestion, I leave it up to you.

tasks/omnibus.py
rdesgroppes: Oh, I didn't realize the `if osx_target?` in `agent.rb`:
```ruby
  if osx_target?
    unless ENV['SKIP_SIGN_MAC'] == 'true'
      code_signing_identity 'Developer ID Application: Datadog, Inc. (JKFCB4CN7C)'
    end
    if ENV['HARDENED_RUNTIME_MAC'] == 'true'
      entitlements_file "#{files_path}/macos/Entitlements.plist"
    end
```
Resolving.

tasks/libs/common/omnibus.py
rdesgroppes: wanted to double-check. Looks fine, indeed! Resolving...

rdesgroppes (approved):
Thank you!

tasks/libs/common/omnibus.py
```diff
@@ -16,190 +16,115 @@
...
+        'SYSTEMROOT',
+        'SSL_CERT_FILE',
+        'TEMP',
+        'TMP',
+        'USERPROFILE',
+        'rvm_bin_path',
+        'rvm_path',
+        'rvm_prefix',
+        'rvm_version',
+    }
```
rdesgroppes, line 126: in the above list, whereas excluding temporary directory & cache management variables makes completely sense, I'm more concerned by variables that are -at least semantically- subject to alter the generated artifacts, for instance:
- signing related: wouldn't different keys/certificates lead to different signatures?
- compilation toolchain related: wouldn't a different `rvm` lead to a different `omnibus-ruby` behavior?

alopezz: I don't think it will matter much in practice (those variables seldom change), but I'm also not going to die on this hill. 
My thinking for these was:
- Signing happens always out of (after) everything that we currently consider cacheable is cached.
- rvm version would only affect ruby behavior if something's very broken with rvm. rvm is a "version manager" for ruby, meaning it's used as a way to get the version of ruby that we want; we should be getting the exact same version of ruby regardless of rvm version.

rdesgroppes: Good to go, then!

PR #39346, opened 2025-07-29
dyninst/uploader: log sometimes on error to upload

ajwerner:
If we fail to send batches, it'd be nice to log something about it. Uncovered that we weren't logging anything in the system-tests when the trace agent URL was misconfigured.

### DIFF
--- a/pkg/dyninst/uploader/batcher.go
+++ b/pkg/dyninst/uploader/batcher.go
@@ -13,6 +13,8 @@ import (
 	"sync"
 	"time"
 
+	"golang.org/x/time/rate"
+
 	"github.com/DataDog/datadog-agent/pkg/util/log"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/dyninst/uploader/batcher.go
```diff
@@ -64,6 +67,11 @@ func newBatcher(name string, sender sender, batcherConfig batcherConfig) *batche
 		state:        newBatcherState(batcherConfig),
 		timer:        timer,
 		sender:       sender,
+		// Used to rate-limit log messages about failed batches.
+		//
+		// TODO: Keep metrics for the number of failed batches and rate limit
+		// the actual attempts to send batches.
+		errLogLimiter: rate.NewLimiter(rate.Every(10*time.Second), 10),
```
piob-io: Is it useful to see a burst of 10 error messages together? If not, I'd limit bursts to 1.

piob-io (approved):

pkg/dyninst/uploader/batcher.go
ajwerner: Done.

PR #39354, opened 2025-07-29
feat: add new service discovery payload data

gengnamstyle:
### What does this PR do?
- Updates **process check** to include service discovery data for linux only
- config changes
  - process check
      - `IsEnabled()` considers service discovery process probe configuration `discovery.enabled` because we want the check to continue to submit service discovery data even when process collection is disabled
  - process collector
      - `isProcessCollectionEnabled()` now considers
          - `process_config.process_collection.enabled`
      - `isServiceDiscoveryEnabled()` removes hardcoded disabling of service discovery collection to now consider
          - `discovery.enabled`
  - discovery check is disabled when `process_config.process_collection.use_wlm` is enabled (since we don't both duplicate discovery data being sent and this makes it easier to compare)
        
## Phase 1 Configuration
image

## Phase 2 Temporary Config `use_wlm` removal + old service discovery check removal
image

### Motivation
- as part of the processes + service discovery consolidation, we are finally sending the new service discovery data to the backend via the process check. RFC - https://datadoghq.atlassian.net/wiki/spaces/cxg/pages/5013111768/RFC+Processes+and+Service+Discovery+Agent+Consolidation

- adds a unit test for the wlm implementation of `ProcessesByPID` that includes service discovery data
- adds a unit test for process collector configuration
- adds unit tests for process check data formatting
- run on ubuntu 24 vm and demonstrated that nginx shows service discovery data by viewing `sudo datadog-agent workload-list`

image

### DIFF
--- a/comp/core/workloadmeta/collectors/internal/process/process_collector.go
+++ b/comp/core/workloadmeta/collectors/internal/process/process_collector.go
@@ -135,14 +135,18 @@ func GetFxOptions() fx.Option {
 }
 
 // isProcessCollectionEnabled returns a boolean indicating if the process collector is enabled
+func (c *collector) isRunningInCoreAgent() bool {
+	return c.config.GetBool("process_config.run_in_core_agent.enabled")
+}
+
+// isProcessCollectionEnabled returns a boolean indicating if the process collector is enabled
 func (c *collector) isProcessCollectionEnabled() bool {
 	return c.config.GetBool("process_config.process_collection.use_wlm")
 }
 
 // isServiceDiscoveryEnabled returns a boolean indicating if service discovery is enabled
 func (c *collector) isServiceDiscoveryEnabled() bool {
-	// TODO: implement the logic to check if service discovery is enabled based on configuration
-	return false
+	return c.systemProbeConfig.GetBool("system_probe_config.enabled") && c.systemProbeConfig.GetBool("discovery.enabled")
 }
 
 // isLanguageCollectionEnabled returns a boolean indicating if language collection is enabled
@@ -160,6 +164,10 @@ func (c *collector) collectionIntervalConfig() time.Duration {
 // is done. It also gets a reference to the store that started it so it
 // can use Notify, or get access to other entities in the store.
 func (c *collector) Start(ctx context.Context, store workloadmeta.Component) error {
+	if !c.isRunningInCoreAgent() {
+		return errors.NewDisabled(componentName, "core process collection not running in core agent")
+	}
+
 	if !c.isProcessCollectionEnabled() && !c.isServiceDiscoveryEnabled() {
 		return errors.NewDisabled(componentName, "process collection and service discovery are disabled")
 	}
--- a/comp/core/workloadmeta/collectors/internal/process/process_collector.go
+++ b/comp/core/workloadmeta/collectors/internal/process/process_collector.go
@@ -136,14 +136,12 @@ func GetFxOptions() fx.Option {
 
 // isProcessCollectionEnabled returns a boolean indicating if the process collector is enabled
 func (c *collector) isProcessCollectionEnabled() bool {
-	return c.config.GetBool("process_config.process_collection.enabled") &&
-		// TODO: process_config.process_collection.use_wlm is temporary and will eventually be removed
-		c.config.GetBool("process_config.process_collection.use_wlm")
+	return c.config.GetBool("process_config.process_collection.enabled")
 }
 
 // isServiceDiscoveryEnabled returns a boolean indicating if service discovery is enabled
 func (c *collector) isServiceDiscoveryEnabled() bool {
-	return c.systemProbeConfig.GetBool("system_probe_config.enabled") && c.systemProbeConfig.GetBool("discovery.enabled")
+	return c.systemProbeConfig.GetBool("discovery.enabled")
 }
 
 // isLanguageCollectionEnabled returns a boolean indicating if language collection is enabled
@@ -161,7 +159,10 @@ func (c *collector) collectionIntervalConfig() time.Duration {
 // is done. It also gets a reference to the store that started it so it
 // can use Notify, or get access to other entities in the store.
 func (c *collector) Start(ctx context.Context, store workloadmeta.Component) error {
-	if !c.isProcessCollectionEnabled() && !c.isServiceDiscoveryEnabled() {
+	// TODO: process_config.process_collection.use_wlm is temporary and will eventually be removed
+	// we want to gate everything for this new collector by the use_wlm config, but eventually
+	// this collector will be gated separately for process_collector OR service discovery
+	if !c.config.GetBool("process_config.process_collection.use_wlm") && !c.isProcessCollectionEnabled() && !c.isServiceDiscoveryEnabled() {
 		return errors.NewDisabled(componentName, "process collection and service discovery are disabled")
 	}
 
--- a/comp/core/workloadmeta/collectors/internal/process/process_collector.go
+++ b/comp/core/workloadmeta/collectors/internal/process/process_collector.go
@@ -162,8 +162,12 @@ func (c *collector) Start(ctx context.Context, store workloadmeta.Component) err
 	// TODO: process_config.process_collection.use_wlm is temporary and will eventually be removed
 	// we want to gate everything for this new collector by the use_wlm config, but eventually
 	// this collector will be gated separately for process_collector OR service discovery
-	if !c.config.GetBool("process_config.process_collection.use_wlm") && !c.isProcessCollectionEnabled() && !c.isServiceDiscoveryEnabled() {
-		return errors.NewDisabled(componentName, "process collection and service discovery are disabled")
+	if !c.config.GetBool("process_config.process_collection.use_wlm") {
+		return errors.NewDisabled(componentName, "wlm process collection disabled")
+	}
+
+	if !c.isProcessCollectionEnabled() && !c.isServiceDiscoveryEnabled() {
+		return errors.NewDisabled(componentName, "wlm process collection and service discovery are disabled")
 	}
 
 	if c.containerProvider == nil {
--- a/comp/core/workloadmeta/collectors/internal/process/process_collector.go
+++ b/comp/core/workloadmeta/collectors/internal/process/process_collector.go
@@ -136,8 +136,7 @@ func GetFxOptions() fx.Option {
 
 // isProcessCollectionEnabled returns a boolean indicating if the process collector is enabled
 func (c *collector) isProcessCollectionEnabled() bool {
-	return c.config.GetBool("process_config.run_in_core_agent.enabled") &&
-		c.config.GetBool("process_config.process_collection.enabled") &&
+	return c.config.GetBool("process_config.process_collection.enabled") &&
 		// TODO: process_config.process_collection.use_wlm is temporary and will eventually be removed
 		c.config.GetBool("process_config.process_collection.use_wlm")
 }
--- a/comp/core/workloadmeta/collectors/internal/process/process_collector.go
+++ b/comp/core/workloadmeta/collectors/internal/process/process_collector.go
@@ -135,13 +135,11 @@ func GetFxOptions() fx.Option {
 }
 
 // isProcessCollectionEnabled returns a boolean indicating if the process collector is enabled
-func (c *collector) isRunningInCoreAgent() bool {
-	return c.config.GetBool("process_config.run_in_core_agent.enabled")
-}
-
-// isProcessCollectionEnabled returns a boolean indicating if the process collector is enabled
 func (c *collector) isProcessCollectionEnabled() bool {
-	return c.config.GetBool("process_config.process_collection.use_wlm")
+	return c.config.GetBool("process_config.run_in_core_agent.enabled") &&
+		c.config.GetBool("process_config.process_collection.enabled") &&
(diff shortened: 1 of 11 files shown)


comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -132,15 +132,19 @@ func GetFxOptions() fx.Option {
 	return fx.Provide(NewProcessCollectorProvider)
 }
 
+// isProcessCollectionEnabled returns a boolean indicating if the process collector is enabled
```
daniel-taf, line 137: incorrect comment

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -132,15 +132,19 @@ func GetFxOptions() fx.Option {
...
+func (c *collector) isRunningInCoreAgent() bool {
+	return c.config.GetBool("process_config.run_in_core_agent.enabled")
+}
+
 // isProcessCollectionEnabled returns a boolean indicating if the process collector is enabled
 func (c *collector) isProcessCollectionEnabled() bool {
 	return c.config.GetBool("process_config.process_collection.use_wlm")
```
daniel-taf: lets check `process_config.run_in_core_agent.enabled` and `process_config.process_collection.enabled` to determine if the process collection is enabled here.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -158,6 +162,10 @@ func (c *collector) collectionIntervalConfig() time.Duration {
 // is done. It also gets a reference to the store that started it so it
 // can use Notify, or get access to other entities in the store.
 func (c *collector) Start(ctx context.Context, store workloadmeta.Component) error {
+	if !c.isRunningInCoreAgent() {
+		return errors.NewDisabled(componentName, "core process collection not running in core agent")
+	}
```
daniel-taf: I don't think we want this here. It will stop service discovery from running without process collection.

pkg/process/procutil/process_model.go
```diff
@@ -99,6 +103,30 @@ type Stats struct {
...
+	AdditionalGeneratedNames []string
+
+	// TracerMetadata contains APM tracer metadata
+	TracerMetadata []tracermetadata.TracerMetadata
+
+	// DDService is the value from DD_SERVICE environment variable
+	DDService string
+
+	// Ports is the list of ports the service is listening on
+	Ports []uint16
```
daniel-taf: Let's have this at the process level. It's not unique information for a "service" and we have thought about adding it to the standard process collection. It will also match better with the ES mapping we have right now: https://github.com/DataDog/k8s-resources/pull/90721. Though that also opens a question on what we should for `APMInstrumented` and `TracerMetadata`. What level should those be at?

pkg/process/checks/process_linux.go
```diff
@@ -52,17 +56,151 @@ func (p *ProcessCheck) processesByPID(collectStats bool) (map[int32]*procutil.Pr
...
+	languagemodels.Unknown: model.Language_LANGUAGE_UNKNOWN,
+	languagemodels.Go:      model.Language_LANGUAGE_GO,
+	languagemodels.Node:    model.Language_LANGUAGE_NODE,
+	languagemodels.Dotnet:  model.Language_LANGUAGE_DOTNET,
+	languagemodels.Python:  model.Language_LANGUAGE_PYTHON,
+	languagemodels.Java:    model.Language_LANGUAGE_JAVA,
+	languagemodels.Ruby:    model.Language_LANGUAGE_RUBY,
+	languagemodels.PHP:     model.Language_LANGUAGE_PHP,
+	languagemodels.CPP:     model.Language_LANGUAGE_CPP,
+}
```
daniel-taf: This should be at the top of the file. Same with `serviceNameSourceMap` along with comments explaining their usage.

pkg/process/checks/process_linux.go
```diff
@@ -52,17 +56,153 @@ func (p *ProcessCheck) processesByPID(collectStats bool) (map[int32]*procutil.Pr
...
+		Gids:     wlmProc.Gids,
+		Stats:    stats,
+		Ports:    ports,
+		Language: wlmProc.Language,
+		Service:  service,
+	}
+}
+
+// formatPorts converts a list of uin16 ports to a int32 PortInfo
+// TODO: because the service discovery response does not distinguish between tcp and udp currently, we currently send everything as TCP
```
daniel-taf, line 131: For agent discovery folks: Why are they not distinguished? As far as I understand, depending on how we want to present this information on Datadog, this can cause an issue where old agents will show UDP connections as TCP.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
gengnamstyle: done!

comp/core/workloadmeta/collectors/internal/process/process_collector.go
gengnamstyle: yup agreed.

pkg/process/procutil/process_model.go
gengnamstyle: moved to process level. 

Currently the agent payload incorporates the `apm instrumented` and `tracermetadata` into a service discovery struct: https://github.com/DataDog/agent-payload/pull/392/files#diff-69b34851fbf6263111121260c46c57084a1fe6e34e9b750a5a673694f2dfe1b0R573-R576

I do not have a strong opinion on this, but if there are any changes we should keep this consistent.

pkg/process/checks/process_linux.go
gengnamstyle: moved and added a commment

pkg/process/checks/process_linux.go
```diff
@@ -51,18 +88,128 @@ func (p *ProcessCheck) processesByPID(collectStats bool) (map[int32]*procutil.Pr
...
+	serviceNames = append(serviceNames, &model.ServiceName{
+		Name:   service.GeneratedName,
+		Source: source,
+	})
+
+	// add dd service name
+	serviceNames = append(serviceNames, &model.ServiceName{
+		Name:   service.DDService,
+		Source: model.ServiceNameSource_SERVICE_NAME_SOURCE_DD_SERVICE,
+	})
```
daniel-taf: is it possible for this to be empty? Would we want to add it then?

pkg/process/procutil/process_model.go
gengnamstyle: talked offline. apm instrumented and tracer metadata will be kept in the service discovery struct

pkg/process/checks/process_linux.go
gengnamstyle: looking at the service discovery code, it does appear possible that it is empty https://github.com/DataDog/datadog-agent/blob/d47c7ddf2c546497f5de54b542aabdbecf14eae5/pkg/collector/corechecks/servicediscovery/usm/service.go#L309

Changed to not add an empty value

pkg/process/checks/process_linux.go
```diff
@@ -8,10 +8,47 @@
...
+	string(usm.Spring):      model.ServiceNameSource_SERVICE_NAME_SOURCE_SPRING,
+	string(usm.JBoss):       model.ServiceNameSource_SERVICE_NAME_SOURCE_JBOSS,
+	string(usm.Tomcat):      model.ServiceNameSource_SERVICE_NAME_SOURCE_TOMCAT,
+	string(usm.WebLogic):    model.ServiceNameSource_SERVICE_NAME_SOURCE_WEBLOGIC,
+	string(usm.WebSphere):   model.ServiceNameSource_SERVICE_NAME_SOURCE_WEBSPHERE,
+}
+
 // useWLMCollection checks the configuration to use the workloadmeta process collector or not in linux
 // TODO: process_config.process_collection.use_wlm is a temporary configuration for refactoring purposes
 func (p *ProcessCheck) useWLMCollection() bool {
```
daniel-taf, line 55: Realized this function also needs to check `process_config.run_in_core_agent.enabled`. The WLM collector only works on the core agent.

pkg/process/checks/process_linux.go
vitkyrka: It wasn't a requirement to do so for SD but since we plant to be use this for things like autodiscovery it makes sense to be want to know which protocol the ports are on. I can send a follow up or separate PR to change the collection to distinguish them.

pkg/process/checks/process_linux.go
gengnamstyle: Doesn't this then mean that service discovery collection is gated by running on the core agent and therefore service discovery data will not be collected if a user specifies to use the process agent? 

Also this implies that the new process collector should not collect service discovery data if we are not running on the core agent.

daniel-taf: Yes, that is a good point. My concern is that using WLM in the process check might now work on the process agent since it uses RemoteWorkloadmeta. We will need to check if there are any limitations for that. Or, we remove the use of the 'process_config.run_in_core_agent' config since we wanted to do that eventually anyway. I'll need to think on it a bit more, so you can actually removed my suggestion for now.

pkg/process/checks/process_linux.go
daniel-taf: I think that would be great. Would make things simpler on the backend so we don't ever need to worry about if the ports are distinguished by the agent or not.

pkg/process/checks/process_linux.go
gengnamstyle: I added a flowchart to the pr, but I think as you suggested, if we do **not** gate by running on the core agent, then we may be sending duplicate process data (one with service discovery data and one without from the process agent)

gengnamstyle: discussed offline. we will remove the `process_config.run_in_core_agent.enabled` in ` p.useWLMCollection()` in the process check and in `isProcessCollectionEnabled()` in the collector because when `process_config.run_in_core_agent.enabled` is disabled and the process agent collects processes we still want to collect service data in the collector.

pkg/process/checks/process_linux.go
```diff
@@ -20,49 +57,170 @@ func (p *ProcessCheck) useWLMCollection() bool {
...
 		procs := make(map[int32]*procutil.Process, len(wlmProcList))
 		for _, wlmProc := range wlmProcList {
-			procs[wlmProc.Pid] = mapWLMProcToProc(wlmProc, statsForProcess[wlmProc.Pid])
+			stats, exists := statsForProcess[wlmProc.Pid]
+			// we need to check if the stats exist because there can be a lag between when a process is stored into WLM and when we query for its stats
+			// ex. a process is stopped but still exists in WLM, so the stats don't exist, and we shouldn't report it anymore
+			if !exists {
+				continue
+			}
+			procs[wlmProc.Pid] = mapWLMProcToProc(wlmProc, stats)
```
daniel-taf: Let's also check the create time from the stats against the create time from WLM, to ensure its actually the same process and not another with the same pid.

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -134,13 +134,14 @@ func GetFxOptions() fx.Option {
...
+	return c.config.GetBool("process_config.process_collection.enabled") &&
+		// TODO: process_config.process_collection.use_wlm is temporary and will eventually be removed
+		c.config.GetBool("process_config.process_collection.use_wlm")
 }
 
 // isServiceDiscoveryEnabled returns a boolean indicating if service discovery is enabled
 func (c *collector) isServiceDiscoveryEnabled() bool {
-	// TODO: implement the logic to check if service discovery is enabled based on configuration
-	return false
+	return c.systemProbeConfig.GetBool("system_probe_config.enabled") && c.systemProbeConfig.GetBool("discovery.enabled")
```
daniel-taf: I believe on slack, there was mention of add `process_config.process_collection.use_wlm` to this condition.

gengnamstyle: added

pkg/process/checks/process_linux.go
gengnamstyle: added with tests

gabedos (approved):
contp owned files related to enabling/disabling collectors look good 👍

daniel-taf (approved):
Code looks good, just need to make sure CI is green.

vitkyrka (approved):

pkg/process/checks/process_test.go
```diff
@@ -3,6 +3,8 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2016-present Datadog, Inc.
 
+//go:build test
```
pgimalac: FYI you don't need to add the `test` build tag in test files (files ending in `_test.go`)
It's only useful for files containing code only used by tests, typically named something like `test_helpers.go`

gengnamstyle: gotcha! I was hoping this could help reduce the package size since the static quality gate is failing for the `static_quality_gate_iot_agent*` 

Would you have any recommendations for that?

pgimalac: There is an ongoing incident about this failure (#incident-41581) so it might not fully be due to this change
Although it seems you're adding 60KiB to the iot package, most likely due to adding [those packages](https://github.com/DataDog/datadog-agent/pull/39354/s#issuecomment-3133949074) to the build

PR #39356, opened 2025-07-29
[dyninst] Increase dynamic type size to 2KiB

piob-io:
We used .5KiB limit from side-eye, to prevent exhausting event buffer. I expect we will generally collect less data, so we can afford more. Eventually we should work on increasing the event buffer size.


grantseltzer (approved):
:-)

piob-io:
Thanks!

PR #39357, opened 2025-07-29
[NDINT-290] Separate out Analytics parsers

ken-schneider:
### What does this PR do?
Separate out the parsers for Versa Analytics data to a separate file to clean up the code.

The parsers are the same, they've just been moved to a new file.

### Motivation
Improve readability of the Versa integration


jmw51798 (approved):

pkg/collector/corechecks/network-devices/versa/client/parser.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	return rows, nil
+}
+
+// parseLinkUsageMetrics parses the raw AaData response into LinkUsageMetrics structs
+func parseLinkUsageMetrics(data [][]interface{}) ([]LinkUsageMetrics, error) {
+	var rows []LinkUsageMetrics
+	for _, row := range data {
+		m := LinkUsageMetrics{}
+		if len(row) < 13 {
```
zoedt: mentioned on the other PR but should we introduce any comment for reasoning if `<` is ok for a specific request call (i.e. additional things may be added but confidence the first 13 fields should be as expected)?

ken-schneider: No, in fact, these should all be `!=`. Our queries call for specific fields and expect them in a specific order in the response and if there's any delta whatsoever, it's a problem.

I've updated the couple places where we were using `<`. Nice catch!

PR #38446, opened 2025-07-03
[ABLD-77] Bump `dda` version to bring `dmgbuild` for macOS

rdesgroppes:
### What does this PR do?
Since available Python build dependencies rely on the version of `dda`, the present change bumps it to bring the `dmgbuild` tool for macOS.

### Motivation
See https://github.com/DataDog/datadog-agent-dev/releases/tag/v0.19.0, and esp. https://github.com/DataDog/datadog-agent-dev/pull/137.

### Possible Drawbacks / Trade-offs
The PR is kept trivial (quickly revertible) as other changes were brought to `dda` since last bump from March 2025.

PR #39361, opened 2025-07-30
Use the same version of `dda` as builds across CI

ofek:
### Motivation

Consistency, and so that new features may be used since the pinned minimum version in this repo rarely gets updated.

### Additional notes

The majority of this PR is merely removing an unnecessary manual installation step since `dda` is already installed in the build images. There are 3 instances where that is not the case:

- GitHub Actions - this reads `.gitlab-ci.yml` to get the current commit of the build images repository in order to fetch the pinned version
- `.gitlab/common/macos.yml` - macOS does not have its own image, this reads an environment variable to find out the current commit
- `.gitlab/benchmarks/benchmarks.yml` - this uses an image that we do not own and does not ship `dda`, now we read an environment variable just like ^

### DIFF
--- a/.github/actions/install-dda/action.yml
+++ b/.github/actions/install-dda/action.yml
@@ -16,7 +16,8 @@ runs:
   - name: Set version
     id: set-version
     run: |-
-      buildimages_commit="$(python -c "from pathlib import Path;import yaml;print(yaml.safe_load(Path('.gitlab-ci.yml').read_text().replace('!reference', 'reference'))['variables']['CI_IMAGE_LINUX'].split('-')[1])")"
+      buildimages_version="$(yq '.variables.CI_IMAGE_LINUX' .gitlab-ci.yml)"
+      buildimages_commit="${buildimages_version#*-}"
       version="$(curl -s https://raw.githubusercontent.com/DataDog/datadog-agent-buildimages/${buildimages_commit}/dda.env | awk -F= '/^DDA_VERSION=/ {print $2}')"
       echo "version=${version}" >> $GITHUB_OUTPUT
     shell: bash
--- a/.github/actions/install-dda/action.yml
+++ b/.github/actions/install-dda/action.yml
@@ -16,7 +16,7 @@ runs:
   - name: Set version
     id: set-version
     run: |-
-      buildimages_commit="$(python -c "from pathlib import Path;print(Path('.gitlab-ci.yml').read_text().split('CI_IMAGE_LINUX_GLIBC_2_23_ARM64:')[1].split()[0].split('-')[1])")"
+      buildimages_commit="$(python -c "from pathlib import Path;import yaml;print(yaml.safe_load(Path('.gitlab-ci.yml').read_text().replace('!reference', 'reference'))['variables']['CI_IMAGE_LINUX_GLIBC_2_23_ARM64'].split('-')[1])")"
       version="$(curl -s https://raw.githubusercontent.com/DataDog/datadog-agent-buildimages/${buildimages_commit}/dda.env | awk -F= '/^DDA_VERSION=/ {print $2}')"
       echo "version=${version}" >> $GITHUB_OUTPUT
     shell: bash
--- a/.github/actions/install-dda/action.yml
+++ b/.github/actions/install-dda/action.yml
@@ -16,7 +16,7 @@ runs:
   - name: Set version
     id: set-version
     run: |-
-      buildimages_commit="$(python -c "from pathlib import Path;import yaml;print(yaml.safe_load(Path('.gitlab-ci.yml').read_text().replace('!reference', 'reference'))['variables']['CI_IMAGE_LINUX_GLIBC_2_23_ARM64'].split('-')[1])")"
+      buildimages_commit="$(python -c "from pathlib import Path;import yaml;print(yaml.safe_load(Path('.gitlab-ci.yml').read_text().replace('!reference', 'reference'))['variables']['CI_IMAGE_LINUX'].split('-')[1])")"
       version="$(curl -s https://raw.githubusercontent.com/DataDog/datadog-agent-buildimages/${buildimages_commit}/dda.env | awk -F= '/^DDA_VERSION=/ {print $2}')"
       echo "version=${version}" >> $GITHUB_OUTPUT
     shell: bash
--- a/.github/actions/install-dda/action.yml
+++ b/.github/actions/install-dda/action.yml
@@ -15,11 +15,14 @@ runs:
   steps:
   - name: Set version
     id: set-version
-    run: echo "version=$(cat .dda/version)" >> $GITHUB_OUTPUT
+    run: |-
+      buildimages_commit="$(python -c "from pathlib import Path;print(Path('.gitlab-ci.yml').read_text().split('CI_IMAGE_LINUX_GLIBC_2_23_ARM64:')[1].split()[0].split('-')[1])")"
+      version="$(curl -s https://raw.githubusercontent.com/DataDog/datadog-agent-buildimages/${buildimages_commit}/dda.env | awk -F= '/^DDA_VERSION=/ {print $2}')"
+      echo "version=${version}" >> $GITHUB_OUTPUT
     shell: bash
 
   - name: Install dda
-    uses: DataDog/datadog-agent-dev@1c61de50a7dfef056026a4ba7cad75239b5ab324
+    uses: DataDog/datadog-agent-dev@1bc144c32a57327b768fb7161081b6a25a5c485e
     with:
       version: ${{ inputs.version || steps.set-version.outputs.version }}
       features: ${{ inputs.features }}
--- a/.github/workflows/assign_issue.yml
+++ b/.github/workflows/assign_issue.yml
@@ -21,16 +21,15 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
-    - name: Install dependencies
-      # Dependencies are installed at runtime. Otherwise it would create a huge image see https://hub.docker.com/r/pytorch/pytorch/tags
-      run: |
-        pip install --upgrade pip
-        pip install --no-compile --no-cache-dir torch transformers "dda==$(cat .dda/version)"
-        dda self dep sync -f legacy-tasks
+    - name: Install dda
+      uses: ./.github/actions/install-dda
+      with:
+        features: legacy-tasks
     - name: Assign issue
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         SLACK_DATADOG_AGENT_BOT_TOKEN : ${{ secrets.SLACK_DATADOG_AGENT_BOT_TOKEN }}
-        DDA_NO_DYNAMIC_DEPS: "1"
-      run: |
-        dda inv -- -e issue.assign-owner -i ${{ github.event.issue.number }}
+      run: >-
+        dda inv --dep torch --dep transformers
+        --
+        -e issue.assign-owner -i ${{ github.event.issue.number }}
--- a/.gitlab/.pre/ci_configuration.yml
+++ b/.gitlab/.pre/ci_configuration.yml
@@ -21,10 +21,6 @@ test_gitlab_compare_to:
   script:
     - GITLAB_TOKEN=$($CI_PROJECT_DIR/tools/ci/fetch_secret.sh $GITLAB_TOKEN write_api) || exit $?; export GITLAB_TOKEN
     - !reference [.setup_agent_github_app]
-    # Python 3.12 changes default behavior how packages are installed.
-    # In particular, --break-system-packages command line option is
-    # required to use the old behavior or use a virtual env. https://github.com/actions/runner-images/issues/8615
-    - pip install "git+https://github.com/DataDog/datadog-agent-dev.git@v$(cat .dda/version)" --break-system-packages
     - dda self dep sync -f legacy-tasks
     - dda inv -- pipeline.compare-to-itself
 
--- a/.gitlab/benchmarks/benchmarks.yml
+++ b/.gitlab/benchmarks/benchmarks.yml
@@ -10,7 +10,9 @@ benchmark:
   tags: ["team:apm-k8s-tweaked-metal-datadog-agent", "specific:true"]
   script:
     - export ARTIFACTS_DIR="$(pwd)/artifacts" && mkdir -p $ARTIFACTS_DIR
-    - pip install "git+https://github.com/DataDog/datadog-agent-dev.git@v$(cat .dda/version)"
+    - export BUILDIMAGES_COMMIT="$(python -c "from pathlib import Path;print(Path('.gitlab-ci.yml').read_text().split('CI_IMAGE_LINUX_GLIBC_2_23_ARM64:')[1].split()[0].split('-')[1])")"
+    - export DDA_VERSION="$(curl -s https://raw.githubusercontent.com/DataDog/datadog-agent-buildimages/${BUILDIMAGES_COMMIT}/dda.env | awk -F= '/^DDA_VERSION=/ {print $2}')"
+    - pip install "git+https://github.com/DataDog/datadog-agent-dev.git@${DDA_VERSION}"
     - dda self dep sync -f legacy-tasks
     - ./test/benchmarks/apm_scripts/capture-hardware-software-info.sh
     - ./test/benchmarks/apm_scripts/run-benchmarks.sh
--- a/.gitlab/benchmarks/benchmarks.yml
+++ b/.gitlab/benchmarks/benchmarks.yml
@@ -10,7 +10,7 @@ benchmark:
   tags: ["team:apm-k8s-tweaked-metal-datadog-agent", "specific:true"]
   script:
     - export ARTIFACTS_DIR="$(pwd)/artifacts" && mkdir -p $ARTIFACTS_DIR
-    - export BUILDIMAGES_COMMIT="${CI_IMAGE_LINUX_GLIBC_2_23_ARM64#*-}"
+    - export BUILDIMAGES_COMMIT="${CI_IMAGE_LINUX#*-}"
     - export DDA_VERSION="$(curl -s https://raw.githubusercontent.com/DataDog/datadog-agent-buildimages/${BUILDIMAGES_COMMIT}/dda.env | awk -F= '/^DDA_VERSION=/ {print $2}')"
     - pip install "git+https://github.com/DataDog/datadog-agent-dev.git@${DDA_VERSION}"
     - dda self dep sync -f legacy-tasks
--- a/.gitlab/benchmarks/benchmarks.yml
+++ b/.gitlab/benchmarks/benchmarks.yml
@@ -10,7 +10,7 @@ benchmark:
   tags: ["team:apm-k8s-tweaked-metal-datadog-agent", "specific:true"]
   script:
     - export ARTIFACTS_DIR="$(pwd)/artifacts" && mkdir -p $ARTIFACTS_DIR
-    - export BUILDIMAGES_COMMIT="$(python -c "from pathlib import Path;print(Path('.gitlab-ci.yml').read_text().split('CI_IMAGE_LINUX_GLIBC_2_23_ARM64:')[1].split()[0].split('-')[1])")"
+    - export BUILDIMAGES_COMMIT="${CI_IMAGE_LINUX_GLIBC_2_23_ARM64#*-}"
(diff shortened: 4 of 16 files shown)


KevinFairise2:
If we go that way there is a high probability that at some point we do an update of dda in buildimages and that pipelines on datadog-agent starts to fail. Ideally a bump of dda on the agent should be coupled with a PR so that we can see if everything goes well before merging it.
If we want consistency between the version of dda in all the jobs we should find a way to retrieve the version of dda that is used inside the buildimage that we pin on datadog-agent.
For example if `  CI_IMAGE_LINUX_GLIBC_2_17_X64: v72007680-abe4fb3f` is used in the rest of the agent pipeline we could pick the version of dda that is pinned in `abe4fb3f` commit on buildimages?

.gitlab/common/macos.yml
```diff
@@ -78,7 +78,8 @@
   # Python 3.12 changes default behavior how packages are installed.
   # In particular, --break-system-packages command line option is
   # required to use the old behavior or use a virtual env. https://github.com/actions/runner-images/issues/8615
-  - python3 -m pip install "git+https://github.com/DataDog/datadog-agent-dev.git@v$(cat .dda/version)" --break-system-packages
+  - export DDA_VERSION="$(curl -s https://raw.githubusercontent.com/DataDog/datadog-agent-buildimages/main/dda.env | awk -F= '/^DDA_VERSION=/ {print $2}')"
+  - python3 -m pip install "git+https://github.com/DataDog/datadog-agent-dev.git@${DDA_VERSION}" --break-system-packages
```
rdesgroppes: I’d recommend avoiding dynamically selecting the dependency version based on another repo’s `main` branch as a sole reference (moving HEAD).
While it might work in the short term, this would introduce [non-reproducible builds](https://reproducible-builds.org/docs/why/) and make troubleshooting or/and rollback difficult.
We should aim at cherishing explicit and versioned build inputs, even for in-house dependencies, as this also improves traceability, auditability, and our colleagues' experience ([POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)) - in addition to the fundamental principle of reproducibility mentioned above.

rdesgroppes (changes requested):

.gitlab/.pre/ci_configuration.yml
```diff
@@ -24,7 +24,8 @@ test_gitlab_compare_to:
     # Python 3.12 changes default behavior how packages are installed.
     # In particular, --break-system-packages command line option is
     # required to use the old behavior or use a virtual env. https://github.com/actions/runner-images/issues/8615
-    - pip install "git+https://github.com/DataDog/datadog-agent-dev.git@v$(cat .dda/version)" --break-system-packages
+    - export DDA_VERSION="$(curl -s https://raw.githubusercontent.com/DataDog/datadog-agent-buildimages/main/dda.env | awk -F= '/^DDA_VERSION=/ {print $2}')"
+    - pip install "git+https://github.com/DataDog/datadog-agent-dev.git@${DDA_VERSION}" --break-system-packages
```
Ishirui: Why are we even installing `dda` here at runtime, shouldn't it be already available in the buildimage ?
I'm only leaving this comment once to avoid spam, but if my understanding is correct this should apply to most other run-time `dda` installs we see in this PR. It would be much better to remove them altogether I think :+1:

.gitlab/benchmarks/benchmarks.yml
```diff
@@ -10,7 +10,8 @@ benchmark:
   tags: ["team:apm-k8s-tweaked-metal-datadog-agent", "specific:true"]
   script:
     - export ARTIFACTS_DIR="$(pwd)/artifacts" && mkdir -p $ARTIFACTS_DIR
-    - pip install "git+https://github.com/DataDog/datadog-agent-dev.git@v$(cat .dda/version)"
+    - export DDA_VERSION="$(curl -s https://raw.githubusercontent.com/DataDog/datadog-agent-buildimages/main/dda.env | awk -F= '/^DDA_VERSION=/ {print $2}')"
+    - pip install "git+https://github.com/DataDog/datadog-agent-dev.git@${DDA_VERSION}"
```
Ishirui: Here we do need a runtime install 👍

.gitlab/common/macos.yml
Ishirui: Would you be OK with what @KevinFairise2 suggested (referring to a file in the other repo but on a pinned commit) ? The problem we're solving here is twofold:
- Removing the link with the `.dda/version` file here, which should _not_ be used for that purpose
- Trying to not introduce yet another source of truth for what `dda` version is in use.

I agree that installing tools at runtime _at all_ is not ideal, and pulling the version with `curl` like this is also pretty bad, but we don't have the capacity to setup a new buildimage for macos jobs as of yet (which would be the proper solution imo)

.gitlab/packaging/oci.yml
```diff
@@ -19,7 +19,8 @@
     # Python 3.12 changes default behavior how packages are installed.
     # In particular, --break-system-packages command line option is
     # required to use the old behavior or use a virtual env. https://github.com/actions/runner-images/issues/8615
-    - python3 -m pip install "git+https://github.com/DataDog/datadog-agent-dev.git@v$(cat .dda/version)" --break-system-packages
+    - export DDA_VERSION="$(curl -s https://raw.githubusercontent.com/DataDog/datadog-agent-buildimages/main/dda.env | awk -F= '/^DDA_VERSION=/ {print $2}')"
+    - python3 -m pip install "git+https://github.com/DataDog/datadog-agent-dev.git@${DDA_VERSION}" --break-system-packages
```
Ishirui, line 23: Ditto

.gitlab/common/macos.yml
rdesgroppes: Perhaps a silly question: is `dda` used anywhere other than in `datadog-agent`? If not, maybe we should consider maintaining it inside the _de facto_ main repo. Worth a read, I think: [Advantages of monorepos](https://danluu.com/monorepo): `Simplified dependencies`, `Tooling`, `Cross-project changes`... isn't it right our case?

@Ishirui, to answer your question (`referring to a file in the other repo but on a pinned commit`), yes it's definitely ~better~ safer than referring to a moving HEAD.

Ishirui: Maybe - that's a much, much larger discussion though, which we have already brought up a couple of times.

In any case I don't think merging `dda` with `datadog-agent` will solve anything as long as we keep relying on other repos like `datadog-agent-buildimages` and `test-infra-definitions` - we'd have to go all the way to see the benefits I think

And even if we do fuse some repos in the future, I would personally prefer for `dda` to remain its own independent tool. Yes it is pretty intrinsically linked to `datadog-agent`, but we are trying to make it as independent as possible (and hopefully we'll get there 😓). The vision for `dda` is to make it more akin to an external tool like, idk, `git` or `uv` - all the repo-specific logic will not be implemented in `dda` directly but in each individual repo through the "local commands" feature (you can extend the capabilities of `dda` for one repo by putting Python modules in the `.dda/extend` folder)

rdesgroppes: > but we are trying to make it as independent as possible
> The vision for `dda` is to make it more akin to an external tool like, idk, `git` or `uv`

Well, I guess I'll have to digest prior background to understand why.
Alright, I agree that this is a much broader debate and that it's unlikely we'll reach a consensus in this thread.

Nevertheless, I'd welcome any (sustainable) progress as my PR #38446 is stuck for about a month despite having done my best to follow the current `dda` paradigm that provides build-time dependencies: DataDog/datadog-agent-dev#137, DataDog/datadog-agent-buildimages#888, DataDog/omnibus-ruby#241, DataDog/omnibus-ruby#242...
I thought it was just a matter of announcing the version bump to the contributors, isn't it any longer?

Ishirui: What it boils down to, and also I think the reason why your PR has been stuck for so long (sorry again ! 😖) is that no one wants to force an update of `dda` on all developers' _local_ environments without a good reason. And "getting the CI working for macos" is, imo, not a good reason to require an update of people's _local_ environments - the two should not even be linked.

There is no single "`dda` version" in use:
- The one defined in `.dda/version` is responsible for enforcing the minimum version each developer needs to have installed _locally_. To be clear: if the version of `dda` you have locally is inferior to the one defined there, _all the `dda` commands you run locally will immediately abort until you update_. This is quite disruptive, and why we are wary of bumping this file and need to make an announcement if we do. 
> Until this PR, this was also used as the source-of-truth for some runtime installs of `dda` in CI, which is *wrong* and needs to be fixed. Once again - the two should never have been linked in the first place.
- The buildimages are built with `dda` preinstalled, the version being pinned in [`datadog-agent-buildimages/dda.env`](https://github.com/DataDog/datadog-agent-buildimages/blob/main/dda.env) - this is the preffered source of truth for what version of `dda` should be installed in CI environments, and can be bumped without too much hassle, and most importantly without disrupting anyone's local env.

So yes, we _could_ bump the version in `.dda/version`, and make an announcement, and probably peeve some devs for breaking their local environment. But we really shouldn't have to do that to fix a CI issue, and this is what that PR aims to fix.

Once this is merged your PR [#38446](https://github.com/DataDog/datadog-agent/pull/38446) will not be necessary anymore, as the macos build will use the same version of `dda` as the buildimages, i.e. the one defined n [`datadog-agent-buildimages/dda.env`](https://github.com/DataDog/datadog-agent-buildimages/blob/main/dda.env), which should be recent enough.

rdesgroppes: I understand you're aiming at decoupling the `dda` version used by developers from the one used in CI.
Whereas I favor strict tooling alignment - as MF put it, "[Put everything in a version controlled mainline](https://martinfowler.com/articles/continuousIntegration.html#PutEverythingInAVersionControlledMainline)" - especially to [avoid surprises](https://books.google.de/books?id=I-NOcVMGWSUC&pg=PA85#v=onepage&q&f=false) when build dependencies differ across branches (`main`, `7.69.x`, etc.), I acknowledge your editorial choice and will go along with it.

spencergilbert (changes requested):
This should, at the very least have retries for every install point.

ofek:
Sorry for the noise, taking this PR out of draft was unintentional as I was just trying to make sure CI passed since it's been using an old version for a long time. I was worried that something would break so fetching the version dynamically was the fastest way to expose any issues. I'm going to put this back into draft and implement the final approach.

.github/workflows/assign_issue.yml
```diff
@@ -21,16 +21,15 @@ jobs:
...
+        features: legacy-tasks
     - name: Assign issue
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         SLACK_DATADOG_AGENT_BOT_TOKEN : ${{ secrets.SLACK_DATADOG_AGENT_BOT_TOKEN }}
-        DDA_NO_DYNAMIC_DEPS: "1"
-      run: |
-        dda inv -- -e issue.assign-owner -i ${{ github.event.issue.number }}
+      run: >-
+        dda inv --dep torch --dep transformers
```
aiuto, line 33: I think a better solution for situations like this is to go down an entirely different path.  Instead of improving the ability to assign issues to `dda`, we should move that functionality to an entirely different tool. That tool can be built, tested, and distributed without any changes to dda. CI workflows should point to it directly.  The command line `dda` tool may invoke it as a very slim wrapper, but the only time a developer might see that is if they call it from the command line.

Besides allowing greater decoupling of unrelated code, that approach allows us to distribute features in an appropriate way per feature.  For example, this one, assign-issues, is probably one that is dependent on the features generally available on github.  My hunch is that we want this in the CI build images that we use. At the other end of the spectrum, a  hypothetical tool that generates new go.mod files based on MODULE.bazel should be checked in with the source tree, so that if we go back to an old release branch we can still build. In general, for any tool, if you would need to go back to an older version to build in an old branch, then it should be part of the source repo rather than in `dda`. If you have to update it in response to changes in corporate policy, or general gitlab infrastructure, then distribute it as part of the CI build images.

.gitlab/common/macos.yml
ofek: The version of `dda` uses the same version installed within the build images now in all cases.

.gitlab/.pre/ci_configuration.yml
ofek: Indeed, removed!

.github/actions/install-dda/action.yml
```diff
@@ -15,11 +15,13 @@ runs:
   steps:
   - name: Set version
     id: set-version
-    run: echo "version=$(cat .dda/version)" >> $GITHUB_OUTPUT
+    run: |-
+      version="$(curl -s https://raw.githubusercontent.com/DataDog/datadog-agent-buildimages/main/dda.env | awk -F= '/^DDA_VERSION=/ {print $2}')"
+      echo "version=${version}" >> $GITHUB_OUTPUT
```
Ishirui: Is there a way we could use buildimages in our GitHub Actions as well, to avoid needing to do this ?

Ishirui (approved):
I think one occurence of `pip install` using `.dda/version` was missed in `.gitlab/benchmarks`. That job does not use any of our buildimages, so we can't outright remove it, but we should still make it depend on the same version as the buildimages.

.github/actions/install-dda/action.yml
ofek: Done!

ajgajg1134 (approved):

clarkb7 (approved):

.github/actions/install-dda/action.yml
```diff
@@ -15,11 +15,14 @@ runs:
   steps:
   - name: Set version
     id: set-version
-    run: echo "version=$(cat .dda/version)" >> $GITHUB_OUTPUT
+    run: |-
+      buildimages_commit="$(python -c "from pathlib import Path;print(Path('.gitlab-ci.yml').read_text().split('CI_IMAGE_LINUX_GLIBC_2_23_ARM64:')[1].split()[0].split('-')[1])")"
```
rdesgroppes: Although it certainly works, I'm afraid the approach is fragile because it parses `.gitlab-ci.yml` as plain text instead of valid YAML, which makes it sensitive to formatting, indentation, or comments.
For **example**:
- what happens if there's no occurrence of the `CI_IMAGE_LINUX_GLIBC_2_23_ARM64:` string in the file?
- or if it appears several times:
    - new key like `OLD_CI_IMAGE_LINUX_GLIBC_2_23_ARM64:`,
    - comment like `# CI_IMAGE_LINUX_GLIBC_2_23_ARM64: warning, also used to query dda version`?

Also, I see there are 13 entries like `CI_IMAGE_LINUX_GLIBC_2_23_ARM64`, especially `CI_IMAGE_WIN_LTSC2022_X64` which is maintained through a distinct `renovate` rule (example bump #38180), which might lead to out-of-sync versions of `dda` being used - unless I'm mistaken.

.github/actions/install-dda/action.yml
```diff
@@ -15,11 +15,14 @@ runs:
...
+      version="$(curl -s https://raw.githubusercontent.com/DataDog/datadog-agent-buildimages/${buildimages_commit}/dda.env | awk -F= '/^DDA_VERSION=/ {print $2}')"
+      echo "version=${version}" >> $GITHUB_OUTPUT
     shell: bash
 
   - name: Install dda
-    uses: DataDog/datadog-agent-dev@1c61de50a7dfef056026a4ba7cad75239b5ab324
+    uses: DataDog/datadog-agent-dev@1bc144c32a57327b768fb7161081b6a25a5c485e
```
rdesgroppes, line 26: I guess this is not possible:
```suggestion
    uses: DataDog/datadog-agent-dev@${version}
```
... right?

.gitlab/benchmarks/benchmarks.yml
```diff
@@ -10,7 +10,9 @@ benchmark:
   tags: ["team:apm-k8s-tweaked-metal-datadog-agent", "specific:true"]
   script:
     - export ARTIFACTS_DIR="$(pwd)/artifacts" && mkdir -p $ARTIFACTS_DIR
-    - pip install "git+https://github.com/DataDog/datadog-agent-dev.git@v$(cat .dda/version)"
+    - export BUILDIMAGES_COMMIT="$(python -c "from pathlib import Path;print(Path('.gitlab-ci.yml').read_text().split('CI_IMAGE_LINUX_GLIBC_2_23_ARM64:')[1].split()[0].split('-')[1])")"
```
rdesgroppes: Same as [above](https://github.com/DataDog/datadog-agent/pull/39361/files#r2256254149).

Plus, since we're in a GitLab definition file, isn't the value of the `CI_IMAGE_LINUX_GLIBC_2_23_ARM64` environment variable available here?

.github/actions/install-dda/action.yml
Ishirui: We discussed this offline, I think Ofek wanted to do proper YAML parsing but the `!reference` tags included everywhere are apparently custom to Gitlab - `pyyaml` errors out when parsing them. So the easiest solution was to do a raw text search - not great, I agree. If there's several occurences of this string, it'll be based on the first one.

Re: the 13 entries - I wasn't here when they were put in place, I guess the idea was that they would be updated separately more often. I have never seen that happen though, all our tooling just does them all at the same time (they are considered as one single `linux-images` dependency by Renovate). The exception, as you point out, is windows, which is treated separately as it is a radically different image.
> We're also actively working on merging all of these images, so hopefully we won't have all those repeats in the near future

.gitlab/benchmarks/benchmarks.yml
Ishirui: +1 on this - the only reason I can think of that it might not be available is because this job is running in a separate "triggered pipeline", but I don't think that's the case ?

.github/actions/install-dda/action.yml
rdesgroppes: > [...] proper YAML parsing [...]
```bash
$ yq '.variables.CI_IMAGE_LINUX_GLIBC_2_23_ARM64' .gitlab-ci.yml
v72499966-f7e12097
```
:new: **Update** (with some kind of validation for the image tag)
```bash
$ hash=$(yq -r '.variables.CI_IMAGE_LINUX_GLIBC_2_23_ARM64 | select(test("^v[0-9]+-[0-9a-f]{7,}$")) | split("-")[1]' .gitlab-ci.yml)
$ echo $hash
f7e12097
```

Ishirui: Well there you go 😅 I guess we have to make sure `yq` is installed in the github runners though

ofek: I updated to the original (equally hacky) approach I had in mind which was to modify the text to remove the custom tags which were preventing loading.

There are some utilities within the legacy invoke scripts that have not yet been migrated to dda but this job needs to be fast and that would require installing dependencies on the fly, so this is a fine compromise for now.

.github/actions/install-dda/action.yml
ofek: That is undesirable because this refers to commits on the branch where the Action lives https://github.com/DataDog/datadog-agent-dev/tree/install

.gitlab/benchmarks/benchmarks.yml
ofek: I'm trying your approach but that's not why I did that. We don't control the image so I don't know the version of Bash that they use and the syntax that I was using in the macOS file requires a recent version, so it was easiest to copy the file reading approach. Let's see if CI passes.

.github/actions/install-dda/action.yml
rdesgroppes: Oh, sorry, I didn't realize the action had its own release cycle... resolving!

rdesgroppes (approved):
I’m approving this PR because it moves us towards more reproducible builds, even if I’m not thrilled by the added indirections. (getting a version now requires parsing a container image tag comprising a `git` hash, pulling a file from another repo at that hash, and again parsing the latter to extract the actual version eventually used for dependency resolution)

As said, the gain outweighs the friction.

.github/actions/install-dda/action.yml
alopezz: Ubuntu runners do seem to have `yq` installed according to https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#tools, in case we still want to try that approach.

.gitlab/benchmarks/benchmarks.yml
rdesgroppes: No worry, it's POSIX: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02

macOS ships with Bash 3.2 by default due to licensing. Fortunately, this syntax is supported in Bash since the very early 2.x versions, so it’s fully compatible.

spencergilbert (approved):
Echoing Regis - we need to keep an eye on flakey failures, and while I don't think it's optimal it is better than where we are today.

PR #39363, opened 2025-07-30
[CWS] Fix Flaky IOUring Connect Test 

theop-dd:
### What does this PR do?

This PR attempts to fix the flakiness of the `IOUring Connect` test.

- Adds a sleep to ensure that the `accept` call is properly set up before we begin tracking events. This might cause some issues.
- Changes the ports used to prevent overlaps. Previously, the test could fail with the error `listen tcp :4242: bind: address already in use`.  
  While we could have used `setsockopt` with `REUSEPORT` and/or `REUSEADDR`, changing the ports gives us more control over the test environment.

### Motivation

Reduce test flakiness and improve reliability.

### DIFF
--- a/pkg/security/tests/connect_test.go
+++ b/pkg/security/tests/connect_test.go
@@ -16,6 +16,7 @@ import (
 	"net/http"
 	"syscall"
 	"testing"
+	"time"
 
 	"github.com/iceber/iouring-go"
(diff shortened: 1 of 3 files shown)


YoannGh (approved):
Thanks for the PR, could you please make sure you remove the print statements before merging the PR?

spikat (approved):

PR #39364, opened 2025-07-30
Introduce dynamic index to run relevant tests

KevinFairise2:
### What does this PR do?

Introduce dynamic index to execute tests as described in https://datadoghq.atlassian.net/wiki/x/q4h2PgE
Create several components that allow to compute a dynamic index:
- **DynamicIndex**: Wrapper around a dict[str, dict[str, set[str]] that represent an index giving a list of test to execute given some changes, with some useful function to dump the index, merge it or load it from a JSON.
- **DynTestBackend**: Represent a backend to store indexes with basic crud operations , first implementation is using a S3 bucket with json files stored.
- **DynTestIndexer**: Used to compute an index, first one implement is a coverage-based indexer that use Go code coverage to compute the index
- **DynTestExecutor**: Used to retrieve the correct index and that given a list of changes return the tests that must be executed.
- **DynTestEvaluator**: Not strictly needed to run the tests but useful to evaluate how efficient the system is and to make sure it would not hide issues.

More details can be found in the README of tasks/libs/dynamic_test and in the docstrings.

Merging that PR will put in place a system to evaluate how efficient it is to run e2e tests based on the coverage computed in a nightly pipeline. A corresponding notebook was created to see the amount of test executed with that new method vs the old way of doing it: https://app.datadoghq.com/notebook/12935247/dynamic-test-execution-evaluation?range=172800000&start=1755600314643&live=true
It will also help to detect errors in the system and too optimistic test skips.

Next steps is to let the system work for some days to assess how efficient it is, to fix the eventual errors that are found. Once this is working the executor should be used in the e2e tests directly to skip the tests that are not needed.
In the meantime we should add coverage support for more tests, at that moment only the host-based tests are able to compute their coverage.

Note: The evaluation is only made on tests that we have info about, which means that all the tests that we are not collecting the coverage yet are not included in the statistics.

The evaluate job compare test executed with that output: https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1102804231

### Motivation

The ultimate goal is make our CI more efficient by providing a way to run only relevant tests on PR. It should reduce the cost of the CI and the flakiness, should also reduce the number of failing tests that end up in main because they were not executed on the PR

### Additional Notes

Some `pip install boto3` are still done at runtime, they will be removed as soon as test-infra-definitions is bumped to the latest version


.gitlab-ci.yml
```diff
@@ -78,12 +79,13 @@ stages:
   - install_script_testing
   - e2e_pre_test
   - e2e_init
-  - e2e
+  - e2e
```
CelianR:
```suggestion
  - e2e
```

tasks/libs/common/datadog_api.py
```diff
@@ -120,3 +120,23 @@ def get_ci_pipeline_events(query, days):
...
+    from datadog_api_client import ApiClient, Configuration
+    from datadog_api_client.v2.api.ci_visibility_tests_api import CIVisibilityTestsApi
+
+    configuration = Configuration()
+    with ApiClient(configuration) as api_client:
+        api = CIVisibilityTestsApi(api_client)
+        # We filter jobs of a single pipeline by its id and job name
+        response = api.list_ci_app_test_events(
+            filter_query=query,
+            page_limit=100,
```
CelianR: Should we increase this to be more efficient? It seems low for the number of jobs we can have but I'm not sure about the max limit

tasks/libs/dynamic_test/README.md
```diff
@@ -0,0 +1,153 @@
+# Dynamic Test Index Utilities
+
```
CelianR, line 2:
```suggestion

**Please see the [documentation](https://datadoghq.atlassian.net/wiki/x/q4h2PgE)**.

```

tasks/libs/dynamic_test/evaluator.py
```diff
@@ -0,0 +1,410 @@
...
+                    value=result.not_executed_failing_count(),
+                    tags=["pipeline_id:" + self.pipeline_id, "index_kind:" + self.kind.value, "job:" + result.job_name],
+                )
+            )
+        send_metrics(series)
+
+    def _evaluate_job(
+        self, job: str, current_job_tests: list[ExecutedTest], predicted_tests: set[str]
+    ) -> EvaluationResult:
+        # Only consider indexed tests, other tests are test the system is currently not able to determine whether they should be executed or not.
```
CelianR:
```suggestion
        # Only consider indexed tests, the system is currently not able to determine whether other tests should be executed or not.
```
Is this right?

.gitlab/dynamic_test/consolidate.yml
```diff
@@ -0,0 +1,10 @@
+index-consolidation:
+  stage: dynamic_test
+  image: registry.ddbuild.io/ci/datadog-agent-buildimages/linux$CI_IMAGE_LINUX_SUFFIX:$CI_IMAGE_LINUX
+  tags: ["arch:amd64"]
```
CelianR: NIT
```suggestion
  tags: ["arch:arm64"]
```

CelianR (approved):

tasks/libs/common/s3.py
```diff
@@ -0,0 +1,215 @@
...
+    s3_prefix = "/".join(s3_path.split("/")[1:])
+
+    # Ensure prefix ends with '/' for proper folder listing
+    if s3_prefix and not s3_prefix.endswith('/'):
+        s3_prefix += '/'
+
+    try:
+        s3_client = boto3.client('s3')
+        # List all objects in the S3 prefix
+        response = s3_client.list_objects_v2(Bucket=s3_bucket_name, Prefix=s3_prefix)
```
ofek: According to the [docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/list_objects_v2.html), this method is limited to 1000 results. What do you think about using pagination? I understand we wouldn't hit this bug right away but I think it's best to fix it now rather than later.

tasks/dyntest.py
```diff
@@ -0,0 +1,43 @@
...
+@task
+def evaluate_index(ctx: Context, bucket_uri: str, commit_sha: str, pipeline_id: str):
+    uploader = S3Backend(bucket_uri)
+    executor = DynTestExecutor(ctx, uploader, IndexKind.PACKAGE, commit_sha)
+    evaluator = DatadogDynTestEvaluator(ctx, IndexKind.PACKAGE, executor, pipeline_id)
+    if not evaluator.initialize():
+        print("Failed to initialize index")
+        return
+    changes = get_modified_files(ctx)
+    results = evaluator.evaluate([os.path.basename(change) for change in changes])
```
ofek: I think this would be a bug, no? From what I can tell here we are passing file names but the evaluator compares against Go package paths from the index. Did you mean to use [`os.path.dirname`](https://docs.python.org/3/library/os.path.html#os.path.dirname)?

tasks/libs/dynamic_test/executor.py
```diff
@@ -0,0 +1,131 @@
...
+            RuntimeError: If no ancestor commit with an index is found
+        """
+        closest_commit = ""
+        for commit in self.backend.list_indexed_keys(kind):
+            is_ancestor = self.ctx.run(f"git merge-base --is-ancestor {commit} {commit_sha}", hide=True, warn=True)
+            if is_ancestor.ok:
+                closest_commit = commit
+                break
+        if closest_commit == "":
+            raise RuntimeError(f"No ancestor commit found for {commit_sha}")
```
ofek, line 103: I think such tooling should never fail CI. I'd much prefer this case runs every test with potentially some telemetry indicating what happened. What do you think?

tasks/libs/dynamic_test/backend.py
```diff
@@ -0,0 +1,185 @@
...
+        download_file_from_s3(
+            s3_path=f"{self.s3_base_path}/{kind.value}/{key}/index.json", local_path=f"{tmp_dir}/index.json"
+        )
+        with open(os.path.join(tmp_dir, "index.json")) as f:
+            return DynamicTestIndex.from_dict(json.load(f))
+
+    def list_indexed_keys(self, kind: IndexKind) -> list[str]:
+        keys = list_sorted_keys_in_s3(f"{self.s3_base_path}/{kind.value}", "index.json")
+
+        return [key.split("/")[0] for key in keys if len(key.split("/")) == 2]
```
ofek: I was perplexed for a moment, perhaps let's make this explicit?

```suggestion
    commit_keys = []
    for key in keys:
        path_parts = key.split("/")

        # A consolidated index path is: <commit_sha>/index.json
        # A partial per-job index path is: <commit_sha>/<job_id>/index.json
        # We must only use the consolidated indices to have a complete dataset.
        if len(path_parts) == 2:
            commit_keys.append(path_parts[0])

    return commit_keys
```

ofek (changes requested):
Great work so far!

tasks/libs/dynamic_test/executor.py
```diff
@@ -0,0 +1,131 @@
...
+
+        Returns:
+            DynamicTestIndex: The loaded index from the closest ancestor commit
+
+        Raises:
+            RuntimeError: If no ancestor commit with an index is found
+        """
+        closest_commit = ""
+        for commit in self.backend.list_indexed_keys(kind):
+            is_ancestor = self.ctx.run(f"git merge-base --is-ancestor {commit} {commit_sha}", hide=True, warn=True)
```
ofek, line 98: This could become very slow eventually, we should track an improvement here.

tasks/libs/dynamic_test/backend.py
```diff
@@ -0,0 +1,185 @@
...
+        # Get system temp folder for downloading index files
+        tmp_dir = tempfile.mkdtemp(prefix="dynidx_consolidate_")
+
+        print(f"Downloading index files from {self.s3_base_path}/{kind.value}/{key}")
+        download_folder_from_s3(s3_path=f"{self.s3_base_path}/{kind.value}/{key}", local_path=tmp_dir)
+
+        consolidated_index = DynamicTestIndex()
+        for folder in os.listdir(tmp_dir):
+            if not os.path.isdir(os.path.join(tmp_dir, folder)):
+                continue
```
ofek, line 166: I think here if a job happens to fail or otherwise does not submit its data then the consolidated index would be incomplete. What do you think about adding a check right before this part that the number of directories matches the expected number of jobs?

ofek (changes requested):
I noticed a few more things.

tasks/dyntest.py
KevinFairise2: This is definitely, I had it fixed in another commit somewhere but forgot it in the last version! Thanks for catching it.

tasks/libs/dynamic_test/executor.py
KevinFairise2: I think an exception is raised here when not index can be found. However I would not expect it to fail the CI. If the executor cannot be properly initialized the code calling it should just not try to use the executor to determine which test to run and just keep doing what we use to do. In most of the case just running them all

tasks/libs/dynamic_test/executor.py
KevinFairise2: Indeed it could be slow if the number of indexed commit becomes huge. The goal is to keep that indexed commit number not so hight by having lifecycle rule that drop the index after some time (for 1 month for example, which would give at most 30 elements)

tasks/libs/dynamic_test/backend.py
KevinFairise2: I added a check and a warning if a partial index is missing while expected.
However having job not uploading their data is something that is expected, at least at the beginning.
Some of the jobs are not computing their code coverage yet so we cannot populate the index with their data. However we can still use that system for the jobs that are creating an index.
For all the jobs that actually created their partial index, the executor will be able to tell which tests are impacted and which one can be skipped. If we do not have data for a specific job we just do not skip anything.
During a first phase the dynamic test system would be used as a test skipper, so we skip test when we know we can and do nothing when we do not have info

tasks/libs/dynamic_test/executor.py
CelianR: If it <= 30 elements I think it is not worth to optimize this and keep it clear

tasks/libs/dynamic_test/backend.py
ofek: After the system is in place for some time, what you think we should do instead of printing out that warning?

ofek (approved):

KevinFairise2:
I think that missing information for a given test/job can always happen. My take on that is that we should try our best to have a complete index ( by adding retries on the pipeline computing the index for example ) but in the case where some parts are missing it should fail safe, i.e if we do not have enought information to determine wether a test should be execued or not we should execute it

PR #39366, opened 2025-07-30
[USM] Support http2 PRIORITY flag

amitslavin:
### What does this PR do?

* This PR adds support for HTTP/2 priority flag.
* Change `HTTP2_MAX_FRAMES_FOR_HEADERS_PARSER_PER_TAIL_CALL` from 16 to 15.
* Change `HTTP2_MAX_TAIL_CALLS_FOR_HEADERS_PARSER` from 15 to 16. 

### Motivation

When priority flag is set, the frame includes 5 additional bytes: 4 bytes for the stream dependency field and 1 byte for the weight. Since we weren’t accounting for these extra bytes in our parser, we were misinterpreting the frame structure, which caused the capture logic to fail.

> The HEADERS frame defines the following flags:
> PRIORITY (0x20): When set, the PRIORITY flag indicates that the Exclusive, Stream Dependency, and Weight fields are present.

link to the HTTP/2 RFC that contains documentation about the issue: https://datatracker.ietf.org/doc/html/rfc7540#section-6.2

I reached the instruction count limit, which explains the changes to the constants we use to iterate over the frames. I increased one and decreased another to maintain the same total number of iterations as before.

I added a UT that includes this flag. It failed before the change and passed after applying my fix.
- [x] Load test.
- [x] Staging cluster deployment.

### Additional Notes
gRPC plain:
Screenshot 2025-07-31 at 11 09 04

gRPC TLS:
Screenshot 2025-07-31 at 11 06 25

### DIFF
--- a/pkg/network/ebpf/c/protocols/http2/decoding-defs.h
+++ b/pkg/network/ebpf/c/protocols/http2/decoding-defs.h
@@ -15,7 +15,7 @@
 // Represents the maximum number of frames we'll process in a single tail call in `handle_headers_frames` program.
 #define HTTP2_MAX_FRAMES_FOR_HEADERS_PARSER_PER_TAIL_CALL 15
 // Represents the maximum number of tail calls to process headers frames.
-// Currently we have up to 240 frames in a packet, thus 15 (15*16 = 240) tail calls is enough.
+// Currently we have up to 240 frames in a packet, thus 16 (15*16 = 240) tail calls is enough.
 #define HTTP2_MAX_TAIL_CALLS_FOR_HEADERS_PARSER 16
 #define HTTP2_MAX_FRAMES_FOR_HEADERS_PARSER (HTTP2_MAX_FRAMES_FOR_HEADERS_PARSER_PER_TAIL_CALL * HTTP2_MAX_TAIL_CALLS_FOR_HEADERS_PARSER)
 // Maximum number of frames to be processed in a single tail call.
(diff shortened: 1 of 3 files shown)


pkg/network/ebpf/c/protocols/http2/decoding-defs.h
```diff
@@ -13,10 +13,10 @@
 #define HTTP2_MAX_FRAMES_FOR_EOS_PARSER (HTTP2_MAX_FRAMES_FOR_EOS_PARSER_PER_TAIL_CALL * HTTP2_MAX_TAIL_CALLS_FOR_EOS_PARSER)
 
 // Represents the maximum number of frames we'll process in a single tail call in `handle_headers_frames` program.
-#define HTTP2_MAX_FRAMES_FOR_HEADERS_PARSER_PER_TAIL_CALL 16
+#define HTTP2_MAX_FRAMES_FOR_HEADERS_PARSER_PER_TAIL_CALL 15
 // Represents the maximum number of tail calls to process headers frames.
 // Currently we have up to 240 frames in a packet, thus 15 (15*16 = 240) tail calls is enough.
```
guyarb: please update the documentation here as well

pkg/network/ebpf/c/protocols/http2/decoding.h
```diff
@@ -931,6 +931,16 @@ static __always_inline void headers_parser(pktbuf_t pkt, void *map_key, conn_tup
         current_stream->tags = tags;
         pktbuf_set_offset(pkt, current_frame.offset);
 
+        // If PRIORITY flag (0x20) set, skip 5-byte priority fields.
+        // See: https://datatracker.ietf.org/doc/html/rfc7540#section-6.2
+        if (current_frame.frame.flags & HTTP2_PRIORITY_FLAG) {
+            pktbuf_advance(pkt, 5);
+            if (current_frame.frame.length > 5) {
+                current_frame.frame.length -= 5;
```
guyarb: I would add a `#define` with `PRIORITY_BUFFER_LEN 5` and use it

guyarb (approved):

PR #39369, opened 2025-07-30
[SYNTH-21092] Use the new data-traceroute lib and use the latest udp code.

Fabrice-Matrat:
### What does this PR do?

This PR uses the new datadog-traceroute lib now that the code has been moved to datadog-traceroute


pkg/ebpf/bytecode/runtime/dynamicinstrumentation.go
paulcacheux, line 1: This whole file should be removed, it's here because the gitignore entry was removed in https://github.com/DataDog/datadog-agent/pull/39208

Fabrice-Matrat: Done

pimlu (approved):
This LGTM, I wonder if we should wait to merge until after the 7.70 release cutoff on Aug 7 next week, to give us maximum time to test via the nightly deploys, deploys to staging, etc? What do you think @AlexandreYang

pkg/networkpath/traceroute/runner/runner.go
```diff
@@ -188,7 +202,26 @@ func (r *Runner) runICMP(cfg config.Config, hname string, target net.IP, maxTTL
...
-	log.Tracef("TCP Results: %+v", pathResult)
+	log.Tracef("ICMP Results: %+v", pathResult)
+
+	return pathResult, nil
+}
+
+func (r *Runner) runUDP(cfg config.Config, hname string, target net.IP, maxTTL uint8, timeout time.Duration) (payload.NetworkPath, error) {
+	destPort := cfg.DestPort
+	if destPort == 0 {
+		destPort = 33434 // TODO: is this the default we want?
```
pimlu, line 213: The previous code used `33434 + uint16(rand.Intn(30))`, which is based off the idea that traditional traceroutes would use `33434 + ttl` as the port for their probe

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/DataDog/datadog-agent
 
-go 1.24.0
+go 1.24.5
```
AlexandreYang: Is that intended? we should probably avoid bumping go version if not necessary.

pkg/networkpath/traceroute/runner/runner.go
AlexandreYang: I think it should be fine to keep it static for now since we already vary the source port.

Except if you see downside of keeping it static @pimlu @Fabrice-Matrat .

pimlu: That seems fine to me

go.mod
Fabrice-Matrat: downgraded

AlexandreYang (approved):
LGTM, If static path and dynamic path are working fine on staging with this agent build.

(discuss offline that this build should be tested on staging before being merged)

jack0x2 (approved):

gjulianm (approved):

PR #39370, opened 2025-07-30
[CWS] Fix kill queue deadlock

YoannGh:
### What does this PR do?

Fix a deadlock when queuing kill actions by ensuring locks are always taken in the same order (ruleDisarmers map lock -> disarmer lock).

### DIFF
--- a/pkg/security/probe/process_killer.go
+++ b/pkg/security/probe/process_killer.go
@@ -145,7 +145,9 @@ func (p *ProcessKiller) KillQueuedPidsAndSetNextAlarm() {
 	now := time.Now()
 	var nextAlarm *time.Time
 	for ruleID, ruleDisarmer := range p.ruleDisarmers {
+		ruleDisarmer.m.Lock()
 		if ruleDisarmer.disarmed || len(ruleDisarmer.killQueue) == 0 {
+			ruleDisarmer.m.Unlock()
 			continue
 		}
 		if now.After(ruleDisarmer.killQueueAlarm) {
@@ -156,6 +158,7 @@ func (p *ProcessKiller) KillQueuedPidsAndSetNextAlarm() {
 				nextAlarm = &ruleDisarmer.killQueueAlarm
 			}
 		}
+		ruleDisarmer.m.Unlock()
 	}
 	if nextAlarm != nil {
 		p.setKillQueueAlarm(nextAlarm)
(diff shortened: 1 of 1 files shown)


pkg/security/probe/process_killer.go
```diff
@@ -257,7 +260,7 @@ func (p *ProcessKiller) updateKillQueueAlarmOnDisarm(disarmer *ruleDisarmer) {
 	// the correct value. Disable the alarm otherwise.
 	// NB: we should not kill anything, except if there is a race between a rule
 	//     disarmament and a kill queue alarm
-	p.KillQueuedPidsAndSetNextAlarm()
+	p.killQueuedPidsAndSetNextAlarm()
```
paulcacheux: Are you sure all callers of `updateKillQueueAlarmOnDisarm` are locking ruleDisarmersLock ? I'm not seeing that in the code

pkg/security/probe/process_killer.go
```diff
@@ -164,6 +156,17 @@ func (p *ProcessKiller) KillQueuedPidsAndSetNextAlarm() {
...
 
+// KillQueuedPidsAndSetNextAlarm performs all pending kills and sets the next alarm if any
+func (p *ProcessKiller) KillQueuedPidsAndSetNextAlarm() {
+	if !p.cfg.RuntimeSecurity.EnforcementEnabled {
+		return
+	}
+
+	p.ruleDisarmersLock.Lock()
+	p.killQueuedPidsAndSetNextAlarm()
+	p.ruleDisarmersLock.Unlock()
```
paulcacheux: defer is probably better, it would allow to unlock in case of a panic

pkg/security/probe/process_killer.go
YoannGh: Indeed none of the callers are locking `ruleDisarmersLock`, this was already the case before and is definitely an issue. I cannot simply take a lock `ruleDisarmersLock` because `disarmer.Lock` is already held here (and other code paths are using the `ruleDisarmersLock` -> `disarmer.Lock` order. Let me see how to fix that

pkg/security/probe/process_killer.go
```diff
@@ -145,7 +145,9 @@ func (p *ProcessKiller) KillQueuedPidsAndSetNextAlarm() {
 	now := time.Now()
 	var nextAlarm *time.Time
 	for ruleID, ruleDisarmer := range p.ruleDisarmers {
+		ruleDisarmer.m.Lock()
 		if ruleDisarmer.disarmed || len(ruleDisarmer.killQueue) == 0 {
+			ruleDisarmer.m.Unlock()
```
paulcacheux: I find this kind of pattern risky, could we extract this block into a method (allowing to use defers) ?

pkg/security/probe/process_killer.go
```diff
@@ -700,7 +712,7 @@ const (
 )
 
 type ruleDisarmer struct {
-	sync.Mutex
+	m               sync.Mutex
```
paulcacheux, line 726: was embedding the sync.Mutex causing any issue ?

YoannGh: No, adding a field name simply makes it easier for the IDE to find where the mutex is used

pkg/security/probe/process_killer.go
YoannGh: Done with https://github.com/DataDog/datadog-agent/pull/39370/commits/0cc112a83ae26faf181394e106eec9036933455c

YoannGh: 9952c031192c9ab1225e660884519da9d977e9d9

paulcacheux (approved):

PR #39372, opened 2025-07-30
trace/stats: Add HTTP Method and Endpoint for resource renaming

e-n-0:
### What does this PR do?

This PR adds the HTTP Method and HTTP Endpoint to the stats protobuf file.
This also includes all required changes to correctly construct a stat span with these new data.

#### Changes

As the method `NewStatSpan` is also used in the dd-trace-go tracer that is imported in some contrib (like serverless), we chose to deprecated this method for this release and replace it by `NewStatSpanWithHTTPEndpoint`.

### Additional Notes

- [RFC](https://docs.google.com/document/d/1n78MrcuX6J9YbPA-mpECS9hHfEvg12e2m2OjqCd2QTM/edit?tab=t.0#heading=h.o5gstqo08gu5)

- Tracer PR that use the new updated stats: https://github.com/DataDog/dd-trace-go/pull/3851

- [Intake PR](https://github.com/DataDog/dd-go/pull/193921)


drichards-87 (approved):

christophe-papazian (approved):

pkg/trace/stats/span_concentrator.go
```diff
@@ -109,15 +112,15 @@ func NewSpanConcentrator(cfg *SpanConcentratorConfig, now time.Time) *SpanConcen
...
-	return sc.NewStatSpan(s.Service, s.Resource, s.Name, s.Type, s.ParentID, s.Start, s.Duration, s.Error, s.Meta, s.Metrics, peerTags)
+	return sc.NewStatSpanWithHTTPEndpoint(s.Service, s.Resource, s.Name, s.Type, s.ParentID, s.Start, s.Duration, s.Error, s.Meta, s.Metrics, peerTags, "", "")
 }
 
-// NewStatSpan builds a StatSpan from the required fields for stats calculation
+// NewStatSpanWithHTTPEndpoint builds a StatSpan from the required fields for stats calculation
 // peerTags is the configured list of peer tags to look for
 // returns (nil,false) if the provided fields indicate a span should not have stats calculated
-func (sc *SpanConcentrator) NewStatSpan(
+func (sc *SpanConcentrator) NewStatSpanWithHTTPEndpoint(
```
ichinaski: I'm not sure what's the best way to add new fields to the `NewStatSpan` constructor, but I feel like this suffix may not be scalable if/when we edit these fields in the future.

Perhaps we can have a more generic `NewStatSpan` that takes a config object, wherein we add the parameters we need to construct the `StatSpan`? 🤔

e-n-0: We did not just added the new fields to the method because otherwise we have a conflicting build issue:
The `dd-trace-go` tracer dependency is imported in the agent (in the serverless cmd for example) and will refuse to compile (because referencing the agent itself during building):
```
# gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer
../../../../../../pkg/mod/gopkg.in/!data!dog/dd-trace-go.v1@v1.72.2/ddtrace/tracer/stats.go:161:80: not enough arguments in call to c.spanConcentrator.NewStatSpan
        have (string, string, string, string, uint64, int64, int64, int32, map[string]string, map[string]float64, []string)
        want (string, string, string, string, uint64, int64, int64, int32, map[string]string, map[string]float64, []string, string, string)
```

We could also rename that with `NewStatSpanWithConfig` and providing a struct of all parameters, and I guess it will be fine too.

e-n-0: I added `NewStatSpanWithConfig` in 96517d0

mackjmr (approved):

ichinaski (approved):
LGTM

PR #39374, opened 2025-07-30
fix(installer): Handle self-stop errors during promotion

BaptisteFoy:
### What does this PR do?
https://github.com/DataDog/datadog-agent/pull/36396 didn't fix the issue for `systemd.Restart`, so the original problem wasn't fixed for promotion

Manual QA


coignetp (approved):
🚀

PR #39375, opened 2025-07-30
[AMCC-85] dogstatsd: use the config API to restore previous values of the config

remeh:
### What does this PR do?

* Fix the override of the config source in `SetBlocklist` and 
* Start using `UnsetForSource` to better use the config API.

### Motivation

Better usage of the config API.

* Ran an Agent
* validated the local config was displayed on `agent config`
* remotely sent a new config, validated it was the one used with `agent config`
* deleted that remote config, validated that `agent config` was again showing the local config

Follow up of https://github.com/DataDog/datadog-agent/pull/39232

### DIFF
--- a/comp/dogstatsd/server/rc.go
+++ b/comp/dogstatsd/server/rc.go
@@ -11,6 +11,7 @@ import (
 	"maps"
 	"slices"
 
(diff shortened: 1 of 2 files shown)


vickenty (approved):

olivielpeau (approved):
✨

PR #39382, opened 2025-07-30
[dyninst/symdb] Capture inlined functions

piob-io:
https://datadoghq.atlassian.net/browse/DEBUG-4241

We capture abstract functions and inlined instances, and then reconcile after reading all packages.
No noticeable performance regression, still takes ~5s to extract symbols from cockroach binary.


pkg/dyninst/symdb/cli/main.go
```diff
@@ -44,33 +44,27 @@ The symbols from the specified binary will be extracted and printed to stdout
 		os.Exit(1)
 	}
 
-	logLevel := os.Getenv("DD_LOG_LEVEL")
```
andreimatei: revert this file

pkg/dyninst/symdb/symdb.go
```diff
@@ -315,6 +317,32 @@ type SymDBBuilder struct {
 	cleanupFuncs []func()
 }
 
+type abstractFunction struct {
```
andreimatei, line 341: give this a comment

pkg/dyninst/symdb/symdb.go
```diff
@@ -728,15 +728,25 @@ func (b *SymDBBuilder) ExtractSymbols(opt ExtractScope) (Symbols, error) {
...
 func (b *SymDBBuilder) reconcileAbstractFunctions() error {
+	abstractFunctions := make([]abstractFunction, 0, len(b.abstractFunctions))
 	for _, af := range b.abstractFunctions {
 		if af.name == "" {
 			// Abstract definition has not be collected, which is legitimate due to filtering.
 			continue
 		}
+		abstractFunctions = append(abstractFunctions, af)
+	}
+	sort.Slice(abstractFunctions, func(i, j int) bool {
```
andreimatei, line 778: give a comment about why we sort

pkg/dyninst/symdb/symdb.go
andreimatei: say that `instances` are accumulated as we explore other functions, and the other fields are populated when we find the abstract definition

pkg/dyninst/symdb/symdb.go
```diff
@@ -1076,85 +1291,115 @@ func pcRangeToLines(r dwarfutil.PCRange, lines []gosym.LineRange) (LineRange, bo
...
+
+func (b *SymDBBuilder) parseVariableLocations(
+	unit *dwarf.Entry,
+	block codeBlock,
+	entry *dwarf.Entry,
+	typeSize uint32,
+	lines []gosym.LineRange,
+	out []LineRange,
+) ([]LineRange, error) {
+	if locField := entry.AttrField(dwarf.AttrLocation); locField != nil {
```
andreimatei: nit: return early if `locField` is nil and unindent the happy path

pkg/dyninst/symdb/symdb.go
```diff
@@ -1076,85 +1291,115 @@ func pcRangeToLines(r dwarfutil.PCRange, lines []gosym.LineRange) (LineRange, bo
...
+	}, typ, nil
+}
+
+func (b *SymDBBuilder) parseVariableLocations(
+	unit *dwarf.Entry,
+	block codeBlock,
+	entry *dwarf.Entry,
+	typeSize uint32,
+	lines []gosym.LineRange,
+	out []LineRange,
```
andreimatei, line 1458: say that results are appended to this slice

pkg/dyninst/symdb/symdb.go
```diff
@@ -315,6 +317,32 @@ type SymDBBuilder struct {
...
+
+type inlinedInstance struct {
+	unit      *dwarf.Entry
+	lines     map[string]gosym.FunctionLines
+	variables []inlinedVariable
+}
+
+type inlinedVariable struct {
+	block codeBlock
+	entry *dwarf.Entry
```
andreimatei: we were talking about resolving and caching the abstract variables eagerly and thus getting rid of accumulating the inlined variables

pkg/dyninst/symdb/symdb.go
```diff
@@ -675,17 +703,112 @@ func (b *SymDBBuilder) ExtractSymbols(opt ExtractScope) (Symbols, error) {
...
+			continue
+		}
+		var file string
+		var startLine uint32
+		var endLine uint32
+		for _, instance := range af.instances {
+			if file == "" {
+				file = instance.lines[af.qualifiedName].File
+			}
+			lines := instance.lines[af.qualifiedName].Lines
```
andreimatei: What do you think about memoizing a min line and max line inside the `FunctionLines` struct, when we build the struct?

pkg/dyninst/symdb/symdb.go
```diff
@@ -675,17 +703,112 @@ func (b *SymDBBuilder) ExtractSymbols(opt ExtractScope) (Symbols, error) {
...
+			for _, line := range lines {
+				if startLine == 0 || line.Line < startLine {
```
andreimatei: We have a `decl_line` on the abstract definition. Shouldn't we use that as the start line?

pkg/dyninst/symdb/symdb.go
```diff
@@ -675,17 +703,112 @@ func (b *SymDBBuilder) ExtractSymbols(opt ExtractScope) (Symbols, error) {
...
+				Variables: variables,
+			},
+		}
+		if af.receiver != "" {
+			t := b.types.getType(af.receiver)
+			if t != nil {
+				t.Methods = append(t.Methods, f)
+			}
+		} else {
+			p := b.packages[af.pkg]
```
andreimatei: does `packages` need to be a field on the `symBuilder`, or could it be a local map in this function built from a slice passed in by the caller?

pkg/dyninst/symdb/symdb.go
```diff
@@ -728,15 +728,25 @@ func (b *SymDBBuilder) ExtractSymbols(opt ExtractScope) (Symbols, error) {
 			res.Packages = append(res.Packages, *pkg)
 		}
 	}
+	sort.Slice(res.Packages, func(i, j int) bool {
```
andreimatei, line 744: should we also sort the types inside packages?

pkg/dyninst/symdb/symdb.go
```diff
@@ -675,17 +703,112 @@ func (b *SymDBBuilder) ExtractSymbols(opt ExtractScope) (Symbols, error) {
 		if err != nil {
 			return Symbols{}, err
 		}
-		// Skip zero-value packages, and also packages that have no symbols in
-		// them. The latter happens for 3rd party packages when we're asking for
-		// 1st party code only.
-		if pkg.Name != "" && (len(pkg.Types) > 0 || len(pkg.Functions) > 0) {
-			res.Packages = append(res.Packages, pkg)
+		if pkg.Name != "" {
+			b.packages[pkg.Name] = &pkg
```
andreimatei: do this instead to not overwrite packages; it seems there are duplicates eating our functions
```
			if existingPkg, ok := b.packages[pkg.Name]; ok {
				existingPkg.Functions = append(existingPkg.Functions, pkg.Functions...)
				existingPkg.Types = append(existingPkg.Types, pkg.Types...)
			} else {
				b.packages[pkg.Name] = &pkg
			}
```
It happens mostly in the standard lib, but also elsewhere. Sometimes I think it happens because of asm functions. Other times I'm not sure.

pkg/dyninst/symdb/cli/main.go
piob-io: Done

pkg/dyninst/symdb/symdb.go
piob-io: Done

pkg/dyninst/symdb/symdb.go
piob-io: Done

pkg/dyninst/symdb/symdb.go
piob-io: Done

pkg/dyninst/symdb/symdb.go
piob-io: Done

pkg/dyninst/symdb/symdb.go
piob-io: Done

pkg/dyninst/symdb/symdb.go
piob-io: I don't like having the gosym library produce derived information, not all users will care about them. I could make this a helper function, if you think it's worth it. This got a bit better now that we process the lines immediately when encountering inlined instance.

pkg/dyninst/symdb/symdb.go
piob-io: We should, done

pkg/dyninst/symdb/symdb.go
piob-io: Could be, I don't have a strong opinion, I did it this way to make it similar to how types get accumulated. I imagine this part will be reworked anyway to support incremental uploads. Happy to change it up in the meantime, if you think it's better.

pkg/dyninst/symdb/symdb.go
piob-io: When it comes to sorting, I settled on the approach to sort the least amount of data. Types don't have to be sorted, because they are accumulated into a slice here, so we end up preserving order of whenver the type reference is first encountered in the dwarf. As above, I expect this code to change when we implement incremental uploads.

pkg/dyninst/symdb/symdb.go
piob-io: Ha! Thanks, done

piob-io:
Thanks! One more commit coming - if we don't have a type for an abstract method, we need to create one.

pkg/dyninst/symdb/symdb.go
```diff
@@ -748,61 +726,36 @@ func (b *SymDBBuilder) ExtractSymbols(opt ExtractScope) (Symbols, error) {
...
+	case ExtractScopeMainModuleOnly:
+		return strings.HasPrefix(pkgName, b.mainModule)
+	case ExtractScopeModulesFromSameOrg:
+		return strings.HasPrefix(pkgName, b.firstPartyPkgPrefix)
+	default:
+		panic(fmt.Sprintf("unsupported extract scope: %d", b.scopeFilter))
+	}
+}
+
 func (b *SymDBBuilder) reconcileAbstractFunctions() error {
```
andreimatei: give this function a comment even if it's smaller now. And maybe rename to `addAbstractFunctions`?

pkg/dyninst/symdb/symdb.go
```diff
@@ -683,17 +719,126 @@ func (b *SymDBBuilder) ExtractSymbols(opt ExtractScope) (Symbols, error) {
...
+				StartLine: int(startLine),
+				EndLine:   int(endLine),
+				Scopes:    nil,
+				Variables: variables,
+			},
+		}
+		if af.receiver != "" {
+			t := b.types.getType(af.receiver)
+			if t == nil {
+				// Some types are empty structures, and functions that
```
andreimatei, line 805: how was this case handled before?

pkg/dyninst/symdb/symdb.go
andreimatei: I think it'd be worth keeping this map more local if you don't mind

pkg/dyninst/symdb/symdb.go
andreimatei: nit: move the comment up a bit to make it clear that copying the functions to the slice is done solely for the sorting.  But this sorting of abstract functions is a little subtle -- the other functions will be listed in the order they show up in the compile unit, and then the abstract ones appear alphabetically. How about we more bluntly sort all the output at the end?

andreimatei (approved):
LGTM nice

pkg/dyninst/symdb/symdb.go
piob-io: Done

pkg/dyninst/symdb/symdb.go
piob-io: Must not be happening for non-inlined functions, we return an error there.

piob-io:
Thank you!

Github doesn't let me respond to the last two comments, but done and done.
I kept current sorting placement - it is next to things that are stored in maps, so it's obvious why we need to sort it, and not other stuff. Otherwise, sorting whole result structure gets bizarre, e.g. should we recurse into all scopes to sort them and variables? Anyway, I believe this is gonna change with incremental uploads.

PR #39384, opened 2025-07-30
Disable temporarily stack cleaner

CelianR:
### What does this PR do?

We have some OOM kills, disabling stackcleaner while we have autoscaling issues.


KevinFairise2 (approved):
😢

PR #39386, opened 2025-07-30
dyninst/symdb: flush the logs on exit

andreimatei:
Otherwise we're sometimes missing a tail.

### DIFF
--- a/pkg/dyninst/symdb/cli/main.go
+++ b/pkg/dyninst/symdb/cli/main.go
@@ -101,6 +101,11 @@ func run(binaryPath string) error {
 	log.Infof("Symbol extraction completed in %s.", time.Since(start))
 	stats := statsFromSymbols(symbols)
 	log.Infof("Symbol statistics for %s: %+v", binaryPath, stats)
+	if !*silent {
+		symbols.Serialize(symdbutil.MakePanickingWriter(os.Stdout))
+	} else {
+		log.Infof("--silent specified; symbols not serialized.")
+	}
 
 	return nil
 }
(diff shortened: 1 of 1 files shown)


pkg/dyninst/symdb/cli/main.go
```diff
@@ -49,13 +49,16 @@ The symbols from the specified binary will be extracted and printed to stdout
 		logLevel = "info"
 	}
 	log.SetupLogger(log.Default(), logLevel)
+	defer log.Flush()
```
ajwerner, line 52: you know this won't get called on `os.Exit`, right?

ajwerner (commented):
I think you should refactor main to call some function that returns an error so that there's only one exit path out of this binary

ajwerner (approved):
I guess you did cover what I wanted. Sorry, just the defer is kinda pointless

pkg/dyninst/symdb/cli/main.go
andreimatei: that's why I call it again below.

all the logic is already part of `run`, but I think this `defer` doesn't hurt

piob-io:
Do we benefit from using this log package here? I noticed the same flushing issue, and just switched this to use fmt.Print (not yet pushed). Is there a better log package that flushes on exit, and immediately on errors too?

pkg/dyninst/symdb/cli/main.go
```diff
@@ -49,13 +49,16 @@ The symbols from the specified binary will be extracted and printed to stdout
...
 	// Start the pprof server.
 	go func() {
 		_ = http.ListenAndServe(fmt.Sprintf("localhost:%d", *pprofPort), nil)
 	}()
 
-	if err := run(*binaryPath); err != nil {
+	err := run(*binaryPath)
+	log.Flush()
+	if err != nil {
 		log.Errorf("Error: %v", err)
```
piob-io, line 60: This error needs to be flushed :/

andreimatei: flushing here flushing there flushing everywhere

PR #39394, opened 2025-07-30
[dyninst/tests] Unredact stack traces now that file paths are stable

piob-io:
We keep redacting stack in e2e test, and that involves 3rd party file paths, that can easily change.
We keep redacting runtime.goexit file path, as it is architecture-dependant.

### DIFF
--- a/pkg/dyninst/end_to_end_test.go
+++ b/pkg/dyninst/end_to_end_test.go
@@ -549,6 +549,10 @@ func waitForLogMessages(
 	{
 		redactors := append(make([]jsonRedactor, 0, len(defaultRedactors)), defaultRedactors...)
 		redactors = append(redactors, redactor(
+			prefixMatcher("/debugger/snapshot/stack"),
+			replacement(`"[stack]"`),
+		))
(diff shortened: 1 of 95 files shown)


pkg/dyninst/end_to_end_test.go
```diff
@@ -548,6 +548,12 @@ func waitForLogMessages(
 	var content []byte
 	{
 		redactors := append(make([]jsonRedactor, 0, len(defaultRedactors)), defaultRedactors...)
+		redactors = append(redactors, redactor(
```
andreimatei, line 551: why do we need this?

andreimatei (approved):

pkg/dyninst/end_to_end_test.go
piob-io: This test probes a binary with 3rd party code, that shows in the stack. Don't want to cause test failures on dependency version update.

piob-io:
Thank you!

ajwerner:
@piob-io sadly this reveals some issue with the stack walking on arm where we get a surprising extra frame

PR #39395, opened 2025-07-30
Global Datadog() can only read/write config, not build it

dustmop:
### What does this PR do?

Change the global `Datadog()` method so that it no longer has the methods from `model.Setup`.

### Motivation

To more cleanly separate the building and usage of the config, remove all `model.Setup` methods from the interface that is returned by the global `Datadog()` accessor.

Introduce the `model.BuildableConfig` interface, which is able to both build and read/write the config. It is only meant to be used during initial construction.

Finally, introduce `config.RevertFinishedBackToBuilder()` which is an escape hatch for OTel, since they build on top of the already built config.

Unit tests already exist to cover these changes


paulcacheux (approved):

gabedos (approved):

pkg/config/model/types.go
```diff
@@ -276,15 +276,26 @@ type Compound interface {
 	ReadConfig(in io.Reader) error
 	MergeConfig(in io.Reader) error
 	MergeFleetPolicy(configPath string) error
+
+	// An escape hatch to build on top of the config after it has completed initialization
+	NewBuilder() BuildableConfig
```
hush-hush: I feel like the name implies that a empty/new builder will be returned.
Also could we expand on the comment explaining that this is a hack made of OTEL and should not be used anywhere else ?

Also, is the expectation that the user will call `BuildSchema` again ? If so, we should explain it.

hush-hush (changes requested):

julien-lebot (approved):

pkg/config/model/types.go
dustmop: Done, this is now `RevertFinishedBackToBuilder()`. Added comments to each method definition, and the relevant callsites, talking about the purpose of the call.

hush-hush (approved):

astuyve (approved):

gjulianm (approved):

liustanley (approved):

mbakht (approved):

PR #39397, opened 2025-07-31
dyninst/procmon: no `go` under a mutex

ajwerner:
There's only ever one analysis in flight at a time already. Let's just keep one goroutine around to deal with that. Interestingly we weren't seeing big wins in the number of dropped events because calling `go` is also the sort of thing that can get you unscheduled.

Relates to [DEBUG-4225](https://datadoghq.atlassian.net/browse/DEBUG-4225).

[DEBUG-4225]: https://datadoghq.atlassian.net/browse/DEBUG-4225?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/dyninst/procmon/process_monitor.go
+++ b/pkg/dyninst/procmon/process_monitor.go
@@ -71,12 +71,13 @@ type ProcessMonitor struct {
 
 	mu struct {
 		sync.Mutex
-		state    state
-		shutdown bool
+		state       state
+		isClosed    bool
+		analyzeChan chan<- uint32
 	}
 
-	wg           sync.WaitGroup
-	shutdownOnce sync.Once
+	wg        sync.WaitGroup
+	closeOnce sync.Once
 }
 
 // NewProcessMonitor creates a new ProcessMonitor that will send updates to the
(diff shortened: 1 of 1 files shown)


pkg/dyninst/procmon/process_monitor.go
```diff
@@ -178,34 +185,45 @@ var analysisFailurePermissionLogLimiter = rate.NewLimiter(rate.Every(10*time.Min
...
+	})
+}
+
+type lockedProcessMonitor ProcessMonitor
+
+func (pm *lockedProcessMonitor) analyzeProcess(pid uint32) {
+	select {
+	case pm.mu.analyzeChan <- pid:
+	default:
+		// This should never happen, but if it does, we'll log it and shutdown
```
tylfin, line 215: If this gets called twice before `pm.analyzeProcess` finishes executing, wouldn't that cause it to fall through into the default case?

ajwerner: Elsewhere we have an invariant that it doesn't. You can look at `state.go`

piob-io (approved):

PR #39404, opened 2025-07-31
[AGENTRUN-602] Update CLCRunnerClient/DCAClient/RemoteTagger to support cross-node TLS verification

misteriaud:
### What does this PR do?
This PR update the CLCRunnerClient/DCAClient/RemoteTagger to support cross-node TLS verification via the `cluster_agent.enable_tls_verification` configuration entry.

When set to `true`, these clients will rely on the global IPC TLC configuration setup by the IPC component.

The CLCRunnerClient and ClusterAgentClient unit tests have been updated to cover the new TLS verification feature.
Manually tested on local kind cluster.


jeremy-hanna (approved):
👍 nice work on this! very clean!

rahulkaukuntla (approved):

pkg/util/clusteragent/clcrunner.go
```diff
@@ -68,14 +68,19 @@ func (c *CLCRunnerClient) init() {
 	c.clcRunnerAPIRequestHeaders = http.Header{}
 	c.clcRunnerAPIRequestHeaders.Set(authorizationHeaderKey, fmt.Sprintf("Bearer %s", authToken))
 
+	// Set TLS config
+	crossNodeClientTLSConfig, err := pkgapiutil.GetCrossNodeClientTLSConfig()
+	if err != nil {
+		c.initErr = fmt.Errorf("failed to get cross-node client TLS config: %w", err)
+		return
```
wdhif, line 75: Would it make sense to allow it to fallback to `InsecureSkipVerify` for a little bit? Or having it behind a config for now?

pkg/util/clusteragent/clusteragent.go
```diff
@@ -166,6 +166,13 @@ func (c *DCAClient) startReconnectHandler(reconnectPeriod time.Duration) {
 
 func (c *DCAClient) initHTTPClient() error {
 	var err error
+
+	// Get Cross Node Client TLS Config
+	tlsConfig, err := pkgapiutil.GetCrossNodeClientTLSConfig()
+	if err != nil {
+		return fmt.Errorf("failed to get cross-node client TLS config: %w", err)
```
wdhif, line 173: Same as https://github.com/DataDog/datadog-agent/pull/39404/files?file-filters%5B%5D=.go&file-filters%5B%5D=.mod&owned-by%5B%5D=wdhif&show-viewed-files=true#r2256604042

pkg/util/clusteragent/clcrunner.go
misteriaud: I did this way to avoid having insecure-by-default client behavior. Having fallback on `InsecureSkipVerify` would mean that if we forget to initialize the `crossNodeClientTLSConfig` for a given codepath, the connection become insecure silently, which makes it very hard to track insecure connection. 
Since the CLCRunnerClient and DCAClient are stored in global variables and initialized outside of the component/dependency framework, it's pretty hard to assert that the TLS configuration has been initialized properly before the clients.
In reality the secure/insecure state is defined by the `cluster_agent.enable_tls_verification` config entry, which is `false` by default. 

The best would be to migrate these clients to the component framework to have clear dependency definition on the IPC component, but this would require a bigger refactor, which is not worth it IMO

wdhif: Understood, we will need to be particularly careful if we want to enable this by default as we would need to "guarantee" that if it's enabled, the TLS configuration has been correctly initialized 👍

wdhif (approved):

pkg/util/clusteragent/clcrunner.go
misteriaud: Exactly, that's why the configuration `DD_CLUSTER_AGENT_ENABLE_TLS_VERIFICATION=true` will be set progressively only for deployments methods that we're sure are supporting it

PR #39405, opened 2025-07-31
Add more tests to orchestrator check processor

xlucas:
Add tests to the processor component.
In order to allow testing time-dependent logic, it changes the processor context contract to accept a clock interface.


pkg/collector/corechecks/cluster/orchestrator/processors/chunking_test.go
```diff
@@ -0,0 +1,211 @@
+package processors
```
xlucas: moved from processor_test.go

stzou (approved):

kangyili (approved):

PR #39409, opened 2025-07-31
Do not rely on Github API to download protoc

KevinFairise2:
### What does this PR do?

Relying on Github API exposes us to Github rate limit, that can be an issue since it is IP based.
By downloading the file directly with `urlretrieve` we are less exposed to that rate limit

### Motivation

Make sure the setup tasks work

### DIFF
--- a/tasks/install_tasks.py
+++ b/tasks/install_tasks.py
@@ -4,10 +4,10 @@ import shutil
 import sys
 import zipfile
 from pathlib import Path
+from urllib.request import urlretrieve
 
 from invoke import Context, Exit, task
 
-from tasks.libs.ciproviders.github_api import GithubAPI
 from tasks.libs.common.color import Color, color_message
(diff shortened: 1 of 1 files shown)


tasks/install_tasks.py
```diff
@@ -4,10 +4,10 @@
 import sys
 import zipfile
 from pathlib import Path
+from urllib.request import urlretrieve
```
Ishirui, line 7: Having a look at the venv installed by `dda` for running tasks, I see `urllib3` is installed, but only as a transitive dependency - it's not defined anywhere in the `pyproject.toml`.

We do have `requests` as an explicit dependency however, and that is even probably where the `urllib` transitive dependency comes from. Could we use that instead ? It's also IIRC the "higher-level" library, so probably better suited for things like this.

Ishirui (approved):

PR #39412, opened 2025-07-31
[EBPF-770] gpu: fix concurrency issues in handling kernel spans

gjulianm:
### What does this PR do?

This PR changes the data structure used to store pending kernel/memory spans in the stream handlers, from a list to a channel. It also changes the naming of kernelSpans/allocations to have a more uniform naming using "span" for consistency.

### Motivation

Avoid panics due to race conditions, where one goroutine can be clearing the pending kernel spans/allocations while another writes into the array. 

I used a channel with non-blocking reads/writes instead of a mutex, as the implementation was simpler and channels are faster than the mutex locking.

Unit tests. Added a specific test with concurrent consumer/producer to reproduce the original issue with the race detector enabled.

### DIFF
--- a/pkg/config/setup/system_probe.go
+++ b/pkg/config/setup/system_probe.go
@@ -438,7 +438,8 @@ func InitSystemProbeConfig(cfg pkgconfigmodel.Config) {
 	cfg.BindEnvAndSetDefault(join(gpuNS, "ring_buffer_pages_per_device"), 32) // 32 pages = 128KB by default per device
 	cfg.BindEnvAndSetDefault(join(gpuNS, "max_kernel_launches_per_stream"), 1000)
 	cfg.BindEnvAndSetDefault(join(gpuNS, "max_mem_alloc_events_per_stream"), 1000)
-	cfg.BindEnvAndSetDefault(join(gpuNS, "max_pending_spans_per_stream"), 1000)
+	cfg.BindEnvAndSetDefault(join(gpuNS, "max_pending_kernel_spans_per_stream"), 1000)
+	cfg.BindEnvAndSetDefault(join(gpuNS, "max_pending_memory_spans_per_stream"), 1000)
 	cfg.BindEnvAndSetDefault(join(gpuNS, "max_streams"), 100)
 	cfg.BindEnvAndSetDefault(join(gpuNS, "max_stream_inactivity_seconds"), 30) // 30 seconds by default, includes two checks at the default interval of 15 seconds
 
--- a/pkg/config/setup/system_probe.go
+++ b/pkg/config/setup/system_probe.go
@@ -438,6 +438,7 @@ func InitSystemProbeConfig(cfg pkgconfigmodel.Config) {
 	cfg.BindEnvAndSetDefault(join(gpuNS, "ring_buffer_pages_per_device"), 32) // 32 pages = 128KB by default per device
 	cfg.BindEnvAndSetDefault(join(gpuNS, "max_kernel_launches_per_stream"), 1000)
 	cfg.BindEnvAndSetDefault(join(gpuNS, "max_mem_alloc_events_per_stream"), 1000)
+	cfg.BindEnvAndSetDefault(join(gpuNS, "max_pending_spans_per_stream"), 1000)
 	cfg.BindEnvAndSetDefault(join(gpuNS, "max_streams"), 100)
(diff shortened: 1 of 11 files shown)


pkg/config/setup/system_probe.go
```diff
@@ -438,6 +438,7 @@ func InitSystemProbeConfig(cfg pkgconfigmodel.Config) {
 	cfg.BindEnvAndSetDefault(join(gpuNS, "ring_buffer_pages_per_device"), 32) // 32 pages = 128KB by default per device
 	cfg.BindEnvAndSetDefault(join(gpuNS, "max_kernel_launches_per_stream"), 1000)
 	cfg.BindEnvAndSetDefault(join(gpuNS, "max_mem_alloc_events_per_stream"), 1000)
+	cfg.BindEnvAndSetDefault(join(gpuNS, "max_pending_spans_per_stream"), 1000)
 	cfg.BindEnvAndSetDefault(join(gpuNS, "max_streams"), 100)
 	cfg.BindEnvAndSetDefault(join(gpuNS, "max_stream_inactivity_seconds"), 30) // 30 seconds by default, includes two checks at the default interval of 15 seconds
```
val06: let's make changes to the naming of the params, before GA...
i propose:

```
gpu_monitoring:
	streams:
		total (changed from max_streams)
		timeout (changed from max_stream_inactivity_seconds)
		max_kernels (changed from max_kernel_launches_per_stream)
		max_pending_spans (changed from max_pending_spans_per_stream)
```

val06: also in the config code, perhaps create a streamConfig struct and have those values inside of it (with respective renaming). 

(for the `timeout`, perhaps you can call the var `timeoutInSeconds` for easier readability of code)

pkg/gpu/stream.go
```diff
@@ -168,6 +169,9 @@ func newStreamHandler(metadata streamMetadata, sysCtx *systemContext, limits str
 		return nil, fmt.Errorf("failed to create memAllocEvents cache: %w", err)
 	}
 
+	sh.pendingKernelSpans = make(chan *kernelSpan, limits.maxPendingSpans)
+	sh.pendingMemorySpans = make(chan *memorySpan, limits.maxPendingSpans)
```
val06: do we really want to use the same size for memory and kernels? i am not aware of such correlation between them

pkg/gpu/stream.go
```diff
@@ -322,42 +344,59 @@ func getAssociatedAllocations(span *kernelSpan) []*memoryAllocation {
...
-func (sh *StreamHandler) getPastData(flush bool) *streamData {
-	if len(sh.kernelSpans) == 0 && len(sh.allocations) == 0 {
-		return nil
+func consumeChannel[T any](ch chan T, count int) []T {
+	items := make([]T, 0, count)
+
+loop:
+	for len(items) < count {
+		select {
+		case item := <-ch:
```
val06: don't you need to check that the channel is not closed?

pkg/gpu/stream.go
```diff
@@ -322,42 +344,59 @@ func getAssociatedAllocations(span *kernelSpan) []*memoryAllocation {
...
+			items = append(items, item)
+		default:
+			// We shouldn't actually hit this, as we should stop consuming the
+			// channel when the count is reached and nothing else is consuming
+			// from the channel, but break just in case to avoid a deadlock
+			break loop
```
val06: instead of that, can't you just use `return items` and remove the loop label?

val06 (commented):
reviewed

hush-hush (approved):

pkg/config/setup/system_probe.go
gjulianm: Makes sense, I can make it in a separate PR

pkg/gpu/stream.go
gjulianm: Changed to use different limits

pkg/gpu/stream.go
gjulianm: I'm not closing the channel, as long as the stream handler stays alive the channel does so too.

pkg/gpu/stream.go
gjulianm: Changed

val06 (approved):

PR #39414, opened 2025-07-31
Pass PULUMI_K8S_DELETE_UNREACHABLE to clean task

KevinFairise2:
### What does this PR do?

When Kubernetes clusters are already deleted/unreachable we should delete them to avoid having the stack stuck in destroying.


CelianR (approved):
Nice catch

PR #39415, opened 2025-07-31
Add dogstatsd metrics to crossorg telemetry

StephenWakely:
### What does this PR do?

Adds selected dogstatsd metrics to metrics posted to cross org telemetry.

### Motivation

It would be useful to observe the usage of the dogstatsd client libraries. Which versions are being used, which ones cause problems with dropped metrics etc...

Since these metrics are posted by dogstatsd and do not form part of the internal agents telemetry, this intercepts the metrics and when a given metric is encountered it adds this as an internal metric. The metrics are added to the cross origin telemetry list.

I tested manually using the Ruby dogstatsd client to check that the dogstatsd metrics and tags are now being sent in the instrumentation telemetry payload.

### DIFF
--- a/pkg/aggregator/agent_telemetry.go
+++ b/pkg/aggregator/agent_telemetry.go
@@ -34,29 +34,36 @@ func newAgentTelemFilterList() utilstrings.FilterList {
 		"datadog.dogstatsd.client.packets_dropped"}, false)
 }
 
+// getDetailsFromSerie extracts the relevant details from the metrics serie
+// that we can use to post to internal telemetry.
+func getDetailsFromSerie(serie *metrics.Serie) (float64, []string) {
+	tags := []string{"", "", ""}
+	serie.Tags.ForEach(func(tag string) {
+		t := strings.SplitN(tag, ":", 2)
+		switch t[0] {
+		case "client":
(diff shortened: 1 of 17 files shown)


pkg/aggregator/agent_telemetry.go
```diff
@@ -0,0 +1,69 @@
...
+		"datadog.dogstatsd.client.packets_sent",
+		"datadog.dogstatsd.client.packets_dropped"}, false)
+}
+
+// getDetailsFromSerie extracts the relevant details from the metrics serie
+// that we can use to post to internal telemetry.
+func getDetailsFromSerie(serie *metrics.Serie) (float64, []string) {
+	tags := []string{"", "", ""}
+	serie.Tags.ForEach(func(tag string) {
+		t := strings.SplitN(tag, ":", 2)
```
remeh, line 43: Maybe worth a `if len(t) > 1`

remeh (commented):
That looks good to me!

In your case I guess it's sufficient, but that will work for series only. Would it be worth mentioning it in the good places?

StephenWakely:
Putting on hold for the minute as there are some security issues around taking external untrusted metrics. Need to think these through before moving forward.

PR #39419, opened 2025-07-31
Profiling: add a retry logic

r1viollet:
#### Retry Logic Implementation:

Add retry mechanism for profile uploads with exponential backoff (3 attempts, 100ms base delay, 2s max delay)
Clone requests to enable proper retry handling while preserving original context timeout
Implement fresh transport creation for retry attempts to avoid stale connection issues

**Connection Pool Issue Mitigation:**
Address the "poisoned connection" problem where:
HTTP client maintains persistent connections to load balancer
Backend failure causes load balancer's backend server to become unhealthy
Client reuses existing connection to load balancer
Load balancer returns HTTP 502 Bad Gateway but keeps connection open
Subsequent requests reuse the same "poisoned" connection → repeated 502s
Solution: Use transport.Clone() for retry attempts to force fresh connections and bypass stale connection pools.

#### Enhanced Error Classification:

Test and classify retryable vs non-retryable errors for both connection-level and HTTP status code errors
Implement proper HTTP status code semantics:
408 Request Timeout: Body read timeouts (more accurate than 502)
503 Service Unavailable: Retryable connection errors
504 Gateway Timeout: Context deadline exceeded
502 Bad Gateway: Non-retryable errors (default)

#### Improved Observability:

Add comprehensive logging around request body read failures with payload size and duration metrics
Enhance retry attempt logging with detailed error context and fresh transport usage
Increase log throttling limits (50/second) for better debugging visibility during issues
Add WARN-level logging for all retry attempts and connection errors for production debugging


pkg/trace/api/profiles.go
```diff
@@ -157,12 +164,148 @@ type multiTransport struct {
...
 }
 
+// isRetryableError determines if an error should trigger a retry
+func isRetryableError(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	// Check for network-level errors that indicate connection issues
+	if netErr, ok := err.(net.Error); ok {
```
ajgajg1134: For these type assertions on an error I would use `errors.As` to ensure that wrapped errors are OK here

e.g.
```
	// Check for network-level errors that indicate connection issues
	var netErr net.Error
	if errors.As(err, &netErr) {
```

pkg/trace/api/profiles.go
```diff
@@ -157,12 +164,148 @@ type multiTransport struct {
...
+			// Common retryable syscall errors
+			switch *syscallErr {
+			case syscall.ECONNREFUSED, syscall.ECONNRESET, syscall.ETIMEDOUT, syscall.ENETUNREACH, syscall.EHOSTUNREACH:
+				return true
+			}
+		}
+	}
+
+	// Check for URL errors that might be retryable
+	if urlErr, ok := err.(*url.Error); ok {
```
ajgajg1134: I think you can drop this check, as far as I can tell a url.Error should never be temporary as I believe these can only happen from failing to parse the provided URL

pkg/trace/api/profiles.go
```diff
@@ -173,27 +316,30 @@ func (m *multiTransport) RoundTrip(req *http.Request) (rresp *http.Response, rer
...
 	}()
+
 	if len(m.targets) == 1 {
-		setTarget(req, m.targets[0], m.keys[0])
-		return m.rt.RoundTrip(req)
+		return retryRoundTrip(req, m.targets[0], m.keys[0])
 	}
+
+	// Multiple targets - read body once for all requests
 	slurp, err := io.ReadAll(req.Body)
```
ajgajg1134: It seems like the retry function here is duplicating the reading of the body here?

ajgajg1134 (changes requested):

pkg/trace/api/profiles.go
```diff
@@ -157,12 +164,148 @@ type multiTransport struct {
...
+	// retryRoundTrip performs a single request with retry logic
+	retryRoundTrip := func(req *http.Request, u *url.URL, apiKey string) (*http.Response, error) {
+		var lastErr error
+		var resp *http.Response
+		var bodyBytes []byte
+
+		// Read the request body once for all retry attempts
+		if req.Body != nil {
+			var err error
+			bodyBytes, err = io.ReadAll(req.Body)
```
r1viollet: so reading the body is not great, though it could get consumed by the initial request even if we need to retry

pkg/trace/api/profiles.go
r1viollet: good catch! Also I'll feel less bad about reading this.

r1viollet:
Closing this one in favour of https://github.com/DataDog/datadog-agent/pull/40045
No strong evidence that agent side retries are needed.

PR #39420, opened 2025-07-31
[dyninst] Address all different framelessness cases

piob-io:
https://datadoghq.atlassian.net/browse/DEBUG-4094

Handle frameless/framefull cases on each architecture, for cfa calculation and stack unwinding.
Correctly determine framelessness for concrete and inlined subroutines, on each architecture.

This PR adds tests and unredacts stacks, polluting the change diff. However, individual commits that change actual logic should have limited test diffs.

### DIFF
--- a/pkg/dyninst/irgen/irgen.go
+++ b/pkg/dyninst/irgen/irgen.go
@@ -140,24 +140,38 @@ func generateIR(
 		cfg.maxDynamicTypeSize,
 		cfg.maxHashBucketsSize,
 	)
-	dwarfRes, err := processDwarf(interests, d, typeCatalog, objFile)
+	pendingSubprograms, err := processDwarf(interests, d, typeCatalog, objFile)
 	if err != nil {
 		return nil, err
 	}
-	pendingSubprograms := dwarfRes.pendingSubprograms
-	abstractIdx := dwarfRes.firstAbstractIdx
 
-	// Determine injection points right after each subprogram's prologue.
-	prologueResults, err := computePrologueResults(
-		objFile, pendingSubprograms, abstractIdx,
-	)
+	// Find prologues need to determine injection points. We make an assumption
+	// that prologue, if function has a frame, should be contained within the
+	// first pc range of a subprogram. This simplifies the logic slightly.
+	prologueSearch := make([]prologueSeachParams, 0, len(pendingSubprograms))
+	for _, sp := range pendingSubprograms {
+		if len(sp.subprogram.OutOfLinePCRanges) > 0 {
+			prologueSearch = append(prologueSearch, prologueSeachParams{
+				unit:    sp.unit,
+				pcRange: sp.subprogram.OutOfLinePCRanges[0],
+			})
+		}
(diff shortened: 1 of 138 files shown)


pkg/dyninst/irgen/irgen.go
```diff
@@ -140,24 +140,38 @@ func generateIR(
...
-	// Determine injection points right after each subprogram's prologue.
-	prologueResults, err := computePrologueResults(
-		objFile, pendingSubprograms, abstractIdx,
-	)
+	// Find prologues need to determine injection points. We make an assumption
+	// that prologue, if function has a frame, should be contained within the
+	// first pc range of a subprogram. This simplifies the logic slightly.
+	prologueSearch := make([]prologueSeachParams, 0, len(pendingSubprograms))
+	for _, sp := range pendingSubprograms {
+		if sp.subprogram.OutOfLinePCRanges != nil {
```
ajwerner: It's more correct to check if `len(sp.subprogram.OutOfLinePCRanges) > 0` in case later we add some sort of subslicing or pooling.

pkg/dyninst/irgen/irgen.go
```diff
@@ -196,101 +210,68 @@ func generateIR(
...
-			cmp.Compare(getPC(a), getPC(b)),
-		)
+	searchParams []prologueSeachParams,
+) (map[uint64]prologueEndLocation, error) {
+	slices.SortFunc(searchParams, func(a, b prologueSeachParams) int {
+		return cmp.Compare(a.pcRange[0], b.pcRange[0])
 	})
 
-	results := make([]prologueResult, len(pending))
+	res := make(map[uint64]prologueEndLocation)
```
ajwerner: nit: if you compact the ranges to eliminate duplicates, you can pass the correct size to this make call to avoid reallocations.

pkg/dyninst/irgen/irgen.go
```diff
@@ -1140,14 +1117,35 @@ func newProbe(
 		}, nil
 	}
 	if subprogram.OutOfLinePCRanges != nil {
-		injectionPoints = append(injectionPoints, prologueEnd)
+		pc := subprogram.OutOfLinePCRanges[0][0]
+		loc := prologueLocs[pc]
```
ajwerner: should we check here and below to see if we have a miss? If we do it'll be just a zero value which could later be confusing.

ajwerner (approved):
LGTM

pkg/dyninst/irgen/irgen.go
piob-io: Done. I used nil, because logically, program either has out of line ranges (concrete or inlined but with outofline instance), and then these ranges can have some length (although always more than zero), or it doesn't have them at all. But this is safer.

pkg/dyninst/irgen/irgen.go
piob-io: Done

pkg/dyninst/irgen/irgen.go
piob-io: Done

PR #39423, opened 2025-07-31
pkg/dyninst: Fix map slots entry size calculation

grantseltzer:
### What does this PR do?

There was an issue with the way I calculated entry size of the swiss map slots array. This fixes it to use info from DWARF instead.

+ test cases
+ a couple style fixes

### Motivation

Fixing map support in dynamic instrumentation

Added more test cases

### DIFF
--- a/pkg/dyninst/decode/types.go
+++ b/pkg/dyninst/decode/types.go
@@ -148,7 +148,7 @@ func newDecoderType(
 		}
 		slotsFieldType, ok := types[slotsField.Type.GetID()]
 		if !ok {
-			return nil, errors.New("type map slot field not found in types: " + s.GroupType.Name)
+			return nil, fmt.Errorf("type map slot field not found in types: %s", s.GroupType.Name)
 		}
 		ctrlField, err := getFieldByName(s.GroupType.RawFields, "ctrl")
 		if err != nil {
@@ -158,18 +158,18 @@ func newDecoderType(
 		ctrlSize := ctrlField.Type.GetByteSize()
 		entryArray, ok := slotsFieldType.(*ir.ArrayType)
 		if !ok {
-			return nil, errors.New("type map slot field is not an array type: " + slotsFieldType.GetName())
+			return nil, fmt.Errorf("type map slot field is not an array type: %s", slotsFieldType.GetName())
 		}
 		noalgstructType, ok := entryArray.Element.(*ir.StructureType)
 		if !ok {
-			return nil, errors.New("type map entry array element is not a structure type: " + entryArray.Element.GetName())
+			return nil, fmt.Errorf("type map entry array element is not a structure type: %s", entryArray.Element.GetName())
 		}
 		keyField, err := getFieldByName(noalgstructType.RawFields, "key")
 		if err != nil {
 			return nil, fmt.Errorf("malformed swiss map header type: %w", err)
 		}
 		if keyField == nil {
-			return nil, errors.New("type map entry array element has no key field: " + entryArray.Element.GetName())
+			return nil, fmt.Errorf("type map entry array element has no key field: %s", entryArray.Element.GetName())
 		}
 		elem, err := getFieldByName(noalgstructType.RawFields, "elem")
 		if err != nil {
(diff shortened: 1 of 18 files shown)


pkg/dyninst/decode/types.go
```diff
@@ -58,14 +58,15 @@ type goHMapBucketType ir.GoHMapBucketType
...
-	valueFieldOffset uint32
+	keyTypeID           ir.TypeID
+	keyTypeName         string
+	keyTypeSize         uint32
+	valueTypeID         ir.TypeID
+	valueTypeName       string
+	valueTypeSize       uint32
+	keyFieldOffset      uint32
+	valueFieldOffset    uint32
+	slotsArrayEntrySize uint32
```
ajwerner, line 69: nit: `slotSize`?

pkg/dyninst/testdata/decoded/sample/testMapLargeKeyLargeValue.json
```diff
@@ -0,0 +1,40 @@
...
+              "m": {
+                "type": "map[[4]int][4]int",
+                "size": "10",
+                "entries": [],
+                "notCapturedReason": "pruned"
+              }
+            }
+          }
+        }
+      }
```
ajwerner: What's going on here? This doesn't seem very big to me

pkg/dyninst/testdata/decoded/sample/testMapWithSmallKeyAndValueMassive.json
```diff
@@ -0,0 +1,40 @@
...
+        "probe": {
+          "id": "testMapWithSmallKeyAndValueMassive",
+          "location": {
+            "method": "main.testMapWithSmallKeyAndValueMassive"
+          }
+        },
+        "captures": {
+          "entry": {
+            "arguments": {
+              "redactMyEntries": {
```
ajwerner: Is this actually massive? I'm a little surprised you need to prune here and I have a feeling it'll change with https://github.com/DataDog/datadog-agent/pull/39356 so rebase that one first and rerun

pkg/dyninst/testdata/decoded/sample/testMapLargeKeyLargeValue.json
grantseltzer: It appears this will be 'fixed' by https://github.com/DataDog/datadog-agent/pull/39356 as well. 


https://github.com/DataDog/datadog-agent/blob/main/pkg/dyninst/decode/marshal.go#L252

The element type size is 520 bytes, and the limit is 512 so this evaluates to 0 groups. Perhaps we can do a partial evaluation? I'm not familiar with how data gets cut off and whether or not we know it's safe.

pkg/dyninst/testdata/decoded/sample/testMapWithSmallKeyAndValueMassive.json
grantseltzer: There's 255 elements that are 2 bytes each. The PR you linked shows the size limit on dynamic types is 512 bytes. So I'm not sure where the limit gets hit?

See my reply on your other comment.

pkg/dyninst/testdata/decoded/sample/testMapLargeKeyLargeValue.json
grantseltzer: This is no longer redacted after rebasing.

pkg/dyninst/testdata/decoded/sample/testMapWithSmallKeyAndValueMassive.json
grantseltzer: This is no longer redacted after rebasing.

pkg/dyninst/decode/types.go
grantseltzer: I prefer the verbosity tbh

pkg/dyninst/testdata/decoded/sample/testMapLargeKeyLargeValue.json
ajwerner: Oh, interesting. In Side-Eye we had some logic to always get at least one element. We can make that change as a follow-up. Thanks!

ajwerner (approved):
LGTM

pkg/dyninst/testdata/decoded/sample/testMapLargeKeyLargeValue.json
grantseltzer: I'll make a JIRA ticket

grantseltzer: https://datadoghq.atlassian.net/browse/DEBUG-4267

PR #39429, opened 2025-07-31
dyninst/symdb: translate package paths with dots properly

andreimatei:
When parsing function and type names, we were not dealing properly with packages
whose last element of the path contains dots; we were not decoding their
names properly. If the last element of the package's import path contains dots,
they are replaced with %2e in DWARF in order to differentiate them from
the dot that separates the package path from the function name. For
example, functions in package
"gopkg.in/square/go-jose.v2" will appear in DWARF as
"gopkg.in/square/go-jose%2ev2.newBuffer".

See https://groups.google.com/g/golang-nuts/c/Can9WXHrqHg/m/kMfx1x6sBgAJ

### DIFF
--- a/pkg/dyninst/symdb/func_name.go
+++ b/pkg/dyninst/symdb/func_name.go
@@ -160,6 +160,14 @@ func parseFuncName(qualifiedName string) (parseFuncNameResult, error) {
 		name = groups[standaloneFuncRENameIdx]
 	}
 
+	// If the last element of the package's import path contains dots, they are
+	// replaced with %2e in DWARF to differentiate them from the dot
+	// that separates the package path from the function name.
+	// For example, functions in package "gopkg.in/square/go-jose.v2" will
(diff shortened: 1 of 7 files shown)


ajwerner:
> I think this only applies to pre-Go-modules code, since after Go modules,
the last element of the import path is not allowed to contain dots.

where’s this coming from?

andreimatei:
> where’s this coming from?

Well, let me answer with a question -- I don't understand how `gopkg.in` works. How can you `import "gopkg.in/go-jose/go-jose.v2"` and then use the functions in it as `jose.JSONWebKeySet` (like `go-jose.v2` is used to find a directory on disk, but then the actual package name from the files inside it is used in the source code)?
Package names (as opposed to their path) cannot have dots in them, so I don't understand when you can end up with dots in the last element of the package name in dwarf.

andreimatei:
Hmm indeed, I see that directory names to not need to match the name of the package used inside them.

piob-io (approved):

PR #39434, opened 2025-07-31
pkg/dyninst: Fix bug in how int16 and int32s are cast

grantseltzer:
### What does this PR do?

Preserves signed bit of int16 and int32
Updates testdata

### Motivation

Found the bug

Existing tests, but now updated to have correct data.


ajwerner (approved):
LGTM

PR #39439, opened 2025-08-01
Refactor static quality gates

JSGette:
### What does this PR do?
This is a global rehaul of Static Quality Gates. Previous implementation suffered from various problems that I am trying to address.
- Suppressed error messages. In case of failing SQG it was not clear why it fails, especially when it comes to infrastructure issues or logical erros and bugs.
- Poor visibility for the users. Even though the information that is published in PR as a comment is pretty clear, it was not clear dealing with the pipeline log. This PR attempts to improve SQG visibility in the pipeline.
- Duplicated code everywhere. No comments, we are bringing everything closer together. For instance, gate metrics handling is exactly the same, thus, there is no need to copy-paste the same code block three times (docker, packages and msi).
- Use of kwargs and reflection to trigger gates. Now all attributes are clearly defined with their types, `entrypoint` is replaced by generic `execute_gate` function that is defined once in the superclass `StaticQualityGate`

### Motivation
We want to improve stability and UX while working with Static Quality Gates. This is the first step on that way.

New unit tests introduced to test basic functionality. Pipeline was used multiple times, the reporting was also tested when one or multiple SQGs are failing.

- New code structure relies more on inheritance and subclasses which is not always very straightforward
- Debug functionality has not been replaced, so it is missing entirely


tasks/static_quality_gates/lib/static_quality_gates_reporter.py
```diff
@@ -0,0 +1,308 @@
...
+        Convert gate name to a user-friendly display name
+
+        Args:
+            gate_name: The technical gate name (e.g., "static_quality_gate_agent_deb_amd64")
+
+        Returns:
+            Human-readable display name (e.g., "Agent DEB (AMD64)")
+        """
+        # Remove the "static_quality_gate_" prefix
+        name = gate_name.replace("static_quality_gate_", "")
```
rdesgroppes, line 29:
```suggestion
        name = gate_name.removeprefix("static_quality_gate_")
```
(which also makes above comment redundant)

tasks/quality_gates.py
```diff
@@ -152,60 +158,69 @@ def _print_quality_gates_report(gate_states: list[dict[str, typing.Any]]):
...
+            metric_handler.register_metric(gate.gate_name, "current_on_wire_size", gate.artifact_on_wire_size)
+            metric_handler.register_metric(gate.gate_name, "current_on_disk_size", gate.artifact_on_disk_size)
 
-    _print_quality_gates_report(gate_states)
+    ctx.run(f"datadog-ci tag --level job --tags static_quality_gates:\"{final_state}\"")
 
+    # Reporting part
+    # Send metrics to Datadog
+    # and then print the summary table
+    # in the job's log
```
rdesgroppes, line 217:
```suggestion
```

rdesgroppes (dismissed):
👍 👏 🔥

release.json
```diff
@@ -18,4 +18,4 @@
         "6": "6.53.1",
         "7": "7.68.2"
     }
-}
```
alopezz: Watch out for this kind of thing (if anything to avoid requiring unnecessary review requests).

tasks/quality_gates.py
```diff
@@ -152,60 +158,69 @@ def _print_quality_gates_report(gate_states: list[dict[str, typing.Any]]):
...
         except AssertionError as e:
-            print(f"Gate {gate} failed ! (AssertionError)")
             final_state = "failure"
-            gate_states.append({"name": gate, "state": False, "error_type": "AssertionError", "message": str(e)})
+            gate_states.append(
+                {"name": gate.gate_name, "state": False, "error_type": "AssertionError", "message": str(e)}
+            )
         except InfraError as e:
-            print(f"Gate {gate} flaked ! (InfraError)\n Restarting the job...")
+            print(f"Gate {gate.gate_name} flaked ! (InfraError)\n Restarting the job...")
```
alopezz, line 188: We could take this opportunity to also fix the punctuation in these messages 😅 though definitely not required.
```suggestion
            print(f"Gate {gate.gate_name} flaked! (InfraError)\n Restarting the job...")
```

tasks/quality_gates.py
```diff
@@ -152,60 +158,69 @@ def _print_quality_gates_report(gate_states: list[dict[str, typing.Any]]):
...
             traceback.print_exception(e)
             ctx.run("datadog-ci tag --level job --tags static_quality_gates:\"restart\"")
             raise Exit(code=42) from e
         except Exception:
```
alopezz, line 192: Maybe this should be `except StaticQualityGateFailed`? Since we now have a specific way to signal gate failure (in the sense of the size actually surpassing the limit), it would make sense to separate that from other exceptions (in fact, it may make sense to let it crash on unhandled exceptions because any other exception should be a bug).

tasks/quality_gates.py
```diff
@@ -152,60 +158,69 @@ def _print_quality_gates_report(gate_states: list[dict[str, typing.Any]]):
...
-        gate_inputs["metricHandler"] = metric_handler
-        gate_inputs["nightly"] = nightly_run
         try:
-            gate_mod = getattr(quality_gates_mod, gate)
-            gate_mod.entrypoint(**gate_inputs)
-            print(f"Gate {gate} succeeded !")
-            gate_states.append({"name": gate, "state": True, "error_type": None, "message": None})
+            gate.execute_gate(ctx)
+            gate_states.append({"name": gate.gate_name, "state": True, "error_type": None, "message": None})
         except AssertionError as e:
```
alopezz, line 182: It might make sense to remove this exception, too. I don't even know why we were using `AssertionError` in the first place.

tasks/static_quality_gates/lib/gates_lib.py
alopezz, line 1: Can be a follow up, but we could rename this to just `gates.py` or something like that, the extra `_lib` is redundant.

tasks/static_quality_gates/lib/gates_lib.py
```diff
@@ -262,3 +273,350 @@ def generate_metric_reports(self, ctx, filename="static_gate_report.json", branc
...
+
+class StaticQualityGate:
+    """
+    Base class for all static quality gates
+    that contains the common attributes for all static quality gates
+    """
+
+    gate_name: str
+    arch: str
+    os: str
```
alopezz, line 296: Since this is going to be computed by subclasses, how about turning into a [(cached) property](https://docs.python.org/3/library/functools.html#functools.cached_property)? The base class could present it as an abstract method / property. Same for `artifact_path`.

tasks/static_quality_gates/lib/gates_lib.py
```diff
@@ -262,3 +273,350 @@ def generate_metric_reports(self, ctx, filename="static_gate_report.json", branc
...
+    max_on_wire_size: int
+    max_on_disk_size: int
+    artifact_on_disk_size: int
+    artifact_on_wire_size: int
```
alopezz, line 300: We discussed this, I think this is not good when they are not initialized, and therefore these types annotations don't really hold.

Note also that these are really only type annotations and that they're essentially no-ops for the code itself.

tasks/static_quality_gates/lib/gates_lib.py
```diff
@@ -262,3 +273,350 @@ def generate_metric_reports(self, ctx, filename="static_gate_report.json", branc
...
+        elif "msi" in self.gate_name:
+            # MSI packages are always amd64 (x86_64) on Windows
+            self.arch = "amd64"
+        else:
+            raise ValueError(f"Unknown arch for gate: {self.gate_name}")
+
+    def print_results(self) -> None:
+        """
+        Print the results of the gate
+        in case of success
```
alopezz, line 325:
```suggestion
        Print the results of the gate in case of success
```

tasks/static_quality_gates/lib/gates_lib.py
```diff
@@ -262,3 +273,350 @@ def generate_metric_reports(self, ctx, filename="static_gate_report.json", branc
...
+            self.artifact_on_disk_size,
+            self.max_on_disk_size,
+        )
+
+    @abstractmethod
+    def _measure_on_disk_and_on_wire_size(self, ctx: Context) -> None:
+        """
+        Measure the size of the artifact on disk and on wire
+        """
+        raise NotImplementedError("This method should be implemented by the subclass")
```
alopezz, line 343: There's a problem with this in that it's not clear what this method is supposed to do. We can infer from the the code that it's expected to populate some internal state, but that's not documented here and feels brittle regardless.

I think this should return the values rather than modify state. That's also somewhat related to why I preferred composition over inheritance here.

tasks/static_quality_gates/lib/gates_lib.py
```diff
@@ -262,3 +273,350 @@ def generate_metric_reports(self, ctx, filename="static_gate_report.json", branc
...
+                self.os = PACKAGE_OS_MAPPING[package_type]
+                if self.os in ['centos', 'suse']:
+                    self.arch = ARCH_MAPPING[self.arch]
+                elif self.os == 'windows':
+                    # MSI packages use x86_64 naming convention
+                    self.arch = ARCH_MAPPING[self.arch]
+                return
+        raise ValueError(f"Unknown os for gate: {self.gate_name}")
+
+    def __init__(self, gate_name: str, gate_max_size_values: dict):
```
alopezz, line 398: `__init__` should always go on top of other "regular" methods.

tasks/static_quality_gates/lib/gates_lib.py
```diff
@@ -262,3 +273,350 @@ def generate_metric_reports(self, ctx, filename="static_gate_report.json", branc
...
+            error_message += f"On disk size (uncompressed artifact size) {self.artifact_on_disk_size / 1024 / 1024} MB is higher than the maximum allowed {self.max_on_disk_size / 1024 / 1024} MB by the gate !\n"
+        if error_message:
+            error_message = color_message(f"{self.gate_name} failed!\n" + error_message, "red")
+            raise StaticQualityGateFailed(error_message)
+
+    def execute_gate(self, ctx: Context):
+        """
+        Execute the quality gate.
+        """
+        from .static_quality_gates_reporter import QualityGateOutputFormatter
```
alopezz, line 364: I would strongly prefer to have the gate return some sort of "GateResults" object (could even be a dictionary) and that the caller would then take care of the printing.

PR #39441, opened 2025-08-01
Fix reinstall of datadog prefixed dependencies during agent upgrade

NouemanKHAL:
### What does this PR do?

Adds an exclude list `DEPS_STARTING_WITH_DATADOG` so that non-integration `datadog-*` packages aren’t mistakenly reinstalled as integrations during an Agent upgrade.

### Motivation

On Agent upgrade we previously treated every `datadog-*` dependency as an integration, accidentally reinstalling non-integration libraries (e.g. datadog-api-client) as third-party integrations. 
This change adds an exclude list of known `datadog-*` dependency that are not integrations, in a attempt to ensure only true integrations are reinstalled as third-party integrations.

Updated the E2E tests under `tests/agent-platform/persisting-integrations`.

The E2E test might fail in the future if the `datadog-api-client` dependency is installed by default from one of the core integrations.

### Additional Notes

I’d appreciate your feedback on the `DEPS_STARTING_WITH_DATADOG` list. To generate it, I:
	1.	Queried the PyPI Simple API for all packages, then filtered by name `datadog-*` using `jq` on the JSON output.
	2.	Converted each top-level integrations-core directory into a `datadog-<name>` entry, replacing underscores (_) with hyphens (-).
	3.	Ran comm -23 `pypi-packages.txt core-packages.txt` to identify the datadog prefixed packages that are not integrations.
	4.	Manually removed `datadog-kasperky` package, as it’s no longer a core integration.


NouemanKHAL:

releasenotes/notes/Fix-reinstall-of-dependencies-on-agent-upgrade-7caeb0c3c790e4e5.yaml
```diff
@@ -0,0 +1,11 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+fixes:
+  - |
+    Fixed an issue on Agent upgrade where all datadog prefixed packages were treated as integrations, preventing non-integration libraries from being properly reinstalled.
```
drichards-87:
```suggestion
    Fixed an issue during Agent upgrade where all Datadog-prefixed packages were treated as integrations, preventing non-integration libraries from being reinstalled.
```

drichards-87 (approved):
Left a suggestion from Docs on the release note and approved the PR.

releasenotes/notes/Fix-reinstall-of-dependencies-on-agent-upgrade-7caeb0c3c790e4e5.yaml
NouemanKHAL: The packages have the prefix`datadog-` like `datadog-api-client`

NouemanKHAL:

steveny91 (approved):

CelianR (approved):

swang392 (approved):

clarkb7 (approved):

omnibus/python-scripts/packages.py
```diff
@@ -186,7 +222,9 @@ def install_diff_packages_file(install_directory, filename, exclude_filename):
         if install_package_name in exclude_packages:
             print(f"Skipping '{install_package_name}' as it's already included in '{exclude_filename}' file")
         else:
-            if install_package_line.startswith('datadog-'):
+            # install_package_line can be a <requirement specifier> which contains a package name and a version constraint
+            # we need to check that install_package_line doesn't contain any dependency name in DEPS_STARTING_WITH_DATADOG
+            if install_package_line.startswith('datadog-') and not any(install_package_line.startswith(dep) for dep in DEPS_STARTING_WITH_DATADOG):
```
clarkb7: do we need to be more precise than `startswith` ? e.g. `datadog-log` will match any name that `datadog-logger` matches., and `datadog-sma` could match a future (for example) integration named `datadog-smart-devices`.

clarkb7 (changes requested):
blocking as follow up to my original comment.

Will `name.startswith("datadog-cli")` block installation/upgrade of [clickhouse](https://github.com/DataDog/integrations-core/tree/master/clickhouse) ?

omnibus/python-scripts/packages.py
clarkb7: other example, `datadog-log` prefix matching [logstash](https://github.com/DataDog/integrations-extras/tree/master/logstash)

NouemanKHAL: Oh, nice catch! Yes we definitely need to! We need to properly extract the dependency name from the install_package_line and then check if it exists in the list.
Thanks @clarkb7 !

NouemanKHAL:

omnibus/python-scripts/packages.py
NouemanKHAL: Fixed!

clarkb7: thanks! can you add a unit test, too?

NouemanKHAL: I wanted to but we have no mocks in place in the post_tests.py, I was considering adding an e2e test case with such libraries, or maybe extract this filtering logic as a method and a unit test for it  post_tests.py.
I'll see what I can do.

NouemanKHAL: Done!

omnibus/python-scripts/post_tests.py
```diff
@@ -67,3 +69,90 @@ def test_post_with_empty_files(self):
         # running rmdir verifies that the directory is empty
         os.rmdir(install_directory)
         os.rmdir(storage_location)
+
+    @unittest.skipIf(os.name == 'nt', "Skip this test on Windows environments")
```
clarkb7, line 73: why skip on windows?

clarkb7 (approved):

omnibus/python-scripts/post_tests.py
clarkb7: are these tests run anywhere else? originally i thought they were only run by the Windows unit test CI, not sure if that's changed,

omnibus/python-scripts/packages.py
clarkb7: thanks!

omnibus/python-scripts/post_tests.py
NouemanKHAL: I just added these tests, and my previous understanding was that the windows persistance logic leaves on the installer, and these test are only checking for the linux way of running the agent CLI commands and pip. If you're more familiar with the windows approach, we could update the assertions to match what we do on the windows side.

PR #39446, opened 2025-08-01
[CWS] Cleanup code related to rule disarmers

YoannGh:
### What does this PR do?

Follow up of #39370.

This PR is only refactoring code and should have no functional impact*.

It mostly simplifies existing code and introduces new methods on rule disarmers to allow using the `Lock`/ `defer Unlock` pattern with the goal of avoiding deadlock in the future.

*: it also fixes a small bug where the `disarmer_type` field of kill action reports would always be `container` once a rule gets dismantled even if the rule was dismantled because of executables.

### DIFF
--- a/pkg/security/probe/process_killer.go
+++ b/pkg/security/probe/process_killer.go
@@ -485,33 +485,10 @@ func (p *ProcessKiller) SendStats(statsd statsd.ClientInterface) {
 	p.perRuleStatsLock.Unlock()
 
 	p.ruleDisarmersLock.Lock()
+	defer p.ruleDisarmersLock.Unlock()
 	for ruleID, disarmer := range p.ruleDisarmers {
-		ruleIDTag := []string{
-			"rule_id:" + string(ruleID),
-		}
-
-		disarmer.m.Lock()
-		for disarmerType, count := range disarmer.disarmedCount {
-			if count > 0 {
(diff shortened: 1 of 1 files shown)


pkg/security/probe/process_killer.go
```diff
@@ -261,6 +261,20 @@ func (p *ProcessKiller) isRuleAllowed(rule *rules.Rule) bool {
...
+	defer p.ruleDisarmersLock.Unlock()
+
+	var disarmer *ruleDisarmer
+	if disarmer = p.ruleDisarmers[ruleID]; disarmer == nil {
+		containerParams, executableParams := p.getDisarmerParams(kill)
+		disarmer = newRuleDisarmer(ruleID, containerParams, executableParams)
+		p.ruleDisarmers[ruleID] = disarmer
+	}
+
+	return disarmer
```
paulcacheux, line 275: Is this a good idea ? Won't the disarmer escape the locked section ?

YoannGh: It will escape the locked section of the `ruleDisarmers` map which is fine. Then the user of the disarmer returned by this method is expected to take care of locking the disarmer itself

spikat (approved):
Honestly it didn't look simpler to me, it's even longer by 45 lines. But yeah I don't see any functional impact so that's an approval

PR #39449, opened 2025-08-01
chore: rewire the cloud run jobs check to move it out of demultiplexer_serverless.go

apiarian-datadog:
### What does this PR do?
Moves a cloud run env var check out of `demultiplexer_serverless.go` and into the serverless-init cmd and serverless pkg.

This brings some of our architectural issues to the forefront, but I think that is better than hiding a surprising cloud-run-jobs-specific check deep in code that isn't in the serverless directories.


nhulston:
The `os` import in `demultiplexer_serverless.go` is not used and can be removed

nhulston (approved):
After removing the `os` import, I tested manually and shutdown metrics still get delivered. Nice work

vickenty (approved):

PR #39452, opened 2025-08-01
dynisnt: run some tests in merge queue to prevent skew

ajwerner:
### What does this PR do?

Run a small set of dyninst tests in the merge queue.

### Motivation

A bunch of the dyninst tests work by executing probes against test binaries. When the set of probes or the test binaries change, we get merge skew. There's some risks that the output of probes changing could conflict, but they are lower than the changes to what we test. This small set of tests should prevent most of the issues.

Fixes https://datadoghq.atlassian.net/browse/DEBUG-4265

### DIFF
--- a/pkg/dyninst/integration_test.go
+++ b/pkg/dyninst/integration_test.go
@@ -367,11 +367,17 @@ func runIntegrationTestSuite(
 			outputs.byTest[t.Name()] = actual
 		}
 		t.Run(fmt.Sprintf("debug=%t", debug), func(t *testing.T) {
+			if debug && testing.Short() {
+				t.Skip("skipping debug with short")
+			}
 			t.Parallel()
 			t.Run("all-probes", func(t *testing.T) { runTest(t, probes) })
 			for i := range probes {
 				probeID := probes[i].GetID()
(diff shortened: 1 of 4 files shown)


ajwerner:
This does make `test_ebpf_*` a bit slower (about 45s), but they are far from the long tail on the merge queue, so it should be okay.
Screenshot from 2025-08-01 12-50-27

pkg/dyninst/integration_test.go
```diff
@@ -367,11 +367,17 @@ func runIntegrationTestSuite(
 			outputs.byTest[t.Name()] = actual
 		}
 		t.Run(fmt.Sprintf("debug=%t", debug), func(t *testing.T) {
+			if debug && testing.Short() {
+				t.Skip("skipping debug in with short")
```
piob-io: typo

piob-io (approved):
Thanks!

pkg/dyninst/integration_test.go
ajwerner: Done. Thanks

ofek (approved):

gjulianm (approved):

PR #39453, opened 2025-08-01
[FA] Add copy and remove_all instructions

coignetp:
### What does this PR do?

Copy stable folder in experiment before starting the config update, and add a `remove_all` instruction
When receiving an old config format, will do the `remove_all` first

Followup of https://github.com/DataDog/datadog-agent/pull/39428

### DIFF
--- a/pkg/fleet/installer/installer.go
+++ b/pkg/fleet/installer/installer.go
@@ -517,7 +517,16 @@ func (i *installerImpl) InstallConfigExperiment(
 	}
 	defer os.RemoveAll(tmpDir)
 
-	// TODO: starts with copying the stable config files
+	// Copy the files from the stable config
+	configRepo := i.configs.Get(pkg)
+	err = configRepo.CopyStable(ctx, tmpDir)
+	if err != nil {
+		return installerErrors.Wrap(
+			installerErrors.ErrFilesystemIssue,
+			fmt.Errorf("could not copy stable config: %w", err),
+		)
+	}
+
 	err = i.writeConfig(tmpDir, rawConfigs)
 	if err != nil {
 		return installerErrors.Wrap(
@@ -526,7 +535,6 @@ func (i *installerImpl) InstallConfigExperiment(
 		)
 	}
 
-	configRepo := i.configs.Get(pkg)
 	err = configRepo.SetExperiment(ctx, version, tmpDir)
 	if err != nil {
 		return installerErrors.Wrap(
@@ -833,9 +841,10 @@ func cleanConfigName(p string) string {
 type configFileAction string
 
 const (
-	configFileActionUnknown configFileAction = ""
-	configFileActionWrite   configFileAction = "write"
-	configFileActionRemove  configFileAction = "remove"
+	configFileActionUnknown   configFileAction = ""
+	configFileActionWrite     configFileAction = "write"
+	configFileActionRemove    configFileAction = "remove"
+	configFileActionRemoveAll configFileAction = "remove_all"
 )
 
 type configFile struct {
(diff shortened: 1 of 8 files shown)


pkg/fleet/installer/installer.go
```diff
@@ -848,18 +857,31 @@ func (i *installerImpl) writeConfig(dir string, rawConfigActions [][]byte) error
...
+					return fmt.Errorf("could not remove config file/directory %s: %w", entryPath, err)
+				}
+			}
+		case configFileActionRemove:
+			for _, file := range configAction.Files {
+				file.Path = cleanConfigName(file.Path)
+
+				if !configNameAllowed(file.Path) {
+					return fmt.Errorf("config file %s is not allowed", file)
+				}
```
BaptisteFoy, line 883: nice

pkg/fleet/installer/repository/repository.go
```diff
@@ -548,6 +571,33 @@ func buildFileMap(rootPath string) (map[string]struct{}, error) {
...
+		}
+
+		targetFilePath := filepath.Join(targetPath, relPath)
+		if info.IsDir() {
+			return os.MkdirAll(targetFilePath, info.Mode())
+		}
+
+		return copyFile(path, targetFilePath)
+	})
+}
```
BaptisteFoy: From what I can see this doesn't preserve file ownership (dd-agent); so unless we re-chown afterwards this is going to be an issue

pkg/fleet/installer/installer.go
clarkb7, line 883: should we move this above the `select` so it applies to all actions?

pkg/fleet/installer/repository/repository_win.go
```diff
@@ -0,0 +1,15 @@
...
+//go:build windows
+
+package repository
+
+import "os"
+
+// copyFileWithPermissions copies a file from src to dst with the same permissions.
+func copyFileWithPermissions(src, dst string, _ os.FileInfo) error {
+	return copyFile(src, dst)
+}
```
clarkb7, line 15: windows permissions are weird when you copy/move files. let's discuss on slack.

pkg/fleet/installer/installer.go
```diff
@@ -517,7 +517,16 @@ func (i *installerImpl) InstallConfigExperiment(
...
-	// TODO: starts with copying the stable config files
+	// Copy the files from the stable config
+	configRepo := i.configs.Get(pkg)
+	err = configRepo.CopyStable(ctx, tmpDir)
+	if err != nil {
+		return installerErrors.Wrap(
+			installerErrors.ErrFilesystemIssue,
+			fmt.Errorf("could not copy stable config: %w", err),
+		)
+	}
```
clarkb7, line 528: why do we need to do this?

BaptisteFoy (approved):

pkg/fleet/installer/installer.go
coignetp: This should not apply to `remove_all`

pkg/fleet/installer/installer.go
coignetp: Answered on slack

pkg/fleet/installer/repository/repository_win.go
clarkb7: discussed offline. We're copying from the stable config dir to the experiment config dir. Both use inherited permissions, so we don't need to change anything here.

clarkb7 (approved):

pkg/fleet/installer/installer.go
clarkb7: remove_all doesn't use the user input path names, it uses a temp dir.

PR #39454, opened 2025-08-01
Use evaluation errors for surfacing issues with decoding

grantseltzer:
### What does this PR do?

- Use evaluation errors for surfacing issues with decoding
    - Errors that occur with either parsing data from events, or symbolicating program counters for stack traces
- Decode() now returns a byte slice instead of taking an io.Writer to allow for flexibility in recovering from errors and returning partial snapshots.

### Motivation

Better error propagation in decoding.

Added fault injection to integration tests

### Additional Notes

[DEBUG-4427
](https://datadoghq.atlassian.net/browse/DEBUG-4427)

### DIFF
--- a/pkg/dyninst/decode/decoder.go
+++ b/pkg/dyninst/decode/decoder.go
@@ -8,23 +8,23 @@
 package decode
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"runtime/debug"
 	"time"
 
 	"github.com/go-json-experiment/json"
+	"github.com/go-json-experiment/json/jsontext"
 	"github.com/google/uuid"
 	pkgerrors "github.com/pkg/errors"
-
 	"golang.org/x/time/rate"
-
-	"github.com/DataDog/datadog-agent/pkg/util/log"
 
 	"github.com/DataDog/datadog-agent/pkg/dyninst/gotype"
 	"github.com/DataDog/datadog-agent/pkg/dyninst/ir"
 	"github.com/DataDog/datadog-agent/pkg/dyninst/output"
 	"github.com/DataDog/datadog-agent/pkg/dyninst/symbol"
+	"github.com/DataDog/datadog-agent/pkg/util/log"
 )
 
 // We don't want to be too noisy about symbolication errors, but we do want to learn
@@ -127,12 +127,14 @@ type typeAndAddr struct {
 	addr   uint64
 }
 
-// Decode decodes the output Event from the BPF program into a JSON format to the specified output writer.
+// Decode decodes the output Event from the BPF program into a JSON format
+// the `output` parameter is appended to and returned as the final output.
+// It is not thread-safe.
 func (d *Decoder) Decode(
 	event Event,
 	symbolicator symbol.Symbolicator,
-	output []byte,
-) (b []byte, probe ir.ProbeDefinition, err error) {
+	buf []byte,
+) (_ []byte, probe ir.ProbeDefinition, err error) {
 	defer d.resetForNextMessage()
 	defer func() {
 		r := recover()
@@ -146,13 +148,27 @@ func (d *Decoder) Decode(
 	}()
 	probe, err = d.snapshotMessage.init(d, event, symbolicator)
 	if err != nil {
-		return nil, nil, err
+		return buf, nil, err
 	}
-	output, err = json.Marshal(&d.snapshotMessage)
-	if err != nil {
-		err = fmt.Errorf("error marshaling snapshot message: %w", err)
+	b := bytes.NewBuffer(buf)
+	enc := jsontext.NewEncoder(b)
+	numAttempts := 0
+	numExpressions := len(d.snapshotMessage.Debugger.Snapshot.Captures.Entry.Arguments.rootType.Expressions)
+	for numAttempts <= numExpressions {
+		// We loop here because when evaluation errors occur, we reduce the amount of data we attempt
+		// to encode and then try again after resetting the buffer.
+		err = json.MarshalEncode(enc, &d.snapshotMessage)
+		if errors.Is(err, errEvaluation) {
+			b = bytes.NewBuffer(buf)
+			enc.Reset(b)
+			numAttempts++
+			continue
+		} else if err != nil {
+			return buf, probe, pkgerrors.Wrap(err, "error marshaling snapshot message")
+		}
+		break
 	}
-	return output, probe, err
+	return b.Bytes(), probe, err
 }
 
 func (d *Decoder) resetForNextMessage() {
@@ -183,12 +199,14 @@ func (s *snapshotMessage) init(
 	symbolicator symbol.Symbolicator,
 ) (ir.ProbeDefinition, error) {
 	s.Service = event.ServiceName
-	s.Debugger.Snapshot = snapshotData{
-		decoder:  decoder,
-		ID:       uuid.New(),
-		Language: "go",
+	s.Debugger = debuggerData{
+		Snapshot: snapshotData{
+			decoder:  decoder,
+			ID:       uuid.New(),
+			Language: "go",
+		},
+		EvaluationErrors: []string{},
 	}
-	s.Debugger.EvaluationErrors = []string{}
 	var rootType *ir.EventRootType
 	var probe ir.ProbeDefinition
 
@@ -264,12 +282,14 @@ func (s *snapshotMessage) init(
 	s.Logger.Version = probe.GetVersion()
 	s.Debugger.Snapshot.Probe.ID = probe.GetID()
 	s.Debugger.Snapshot.Stack.frames = stackFrames
+
 	s.Debugger.Snapshot.Captures.Entry.Arguments = argumentsData{
 		event:            event,
 		rootType:         rootType,
 		rootData:         s.rootData,
 		decoder:          decoder,
 		evaluationErrors: &s.Debugger.EvaluationErrors,
+		skipIndicies:     make([]byte, len(rootType.Expressions)),
 	}
 	return probe, nil
 }
@@ -277,14 +297,14 @@ func (s *snapshotMessage) init(
 func symbolicate(event Event, stackHash uint64, symbolicator symbol.Symbolicator) ([]symbol.StackFrame, error) {
 	pcs, err := event.StackPCs()
 	if err != nil {
-		return nil, fmt.Errorf("error getting stack pcs %w", err)
+		return nil, fmt.Errorf("error getting stack pcs: %w", err)
 	}
 	if len(pcs) == 0 {
 		return nil, errors.New("no stack pcs found")
 	}
 	stackFrames, err := symbolicator.Symbolicate(pcs, stackHash)
 	if err != nil {
-		return nil, fmt.Errorf("error symbolicating stack %w", err)
+		return nil, fmt.Errorf("error symbolicating stack: %w", err)
 	}
 	return stackFrames, nil
 }
--- a/pkg/dyninst/decode/decoder.go
+++ b/pkg/dyninst/decode/decoder.go
@@ -153,16 +153,14 @@ func (d *Decoder) Decode(
 	}
 	b := bytes.NewBuffer(buf)
 	enc := jsontext.NewEncoder(b)
-	numAttempts := 0
 	numExpressions := len(d.snapshotMessage.Debugger.Snapshot.Captures.Entry.Arguments.rootType.Expressions)
-	for numAttempts <= numExpressions {
-		// We loop here because when evaluation errors occur, we reduce the amount of data we attempt
-		// to encode and then try again after resetting the buffer.
+	// We loop here because when evaluation errors occur, we reduce the amount of data we attempt
+	// to encode and then try again after resetting the buffer.
+	for range numExpressions + 1 { // +1 for the initial attempt
 		err = json.MarshalEncode(enc, &d.snapshotMessage)
 		if errors.Is(err, errEvaluation) {
 			b = bytes.NewBuffer(buf)
 			enc.Reset(b)
-			numAttempts++
 			continue
 		} else if err != nil {
 			return buf, probe, pkgerrors.Wrap(err, "error marshaling snapshot message")
--- a/pkg/dyninst/decode/decoder.go
+++ b/pkg/dyninst/decode/decoder.go
@@ -265,11 +265,11 @@ func (s *snapshotMessage) init(
 	s.Debugger.Snapshot.Probe.ID = probe.GetID()
 	s.Debugger.Snapshot.Stack.frames = stackFrames
 	s.Debugger.Snapshot.Captures.Entry.Arguments = argumentsData{
-		event:        event,
-		rootType:     rootType,
-		rootData:     s.rootData,
-		decoder:      decoder,
-		debuggerData: &s.Debugger,
+		event:            event,
+		rootType:         rootType,
+		rootData:         s.rootData,
+		decoder:          decoder,
+		evaluationErrors: &s.Debugger.EvaluationErrors,
 	}
 	return probe, nil
 }
--- a/pkg/dyninst/decode/decoder.go
+++ b/pkg/dyninst/decode/decoder.go
@@ -76,9 +76,10 @@ type Decoder struct {
 	typeNameResolver     TypeNameResolver
 
 	// These fields are initialized and reset for each message.
-	snapshotMessage   snapshotMessage
-	dataItems         map[typeAndAddr]output.DataItem
-	currentlyEncoding map[typeAndAddr]struct{}
+	snapshotMessage    snapshotMessage
+	dataItems          map[typeAndAddr]output.DataItem
+	currentlyEncoding  map[typeAndAddr]struct{}
+	skipIndiciesBuffer []byte
 }
 
 // NewDecoder creates a new Decoder for the given program.
@@ -93,10 +94,10 @@ func NewDecoder(
 		stackFrames:          make(map[uint64][]symbol.StackFrame),
 		typesByGoRuntimeType: make(map[uint32]ir.TypeID),
 		typeNameResolver:     typeNameResolver,
-
-		snapshotMessage:   snapshotMessage{},
-		dataItems:         make(map[typeAndAddr]output.DataItem),
-		currentlyEncoding: make(map[typeAndAddr]struct{}),
+		snapshotMessage:      snapshotMessage{},
+		dataItems:            make(map[typeAndAddr]output.DataItem),
+		currentlyEncoding:    make(map[typeAndAddr]struct{}),
+		skipIndiciesBuffer:   make([]byte, 1),
 	}
 	for _, probe := range program.Probes {
 		for _, event := range probe.Events {
@@ -289,7 +290,7 @@ func (s *snapshotMessage) init(
 		rootData:         s.rootData,
 		decoder:          decoder,
 		evaluationErrors: &s.Debugger.EvaluationErrors,
-		skipIndicies:     make([]byte, len(rootType.Expressions)),
+		skipIndicies:     decoder.getSkipIndiciesBuffer(len(rootType.Expressions)),
 	}
 	return probe, nil
 }
@@ -307,4 +308,20 @@ func symbolicate(event Event, stackHash uint64, symbolicator symbol.Symbolicator
 		return nil, fmt.Errorf("error symbolicating stack: %w", err)
 	}
 	return stackFrames, nil
+}
+
+// getSkipIndiciesBuffer returns a zeroed byte slice of the required size,
+// reusing the internal buffer when possible to avoid allocations.
+func (d *Decoder) getSkipIndiciesBuffer(numExpressions int) []byte {
+	requiredBytes := (numExpressions + 7) / 8
+	if cap(d.skipIndiciesBuffer) < requiredBytes {
+		d.skipIndiciesBuffer = make([]byte, requiredBytes)
+	} else {
+		// Reuse existing buffer, just resize and clear
+		d.skipIndiciesBuffer = d.skipIndiciesBuffer[:requiredBytes]
+		for i := range d.skipIndiciesBuffer {
+			d.skipIndiciesBuffer[i] = 0
+		}
+	}
+	return d.skipIndiciesBuffer
 }
--- a/pkg/dyninst/decode/decoder.go
+++ b/pkg/dyninst/decode/decoder.go
@@ -10,7 +10,6 @@ package decode
 import (
 	"errors"
 	"fmt"
-	"io"
 	"runtime/debug"
 	"time"
 
@@ -18,11 +17,19 @@ import (
 	"github.com/google/uuid"
 	pkgerrors "github.com/pkg/errors"
 
+	"golang.org/x/time/rate"
+
+	"github.com/DataDog/datadog-agent/pkg/util/log"
+
 	"github.com/DataDog/datadog-agent/pkg/dyninst/gotype"
 	"github.com/DataDog/datadog-agent/pkg/dyninst/ir"
 	"github.com/DataDog/datadog-agent/pkg/dyninst/output"
 	"github.com/DataDog/datadog-agent/pkg/dyninst/symbol"
 )
+
+// We don't want to be too noisy about symbolication errors, but we do want to learn
+// about them and we don't want to bail out completely.
+var symbolicateErrorLogLimiter = rate.NewLimiter(rate.Every(1*time.Minute), 10)
 
 type probeEvent struct {
 	event *ir.Event
@@ -124,8 +131,8 @@ type typeAndAddr struct {
 func (d *Decoder) Decode(
 	event Event,
 	symbolicator symbol.Symbolicator,
-	out io.Writer,
-) (probe ir.ProbeDefinition, err error) {
+	output []byte,
+) (b []byte, probe ir.ProbeDefinition, err error) {
 	defer d.resetForNextMessage()
 	defer func() {
 		r := recover()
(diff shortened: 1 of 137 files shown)


ajwerner (commented):
I'm not so sure about this. What errors in type construction are anything other than bugs and assertion failures?

grantseltzer:
> I'm not so sure about this. What errors in type construction are anything other than bugs and assertion failures?

It'd be any possible issue with irgen not populating `Types` correctly, so yes it'd likely be for bugs which is why there's errors in newDecoderType. 

If there's an issue with just one type, why should it prevent us from getting events altogether? The type that's not captured would ideally surface with a notCapturedReason.

ajwerner:
I'm of two minds about the thing. In general making things fallible in the data plane tends to imply that there's an expected failure mode. What I'd worry about is that we'll never learn about these situations as a team and we'll just have bugs sitting around forever and users having bad times whereas if we send failure diagnostics were much more likely to address what is certainly a bug on our side.

pkg/dyninst/decode/marshal.go
```diff
@@ -66,7 +66,8 @@ type captureData struct {
 }
 
 type capturePointData struct {
-	Arguments argumentsData `json:"arguments"`
+	Arguments     jsontext.Value `json:"arguments"`
+	ArgumentsData argumentsData  `json:"-"`
```
ajwerner: why export it?

grantseltzer:
> I'm of two minds about the thing. In general making things fallible in the data plane tends to imply that there's an expected failure mode. What I'd worry about is that we'll never learn about these situations as a team and we'll just have bugs sitting around forever and users having bad times whereas if we send failure diagnostics were much more likely to address what is certainly a bug on our side.

In my mind this gives us more avenues to surface errors. We can set a diagnostic on the probe as well as have not captured reasons set.

ajwerner:
>  In my mind this gives us more avenues to surface errors. We can set a diagnostic on the probe as well as have not captured reasons set.

That's certainly not what this change is doing. It's creating an avenue for a code invariant violation to be hidden (if the type isn't needed, the code still shouldn't fail here and we'll just now never know). Also you haven't done the other plumbing for this error to be presented. If you want to implement an end-to-end flow that improves the user experience, I'm supportive. One thing I note is you also didn't fix the decoder constructor for the rcscrape package. 

On the whole my feedback on this change is that it's not worth more bickering from me, so if you insist on this lazy construction, fine -- follow through on making it work fully.

I especially see no reason to make this change prior to changes that do something good with errors that occur during the decode process. I'd put this on hold until after such a change that could utilize it.

Furthermore, it's hard to test this in a realistic scenario because we never expect these errors to happen.

pkg/dyninst/decode/decoder.go
```diff
@@ -91,11 +85,29 @@ func (d *Decoder) Decode(
...
+		}
+	}
 	probe, err = d.snapshotMessage.init(d, event, symbolicator)
 	if err != nil {
 		return nil, err
 	}
 	defer d.snapshotMessage.clear()
+	b := []byte{}
+	buf := bytes.NewBuffer(b)
+	err = json.MarshalWrite(buf, &d.snapshotMessage.Debugger.Snapshot.Captures.Entry.ArgumentsData)
```
ajwerner: This extra set of allocations on the happy path is somewhat sad

grantseltzer:
> > In my mind this gives us more avenues to surface errors. We can set a diagnostic on the probe as well as have not captured reasons set.
> 
> That's certainly not what this change is doing. It's creating an avenue for a code invariant violation to be hidden (if the type isn't needed, the code still shouldn't fail here and we'll just now never know). Also you haven't done the other plumbing for this error to be presented. If you want to implement an end-to-end flow that improves the user experience, I'm supportive. One thing I note is you also didn't fix the decoder constructor for the rcscrape package.
> 
> On the whole my feedback on this change is that it's not worth more bickering from me, so if you insist on this lazy construction, fine -- follow through on making it work fully.
> 
> I especially see no reason to make this change prior to changes that do something good with errors that occur during the decode process. I'd put this on hold until after such a change that could utilize it.
> 
> Furthermore, it's hard to test this in a realistic scenario because we never expect these errors to happen.

I'm being intentional about keeping PRs small. I marked this change while making larger changes to surfacing of not captured reasons. It's a draft for a reason. I think you're being a bit unfair to be honest.

pkg/dyninst/decode/decoder.go
```diff
@@ -89,15 +82,27 @@ type typeAndAddr struct {
...
+		d.decoderTypes = make(map[ir.TypeID]decoderType, len(d.program.Types))
+		for _, t := range d.program.Types {
+			decoderType, err := newDecoderType(t, d.program.Types)
+			if err != nil {
+				log.Tracef("error getting decoder type for type %s: %w", t.GetName(), err)
+				continue
+			}
+			d.decoderTypes[t.GetID()] = decoderType
+		}
+	}
```
ajwerner: I want to push back on making this lazy. Imagine we have a bug where we're going to fail on this. Now every time we get an event we'll hit this code. It will never succeed. But even further, I think it's bad to have this work happening in the data plane path. Right now we construct the decoder before attachment which is a good thing. Also, it feels right to me that we fail with probe diagnostics and in a loud way if we can't construct a decoder.

I think having the opportunity to return an error from decoder construction was a good thing. It is akin to a failure to compile the probe. I see no reason to make the change, and I especially see no reason to couple it with this PR.

pkg/dyninst/decode/decoder.go
```diff
@@ -89,15 +82,27 @@ type typeAndAddr struct {
...
 	probe, err = d.snapshotMessage.init(d, event, symbolicator)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 	defer d.snapshotMessage.clear()
-	err = json.MarshalWrite(out, &d.snapshotMessage)
-	return probe, err
+
+	buf := bytes.NewBuffer(output)
```
ajwerner: this is the same thing as `bytes.NewBuffer(nil)` but is more confusing.

pkg/dyninst/decode/swiss_maps.go
```diff
@@ -0,0 +1,107 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
```
ajwerner: Pulling this out would have been a good thing to do in its own commit. It's hard to tell what's just code movement and what needs to be re-reviewed from scratch

pkg/dyninst/decode/decoder.go
```diff
@@ -186,17 +181,15 @@ func (s *snapshotMessage) init(
...
 	if !ok {
-		pcs, err = event.StackPCs()
-		if err != nil {
-			return probe, fmt.Errorf("error getting stack pcs %w", err)
-		}
-		stackFrames, err = symbolicator.Symbolicate(pcs, header.Stack_hash)
+		stackFrames, err = symbolicate(event, header.Stack_hash, symbolicator)
 		if err != nil {
-			return probe, fmt.Errorf("error symbolicating stack %w", err)
+			s.Debugger.EvaluationErrors = append(s.Debugger.EvaluationErrors,
```
ajwerner, line 263: this is the sort of thing we'll want to hear about. Can you add a rate limiter and log this error?

ajwerner, line 263: We need to store the fact that this failed in some form or another. We're not going to get the stack again so then we'll just keep getting the error about how we don't have the PCs. That's a bad experience.

pkg/dyninst/decode/decoder.go
ajwerner, line 1: You could have a test in the decode package that trivially covers the symbolication failure scenario methinks. That'd be better than this very-far-away testing.

pkg/dyninst/testdata/decoded/sample/faulty-stackC.json
ajwerner, line 1: It's best to do the rewriting in the commit that causes the changes to the output. It's important to see what changes in which commit as a way to express the intention of the change

pkg/dyninst/decode/decoder.go
grantseltzer: It does suck that once this fails it will always fail - but what do you have in mind? It will always be accurate. Perhaps probe status is appropriate?

grantseltzer: I can certainly log it, what do you mean by rate limiter?

grantseltzer: As per IRL convo, Event.StackPCs() can return an empty slice of pcs and no error which messes with this logic. I've added a check in symbolicate() to return an error if the pc slice is nil.

grantseltzer: As per irl convo, I added a rate limiter :^)

grantseltzer:
I've separated the faulty integration test into its own file, but it definitely can be cleaned up with more reuse. While I can do that if you encourage it, I will say that I like having it still run under `TestDyninst`.

grantseltzer:
I separated the testing further, lmk if this works!

pkg/dyninst/utilities_test.go
piob-io, line 1: We already have a file for test utilities - pkg/dyninst/dyninsttest/util.go let's put this there.

pkg/dyninst/integration_test.go
piob-io, line 1: Could the test structure changes happen in a separate pr?

pkg/dyninst/decode/decoder.go
```diff
@@ -128,7 +136,7 @@ func (s *snapshotMessage) init(
 		ID:       uuid.New(),
 		Language: "go",
 	}
-
+	s.Debugger.EvaluationErrors = []string{}
```
piob-io: Why not rely on nil slice semantically meaning an empty one?

pkg/dyninst/decode/decoder.go
```diff
@@ -206,15 +212,29 @@ func (s *snapshotMessage) init(
...
 
 func (s *snapshotMessage) clear() {
 	s.Debugger.Snapshot = snapshotData{}
+	clear(s.Debugger.EvaluationErrors)
+}
+
+func symbolicate(event Event, stackHash uint64, symbolicator symbol.Symbolicator) ([]symbol.StackFrame, error) {
+	pcs, err := event.StackPCs()
+	if err != nil || pcs == nil {
+		return nil, fmt.Errorf("error getting stack pcs %w", err)
```
piob-io: We need a separate error message if pcs is nil, since err will be nil then. How about we keep the error message from the failure to symbolicate the stack the first time, when we did have pcs? Then we can avoid this artifical error condition that stems from us not re-emitting pcs on repeated stacks.

pkg/dyninst/decode/marshal.go
```diff
@@ -66,14 +69,15 @@ type captureData struct {
...
 type argumentsData struct {
-	rootData []byte
-	rootType *ir.EventRootType
-	event    Event
-	decoder  *Decoder
+	rootData     []byte
+	rootType     *ir.EventRootType
+	event        Event
+	decoder      *Decoder
+	debuggerData *debuggerData
```
piob-io: Why do we need this cyclic reference?

piob-io: If this is only for evaluationErrors, could we only include evaluation errors (as a pointer to slice)? This way there will be no cycle, less cognitive burden.

pkg/dyninst/decode/marshal.go
```diff
@@ -86,12 +90,38 @@ func expressionIsPresent(bitset []byte, expressionIndex int) bool {
...
+}
 
+func (ad *argumentsData) MarshalText() ([]byte, error) {
+	var (
+		err error
+		buf = bytes.NewBuffer([]byte{})
+		enc = jsontext.NewEncoder(buf)
+	)
+	if err = writeTokens(enc, jsontext.BeginObject); err != nil {
+		ad.debuggerData.EvaluationErrors = append(ad.debuggerData.EvaluationErrors, err.Error())
```
piob-io: Since we return the error here anyway, could the caller take care of adding that error to evaluation errors? Otherwise we are splitting responsibilities on handling an error, triggering question what code is responsible for what part of dealing with errors...

piob-io:
> I've separated the faulty integration test into its own file, but it definitely can be cleaned up with more reuse. While I can do that if you encourage it, I will say that I like having it still run under `TestDyninst`.

What's the benefit of having an integration test over targeted unit tests for this?

ajwerner (changes requested):
I don't think I get it. The point of returning slice was so that upon an error, we could stop trying to marshal the expression that failed, and retry the whole thing again. What I expected this to mean is that we'd need a loop. Like the point is to have expression failures not result in an error being returned, and yet other than for symbolication, they still are. 

It's also at that point around each expression that I'd expect a defered recover to turn panics into evaluation errors. 

Let's make sure there are tests that demonstrate the goal of the PR.

One way to inject a panic is to mutate the type map of the decoder to shove a fake decoder type in that panics. You could have yet another than returns an error.

pkg/dyninst/decode/decoder.go
```diff
@@ -120,12 +129,12 @@ type typeAndAddr struct {
 	addr   uint64
 }
 
-// Decode decodes the output Event from the BPF program into a JSON format to the specified output writer.
+// Decode decodes the output Event from the BPF program into a JSON format
```
ajwerner, line 131: make it clear that it will append the data to output and return that

pkg/dyninst/decode/decoder.go
```diff
@@ -139,13 +148,24 @@ func (d *Decoder) Decode(
 	}()
 	probe, err = d.snapshotMessage.init(d, event, symbolicator)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
-	err = json.MarshalWrite(out, &d.snapshotMessage)
-	if err != nil {
-		err = fmt.Errorf("error marshaling snapshot message: %w", err)
+	output = output[:0]
```
ajwerner: It might be nice to have a comment near the loop explaining why it exists

pkg/dyninst/decode/decoder.go
ajwerner: yeah, I too do not understand this. What if instead, for clearity, we explicitly reset all of s.Debugger?

pkg/dyninst/decode/decoder.go
```diff
@@ -139,13 +148,24 @@ func (d *Decoder) Decode(
...
+	b := bytes.NewBuffer(output)
+	enc := jsontext.NewEncoder(b)
+	for {
```
ajwerner: it's dangerous to have a completely unbounded loop. Perhaps make the condition that if the number of iterations equals the number of expressions, we need to bail

pkg/dyninst/decode/decoder.go
```diff
@@ -139,13 +148,24 @@ func (d *Decoder) Decode(
...
+		err = json.MarshalEncode(enc, &d.snapshotMessage)
+		if errors.Is(err, errEvaluation) {
+			output = output[:0]
```
ajwerner: Generally when you pass in a buffer and return something else, the semantics are that you're appending to it. I don't think those are the semantics you've achieved here.

pkg/dyninst/decode/decoder.go
```diff
@@ -260,10 +278,27 @@ func (s *snapshotMessage) init(
...
+		evaluationErrors: &s.Debugger.EvaluationErrors,
+		skipIndicies:     make([]bool, len(rootType.Expressions)),
 	}
 	return probe, nil
 }
+
+func symbolicate(event Event, stackHash uint64, symbolicator symbol.Symbolicator) ([]symbol.StackFrame, error) {
+	pcs, err := event.StackPCs()
+	if err != nil {
+		return nil, fmt.Errorf("error getting stack pcs %w", err)
```
ajwerner: nit: usually we put a `: %w` when wrapping.

pkg/dyninst/decode/decoder.go
```diff
@@ -260,10 +278,27 @@ func (s *snapshotMessage) init(
...
+	}
+	if len(pcs) == 0 {
+		return nil, errors.New("no stack pcs found")
+	}
+	stackFrames, err := symbolicator.Symbolicate(pcs, stackHash)
+	if err != nil {
+		return nil, fmt.Errorf("error symbolicating stack %w", err)
```
ajwerner: nit: usually we put a `: %w` when wrapping.

pkg/dyninst/decode/marshal.go
```diff
@@ -66,10 +67,12 @@ type capturePointData struct {
...
 type argumentsData struct {
-	rootData []byte
-	rootType *ir.EventRootType
-	event    Event
-	decoder  *Decoder
+	rootData         []byte
+	rootType         *ir.EventRootType
+	event            Event
+	decoder          *Decoder
+	evaluationErrors *[]string
```
ajwerner, line 74: why `*[]string`, feels bizarre to me.

pkg/dyninst/decode/marshal.go
```diff
@@ -90,58 +93,95 @@ func expressionIsPresent(bitset []byte, expressionIndex int) bool {
...
-	if err = writeTokens(enc, jsontext.BeginObject); err != nil {
+	if err := writeTokens(enc, jsontext.BeginObject); err != nil {
 		return err
 	}
 
 	if ad.rootType.PresenceBitsetSize > uint32(len(ad.rootData)) {
-		return fmt.Errorf("presence bitset is out of bounds: %d > %d",
-			ad.rootType.PresenceBitsetSize, len(ad.rootData),
-		)
+		*ad.evaluationErrors = append(*ad.evaluationErrors, "presence bitset is out of bounds")
```
ajwerner: Is this case covered? It seems to me like as the code is written, this would infinite loop, right?

pkg/dyninst/decode/decoder.go
```diff
@@ -260,10 +278,27 @@ func (s *snapshotMessage) init(
...
-		event:    event,
-		rootType: rootType,
-		rootData: s.rootData,
-		decoder:  decoder,
+		event:            event,
+		rootType:         rootType,
+		rootData:         s.rootData,
+		decoder:          decoder,
+		evaluationErrors: &s.Debugger.EvaluationErrors,
+		skipIndicies:     make([]bool, len(rootType.Expressions)),
```
ajwerner: this feels like one of these things we should reuse if we can. Also, it could be a bitset rather than a bool slice

pkg/dyninst/testdata/decoded/sample/stackC.json
```diff
@@ -62,7 +62,8 @@
             "arguments": {}
           }
         }
-      }
+      },
+      "evaluationErrors": []
```
ajwerner: It seems almost certain to me that we should omit this if there were no evaluation errors

pkg/dyninst/decode/decoder.go
```diff
@@ -8,22 +8,31 @@
...
 
 	"github.com/go-json-experiment/json"
+	"github.com/go-json-experiment/json/jsontext"
 	"github.com/google/uuid"
 	pkgerrors "github.com/pkg/errors"
 
+	"golang.org/x/time/rate"
+
+	"github.com/DataDog/datadog-agent/pkg/util/log"
+
```
ajwerner: nit: stray newline

pkg/dyninst/decode/decoder.go
```diff
@@ -8,22 +8,31 @@
...
-	"io"
 	"runtime/debug"
 	"time"
 
 	"github.com/go-json-experiment/json"
+	"github.com/go-json-experiment/json/jsontext"
 	"github.com/google/uuid"
 	pkgerrors "github.com/pkg/errors"
 
+	"golang.org/x/time/rate"
```
ajwerner, line 21: nit: doesn't this belong mixed in with the other 3rdparty?

ajwerner (commented):
Getting close

pkg/dyninst/decode/marshal.go
grantseltzer: So that we can append to it without a pointer to `debuggerData`. I previously had that but Piotr asked me to be more precise with what i'm referencing.

pkg/dyninst/decode/decoder.go
grantseltzer: ~~I've changed it to be a bitset, I agree i'd like to reuse it but could use some advice on how to do so. Doing something simple like `x = x[:0]` won't work because it needs to be correctly populated with number of bits at least as long as the number of expressions.~~

~So we'd have to add a helper which checks required and existing length, grows it accordingly if needed, and and 0's all individual elements. Is this how you were thinking about it and is it worth the added instructions? I can work on a benchmark if you think it's worth investigating.~

Nevermind it's really not that much and we already have a benchmark

pkg/dyninst/decode/marshal.go
ajwerner: I don't get it... I don't think that's what he was asking for. I think he was asking for `nil` as opposed to `[]string{}` (which will not `== nil`)/

pkg/dyninst/decode/marshal.go
```diff
@@ -90,58 +93,107 @@ func expressionIsPresent(bitset []byte, expressionIndex int) bool {
...
-		)
-		if err != nil {
-			return fmt.Errorf(
-				"error parsing data for field %s: %w",
-				ad.rootType.Name, err,
-			)
+		if err := ad.processExpression(enc, expr, presenceBitSet, i); errors.Is(err, errEvaluation) {
+			// This expression resulted in an evaluation error, we mark it to be skipped
+			// and will try again
+			setSkipped(ad.skipIndicies, i)
```
ajwerner, line 190: given you already did `errors.Is` I think it's bizarre to return errEvalutation here as opposed to just returning err

pkg/dyninst/decode/decoder.go
```diff
@@ -139,13 +149,27 @@ func (d *Decoder) Decode(
...
+		return buf, nil, err
 	}
-	err = json.MarshalWrite(out, &d.snapshotMessage)
-	if err != nil {
-		err = fmt.Errorf("error marshaling snapshot message: %w", err)
+	b := bytes.NewBuffer(buf)
+	enc := jsontext.NewEncoder(b)
+	numAttempts := 0
+	numExpressions := len(d.snapshotMessage.Debugger.Snapshot.Captures.Entry.Arguments.rootType.Expressions)
+	for numAttempts <= numExpressions {
```
ajwerner: nit: `for range numExpressions` and lose `numAttempts`

pkg/dyninst/decode/decoder.go
```diff
@@ -86,10 +94,10 @@ func NewDecoder(
...
 		typesByGoRuntimeType: make(map[uint32]ir.TypeID),
 		typeNameResolver:     typeNameResolver,
-
-		snapshotMessage:   snapshotMessage{},
-		dataItems:         make(map[typeAndAddr]output.DataItem),
-		currentlyEncoding: make(map[typeAndAddr]struct{}),
+		snapshotMessage:      snapshotMessage{},
+		dataItems:            make(map[typeAndAddr]output.DataItem),
+		currentlyEncoding:    make(map[typeAndAddr]struct{}),
+		skipIndiciesBuffer:   make([]byte, 1),
```
ajwerner, line 100: Doesn't this mean the code is going to panic if you ever have more than 8 expressions? Instead I think in init you need to make sure this is big enough... 

Also, don't you need to zero it? Otherwise you'll just forever skip an index once you've skipped it. Might be nice to test that after you inject an error and then fix what caused it, that the right thing pops out?

pkg/dyninst/decode/decoder_test.go
```diff
@@ -767,12 +764,81 @@ func TestDecoderPanics(t *testing.T) {
...
+		Event:       output.Event(input),
+		ServiceName: "foo"},
+		&noopSymbolicator{},
+		[]byte{},
+	)
+	require.NoError(t, err)
+	require.Contains(t, string(out), "no decoder type found")
+}
+
+func TestDecoderFailsOnEvaluationErrorAndRetainsPassedBuffer(t *testing.T) {
```
ajwerner, line 811: Can you write a comment explaining what this test is testing? I think you should additionally assert that the prefix of the returned output also has the same prefix as buf... Also, why are you doing it twice? Maybe there's a reason but it's not self-evident, so commentary would be appreciated.

pkg/dyninst/decode/decoder.go
grantseltzer: In `init()` we call `getSkipIndiciesBuffer` which will expand it as needed, even on the first use.

This function also zeroes it.

pkg/dyninst/decode/marshal.go
grantseltzer: This is the discussion i'm referencing:

https://github.com/DataDog/datadog-agent/pull/39454#discussion_r2267396990

pkg/dyninst/decode/decoder.go
```diff
@@ -259,11 +281,45 @@ func (s *snapshotMessage) init(
...
+// getSkipIndiciesBuffer returns a zeroed byte slice of the required size,
+// reusing the internal buffer when possible to avoid allocations.
+func (d *Decoder) getSkipIndiciesBuffer(numExpressions int) []byte {
+	requiredBytes := (numExpressions + 7) / 8
+	if cap(d.skipIndiciesBuffer) < requiredBytes {
+		d.skipIndiciesBuffer = make([]byte, requiredBytes)
+	} else {
+		// Reuse existing buffer, just resize and clear
+		d.skipIndiciesBuffer = d.skipIndiciesBuffer[:requiredBytes]
+		for i := range d.skipIndiciesBuffer {
```
ajwerner, line 320: nit: could use `clear` (a thing to know about)

pkg/dyninst/decode/marshal.go
ajwerner: Okay

ajwerner (approved):
LGTM

PR #39455, opened 2025-08-01
[OTEL-2730] Add container ID detection logic in infraattributes processor

jade-guiton-dd:
### What does this PR do?

This PR modifies the `infraattributes` processor (used in the Agent's OTLP ingest and DDOT pipelines) to leverage the Tagger's `GenerateContainerIDFromOriginInfo` method, in order to determine container IDs based on auxiliary information (PID, Cgroup Inode, Pod UID + container name), when `container.id` is missing from an OTel payload. This auxiliary information is extracted from resource attributes based on OTel semantic conventions.

I also implemented extraction of two "custom" resource attributes which don't exist in semantic conventions:
- `datadog.container.cgroup_inode` for users who might want to leverage inode-based detection;
- `datadog.container.is_init` since the container name-based detection seems to require determining whether we are dealing with an init-container or not.

### Motivation

We can expect in-agent container ID detection to become crucial for Agent OTLP ingest / DDOT users as more and more of them transition to a setup (cgroups v2 + k8s) where in-container detection of container IDs becomes impossible. See [this document](https://docs.google.com/document/d/1XP0dXjl46_y7JhV6ke6tM7J7gL4JF4HFc8YJW2LXDKI) for more detail.

While DDOT users could reasonably set up the `k8sattributes` processor to perform container name-based detection (with the added benefit that the pod can be detected from the source IP), this is not possible with OTLP ingest's fixed pipeline.

Added some basic unit tests to check extraction of container IDs with a mock tagger.

### RC validation

I haven't found an easy way to manually test this PR locally, but it should be reasonably easy with a release candidate. I can provide some code to do it or do it myself.

- Run the Agent in a Kubernetes environment with OTLP ingest enabled
- Send an OTLP logs payload to it containing valid `k8s.container.name` (hardcoded) and `k8s.pod.uid` (from K8s downward API) resource attributes
- Check that the final log contains a corresponding `container_id` tag.

Doing the same with a trace payload and a metric payload will also validate the changes in #40203.

### Additional Notes

I plan to make a second PR adding the infraattributes processor in all OTLP ingest pipelines (instead of just for logs), and disabling the equivalent logic in the serializerexporter / trace agent.


jade-guiton-dd:
The failures to upload to Codecov seem to be happening on main too, so I'm ignoring them for now.

cswatt (approved):

zhuminyi (approved):

comp/otelcol/otlp/components/processor/infraattributesprocessor/common.go
```diff
@@ -113,6 +121,37 @@ func (p infraTagsProcessor) ProcessTags(
...
+		originInfo.ExternalData.ContainerName = k8sContainerName.AsString()
+	}
+
+	// Ad-hoc attributes for data not covered by K8s semantic conventions
+	if cgroupInode, ok := attrs.Get("datadog.container.cgroup_inode"); ok && cgroupInode.Type() == pcommon.ValueTypeInt {
+		originInfo.LocalData.Inode = uint64(cgroupInode.Int())
+	}
+	if initContainer, ok := attrs.Get("datadog.container.is_init"); ok && initContainer.Type() == pcommon.ValueTypeBool {
+		originInfo.ExternalData.Init = initContainer.Bool()
+	}
```
songy23, line 150: Are these documented somewhere?

songy23 (approved):

comp/otelcol/otlp/components/processor/infraattributesprocessor/common.go
jade-guiton-dd: They are not, they're new and completely ad-hoc for this feature. I'm thinking of rewriting the Google Doc as a Confluence page, so I'll probably document these there, for supporting the rare customers who want to use them. But maybe we should consider writing a public-facing doc page about container ID resolution 🤔

songy23: May be worth adding to the readme here: http://github.com/DataDog/datadog-agent/tree/main/comp/otelcol/otlp/components/processor/infraattributesprocessor#expected-attributes

PR #39459, opened 2025-08-01
[NDMII-3574] Update NDM autodiscovery Digest

Pierre-L42:
### What does this PR do?

This PR updates the autodiscovery `Digest` function by adding a parameter that specifies whether to calculate the hash based on the `authentications` config or based on the old config used to authenticate to a device.
We still need to support old hash because it is used to compute the name of autodiscovery cache file. So we can migrate to the new hash by renaming the file with old hash when we detect it.

Changed the autodiscovery status code to make it not depend on `Digest` function by using a `subnetIndex` to sort subnets.

### Motivation

https://dd.slack.com/archives/C07RHJTEFRD/p1754039311773789

Tested on Mimic:
- Tested that the cache with old hash format is renamed with the new hash format.
- Tested that `authentications` config works again.
- Tested that configs without `authentications` still work.


louis-cqrl (approved):

pkg/snmp/snmp.go
```diff
@@ -234,22 +234,39 @@ func NewListenerConfig() (ListenerConfig, error) {
 }
 
 // Digest returns an hash value representing the data stored in this configuration, minus the network address
-func (c *Config) Digest(address string) string {
+// TODO: Remove support for legacy format when Agent reaches version 7.76+: see https://github.com/DataDog/datadog-agent/pull/39459
+func (c *Config) Digest(address string, useLegacyFormat bool) string {
```
TCheruy: For readability, I think it would be better to have two separate functions: `Digest` and `LegacyDigest`. We'll have some code duplication but i think its ok here

comp/core/autodiscovery/listeners/snmp.go
```diff
@@ -206,18 +207,16 @@ func (l *SNMPListener) checkDevice(job snmpJob) {
 		}
 
 		deviceInfo := l.checkDeviceInfo(authentication, job.subnet.config.Port, deviceIP)
+		l.createService(entityID, job.subnet, deviceIP, deviceInfo, authIndex, true)
```
TCheruy, line 210: Thank you for doing these small code improvements, this is great!

TCheruy (approved):
LGTM! Small nit

PR #39460, opened 2025-08-01
Fix JUnit uploads blocked by undrained output streams

rdesgroppes:
### What does this PR do?
This change improves the upload of JUnit reports, especially in the `tests_windows-x64` job.

It addresses two inefficiencies in how subprocesses are managed:
1. Upload processes were launched with `stdout`/`stderr` pipes, requiring the parent process to actively drain them. If not drained promptly, the child process blocks, ignoring graceful termination signals.
  In practice, subprocesses were only drained one at a time per caller thread using `communicate`.
2. Multiple subprocesses were launched from the same thread in the thread pool, exceeding the intended parallelism and defeating controlled scheduling.

The fix lets subprocesses inherit `stdout`/`stderr` directly (no pipes) and ensures each upload process runs in its own thread pool task.

### Motivation
- Failed jobs with seemingly lingering background processes, for instance: https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1058500308
- Slack thread: https://dd.slack.com/archives/C06PBHLD4DQ/p1754040098218369
- Might help address [ACIX-968]

Temporarily forced `RUN_UNIT_TESTS: "on"` in an [intermediate commit](https://github.com/DataDog/datadog-agent/pull/39460/commits/55c48733f1e6e04cb0b28930e4dde60dca7d866a) (reverted since).
Output is no longer delayed and now renders verbatim: https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1061859337

The original `ThreadPoolExecutor` is preserved by the change, but we could later consider switching to `asyncio` (Python's answer to writing clean, efficient, and scalable code for concurrent I/O operations).

### Additional Notes
[ACIX-968]: https://datadoghq.atlassian.net/browse/ACIX-968?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tasks/libs/common/junit_upload_core.py
+++ b/tasks/libs/common/junit_upload_core.py
@@ -2,15 +2,15 @@ import io
 import os
 import platform
 import re
-import sys
 import tarfile
 import tempfile
 import xml.etree.ElementTree as ET
 from collections import defaultdict
+from collections.abc import Iterator
 from concurrent.futures import ThreadPoolExecutor
 from pathlib import Path
 from shutil import which
-from subprocess import PIPE, CalledProcessError, Popen
+from subprocess import check_call
 
 from invoke.exceptions import Exit
 
@@ -104,8 +104,7 @@ def junit_upload_from_tgz(junit_tgz, result_json, codeowners_path=".github/CODEO
         # Upload junit on a per-team basis (all folders except the one part of the original archive)
         team_folders = [item for item in working_dir.iterdir() if item.is_dir() and item not in xml_folders]
         with ThreadPoolExecutor() as executor:
-            for log in executor.map(upload_junitxmls, team_folders):
-                print(log)
+            _upload_junitxmls(team_folders, executor)
 
 
 def get_flaky_failures_and_marked_flaky_tests_from_test_output(result_json):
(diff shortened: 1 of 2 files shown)


CelianR (approved):
I'm not 100% sure that this will fix the issue. It's maybe a runner issue (@KevinFairise2 asked CI execution for that). I'm not against trying this however, maybe you could try by running 10 times the windows job? Like adding `parallel: 10` to the impacted jobs

rdesgroppes:
> I'm not 100% sure that this will fix the issue.

While I'm pretty sure the change won't fix 100% of the issue**s**, it will at least cover the following ones:
```python
    for flags, files in xml_files.items():
        args = set_tags(owner, flavor, flags, additional_tags, files[0])
        args.extend(files)
        processes.append(Popen(datadog_ci_command + args, bufsize=-1, env=process_env, stdout=PIPE, stderr=PIPE))
        # 1.             ^ starts len(xml_files) processes in the background, disregarding ThreadPoolExecutor's max_workers

    for process in processes:
        stdout, stderr = process.communicate()
        # 2.                     ^ takes only 1 process in the foreground at a time, which means next background processes:
        #                          - keep executing until one of their output streams saturates
        #                          - are unlikely to react to graceful termination signals (let alone once blocked)
        print(stdout)
        # 3.  ^ buffered standard output is printed at once
        #       - all lines are misleadingly prefixed with the same timestamp (down to the ms, despite 30+s between first and last ones)
        #       - rendered as bytes whereas the original output is UTF8-encoded (console shows `\n` instead of actual newlines)
        print(f" Uploaded {len(tuple(team_dir.iterdir()))} files for {team_dir.name}")
        if stderr:
            print(f"Failed uploading junit:\n{stderr.decode()}", file=sys.stderr)
            # 4.  ^ same as above for buffered standard error
            raise CalledProcessError(process.returncode, datadog_ci_command)
            # 5.  ^ assumes that an error message implies a nonzero exit code
            #       - transient errors may be logged, which should not prevent a programmatic recovery (successful retry)
            #       - if there's no error message, then an actual nonzero exit code will be lost
    return ""  # For ThreadPoolExecutor.map. Without this it prints None in the log output.
    # 6.   ^ workaround mandated by the approach
```

KevinFairise2:
Interesting detailed explanation! I think that part of the fix:
`# 1.             ^ starts len(xml_files) processes in the background, disregarding ThreadPoolExecutor's max_workers`
Could help with the issue we are currently seeing. Let's have it merged and see if we still see flakiness in that job

KevinFairise2 (approved):

rdesgroppes:
> Interesting detailed explanation! I think that part of the fix: `# 1. ^ starts len(xml_files) processes in the background, disregarding ThreadPoolExecutor's max_workers` Could help with the issue we are currently seeing. Let's have it merged and see if we still see flakiness in that job

Next leads:
1. fix [ThreadPoolExecutor.max_workers](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor) to something we want? (instead of the default: `min(32, os.cpu_count() + 4)` since Python 3.8, `(os.process_cpu_count() or 1) + 4` since Python 3.13)
2. use [ProcessPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ProcessPoolExecutor) instead?
3. use [asyncio](https://docs.python.org/3/library/asyncio-subprocess.html)?
4. prevent `datadog-ci` from proceeding with concurrent uploads (default behavior) since the Python wrapper already performs parallelism?

rdesgroppes:
> The backport to `7.69.x` failed:
> [...]
> Then, create a pull request where the `base` branch is `7.69.x` and the `compare`/`head` branch is `backport-39460-to-7.69.x`.

Manually created with way less steps: (who needs worktrees?)
- #40068

PR #39462, opened 2025-08-01
tasks/system_probe.py: use -buildvcs=false more

ajwerner:
On the vagrant vms where the fs is slow, invoking git is the dominant cost of these tasks. This stops doing the go tool from invoking git. We weren't relying on the git information here.

For:
```bash
rm -rf /git/datadog-agent/pkg/dyninst/testprogs/binaries
time dda inv -e system-probe.build-dyninst-test-programs
```

Before:
```
real	3m59.660s
user	1m5.437s
sys	2m43.179s
```

After:
```
real	0m10.396s
user	0m23.133s
sys	0m9.839s
```

### DIFF
--- a/tasks/system_probe.py
+++ b/tasks/system_probe.py
@@ -1080,7 +1080,9 @@ def go_package_dirs(packages, build_tags):
     else:
         packages_arg = " ".join(packages)
 
-    cmd = f"go list -find -f \"{format_arg}\" -mod=readonly -tags \"{buildtags_arg}\" {packages_arg}"
+    # Disable buildvcs to avoid attempting to invoke git, which can be slow in
+    # VMs and we don't need its output here.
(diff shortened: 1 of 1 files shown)


tasks/system_probe.py
```diff
@@ -2246,7 +2246,7 @@ def ninja_add_dyninst_test_programs(
             pool="gobuild",
             variables={
                 "go": go_path,
-                "extra_arguments": "-trimpath",
+                "extra_arguments": "-trimpath -buildvcs=false",
```
andreimatei: add a comment about why the unusual parameter

andreimatei (approved):

tasks/system_probe.py
ajwerner: Done.

gjulianm (approved):

PR #39466, opened 2025-08-01
[AGENTCFG-87] Panic-ing if multiple transformers are set for an envvar

rahulkaukuntla:
### What does this PR do?

Something like this should not work:

> conf.ParseEnvAsSlice("a", <func>)
> 
> conf.ParseEnvAsSliceMapString("a", <func>)

so I just check if a transformer has already been set for an envvar when registering a new transfer for said envvar, and panic if it has been set already.

I added unit tests.


dustmop (changes requested):
Could you add this fix to viper as well? Unfortunately the `Viper` object doesn't provide access to what env vars are already registered, so we will need to add it in the wrapper code in `pkg/config/viperconfig/viper.go`, specifically the `setEnvTransformer` method.

pkg/config/nodetreemodel/config.go
```diff
@@ -592,30 +592,52 @@ func (c *ntmConfig) insertNodeFromString(curr InnerNode, key string, envval stri
...
 	c.Lock()
 	defer c.Unlock()
+	if _, exists := c.envTransform[strings.ToLower(key)]; exists {
+		panic(fmt.Sprintf("env transform for %s already exists", key))
+	}
 	c.envTransform[strings.ToLower(key)] = func(k string) interface{} { return fn(k) }
 }
 
+// ClearEnvTransformer tells the config to assume that no env transformer is registered for the given key
+func (c *ntmConfig) ClearEnvTransformer(key string) {
```
dustmop: This method isn't needed. It's only used in the NodeTreeModel test. Instead, just use a different key for the 2nd part of the test.

pkg/collector/python/test_datadog_agent.go
```diff
@@ -129,6 +129,7 @@ func testEmitAgentTelemetry(t *testing.T) {
 func testObfuscaterConfig(t *testing.T) {
 	pkgconfigmodel.CleanOverride(t)
 	conf := pkgconfigmock.New(t)
+	conf.CompletelyClearEnvTransformers()
```
dustmop: Is this needed? The config just got created and hasn't had any env transformers added.

cmd/otel-agent/config/agent_config.go
```diff
@@ -99,6 +99,7 @@ func NewConfigComponent(ctx context.Context, ddCfg string, uris []string) (confi
 
 	// Set the global agent config
 	pkgconfig := pkgconfigsetup.Datadog()
+	pkgconfig.CompletelyClearEnvTransformers()
```
dustmop: No need to add this here. Short term: have `SetEnvPrefix` clear the env transformers for Viper. Long term: once https://github.com/DataDog/datadog-agent/pull/39395 is merged we'll have better control over how this double building of the config works.

pkg/config/model/types.go
```diff
@@ -247,6 +247,8 @@ type Setup interface {
 	ParseEnvAsMapStringInterface(key string, fx func(string) map[string]interface{})
 	ParseEnvAsSliceMapString(key string, fx func(string) []map[string]string)
 	ParseEnvAsSlice(key string, fx func(string) []interface{})
+	ClearEnvTransformer(key string)
+	CompletelyClearEnvTransformers()
```
dustmop: Adding these to the public interface isn't necessary. We don't want other places in the code calling these methods, as they're not something callers should care about doing.

pkg/config/nodetreemodel/config_test.go
```diff
@@ -1153,3 +1153,33 @@ func TestSequenceID(t *testing.T) {
...
+	assert.NotPanics(t, func() {
+		config.ParseEnvAsStringSlice("list_of_nums", func(in string) []string {
+			return strings.Split(in, ",")
+		})
+	}, "env transform for list_of_nums works if set once")
+	config.ClearEnvTransformer("list_of_nums")
+
+	assert.PanicsWithValue(t, "env transform for list_of_nums already exists", func() {
+		config.ParseEnvAsSlice("list_of_nums", func(in string) []interface{} {
+			vals := []interface{}{}
```
dustmop: This function body should be replaced by something simpler, so as to not distract from what the test's purpose is. Something like `return []string{"a", "b"}` is fine instead.

pkg/config/viperconfig/viper_test.go
```diff
@@ -525,3 +526,33 @@ func TestSequenceID(t *testing.T) {
...
+	assert.NotPanics(t, func() {
+		config.ParseEnvAsStringSlice("list_of_nums", func(in string) []string {
+			return strings.Split(in, ",")
+		})
+	}, "env transform for list_of_nums works if set once")
+	config.ClearEnvTransformer("list_of_nums")
+
+	assert.PanicsWithValue(t, "env transform for list_of_nums already exists", func() {
+		config.ParseEnvAsSlice("list_of_nums", func(in string) []interface{} {
+			vals := []interface{}{}
```
dustmop: Same here, please replace this function body with something simpler.

dustmop (changes requested):

jeremy-hanna (approved):
👍 for agent-runtimes files

pkg/config/nodetreemodel/config_test.go
rahulkaukuntla: Done.

pkg/config/viperconfig/viper_test.go
rahulkaukuntla: Done.

pkg/config/nodetreemodel/config.go
rahulkaukuntla: Sounds good, just using a different key as suggested.

pkg/config/model/types.go
rahulkaukuntla: Makes sense, removed.

cmd/otel-agent/config/agent_config.go
rahulkaukuntla: Per your suggestion, I added a line to SetEnvPrefix to also reset the existing transformers slice when calling it, so now there's no need to add methods to the `pkg/config/model/types.go` file.

pkg/collector/python/test_datadog_agent.go
rahulkaukuntla: I don't understand why, but yes, it seems that env transformers are added (for example, `dogstatsd_mapper_profiles`). As such, I just used the suggestion you mentioned [here](https://github.com/DataDog/datadog-agent/pull/39466#discussion_r2258287563) to resolve this issue.

pkg/collector/python/test_datadog_agent.go
```diff
@@ -129,6 +129,7 @@ func testEmitAgentTelemetry(t *testing.T) {
 func testObfuscaterConfig(t *testing.T) {
 	pkgconfigmodel.CleanOverride(t)
 	conf := pkgconfigmock.New(t)
+	conf.SetEnvPrefix("")
 	pkgconfigsetup.InitConfig(conf)
```
dustmop: This is the better fix:

```suggestion
	conf := pkgconfigmock.New(t)
```

The call to `mock.New` already calls `InitConfig`, [right here](https://github.com/DataDog/datadog-agent/blob/42da4510cc398ca6eb0cf2e3256e94120253169b/pkg/config/mock/mock.go#L54). Meaning all of the env transfers are already setup by the constructor, and this second call would erroneously add them again.

dustmop (approved):
LGTM, with one last suggestion to improve a test. Once that fixed, this is good to merge.

pkg/collector/python/test_datadog_agent.go
rahulkaukuntla: Nice, thanks for the look. incident 41849 is still blocking the merging of this, so I'll merge once it's resolved.

PR #39471, opened 2025-08-01
feat(trace-agent): [SVLS-7262] optional remote tagger

apiarian-datadog:
### What does this PR do?
Changes the remote tagger used in the trace-agent to an optional remote tagger. If the remote tagger is disabled, the noop tagger will be used.

### Motivation
The Azure App Services Extension uses the trace-agent but cannot provide a core agent to support the remote tagger which was made non-optional in [7.61.0](https://github.com/DataDog/datadog-agent/releases/tag/7.61.0). Since there are no containers for us to tag in Azure App Services environments, we don't actually need the remote tagger there, so the noop tagger is sufficient.

- [x] Tests in actual AAS Extension builds in both the AAS Windows Extension and AAS Linux Wrapper

This exposes all users of the trace-agent to this new setting. The setting defaults to false (not disabling the remote tagger). But clear documentation may be required to ensure that users understand the dynamics.


comp/core/tagger/impl-optional-remote/optional-remote.go
```diff
@@ -0,0 +1,80 @@
...
+		IPC:       req.IPC,
+	}
+
+	remoteProvides, err := remote.NewComponent(remoteRequires)
+	if err != nil {
+		return Provides{}, err
+	}
+
+	return Provides{remoteProvides}, nil
+}
```
GustavoCaso:
```suggestion
func NewComponent(req Requires) (Provides, error) {
	if req.OptionalRemoteParams.Disable() {
		return Provides{
			remote.Provides{
				Comp: noop.NewComponent(),
			},
		}, nil
	}

	remoteRequires := remote.Requires{
		Lc:        req.Lc,
		Params:    req.RemoteParams,
		Config:    req.Config,
		Log:       req.Log,
		Telemetry: req.Telemetry,
		IPC:       req.IPC,
	}

	remoteProvides, err := remote.NewComponent(remoteRequires)
	if err != nil {
		return Provides{}, err
	}

	return Provides{remoteProvides}, nil
}
```

We can simplify the disabled branch. The IPC endpoint in this case is not necessary, as I don't think in the Azure environment, people are going to execute the `agent tagger-list` command, as there is no agent binary 😄

apiarian-datadog: oh, i'll give that a try. i think there was some sort of error when i omitted the endpoint, but i don't remember now

apiarian-datadog: yup, this works!

drichards-87 (approved):

releasenotes/notes/azure-app-services-extension-trace-agent-fix-c1238af92f2f015b.yaml
```diff
@@ -0,0 +1,6 @@
+---
+fixes:
+  - |
+    Restores the ability to run the trace-agent in the Azure App Services
```
jcstorms1: Nit: Rather than saying `run the trace-agent` I'd say `update the trace-agent`. We've always been able to run it but we were unable to update from a pinned version.

jcstorms1 (approved):

releasenotes/notes/azure-app-services-extension-trace-agent-fix-c1238af92f2f015b.yaml
apiarian-datadog: aah, fixed!

GustavoCaso (approved):

ajgajg1134 (approved):

zhuminyi:
I noticed that the remoteTagger.Options struct already has an unused [Disabled](https://github.com/DataDog/datadog-agent/blob/main/comp/core/tagger/impl-remote/remote.go) field that could solve this issue more simply. Have you considered using that and returning a noop tagger when disabled

zhuminyi (approved):

PR #39474, opened 2025-08-01
Fetch Host CCRID for Azure, add to v5 Metadata Payload

dustmop:
### What does this PR do?

When we fetch the HostAliases at startup, also return the cloud provider name. Then fetch the CCRID for the Host, only for Azure currently. The new config flag `collect_ccrid` can be used to disable this fetch, in case it causes delays at startup.

### Motivation

Send Host CCRIDs to the backend to uniquely identify hosts.

Tested manually on Azure.

Unit tests updated for GetHostAliases.

Since this is adding another http request to startup, it could cause additional slowdown.


pkg/util/cloudproviders/azure/azure.go
```diff
@@ -174,6 +174,23 @@ func getHostnameWithConfig(ctx context.Context, config model.Config) (string, er
...
 
+var hostCCRIDFetcher = cachedfetch.Fetcher{
+	Name: "Azure Host CCRID",
+	Attempt: func(ctx context.Context) (interface{}, error) {
+		rg, err := getResponse(ctx,
+			metadataURL+"/metadata/instance/compute/resourceId?api-version=2021-02-01&format=text")
+		if err != nil {
+			return "", fmt.Errorf("unable to query metadata endpoint: %s", err)
+		}
+		return rg, nil
```
vksrinivasan1, line 185: For azure specifically, I think we should lowercase these. Azure APIs are inconsistent in the casing they return back for resource IDs so the azure ints team [gave the directive](https://docs.google.com/document/d/1WyATANotBEXJg9gK70RJTliewuVMUkAnKDT7YV19ml4/edit?tab=t.0#heading=h.9zts2pof3eet) to lowercase azure CCRIDs when associating it with data. cc @parsons90 to confirm

jeremy-hanna (approved):

pkg/util/cloudproviders/azure/azure.go
dustmop: Thanks for the tip, added another commit.

rahulkaukuntla (approved):

pkg/util/cloudproviders/azure/azure.go
```diff
@@ -174,6 +174,30 @@ func getHostnameWithConfig(ctx context.Context, config model.Config) (string, er
 	return name, nil
 }
 
+var hostCCRIDFetcher = cachedfetch.Fetcher{
```
andresromerodatadog, line 177: Does this cache it for the entire runtime?

One thing that can happen to compute instances is that they can be moved to a new resource group, which changes their CCRID: https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/move-resource-group-and-subscription?tabs=azure-cli

I'm wondering how long before we catch this and update it. The instance remains funcitonal the whole time so the agent would still be running.

pkg/util/cloudproviders/azure/azure.go
```diff
@@ -174,6 +174,30 @@ func getHostnameWithConfig(ctx context.Context, config model.Config) (string, er
...
+
+// GetHostCCRID returns the Canonical Cloud Resource ID for the Azure host
+func GetHostCCRID(ctx context.Context) (string, error) {
+	caseInsensitiveCCRID, err := hostCCRIDFetcher.FetchString(ctx)
+	if err != nil {
+		return "", err
+	}
+	// Azure APIs are inconsistent with handling case, so it is recommended to
+	// lower-case returned strings that represent stable IDs
+	// https://docs.google.com/document/d/1WyATANotBEXJg9gK70RJTliewuVMUkAnKDT7YV19ml4/edit?tab=t.0#heading=h.9zts2pof3eet
```
andresromerodatadog: I don't usually contribute to this repo, is linking to internal google docs normal?

Alternatively we can link the page that says ids are case insensitive for compute: https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/move-resource-group-and-subscription?tabs=azure-cli

andresromerodatadog (commented):
Do you have an example of an ID returned by this code? I just want to verify it aligns with what we expect.

pkg/util/cloudproviders/azure/azure.go
dustmop: It's done on occasion for RFCs, but I agree that using a public doc is better. I will update.

pkg/util/cloudproviders/azure/azure.go
dustmop: Yes, it would be cached for the Agent's entire lifetime.

What initiates moving compute instances to a new resource group? Would this require manual intervention or can it happen automatically? If it's automatic is there some sort of event or notification that the Agent could receive in order to know to refresh this value? Otherwise I think we could advice customers to simply restart the Agent when this happens.

dustmop:
> Do you have an example of an ID returned by this code? I just want to verify it aligns with what we expect.

Looks like this, some fields have been changed: `/subscriptions/1234abcd-78ef-ab12-cd45-10abc2345def/resourceGroups/MYRESOURCES/providers/Microsoft.Compute/virtualMachines/my-vm-name`

pkg/util/cloudproviders/azure/azure.go
andresromerodatadog: It can be a customer in the portal, or with the CLI, or with the API. It doesn't require manual intervention but my guess is that it is rare (like when rearchitecting subscriptions). 

I can't find any automatic event or notification that would let the agent know, either. The VM isn't told it is being moved.

My feeling is rather than caching it forever and depending on customers to do "the right thing", we could cache it with a pretty long TTL just so that we get consistency eventually. The worst outcome here is the customer never restarts the agent and we have mismatched CCRIDs forever.

andresromerodatadog:
> > Do you have an example of an ID returned by this code? I just want to verify it aligns with what we expect.
> 
> Looks like this, some fields have been changed: `/subscriptions/1234abcd-78ef-ab12-cd45-10abc2345def/resourceGroups/MYRESOURCES/providers/Microsoft.Compute/virtualMachines/my-vm-name`

Assuming this is before the lowercasing, this looks good to me.

comp/metadata/inventoryhost/inventoryhostimpl/inventoryhost.go
```diff
@@ -266,6 +267,7 @@ func (ih *invHost) fillData() {
 
 	ih.data.CloudProviderSource = cloudproviders.GetSource(cloudProvider)
 	ih.data.CloudProviderHostID = cloudproviders.GetHostID(context.Background(), cloudProvider)
+	ih.data.CanonicalCloudResourceID = cloudproviders.GetHostCCRID(context.Background(), []string{cloudProvider})
```
PCaponetti: not sure if you meant to check the flag `host_disable_ccrid_fetch` here as well?

pkg/util/cloudproviders/azure/azure.go
PCaponetti: Curious what "long" means for TTL here. I don't know the full set of implications of having the wrong CCRID (as we haven't designed all the consumers of this data). I know the initial value add is on new resource startup so we likely will at least get that right. Is it expensive to fetch from the metadata API (maybe lazily on event submission?)? Is there any mechanism where we can communicate back to the hosts (guessing no) to invalidate cache when we detect a change in resource ID (hard to detect, just curious what's possible)?

comp/metadata/inventoryhost/inventoryhostimpl/inventoryhost.go
dustmop: It's not important for the Inventory Host payload. The other fetch is for the Host "v5" Payload, which happens at startup and so is timing-sensitive, it affects how it takes for the Agent to startup. Inventory Host happens later so the time that it takes doesn't matter as much.

pkg/util/cloudproviders/azure/azure.go
dustmop: My mistake, I misunderstood how the code worked. The cache doesn't prevent future fetches, rather it is [only used if contacting the cloud API fails](https://github.com/DataDog/datadog-agent/blob/d07dfbec8455969731f849e3d55c967ae970e60c/pkg/util/cachedfetch/fetcher.go#L16). If the API just returns a different value then the next fetch will see that value.

andresromerodatadog: Oh then this is absolutely fine from this perspective. I'll approve.

andresromerodatadog (approved):

PCaponetti (approved):

pkg/util/cloudproviders/cloudproviders.go
```diff
@@ -118,9 +119,10 @@ var hostAliasesDetectors = []cloudProviderAliasesDetector{
 	{name: kubernetes.CloudProviderName, callback: kubernetes.GetHostAliases},
 }
 
-// GetHostAliases returns the hostname aliases from different provider
-func GetHostAliases(ctx context.Context) []string {
+// GetHostAliases returns the hostname aliases and the name of the possible cloud providers
+func GetHostAliases(ctx context.Context) ([]string, []string) {
```
hush-hush: We should never detect multiple cloudproviders at the same time, no ? maybe relying on `DetectCloudProvider` would be better in order to use `GetHostCCRID` later.

WDYT ?

pkg/util/cloudproviders/cloudproviders.go
```diff
@@ -138,13 +140,37 @@ func GetHostAliases(ctx context.Context) []string {
...
+}
+
+type cloudProviderCCRIDDetector func(context.Context) (string, error)
+
+var hostCCRIDDetectors = map[string]cloudProviderCCRIDDetector{
+	azure.CloudProviderName: azure.GetHostCCRID,
+}
+
+// GetHostCCRID returns the host CCRID from the first provider that works
+func GetHostCCRID(ctx context.Context, cloudproviders []string) string {
```
hush-hush: Passing multiple cloudproviders to fetch an unique ID is not intuitive I think. You don't know as a caller which cloudprovider answered even if in real life we know that only one should return something.

hush-hush (changes requested):

pkg/util/cloudproviders/cloudproviders.go
dustmop: I would rather not add a third call to startup by relying on `DetectCloudProvider`, if possible. The thing with `GetHostAliases` is that its written as though multiple cloud providers might return aliases (because some of them use the same endpoints?), but maybe that's just a mistake in how its written.

dustmop: Okay, talked offline, now returning the first cloudprovider that returns host aliases, and warning if others respond with something.

flushentitypacket (approved):

pkg/config/setup/config.go
```diff
@@ -289,6 +289,7 @@ func InitConfig(config pkgconfigmodel.Setup) {
 	config.BindEnvAndSetDefault("check_sampler_expire_metrics", true)
 	config.BindEnvAndSetDefault("check_sampler_context_metrics", false)
 	config.BindEnvAndSetDefault("host_aliases", []string{})
+	config.BindEnvAndSetDefault("host_disable_ccrid_fetch", false)
```
hush-hush: We have a bunch of `collect_XXX` settings for the metadata (`collect_gce_tags`, `collect_ec2_tags`, `collect_ec2_instance_info`, `collect_gpu_tags`). What do you think about `collect_ccrid` instead ?

We can change this in a later PR if needed.

hush-hush (approved):

pkg/config/setup/config.go
dustmop: Sure thing, changed to `collect_ccrid`

PR #39476, opened 2025-08-01
[pkg/clusteragent/autoscaling/workload] Fix unit test timestamp to prevent timezone mismatch

jennchenn:
### What does this PR do?

Do not cast timestamp used in autoscaling unmarshal unit test to prevent failures when run locally

### Motivation

Unit test failed to pass locally due to timestamp discrepancy

1. Run the unit test: `TZ="Europe/Berlin" go test -v -tags "kubeapiserver test" ./pkg/clusteragent/autoscaling/workload -run TestMarshalUnmarshal -count=1`
2. Verify it passes


pkg/clusteragent/autoscaling/workload/dump_test.go
```diff
@@ -175,7 +175,7 @@ Settings: map[key:value]
 
 func TestMarshalUnmarshal(t *testing.T) {
 	// json serialization drops nanoseconds; strip it here
-	testTime := time.Unix(time.Now().Unix(), 0).UTC()
+	testTime := time.Unix(time.Now().Unix(), 0)
```
vboulineau: Why does it fail? The code should not be sensitive to that and all operations should use UTC timestamps

dustmop: The serialization step causes this to fail. Seems like something in the marshaller is looking at the locale. I tried fixing this on a [previous PR](https://github.com/DataDog/datadog-agent/pull/38560) but it wasn't the correct approach and [got reverted](https://github.com/DataDog/datadog-agent/pull/38632).

jennchenn: I changed the serialization logic so it uses UTC now - it seems there was an issue with unmarshalling `metav1.Time` and it would convert those timestamps to the local timezone instead

vboulineau (approved):

PR #39281, opened 2025-07-25
retry msi error 1601

clarkb7:
### What does this PR do?
retry when the MSI returns [exit code 1601](https://learn.microsoft.com/en-us/windows/win32/msi/error-codes) (`ERROR_INSTALL_SERVICE_FAILURE`) during remote updates on Windows

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1679
Be more resilient to `msiserver` issues

### Describe how you validated your changes

added to existing unit tests.

For manual testing, disable or force stop the msiserver service.

### Additional Notes

This error is distinct from the `Error 1719. The Windows Installer Service could not be accessed` error described at:
- https://support.gfi.com/article/114416-error-fatal-error-during-installation-when-installing-net-4-framework-on-a-64-bit-machine
- https://www.advancedinstaller.com/forums/viewtopic.php?t=50343\

These cases contain a log like
```
CustomAction XXX returned actual error code 1601
```
but the MSI actually fails with exit code 1603 (generic failure), so we can't identify these cases on exit code alone. maybe because the error occurred within a custom action? Will look into parsing the error out of the log file in a follow-up PR.

PR #39477, opened 2025-08-01
retry msi error 1719

clarkb7:
### What does this PR do?
look at the MSI log for more information to determine whether or not to retry when MSI returns a generic exit code, during remote updates on Windows

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1679
follow up to https://github.com/DataDog/datadog-agent/pull/39281

added unit test

it's not clear if a retry after this error occurs will actually succeed. The log contains error 1601 so it seems to be a case similar to https://github.com/DataDog/datadog-agent/pull/39281 (intermittent connection error / service killed) and may be retryable.

However these support articles point to a registry key (misconfiguration?) as a potential source, and it's the only way I've been able to reproduce this precise sequence of log messages. However I can't find further information about this registry key and why it might exist/have been created.
- https://support.gfi.com/article/114416-error-fatal-error-during-installation-when-installing-net-4-framework-on-a-64-bit-machine
- https://www.advancedinstaller.com/forums/viewtopic.php?t=50343\

In live cases we've seen, the error occurred after the Agent was successfully uninstalled. So this would seem to indicate that the `msiserver` service was indeed functional immediately prior to us seeing the error.

Enabling verbose logging (`/l*v`) provides more precise information about the cause of the 1601/1719 error, but MSI verbose logs are /really/ verbose (> 50 MB) and we don't want that at this time.


BaptisteFoy (approved):

releasenotes/notes/windows-remote-update-retry-1719-8d74b38c23cf5336.yaml
```diff
@@ -0,0 +1,6 @@
+---
+enhancements:
+  - |
+    Remote Agent management on Windows now automatically retries when a MSI
+    custom action returns error 1601 (``ERROR_INSTALL_SERVICE_FAILURE``) and
+    causes ``Error 1719. The Windows Installer Service could not be accessed``.
```
brett0000FF:
```suggestion
    Remote Agent Management on Windows now automatically retries when an MSI
    custom action returns error 1601 (``ERROR_INSTALL_SERVICE_FAILURE``) and
    causes ``Error 1719. The Windows Installer Service could not be accessed``.
```

brett0000FF (approved):

pkg/fleet/installer/msi/msiexec.go
```diff
@@ -356,25 +380,32 @@ func (m *Msiexec) Run(ctx context.Context) ([]byte, error) {
...
-			return nil, backoff.Permanent(err)
+		// Process log file
+		logFileBytes, logErr := m.openAndProcessLogFile()
+		if logErr != nil {
+			err = errors.Join(err, logErr)
+		}
+		if err != nil {
+			// An error occurred, check if it's retryable or permanent
+			if isRetryableExitCode(err) {
+				return logFileBytes, err
```
alexn-dd, line 391: This case is not related to the log file.  Should this be `return nil, err`?

alexn-dd (approved):

pkg/fleet/installer/msi/msiexec.go
clarkb7: we want to always return the log messages so that the caller can examine/report it. It will contain more error information than the exit code can supply. However this goes against Golang convention of caller not rely on other return values if an error is returned. We should probably return the log data through a custom error type instead.
https://datadoghq.atlassian.net/browse/WINA-1693

PR #39497, opened 2025-08-04
Hash omnibus files instead of trying to guess where they last changed

alopezz:
### What does this PR do?

It computes the hash of all the individual files that could impact an omnibus build as part of the git cache key computation, instead of trying to base it on the diff against the last detected changes. This is easier to reason about and easier to get right.

### Motivation

While working on #39342, and rebasing I got "[cacerts dirtied the cache](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1061788064/viewer#L499)" which matches a [recent change](https://github.com/DataDog/datadog-agent/commit/e3847e6711acc067f311f39308733967ff9d3234) (a change that happened between rebases). This means that the mechanism to account for changes in omnibus files when computing the git cache key is not working properly.

I validated manually via debug printing that all the expected files ran through the hasher (I could leave debug printing in if we think that would be helpful).

There's a chance that hashing all the contents is somewhat slower than doing the git thing that we were doing, but in practice it's not really noticeable.

### DIFF
--- a/tasks/libs/common/omnibus.py
+++ b/tasks/libs/common/omnibus.py
@@ -222,7 +222,7 @@ def _hash_paths(hasher, paths: list[str]):
             for filepath in sorted(all_files_under(path)):
                 hash_file(filepath)
         else:
-            raise Exception("provided paths must exist and be either a folder or a regular file")
+            raise ValueError("provided paths must exist and be either a folder or a regular file")
 
 
 def get_dd_api_key(ctx):
--- a/tasks/libs/common/omnibus.py
+++ b/tasks/libs/common/omnibus.py
@@ -202,22 +202,27 @@ def _get_environment_for_cache() -> dict:
     return dict(filter(env_filter, sorted(os.environ.items())))
 
 
-def _last_omnibus_changes(ctx):
-    omnibus_invalidating_files = ['omnibus/config/', 'omnibus/lib/', 'omnibus/omnibus.rb']
-    omnibus_last_commit = ctx.run(
-        f'git log -n 1 --pretty=format:%H {" ".join(omnibus_invalidating_files)}', hide='stdout'
-    ).stdout
-    # The commit sha1 is likely to change between a PR and its merge to main
-    # In order to work around this, we hash the commit diff so that the result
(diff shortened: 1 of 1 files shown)


tasks/libs/common/omnibus.py
```diff
@@ -202,22 +202,27 @@ def env_filter(item):
...
+                yield os.path.join(root, filename)
+
+    for path in sorted(paths):
+        if os.path.isfile(path):
+            hash_file(path)
+        elif os.path.isdir(path):
+            for filepath in sorted(all_files_under(path)):
+                hash_file(filepath)
+        else:
+            raise Exception("provided paths must exist and be either a folder or a regular file")
```
rdesgroppes:
```suggestion
            raise ValueError("provided paths must exist and be either a folder or a regular file")
```
(avoid generic exceptions)

tasks/libs/common/omnibus.py
```diff
@@ -202,22 +202,27 @@ def env_filter(item):
...
-    hash = hashlib.sha1()
-    hash.update(str.encode(omnibus_last_changes))
-    result = hash.hexdigest()
-    print(f'Hash for last omnibus changes is {result}')
-    return result
+def _hash_paths(hasher, paths: list[str]):
+    """Update hashlib.hash object `hasher` by recursive hashing of the contents provided in `paths`."""
+
+    def hash_file(filepath):
+        with open(filepath, 'rb') as f:
```
rdesgroppes, line 213: shouldn't we add the (relative) path itself? (to track file moves)

alopezz: The idea being that a file being renamed without changed contents (and still being in the same order after sorting) would give us the same hash? Sounds unlikely but indeed it doesn't hurt to account for it.

alopezz: Added this in [3b0aeae](https://github.com/DataDog/datadog-agent/pull/39497/commits/3b0aeaeef0ee876c7b6615548f1c3ab310ce2c5b).

rdesgroppes (approved):

PR #39499, opened 2025-08-04
Bootstrapping CCRID detection in the Agent

hush-hush:
### What does this PR do?

Bootstrapping CCRID detection in the Agent.

Running the CI is enough.


louis-cqrl (approved):

comp/metadata/host/hostimpl/utils/meta.go
```diff
@@ -37,6 +37,7 @@ type Meta struct {
 	ClusterName               string   `json:"cluster-name,omitempty"`
 	LegacyResolutionHostname  string   `json:"legacy-resolution-hostname,omitempty"`
 	HostnameResolutionVersion int      `json:"hostname-resolution-version,omitempty"`
+	CCRID                     string   `json:"ccrid,omitempty"`
```
dustmop, line 40: Given that the other fields on this struct are rather long, what do you think about replacing the field name with a non-acronym version: `CanonicalCloudResourceID`? We can still use `ccrid` for the serialization format. It should help with readability for future devs that are not as familiar with the acronym.

comp/metadata/host/hostimpl/utils/meta.go
```diff
@@ -74,6 +75,7 @@ func getMeta(ctx context.Context, conf model.Reader, hostnameComp hostnameinterf
 		InstanceID:                instanceID,
 		AgentHostname:             agentHostname,
 		HostnameResolutionVersion: 1,
+		CCRID:                     cloudproviders.GetCCRID(ctx),
```
dustmop, line 78: If you look at how `GetHostAliases` works, it does extra work to be fast because this code runs at Agent startup. `GetHostAliases` sends requests to all possible metadata endpoints at once, rather than iterating over them and blocking on each. In order to not do this twice, [my version](https://github.com/DataDog/datadog-agent/pull/39474/files) of this code changes `GetHostAliases` to return both the aliases and the cloudname, and then `GetCCRID` uses that cloudname to only lookup the CCRID for that exact cloudenv. What do you think?

PR #39503, opened 2025-08-04
[EBPF-773] uprobes: fix attachment when multiple rules match a target

gjulianm:
### What does this PR do?

This PR fixes the uprobe attachment process when multiple rules match a file. Previously, if any of the symbols didn't match, the attachment process would stop and no probes would be attached. With this change, rules that have all their symbols available will attach their probes, while the other rules won't get attached.

### Motivation

Fix attachment when multiple rules are active, specially for static binaries. For example, in GPUM we had rules to attach to CUDA functions and to Libc functions. If a binary was statically compiled against CUDA but dynamically against libc, it would fail to match all functions as the libc functions weren't available.

Also fixes https://datadoghq.atlassian.net/browse/EBPF-769

Added unit tests.

### Additional Notes

The GoTLS inspector does not support multiple rules. It is not a behavior change, as the GoTLS attacher doesn't use multiple rules. Additionally, this inspector doesn't support other features such as best-effort symbols either.


pkg/network/usm/ebpf_gotls_helpers.go
```diff
@@ -111,7 +117,7 @@ func (p *goTLSBinaryInspector) Inspect(fpath utils.FilePath, requests []uprobes.
 	elapsed := time.Since(start)
 	p.binAnalysisMetric.Add(elapsed.Milliseconds())
 
-	return inspectionResult.Functions, nil
+	return map[int]*uprobes.InspectionResult{0: {SymbolMap: inspectionResult.Functions}}, nil
```
guyarb, line 120: Why do we have map here and not an array?

gjulianm: I prefer to have a map so that we don't depend on the order, a "rule id" seems more robust to me

guyarb (approved):

val06 (approved):

PR #39506, opened 2025-08-04
dyninst/ebpf: add a trie for pointer tracking

ajwerner:
Before this change we just used a list. This meant that we needed to keep the number of pointers we could chase to be quite small because we didn't want quadratic behavior. Before this change, the limit on the number of pointers was 128 in total. Now that number has been lifted to 1024. The memory usage has gone up to an extra 4 bytes per pointer: from 1536 total to 16384. This is still quite efficient given what we're doing.

The structure can, with the same widths, be used for up to 2048 pointers which will be exactly 32KiB in size. This is handily the maximum size for a single per-cpu array item.

It will take some more clever tricks to extend the structure beyond 2048 items, but we can cross that bridge when we get there.

Fixes [DEBUG-4277](https://datadoghq.atlassian.net/browse/DEBUG-4277).

[DEBUG-4277]: https://datadoghq.atlassian.net/browse/DEBUG-4277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/dyninst/ebpf/chased_pointers_trie.h
```diff
@@ -0,0 +1,323 @@
...
+
+// Count leading zeros in 64-bit value.
+__attribute__((always_inline)) static inline uint8_t clz64(uint64_t x) {
+    uint8_t n = 64;
+    if (x) {
+        n = 0;
+        if (!(x & 0xFFFFFFFF00000000)) {
+            n += 32;
+            x <<= 32;
+        }
```
piob-io: Just call clz32 after one check? Less code, it should all get optimized since it's all inlined.

pkg/dyninst/ebpf/chased_pointers_trie.h
```diff
@@ -0,0 +1,323 @@
...
+        if (!(x & 0x80000000)) {
+            n += 1;
+        }
+    }
+    return n;
+}
+
+typedef uint8_t chased_pointers_trie_insert_result_t;
+#define CHASED_POINTERS_TRIE_EXISTS 0
```
piob-io: How about an enum instead?

pkg/dyninst/ebpf/chased_pointers_trie.h
```diff
@@ -0,0 +1,323 @@
...
+
+// the 12th bit is set.
+#define CPT_IS_LEAF(node) ((node) & CPT_LEAF_BIT)
+
+// Mask out leaf bit.
+#define CPT_NODE_MASK 0x7FF
+
+#ifdef CGO_TEST
+static
+#endif
```
grantseltzer: What does this accomplish?

pkg/dyninst/ebpf/chased_pointers_trie.h
```diff
@@ -0,0 +1,323 @@
...
+    void
+    chased_pointers_trie_init(chased_pointers_trie_t *trie) {
```
piob-io: Could this be on one line with no intendation:
void chased_pointers_trie_init(chased_pointers_trie_t *trie) {

piob-io (approved):
Neat!

pkg/dyninst/ebpf/chased_pointers_trie.h
ajwerner: Removed. I was having some cgo bindings issues before.

pkg/dyninst/ebpf/chased_pointers_trie.h
ajwerner: Done.

pkg/dyninst/ebpf/chased_pointers_trie.h
ajwerner: Done.

pkg/dyninst/ebpf/chased_pointers_trie.h
ajwerner: Done -- it was clang-format but I got rid of the weirdness

ajwerner:
The leaf nodes we're being packed as I wanted them to. They are now and I added assertions to make sure that's true. It's a little bit janky but I think it's worthwhile.

PR #39509, opened 2025-08-04
[ACIX-925] Update python buffering

CelianR:
### What does this PR do?

The gitlab sections were broken sometimes, set python buffer and dda interactive options to avoid that.

The issue was that the special characters were re-encoded by dda and that the buffering broke the two lines of the header in two.

### Additional Notes

[PoC](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1062805856)


Ishirui (approved):
🔥🔥🔥

Should we set these vars in the buildimages instead ? Probably just a matter of preference

PR #39513, opened 2025-08-04
feat(fleet): Add datadog-installer.exe as OCI layer

BaptisteFoy:
### What does this PR do?
Adds `datadog.exe` into its own layer of the Datadog Agent OCI package, similarly to Linux. This allows us to pull the installer that will do the upgrade directly from the OCI instead of extracting it from the MSI; potentially saving us a lot of time during upgrades.

E2E

new process tree shows `datadog-installer.exe` daemon running `installer.exe` directly instead of using `msiexec.exe` to extract it
| before | after |
| -- | -- |
| image | image |

network traffic reduced from ~480 MB to ~180 MB
| before | after |
| -- | -- |
| image | image | 

### Additional Notes
Need a follow-up investigation on why old code downloads the image three times instead of twice as expected.


rdesgroppes (approved):
LGTM

arbll (approved):

pkg/fleet/installer/bootstrap/bootstrap_windows.go
```diff
@@ -62,6 +62,28 @@ func downloadInstaller(ctx context.Context, env *env.Env, url string, tmpDir str
...
 	}
 
+	installerBinPath := filepath.Join(tmpDir, "installer.exe")
+	err = downloadedPackage.ExtractLayers(oci.DatadogPackageInstallerLayerMediaType, installerBinPath) // Returns nil if the layer doesn't exist
+	if err != nil {
+		return nil, fmt.Errorf("failed to extract layers: %w", err)
+	}
+	if _, err := os.Stat(installerBinPath); err != nil {
+		return downloadInstallerOld(ctx, env, url, tmpDir) // Fallback to the old method if the file doesn't exist
+	}
```
clarkb7: add comments and keep `datadog-installer.exe` name
```suggestion
	// Download just datadog-installer.exe from its own layer
	installerBinPath := filepath.Join(tmpDir, "datadog-installer.exe")
	err = downloadedPackage.ExtractLayers(oci.DatadogPackageInstallerLayerMediaType, installerBinPath) // Returns nil if the layer doesn't exist
	if err != nil {
		return nil, fmt.Errorf("failed to extract layers: %w", err)
	}
	if _, err := os.Stat(installerBinPath); err != nil {
		// Fallback to the old method if the file/layer doesn't exist
		// this is expected for versions earlier than 7.70
		return downloadInstallerOld(ctx, env, url, tmpDir) 
	}
```

pkg/fleet/installer/bootstrap/bootstrap_windows.go
```diff
@@ -62,6 +62,28 @@ func downloadInstaller(ctx context.Context, env *env.Env, url string, tmpDir str
...
+	return iexec.NewInstallerExec(env, installerBinPath), nil
+}
+
+// downloadInstallerOld downloads the installer package from the registry and returns the path to the executable.
+func downloadInstallerOld(ctx context.Context, env *env.Env, url string, tmpDir string) (*iexec.InstallerExec, error) {
```
clarkb7:
```suggestion
// downloadInstallerOld downloads the installer package from the registry and returns the path to the executable.
//
// Should only be called for versions earlier than 7.70. This downloads the layer containing the MSI and then
// uses MSI admin install to extract `datadog-installer.exe` from the MSI.
func downloadInstallerOld(ctx context.Context, env *env.Env, url string, tmpDir string) (*iexec.InstallerExec, error) {
```

clarkb7 (approved):

PR #39519, opened 2025-08-04
[AGENTCFG-326] Fetching the AWS Host CCRID

rahulkaukuntla:
### What does this PR do?

We need to fetch the ARN of an EC2 Resource to help with [this RFC](https://docs.google.com/document/d/18v9yDZpUWxl6y5Vvi2Z08j_yYwxzNEjgK4OHclulTMc/edit?pli=1&tab=t.9h3c53hm6v9v#heading=h.vrc32trb5vis). However, there is no api that just gives us the ARN, so we have to construct it by querying for [instance metadata](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html).

I added a unit test--you can manually QA this by running the agent in a EC2 instance, and then emitting a log statement for the payload [here](https://github.com/DataDog/datadog-agent/blob/5fee599a6ac41d23fc79fdef312aa90b7a62f4d9/comp/metadata/host/hostimpl/utils/host.go#L208)


pkg/config/setup/config.go
```diff
@@ -289,6 +289,7 @@ func InitConfig(config pkgconfigmodel.Setup) {
 	config.BindEnvAndSetDefault("check_sampler_expire_metrics", true)
 	config.BindEnvAndSetDefault("check_sampler_context_metrics", false)
 	config.BindEnvAndSetDefault("host_aliases", []string{})
+	config.BindEnvAndSetDefault("host_disable_ccrid_fetch", false)
```
GustavoCaso: Can we configure it to be `host_ccrid_fetch` instead of being a negation?

rahulkaukuntla: Dustin decided to go with `collect_ccrid`, so I have cherry-picked his change.

comp/metadata/host/hostimpl/utils/meta.go
```diff
@@ -65,15 +66,22 @@ func getMeta(ctx context.Context, conf model.Reader, hostnameComp hostnameinterf
 		agentHostname = hostnameData.Hostname
 	}
 
+	ccrid := ""
+	hostAliases, cloudname := cloudproviders.GetHostAliases(ctx)
+	if !conf.GetBool("collect_ccrid") {
```
GustavoCaso: By default, the value of  `collect_ccrid` is `false`, so the condition is evaluating to `true`. I think what we want is:
```
if conf.GetBool("collect_ccrid") {
....
```

pkg/util/cloudproviders/azure/azure.go
```diff
@@ -174,6 +174,30 @@ func getHostnameWithConfig(ctx context.Context, config model.Config) (string, er
...
+			metadataURL+"/metadata/instance/compute/resourceId?api-version=2021-02-01&format=text")
+		if err != nil {
+			return "", fmt.Errorf("unable to query metadata endpoint: %s", err)
+		}
+		return rg, nil
+	},
+}
+
+// GetHostCCRID returns the Canonical Cloud Resource ID for the Azure host
+func GetHostCCRID(ctx context.Context) (string, error) {
```
GustavoCaso: We should add a test for this new function

pkg/util/cloudproviders/cloudproviders.go
```diff
@@ -138,13 +139,39 @@ func GetHostAliases(ctx context.Context) []string {
...
+
+type cloudProviderCCRIDDetector func(context.Context) (string, error)
+
+var hostCCRIDDetectors = map[string]cloudProviderCCRIDDetector{
+	azure.CloudProviderName: azure.GetHostCCRID,
+	ec2.CloudProviderName:   ec2.GetHostCCRID,
+}
+
+// GetHostCCRID returns the host CCRID from the first provider that works
+func GetHostCCRID(ctx context.Context, cloudname string) string {
```
GustavoCaso: We should a test for this function as well 😄

pkg/util/ec2/ccrid_fetch_test.go
```diff
@@ -0,0 +1,44 @@
...
+// Copyright 2016-present Datadog, Inc.
+
+package ec2
+
+import (
+	"context"
+	"testing"
+)
+
+func TestGetHostCCRID(t *testing.T) {
```
GustavoCaso, line 13: 👍

GustavoCaso (commented):
Code LGTM. We are missing a few tests. Once the test are added we can merge 🚀

pkg/util/ec2/ccrid_fetch.go
```diff
@@ -0,0 +1,107 @@
...
+		return "", err
+	}
+	s, ok := val.(string)
+	if !ok {
+		return "", fmt.Errorf("%s returned non-string", f.Name)
+	}
+	return s, nil
+}
+
+var regionFetcher = fetcher{
```
dustmop: Can you change this to use `cachedfetch.Fetcher` like is done [here](https://github.com/DataDog/datadog-agent/blob/97ff31827a34dee7ddbc959761929bdc24a3e0fd/pkg/util/ec2/ec2.go#L56)? That way, if the endpoint fails sometime, the previously retrieved value will be returned.

dustmop (changes requested):

pkg/util/cloudproviders/azure/azure.go
rahulkaukuntla: Since these are related to the changes that were made to Dustin's PR, he said that he'll do them in a [separate PR](https://github.com/DataDog/datadog-agent/pull/39636).

pkg/util/cloudproviders/cloudproviders.go
rahulkaukuntla: Since these are related to the changes that were made to Dustin's PR, he said that he'll do them in a [separate PR](https://github.com/DataDog/datadog-agent/pull/39636).

pkg/util/ec2/ccrid_fetch.go
rahulkaukuntla: Done [here](https://github.com/DataDog/datadog-agent/pull/39519/commits/93b4e1e448c7faa0978f5abb6f016daa1a6779e6)

pkg/util/cloudproviders/azure/azure.go
dustmop: Tests being added here: https://github.com/DataDog/datadog-agent/pull/39636

dustmop (approved):
Looks good to me, thanks!

jeremy-hanna (approved):

PR #39521, opened 2025-08-04
ignore top level bazel symlinks

aiuto:
### What does this PR do?

Ignores bazel top level symlinks.


git-thuerk-done (approved):

ofek (approved):
Do these refer to actual symbolic links? Asking because I'm curious about what happens when that feature is not enabled on Windows machines (as it is often not).

rdesgroppes (approved):

aiuto:
> Do these refer to actual symbolic links? Asking because I'm curious about what happens when that feature is not enabled on Windows machines (as it is often not).

They are symlinks on macos and linux.  Bazel on windows probably requires symlinks enabled.
But it does not matter w.r.t. ignoring this.  You still want git to ignore the bazel-* folders, symlink or real.

aiuto:
Droping this because of my unsigned commit problem

PR #39522, opened 2025-08-04
[OTAGENT-486] Support for DDOT receiving remote config changes. 

tim-liakh-datadog:
### What does this PR do?
The purpose of this PR is to prepare for DDOT remote config changes. It does this by:

1. Accepting and expecting otel-config.yaml to get passed down from remote
2. Adding the otel-config.yaml as an additional config within otel.


cmd/otel-agent/subcommands/run/command.go
```diff
@@ -122,7 +124,22 @@ func runOTelAgentCommand(ctx context.Context, params *cliParams, opts ...fx.Opti
...
+
+	// Add fleet policy config if DD_FLEET_POLICIES_DIR is set
+	if fleetPoliciesDir := os.Getenv("DD_FLEET_POLICIES_DIR"); fleetPoliciesDir != "" {
+		fleetConfigPath := filepath.Join(fleetPoliciesDir, "otel-config.yaml")
+		// Check if the fleet config file exists before adding it
+		if _, err := os.Stat(fleetConfigPath); err == nil {
+			uris = append(uris, "file:"+fleetConfigPath)
+			fmt.Printf("Using fleet policy config: %s\n", fleetConfigPath)
+		}
+	}
```
BaptisteFoy: nit: you should probably make sure it's an absolute path & resolve it (it'll be a symlink)

cmd/otel-agent/subcommands/run/command.go
```diff
@@ -122,7 +124,33 @@ func runOTelAgentCommand(ctx context.Context, params *cliParams, opts ...fx.Opti
...
+	if fleetPoliciesDir := os.Getenv("DD_FLEET_POLICIES_DIR"); fleetPoliciesDir != "" {
+		absFleetPoliciesDir, err := filepath.Abs(fleetPoliciesDir)
+		if err != nil {
+			fmt.Printf("Warning: failed to get absolute path for fleet policies dir %s: %v\n", fleetPoliciesDir, err)
+		} else {
+			resolvedFleetPoliciesDir, err := filepath.EvalSymlinks(absFleetPoliciesDir)
+			if err != nil {
+				fmt.Printf("Warning: failed to resolve symlinks for fleet policies dir %s: %v\n", absFleetPoliciesDir, err)
+				resolvedFleetPoliciesDir = absFleetPoliciesDir
+			}
```
BaptisteFoy: EvalSymlinks unfortunately returns a relative path if the symlink has been declared as relative. I would evaluate symlinks first then get the absolute path.

Also, I would recommend having distinct behaviours if the error is `os.ErrNotExists` or not, as the former is expected behaviour

cmd/otel-agent/subcommands/run/command.go
```diff
@@ -122,7 +124,33 @@ func runOTelAgentCommand(ctx context.Context, params *cliParams, opts ...fx.Opti
...
-	uris := append(params.ConfPaths, params.Sets...)
+
+	// Apply overrides
+	uris := append([]string{}, params.ConfPaths...)
+
+	// Add fleet policy config if DD_FLEET_POLICIES_DIR is set
+	if fleetPoliciesDir := os.Getenv("DD_FLEET_POLICIES_DIR"); fleetPoliciesDir != "" {
+		absFleetPoliciesDir, err := filepath.Abs(fleetPoliciesDir)
+		if err != nil {
+			fmt.Printf("Warning: failed to get absolute path for fleet policies dir %s: %v\n", fleetPoliciesDir, err)
```
BaptisteFoy: For my own understanding: is it possible to use a logger here?

pkg/fleet/daemon/remote_config.go
```diff
@@ -125,6 +126,9 @@ func handleInstallerConfigUpdate(h handleConfigsUpdate) func(map[string]state.Ra
 			if len(legacyConfigs.Configs.APMLibrariesYAML) > 0 {
 				installerConfig.Files = append(installerConfig.Files, installerConfigFile{Path: "/application_monitoring.yaml", Contents: legacyConfigs.Configs.APMLibrariesYAML})
 			}
+			if len(legacyConfigs.Configs.OtelConfigYAML) > 0 {
```
BaptisteFoy: I believe you need to add otel-config.yaml to `allowedConfigFiles` too!

cmd/otel-agent/subcommands/run/command.go
tim-liakh-datadog: I am not fully sure, but it should be possible. From the lines above I saw fmt.Println getting used so I followed the standard.

cmd/otel-agent/subcommands/run/command.go
```diff
@@ -122,7 +124,41 @@ func runOTelAgentCommand(ctx context.Context, params *cliParams, opts ...fx.Opti
...
+			fleetConfigPath := filepath.Join(absFleetPoliciesDir, "otel-config.yaml")
+
+			if _, err := os.Stat(fleetConfigPath); err == nil {
+				uris = append(uris, "file:"+fleetConfigPath)
+				fmt.Printf("Using fleet policy config: %s\n", fleetConfigPath)
+			} else if !os.IsNotExist(err) {
+				fmt.Printf("Warning: failed to access fleet policy config %s: %v\n", fleetConfigPath, err)
+			}
+		}
+	}
```
liustanley: Could you move this logic into a helper function?

cmd/otel-agent/subcommands/run/command.go
```diff
@@ -122,7 +124,41 @@ func runOTelAgentCommand(ctx context.Context, params *cliParams, opts ...fx.Opti
...
+				fmt.Printf("Warning: failed to resolve symlinks for fleet policies dir %s: %v\n", fleetPoliciesDir, err)
+			}
+			resolvedFleetPoliciesDir = fleetPoliciesDir
+		}
+
+		// Make it absolute
+		absFleetPoliciesDir, err := filepath.Abs(resolvedFleetPoliciesDir)
+		if err != nil {
+			fmt.Printf("Warning: failed to get absolute path for fleet policies dir %s: %v\n", resolvedFleetPoliciesDir, err)
+		} else {
```
liustanley: Once this is in a helper function, you can add a return in the `if err != nil` case and remove this else.

cmd/otel-agent/subcommands/run/command.go
```diff
@@ -122,7 +124,41 @@ func runOTelAgentCommand(ctx context.Context, params *cliParams, opts ...fx.Opti
...
+			fleetConfigPath := filepath.Join(absFleetPoliciesDir, "otel-config.yaml")
+
+			if _, err := os.Stat(fleetConfigPath); err == nil {
+				uris = append(uris, "file:"+fleetConfigPath)
+				fmt.Printf("Using fleet policy config: %s\n", fleetConfigPath)
+			} else if !os.IsNotExist(err) {
+				fmt.Printf("Warning: failed to access fleet policy config %s: %v\n", fleetConfigPath, err)
+			}
```
liustanley:
```suggestion
			_, err := os.Stat(fleetConfigPath)
			if err != nil && !os.IsNotExist(err) {
				fmt.Printf("Warning: failed to access fleet policy config %s: %v\n", fleetConfigPath, err)
			}
			if err == nil {
				uris = append(uris, "file:"+fleetConfigPath)
				fmt.Printf("Using fleet policy config: %s\n", fleetConfigPath)
			}
```
Avoid using `else if` when possible

pkg/fleet/daemon/remote_config.go
```diff
@@ -105,6 +105,7 @@ func handleInstallerConfigUpdate(h handleConfigsUpdate) func(map[string]state.Ra
 					SecurityAgentYAML json.RawMessage `json:"security-agent.yaml,omitempty"`
 					SystemProbeYAML   json.RawMessage `json:"system-probe.yaml,omitempty"`
 					APMLibrariesYAML  json.RawMessage `json:"application_monitoring.yaml,omitempty"`
+					OtelConfigYAML    json.RawMessage `json:"otel-config.yaml,omitempty"`
```
liustanley: nit: Change to "OTelConfigYAML"

cmd/otel-agent/subcommands/run/command.go
liustanley: It isn't possible to use a logger here because this point in the function is before we get the logger from fx.

liustanley (commented):
Mostly looks good to me, just some comments on formatting

BaptisteFoy (approved):
LGTM from a functional point of view

liustanley (approved):
LGTM, you can add Milestone 7.70 if you merge today

PR #39525, opened 2025-08-05
dyninst/symdb: fix handling of types in packages with dots

andreimatei:
Fix a bug where types from packages containing a dot in the component of
the import path where added to the wrong package, or sometimes not added
at all. We were using the unescaped import patch when parsing out the
type's package, which made the regex we use for parsing incorrect. Now
we use unescaping utilities copied from the compiler, which are more
robust.

Also apply unescaping to the type names of variables; before, we were
producing variables with escaped type names in the output.

### DIFF
--- a/pkg/dyninst/symdb/symdb.go
+++ b/pkg/dyninst/symdb/symdb.go
@@ -13,7 +13,6 @@ import (
 	"errors"
 	"fmt"
 	"math"
-	"regexp"
 	"runtime/debug"
 	"sort"
 	"strings"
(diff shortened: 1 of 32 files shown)


ajwerner:
reading https://github.com/golang/go/blob/7a1679d7ae32dd8a01bd355413ee77ba517f5f43/src/cmd/compile/internal/ir/func.go#L334-L393 and friends, it seems like you've fixated too much on just `.`. There's other characters that get escaped. May as well just do the right thing in general, no? Also, in case there are lots and lots of escapes, no need to do the string concats -- may as well just build the string in a byte slice and convert.

ajwerner:
> The period needs escaping only in the  last segment of the path, and it makes for happier users if we escape that as little as possible.

Interesting! https://github.com/golang/go/blob/7a1679d7ae32dd8a01bd355413ee77ba517f5f43/src/cmd/internal/objabi/path.go#L13-L17

pkg/dyninst/symdb/symdb.go
```diff
@@ -423,26 +423,29 @@ func (c *typesCollection) getType(name string) *Type {
...
-		// escaping so that our cache key matches the actual package name.
-		escapedDot := "%2e"
-		i := strings.LastIndex(name, escapedDot)
-		if i >= 0 {
-			// Replace %2e with '.' in the type name. This is how DWARF encodes
-			// dots in package names.
-			name = name[:i] + "." + name[i+len(escapedDot):]
-		}
-	}
+	sanitizedName := unescapePackageName(t.Common().Name)
```
piob-io: Strictly speaking, isn't this an unsanitizedName?

I think it's worth reusing the same adjectives. Now we have unescape function, and then sanitized and raw variables. How about we call the variables unescapedName and name, or name and escapedName? Or we call the function sanitizePackageName (but I think it's the other direction).

piob-io (approved):

pkg/dyninst/symdb/symdb.go
andreimatei: switched to `unescapedNamed`

andreimatei:
switched to unescaping utilities from the compiler

PR #39526, opened 2025-08-05
dyninst/symdb: introduce a streaming api

andreimatei:
Introduce an iterator over a binary's packages so that the symbols can
be streamed out without accumulating them all in memory. For now this
implies configuring the extraction process to ignore abstract functions
and inlined instances, since they can appear in compile units
corresponding to other packages.

Even when using this streaming API, there is still a types cache that
accumulates types across packages (populated when resolving variables).
We'll see if that's a problem for memory use.

### DIFF
--- a/pkg/dyninst/symdb/symdb.go
+++ b/pkg/dyninst/symdb/symdb.go
@@ -12,6 +12,7 @@ import (
 	"debug/dwarf"
 	"errors"
 	"fmt"
+	"iter"
 	"math"
 	"runtime/debug"
 	"slices"
@@ -310,8 +311,8 @@ type SymDBBuilder struct {
 	loclistReader *loclist.Reader
 	// The size of pointers for the binary's architecture, in bytes.
 	pointerSize int
-	// Filtering mode.
-	scopeFilter ExtractScope
+	options     ExtractOptions
+
 	// The module path of the Go module containing the main function. Empty if
 	// unknown.
 	mainModule string
@@ -323,10 +324,15 @@ type SymDBBuilder struct {
 
 	// The compile unit currently being processed by explore* functions.
 	currentCompileUnit        *dwarf.Entry
+	currentCompileUnitName    string
 	filesInCurrentCompileUnit []string
 
 	abstractFunctions map[dwarf.Offset]*abstractFunction
-	types             typesCollection
+
+	// typesCache will accumulate types as we look them up to resolve variables
+	// and functions. The cache is indexed by DWARF offset.
+	typesCache *dwarfutils.TypeFinder
+	types      typesCollection
 
 	// Stack of blocks currently being explored. Variable location lists can
 	// make references to the current block and its PC ranges; they will look at
@@ -367,10 +373,6 @@ type typesCollection struct {
 	mainModule          string
 	firstPartyPkgPrefix string
 
-	// typesCache will accumulate types as we look them up to resolve variables
-	// and functions. The cache is indexed by DWARF offset.
-	typesCache *dwarfutils.TypeFinder
-
 	// Map from the type's name, as it appears in DWARF, to the type info. Only
 	// some of the types from typesCache are to be exported to SymDB and thus
 	// are present here -- we ignore some types we don't support, and we
@@ -381,8 +383,8 @@ type typesCollection struct {
 	packages map[string][]*Type
 }
 
-func (c *typesCollection) resolveType(offset dwarf.Offset) (typeInfo, error) {
-	typ, err := c.typesCache.FindTypeByOffset(offset)
+func (b *SymDBBuilder) resolveType(offset dwarf.Offset) (typeInfo, error) {
+	typ, err := b.typesCache.FindTypeByOffset(offset)
 	if err != nil {
 		return typeInfo{}, err
 	}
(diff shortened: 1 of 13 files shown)


pkg/dyninst/symdb/symdb.go
```diff
@@ -309,9 +310,9 @@ type SymDBBuilder struct {
 	// The location list reader used to read location lists for variables.
 	loclistReader *loclist.Reader
 	// The size of pointers for the binary's architecture, in bytes.
-	pointerSize int
-	// Filtering mode.
-	scopeFilter ExtractScope
+	pointerSize    int
+	extractOptions ExtractOptions
```
piob-io: Maybe just "options"?

pkg/dyninst/symdb/symdb.go
```diff
@@ -439,9 +449,10 @@ func unescapePackageName(s string) string {
 	return s
 }
 
-// addType adds a type to the collection if it is not already present.
-// Unsupported types are ignored and no error is returned.
-func (c *typesCollection) addType(t godwarf.Type) error {
+// addTypeMaybe adds a type to the collection if it is not already present.
+// Unsupported types are ignored and no error is returned. If pkgFilter is not
+// empty, then only types from that package are added.
+func (c *typesCollection) addTypeMaybe(t godwarf.Type, pkgFilter string) error {
```
piob-io: maybeAddType?

pkg/dyninst/symdb/symdb.go
```diff
@@ -755,23 +759,119 @@ func (b *SymDBBuilder) ExtractSymbols() (Symbols, error) {
...
+		return func(_yield func(pkg Package) bool) {},
+			func() error {
+				return fmt.Errorf("cannot overate over packages when IncludeInlinedFunctions is set")
+			}
+	}
+	return b.packagesIteratorInner()
+}
+
+func (b *SymDBBuilder) packagesIteratorInner() (iter.Seq[Package], func() error) {
+	// err is captured by the returned error-checking function, and is assigned
```
piob-io: Could we return iter.Seq2[Package, Error] instead, forcing more explicit handling of the error?

pkg/dyninst/symdb/symdb.go
```diff
@@ -755,23 +759,119 @@ func (b *SymDBBuilder) ExtractSymbols() (Symbols, error) {
...
+	if b.extractOptions.IncludeInlinedFunctions {
+		return func(_yield func(pkg Package) bool) {},
+			func() error {
+				return fmt.Errorf("cannot overate over packages when IncludeInlinedFunctions is set")
+			}
+	}
+	return b.packagesIteratorInner()
+}
+
+func (b *SymDBBuilder) packagesIteratorInner() (iter.Seq[Package], func() error) {
```
piob-io: I really like the iterator approach. What I think it could further help us with is - we have these variables in the builder, that are closely tied to the "exploration", and that must be correctly updated, and are implicitly depended on from "explore" functions - like "current package". Now that we have an iterator with state, could we move them in here, and pass them more explicitly?

No need to address this in this pr, just hoping for better future.

pkg/dyninst/symdb/symdb.go
```diff
@@ -755,23 +759,119 @@ func (b *SymDBBuilder) ExtractSymbols() (Symbols, error) {
...
+			var pkg Package
+			pkg, err = b.exploreCompileUnit(entry, entryReader)
+			if err != nil {
+				break
+			}
+			if pkg.Name == "" {
+				continue
+			}
+
+			// If we're not dealing with inlined functions, then move all
```
piob-io, line 840: I feel like it would be clearer if dealing with accumulation would be kept closer together. Could we move this part to the caller - ExtractSymbols ?

piob-io (approved):

pkg/dyninst/symdb/symdb.go
andreimatei: done

pkg/dyninst/symdb/symdb.go
andreimatei: done

pkg/dyninst/symdb/symdb.go
andreimatei: done

pkg/dyninst/symdb/symdb.go
andreimatei: not really, as discussed

pkg/dyninst/symdb/symdb.go
andreimatei: I appended a commit changing the API.
Now the exploration functions are exposed as free-standing function, and the struct with the iteration state is unexported. See if you like.

piob-io: It's really nice. I was thinking about simplifying the state from the other end (from the inside), but this helps with the complexity too. The currentCompileUnit, currentCompileUnitName, filesInCurrentCompileUnit and blockStack are things that are valid within certain execution scope (while exploring specific package or block). But they are stored within the current iterator object, which does more, and calling explore functions "out of bound", without taking care for these variables can fail in a quite weird way. package-scoped things I think could live inside some object that is valid only during single package iterator step. Blocks could maybe be stored on the stack itself? This is not a blocking issue.

PR #39530, opened 2025-08-05
[CONTP-849] Support cluster agent leader re-election in E2E test

zhuminyi:
### What does this PR do?
Add cluster agent leader re-election to eks test. 
The test will delete cluster-agent leader pod and re-elect a new leader.

### DIFF
--- a/test/new-e2e/tests/containers/eks_test.go
+++ b/test/new-e2e/tests/containers/eks_test.go
@@ -3,6 +3,12 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2023-present Datadog, Inc.
 
+// EKS-specific e2e tests focusing on Fargate and AWS-specific Kubernetes features.
+// These tests should only depend on:
+// - Node agents (for container/pod metrics and logs)
+// - Application workloads (nginx, dogstatsd, etc.)
+// - EKS-specific features (Fargate, etc.)
+
 package containers
 
 import (
--- a/test/new-e2e/tests/containers/k8s_cluster_agent_test.go
+++ b/test/new-e2e/tests/containers/k8s_cluster_agent_test.go
@@ -1,0 +1,116 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2023-present Datadog, Inc.
+
+// This test suite contains cluster agent specific tests that require isolation.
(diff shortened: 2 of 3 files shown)


test/new-e2e/tests/containers/eks_test.go
```diff
@@ -261,3 +266,30 @@ func (suite *eksSuite) TestNginxFargate() {
...
+// TestClusterAgentLeaderElectionWithRestart tests cluster agent leader election by restarting the leader
+// This test is safe to run in eks_test.go because the other EKS tests don't depend on the cluster agent
+// (they focus on node-level metrics and logs, which are handled by node agents)
+func (suite *eksSuite) TestClusterAgentLeaderElectionWithRestart() {
+	// Get the current leader
+	originalLeaderPodName := suite.testDCALeaderElection(false)
+	suite.Require().NotEmpty(originalLeaderPodName, "Original leader DCA pod name should not be empty")
+	suite.T().Logf("Current cluster agent leader: %s", originalLeaderPodName)
+
+	// Restart the leader pod by deleting it
```
sblumenthal: nit: "Restart" here is a bit misleading, maybe replace with "reshuffle"? If so, same with the test name

I understand the intention of the test, and the use of the wording, but "restarting a pod" by deleting it is the kind of thing that people will see and do a double take on

sblumenthal: Or even "re-elect" would probably be the best wording to use?

gabedos (approved):
LGTM, could be worth adding a warning to the top of the `eks_test.go` file to increase visibility of the constriction you added here stating that any DCA dependent tests should not be added in this scenario due to increased flakiness resulting from restarting cluster-agents.

zhuminyi:
> LGTM, could be worth adding a warning to the top of the `eks_test.go` file to increase visibility of the constriction you added here stating that any DCA dependent tests should not be added in this scenario due to increased flakiness resulting from restarting cluster-agents.

Good idea, warning added

test/new-e2e/tests/containers/eks_test.go
```diff
@@ -3,18 +3,35 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2023-present Datadog, Inc.
 
+// This test suite contains TestClusterAgentLeaderElectionWithReelection which
```
davidor: I'm a bit confused about this.

eksSuite embeds `k8sSuite` https://github.com/DataDog/datadog-agent/pull/39530/files#diff-bcf487461201693f75440b908a989a0507d1fb5323b94aec544b2a5e66ef876aL21

So in EKS we also run the tests there, right?

PR #32877, opened 2025-01-10
[CONTINT-4506] add resource labels and annotations as tags to orchestrator

ewoodthomas:
### What does this PR do?

Adds Kubernetes resource labels and annotations as tags in the orchestrator check when configured with `kubernetesResourcesLabelsAsTags` and `kubernetesResourcesAnnotationsAsTags`. Improves tagging consistency.

### Motivation

Support for `kubernetesResourcesLabelsAsTags` and `kubernetesResourcesAnnotationsAsTags` has already been released in official documentation. However, the orchestrator check does not emit these tags.

### Describe how you validated your changes

Added unit tests that cover tag addition for every k8s resource type the orchestrator check supports.

Manually tested:

1. Deploy the agent with the following added to the config:
```
datadog:
  ...
  kubernetesResourcesLabelsAsTags:
    clusterroles.rbac.authorization.k8s.io:
      foo1: bar1
  kubernetesResourcesAnnotationsAsTags:
    clusterroles.rbac.authorization.k8s.io:
      foo2: bar2
```

2. Deploy a clusterrole with the following config:
```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: example-cluster-role
  labels:
    role-type: management
    team: devops
    environment: production
    foo1: example-value
  annotations:
    foo2: example-anno
rules:
  - apiGroups: [""]
    resources: ["pods"]
    verbs: ["list", "get"]
```

3. Check the orchestrator explorer and verify the tags `bar1:example-value` and `bar2:example-anno` are present for `example-cluster-role`

I did not manually test every resource, but unit tests alone validate the changes.

### Additional Notes

The changes for this could likely be limited to the transformers. We would just need to call some initialization when starting the orchestrator check to `GetMetadataTags` to avoid getting it every time a resource is extracted.

The main driver for the changes to the processors and collectors is to easily allow support for CRs and CRDs in the future.  To know the resource type, we need the `name` and `version` passed into the collector. Hence, we need to change the collectors to pass this information down to the Processor.

Feel free to suggest other ways of going about this, I've gone back and forth between a few different ideas.

PR #39541, opened 2025-08-05
CAP-2788 fix `labels as tags` and `annotations as tags` for pod collection

kangyili:
### What does this PR do?

This PR fixes  an oversight of https://github.com/DataDog/datadog-agent/pull/32877 . We updated orchestrator check in cluster agent but we didn’t update it for the pod check in node agent

### QA

I have datadog agent with env
```
agents:
  containers:
    agent:
      env:
        - name: DD_KUBERNETES_POD_LABELS_AS_TAGS
          value: '{"dagster/code-location":"dagster.code_location","dagster/job":"dagster.job_name"}'
```

I have a cronjob
```
apiVersion: batch/v1
kind: CronJob
metadata:
  name: example-cronjob
spec:
  schedule: "* * * * *" 
  jobTemplate:
    spec:
      ttlSecondsAfterFinished: 1800 
      template:
        metadata:
          labels:
            dagster/code-location: "test"
            dagster/job: "test-job"
        spec:
          restartPolicy: Never
          containers:
          - name: example
            image: busybox
            command: ["sh", "-c", "echo Starting job; sleep 60; echo Job done"]
```
On completed pods, I can find tags `dagster.code_location` and `dagster.job_name`

Screenshot 2025-08-05 at 15 50 06


pkg/collector/corechecks/orchestrator/pod/pod.go
```diff
@@ -167,6 +168,9 @@ func (c *Check) Run() error {
 	}
 
 	groupID := nextGroupID()
+	metadataAsTags := utils.GetMetadataAsTags(c.cfg)
```
mbotarro, line 172: Should we use `podName` and `podVersion` constants as in the cluster check?
https://github.com/DataDog/datadog-agent/blob/453fc36fa12026e4738cc64d6fb804fed1cf22df/pkg/collector/corechecks/cluster/orchestrator/collectors/k8s/pod_unassigned.go#L46

This way we don't need to hardcode `"pods"` and avoid future regressions  if `getResourceType` or `GetResourceAnnotations` changes.

mbotarro (approved):
Left a minor nit. Once it's addressed we should be good to merge.

pkg/collector/corechecks/orchestrator/pod/pod.go
kangyili: thank you for the review 
https://github.com/DataDog/datadog-agent/commit/2afc00820bcbff10a00f189fdb6ee2c4f62b3edf

mbotarro: I see why you not used it in the first place. The constants defined [here](https://github.com/DataDog/datadog-agent/blob/634cc25dead676d817f0dd00ca2280949b2abda7/pkg/collector/corechecks/cluster/orchestrator/collectors/k8s/types.go#L15) are private to the `cluster` checks package.

I don't think it's a good idea to copy/paste them to the `corechecks` because we'll have a duplicated version that can diverge over time.

For this reason, my suggestion is:
* Let's revert back to use your previous version with hardcoded `"pods"` to unblock this PR
* Let's plan to move the `cluster/orchestrator/collector/k8s/types.go` to a util package that can be used by both cluster and node checks. This can be done in a separate PR.

Wdyt?

pkg/collector/corechecks/cluster/orchestrator/util/tags_test.go
```diff
@@ -3,7 +3,7 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2016-present Datadog, Inc.
 
-//go:build kubeapiserver && orchestrator
```
mbotarro, line 6: For my understanding, what is the impact of removing `kubeapiserver` here?

pkg/collector/corechecks/orchestrator/pod/pod.go
mbotarro: Never mind, I just realised you already did both in this PR.

pkg/collector/corechecks/orchestrator/pod/pod_test.go
```diff
@@ -193,14 +196,18 @@ func (suite *PodTestSuite) TestPodCheck() {
 		sorted("kube_api_version:v1"))
 	require.Equal(suite.T(),
 		sorted(suite.sender.pods[0].(*process.CollectorPod).Pods[0].Tags...),
-		sorted("kube_condition_podscheduled:true", "pod_status:pending"))
+		sorted("kube_condition_podscheduled:true", "pod_status:pending",
+			"dd_component:kube-proxy", "dd_tier:node",
+			"config_hash:260c2b1d43b094af6d6b4ccba082c2db", "config_source:file"))
```
mbotarro, line 201: 💯

mbotarro (approved):

pkg/collector/corechecks/cluster/orchestrator/util/tags_test.go
kangyili: This is a simple utility package with no dependencies on the kubeapiserver, such as informers . It's shared between both the pod check and cluster check.

The cluster check requires  `kubeapiserver && orchestrator`, while the pod check requires `kubelet &&orchestrator`. So I think it makes sense for the shared package to include only the 'orchestrator' build tag.

mbotarro: Makes sense. Thanks for clarifying it.

PR #39413, opened 2025-07-31
[AMCC-85] dogstatsd: emit a signal on control reconfiguration.

remeh:
### What does this PR do?

Add some telemetry on dogstatsd control subsystem reconfiguration.

### Motivation

Easier monitoring on the feature in graphs.

### Describe how you validated your changes

Reconfigured the Agent and validated the values.

PR #39542, opened 2025-08-05
Revert "[AMCC-85] dogstatsd: emit a signal on control reconfiguration."

alopezz:
Reverts DataDog/datadog-agent#39413

Attempting a revert since this PR violated the gates when it landed on main.


alopezz:
Not needed anymore, we brought the disk size down via https://github.com/DataDog/datadog-agent/pull/39545.

PR #39545, opened 2025-08-05
Don't ships systemd headers

alopezz:
### What does this PR do?

It excludes systemd headers from final Agent (including iot-agent) packages.

### Motivation

#incident-41581

These files are only used for building components of the Agent, and customers are not meant to use them in any way. Therefore, they're dead weight that we can remove.

CI passing, size reduction, manual check that the files are no longer present.


JSGette (approved):
LGTM!

PR #39549, opened 2025-08-05
[NDINT-290] Add QoS, DIA, and Site Metrics

ken-schneider:
### What does this PR do?
Adds QoS, DIA, Site metrics to the Versa integration.

For QoS metrics are collected by the various forwarding classes (by type, not individual FC) as well as in aggregate.

For each class we collect:
- Volume Sent (bytes)
- Volume Dropped (bytes)
- Bandwidth (bytes per second)

For aggregated totals we collect:
- Volume Sent (bytes)
- Volume Dropped (bytes)
- Bandwidth Sent (bytes per second)
- Percentage dropped (%)

For DIA metrics we collect:
- Volume Sent/Received (bytes)
- Bandwidth Sent/Received (bytes per second)

For Site metrics we collect:
- Availability (percentage)
- Volume Sent/Received (bytes)
- Bandwidth Sent/Received (bytes per second)

### Motivation
Add QoS, DIA, and Site support for the Versa integration

Tested against our Versa environment as well as unit tests based on other analytics metrics structures


pkg/collector/corechecks/network-devices/versa/client/client.go
```diff
@@ -476,6 +514,27 @@ func (client *Client) GetLinkUsageMetrics(tenant string) ([]LinkUsageMetrics, er
 	)
 }
 
+// GetSiteMetrics gets site metrics for a Versa tenant
+func (client *Client) GetSiteMetrics(tenant string) ([]SiteMetrics, error) {
+	return getPaginatedAnalytics(
```
zoedt, line 519: very nice !!

pkg/collector/corechecks/network-devices/versa/client/parser.go
```diff
@@ -174,6 +174,48 @@ func parseLinkUsageMetrics(data [][]interface{}) ([]LinkUsageMetrics, error) {
 	return rows, nil
 }
 
+// parseSiteMetrics parses the raw AaData response into SiteMetrics structs
+func parseSiteMetrics(data [][]interface{}) ([]SiteMetrics, error) {
+	var rows []SiteMetrics
+	for _, row := range data {
+		m := SiteMetrics{}
+		if len(row) < 10 {
```
zoedt: should we keep consistency for `<` or `!=` for this length check? (below have `!=` but maybe it only really matters per call in the end)

pkg/collector/corechecks/network-devices/versa/client/client.go
```diff
@@ -538,18 +603,48 @@ func (client *Client) GetTunnelMetrics(tenant string) ([]TunnelMetrics, error) {
...
+			"volume-rx",
+			"bw-tx",
+			"bw-rx",
+		},
+		parseDIAMetrics,
+	)
+}
+
 // buildAnalyticsPath constructs a Versa Analytics query path in a cleaner way so multiple metrics can be added.
+// TODO: maybe this becomes a struct function. Modifications will be easier and the function signature will be
```
zoedt, line 631: i like this idea and was thinking it would be nice to see the fields modified w/ how many different values can be for each one and getting lost in all of them !

zoedt (approved):
overall lgtm just tiny picks here and there + whatever's needed for CI !

pkg/collector/corechecks/network-devices/versa/client/parser.go
ken-schneider: acked on the other PR where I moved the parsers, this is updated to `!=` as well

pkg/collector/corechecks/network-devices/versa/client/client.go
ken-schneider: Yeah, it's just a lot of passing the same things around.

I'll make a follow up PR to address this

PR #39551, opened 2025-08-05
Jmw/investigate netflow performance and scalability

jmw51798:


comp/netflow/config/config.go
```diff
@@ -36,6 +36,13 @@ type NetflowConfig struct {
 	PrometheusListenerEnabled bool   `mapstructure:"prometheus_listener_enabled"`
 
 	ReverseDNSEnrichmentEnabled bool `mapstructure:"reverse_dns_enrichment_enabled"`
+
+	DropFlowsBeforeAggregator  bool `mapstructure:"drop_flows_before_aggregator"`
+	DropFlowsBeforeEPForwarder bool `mapstructure:"drop_flows_before_ep_forwarder"`
+	SkipHashCollisionDetection bool `mapstructure:"skip_hash_collision_detection"`
+	AggregationHashUseSyncPool bool `mapstructure:"aggregation_hash_use_sync_pool"`
+	PortRollupUseFixedSizeKey  bool `mapstructure:"portrollup_use_fixed_size_key"`
+	GetMemoryStats             bool `mapstructure:"get_memory_stats"`
```
jmw51798: These are config fields for running various experimental code changes.

comp/netflow/flowaggregator/flowaccumulator.go
```diff
@@ -135,7 +203,9 @@ func (f *flowAccumulator) add(flowToAdd *common.Flow) {
 		f.addRDNSEnrichment(aggHash, flowToAdd.SrcAddr, flowToAdd.DstAddr)
 	} else {
 		// use go routine for hash collision detection to avoid blocking critical path
-		go f.detectHashCollision(aggHash, *aggFlow.flow, *flowToAdd)
+		if !f.skipHashCollisionDetection {
```
jmw51798, line 257: Profiling shows a significant amount of overhead for the creation of the goroutines to do the hash collision detection, and the collision detection function is pretty short.  As such it probably makes sense to run the hash detection in-line.

comp/netflow/portrollup/portrollup.go
```diff
@@ -83,6 +115,26 @@ func (prs *EndpointPairPortRollupStore) Add(sourceAddr []byte, destAddr []byte,
 	prs.AddToStore(prs.newStore, srcToDestKey, destToSrcKey, sourceAddr, destAddr, sourcePort, destPort, prs.IsEphemeralFromKeys(srcToDestKey, destToSrcKey))
 }
 
+// addWithFixedSizeKey implements the new fixed-size key approach for IPv4 only
+func (prs *EndpointPairPortRollupStore) addWithFixedSizeKey(sourceAddr []byte, destAddr []byte, sourcePort uint16, destPort uint16) {
```
jmw51798, line 140: This is the change I mentioned to use a fixed size key to avoid the multiple extraneous memory allocations when generating a string key and to reduce portRollup.Add() function runtime by ~33%.

PR #39553, opened 2025-08-05
[NDMII-3569] Cache failures count in autodiscovery

Pierre-L42:
### What does this PR do?

This PR adds caching the number of connection failures to devices in autodiscovery.

### Motivation

Some customers restarts their Agents continuously and so the connection failures never reaches the failures limit, so we'll create checks for devices that we didn't successfully connect for a long time.

Tested on Mimic, failures are stored for devices in cache and is well used across agent restarts.


comp/core/autodiscovery/listeners/snmp.go
```diff
@@ -442,10 +441,38 @@ func (l *SNMPListener) checkDevices() {
...
+			}
+		} else {
+			writeCache = writeCache && (device.Failures != deviceFailures)
+			device.Failures = deviceFailures
+			subnet.devices[entityID] = device
+		}
+
+		if writeCache {
+			l.writeCache(subnet)
+		}
```
TCheruy: I'm not sure I see why this logic is needed, could you please elaborate?
My understanding is that the 
```
if _, present := l.services[entityID]; present {
```
check is meant to avoid re-creating a service (when scanning the subnet) if it was already created from the cache. With that change, it looks like we'll override the failures with 0 in the cache in the following situation?:
- a device service gets created from the cache with a Failure count of 1
- later on, while scanning the subnet, we reach the same device IP
- checkDevice() ([here](https://github.com/DataDog/datadog-agent/pull/39553/files#diff-18cf895ffa7f7f2ddc107902831b3ee19763f4b528ed0bb4f4c54559ad3b68afR210)) creates the service with a failure count of 0
- we override the failure counts with 0 here

comp/core/autodiscovery/listeners/snmp.go
```diff
@@ -525,22 +553,47 @@ func (l *SNMPListener) registerService(pendingDevice devicededuper.PendingDevice
...
+	}
+
+	var failures int
+	isPendingDevice := false
+	writeCache := false
+
+	device, exists := subnet.devices[entityID]
+	if !exists {
+		// This case should happen only when device is a pending device (service created but not registered yet)
+		newFailures, found := l.deviceDeduper.IncreasePendingDeviceFailures(svc.deviceIP)
```
TCheruy: Do we need to keep track of the failure count for pending devices?

My understanding is that the failure count will only increase if the service was created and the auto-discovery re-scans the subnet but the device is now reachable. This only happens once the subnet was fully scanned, so at this point, pending devices should have either been promoted to non-pending devices, or been discarded?

TCheruy (commented):
Great PR! Left a few implementation questions

comp/core/autodiscovery/listeners/snmp.go
Pierre-L42: Yes this is expected because we want to reset the failure count to 0 when we successfully connect to the device, as it was before

comp/core/autodiscovery/listeners/snmp.go
Pierre-L42: Kinda related to the above comment as failure count can be reset

comp/core/autodiscovery/listeners/snmp.go
TCheruy: Ah, this is to makes sure we only track **consecutive** failures!

comp/core/autodiscovery/listeners/snmp.go
```diff
@@ -527,22 +542,34 @@ func (l *SNMPListener) registerService(pendingDevice devicededuper.PendingDevice
...
+		l.delService <- svc
+		delete(l.services, entityID)
+		delete(subnet.devices, entityID)
+
+		writeCache = true
+	}
+
+	if writeCache {
+		l.writeCache(subnet)
 	}
```
TCheruy: I think we can just early return if the device does not exists here
```suggestion
	device, exists := subnet.devices[entityID]
	if !exists {
	  return
	}
		
	device.Failures++
	subnet.devices[entityID] = device

	if l.config.AllowedFailures != -1 && device.Failures >= l.config.AllowedFailures {
		l.delService <- svc
		delete(l.services, entityID)
		delete(subnet.devices, entityID)
	}

	l.writeCache(subnet)
```

TCheruy (approved):
Looks good!

pkg/snmp/devicededuper/device_deduper.go
```diff
@@ -34,6 +34,7 @@ type PendingDevice struct {
 	AuthIndex  int
 	WriteCache bool
 	IP         string
+	Failures   int
```
TCheruy, line 37: missed this but i don't think we need this 😄

PR #39561, opened 2025-08-05
APMSP-2191: Fix prob sampler dropping all OTLP traces

ajgajg1134:
### What does this PR do?
This fixes an issue where the probability sampler would drop all incoming OTLP trace data

### Motivation
We want the probability sampler to work on DD and OTLP data that it processes.

Added a new unit test to cover this exact case. Additionally I also tested an OTEL sample app and verified that traces are now visible in the UI.

### DIFF
--- a/pkg/trace/agent/agent.go
+++ b/pkg/trace/agent/agent.go
@@ -689,19 +689,24 @@ func (a *Agent) runSamplers(now time.Time, ts *info.TagStats, pt traceutil.Proce
 
 	if a.conf.ProbabilisticSamplerEnabled {
 		samplerName = sampler.NameProbabilistic
+		probKeep := false
+
 		if rare {
 			samplerName = sampler.NameRare
-			return true, true
-		}
-		if a.ProbabilisticSampler.Sample(pt.Root) {
+			probKeep = true
+		} else if a.ProbabilisticSampler.Sample(pt.Root) {
 			pt.TraceChunk.Tags[tagDecisionMaker] = probabilitySampling
-			return true, true
-		}
-		if traceContainsError(pt.TraceChunk.Spans, false) {
+			probKeep = true
+		} else if traceContainsError(pt.TraceChunk.Spans, false) {
 			samplerName = sampler.NameError
-			return a.ErrorsSampler.Sample(now, pt.TraceChunk.Spans, pt.Root, pt.TracerEnv), true
+			probKeep = a.ErrorsSampler.Sample(now, pt.TraceChunk.Spans, pt.Root, pt.TracerEnv)
 		}
-		return false, true
+		if probKeep {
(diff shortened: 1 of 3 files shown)


joepeeples (approved):

pkg/trace/agent/agent.go
```diff
@@ -689,19 +689,24 @@ func (a *Agent) runSamplers(now time.Time, ts *info.TagStats, pt traceutil.Proce
...
-		}
-		if traceContainsError(pt.TraceChunk.Spans, false) {
+			probKeep = true
+		} else if traceContainsError(pt.TraceChunk.Spans, false) {
 			samplerName = sampler.NameError
-			return a.ErrorsSampler.Sample(now, pt.TraceChunk.Spans, pt.Root, pt.TracerEnv), true
+			probKeep = a.ErrorsSampler.Sample(now, pt.TraceChunk.Spans, pt.Root, pt.TracerEnv)
+		}
+		if probKeep {
+			pt.TraceChunk.Priority = int32(sampler.PriorityAutoKeep)
```
ichinaski, line 705: What if the trace already had an explicitly set priority? Should we first honor such decisions upstream? 🤔

ajgajg1134: Good question! And nope! When using the probability sampler any and all priorities are overwritten. Head based sampling (aka when the priority has already been set) is incompatible with probability sampling

ichinaski: Great, thanks for the clarification, that makes sense.

ichinaski (approved):

mscrivo:
I think this commit has caused probabilistic sampling to not work properly in conjunction with the datadog ruby gem. After deploying 7.70 (and every version thereafter) our ingestion rates look like this:

image

We have this service configured to sample only 5% of traces using: `c.tracing.sampling.default_rate = 0.05`

ajgajg1134:
> I think this commit has caused probabilistic sampling to not work properly in conjunction with the datadog ruby gem. After deploying 7.70 (and every version thereafter) our ingestion rates look like this:
> 
> image
> We have this service configured to sample only 5% of traces using: `c.tracing.sampling.default_rate = 0.05`

Hello! To better investigate your issue can you please open a Datadog Support ticket? That will let the team take a closer look at your exact setup here.

mscrivo:
> > I think this commit has caused probabilistic sampling to not work properly in conjunction with the datadog ruby gem. After deploying 7.70 (and every version thereafter) our ingestion rates look like this:
> > 
> > image
> > We have this service configured to sample only 5% of traces using: `c.tracing.sampling.default_rate = 0.05`
> 
> Hello! To better investigate your issue can you please open a Datadog Support ticket? That will let the team take a closer look at your exact setup here.
> 

Already done. Really hope they can figure it out because we can't update to any new versions as it stands.

PR #39562, opened 2025-08-05
add lambda tag to http header

michael-zhao459:
### What does this PR do?

This PR adds _dd.origin:lambda as a HTTP header tag to send to DSM backend if the data is coming from a lambda. 

Wrote 2 unit tests, 1 where the tag should not be added if not a lambda and 1 test where the tag should be added if is a lambda. Tested with AWS Sandbox account, 
Screenshot 2025-08-05 at 3 40 13 PM
backend is capable of querying the metric despite having DD_EXTENSION_VERSION = compatibility. 

This is the same tag that the Lambda next-gen extension layer adds https://github.com/DataDog/datadog-lambda-extension/blob/main/bottlecap/src/traces/proxy_flusher.rs#L63-L65

### DIFF
--- a/pkg/trace/api/pipeline_stats.go
+++ b/pkg/trace/api/pipeline_stats.go
@@ -57,6 +57,9 @@ func (r *HTTPReceiver) pipelineStatsProxyHandler() http.Handler {
 		tag := fmt.Sprintf("orchestrator:fargate_%s", strings.ToLower(string(orch)))
 		tags = tags + "," + tag
 	}
+	if r.conf.LambdaFunctionName != "" {
+		tags = tags + "," + "_dd.origin:lambda"
+	}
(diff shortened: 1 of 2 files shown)


pkg/trace/api/pipeline_stats.go
```diff
@@ -57,6 +57,9 @@ func (r *HTTPReceiver) pipelineStatsProxyHandler() http.Handler {
 		tag := fmt.Sprintf("orchestrator:fargate_%s", strings.ToLower(string(orch)))
 		tags = tags + "," + tag
 	}
+	if r.conf.LambdaFunctionName != "" {
```
michael-zhao459, line 60: Approach taken from https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/api/profiles.go#L85-L88 Profiling team

piochelepiotr (approved):

ichinaski (approved):

PR #39563, opened 2025-08-05
[CIEXE-1103]Make vault read generalizable for omnibus tests

sjain313:
### Motivation
CI team is creating dedicated gitlab runners for datadog-agent so that datadog-agents job dont get impacted by other repos. As part of this we would need to replace hardcoded vault path to be generalizable so we can smoothly migrate from shared gitlab runners to dedicated ones

It seems we are running `omnibus` unit tests mocked to be on linux in macos runners. They are also run on kubernetes runners. So this diff handles both cases. 

We can see the CI jobs ran fine and hence we can confirm that they could read the secret successfully from vault.

These tests especially : 
invoke_unit_tests
tests_macos_gitlab_amd64
tests_macos_gitlab_arm64

### DIFF
--- a/.gitlab/source_test/tooling_unit_tests.yml
+++ b/.gitlab/source_test/tooling_unit_tests.yml
@@ -10,4 +10,4 @@ invoke_unit_tests:
   script:
     - python3 -m pip install "git+https://github.com/DataDog/datadog-agent-dev.git@v$(cat .dda/version)" --break-system-packages
     - python3 -m dda self dep sync -f legacy-tasks
-    - POD_NAMESPACE=${POD_NAMESPACE} dda inv -- -e invoke-unit-tests.run
+    - dda inv -- -e invoke-unit-tests.run
--- a/.gitlab/source_test/tooling_unit_tests.yml
+++ b/.gitlab/source_test/tooling_unit_tests.yml
@@ -10,4 +10,4 @@ invoke_unit_tests:
   script:
     - python3 -m pip install "git+https://github.com/DataDog/datadog-agent-dev.git@v$(cat .dda/version)" --break-system-packages
     - python3 -m dda self dep sync -f legacy-tasks
-    - dda inv -- -e invoke-unit-tests.run
+    - POD_NAMESPACE=${POD_NAMESPACE} dda inv -- -e invoke-unit-tests.run
--- a/tasks/libs/common/omnibus.py
+++ b/tasks/libs/common/omnibus.py
@@ -155,9 +155,6 @@ def _hash_paths(hasher, paths: list[str]):
 
 
 def get_dd_api_key(ctx):
-    print(f'sjain:sys.platform: {sys.platform}')
-    for key, value in os.environ.items():
-        print(f"{key}: {value}")
     if sys.platform == 'win32':
         cmd = f'aws.exe ssm get-parameter --region us-east-1 --name {os.environ["API_KEY_ORG2"]} --with-decryption --query "Parameter.Value" --out text'
     elif sys.platform == 'darwin':
--- a/tasks/libs/common/omnibus.py
+++ b/tasks/libs/common/omnibus.py
@@ -160,7 +160,7 @@ def get_dd_api_key(ctx):
     elif sys.platform == 'darwin':
         cmd = f'vault kv get -field=token kv/aws/arn:aws:iam::486234852809:role/ci-datadog-agent/{os.environ["AGENT_API_KEY_ORG2"]}'
     else:
-        cmd = f'vault kv get -field=token kv/k8s/{os.environ["POD_NAMESPACE"]}/datadog-agent/{os.environ["AGENT_API_KEY_ORG2"]}'
+        cmd = f'vault kv get -field=token kv/k8s/{os.environ.get('POD_NAMESPACE', 'gitlab-runner')}/datadog-agent/{os.environ["AGENT_API_KEY_ORG2"]}'
     return ctx.run(cmd, hide=True).stdout.strip()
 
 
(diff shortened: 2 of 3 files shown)


tasks/unit_tests/omnibus_tests.py
```diff
@@ -42,6 +42,7 @@ def _run_calls_to_string(mock_calls):
         'S3_OMNIBUS_GIT_CACHE_BUCKET': 'omnibus-cache',
         'API_KEY_ORG2': 'api-key',
         'AGENT_API_KEY_ORG2': 'agent-api-key',
+        'POD_NAMESPACE': os.environ.get('POD_NAMESPACE', 'gitlab-runner'),
```
sjain313, line 45: We need to have a default value here because these unit tests are also called from `macos` runners  but [`system` is patched as `linux`](https://github.com/DataDog/datadog-agent/pull/39563/files#diff-407acb07e932b50660fc4999612d12a05d6046a1716a53125e1010050c898cebR33). So even though the unit tests are run in macos runners(run on EC2) they try to access vault and dont have K8 `POD_NAMESPACE` env var set.

To confirm - Do we intend to run linux patched unit tests in macos CI job `tests_macos_gitlab_amd64` ? We already have a separate job to run unit tests `invoke_unit_tests`

.gitlab/source_test/tooling_unit_tests.yml
```diff
@@ -10,4 +10,4 @@ invoke_unit_tests:
   script:
     - python3 -m pip install "git+https://github.com/DataDog/datadog-agent-dev.git@v$(cat .dda/version)" --break-system-packages
     - python3 -m dda self dep sync -f legacy-tasks
-    - dda inv -- -e invoke-unit-tests.run
+    - POD_NAMESPACE=${POD_NAMESPACE} dda inv -- -e invoke-unit-tests.run
```
rdesgroppes: Aren't environment variables exported by default?

tasks/unit_tests/omnibus_tests.py
rdesgroppes: I'm not sure I get it, esp. what "system is patched as linux" means. Could you enlighten me, please? (I'm still learning...)

sjain313: oh its L33 - `@mock.patch('sys.platform', 'linux')`. I am wondering why we need to do this ? 

The test will derive the platform based on whatever runner it is running on but it seems like we want to run the tests on macos runner but have the unit test think its running on linux machine(because of the above patch), I am curious why ?

.gitlab/source_test/tooling_unit_tests.yml
sjain313: yup, fixed

sjain313:
I retried the failed docker build CI jobs but they still timeout. I don't think my changes should impact them.

tasks/unit_tests/omnibus_tests.py
rdesgroppes: Doing some archaeology... please hold the line!

tasks/libs/common/omnibus.py
```diff
@@ -160,7 +160,7 @@ def get_dd_api_key(ctx):
     elif sys.platform == 'darwin':
         cmd = f'vault kv get -field=token kv/aws/arn:aws:iam::486234852809:role/ci-datadog-agent/{os.environ["AGENT_API_KEY_ORG2"]}'
     else:
-        cmd = f'vault kv get -field=token kv/k8s/gitlab-runner/datadog-agent/{os.environ["AGENT_API_KEY_ORG2"]}'
+        cmd = f'vault kv get -field=token kv/k8s/{os.environ["POD_NAMESPACE"]}/datadog-agent/{os.environ["AGENT_API_KEY_ORG2"]}'
```
jorgetomtz: Should you use `os.environ.get('POD_NAMESPACE', 'gitlab-runner')` here to preserve the default behavior ie current hardcoded case? Then no need to update unit test

rdesgroppes: I second this suggestion. That would grant you my approval while buying me some time to answer your [above question](https://github.com/DataDog/datadog-agent/pull/39563/files#r2257756956).

sjain313: No strong preference so changed :) 

fwiw - my thought here was that to not unknowingly fallback on 'gitlab-runner` namespace which we are migrating off of and rather fail CI job so we can investigate it. Unit tests were the only caller which I fixed in my other change.

tasks/unit_tests/omnibus_tests.py
rdesgroppes: As [agreed below](https://github.com/DataDog/datadog-agent/pull/39563#discussion_r2260881214), I'll keep digging off this PR which LGTM.

rdesgroppes (approved):

rdesgroppes:
> I retried the failed docker build CI jobs but they still timeout. I don't think my changes should impact them.

You'll never rebase enough.

tasks/unit_tests/omnibus_tests.py
rdesgroppes: That's #39633.

estherk15 (approved):

sjain313:
Just learned about an ongoing incident related to quality tests [#incident-41756](https://dd.enterprise.slack.com/archives/C099D1ZHEA2) , so trying my luck again with rebase based on [incident comments.](https://dd.slack.com/archives/C099D1ZHEA2/p1754598976618849)

louis-cqrl (approved):

PR #39565, opened 2025-08-05
Cleanup datadog-installer tmp directory with GarbageCollector and Purge

mrafi97:
### What does this PR do?
This PR adds additional functionality GarbageCollector and Purge so that the tmp directory is cleaned up. This is to avoid issues with tmp taking up too much disk space after several fleet automation upgrades.

To help with this, the datadog-installer.exe in tmp is moved to the system temp directory and run from there. This will help avoid issues with cleaning up tmp due to the exe being used.

The file path for this is `%TEMP%\datadog-installer\datadog-installer*\datadog-installer.exe`
With each subsequent upgrade/downgrade a new temp directory is made in `%TEMP%\datadog-installer\`

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1478

Added some tests to make sure that tmp is cleaned up and that datadog-installer.exe is moved to the system temp directory.
Existing CI tests should ensure that agent installation and upgrades work as expected.

### DIFF
--- a/pkg/fleet/installer/bootstrap/bootstrap_windows_test.go
+++ b/pkg/fleet/installer/bootstrap/bootstrap_windows_test.go
@@ -11,6 +11,8 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+
+	"github.com/DataDog/datadog-agent/pkg/fleet/installer/paths"
 )
 
 func TestGetInstallPath(t *testing.T) {
@@ -46,6 +48,21 @@ func TestGetInstallPathSystemTemp(t *testing.T) {
 	// create temp directory
 	tmpDir := t.TempDir()
 
+	// backup original CreateSystemTempDir() function and restore the original function once we're done
+	// Use a different function that creates a directory in the test's temp directory
+	var originalCreateSystemTempDir = paths.CreateSystemTempDir
+	defer func() {
+		paths.CreateSystemTempDir = originalCreateSystemTempDir
+	}()
+
+	paths.CreateSystemTempDir = func() (string, error) {
+		err := os.Mkdir(filepath.Join(tmpDir, "datadog-installer"), 0755)
+		if err != nil {
+			return "", err
+		}
+		return filepath.Join(tmpDir, "datadog-installer"), nil
+	}
+
 	// add an exe to the temp directory
 	exePath := filepath.Join(tmpDir, "datadog-installer.exe")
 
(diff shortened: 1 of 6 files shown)


domalessi (approved):

pkg/fleet/installer/paths/installer_paths_windows.go
```diff
@@ -561,3 +561,35 @@ func resetPermissionsForTree(path string) error {
...
+var CreateSystemTempDir = createSystemTempDir
+
+func createSystemTempDir() (string, error) {
+
+	sddl := "O:BAG:BAD:PAI(A;OICI;FA;;;SY)(A;OICI;FA;;;BA)(A;CI;0x1200a9;;;WD)"
+
+	// os.TempDir() uses GetTempPath, returning the first non-empty value from %TMP%, %TEMP%, %USERPROFILE%, or the Windows directory.
+	//
+	// https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppath2w
+	tempDir := os.TempDir()
```
jack0x2, line 584: should we hardcode this to the systemtemp directory?

pkg/fleet/installer/bootstrap/bootstrap_windows_test.go
```diff
@@ -41,3 +43,66 @@ func TestGetInstallPath(t *testing.T) {
...
 	}
 }
+
+func TestGetInstallPathSystemTemp(t *testing.T) {
+	// create temp directory
+	tmpDir := t.TempDir()
+
+	// backup original CreateSystemTempDir() function and restore the original function once we're done
+	// Use a different function that creates a directory in the test's temp directory
+	var originalCreateSystemTempDir = paths.CreateSystemTempDir
```
jack0x2, line 53: Why do we have to back up this function?

pkg/fleet/installer/installer_test.go
```diff
@@ -405,7 +405,28 @@ func TestPurge(t *testing.T) {
...
+
+		oldPurgeTmpDirectory := purgeTmpDirectory
+		purgeTmpDirectory = func(_ string) error {
+			err := os.RemoveAll(tmpPath)
+			if err != nil {
+				t.Fatalf("could not delete tmp directory: %v", err)
+			}
+			return nil
+		}
+		defer func() {
```
jack0x2, line 421: same as before to we need a backup of the original?

pkg/fleet/installer/paths/installer_paths_windows.go
jack0x2: From the docs:

When calling this function from a process running as SYSTEM it will return the path C:\Windows\SystemTemp, which is inaccessible to non-SYSTEM processes. For non-SYSTEM processes, GetTempPath2 will behave the same as [GetTempPath](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppatha).

For system processes, the GetTempPath2 function checks for the existence of the environment variable SystemTemp. If this environment variable is set, it will use the value of the environment variable as the path instead of the default system provided path on the C: drive.

So it looks like as system we return the system path always.

mrafi97: Yup that's pretty much why we're using os.TempDir(), takes the guess work out of whether we should be using `Windows\SystemTemp` or `Windows\Temp` and just letting Windows give the right path. Also from my understanding not all versions of Windows 2016 has `Windows\SystemTemp`. From the docs:

> When Windows updates released in March 2025 and later are installed, this API will be supported on Windows 10, version 1607 (Build 14393.7876) and later and Windows Server 2016 Build 14393.7876 and later versions. See [KB5053594](https://support.microsoft.com/topic/march-11-2025-kb5053594-os-build-14393-7876-831b6318-8f05-4c41-b413-509fb89baa34) for more information.

pkg/fleet/installer/bootstrap/bootstrap_windows_test.go
mrafi97: When I was looking up how to create mocks of some of these functions, it was mentioned that it's good to backup the original function and restore it with defer. That way any other tests that use that function will have its original function definition rather than mocked one that was made for this test: https://www.pixelstech.net/article/1639186594-Restore-mocked-variables-in-GoLang-unit-test

jack0x2 (approved):

PR #39570, opened 2025-08-06
Initial version of MODULE.bazel

aiuto:
Initial version of MODULE.bazel

This brings in some fundamental rule sets and a dependency on zlib.

The rule sets are added no so that we can swarm on individual new pieces
of code without hitting merge conflicts on MODULE.bazel.

It does not modify CI at this time.

Try to build something.  On both linux and macos. Didn't do windows yet.
```
bazel build @zlib//:all
ar tv   bazel-bin/external/zlib+/libz.a
```

tickets:
https://datadoghq.atlassian.net/browse/ABLD-130
https://datadoghq.atlassian.net/browse/ABLD-139


.bazelrc
```diff
@@ -0,0 +1,23 @@
...
+build --verbose_failures
+
+# Set up bazel HOST platform specific flags
+#
+# These configurations set flags for the OS we are building *on*, not the OS we
+# are building *for*.
+common --enable_platform_specific_config
+build:linux --sandbox_add_mount_pair=/tmp
+build:macos --sandbox_add_mount_pair=/var/tmp
+build:windows --sandbox_add_mount_pair=C:\Temp
```
rdesgroppes: I think the workaround is no longer needed:
https://github.com/bazelbuild/bazel/commit/fb6658c86164b81205a056a9a54975a62f2f957a

.bazelrc
```diff
@@ -0,0 +1,23 @@
+# Do not edit this file without a review from @DataDog/agent-build
```
JSGette, line 1: How about including best practice flags right from the start? The article isn't very fresh and some flags were flipped since then, but I would still add the rest:
https://blog.aspect.build/bazelrc-flags

So the list of flags we should add:

-  `build --sandbox_default_allow_network=false` - to restrict network access of sandboxes
-  `build --incompatible_strict_action_env` - to strip PATH envvar and not pass other envvars into actions
-  `build --modify_execution_info=PackageTar=+no-remote` - packaging are usually cheap in terms of compute and different almost every time due to volatile data included into final packages, thus, those shouldn't land in CAS nor built remotely in the future
-  `build --nolegacy_external_runfiles` - improves sandbox performance
-  `startup --host_jvm_args=-DBAZEL_TRACK_SOURCE_DIRECTORIES=1` - not very relevant for CI, but should help locally ([details](https://github.com/bazelbuild/bazel/issues/25834))
-  `build --experimental_remote_merkle_tree_cache --experimental_remote_merkle_tree_cache_size=[XX]` - this should reduce number of FindMissingBlobs requests and improve cache lookup performance
-  `build --remote_local_fallback` - we don't want to fail the build if cache isn't available
-  `build --incompatible_remote_results_ignore_disk` - writes to disk cache if `--noremote_upload_local_results` is set along with `--disk_cache` which should be a default behavior locally
-  `build --incompatible_default_to_explicit_init_py` - [details](https://github.com/bazelbuild/bazel/issues/10076)
-  `test --test_verbose_timeout_warnings` - to print warnings if test's duration doesn't correspond its size
-  `build --experimental_inmemory_sandbox_stashes` - bazel reuses sandboxes by default, this flag keeps stashed sandboxes in-memory that reduces IO footprint, might lead to high RAM consumption, though.
-  `test --build_tests_only` - tell bazel to only build test-relevant targets, especially convenient locally
- `--experimental_use_hermetic_linux_sandbox` - to use hardlinks and not mount root, this will only mount directories mentioned in `--sandbox_add_mount_pair `


As a last line to allow local experiments without the need to rewrite original values in `.bazelrc`:
`try-import %workspace%/.bazelrc.user`

KevinFairise2, line 1: Is it really useful? Since you are going to be codeowner, you will be asked for a review on any update of that file

KevinFairise2 (approved):

.bazelrc
```diff
@@ -0,0 +1,23 @@
...
+#
+# These configurations set flags for the OS we are building *on*, not the OS we
+# are building *for*.
+common --enable_platform_specific_config
+build:linux --sandbox_add_mount_pair=/tmp
+build:macos --sandbox_add_mount_pair=/var/tmp
+build:windows --sandbox_add_mount_pair=C:\Temp
+# Point to the mingw location and other settings that point to local stuff
+# TODO: This should eventually go somewhere else (to avoid requiring specific environment setups)
+build:windows --action_env=PATH --define=MINGW_PATH=c:/tools/msys64/mingw64 --define=GCC_VERSION="12.2.0"
```
alopezz: We may prefer to leave this out until we actually include the toolchain bits that these define's refer to.

MODULE.bazel
```diff
@@ -0,0 +1,23 @@
...
+bazel_dep(name = "rules_cc", version = "0.1.1")
+bazel_dep(name = "rules_proto", version = "7.1.0")
+bazel_dep(name = "protobuf", version = "29.3", repo_name = "com_google_protobuf")
+bazel_dep(name = "rules_rust", version = "0.59.2")
+
+git_override(
+    module_name = "rules_foreign_cc",
+    commit = "1b198e7e2f733f65935a78389b5a8cdedc254cb0",
+    remote = "https://github.com/bazel-contrib/rules_foreign_cc",
+)
```
alopezz: This change has been released now, so the git override is no longer needed. I think it would make sense to start from a clean slate here and just use a more recent version of the `rules_foreign_cc` `bazel_dep` instead..

MODULE.bazel
```diff
@@ -0,0 +1,23 @@
+bazel_dep(name = "rules_go", version = "0.53.0")
+bazel_dep(name = "gazelle", version = "0.42.0")
+bazel_dep(name = "platforms", version = "0.0.11")
+bazel_dep(name = "rules_pkg", version = "1.1.0")
+bazel_dep(name = "rules_python", version = "1.2.0")
+bazel_dep(name = "rules_foreign_cc", version = "0.13.0")
+bazel_dep(name = "rules_cc", version = "0.1.1")
+bazel_dep(name = "rules_proto", version = "7.1.0")
+bazel_dep(name = "protobuf", version = "29.3", repo_name = "com_google_protobuf")
+bazel_dep(name = "rules_rust", version = "0.59.2")
```
alopezz: Why include dependencies we're not using yet? I feel like it would be better to add these as they are needed, to try to make sure that whatever we need is actually used.

.bazelrc
aiuto: Sounds good. Since we are not trying to use Bazel7, we are fine

.bazelrc
aiuto: I do find these comments useful. They are a warning that even a global owner should be careful with changes.  It's like a "Warning: High Voltage" sign on an electrical panel.

.bazelrc
aiuto: Removed.

MODULE.bazel
aiuto: I was trying to avoid us all stomping on the same set of deps while we add new  dependencies. But, I'm not going to die on that. Removed except for platforms, which is foundational.

MODULE.bazel
aiuto: Removed

.bazelrc
aiuto: Given that list of flags, I think we need a separate PR just for that.
I worry a lot about incompatible and experimental flags. I know how the Bazel team does not really care about carrying them forward in good ways.

aiuto:
Going to drop and recreate.  My commits were not being verified. I *think* it is because tony.aiuto@datadoghq.com was not my primary github address.

.bazelrc
JSGette: Agreed, let's track it separately: https://datadoghq.atlassian.net/browse/ABLD-142

One flag I would still ask you to add, though. Which is `build --incompatible_strict_action_env`, this one ensures that `PATH` is exactly the same on each and every machine and that other envvars aren't leaking into action command

PR #39574, opened 2025-08-06
dyninst: introduce DiskCache for decompressed sections

ajwerner:
### What does this PR do?

This PR does the work to prepare the Go DI subsystem to use file-backed mmaps for decompressed dwarf sections.

### Motivation

DWARF data can be large, generally needs random access, and is shipped compressed by default. The system-probe doesn't have the luxury of lots of RAM. To deal with that, we can use disk space to enable the kernel to page out the portions of the decompressed sections we do not need and to not count these sections as part of our resident working set.

I tested manually as well as using the automated tests.

### Additional Notes

A related PR stacked on this one will actually do the plumbing to hook this work up to the primary implementation.

Part of [DEBUG-4292](https://datadoghq.atlassian.net/browse/DEBUG-4292)

[DEBUG-4292]: https://datadoghq.atlassian.net/browse/DEBUG-4292?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/dyninst/object/disk_cache.go
```diff
@@ -0,0 +1,544 @@
...
+	// MaxTotalBytes is the maximum aggregate size of all sections that can be
+	// cached, in bytes.
+	MaxTotalBytes uint64
+}
+
+func (cfg *DiskCacheConfig) validate() error {
+	if cfg.DirPath == "" {
+		return fmt.Errorf("dirPath must not be empty")
+	}
+	if cfg.RequiredDiskSpaceBytes == 0 {
```
piob-io: Why not allow zero? To effectively disable this protection?

pkg/dyninst/htlhash/htlhash.go
```diff
@@ -0,0 +1,104 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+// Package htlhash computes the htl hash of an executable.
```
andreimatei: // Package htlhash computes the htl hash (head, tail, length) of an executable.

pkg/dyninst/object/elf.go
```diff
@@ -115,20 +115,38 @@ func IsStrippedBinaryError(err error) bool {
 	return errors.As(err, &decodeErr)
 }
 
-// OpenElfFile opens an elf file from a path.
+// InMemoryElfFileLoader is an elf file loader that loads elf files from memory.
```
andreimatei: "from memory" or more like "into memory"?

pkg/dyninst/object/elf.go
```diff
@@ -115,20 +115,38 @@ func IsStrippedBinaryError(err error) bool {
...
+	_ struct{} // prevent instantiation
+}
+
+// Load loads an elf file from the given path.
+func (l *InMemoryElfFileLoader) Load(path string) (*ElfFile, error) {
+	return OpenElfFile(path)
+}
+
+// NewInMemoryElfFileLoader creates an InMemoryElfFileLoader that will load elf
+// files from disk and use anonymous memory mappings for the compressed section
```
andreimatei, line 130: I think we need more words here to explain what the deal is with this memory mappings. The whole heap generally uses "anonymous memory mappings"; I think we want to say that we're not using the Go heap in order to not affect the GC thresholds (right?)

pkg/dyninst/object/disk_cache.go
```diff
@@ -0,0 +1,544 @@
...
+		}
+		err = os.Remove(filepath.Join(cfg.DirPath, name))
+		if err != nil && !os.IsNotExist(err) {
+			if failedToRemove < 5 { // warn, but do not spam the logs
+				log.Warnf("failed to remove cache file on startup: %v", err)
+			}
+			failedToRemove++
+		}
+	}
+	// In case there were tons of files we failed to remove, warn.
```
piob-io: I'd think that tons would be more than zero?

pkg/dyninst/object/disk_cache.go
```diff
@@ -0,0 +1,544 @@
...
+func newDiskCache(cfg DiskCacheConfig, diskUsageReader diskUsageReader) (*DiskCache, error) {
+	if err := cfg.validate(); err != nil {
+		return nil, err
+	}
+	if err := prepareCacheDir(cfg); err != nil {
+		return nil, err
+	}
+	c := &DiskCache{
+		dirPath:      cfg.DirPath,
+		maxTotalSize: cfg.MaxTotalBytes,
```
andreimatei: nit: choose one of `totalBytes` or `totalSize`

pkg/dyninst/object/mmap.go
```diff
@@ -25,11 +25,19 @@ type MMappingElfFile struct {
...
-	Data    []byte
+	data    []byte
 	mmaped  []byte
 	cleanup runtime.Cleanup
 }
 
+// Data returns the data of the mmapped section.
+//
+// The returned data is valid until Close is called or the MMappedData is
+// garbage collected.
```
andreimatei: I think it's worth making this more clear and saying that someone must hold a reference to the `MMappedData`; in particular they should hold a reference in order to call `MMappedData.Close()`

pkg/dyninst/object/disk_cache.go
```diff
@@ -0,0 +1,545 @@
...
+type DiskCacheConfig struct {
+	// DirPath is the directory to store cached sections.
+	DirPath string
+	// RequiredDiskSpaceBytes is the minimum number of bytes that must remain
+	// available on the underlying filesystem *after* writing a decompressed
+	// section.
+	RequiredDiskSpaceBytes uint64
+	// RequiredDiskSpacePercent is the minimum percentage of total disk capacity
+	// that must remain free *after* writing a decompressed section.
+	RequiredDiskSpacePercent float64
```
andreimatei, line 44: say how these two `Required...` fields work with one another

pkg/dyninst/object/disk_cache.go
```diff
@@ -0,0 +1,545 @@
...
+
+	entryToDelete, err := decrementRefCount()
+	if err != nil {
+		log.Errorf("release invariant violation: %w", err)
+		return
+	}
+	if entryToDelete == nil {
+		return
+	}
+	defer close(entryToDelete.deleted)
```
andreimatei: would the code from here on look better as a method on `cacheEntry`?

andreimatei: or rather maybe this whole `DiskCache.release()` would look better as a method on `cacheEntry` with a single `c.mu.Lock() / defer c.mu.Unlock()`, and with another pair of `Unlock()/Lock()` in the middle around the file deletion. It's kinda hard to read as it is written now, and the whole point about a mutex not being held around the file deletion is obscured.

andreimatei (commented):
LGTM

pkg/dyninst/object/disk_cache.go
```diff
@@ -0,0 +1,545 @@
...
+	cr compressedFileRange,
+	mef *MMappingElfFile,
+) error {
+	if err := checker.check(uint64(cr.uncompressedLength)); err != nil {
+		return err
+	}
+
+	cacheDir := filepath.Dir(outputPath)
+	cacheName := filepath.Base(outputPath)
+	tmpFile, err := os.CreateTemp(cacheDir, fmt.Sprintf(".tmp.%s.*", cacheName))
```
piob-io: This tmpFile could be unnamed (using O_TMPFILE), so we don't leak it (temporarily, but still) on crash. Once everything succeeds, an unnamed file can be linked (replacing the current rename).

pkg/dyninst/object/disk_cache.go
```diff
@@ -0,0 +1,545 @@
...
+	// Ensure decompression (only first goroutine will actually perform work).
+	entry.decompress.Do(func() {
+		entry.decompress.err = decompressToDisk(
+			&c.checker, entry.path, cr.compressedFileRange, mef,
+		)
+	})
+	if entry.decompress.err != nil {
+		return nil, entry.decompress.err
+	}
+	md, err := mmapDecompressedData(entry.path)
```
piob-io: Why don't we cover mmap also under "do once"? It's logically correct, but why not optimize given that we already have some "do once" logic, and we do our own refcounting for the file anyway?

piob-io (commented):
Neat!

pkg/dyninst/object/disk_cache.go
ajwerner: Done

pkg/dyninst/object/disk_cache.go
ajwerner: Done.

pkg/dyninst/object/disk_cache.go
ajwerner: Done.

pkg/dyninst/object/disk_cache.go
ajwerner: Done.

pkg/dyninst/object/disk_cache.go
ajwerner: Done.

pkg/dyninst/object/disk_cache.go
ajwerner: I did now anchor this logic on the `cacheEntry` and have it hold on to the cache and directly hold on to the one mmap. I also added more commentary.

pkg/dyninst/object/disk_cache.go
ajwerner: I think I'll do as you suggest, in a sense. I think having the `/proc/<pid>/maps` point the finger at a file path that implicates us is a good idea, but I also think it's nice to delete this startup logic. I also don't want to risk bad interactions between different processes. I've gotten rid of the removal of files on startup. There's a tiny window where we might leave an abandoned empty file, but I think that's okay. At least for this iteration. Go doesn't have great support for O_TMPFILE out of the box anyway.

pkg/dyninst/object/elf.go
ajwerner: Yeah that was some AI slop.

pkg/dyninst/object/elf.go
ajwerner: Said more.

pkg/dyninst/object/mmap.go
ajwerner: I said more things. I think it's better but it's in some ways less accurate.

pkg/dyninst/htlhash/htlhash.go
ajwerner: Done.

ajwerner (commented):
Alright folks, I reworked this to eagerly delete the files before we write any bytes into them (akin to what Piotr suggested). I also made the cleanup on GC more sound and tested it. Now that the files are gone, the tests poke around in the proc maps to validate the lifetimes of the data.

Thanks for the reviews!

piob-io (approved):

ajwerner:
@andreimatei I'm merging it but happy to address more feedback in a follow-up.

PR #39575, opened 2025-08-06
dyninst/module: plumb configuration for the DiskCache

ajwerner:
This PR hooks up #39574 to dyninst and introduces the configuration options to control it.

Closes [DEBUG-4292](https://datadoghq.atlassian.net/browse/DEBUG-4292)

[DEBUG-4292]: https://datadoghq.atlassian.net/browse/DEBUG-4292?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/config/setup/system_probe.go
+++ b/pkg/config/setup/system_probe.go
@@ -43,6 +43,11 @@ const (
 	// defaultBTFOutputDir is the default path for extracted BTF
 	defaultBTFOutputDir = "/var/tmp/datadog-agent/system-probe/btf"
 
+	// defaultDynamicInstrumentationDebugInfoDir is the default path for debug
+	// info for dynamic instrumentation this is the directory where the debug
+	// info is decompressed into during processing.
+	defaultDynamicInstrumentationDebugInfoDir = "/var/tmp/datadog-agent/system-probe/dynamic-instrumentation/decompressed-debug-info"
+
 	// defaultAptConfigDirSuffix is the default path under `/etc` to the apt config directory
 	defaultAptConfigDirSuffix = "/apt"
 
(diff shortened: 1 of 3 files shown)


pkg/config/setup/system_probe.go
```diff
@@ -43,6 +43,11 @@ const (
 	// defaultBTFOutputDir is the default path for extracted BTF
 	defaultBTFOutputDir = "/var/tmp/datadog-agent/system-probe/btf"
 
+	// defaultDynamicInstrumentationDebugInfoDir is the default path for debug
+	// info for dynamic instrumentation this is the directory where the debug
+	// info is decompressed into during processing.
+	defaultDynamicInstrumentationDebugInfoDir = "/var/tmp/datadog-agent/system-probe/dynamic-instrumentation/decompressed-debug-info"
```
hush-hush, line 49: What about Windows ?

hush-hush (approved):
Looks good for agent configuration. I added 1 question about windows support.

pkg/config/setup/system_probe.go
ajwerner: This feature doesn't work on windows and the consumer of this setting has a build tag for `linux_bpf`, but I'm open to suggestions if there's something more appropriate to use here.

piob-io (approved):

pkg/config/setup/system_probe.go
```diff
@@ -169,12 +174,17 @@ func InitSystemProbeConfig(cfg pkgconfigmodel.Config) {
...
 	cfg.BindEnvAndSetDefault(join(spNS, "attach_kprobes_with_kprobe_events_abi"), false, "DD_ATTACH_KPROBES_WITH_KPROBE_EVENTS_ABI")
 
-	// User Tracer
+	// Dynamic Instrumentation settings
 	cfg.BindEnvAndSetDefault(join(diNS, "enabled"), false, "DD_DYNAMIC_INSTRUMENTATION_ENABLED")
 	cfg.BindEnvAndSetDefault(join(diNS, "offline_mode"), false, "DD_DYNAMIC_INSTRUMENTATION_OFFLINE_MODE")
 	cfg.BindEnvAndSetDefault(join(diNS, "probes_file_path"), false, "DD_DYNAMIC_INSTRUMENTATION_PROBES_FILE_PATH")
 	cfg.BindEnvAndSetDefault(join(diNS, "snapshot_output_file_path"), false, "DD_DYNAMIC_INSTRUMENTATION_SNAPSHOT_FILE_PATH")
 	cfg.BindEnvAndSetDefault(join(diNS, "diagnostics_output_file_path"), false, "DD_DYNAMIC_INSTRUMENTATION_DIAGNOSTICS_FILE_PATH")
+	cfg.BindEnvAndSetDefault(join(diNS, "debug_info_disk_cache", "enabled"), true)
```
val06, line 183: do we want to enable it by default? 

(I assumed here it is a new feature, so perhaps it is good to have it behind a disabled feature flag to get more confidence)

val06 (approved):
approved with comment for files owned by ebpf-platform

pkg/config/setup/system_probe.go
ajwerner: It’s a good question. We’re only rolling the feature out internally this month. The primary failure mode of this patch is that we fail to use DI because there isn’t enough disk space. The failure mode we’re addressing with it is OOMs. Prior to this change (and the code it enabled) we were storing the data in RAM (not on the Go heap, but still in RAM.). 

Another note that may help is that this feature is behind a feature flag and only enabled on staging for internal users right now. Even then you have to place probes for this to kick in. We also control the minimum version of the agent to which we’ll send probes. We’re only committed to internal use for 7.70. If things work very well and we have a high degree of confidence we’ll look at previewing for external users at 7.70, but I’m not necessarily counting on it. 

Given the OOMs I think this has to be the default. Before too too long I hope we can offload this dwarf processing and not have to be doing this.

ajwerner:
Thank you for the review!

PR #39576, opened 2025-08-06
CONTP-913: Add cluster check to inventory metadata collection

zhuminyi:
### What does this PR do?
This PR implements cluster check inventory metadata collection for the Datadog Cluster Agent by creating a new clusterchecks metadata component. The implementation provides:

- New /metadata/cluster-checks API endpoint that exposes cluster check configurations
- Dedicated clustercheck_metadata payload separate from the existing inventorychecks component
- Leader-only reporting to ensure only the leader cluster agent sends cluster-wide metadata
- Runtime cluster check status collection including dispatch information, node counts, and warmup state
- Check_metadata file in Flare 

### Motivation
Previously, cluster check configurations (like kubernetes_state_core) were not appearing in the agent's inventory metadata, which made it difficult to:

- Monitor cluster check deployment - No visibility into which cluster checks are configured and running
- Debug cluster check issues - Missing metadata in flares when troubleshooting CLC (Cluster Level Check) runners

The new implementation ensures that:

- Cluster agents expose only cluster check metadata via /metadata/cluster-checks
- Node agents continue using /metadata/inventory-checks for local checks
- Backend can distinguish between cluster and node agent check metadata (different payloads)

Flare
Run agent flare in dca, download flare zip and cluster-checks-metadata.json can be found. 
```
{
    "clustername": "minyi-test-gke",
    "cluster_id": "d1bd4888-2990-4312-a610-c9eae75acba4",
    "timestamp": 1755060822849588320,
    "clustercheck_metadata": {
        "kubernetes_state_core": [
            {
                "config.hash": "kubernetes_state_core:88b171c1251ce1d2",
                "config.provider": "file",
                "config.source": "file:/etc/datadog-agent/conf.d/kubernetes_state_core.yaml",
                "errors": "",
                "init_config": "",
                "instance_config": "collectors:\n- nodes\n- pods\n- services\n- resourcequotas\n- replicationcontrollers\n- limitranges\n- persistentvolumeclaims\n- persistentvolumes\n- namespaces\n- endpoints\n- daemonsets\n- deployments\n- replicasets\n- statefulsets\n- cronjobs\n- jobs\n- horizontalpodautoscalers\n- poddisruptionbudgets\n- storageclasses\n- volumeattachments\n- ingresses\nempty_default_hostname: true\npod_collection_mode: cluster_unassigned\nskip_leader_election: true\ntags:\n- cluster_name:minyi-test-gke\n- kube_cluster_name:minyi-test-gke\n- orch_cluster_id:d1bd4888-2990-4312-a610-c9eae75acba4\n",
                "node_name": "datadog-agent-linux-clusterchecks-bf4ddbd6d-w5q5k",
                "status": "DISPATCHED"
            }
        ]
    },
    "clustercheck_status": {
        "dangling_count": 0,
        "node_count": 1
    },
    "uuid": "350c8479-acf2-29ac-da36-97deaa5939e6"
}
```

### Additional Notes

This feature is enabled by dca configs:
```
inventories_checks_configuration_enabled
inventories_enabled
enable_metadata_collection
```


comp/metadata/clusterchecks/impl/clusterchecks.go
```diff
@@ -0,0 +1,277 @@
...
+	} else {
+		// For non-cluster agents, create a no-op payload
+		cc.InventoryPayload = util.CreateInventoryPayload(
+			deps.Conf,
+			deps.Log,
+			nil,
+			cc.getPayloadAsMarshaler,
+			"cluster-checks-metadata.json",
+		)
+	}
```
misteriaud: **❓ question:** ‏I'm not sure to understand why we need to return a no-op InventoryPayload. It's probably better to return a `nil` MetadataProvider (`nil` providers are filtered by the metadata runner component [here](https://github.com/DataDog/datadog-agent/blob/main/comp/metadata/runner/runnerimpl/runner.go#L67))

comp/metadata/clusterchecks/def/component.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"net/http"
+)
+
+// team: container-platform
+
+// Component is the component interface for clusterchecks metadata
+type Component interface {
+	// SetClusterHandler sets the cluster checks handler for collecting cluster check metadata
+	SetClusterHandler(handler interface{})
```
misteriaud: **💬 suggestion:** I would highly recommend to refactor first the ClusterCheckHandler into a component in order to define a clear dependency using Fx and avoid using this kind of anti-pattern. 
It seems that `clustercheck.NewHandler` [only require 2 dependencies](https://github.com/DataDog/datadog-agent/blob/main/pkg/clusteragent/clusterchecks/handler.go#L63-L64) which are already components (`autodiscover` and `tagger`).

comp/metadata/clusterchecks/impl/clusterchecks.go
zhuminyi: updated

comp/metadata/clusterchecks/def/component.go
zhuminyi: clusterchecks component will be part of this PR https://github.com/DataDog/datadog-agent/pull/39901

comp/core/clusterchecks/def/component.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/DataDog/datadog-agent/pkg/clusteragent/clusterchecks/types"
+)
+
+// team: container-platform
+
+// Component is the interface for the cluster checks handler
+type Component interface {
+	// Run is the main goroutine for the handler. It has to
+	// be called in a goroutine with a cancellable context.
+	Run(ctx context.Context)
```
davidor: I think this shouldn't be exposed, we should use fx lifecycles instead.

comp/core/clusterchecks/def/component.go
```diff
@@ -0,0 +1,57 @@
...
+
+	// API Methods
+
+	// RejectOrForwardLeaderQuery rejects or forwards the query based on leadership status
+	RejectOrForwardLeaderQuery(rw http.ResponseWriter, req *http.Request) bool
```
davidor: I think this is an implementation detail that shouldn't be exposed in the interface.
I think this is only used by functions defined below, so we should refactor to avoid exposing this.

comp/core/clusterchecks/def/component.go
```diff
@@ -0,0 +1,57 @@
...
+	"context"
+	"net/http"
+
+	"github.com/DataDog/datadog-agent/pkg/clusteragent/clusterchecks/types"
+)
+
+// team: container-platform
+
+// Component is the interface for the cluster checks handler
+type Component interface {
```
davidor: I’m not sure about this interface. Many of the exported functions are called only from `cmd/cluster-agent/api/v1/clusterchecks.go`, which defines the Cluster Check endpoints.

If so, exposing these functions solely to define the Cluster Check endpoints seems a bit odd.

davidor: If we create a component for this (though I’m not sure it’s needed yet), we should move all the code under `pkg/clusteragent/clusterchecks` into the component’s impl folder. Otherwise, other callers may call the implementation directly instead of using the component.

comp/core/clusterchecks/def/component.go
zhuminyi: Code moved to https://github.com/DataDog/datadog-agent/pull/39901

comp/core/clusterchecks/def/component.go
zhuminyi: moved.

zhuminyi: moved

comp/metadata/clusterchecks/impl/clusterchecks.go
```diff
@@ -0,0 +1,305 @@
...
+	Provider runnerimpl.Provider
+}
+
+// NewComponent creates a new clusterchecks component
+func NewComponent(deps Requires) Provides {
+	// Get cluster identification
+	clusterName := clustername.GetClusterName(context.TODO(), "")
+	clusterID, err := clustername.GetClusterID()
+	if err != nil {
+		clusterID = "" // Handle error gracefully like clusteragent
```
dustmop, line 101: Could we get a log.Debug for the `err` here?

comp/metadata/clusterchecks/impl/clusterchecks.go
```diff
@@ -0,0 +1,305 @@
...
+	state, err := cc.clusterHandler.GetState()
+	if err != nil {
+		cc.log.Debugf("Error getting cluster check state: %s", err)
+		return
+	}
+
+	// Add status information to payload
+	// dangling_count: how many checks are not assigned to any node
+	payload.ClusterCheckStatus["dangling_count"] = len(state.Dangling)
+	//node_count: how many nodes are available for check distribution
```
dustmop: nit: please add a space after `//` and before `node_count` in the comment

comp/metadata/clusterchecks/impl/clusterchecks_noop.go
```diff
@@ -0,0 +1,54 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+//go:build !clusterchecks
```
dustmop: The preferred way to handle a no-op implementation would be to move this implementation into `comp/metadata/clusterchecks/impl-noop`, as described [by the docs](https://datadoghq.dev/datadog-agent/components/creating-components/#file-hierarchy). This would make the build tag unnecessary, instead the top-level package that is importing this one would decide which package to import, the real implementation or the no-op. If the `clusterchecks` build tag is deeply ingrained and shouldn't be removed (I'm guessing this is the case) the decision of which implementation to pick should be done higher in the call chain and shouldn't be solely done in the component itself. For example it could happen in the `fx/fx.go` file which returns the Compnent in the `Module` function.

comp/metadata/clusterchecks/impl/mock.go
```diff
@@ -0,0 +1,31 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+package clusterchecksimpl
```
dustmop: Similarly, we recommend to put the mock into `comp/metadata/clusterchecks/mock/mock.go`.

comp/metadata/clusterchecks/impl/clusterchecks.go
```diff
@@ -0,0 +1,305 @@
...
+		cc.clusterHandler = h
+		cc.log.Debug("Cluster handler set for metadata component")
+	} else {
+		cc.log.Warn("Failed to set cluster handler: invalid type")
+	}
+}
+
+// isLeader checks if the cluster agent is the leader using GetState.
+// The handler's GetState() returns:
+// - NotRunning == "" when state is leader
```
dustmop: How about creating string constants for these possible states? Seems better than only referencing them in comments.

comp/metadata/clusterchecks/impl/clusterchecks_noop.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/DataDog/datadog-agent/pkg/serializer"
+)
+
+type noopImpl struct{}
+
+// Requires defines the dependencies for the clusterchecks metadata component (no-op)
+type Requires struct {
+	Log        log.Component
+	Conf       config.Component
+	Serializer serializer.MetricSerializer
```
dustmop: No need to require these three dependencies for the no-op: different implementations can have different requirements (and providers). In fact, it's better to leave them out, because then callers that only use the no-op won't pick up these unnecessary pieces.

pkg/flare/clusteragent/archive_dca.go
```diff
@@ -140,6 +143,27 @@ func getMetadataMap(fb flaretypes.FlareBuilder) error {
...
+func getClusterChecksMetadata(fb flaretypes.FlareBuilder, client ipc.HTTPClient) error {
+	targetURL := url.URL{
+		Scheme: "https",
+		Host:   fmt.Sprintf("localhost:%v", pkgconfigsetup.Datadog().GetInt("cluster_agent.cmd_port")),
+		Path:   "/metadata/cluster-checks",
+	}
+
+	r, err := client.Get(targetURL.String(), ipchttp.WithCloseConnection)
+	if err != nil {
+		if r != nil && string(r) != "" {
```
dustmop:
```suggestion
		if len(r) > 0 {
```

comp/metadata/clusterchecks/impl/clusterchecks_noop.go
zhuminyi: removed

comp/metadata/clusterchecks/impl/clusterchecks_noop.go
zhuminyi: noop removed since it is not used anywhere.

dustmop (approved):
LGTM!

misteriaud (approved):
LGTM for ARUN owned files 👍

comp/metadata/clusterchecks/def/component.go
```diff
@@ -0,0 +1,22 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+// Package clusterchecks provides the clusterchecks metadata component
```
davidor, line 6: The rest of files in `comp/metadata/clusterchecks` have `//go:build clusterchecks`. Should this one too?

comp/metadata/clusterchecks/def/component.go
```diff
@@ -0,0 +1,22 @@
...
+
+// team: container-platform
+
+// Component is the component interface for clusterchecks metadata
+type Component interface {
+	// WritePayloadAsJSON writes the cluster checks payload as JSON to HTTP response
+	WritePayloadAsJSON(w http.ResponseWriter, r *http.Request)
+
+	// SetClusterHandler sets the cluster handler for collecting metadata
+	SetClusterHandler(handler interface{})
```
davidor: I don't understand why this is `interface{}`.
We should be more specific. The implementation of this function calls `handler.(*pkgclusterchecks.Handler)` so I think that's the type we should use here.

comp/metadata/clusterchecks/impl/clusterchecks_test.go
```diff
@@ -0,0 +1,257 @@
...
+	// "github.com/stretchr/testify/assert"
+	// "github.com/stretchr/testify/require"
+	// "github.com/DataDog/datadog-agent/comp/core/autodiscovery/integration"
+	// logmock "github.com/DataDog/datadog-agent/comp/core/log/mock"
+	// "github.com/DataDog/datadog-agent/pkg/clusteragent/clusterchecks/types"
+	// configmock "github.com/DataDog/datadog-agent/pkg/config/mock"
+)
+
+// Tests are temporarily disabled while refactoring the clusterchecks component
+// TODO: Re-enable tests after component refactoring is complete
```
davidor: So should they be uncommented now?

comp/metadata/clusterchecks/impl/clusterchecks_test.go
```diff
@@ -0,0 +1,257 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+//go:build clusterchecks && test
```
davidor: I don't think we need `test` here. The file name ends with `_test.go`.

comp/metadata/clusterchecks/impl/clusterchecks.go
```diff
@@ -0,0 +1,313 @@
...
+//
+// In this case, the payload can't be split any further.
+func (p *Payload) SplitPayload(_ int) ([]marshaler.AbstractMarshaler, error) {
+	return nil, fmt.Errorf("could not split cluster checks payload any more, payload is too big for intake")
+}
+
+type clusterChecksImpl struct {
+	util.InventoryPayload
+
+	m sync.RWMutex
```
davidor: What is this mutex protecting? Is it the `clusterHandler`? If so, can we make it clearer with this env var name?

comp/metadata/clusterchecks/impl/clusterchecks.go
```diff
@@ -0,0 +1,313 @@
...
+}
+
+// NewComponent creates a new clusterchecks component
+func NewComponent(deps Requires) Provides {
+	// Get cluster identification
+	clusterName := clustername.GetClusterName(context.TODO(), "")
+	clusterID, err := clustername.GetClusterID()
+	if err != nil {
+		deps.Log.Debugf("Error retrieving cluster ID: %v", err)
+		clusterID = "" // Handle error gracefully like clusteragent
```
davidor, line 101: Just to clarify, does this mean we might send payloads with an empty cluster ID?
I’m not sure what the impact would be. Could that lead to data from different clusters getting mixed up if something fails?

comp/metadata/clusterchecks/impl/clusterchecks.go
```diff
@@ -0,0 +1,313 @@
...
+func (p *Payload) MarshalJSON() ([]byte, error) {
+	type PayloadAlias Payload
+	return json.Marshal((*PayloadAlias)(p))
+}
+
+// SplitPayload implements marshaler.AbstractMarshaler#SplitPayload.
+//
+// In this case, the payload can't be split any further.
+func (p *Payload) SplitPayload(_ int) ([]marshaler.AbstractMarshaler, error) {
+	return nil, fmt.Errorf("could not split cluster checks payload any more, payload is too big for intake")
```
davidor: I don't understand this error message.
It says that the payload is too big, so it can't be split ?

comp/metadata/clusterchecks/impl/clusterchecks.go
```diff
@@ -0,0 +1,313 @@
...
+	"github.com/DataDog/datadog-agent/pkg/util/kubernetes/clustername"
+	"github.com/DataDog/datadog-agent/pkg/util/uuid"
+)
+
+type metadata map[string]interface{}
+
+// State constants for cluster checks handler
+const (
+	// StateLeader indicates the cluster agent is the leader (NotRunning is empty)
+	StateLeader = ""
```
davidor: `StateFollower` and `StateNotReady` are not used, so they should be deleted.
`StateLeader` is only used in a comparison below:
```
// Check if we're the leader (NotRunning is empty)
return state.NotRunning == StateLeader
```
We can delete it as well and just compare with "". The comment already makes it clear.

comp/metadata/clusterchecks/def/component.go
```diff
@@ -0,0 +1,22 @@
...
+package clusterchecks
+
+import (
+	"net/http"
+)
+
+// team: container-platform
+
+// Component is the component interface for clusterchecks metadata
+type Component interface {
```
davidor, line 18: Not directly related to this PR, but I noticed that some components under `comp/metadata` expose different interfaces.
I’m not very familiar with them, but I wonder if it would make sense to standardize things (not in this PR).

comp/metadata/clusterchecks/def/component.go
zhuminyi: This is component interface and usually doesn't have build tag

comp/metadata/clusterchecks/impl/clusterchecks_test.go
zhuminyi: without clusterchecks component, it doesn't work at all. I will remove it for now.

comp/metadata/clusterchecks/impl/clusterchecks.go
zhuminyi: backend eventplatform has a logic to detect the case that both clustername and clusterid are missing before adding it to the database

comp/metadata/clusterchecks/impl/clusterchecks.go
zhuminyi: Metadata payloads are typically all-or-nothing. The error message is following other metadata payload pattern.

comp/metadata/clusterchecks/impl/clusterchecks.go
davidor: Would it make sense to avoid sending anything at all in the first place?
If we decide to keep this, I think it'd be good to add a comment explaining what you mentioned.

comp/metadata/clusterchecks/impl/clusterchecks.go
davidor: Thanks for clarifying. I don’t think the error message reflects what you described in your comment.

comp/metadata/clusterchecks/impl/clusterchecks.go
zhuminyi: yeah, I added skipping logic and updated the comment. So the logic is that we skip sending payload to the backedn if it is without identifier, but keep it in the local flare
```
	// Note: We intentionally generate payloads even when clustername or clusterID are empty
	// for visibility in flares/endpoints, but getPayloadAsMarshaler() will return nil
	// to prevent sending to backend when both identifiers are missing. Flares report is using
	// getAsJSON.
```

comp/metadata/clusterchecks/impl/clusterchecks.go
zhuminyi: isLeader is now returning state.NotRunning == ""

comp/metadata/clusterchecks/impl/clusterchecks.go
zhuminyi: updated

davidor (approved):

sarah-witt (approved):

Kyle-Neale (approved):

PR #39589, opened 2025-08-06
[OTAGENT-479]: Drop prometheus internal metrics

ogaca-dd:
### What does this PR do?
This PR removes some internal metrics for the auto configured prometheus:
 * `scrape_.*`
* `up`
* `promhttp_metric_handler_errors_total`


comp/otelcol/converter/impl/prometheus.go
```diff
@@ -176,20 +176,46 @@ func addPrometheusReceiver(conf *confmap.Conf, promServerAddr string) {
...
 			EnhancedName: ddExporterName,
 		}, pipelineName)
 	}
 }
 
+func getProcessorInternalPipeline() component {
+	name := "filter/drop-prometheus-internal-metrics"
+	return component{
+		Type:         "processors",
+		Name:         name,
```
songy23, line 202: I remember this field is supposed to be just the name of the package, i.e. filter in this case

songy23 (commented):
Probably you'll also need to fix the otel e2e tests, other than that LGTM

songy23 (approved):

PR #21935, opened 2024-01-08
(fleet) inject the agent install path at build time

arbll:
### What does this PR do?

This PR's goal is to make the install path `/opt/datadog-agent` something configurable in the agent code at build time.

### Motivation

This will allow us to build packages of the agent that can be installed in alternative paths (ex: `/opt/datadog/agent/v7.50/`). Building such packages is required for the upcoming remote updater.

### Additional Notes

- I have only dealt with linux for now as this is the platform we are starting with for the updater.
- Beyond this PR, we should never use the `/opt/datadog-agent` path outside of tests on linux.

PR #39595, opened 2025-08-06
upgrade(agent): Get the agent install path from its own executable

BaptisteFoy:
### What does this PR do?
https://github.com/DataDog/datadog-agent/pull/21935 introduced a way to have multiple builds of the agent with different paths. For Fleet Automation to work properly we need the agent path to be a bit more flexible. This PR removes the hardcoded install path, relying instead on the dynamic `os.Executable`. 

There will be a follow-up PR related to the `rpath`

### Motivation
Allow installation of agent plugins through Fleet Automation requires being able to have experiments of the same version with different plugins. This means that the same agent version must live twice on disk on different paths.

CI pipeline, manual QA with both the OCI & the Deb

### Additional Notes
I had to add an escape hatch for KMT tests, as the binary is not executed in /opt/datadog-agent but dependencies are there


BaptisteFoy:

BaptisteFoy:

BaptisteFoy:

BaptisteFoy:

BaptisteFoy:

BaptisteFoy:

BaptisteFoy:

BaptisteFoy:

BaptisteFoy:

arbll (approved):

hush-hush (approved):

test/new-e2e/system-probe/test/micro-vm-init.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 set -eEuo pipefail
 
+export DD_TEST_INSTALL_PATH_OVERRIDE="/opt/datadog-agent"
```
aiuto, line 7: Is this where it would install anyway?
It appears that nothing else in the test changed, which makes me think that not setting this would still work.

aiuto (commented):
The code looks sound overall, but I wonder if the test is complete.

test/new-e2e/system-probe/test/micro-vm-init.sh
BaptisteFoy: Unfortunately it's not the case, as you can see for instance in [this job](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1068728480#L2197) ([pipeline](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/72952440)).

The error is ```cannot load program: runtime compilation failed: error getting helper availability: get available helpers: preprocess helpers: clang: fork/exec /opt/kmt-ramfs/system-probe-tests/pkg/network/embedded/bin/clang-bpf: no such file or directory```; meaning that the test itself is executed from `/opt/kmt-ramfs/system-probe-tests/pkg/network/<something>/<something_else>`, but the `clang-bpf` binary is in `/opt/datadog-agent`.

With this PR's change the search path for the `clang-bpf` binary changes unless we force it with this environment variable. This was the simplest approach I could find, as many KMT tests in many places failed with the same error but a different execution path.

---

As a side note, the agent can be installed elsewhere today when using Remote Agent Management (`/opt/datadog-packages/datadog-agent/<version>`); and this PR & the rpath one mentioned in the description will allow it to be installed in semi-arbitrary directories in `/opt/datadog-packages/datadog-agent`

gjulianm: This makes sense to me, as the test binaries are running from a different path as the agent would usually do. However, can we add a comment explaining why we need this line?

gjulianm (approved):
LGTM, I'd suggest just adding the comment I mentioned to explain why we need that env var

tasks/libs/common/utils.py
```diff
@@ -246,10 +246,6 @@ def get_build_flags(
 
     rtloader_lib, rtloader_headers, rtloader_common_headers = get_rtloader_paths(embedded_path, rtloader_root)
 
-    # setting the install path, allowing the agent to be installed in a custom location
-    if sys.platform.startswith('linux') and install_path:
```
KevinFairise2, line 250: If we remove that part is the `install_path` flag of that function still behaving as we expect it? I feel like we can now no longer decide the install path at build time, and only at runtime with an environment variable?

KevinFairise2 (approved):

tasks/libs/common/utils.py
BaptisteFoy: The `init` function forces us to compute the binary path at runtime, so even if we keep this part it will be overwritten

KevinFairise2: I guess it means that it is no longer used and there is no risk that we break existing things by removing it, right?

BaptisteFoy: Yep

PR #39598, opened 2025-08-06
[incident-41677] Disable docker_build cache

CelianR:

### DIFF
--- a/.gitlab/container_build/docker_linux.yml
+++ b/.gitlab/container_build/docker_linux.yml
@@ -43,6 +43,8 @@
         # that can't be usefully cached, to the remote cache.
         echo "Building ${CACHE_TARGET} for cache"
         docker buildx build --pull --platform linux/$ARCH \
+          --cache-from="" \
+          --cache-to="" \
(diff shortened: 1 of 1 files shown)


.gitlab/container_build/docker_linux.yml
```diff
@@ -43,9 +43,7 @@
         # that can't be usefully cached, to the remote cache.
         echo "Building ${CACHE_TARGET} for cache"
         docker buildx build --pull --platform linux/$ARCH \
-          ${CACHE_SOURCE} \
-          ${CACHE_TO} \
-          ${DOCKER_NO_CACHE} \
+          --no-cache \
```
alopezz: Rather than doing this, I think it might be better to set the environment variables right before this, that is:
```
CACHE_TO=""
CACHE_SOURCE=""
```

You could also set `DOCKER_NO_CACHE=--no-cache` though I don't think that's really necessary as I believe the remote caching to be a more likely culprit than the local cache, so I'd probably try not changing that.

PR #39606, opened 2025-08-06
Initial version of MODULE.bazel

aiuto:
### What does this PR do?

This create the minimal MODULE.bazel and adds dependency on zlib.

The rule sets are added no so that we can swarm on individual new pieces of code without hitting merge conflicts on MODULE.bazel.

It does not modify CI at this time.

Try to build something. On both linux and macos. Didn't do windows yet.

```
bazel build @zlib//:all
ar tv   bazel-bin/external/zlib+/libz.a
```

### Tickets
- https://datadoghq.atlassian.net/browse/ABLD-130
- https://datadoghq.atlassian.net/browse/ABLD-139

NOTE: This is a reland of https://github.com/DataDog/datadog-agent/pull/39570.  The commits did not verify in that, so it's easier to restart from the final version of that.


.bazelversion
```diff
@@ -0,0 +1 @@
+8.2.1
```
JSGette: #ff2r

out of curiousity: why not 8.3.1?
https://github.com/bazelbuild/bazel/releases/tag/8.3.1

.gitignore
```diff
@@ -25,6 +25,9 @@ venv*/
 *.log
 /main
 
+# Bazel output folders
```
rdesgroppes:
```suggestion
# Bazel convenience symlinks to output folders
```

.bazelversion
aiuto: Good catch. I had that in the other PR, but my reconstruction failed.
Updated.

.gitignore
aiuto: I wrote "output folders" because of Ofek's question in the first version of the PR where he asked if they were always symlinks - or are they potentially real folders on Windows. I was hedging. But, even if they are real on some windows installs, saying symlink is more often correct for most people..

ofek (approved):

KevinFairise2 (approved):

PR #39608, opened 2025-08-06
[CNM-4626] Add ICMP Mode to dynamic paths

pimlu:
### What does this PR do?
This PR adds "ICMP Mode" to dynamic paths, which replaces TCP/UDP pathtests with ICMP for networks that block everything else but ICMP.

You can enable it as `datadog.collector.icmp_mode`:
1. `none`: Default setting, no changes
2. `tcp`: Change TCP connections into ICMP pathtests
3. `udp`: Change UDP connections into ICMP pathtests
4. `all`: Change all connections into ICMP pathtests

### Motivation
Customer encountered this issue with their firewalls

New tests should pass (checks that connections are scheduled correctly, and the different ICMP modes).


akarpz (approved):

pkg/config/setup/config.go
```diff
@@ -493,6 +493,8 @@ func InitConfig(config pkgconfigmodel.Setup) {
 	config.BindEnvAndSetDefault("network_path.collector.disable_intra_vpc_collection", false)
 	config.BindEnvAndSetDefault("network_path.collector.source_excludes", map[string][]string{})
 	config.BindEnvAndSetDefault("network_path.collector.dest_excludes", map[string][]string{})
+	config.SetKnown("network_path.collector.tcp_method")
```
pimlu: This already was in use but I forgot to call `SetKnown` on it, so there were spurious warnings.

hush-hush: Please use `BindEnvAndSetDefault` instead, we want a default for all settings going forward

hush-hush (changes requested):

hush-hush (approved):

PR #39619, opened 2025-08-07
[ACIX-616] Update e2e cleanup for remote stack cleaning

CelianR:
### What does this PR do?

We have issues with this job ([error](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1067208689)). Updating this to:

1. Fix issues we currently have (cannot serialize invoke context, cannot use local function to map)
2. Add back the old way of cleaning in case we need to disable stack cleaner

Example working [job](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1068840832).


.gitlab/e2e/e2e.yml
```diff
@@ -1046,11 +1046,19 @@ new-e2e-cleanup-on-failure:
...
     - PULUMI_CONFIG_PASSPHRASE=$($CI_PROJECT_DIR/tools/ci/fetch_secret.sh $AGENT_QA_E2E pulumi_passphrase) || exit $?; export PULUMI_CONFIG_PASSPHRASE
     - pulumi login "s3://dd-pulumi-state?region=us-east-1&awssdk=v2&profile=$AWS_PROFILE"
-    - dda inv -- -e new-e2e-tests.cleanup-remote-stacks --stack-regex "ci-init-$CI_PIPELINE_ID-.*" --pulumi-backend=dd-pulumi-state
+    - |
+      if [ "$REMOTE_STACK_CLEANING" = "true" ]; then
+        # If destroy locally, check only the eks stacks
+        STACK_REGEX="ci-init-$CI_PIPELINE_ID.*eks.*"
+      else
+        STACK_REGEX="ci-init-$CI_PIPELINE_ID-.*"
+      fi
```
KevinFairise2, line 1071: Why do we need that if?

KevinFairise2 (approved):

.gitlab/e2e/e2e.yml
CelianR: I prefer to keep the previous behaviour in case we need to disable stack cleaner. I can remove that once it is fully migrated

PR #39622, opened 2025-08-07
Use resolved `git` hashes in Omnibus environment

rdesgroppes:
### What does this PR do? / Motivation
This is aimed at reducing the number of `Omnibus git cache mutation` alerts triggered when using a "floating" `git` ref such as a branch name.

I ran the script with hardcoded overrides denoting "moving" refs of external repositories (i.e. `main`, `master`, `datadog-5.5.0`): https://github.com/DataDog/datadog-agent/commit/727aa50ea4a01f1180f378441d5eb1ec22bc988f#diff-0a1a69df7ccd1e457863118f2c0908aec6b4b3ad15cf698562bdaeec824aba01R132
The behavior was as expected: ([example](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1069081597))
```sh
$ dda inv -- -e omnibus.build --base-dir $OMNIBUS_BASE_DIR --skip-deps --go-mod-cache="$GOPATH/pkg/mod" --system-probe-bin=/tmp/system-probe --flavor "$FLAVOR" --config-directory "$CONFIG_DIR" --install-directory "$INSTALL_DIR"
/go/src/github.com/DataDog/datadog-agent/tasks/omnibus.py:142: UserWarning: Multiple candidates for https://github.com/DataDog/integrations-core.git@master: ('refs/heads/external-pr/Yufeireal/master', 'refs/heads/external-pr/footprint-it-solutions/master', 'refs/heads/master') - choosing shortest: refs/heads/master
```

None foreseen.

### Additional Notes

### DIFF
--- a/tasks/omnibus.py
+++ b/tasks/omnibus.py
@@ -118,12 +118,28 @@ def get_omnibus_env(
     if go_mod_cache:
         env['OMNIBUS_GOMODCACHE'] = go_mod_cache
 
-    env_override = ['INTEGRATIONS_CORE_VERSION', 'OMNIBUS_RUBY_VERSION']
-    for key in env_override:
+    external_repos = {
+        "INTEGRATIONS_CORE_VERSION": "https://github.com/DataDog/integrations-core.git",
+        "OMNIBUS_RUBY_VERSION": "https://github.com/DataDog/omnibus-ruby.git",
+    }
+    for key, url in external_repos.items():
         value = os.environ.get(key)
         # Only overrides the env var if the value is a non-empty string.
         if value:
             env[key] = value
+        ref = env[key]
+        if not re.fullmatch(r"[0-9a-f]{4,40}", ref):  # resolve only "moving" refs, such as `own/branch`
+            candidates = [line.split() for line in ctx.run(f"git ls-remote --refs {url} '{ref}'").stdout.splitlines()]
+            if not candidates:
+                warnings.warn(f"No candidate for {url}@{ref} - leaving untouched", stacklevel=1)
+                continue
+            sha1, shortest_ref = min(candidates, key=lambda c: len(c[1]))
+            if len(candidates) > 1:  # happens when a branch name mimics its base or target, such as `my/own/branch`
+                warnings.warn(
+                    f"Multiple candidates for {url}@{ref}: {[c[1] for c in candidates]} - choosing shortest: {shortest_ref} -> {sha1}",
+                    stacklevel=1,
+                )
+            env[key] = sha1
 
     if sys.platform == 'darwin':
         env['MACOSX_DEPLOYMENT_TARGET'] = '11.0' if os.uname().machine == "arm64" else '10.12'


tasks/omnibus.py
```diff
@@ -121,9 +122,9 @@ def get_omnibus_env(
     env_override = ['INTEGRATIONS_CORE_VERSION', 'OMNIBUS_RUBY_VERSION']
     for key in env_override:
         value = os.environ.get(key)
-        # Only overrides the env var if the value is a non-empty string.
+        # Only overrides the env var taking the effective SHA-1 if the value is a non-empty string.
         if value:
-            env[key] = value
+            env[key] = subprocess.check_output(("git", "rev-parse", "--verify", value), text=True).rstrip()
```
alopezz: This only resolves the reference if the value got through the environment, and not via release.json (from [line 99](https://github.com/DataDog/datadog-agent/blob/b1469e6cec26d68520c8da0b61b269b77b2bca80/tasks/omnibus.py#L99)), I think.

alopezz: Maybe something like this could work here:

```python
    for key in env_override:
        value = os.environ.get(key) or env[key]
        env[key] = subprocess.check_output(("git", "rev-parse", "--verify", value), text=True).rstrip()
```

Though in that case I would probably also rename `env_override` to something else (`external_repo_refs`?).

alopezz: Now on second thought, no, I think this doesn’t work, because these references are for external repos, not the current one.

rdesgroppes: Fixed and fixed.

tasks/omnibus.py
```diff
@@ -118,12 +119,30 @@ def get_omnibus_env(
...
+    for key, url in external_repos.items():
         value = os.environ.get(key)
         # Only overrides the env var if the value is a non-empty string.
         if value:
             env[key] = value
+        ref = env[key]
+        if not re.fullmatch(r"[0-9a-f]{4,40}", ref):  # resolve only "moving" refs, such as `own/branch`
+            candidates = tuple(
+                line.split() for line in subprocess.check_output(("git", "ls-remote", url, ref), text=True).splitlines()
+            )
```
alopezz: Why `tuple`?

```suggestion
            candidates = [
                line.split() for line in subprocess.check_output(("git", "ls-remote", url, ref), text=True).splitlines()
            ]
```

tasks/omnibus.py
```diff
@@ -118,12 +119,30 @@ def get_omnibus_env(
...
+    }
+    for key, url in external_repos.items():
         value = os.environ.get(key)
         # Only overrides the env var if the value is a non-empty string.
         if value:
             env[key] = value
+        ref = env[key]
+        if not re.fullmatch(r"[0-9a-f]{4,40}", ref):  # resolve only "moving" refs, such as `own/branch`
+            candidates = tuple(
+                line.split() for line in subprocess.check_output(("git", "ls-remote", url, ref), text=True).splitlines()
```
alopezz: For consistency we should probably use `ctx.run` to manage subprocesses. Also, adding [--refs](https://git-scm.com/docs/git-ls-remote#_output) looks somewhat safer.

```suggestion
                line.split() for line in ctx.run(f'git ls-remote --refs "{url}" "ref"').stdout.splitlines()
```

alopezz (approved):
I'm mostly fine with this. It would be even better with a couple of tests 😉 but I'll leave that up to you.

tasks/omnibus.py
rdesgroppes: Indeed arguable => applying your suggestion, thanks!

tasks/omnibus.py
rdesgroppes: Hum, seems I shouldn't have moved to `ctx.run`:
- #40227

PR #39624, opened 2025-08-07
workloadmeta: add support for container owner in protobuf (and remote collector)

paulcacheux:
### What does this PR do?

This PR adds support for the container owner field in the protobuf format allowing the remote workloadmeta store to access this information. The main goal of this is to allow remote consumers to go from the container to the owning pod and read some pod specific data (like pod annotations or namespace).


gabedos (approved):
lgtm

PR #39632, opened 2025-08-07
[AGENTRUN-615] Enable Cluster CA-Signed IPC Certificate

misteriaud:
### What does this PR do?
This PR adds support for loading a Cluster CA to enable TLS trust chains for cross-node communication. Specifically, it:

- Signs the IPC certificate with the Cluster CA, establishing a cross-node trust chain.
- If `cluster_trust_chain.enable_tls_verification` is set to `true`, uses the Cluster CA for DCAClient and CLCClient verification.
- Introduces two new configuration entries to specify the Cluster CA certificate and key file paths:
  - `cluster_trust_chain.ca_cert_file_path`
  - `cluster_trust_chain.ca_key_file_path`
- When the Cluster CA is configured, adds additional SANs to the IPC certificate so clients can be properly identified (uses `configutils.GetClusterAgentEndpoint` for DCA and `clc_runner_host` for CLC runner).
- The NodeAgent does not require additional SANs, as it has no external API consumers.

### Motivation
Enabling IPC certificate signing with the Cluster CA allows secure, verified TLS communication between nodes, enhancing trust and security for cross-node communications.

### Validation
- Increased unit test coverage for `pkg/api/security/cert`.
- Tested in a local kind cluster with NodeAgent, ClusterAgent, and CLCRunner enabled with the following scenarios

#### Client Verification enabled & Server self-signing IPC cert
Expectation: failure due to bad TLS handshake
Results:
| Client \\ Server | NodeAgent | CLCRunner | DCA     |
| ---------------- | --------- | --------- | ------- |
| NodeAgent        | x         | x         | failure |
| CLCRunner        | x         | x         | failure |
| DCA              | x         | failure   | x       |

#### Client Verification enabled & Server signing IPC cert with Cluster CA
Expectation: success
Results:
| Client \\ Server | NodeAgent | CLCRunner | DCA     |
| ---------------- | --------- | --------- | ------- |
| NodeAgent        | x         | x         | Success |
| CLCRunner        | x         | x         | Success |
| DCA              | x         | Success   | x       |

### Additional Notes
A e2e test will be implemented to cover these changes in a following PR.


hush-hush (approved):
Looks good for Agent-Configuration

comp/core/ipc/impl/ipc_test.go
```diff
@@ -112,3 +123,358 @@ func TestDeadline(t *testing.T) {
 	assert.Error(t, err)
 	assert.LessOrEqual(t, duration, mockConfig.GetDuration("auth_init_timeout")+time.Second)
 }
+
+// createTestCA creates a test CA certificate and private key files for testing cluster trust chain
+func createTestCA(t *testing.T) (string, string, *x509.Certificate) {
+	// Generate CA private key
```
GustavoCaso: Should we add a call to `t.Helper()` in this function?

comp/core/ipc/impl/ipc_test.go
```diff
@@ -112,3 +123,358 @@ func TestDeadline(t *testing.T) {
...
+
+	err = os.WriteFile(caKeyPath, caPrivKeyPEM, 0600)
+	require.NoError(t, err)
+
+	return caCertPath, caKeyPath, caCert
+}
+
+// setupBasicIPCConfig creates a basic IPC configuration for testing
+func setupBasicIPCConfig(t *testing.T) model.Config {
+	mockConfig := configmock.New(t)
```
GustavoCaso, line 183: Same comment about `t.Helper()` 😄

pkg/api/security/cert/cert_getter.go
```diff
@@ -66,16 +72,112 @@ func (certificateFactory) Deserialize(raw []byte) (Certificate, error) {
...
+
+	clientConfig, serverConfig, err := GetTLSConfigFromCert(cert.cert, cert.key)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("error while setting TLS configs: %w", err)
+	}
+	return clientConfig, serverConfig, clusterClientConfig, err
+}
+
+// ReadClusterCA reads the cluster CA certificate and key from the given path
+func ReadClusterCA(caCertPath, caKeyPath string) (*x509.Certificate, any, error) {
```
GustavoCaso: Should we move this function to `pkg/api/security/cert/cert_getter_dca.go`?

pkg/api/security/cert/cert_getter.go
```diff
@@ -66,16 +72,112 @@ func (certificateFactory) Deserialize(raw []byte) (Certificate, error) {
...
+	caKeyPEM, err := os.ReadFile(caKeyPath)
+	if err != nil {
+		return nil, nil, fmt.Errorf("unable to read cluster CA key file: %w", err)
+	}
+
+	// Parse the cluster CA cert
+	block, _ := pem.Decode(caCertPEM)
+	if block == nil {
+		return nil, nil, fmt.Errorf("unable to decode cluster CA cert PEM")
+	}
```
GustavoCaso: Do you think we should return error message if any?

pkg/api/security/cert/cert_getter.go
```diff
@@ -66,16 +72,112 @@ func (certificateFactory) Deserialize(raw []byte) (Certificate, error) {
...
+	caCert, err = x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		return nil, nil, fmt.Errorf("unable to parse cluster CA cert file: %w", err)
+	}
+
+	// Parse the cluster CA key
+	block, _ = pem.Decode(caKeyPEM)
+	if block == nil {
+		return nil, nil, fmt.Errorf("unable to decode cluster CA key PEM")
+	}
```
GustavoCaso: Similar comment about returning the error information

pkg/api/security/cert/cert_getter.go
```diff
@@ -66,16 +72,112 @@ func (certificateFactory) Deserialize(raw []byte) (Certificate, error) {
...
+
+	var caPrivKey any
+
```
GustavoCaso: Nit: To improve readability, I would create a dedicated error variable that is used in the `switch` statement. 

```
var caParseErr error
```

That way we are not reusing the `err` variable from above 😄

pkg/api/security/cert/cert_getter_dca.go
```diff
@@ -0,0 +1,136 @@
...
+	caCert                *x509.Certificate
+	caPrivKey             any
+}
+
+// readClusterCAConfig reads cluster CA configuration and files from disk once
+// Returns nil if no cluster CA is configured
+func readClusterCAConfig(config configModel.Reader) (*clusterCAData, error) {
+	enableTLSVerification := config.GetBool("cluster_trust_chain.enable_tls_verification")
+	clusterCAPath := config.GetString("cluster_trust_chain.ca_cert_file_path")
+	clusterCAKeyPath := config.GetString("cluster_trust_chain.ca_key_file_path")
```
GustavoCaso, line 85: We are checking if either of them is empty to early return, but what if one is empty and the other is not? Would the function `ReadClusterCA(clusterCAPath, clusterCAKeyPath)` work as expected?

pkg/api/security/cert/cert_getter_dca.go
```diff
@@ -0,0 +1,136 @@
...
+		enableTLSVerification: enableTLSVerification,
+		caCert:                caCert,
+		caPrivKey:             caPrivKey,
+	}, nil
+}
+
+// buildClusterClientTLSConfig creates the TLS configuration for cluster client communication
+// using pre-read cluster CA data
+func (c *clusterCAData) buildClusterClientTLSConfig() (*tls.Config, error) {
+	// Default to insecure configuration
```
GustavoCaso, line 110: Nit: I would place the early return condition first 

```go
if !c.enableTLSVerification {
    return &tls.Config{
		InsecureSkipVerify: true,
	}
}

...
```

That way is easier to read 😄

pkg/api/security/cert/cert_getter_dca.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+
+		serverHost, _, err = net.SplitHostPort(parsedURL.Host)
+		if err != nil {
+			return fmt.Errorf("unable to get pod IP from cluster agent endpoint: %w", err)
+		}
+	}
+
+	// If the process is a CLC Runner, add the CLC Runner host to the SANs
+	if isCLC {
```
GustavoCaso: Can the cluster agent be the CLC runner as well? I imagine yes, but just double-checking for my own learning 😄

pkg/config/setup/config.go
```diff
@@ -596,8 +596,14 @@ func InitConfig(config pkgconfigmodel.Setup) {
 	config.BindEnvAndSetDefault("cluster_agent.kube_metadata_collection.resources", []string{})
 	config.BindEnvAndSetDefault("cluster_agent.kube_metadata_collection.resource_annotations_exclude", []string{})
 	config.BindEnvAndSetDefault("cluster_agent.cluster_tagger.grpc_max_message_size", 4<<20) // 4 MB
-	// Enable TLS verification for Agent cross-node communications (NodeAgent->DCA / CLC->DCA / DCA->CLC).
-	config.BindEnvAndSetDefault("cluster_agent.enable_tls_verification", false)
```
GustavoCaso, line 605: Just double-checking that we are not breaking production for any customer. This key is only used internally, right? Mostly to assess that the rename can be perform with a breaking change 😄

GustavoCaso (approved):
Amazing work 🎉 

I left a few suggestions but they are not blockers 😄

pkg/api/security/cert/cert_getter.go
misteriaud: The thing is that `pem.Decode` does not return any parsing error if it's unable to parse the PEM buffer. So we can't return any more details.

pkg/api/security/cert/cert_getter.go
misteriaud: Same answer as above, no error returned by `pem.Decode(caKeyPEM)`

pkg/api/security/cert/cert_getter_dca.go
misteriaud: `readClusterCA` needs to have access  both cert and key files to work.

pkg/api/security/cert/cert_getter_dca.go
misteriaud: Excellent suggestion, indeed it's much easier to see the flow !

pkg/api/security/cert/cert_getter_dca.go
misteriaud: Good point ! The CLC runner is basically a NodeAgent with only purpose to run checks, so it can't be a ClusterAgent.
But I didn't covered the scenario where the ClusterAgent runs with the config `clc_runner_enabled: true`. I suppose this configuration would be ignored by the rest of the code, so I just replace the `if` by an `else if` because both are incompatibles. 
Updated in fcfb1f889fe27867eb66c1ffbf61fce7b3d5cbba.

comp/core/ipc/impl/ipc_test.go
misteriaud: Yes ! Done in 763d828da41b106a850c4ea43aa6d9bf2fecf8c0.

comp/core/ipc/impl/ipc_test.go
misteriaud: Yes ! Done in 763d828da41b106a850c4ea43aa6d9bf2fecf8c0.

pkg/api/security/cert/cert_getter.go
misteriaud: Indeed, done in fcfb1f889fe27867eb66c1ffbf61fce7b3d5cbba.

pkg/api/security/cert/cert_getter.go
misteriaud: Good suggestion, done in fcfb1f889fe27867eb66c1ffbf61fce7b3d5cbba.

pkg/api/security/cert/cert_getter_dca.go
misteriaud: Done in fcfb1f889fe27867eb66c1ffbf61fce7b3d5cbba.

pkg/config/setup/config.go
misteriaud: This config has been introduced in a PR I merged two weeks ago (#39404), so no customer used it yet.

pkg/api/security/cert/cert_getter_dca.go
GustavoCaso: If that is the case, I would make it the condition like this:

```go
// If no cluster CA path is configured, return nil (not an error)
	if !(clusterCAPath != "" && clusterCAKeyPath != "") {
		return &clusterCAData{
			enableTLSVerification: enableTLSVerification,
		}, nil
	}

    ....
```

wdhif (approved):

PR #5010, opened 2020-02-27
[omnibus] Fix MacOS build conflicting with system Python 3

KSerrania:
### What does this PR do?

Work around an implementation detail of Python 3 on MacOS making the build fail (because the embedded pip3 tries to use the system python3 instead of the embedded one).

### Motivation

Make the MacOS build work with Python 3 installed on the system.

### Full explanation

On MacOS, python3 does an ugly hack to circumvent the limits imposed on processes by Apple on framework builds (which is the case when installing python3 with brew).

This hack consists in setting `__PYVENV_LAUNCHER__` to the location of the system python3 executable, and using it instead of `argv[0]` (which is set by the OS). The problem is that this variable is not unset after it's read, so it's still set in all commands launched by omnibus.

In particular, it's set during the pip installation (in `pip3.rb`), which makes the `/opt/datadog-agent/embedded/bin/python3 setup.py install --prefix=/opt/datadog-agent/embedded` command link pip3 to the system python instead of the embedded python (since that's what the `__PYVENV_LAUNCHER__` var is pointing at), bypassing the `--prefix` option and making the build fail.

More info here: https://stackoverflow.com/questions/26323852/whats-the-meaning-of-pyvenv-launcher-environment-variable

This might be fixed by: https://github.com/pypa/virtualenv/pull/1648, so we should check again if this hack is still needed in future releases.

PR #39633, opened 2025-08-07
Revert "Fix MacOS build conflicting with system Python 3"

rdesgroppes:
### What does this PR do? / Motivation
The present change is a follow-up of #39563, where it was found that:
> We need to have a default value here because these unit tests are also called from `macos` runners  but [`system` is patched as `linux`](https://github.com/DataDog/datadog-agent/pull/39563/files#diff-407acb07e932b50660fc4999612d12a05d6046a1716a53125e1010050c898cebR33). So even though the unit tests are run in macos runners(run on EC2) they try to access vault and dont have K8 `POD_NAMESPACE` env var set.

It turns out the temporary fix introduced by #5010 in 2020 is no longer needed:
- upstream bug report: https://github.com/pypa/virtualenv/issues/1643
- upstream fix: https://github.com/pypa/virtualenv/pull/1648
- released with `virtualenv` v20.0.5: https://virtualenv.pypa.io/en/latest/changelog.html#v20-0-5-2020-02-21

This therefore reverts commit d16a8a4da3baadc1ada4636513476e5b3df169cb: "[omnibus] Fix MacOS build conflicting with system Python 3 (#5010)", and adjusts consequences accordingly:
 - the `POD_NAMESPACE` environment variable must be present (no default) for `linux`, and only for `linux`,
- patches in tests get tailored to the very needs, with 3 `platforms` to cover implying 3 "sub" tests,
- covering `win32` unearthes unmatched command patterns due to suffixes (`.bat`, `.exe`) being used for certain commands (like `aws.exe` and `omnibus.bat`, unlike `git`) which requires further adaptation, especially to understand which patterns are not found.

### DIFF
--- a/tasks/omnibus.py
+++ b/tasks/omnibus.py
@@ -36,14 +36,7 @@ def omnibus_run_task(ctx, task, target_project, base_dir, env, log_level="info",
         if host_distribution:
             overrides_cmd += f" --override=host_distribution:{host_distribution}"
 
-        omnibus = "bundle exec omnibus"
-        if sys.platform == 'win32':
-            omnibus = "bundle exec omnibus.bat"
-        elif sys.platform == 'darwin':
-            # HACK: This is an ugly hack to fix another hack made by python3 on MacOS
-            # The full explanation is available on this PR: https://github.com/DataDog/datadog-agent/pull/5010.
-            omnibus = "unset __PYVENV_LAUNCHER__ && bundle exec omnibus"
-
+        omnibus = f"bundle exec {"omnibus.bat" if sys.platform == "win32" else "omnibus"}"
         cmd = "{omnibus} {task} {project_name} --log-level={log_level} {overrides}"
         args = {
             "omnibus": omnibus,
--- a/tasks/omnibus.py
+++ b/tasks/omnibus.py
@@ -30,11 +30,11 @@ from tasks.libs.releasing.version import get_version, load_dependencies
 
 def omnibus_run_task(ctx, task, target_project, base_dir, env, log_level="info", host_distribution=None):
     with ctx.cd("omnibus"):
-        overrides_cmd = ""
+        overrides = []
         if base_dir:
-            overrides_cmd = f"--override=base_dir:{base_dir}"
+            overrides.append(f"--override=base_dir:{base_dir}")
         if host_distribution:
-            overrides_cmd += f" --override=host_distribution:{host_distribution}"
+            overrides.append(f"--override=host_distribution:{host_distribution}")
 
         omnibus = f"bundle exec {"omnibus.bat" if sys.platform == "win32" else "omnibus"}"
         cmd = "{omnibus} {task} {project_name} --log-level={log_level} {overrides}"
@@ -43,7 +43,7 @@ def omnibus_run_task(ctx, task, target_project, base_dir, env, log_level="info",
             "task": task,
             "project_name": target_project,
             "log_level": log_level,
-            "overrides": overrides_cmd,
+            "overrides": " ".join(overrides),
         }
 
         with gitlab_section(f"Running omnibus task {task}", collapsed=True):
--- a/tasks/unit_tests/omnibus_tests.py
+++ b/tasks/unit_tests/omnibus_tests.py
@@ -87,11 +87,20 @@ class TestOmnibusCache(unittest.TestCase):
         """Assert the given line patterns appear in the given order in `msg`."""
         commands = _run_calls_to_string(self.mock_ctx.run.mock_calls)
         # Match patterns independently to avoid pitfalls of regex merging while enabling precise mismatch reporting
+        unmatched_patterns = []
+        pos = 0
         for pattern in line_patterns:
-            self.assertIsNotNone(
-                re.search(pattern, commands, re.MULTILINE),
-                f'Failed to match pattern `{pattern}` among {pformat(commands)}',
-            )
+            match = re.search(pattern, commands[pos:], re.MULTILINE)
+            if match:
+                pos += match.end()
+            else:
+                unmatched_patterns.append(pattern)
+        if unmatched_patterns:
+            self.fail(f"""Failed to match patterns in order:
+{pformat(unmatched_patterns)}
+... among commands:
+{pformat(commands)}
+""")
 
     @_for_each_platform
     def test_successful_cache_hit(self):
--- a/tasks/unit_tests/omnibus_tests.py
+++ b/tasks/unit_tests/omnibus_tests.py
@@ -2,6 +2,7 @@ import functools
 import os
 import re
 import unittest
+from pprint import pformat
 from unittest import mock
 
 from invoke.context import MockContext
@@ -61,6 +62,7 @@ class TestOmnibusCache(unittest.TestCase):
                     mock.patch("sys.platform", platform),
                     mock.patch.dict(os.environ, get_dd_api_key_env),
                 ):
+                    self.setUp()  # because `unittest` doesn't offer a subtest setup hook yet
                     test_func(self, *args, **kwargs)
 
         return wrapper
@@ -81,15 +83,15 @@ class TestOmnibusCache(unittest.TestCase):
         for pattern, result in patterns:
             self.mock_ctx.set_result_for('run', re.compile(pattern), result)
 
-    def assertRunLines(self, line_patterns):
+    def assertRunLines(self, *line_patterns):
         """Assert the given line patterns appear in the given order in `msg`."""
         commands = _run_calls_to_string(self.mock_ctx.run.mock_calls)
-
-        pattern = '(\n|.)*'.join(line_patterns)
-        return self.assertIsNotNone(
-            re.search(pattern, commands, re.MULTILINE),
-            f'Failed to match pattern {line_patterns}.',
-        )
+        # Match patterns independently to avoid pitfalls of regex merging while enabling precise mismatch reporting
+        for pattern in line_patterns:
+            self.assertIsNotNone(
+                re.search(pattern, commands, re.MULTILINE),
+                f'Failed to match pattern `{pattern}` among {pformat(commands)}',
+            )
 
     @_for_each_platform
     def test_successful_cache_hit(self):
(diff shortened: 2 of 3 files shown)


tasks/unit_tests/omnibus_tests.py
```diff
@@ -30,7 +32,6 @@ def _run_calls_to_string(mock_calls):
     return '\n'.join(commands_run)
 
 
-@mock.patch('sys.platform', 'linux')
```
alopezz, line 33: I think this may make tests fail on windows. As long as the code's behavior depends on this global variable it becomes an input to tests and thus it's better brought under control.

rdesgroppes: The problems I see:
1. `@mock.patch('sys.platform', 'linux')` prevents from covering `darwin` and `win32` code paths,
2. contributors are confused leading them to workarounds such as `os.environ.get('POD_NAMESPACE', 'gitlab-runner')` instead of the intended `os.environ["POD_NAMESPACE"]` - so was I.

Unless I'm mistaken, tests pass on Windows. Do you foresee only a subset of tests was run and we'll face a problem after merging to the mainline?

alopezz: > Unless I'm mistaken, tests pass on Windows.

You can try by setting the platform here to 'win32' 😄 .  But I think [this assertion](https://github.com/DataDog/datadog-agent/blob/cb1684d631a4f043af5c526c32e19a6ea6d72e97/tasks/unit_tests/omnibus_tests.py#L148) would break because [this command](https://github.com/DataDog/datadog-agent/blob/cb1684d631a4f043af5c526c32e19a6ea6d72e97/tasks/omnibus.py#L39) changes.

> @mock.patch('sys.platform', 'linux') prevents from covering darwin and win32 code paths,

On the contrary, this allows us to cover those other 2 paths explicitly, and to test all 3 regardless of the platform where tests are running if we do so. Again, it's about controlling the inputs of your code in the test.

> contributors are confused leading them to workarounds such as os.environ.get('POD_NAMESPACE', 'gitlab-runner') instead of the intended os.environ["POD_NAMESPACE"] - so was I.

That doesn't seem related to this. What that calls for is more isolation (i.e. patching those env vars), not less.

rdesgroppes: Wow, I seem to be completely lost here. How setting `sys.platform = "linux"` does cover:
```py
    if sys.platform == 'win32':
        #              ^ this?
        cmd = f'aws.exe ssm get-parameter --region us-east-1 --name {os.environ["API_KEY_ORG2"]} --with-decryption --query "Parameter.Value" --out text'
    elif sys.platform == 'darwin':
        #                ^ and that?
        cmd = f'vault kv get -field=token kv/aws/arn:aws:iam::486234852809:role/ci-datadog-agent/{os.environ["AGENT_API_KEY_ORG2"]}'
    else:
        # only covered part, isn't it?
        cmd = f'vault kv get -field=token kv/k8s/{os.environ["POD_NAMESPACE"]}/datadog-agent/{os.environ["AGENT_API_KEY_ORG2"]}'
```

tasks/omnibus.py
```diff
@@ -36,14 +36,7 @@ def omnibus_run_task(ctx, task, target_project, base_dir, env, log_level="info",
...
 
-        omnibus = "bundle exec omnibus"
-        if sys.platform == 'win32':
-            omnibus = "bundle exec omnibus.bat"
-        elif sys.platform == 'darwin':
-            # HACK: This is an ugly hack to fix another hack made by python3 on MacOS
-            # The full explanation is available on this PR: https://github.com/DataDog/datadog-agent/pull/5010.
-            omnibus = "unset __PYVENV_LAUNCHER__ && bundle exec omnibus"
-
+        omnibus = f"bundle exec omnibus{".bat" if sys.platform == "win32" else ""}"
```
aiuto: I'm having trouble parsing this - especially the ""}" at the end.  I can figure it out, but maybe just the simple thing is better
```
omnibus = "bundle exec %s" % "omnibus.bat" if sys.platform=win3w else "omnibus"
```

tasks/omnibus.py
```diff
@@ -36,14 +36,7 @@ def omnibus_run_task(ctx, task, target_project, base_dir, env, log_level="info",
...
         cmd = "{omnibus} {task} {project_name} --log-level={log_level} {overrides}"
```
aiuto, line 40: I see the pattern here is to create command lines that get passed to ctx.run().
Is that really what we do?  It's so much safer to pass a vector of args, and let ctx run properly quote per OS when it does the fork/exec.  If we have that option, we could start migrating to it.

tasks/omnibus.py
rdesgroppes: I agree with your parser, and will change accordingly.

tasks/omnibus.py
rdesgroppes: I share your point of view, but I think it would require a major overhaul to do it consistently and not just in the following line... And we need to put this in the context of the transition to Bazel, where most of these imperative lines will be transformed into rules.

alopezz: This may work but the [type annotation of the run function](https://docs.pyinvoke.org/en/stable/api/runners.html#invoke.runners.Runner.run) asks for a string.

paulcacheux:
probably a rebase error here

tasks/omnibus.py
rdesgroppes: Compromise:
```py
omnibus = f"bundle exec {"omnibus.bat" if sys.platform == "win32" else "omnibus"}"
```

rdesgroppes:
> probably a rebase error here

Indeed, thanks @paulcacheux!

tasks/unit_tests/omnibus_tests.py
rdesgroppes: ✅ Done, now running these tests for all platforms irrelevantly from the host OS.

tasks/omnibus.py
rdesgroppes: Resolving because we won't do an extensive refactoring in this round.

tasks/unit_tests/omnibus_tests.py
```diff
@@ -165,17 +187,20 @@ def test_cache_hit_with_corruption(self):
         # We're satisfied if we ran the build despite that failure
         self.assertRunLines([r'bundle exec omnibus build agent'])
 
+    @_for_each_platform()
     def test_cache_is_disabled_by_unsetting_env_var(self):
-        del os.environ['OMNIBUS_GIT_CACHE_DIR']
```
rdesgroppes, line 169: Note to reviewers: this used to mutate the *global* `os.environ` (`os` imported by `omnibus_tests.py`).

tasks/unit_tests/omnibus_tests.py
```diff
@@ -165,17 +187,20 @@ def test_cache_hit_with_corruption(self):
...
         self._set_up_default_command_mocks()
 
-        omnibus.build(self.mock_ctx)
+        with mock.patch.dict("os.environ") as env:
+            del env["OMNIBUS_GIT_CACHE_DIR"]
```
rdesgroppes, line 198: ... whereas this only mutates the test call-specific patched `os.environ` (specific to sub-test execution stack).

tasks/unit_tests/omnibus_tests.py
```diff
@@ -40,27 +41,45 @@ def _run_calls_to_string(mock_calls):
...
+    @staticmethod
+    def _for_each_platform(calls_get_dd_api_key: bool = False):
+        def decorator(test_func):
+            @functools.wraps(test_func)
+            def wrapper(self, *args, **kwargs):
+                for platform, get_dd_api_key_env in {
+                    "darwin": {"AGENT_API_KEY_ORG2": "agent-api-key"},
+                    "linux": {"AGENT_API_KEY_ORG2": "agent-api-key", "POD_NAMESPACE": "pod-ns"},
+                    "win32": {"API_KEY_ORG2": "api-key"},
+                }.items():
```
alopezz: This is a point where I miss pytest 😬; interesting way to solve this regardless.

tasks/unit_tests/omnibus_tests.py
```diff
@@ -40,27 +41,45 @@ def _run_calls_to_string(mock_calls):
...
+                    with (
+                        self.subTest(platform=platform),
+                        mock.patch("sys.platform", platform),
+                        mock.patch.dict(os.environ, get_dd_api_key_env) if calls_get_dd_api_key else nullcontext(),
```
alopezz: Question: would there be any drawbacks to making this patching unconditional?

tasks/unit_tests/omnibus_tests.py
```diff
@@ -76,6 +95,7 @@ def assertRunLines(self, line_patterns):
             f'Failed to match pattern {line_patterns}.',
         )
 
+    @_for_each_platform()
```
alopezz: I can't comment directly on the line, but I'm surprised that this assertion (and similar ones throughout the test suite):

https://github.com/DataDog/datadog-agent/blob/c27725383bd9d6d1e63194d3ddbb2caf868c597b/tasks/unit_tests/omnibus_tests.py#L228

doesn't fail when `sys.platform == "win32"` given

https://github.com/DataDog/datadog-agent/blob/c27725383bd9d6d1e63194d3ddbb2caf868c597b/tasks/omnibus.py#L39.

🤔 Maybe a bug on the test (on the assertion helper perhaps? )?

alopezz (commented):
Looks good to me, my only real concern is the one about the assertion on `bundle exec omnibus` when win32 should have `bundle exec omnibus.py`.

tasks/unit_tests/omnibus_tests.py
rdesgroppes: I also miss `pytest`, but [nose](https://nose.readthedocs.io/en/latest/) even more.

tasks/unit_tests/omnibus_tests.py
rdesgroppes: Quite the other way around. Let me shrink it down.

tasks/unit_tests/omnibus_tests.py
alopezz: Thanks for the catch!

tasks/unit_tests/omnibus_tests.py
rdesgroppes: ✅ Done in 2nd commit.

tasks/unit_tests/omnibus_tests.py
rdesgroppes: Yes, there's was a subtle bug... the following method turns out to be only invoked prior to executing a `test_*` method, but not in-between sub-tests:
```py
    def setUp(self):
        self.mock_ctx = MockContextRaising(run={})
```
✅ Please find a fix in the 3rd commit.

alopezz (approved):
Thanks!

PR #39635, opened 2025-08-07
Migrate processes to WorkloadFilter

vitkyrka:
### What does this PR do?

Add processes support to WorkloadFilter and migrate the one existing
configuration to use it.

### Motivation

https://datadoghq.atlassian.net/browse/DSCVR-177

Tests added.
Existing process config tests minimally modified to use new infra.

### DIFF
--- a/comp/core/workloadfilter/def/types.go
+++ b/comp/core/workloadfilter/def/types.go
@@ -6,7 +6,10 @@
 package workloadfilter
 
 import (
+	"strings"
+
 	typedef "github.com/DataDog/datadog-agent/comp/core/workloadfilter/def/proto"
+	workloadmeta "github.com/DataDog/datadog-agent/comp/core/workloadmeta/def"
 )
 
 // Result is an enumeration that represents the possible results of a filter evaluation.
@@ -41,6 +44,7 @@ const (
 	ServiceType   ResourceType = "service"
 	EndpointType  ResourceType = "endpoint"
 	ImageType     ResourceType = "image"
+	ProcessType   ResourceType = "process"
 )
 
 //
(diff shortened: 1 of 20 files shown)


comp/core/workloadfilter/def/types.go
```diff
@@ -218,3 +222,54 @@ const (
...
+// Process Definition
+//
+
+// Process represents a filterable process object.
+type Process struct {
+	*typedef.FilterProcess
+}
+
+// CreateProcess creates a Filterable Process object from a workloadmeta.Process.
+func CreateProcess(process *workloadmeta.Process) *Process {
```
gabedos, line 236: Nit: In a recent PR, we've moved the Create methods to a separate subpackage to better manage dependencies. You should be able to find it now under `comp/core/workloadfilter/util/workloadmeta`

pkg/process/checks/process.go
```diff
@@ -701,29 +701,13 @@ func initScrubber(config pkgconfigmodel.Reader, scrubber *procutil.DataScrubber)
...
-		}
+func isDisallowListed(cmdline []string, filter workloadfiltercomp.Component) bool {
+	process := &workloadfiltercomp.Process{
+		FilterProcess: &workloadfiltertypedef.FilterProcess{
+			Cmdline: strings.Join(cmdline, " "),
+		},
 	}
-	return false
+
+	return filter.IsProcessExcluded(process, [][]workloadfiltercomp.ProcessFilter{{workloadfiltercomp.LegacyProcessDisallowlist}})
```
gabedos, line 712: Using the workloadFiltering component with `cel-go` will result in the CEL dependencies being pulled. The process-agent binary is now pulling in more dependencies as seen here https://github.com/DataDog/datadog-agent/pull/39635#issuecomment-3165103354. This will result in failing `static_quality_gates`.

We may need to first look into the feasibility of using build tags to fallback on legacy options in the process Agent. Will the new consolidated process check + service discovery and process autodiscovery be only executed on the Core Agent? or will this implementation also be supported on the Process Agent (that is being deprecated)?

vitkyrka: AFAIK it will only be supported on the core agent.  We could just avoid changing the existing process denylist option, since the CEL support is not required for that. I mostly did it just to have a user for the new process CEL support.  The process CEL support will actually be useful for new features like process autodiscovery and filtering for logs discovery, and those things are done in the core agent.

PR #39637, opened 2025-08-07
Add deploy jobs for DDOT deb/rpm

liustanley:
### What does this PR do?
Adds deploy jobs for DDOT deb, rpm, and rpm suse


spencergilbert (approved):
Seems correct, assuming the config on the agent-release-management side is updated and correct

PR #39638, opened 2025-08-07
Fix stabilization window implementation

vboulineau:
### What does this PR do?

Fix the stabilization window implementation. 3 issues were detected in the current implementation:
- The major one, using only actions data does not work as we need actual recommendation history to make the correct decision
- The second one is that we were applying stabilization window after applying scaling speed rules
- The third one is that we were restricting the set of recommendations we were looking at to determine `min` and `max` boundaries, disappears as we're now looking at raw recommendations instead of actions.

### Motivation

Provide accurate results when using stabilization window.

The change can be validated by using workloads that have flapping replica calculations and verifying the expected behaviour (which is greatly explained in this issue: https://github.com/kubernetes/kubernetes/issues/96671).

This PR does not implement the serialization of recommendation history in the CR, as no field is available for it.
As a best effort, we read from the past actions to get _some_ of the past recommendations.

On a leader change or Cluster Agent restart, the stabilization window will reset. Note that old data may be present in the followers when they were once leader, it's not an issue as data are discarded based on timestamp.

### Additional Notes

### DIFF
--- a/pkg/clusteragent/autoscaling/workload/controller_horizontal.go
+++ b/pkg/clusteragent/autoscaling/workload/controller_horizontal.go
@@ -12,6 +12,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"slices"
 	"time"
 
 	autoscalingv1 "k8s.io/api/autoscaling/v1"
@@ -108,12 +109,7 @@ func (hr *horizontalController) performScaling(ctx context.Context, podAutoscale
 		autoscalerInternal.UpdateFromHorizontalAction(nil, err)
 		return autoscaling.NoRequeue, nil
 	}
-	// We are already scaled
-	if horizontalAction == nil {
-		autoscalerInternal.UpdateFromHorizontalAction(nil, nil)
-		return autoscaling.NoRequeue, nil
-	}
-	// Target replicas has not changed due to scaling rules
+	// Target replicas has not changed because we are already scaled or due to scaling rules
 	if horizontalAction.FromReplicas == horizontalAction.ToReplicas {
 		autoscalerInternal.UpdateFromHorizontalAction(horizontalAction, nil)
 		if nextEvalAfter > 0 {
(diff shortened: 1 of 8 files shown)


clamoriniere (approved):
looks fine

pkg/clusteragent/autoscaling/workload/controller_horizontal.go
```diff
@@ -108,12 +109,7 @@ func (hr *horizontalController) performScaling(ctx context.Context, podAutoscale
 		autoscalerInternal.UpdateFromHorizontalAction(nil, err)
 		return autoscaling.NoRequeue, nil
 	}
-	// We are already scaled
-	if horizontalAction == nil {
```
jennchenn, line 112: Can't `horizontalAction` still be `nil`? Why did you remove this check?
https://github.com/DataDog/datadog-agent/blob/98983bff124127c8fffcf5a9f4103ded6164a400/pkg/clusteragent/autoscaling/workload/controller_horizontal.go#L161-L163

pkg/clusteragent/autoscaling/workload/model/pod_autoscaler.go
```diff
@@ -292,6 +308,14 @@ func (p *PodAutoscalerInternal) UpdateFromStatus(status *datadoghqcommon.Datadog
 
 		if len(status.Horizontal.LastActions) > 0 {
 			p.horizontalLastActions = status.Horizontal.LastActions
+
+			// TODO: Store the recommendations history as well, the last actions miss data
+			for _, recommendation := range status.Horizontal.LastActions {
+				p.horizontalLastRecommendations = addRecommendationToHistory(recommendation.Time.Time, p.horizontalRecommendationsRetention, p.horizontalLastRecommendations, HorizontalScalingValues{
+					Timestamp: recommendation.Time.Time,
+					Replicas:  recommendation.ToReplicas,
```
jennchenn, line 316: To confirm `horizontalLastRecommendations` always stores the constrained recommendation and not the actual recommendation?

jennchenn (approved):

pkg/clusteragent/autoscaling/workload/controller_horizontal.go
vboulineau: Previously we could report `nil, nil` indeed but it's not the case anymore, we report only `nil, err`. `nil, nil` was used to convey `no scale`, but the case was already checked after

pkg/clusteragent/autoscaling/workload/model/pod_autoscaler.go
vboulineau: Good catch it should be the unconstrained one, we only store unconstrained recommendations

PR #39642, opened 2025-08-07
[Dynamic Instrumentation] Symbols Upload

GreenMatan:
### What does this PR do?
Completes the work started in #39020 - utilizing symbols extraction functionality and uploads it to debugger's backend for tracked Go processes.

The debugger backend delivers symdb upload request through RC. When a service receives this request, the system-probe reads it's DWARF and start extracting symbols in hierarchy of Package -> Function -> Args/Locals.
The symbols are uploaded with an iterator to avoid excessive memory usage during the upload.  Currently, we cap the upload batches to packages accumulating to >= 10k functions.

### Motivation
Better experience when creating probes through DI UI. With symbols users are able to simply select which function they want to probe without Source code integration and without intimate knowledge of the way Go constructs functions fully qualified names.

### DIFF
--- a/pkg/dyninst/module/state.go
+++ b/pkg/dyninst/module/state.go
@@ -21,11 +21,12 @@ import (
 
 type processState struct {
 	procRuntimeID
-	executable       actuator.Executable
-	symbolicator     symbol.Symbolicator
-	symbolicatorFile io.Closer
-	symbolicatorErr  error
-	gitInfo          procmon.GitInfo
+	executable         actuator.Executable
+	symbolicator       symbol.Symbolicator
+	symbolicatorFile   io.Closer
+	symbolicatorErr    error
+	gitInfo            procmon.GitInfo
+	symdbUploadStarted bool
 }
 
 type processStore struct {
@@ -63,9 +64,11 @@ func (ps *processStore) ensureExists(update *rcscrape.ProcessUpdate) procRuntime
 	if !ok {
 		proc = &processState{
 			procRuntimeID: procRuntimeID{
-				ProcessID: update.ProcessID,
-				runtimeID: update.RuntimeID,
-				service:   update.Service,
+				ProcessID:   update.ProcessID,
+				runtimeID:   update.RuntimeID,
+				service:     update.Service,
+				version:     update.Version,
+				environment: update.Environment,
 			},
 			executable: update.Executable,
 			gitInfo:    update.GitInfo,
@@ -141,4 +144,17 @@ func (ps *processStore) getRuntimeID(procID actuator.ProcessID) (procRuntimeID, 
 		return procRuntimeID{}, false
 	}
 	return p.procRuntimeID, true
+}
+
+func (ps *processStore) markSymdbUploadStarted(pid actuator.ProcessID) bool {
+	ps.mu.Lock()
+	defer ps.mu.Unlock()
+
+	proc, ok := ps.processes[pid]
+	if !ok || proc.symdbUploadStarted {
+		return false // Process gone or already started
+	}
+
+	proc.symdbUploadStarted = true
+	return true
 }
--- a/pkg/dyninst/module/state.go
+++ b/pkg/dyninst/module/state.go
@@ -21,12 +21,11 @@ import (
 
 type processState struct {
 	procRuntimeID
-	executable         actuator.Executable
-	symbolicator       symbol.Symbolicator
-	symbolicatorFile   io.Closer
-	symbolicatorErr    error
-	gitInfo            procmon.GitInfo
-	symdbUploadStarted bool
+	executable       actuator.Executable
+	symbolicator     symbol.Symbolicator
+	symbolicatorFile io.Closer
+	symbolicatorErr  error
+	gitInfo          procmon.GitInfo
 }
 
 type processStore struct {
@@ -144,17 +143,4 @@ func (ps *processStore) getRuntimeID(procID actuator.ProcessID) (procRuntimeID, 
 		return procRuntimeID{}, false
 	}
 	return p.procRuntimeID, true
-}
-
-func (ps *processStore) markSymdbUploadStarted(pid actuator.ProcessID) bool {
-	ps.mu.Lock()
-	defer ps.mu.Unlock()
-
-	proc, ok := ps.processes[pid]
-	if !ok || proc.symdbUploadStarted {
-		return false // Process gone or already started
-	}
-
-	proc.symdbUploadStarted = true
-	return true
 }
--- a/pkg/dyninst/module/symdb.go
+++ b/pkg/dyninst/module/symdb.go
@@ -1,0 +1,63 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-present Datadog, Inc.
+
+//go:build linux_bpf
+
+package module
+
+import (
+	"github.com/DataDog/datadog-agent/pkg/dyninst/actuator"
+	"github.com/DataDog/datadog-agent/pkg/dyninst/symdb"
+	"github.com/DataDog/datadog-agent/pkg/dyninst/uploader"
+	"github.com/DataDog/datadog-agent/pkg/util/log"
+)
+
+// SymDBUploader interface for uploading SymDB data
+type SymDBUploader interface {
+	Enqueue(eventMetadata *uploader.EventMetadata, symdbRoot *uploader.SymDBRoot) error
+}
+
+type symdbManager struct {
+	uploader SymDBUploader
+	store    *processStore
+}
+
+func newSymdbManager(uploader SymDBUploader, store *processStore) *symdbManager {
+	return &symdbManager{
+		uploader: uploader,
+		store:    store,
+	}
+}
+
+func (m *symdbManager) requestUpload(runtimeID procRuntimeID, executable actuator.Executable) {
+	if !m.store.markSymdbUploadStarted(runtimeID.ProcessID) {
+		return
+	}
+
+	go m.performUpload(runtimeID, executable)
+}
+
+func (m *symdbManager) performUpload(runtimeID procRuntimeID, executable actuator.Executable) {
+	symbols, err := symdb.ExtractSymbols(executable.Path, symdb.ExtractOptions{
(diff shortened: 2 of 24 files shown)


pkg/dyninst/uploader/symdb.go
```diff
@@ -81,35 +72,41 @@ type LineRange struct {
...
 }
 
-type EventMetadata struct {
-	DDSource  string `json:"ddsource"`
-	Service   string `json:"service"`
-	RuntimeID string `json:"runtimeId"`
-}
-
 type SymDBUploader struct {
+	service string
```
andreimatei: these fields are left-over from when I was playing around; they're not needed here

pkg/dyninst/actuator/state.go
```diff
@@ -224,7 +224,7 @@ func handleProcessesUpdated(
 	var before, after []probeKey
 	anythingChanged := func(
 		p *process,
-		probesAfterUpdate []ir.ProbeDefinition,
+		pu ProcessUpdate,
```
andreimatei: revert changes to this file? You didn't seem to need the change, unless you have an ulterior motive.

pkg/dyninst/module/symdb.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/DataDog/datadog-agent/pkg/dyninst/uploader"
+	"github.com/DataDog/datadog-agent/pkg/util/log"
+)
+
+// SymDBUploader interface for uploading SymDB data
+type SymDBUploader interface {
+	Enqueue(pkg symdb.Package) error
+}
+
+type symdbManager struct {
```
andreimatei, line 24: please give this guy a comment saying what it's for

pkg/dyninst/rcscrape/scraper.go
```diff
@@ -144,6 +144,7 @@ func (s *scraperSink) HandleEvent(ev output.Event) error {
 		if err != nil {
 			return err
 		}
+		symdbEnabled = true // TODO debugging purposes, won't be merged
```
andreimatei: don't forget about this

pkg/dyninst/module/state.go
```diff
@@ -142,3 +145,16 @@ func (ps *processStore) getRuntimeID(procID actuator.ProcessID) (procRuntimeID,
...
+func (ps *processStore) markSymdbUploadStarted(pid actuator.ProcessID) bool {
+	ps.mu.Lock()
+	defer ps.mu.Unlock()
+
+	proc, ok := ps.processes[pid]
+	if !ok || proc.symdbUploadStarted {
+		return false // Process gone or already started
+	}
+
+	proc.symdbUploadStarted = true
```
andreimatei: I might well be missing something, but I think this is not quite right -- we set `proc.symdbUploadStarted = true` and we never set it to `false`, which means that we'll never upload the symbols a second time for a process. I think we should treat `Process.ShouldUploadSymDB` as "edge-triggered" -- i.e. do the upload every time it changes from `false -> true`. Also, I think we should cancel an ongoing upload when it switches `true -> false` (*) (the way things currently work, once a `ProcessUpdate` has `ShouldUploadSymDB == true`, future ones will also have the flag set as long as the respective RC predicate continues targeting the tracer.

I realize that it's not expected for the backend to select a tracer for doing an upload, then un-select it, then select it again, but I don't think this code should rely on that.

(*) we could associate a `ctx` and `CancelFunc` to an upload

pkg/dyninst/uploader/symdb.go
```diff
@@ -0,0 +1,397 @@
...
+	// Add language specifics for line ranges if available
+	if len(variable.AvailableLineRanges) > 0 {
+		ranges := make([]LineRange, len(variable.AvailableLineRanges))
+		for i, r := range variable.AvailableLineRanges {
+			ranges[i] = LineRange{
+				Start: r[0],
+				End:   r[1],
+			}
+		}
+		/*symbol.LanguageSpecifics = &LanguageSpecifics{
```
andreimatei: get rid of this or write a TODO

pkg/dyninst/module/symdb.go
```diff
@@ -0,0 +1,74 @@
...
+			Scope:                   symdb.ExtractScopeModulesFromSameOrg,
+			IncludeInlinedFunctions: false,
+		})
+	if err != nil {
+		log.Errorf("SymDB: failed to read symbols for process %v (executable: %s): %v",
+			runtimeID.ProcessID, executable.Path, err)
+		return
+	}
+
+	up := uploader.NewSymDBUploader(
```
andreimatei: I think this leaks a goroutine in the `Uploader`; `Stop()` needs to be called.

pkg/dyninst/uploader/symdb.go
```diff
@@ -0,0 +1,397 @@
...
+	SymbolTypeLocal       SymbolType = "local"
+)
+
+// Symbol represents a variable, parameter, or field in the SymDB schema
+type Symbol struct {
+	Name       string     `json:"name"`
+	Type       string     `json:"type"`
+	SymbolType SymbolType `json:"symbol_type"`
+	Line       *int       `json:"line,omitempty"`
+}x
```
andreimatei: xx

pkg/dyninst/module/symdb.go
andreimatei: nevermind, I rewrote this uploader

andreimatei (commented):
I pushed a commit removing the dependency on the `Batcher` since I had issues with it; hope you don't mind. Please take a look.

pkg/dyninst/procmon/analyze.go
```diff
@@ -131,7 +131,7 @@ func analyzeProcess(
 			"failed to analyze environ", err,
 		)
 	}
-	if ddEnv.serviceName == "" || !ddEnv.diEnabled {
+	if ddEnv.serviceName == "" || ddEnv.serviceVersion == "" || ddEnv.environmentName == "" || !ddEnv.diEnabled {
 		log.Tracef(
 			"process %d is not interesting: service name is %q, %s=%t",
```
andreimatei, line 136: let's include all the relevant env vars in here

pkg/dyninst/procmon/analyze.go
```diff
@@ -131,7 +131,7 @@ func analyzeProcess(
 			"failed to analyze environ", err,
 		)
 	}
-	if ddEnv.serviceName == "" || !ddEnv.diEnabled {
+	if ddEnv.serviceName == "" || ddEnv.serviceVersion == "" || ddEnv.environmentName == "" || !ddEnv.diEnabled {
```
andreimatei: Let's make sure we're all on the same page here -- that we need all three of service name, env, and version in order to monitor a process. All of these are indeed needed for that process to be targeted for probing through RC?
From the UI, I seem to be able to create probes without filtering on either env or version, AFAICS.

cc @ajwerner @piob-io

GreenMatan: yes, all three are required in all other languages, as [described in our docs](https://docs.datadoghq.com/tracing/dynamic_instrumentation/#:~:text=Unified%20Service%20Tagging%20tags%20service%2C%20env%2C%20and%20version%20are%20applied%20to%20your%20deployment.). I think it's safe to require them. Also, in the upcoming 'on by default' efforts, we'd like to target preprod envs which will require anyhow to set the environment.
WDYT?

pkg/dyninst/procmon/analyze.go
GreenMatan: 🤔 , which ones am I missing?

pkg/dyninst/module/state.go
GreenMatan: the problem is, SymbolUpload request through RC is never changing back to `false` and we receive `symdbUpload=true` every few seconds through our config probes. What it means is that there is no notion of 'reuploading' symbols for a given service, as far as I can tell. Even if we had a way to know if we were requested to reupload symbols, does it make sense for Go? We have the same binary in hand and we don't care about dynamically loaded modules.

I'll do the work to cancel in flight symdb upload requests if the the flag turns to `false` (it's currently not supported but might be in the future).

pkg/dyninst/actuator/state.go
GreenMatan: yes indeed, in the first iteration of implementing symdb upload I was thinking to include it in different code paths that required to change `handleProcessesUpdated` in a way that detects not only probe changes but also symdb upload flag changes.
I'll revert this change.

pkg/dyninst/procmon/analyze.go
andreimatei: I've undone this, going back to requiring only DD_SERVICE, as per the slack conversation -- it seems that we want to continue supporting processes without env and version. I'll test manually that the symdb backend likes what we upload when those aren't set.

pkg/dyninst/testprogs/progs/go.mod
```diff
@@ -9,7 +9,8 @@ require (
 	github.com/DataDog/dd-trace-go/v2 v2.3.0-dev
```
andreimatei: you had upgraded it to 2.3.0-dev.1; I've undone that assuming that there was no particular reason for it.

Separately, I'll update to 2.2.3 and delete the `replace` below once that's tagged.

pkg/dyninst/actuator/state.go
andreimatei: I've reverted this file.

pkg/dyninst/module/symdb.go
andreimatei: added a comment in a later commit

pkg/dyninst/procmon/analyze.go
andreimatei: the log message only includes the service name. But now this is good again, since we no longer check for env and version, so this comment is obsolete.

andreimatei (approved):
I massaged all the commits so that they all compile and pass tests.
I'm approving this now; I think it's good to go.
Do a pass and give your stamp when you can.

pkg/dyninst/procmon/analyze.go
GreenMatan: Keeping this conv open until it's fully tested

pkg/dyninst/module/state.go
GreenMatan: This conv is no longer relevant, you already implemented aborting in flight symdb uploads
Resolving

dustmop (approved):

PR #39645, opened 2025-08-07
dyninst: implement hmap support

ajwerner:
This adds support for maps in go 1.23 and prior.

Fixes [DEBUG-4185](https://datadoghq.atlassian.net/browse/DEBUG-4185).

Testing will come in #39378

[DEBUG-4185]: https://datadoghq.atlassian.net/browse/DEBUG-4185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/dyninst/decode/types.go
+++ b/pkg/dyninst/decode/types.go
@@ -19,7 +19,7 @@ import (
 	"github.com/go-json-experiment/json/jsontext"
 
 	"github.com/DataDog/datadog-agent/pkg/dyninst/ir"
-	"github.com/DataDog/datadog-agent/pkg/util/log"
+	"github.com/DataDog/datadog-agent/pkg/dyninst/output"
 )
 
 // decoderType is a decoder-specific representation of an ir.Type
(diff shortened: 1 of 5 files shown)


pkg/dyninst/decode/types.go
```diff
@@ -267,7 +292,54 @@ func newDecoderType(
...
+		if err != nil {
+			return nil, fmt.Errorf("malformed hmap header type: %w", err)
+		}
+		bucketsField, err := getFieldByName(s.RawFields, "buckets")
+		if err != nil {
+			return nil, fmt.Errorf("malformed hmap header type: %w", err)
+		}
+		oldBucketsField, err := getFieldByName(s.RawFields, "oldbuckets")
+		if err != nil {
+			return nil, fmt.Errorf("malformed hmap header type: %w", err)
```
grantseltzer, line 305: Can you make these errors messages more specific? When errors are the same strings in different places it's harder to track down where they originated from.

grantseltzer (approved):

pkg/dyninst/decode/types.go
ajwerner: but each of the errors coming from `getFieldByName` has the name...

ajwerner: It's also parallel to what we have in swiss maps. Feel free to push on me for a follow-up if my answers do not please you

PR #39657, opened 2025-08-08
[ksm] Migrate from Kubelet pod watcher to workloadmeta

davidor:
### What does this PR do?

This PR updates the KSM check to use workloadmeta instead of the Kubelet pod watcher to get pod info.

Now, when the KSM check is configured to emit pod metrics from the node Agent it gets pod information from the kubelet using the Pod Watcher defined in the package that contains the Kubelet client. This PR changes the KSM check so that it uses workloadmeta instead. This is similar to the change that I did for the prometheus autodiscovery provider in https://github.com/DataDog/datadog-agent/pull/39238

This change is part of the effort to make workloadmeta the single source of truth for pod data. Benefits include:
- Consistent data across Agent components.
- Code easier to maintain and reason about.
- Improved performance by avoiding redundant data and extra calls to the Kubelet.

Summary of the changes:
- Added new fields to the `KubernetesPod` struct in workloadmeta that KSM needs. These fields are only used by KSM, so I didn’t add protobuf definitions for them.
- Updated the kubelet workloadmeta collector to populate the new fields.
- Replaced the kubelet-based reflector in KSM with one that uses workloadmeta.
- Updated the pod conversion logic in KSM to use workloadmeta pods instead of pods from the kubelet package.

- Added unit tests for the new code.
- Ran manual tests to compare `kubernetes_state.pod.*`, `kubernetes_state.container.*`, and `kubernetes_state.init_container.*` metrics with the current version.


misteriaud (approved):

comp/core/workloadmeta/def/types.go
```diff
@@ -801,9 +824,67 @@ func (p KubernetesPod) String(verbose bool) string {
 		_, _ = fmt.Fprintln(&sb, "Kube Services:", sliceToString(p.KubeServices))
 		_, _ = fmt.Fprintln(&sb, "Namespace Labels:", mapToString(p.NamespaceLabels))
 		_, _ = fmt.Fprintln(&sb, "Namespace Annotations:", mapToString(p.NamespaceAnnotations))
+
 		if !p.FinishedAt.IsZero() {
 			_, _ = fmt.Fprintln(&sb, "Finished At:", p.FinishedAt)
 		}
+
+		if !p.CreationTimestamp.IsZero() {
+			_, _ = fmt.Fprintln(&sb, "Creation Timestamp:", p.CreationTimestamp)
```
L3n41c, line 833: I’m not sure the `_, _ = …` pattern is strictly needed.
At least, there are some other places in `datadog-agent` codebase where the return values of `fmt.Fprintln` are ignored without raising a linter violation like: https://github.com/DataDog/datadog-agent/blob/37d1b8d448852791921dcd3a34f0182c500b4a26/cmd/agent/subcommands/flare/command.go#L223-L233

```suggestion
			fmt.Fprintln(&sb, "Creation Timestamp:", p.CreationTimestamp)
```

comp/core/workloadmeta/def/types.go
```diff
@@ -801,9 +824,67 @@ func (p KubernetesPod) String(verbose bool) string {
...
+		if p.HostIP != "" {
+			_, _ = fmt.Fprintln(&sb, "Host IP:", p.HostIP)
+		}
+
+		_, _ = fmt.Fprintln(&sb, "Host Network:", p.HostNetwork)
+
+		if len(p.InitContainerStatuses) > 0 {
+			_, _ = fmt.Fprintln(&sb, "----------- Init Container Statuses -----------")
+			for _, cs := range p.InitContainerStatuses {
+				_, _ = fmt.Fprint(&sb, cs.String(verbose))
```
L3n41c, line 853: Definitively not a big deal for such a piece of code invoked only by debug CLI commands, but it might have been more efficient to pass the `strings.Builder` directly to the function in charge of printing the `KubernetesContainerStatus` structure instead of having to go through a `String`.

```suggestion
				cs.WriteTo(&sb, verbose)
```

comp/core/workloadmeta/def/types.go
```diff
@@ -842,6 +924,154 @@ func (o KubernetesPodOwner) String(verbose bool) string {
...
+	_, _ = fmt.Fprintln(&sb, "ContainerID:", cs.ContainerID)
+	_, _ = fmt.Fprintln(&sb, "Name:", cs.Name)
+	_, _ = fmt.Fprintln(&sb, "Image:", cs.Image)
+	_, _ = fmt.Fprintln(&sb, "ImageID:", cs.ImageID)
+	_, _ = fmt.Fprintln(&sb, "Ready:", cs.Ready)
+	_, _ = fmt.Fprintln(&sb, "RestartCount:", cs.RestartCount)
+	_, _ = fmt.Fprint(&sb, cs.State.String(false))
+	_, _ = fmt.Fprint(&sb, cs.LastTerminationState.String(false))
+	return sb.String()
+}
```
L3n41c, line 1027: Followup of previous comment.
If we pass the `strings.Builder` directly, we can be slightly more efficient.

```suggestion
func (cs KubernetesContainerStatus) WriteTo(sb *strings.Builder, _verbose bool) {
	fmt.Fprintln(sb, "ContainerID:", cs.ContainerID)
	fmt.Fprintln(sb, "Name:", cs.Name)
	fmt.Fprintln(sb, "Image:", cs.Image)
	fmt.Fprintln(sb, "ImageID:", cs.ImageID)
	fmt.Fprintln(sb, "Ready:", cs.Ready)
	fmt.Fprintln(sb, "RestartCount:", cs.RestartCount)
	fmt.Fprint(sb, cs.State.String(false))
	fmt.Fprint(sb, cs.LastTerminationState.String(false))
}
```

comp/core/workloadmeta/collectors/internal/kubelet/kubelet.go
```diff
@@ -172,15 +172,21 @@ func parsePods(pods []*kubelet.Pod, collectEphemeralContainers bool) []workloadm
...
+				Name:       o.Name,
+				ID:         o.ID,
+				Controller: o.Controller,
 			})
 		}
 
 		PodSecurityContext := extractPodSecurityContext(&pod.Spec)
 		RuntimeClassName := extractPodRuntimeClassName(&pod.Spec)
 
+		var startTime *time.Time
```
gabedos, line 185: Is there a particular reason why we are storing the start time as nil when it's zero? Why can't we just directly pass the value through directly?
`StartTime: &pod.Status.StartTime`

davidor: This might be a detail without practical implications, but the Pod in the kubernetes library (`corev1.Pod`) defines this as a `*time.Time`, but when we parse the pod from the Kubelet we store it as a `time.Time`: https://github.com/DataDog/datadog-agent/blob/64e4aef3d9a2f32f4124366fc734c593bddcc248/pkg/util/kubernetes/kubelet/types_kubelet.go#L209

That’s why I needed this here. However, I think the real fix would be to change the `startTime` referenced above to `*time.Time`. That way, our kubelet type would match what corev1.Pod defines.

pkg/kubestatemetrics/builder/workloadmeta_conversion.go
```diff
@@ -0,0 +1,409 @@
...
+
+	if resources.CPULimit != nil || resources.MemoryLimit != nil {
+		result.Limits = make(corev1.ResourceList)
+	}
+
+	// CPU resources are stored as percentages (0-100*numCPU) in workloadmeta
+	// but Kubernetes ResourceList expects them as core fractions (e.g., "0.1" for 10% of one core)
+	if resources.CPURequest != nil {
+		cpuCores := *resources.CPURequest / 100.0
+		result.Requests[corev1.ResourceCPU] = resource.MustParse(fmt.Sprintf("%.3f", cpuCores))
```
L3n41c: Having to go through a `string` to convert a `float` to a `resource.Quantity` is suboptimal.
`resource.NewMilliQuantity(…)` should be more efficient.

```suggestion
		result.Requests[corev1.ResourceCPU] = resource.NewMilliQuantity(int64(cpuCores*1000), resource.DecimalSI)
```

pkg/kubestatemetrics/builder/workloadmeta_conversion.go
```diff
@@ -0,0 +1,409 @@
...
+	}
+	if resources.MemoryRequest != nil {
+		result.Requests[corev1.ResourceMemory] = resource.MustParse(fmt.Sprintf("%d", *resources.MemoryRequest))
```
L3n41c: Having to go through a `string` to convert an `int` to a `resource.Quantity` is suboptimal.
`resource.NewQuantity(…)` should be more efficient.

```suggestion
		result.Requests[corev1.ResourceMemory] = resource.NewQuantity(*resources.MemoryRequest, resource.BinarySI)
```

pkg/kubestatemetrics/builder/workloadmeta_conversion.go
```diff
@@ -0,0 +1,409 @@
...
+	}
+	if resources.CPULimit != nil {
+		cpuCores := *resources.CPULimit / 100.0
+		result.Limits[corev1.ResourceCPU] = resource.MustParse(fmt.Sprintf("%.3f", cpuCores))
+	}
+	if resources.MemoryLimit != nil {
+		result.Limits[corev1.ResourceMemory] = resource.MustParse(fmt.Sprintf("%d", *resources.MemoryLimit))
```
L3n41c: Having to go through a `string` to convert an `int` to a `resource.Quantity` is suboptimal.
`resource.NewQuantity(…)` should be more efficient.

```suggestion
		result.Limits[corev1.ResourceMemory] = resource.NewQuantity(*resources.MemoryLimit, resource.BinarySI)
```

pkg/kubestatemetrics/builder/workloadmeta_conversion.go
```diff
@@ -0,0 +1,409 @@
...
+	if resources.CPURequest != nil {
+		cpuCores := *resources.CPURequest / 100.0
+		result.Requests[corev1.ResourceCPU] = resource.MustParse(fmt.Sprintf("%.3f", cpuCores))
+	}
+	if resources.MemoryRequest != nil {
+		result.Requests[corev1.ResourceMemory] = resource.MustParse(fmt.Sprintf("%d", *resources.MemoryRequest))
+	}
+	if resources.CPULimit != nil {
+		cpuCores := *resources.CPULimit / 100.0
+		result.Limits[corev1.ResourceCPU] = resource.MustParse(fmt.Sprintf("%.3f", cpuCores))
```
L3n41c: Having to go through a `string` to convert a `float` to a `resource.Quantity` is suboptimal.
`resource.NewMilliQuantity(…)` should be more efficient.

```suggestion
		result.Limits[corev1.ResourceCPU] = resource.NewMilliQuantity(int64(cpuCores*1000), resource.DecimalSI)
```

pkg/kubestatemetrics/builder/workloadmeta_pods.go
```diff
@@ -0,0 +1,150 @@
...
+	// the collectors configured (by default it's the pods one plus
+	// "pods_extended")
+	stores []cache.Store
+
+	started bool
+}
+
+func newWorkloadmetaReflector(wmeta workloadmeta.Component, namespaces []string) (workloadmetaReflector, error) {
+	if wmeta == nil {
+		return workloadmetaReflector{}, fmt.Errorf("workloadmeta cannot be nil")
```
L3n41c: As there is no string formatting, we can use:

```suggestion
		return workloadmetaReflector{}, errors.New("workloadmeta cannot be nil")
```

pkg/kubestatemetrics/builder/workloadmeta_pods.go
```diff
@@ -0,0 +1,150 @@
...
+	return workloadmetaReflector{
+		namespaces:         namespaces,
+		watchAllNamespaces: watchAllNamespaces,
+		wmeta:              wmeta,
+	}, nil
+}
+
+func (wr *workloadmetaReflector) addStore(store cache.Store) error {
+	if wr.started {
+		return fmt.Errorf("cannot add store after reflector has started")
```
L3n41c: In the absence of formatter, `errors.New` is more efficient than `fmt.Errorf`.

```suggestion
		return errors.New("cannot add store after reflector has started")
```

pkg/kubestatemetrics/builder/workloadmeta_pods.go
```diff
@@ -0,0 +1,150 @@
...
+
+	wr.stores = append(wr.stores, store)
+	return nil
+}
+
+// start starts the workloadmeta reflector. It should be called only once after all the
+// stores have been added.
+func (wr *workloadmetaReflector) start(ctx context.Context) error {
+	if wr.started {
+		return fmt.Errorf("reflector already started")
```
L3n41c:
```suggestion
		return errors.New("reflector already started")
```

L3n41c (approved):

pkg/kubestatemetrics/builder/workloadmeta_conversion.go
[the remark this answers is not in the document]
L3n41c: Given the string format passed to `fmt.Sprintf("%.3f", …)`, I would expect the resulting string to fail to be parsed by `resource.MustParse(…)` only in very pathological situations like with `NaN` or `Inf` !

Anyway, the best would be to use `resource.NewQuantity` or `resource.NewMilliQuantity` instead.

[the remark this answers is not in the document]
L3n41c: Given the string format passed to `fmt.Sprintf("%.3f", …)`, I would expect the resulting string to fail to be parsed by `resource.MustParse(…)` only in very pathological situations like with `NaN` or `Inf` !

Anyway, the best would be to use `resource.NewQuantity` or `resource.NewMilliQuantity` instead.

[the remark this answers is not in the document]
L3n41c: Given the string format passed to `fmt.Sprintf("%.3f", …)`, I would expect the resulting string to fail to be parsed by `resource.MustParse(…)` only in very pathological situations like with `NaN` or `Inf` !

Anyway, the best would be to use `resource.NewQuantity` or `resource.NewMilliQuantity` instead.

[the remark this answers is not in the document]
L3n41c: Given the string format passed to `fmt.Sprintf("%.3f", …)`, I would expect the resulting string to fail to be parsed by `resource.MustParse(…)` only in very pathological situations like with `NaN` or `Inf` !

Anyway, the best would be to use `resource.NewQuantity` or `resource.NewMilliQuantity` instead.

comp/core/workloadmeta/def/types.go
davidor: Yes, I noticed that. However, all the other usages of `Fprintln` in the file that I updated use the` _, _ = ...` pattern so I think it's better to leave it like that for consistency.

comp/core/workloadmeta/def/types.go
davidor: I agree, but the existing code in this file uses the `String()` pattern, so I prefer to leave it like that for consistency.
We could change all usages in a different PR, but as you said, I don't think it's worth it given that this is just a CLI command for debugging.

pkg/kubestatemetrics/builder/workloadmeta_conversion.go
davidor: Fixed :+1:

pkg/kubestatemetrics/builder/workloadmeta_conversion.go
davidor: Fixed :+1:

pkg/kubestatemetrics/builder/workloadmeta_conversion.go
davidor: Fixed :+1:

pkg/kubestatemetrics/builder/workloadmeta_pods.go
davidor: Fixed :+1:

pkg/kubestatemetrics/builder/workloadmeta_conversion.go
davidor: Fixed :+1:

pkg/kubestatemetrics/builder/workloadmeta_pods.go
davidor: Fixed :+1:

pkg/kubestatemetrics/builder/workloadmeta_pods.go
davidor: Fixed :+1:

pkg/kubestatemetrics/builder/workloadmeta_conversion.go
```diff
@@ -0,0 +1,409 @@
...
+		result[i] = corev1.Container{
+			Name:      container.Name,
+			Image:     container.Image.Name,
+			Resources: convertWorkloadmetaContainerResources(container.Resources),
+		}
+
+		// Lookup full container information from workloadmeta
+		containerEntity, err := wmeta.GetContainer(container.ID)
+		if err != nil {
+			log.Trace("Failed to get full container entity for ID %s: %v", container.ID, err)
```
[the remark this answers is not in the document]
davidor: Fixed :+1:

davidor:
@L3n41c Thanks for the review. I addressed your comments.

L3n41c (approved):

davidor:
Rebased on top of main to make CI pass.

gabedos (approved):

PR #39664, opened 2025-08-08
[SQG] Rework static quality gates.

JSGette:
### What does this PR do?
Previous implementation suffered from various issues:
- Code duplication
- Suppressed errors and error traces
- Poor test coverage
- Unclear invocation parameters due to
usage of reflection and kwargs
    
This tries to address those issues by redesigning the entire approach on how we treat static quality
gates and trigger them. We no longer have 3 different libraries to work with packages, docker images and MSI.
They are all merged into the same gates.py. We also introduce ArtifactMeasurer protocol that describes
measure function that later differs for packages and docker images. We also introduce several immutable
dataclasses and value checks to prevent initialization errors.

Another important part is visualization in the pipeline and Datadog. For one part we now print quick
results per each quality gate with color coding allowing to quickly detect whether the execution was
successful or not. At the end of the execution we print a summary that shows current sizes against max
allowed values and the remaining budget left.

### Motivation
This is an attempt to make SQG more maintainable.

New unit tests are added, old tests adjusted to work with new code.

Debug features are not yet available


tasks/static_quality_gates/gates.py
```diff
@@ -0,0 +1,866 @@
...
+
+
+class StaticQualityGateFailed(Exception):
+    """
+    Exception raised when a static quality gate fails
+    """
+
+    def __init__(self, message: str):
+        self.message = color_message(message, "red")
+        super().__init__(self.message)
```
rdesgroppes: The `message` attribute was intentionally removed from Python 3. To retrieve the message from an exception `e`, the need is now covered by `str(e)`.
If one still wants to achieve `e.message` despite the above mentioned design choice, one may add a property instead, for instance:
```suggestion
        super().__init__(color_message(message, "red"))

    @property
    def message(self):
        return self.args[0]
```

tasks/static_quality_gates/gates.py
```diff
@@ -0,0 +1,866 @@
...
+    os: str
+
+    def __post_init__(self):
+        """Validate configuration data"""
+        if not self.gate_name:
+            raise ValueError("gate_name cannot be empty")
+        if self.max_on_wire_size <= 0:
+            raise ValueError("max_on_wire_size must be positive")
+        if self.max_on_disk_size <= 0:
+            raise ValueError("max_on_disk_size must be positive")
```
rdesgroppes:
```suggestion
            raise ValueError("max_on_wire_size must be strictly positive")
        if self.max_on_disk_size <= 0:
            raise ValueError("max_on_disk_size must be strictly positive")
```

tasks/static_quality_gates/gates.py
```diff
@@ -0,0 +1,866 @@
...
+            wire_size, disk_size = self._calculate_package_sizes(ctx, config, artifact_paths)
+
+            # For packages, the primary artifact path is the main package file
+            primary_path = artifact_paths.get('primary', artifact_paths.get('msi', ''))
+
+            return ArtifactMeasurement(artifact_path=primary_path, on_wire_size=wire_size, on_disk_size=disk_size)
+        except Exception as e:
+            if isinstance(e, StaticQualityGateFailed | InfraError):
+                raise
+            raise StaticQualityGateFailed(f"Failed to measure package {config.gate_name}: {e}") from e
```
rdesgroppes: The following adjustment would leverage Python’s exception hierarchy and control flow:
```suggestion
        except (StaticQualityGateFailed, InfraError):
            raise
        except Exception as e:
            raise StaticQualityGateFailed(f"Failed to measure package {config.gate_name}") from e
```
Also: no `: {e}` in the message: avoids redundancy, because `from e` already includes the underlying exception in the output.

rdesgroppes (approved):
Fine by me, feel free to ignore my minor comments.
Thanks for your commendable patience and the great outcome!

tasks/static_quality_gates/gates.py
JSGette: Removed message and passing a string directly

tasks/quality_gates.py
```diff
@@ -152,61 +159,93 @@ def _print_quality_gates_report(gate_states: list[dict[str, typing.Any]]):
...
-    ctx.run(f"datadog-ci tag --level job --tags static_quality_gates:\"{final_state}\"")
+        finally:
+            metric_handler.register_gate_tags(
+                gate.config.gate_name,
+                gate_name=gate.config.gate_name,
+                arch=gate.config.arch,
+                os=gate.config.os,
+                pipeline_id=os.environ["CI_PIPELINE_ID"],
+                ci_commit_ref_slug=os.environ["CI_COMMIT_REF_SLUG"],
+                ci_commit_sha=os.environ["CI_COMMIT_SHA"],
```
KevinFairise2, line 204: Any reason to not take these attributes when the metric handler is instantiated, I think these one should not change depending on the gate

ofek (approved):
Looks like a reasonable change, thanks a lot!

KevinFairise2 (approved):
That's way clearer than it used to be! Good job.
Just one minor comment that you can feel free to ignore

tasks/quality_gates.py
JSGette: I think, I used this approach because of its flexibility. I didn't want to touch GateMetricHandler just yet as it's already pretty big change whereas tags function adds arbitrary fields to metadata which is exactly what I wanted to achieve

PR #39665, opened 2025-08-08
[CHAOSPLT-1027] Add fuzz test for dyninst

edznux-dd:
### What does this PR do?

This PR adds a few initial fuzz harnesses for the dynisnt "domains". These are quite early and fairly naive, but it'll help get started and allow for iterative improvements.

### Motivation

The goal is to find bugs and improve the reliability of the agent by catch panics, DoS and other issues, sometimes security sensitive automatically, with fuzzing.
With the automatic on-boarding in the fuzzing infrastructure, we are trying to reach increase the code coverage of fuzz tests.

```bash
cd ~/dd/datadog-agent/pkg/dyninst/object
go test -fuzz=FuzzParseModuleData -tags=linux_bpf -run=FuzzParseModuleData

cd ~/dd/datadog-agent/pkg/dyninst/dwarf/loclist
go test -fuzz=FuzzParseInstructions -tags=linux_bpf -run=FuzzParseInstructions

cd ~/dd/datadog-agent/pkg/dyninst/procmon
go test -fuzz=FuzzAnalyzeProcess -tags=linux_bpf -run=FuzzAnalyzeProcess
```

### Additional Notes

These test won't run in your CI pipeline directly but on a scheduled run on main, every day :)
This avoid flaky failures + very long fuzz test duration.

We have enrichment, triaging and notification capabilities to keep you alerted.


KevinFairise2 (approved):

pkg/dyninst/decode/decoder_test.go
```diff
@@ -0,0 +1,312 @@
...
+	symbolicator := &mockSymbolicator{}
+
+	f.Fuzz(func(_ *testing.T, eventData []byte, serviceName string) {
+		event := Event{
+			Event:       output.Event(eventData),
+			ServiceName: serviceName,
+		}
+
+		var buf bytes.Buffer
+		probe, err := decoder.Decode(event, symbolicator, &buf)
```
ajwerner: The thing about the probe that's returned here is that it's just one of the values inside that test program -- it's always going to be that one probe you created in createTestProgram. Verifying anything about it down below seems uninteresting. 

I think this is a good area to fuzz, but I don't think the fuzzer is going to end up hitting very many cases here. I think we'd be better off creating more complex programs and a more complex seed corpus and then fuzzing from there. What do you think about pulling this decoding fuzzing out of this PR and putting it on another PR stacked against this one. Then I can help collaborate on making it better.

pkg/dyninst/dwarf/loclist/fuzz_test.go
```diff
@@ -0,0 +1,19 @@
...
+package loclist
+
+import "testing"
+
+func FuzzParseInstructions(f *testing.F) {
+	f.Fuzz(func(t *testing.T, data []byte, ptrSize uint8, totalByteSize uint32) {
+		_, err := ParseInstructions(data, ptrSize, totalByteSize)
+		if err != nil {
+			t.Skip("Failed to parse instructions")
+		}
```
ajwerner: Does this t.Skip add any value?

pkg/dyninst/object/fuzz_test.go
```diff
@@ -0,0 +1,170 @@
...
+		f.Add(validELF[:len(validELF)/2])
+	}
+
+	// Add empty and minimal data
+	f.Add([]byte{})
+	f.Add([]byte{0x7f, 0x45, 0x4c, 0x46}) // ELF magic only
+
+	f.Fuzz(func(t *testing.T, data []byte) {
+		// TODO: use of an in memory file instead of a temporary file to avoid massive IO
+		tmpFile, err := os.CreateTemp("", "elf-file")
```
ajwerner, line 44: This is motivating for me. The `object` package APIs have gotten sprawling. I have a PR inbound to make this trivial. Might come in after lunch.

pkg/dyninst/object/fuzz_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+	// Add various malformed ELF structures
+	corruptedELF := make([]byte, len(validELF))
+	copy(corruptedELF, validELF)
+	// Corrupt the section headers
+	if len(corruptedELF) > 100 {
+		corruptedELF[50] = 0xFF
+		corruptedELF[51] = 0xFF
+	}
+	f.Add(corruptedELF)
```
ajwerner, line 31: thoughts on using some real files?

ajwerner (commented):
Neat! Thanks for doing it! I'll follow up with a PR that'll make the object fuzzing better in a bit.

pkg/dyninst/object/fuzz_test.go
edznux-dd: that would be great!

I don't have an easy way that is accesible in the datadog-agent to push for arbitrary binary, but, if we can either `embed` them with Go, or convert them to fuzz input format and upload them manually (once) that would be extremely useful to get decent coverage!

pkg/dyninst/object/fuzz_test.go
edznux-dd: Nice, I'll try to merge this PR and feel free to update in yours afterward?

pkg/dyninst/dwarf/loclist/fuzz_test.go
edznux-dd: Changing to return, I agree, probably unecessary.

I don't think I can do anything with the ir.Piece returned from it, to exercise more methods / functions on that package.

pkg/dyninst/decode/decoder_test.go
edznux-dd: Sure, let me remove that :)

Happy to see that it's an interesting and valuable part, but I can't claim expertise on that codebase, so i'm fine with that :) 
thanks for the feedback!

pkg/dyninst/object/fuzz_test.go
ajwerner: We have some tooling for this. I'll hook it up in a follow up.

pkg/dyninst/object/fuzz_test.go
ajwerner: Sounds good.

ajwerner (approved):
Merging now works. I'll file some JIRAs for myself to follow up on the things we discussed.

PR #39672, opened 2025-08-08
feat: create new checksum fingerprinting feature to track truncation and rotation for files

lucastemb:
### What does this PR do?

This PR implements improvements to log rotation detection by introducing a fingerprinting mechanism that enhances Datadog Agent’s ability to accurately identify log file rotations. The change replaces or supplements the current inode-based detection logic with a content-based fingerprint, improving resilience to edge cases such as inode reuse or file truncation.

### Motivation
The existing log rotation detection mechanism can misidentify log file continuations or rotations under certain conditions, particularly in environments where inode reuse is common or files are renamed without changes to metadata. These failures lead to incorrect log collection behavior such as duplicated logs, data loss, or broken pipelines.

This PR addresses those issues by implementing a fingerprinting strategy based on content hashes of the beginning of the file, which remains stable across common log rotation patterns. This change aims to:

- Improve rotation detection accuracy
- Reduce false positives in file change detection
- Enable safe and performant tracking of log files over time

The design and rationale are described in detail in [RFC: Log Rotation Detection Improvements via Fingerprinting](https://datadoghq.atlassian.net/wiki/spaces/AL/pages/5095229352).

Checksum fingerprinting may run marginally slower due to the Agent now having to read the content of each file to be tailed.


cswatt (approved):

pgimalac (approved):
LGTM for Agent Runtimes

comp/logs/agent/config/config.go
```diff
@@ -160,6 +164,64 @@ func IsExpectedTagsSet(coreConfig pkgconfigmodel.Reader) bool {
 	return ExpectedTagsDuration(coreConfig) > 0
 }
 
+// GlobalFingerprintConfig returns the global fingerprint configuration to apply to all logs.
+func GlobalFingerprintConfig(coreConfig pkgconfigmodel.Reader) (*FingerprintConfig, error) {
+	var err error
+	// Get the complete fingerprint config map without applying defaults immediately
+	raw := coreConfig.Get("logs_config.fingerprint_config")
+	// Let the config system handle defaults
+	config := &FingerprintConfig{}
```
dustmop: Here you're declaring `config` as a pointer to `FingerprintConfig`, but then creating a pointer to that when passing it to `json.Unmarshal` and `UnmarshalKey` a few lines later, meaning you're passing in a double-pointer (`**FingerprintConfig`). Could you change this line to instead be:

```
config := FingerprintConfig{}
```

and then you'll need to fix up the call to `ValidateFingerprintConfig` and the return value.

comp/logs/agent/config/config.go
```diff
@@ -160,6 +164,64 @@ func IsExpectedTagsSet(coreConfig pkgconfigmodel.Reader) bool {
...
+
+	if s, ok := raw.(string); ok && s != "" {
+		log.Debugf("GlobalFingerprintConfig: unmarshaling from string: %s", s)
+		err = json.Unmarshal([]byte(s), &config)
+	} else {
+		log.Debugf("GlobalFingerprintConfig: using structure.UnmarshalKey")
+		// Fallback to structure.UnmarshalKey
+		err = structure.UnmarshalKey(coreConfig, "logs_config.fingerprint_config", &config, structure.ConvertEmptyStringToNil)
```
dustmop: Do we need to use `structure.ConvertEmptyStringToNil` here or could we get away with leaving it out? I think it only exists to support some legacy tests.

comp/logs/agent/config/config.go
```diff
@@ -160,6 +164,64 @@ func IsExpectedTagsSet(coreConfig pkgconfigmodel.Reader) bool {
...
 
+// GlobalFingerprintConfig returns the global fingerprint configuration to apply to all logs.
+func GlobalFingerprintConfig(coreConfig pkgconfigmodel.Reader) (*FingerprintConfig, error) {
+	var err error
+	// Get the complete fingerprint config map without applying defaults immediately
+	raw := coreConfig.Get("logs_config.fingerprint_config")
+	// Let the config system handle defaults
+	config := &FingerprintConfig{}
+
+	if s, ok := raw.(string); ok && s != "" {
```
dustmop: I don't love that this setting has a variadic type (either a string or a map[string]) but I see that it is similar to how [GlobalProcessingRules](https://github.com/DataDog/datadog-agent/blob/b19befd8e0bf9ce4b3cc64edcf7f06f561fbb220/comp/logs/agent/config/config.go#L74C6-L74C27) works, and we don't currently have an alternative that provides this functionality. I think this is fine for now, but as we work on making the config has more structured types, we'll probably have to revisit this and simplify it.

pkg/config/setup/config.go
```diff
@@ -1575,6 +1595,15 @@ func logsagent(config pkgconfigmodel.Setup) {
 	config.BindEnvAndSetDefault("logs_config.socks5_proxy_address", "")
 	// disable distributed senders
 	config.BindEnvAndSetDefault("logs_config.disable_distributed_senders", false)
+	// determines fingerprinting strategy to detect rotation and truncation
+	config.BindEnvAndSetDefault("logs_config.fingerprint_enabled_experimental", DefaultExperimentalFingerprintingEnabled)
+	// consolidated fingerprint configuration with defaults
+	config.BindEnvAndSetDefault("logs_config.fingerprint_config", map[string]interface{}{
```
dustmop: Normally, we would instead use `BindEnvAndSetDefault` for each of these settings, such as:

```
config.BindEnvAndSetDefault("logs_config.fingerprint_config.count", DefaultFingerprintingMaxLines)
```
etc, but I think that's incompatible with allowing "logs_config.fingerprint_config" to be a `map[string]` OR json `string`.

Do we know if it's possible in prod for the setting "logs_config.fingerprint_config" to end up with a json string assigned as its value, or is that capability only used by tests?

lucastemb: @dustmop At the moment, I think it's a capability used by tests, but I do think that it can be assigned as a json string if it's passed in as an env var. 

This does echo a similar sentiment that @hush-hush shared in a previous comment on this [PR](https://github.com/DataDog/datadog-agent/pull/38010#discussion_r2253822339), so I will be removing the env vars related stuff.

comp/logs/agent/config/config.go
lucastemb: 🤔  Running `dda inv test` seems to indicate that it's not needed.

pkg/logs/launchers/file/launcher.go
```diff
@@ -55,6 +73,14 @@ type Launcher struct {
 	scanPeriod             time.Duration
 	flarecontroller        *flareController.FlareController
 	tagger                 tagger.Component
+	//Stores pertinent information about old tailer when rotation occurs and fingerprinting isn't possible
```
dustmop: Please add a space after the slashes as it is [considered idiomatic](https://www.digitalocean.com/community/tutorials/how-to-write-comments-in-go#ordinary-comments), except for directives like `//go:*`.

pkg/config/setup/config.go
```diff
@@ -40,6 +40,26 @@ import (
...
+	DefaultFingerprintingMaxBytes = 100000
+
+	// DefaultLinesOrBytesToSkip is the default number of lines (or bytes) to skip when reading a file.
+	// Whether we skip lines or bytes is dependent on whether we choose to compute the fingerprint by lines or by bytes.
+	DefaultLinesOrBytesToSkip = 0
+
+	// DefaultFingerprintingMaxLines is the default maximum number of lines to read before computing the fingerprint.
+	DefaultFingerprintingMaxLines = 1
+
+	//DefaultFingerprintStrategy is the default strategy for computing the checksum fingerprint.
```
dustmop: same here please

comp/logs/agent/config/config.go
```diff
@@ -160,6 +164,53 @@ func IsExpectedTagsSet(coreConfig pkgconfigmodel.Reader) bool {
...
+	var err error
+	config := FingerprintConfig{}
+	err = structure.UnmarshalKey(coreConfig, "logs_config.fingerprint_config", &config)
+	if err != nil {
+		return nil, err
+	}
+	log.Debugf("GlobalFingerprintConfig: after unmarshaling - FingerprintStrategy: %s, Count: %d, CountToSkip: %d, MaxBytes: %d",
+		config.FingerprintStrategy, config.Count, config.CountToSkip, config.MaxBytes)
+
+	//Return the config and validate the fingerprintConfig as well
```
dustmop: same here

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,286 @@
...
+			}
+		}
+	}
+
+	// Read lines for hashing
+	var buffer []byte
+	linesRead := 0
+	bytesRead := 0
+	for linesRead < maxLines && scanner.Scan() {
+		//TODO: Check for error here
```
dustmop: same here

pkg/logs/types/fingerprint.go
```diff
@@ -0,0 +1,73 @@
...
+	InvalidFingerprintValue = 0
+)
+
+// Fingerprint struct that stores both the value and config used to derive that value
+type Fingerprint struct {
+	Value  uint64
+	Config *FingerprintConfig
+}
+
+func (f *Fingerprint) String() string {
```
dustmop, line 25: Please add a comment for this method also

dustmop (approved):
Thanks for the fixes! A have a few small nits but nothing blocking, approving now.

comp/logs/agent/config/integration_config.go
```diff
@@ -35,6 +35,28 @@ const (
 	SHIFTJIS string = "shift-jis"
 )
 
+// FingerprintStrategy defines the strategy used for fingerprinting (line_checksum or byte_checksum)
+type FingerprintStrategy string
```
ddrthall: This looks like a duplicate type, can shift all instances over to the types/fingerprint.go version

comp/logs/auditor/impl/auditor.go
```diff
@@ -38,6 +39,8 @@ type RegistryEntry struct {
 	Offset             string
 	TailingMode        string
 	IngestionTimestamp int64
+	Fingerprint        uint64
```
ddrthall: There isn't a significant advantage to storing discrete fingerprint values and fingerprint configs now that the fingerprint type has been safely removed from the circular dependency loop -  Fingerprint and FingerprintConfig can be reduced to a single *types.Fingerprint field

comp/logs/auditor/impl/auditor.go
```diff
@@ -234,7 +267,27 @@ func (a *registryAuditor) run() {
 			}
 			// update the registry with the new entry
 			for _, msg := range payload.MessageMetas {
-				a.updateRegistry(msg.Origin.Identifier, msg.Origin.Offset, msg.Origin.LogSource.Config.TailingMode, msg.IngestionTimestamp)
+				var fingerprintConfig *types.FingerprintConfig
+				if msg.Origin.LogSource.Config.FingerprintConfig != nil {
```
ddrthall: Do we want to be dynamically identifying a fingerprint config in the auditor? Under what circumstances would we want to be pulling either the log source config or the global config compared to the one stored within the origin's Fingerprint?

pkg/logs/launchers/file/launcher.go
```diff
@@ -170,26 +201,62 @@ func (s *Launcher) scan() {
...
+				if didRotate {
+					// restart tailer because of file-rotation on file
+					succeeded := s.restartTailerAfterFileRotation(tailered, file)
+					if !succeeded {
+						// the setup failed, let's try to tail this file in the next scan
+						continue
+					}
+				}
 			}
 		} else {
```
ddrthall: This logic can be safely compressed in a couple of ways:
1. We don't need to both check IsFingerprintingEnabled and ShouldFileFingerprint - the latter includes the former, we should only ever need to call ShouldFileFingerprint.

2. This will be mentioned elsewhere, but if you ever find yourself casting an interface to a discrete struct you've likely done something wrong. In this case we don't actually need the interface and can safely delete it in favor of the struct. If we do end up wanting the interface for any reason, it should always include the functions referenced here, such as ShouldFingerprintFile.

3. We're duplicating a lot of logic for rotation typing and detection. We would ideally want something akin to
```
if ShouldFileFingerprint {
    if DidRotateViaFingerprint {
         rotateTailerWithoutRestart
    }
} else {
    if DidRotate {
        restartTailerAfterFileRotation
    }
}
```
There will be a lot more error handling, of course, but I don't think there's a reason to ever call any of these functions in more than one flow

pkg/logs/launchers/file/launcher.go
```diff
@@ -212,20 +279,51 @@ func (s *Launcher) scan() {
...
 			filesTailed[scanKey] = true
 			continue
 		}
+
+		// Check if we have stored info for this file from a previous rotation
+		oldInfo, hasOldInfo := s.oldInfoMap[scanKey]
+		if hasOldInfo {
+			// Use stored info for rotation detection
+			var fingerprint *types.Fingerprint
+			if s.fingerprinter.IsFingerprintingEnabled() {
```
ddrthall: This is also very compression friendly - some similar notes to above regarding the removal of IsFingerprintingEnabled and the removal of the interface casting. 
We can also move the fingerprint check completely outside of the hasOldInfo if-else block, so we don't need to duplicate the logic across both flows.

pkg/logs/launchers/file/launcher.go
```diff
@@ -212,20 +279,51 @@ func (s *Launcher) scan() {
...
-			}
-			tailersLen++
+		_, isTailed := s.tailers.Get(scanKey)
+		if isTailed {
 			filesTailed[scanKey] = true
 			continue
 		}
+
+		// Check if we have stored info for this file from a previous rotation
+		oldInfo, hasOldInfo := s.oldInfoMap[scanKey]
```
ddrthall: We do need some way to clean out oldInfo that was never inserted in to a tailer and is no longer associated with a file. The easiest way I can think of doing this is to to copy and replace oldInfo right before we begin pass 2, e.g. 
```
lastIterationOldInfo := s.oldInfoMap
s.oldInfoMap = make(map[string]*oldTailerInfo)
```

We then repopulate oldInfo around here - inserting key-value pairs back in to s.oldInfoMap from lastIterationOldInfo only if there is still an untailed file associated with the scan key in lastIteration... and that file remains un-fingerprintable

pkg/logs/launchers/file/launcher.go
```diff
@@ -293,31 +391,43 @@ func (s *Launcher) launchTailers(source *sources.LogSource) {
 			continue
 		}
 
+		var fingerprint *types.Fingerprint
+		if s.fingerprinter.IsFingerprintingEnabled() {
```
ddrthall: Another prime compression spot - removal of IsFingerprintingEnabled and the removal of the interface casting.

pkg/logs/launchers/file/launcher.go
```diff
@@ -379,11 +572,15 @@ func (s *Launcher) restartTailerAfterFileRotation(oldTailer *tailer.Tailer, file
 	// We will keep track of the rotated tailer until it is finished.
 	s.rotatedTailers = append(s.rotatedTailers, oldTailer)
 	s.tailers.Add(newTailer)
+
+	// Clean up any stale old info for this file path
+	delete(s.oldInfoMap, file.Path)
```
ddrthall: This is likely extraneous - we don't expect to have anything in oldInfoMap for files that will ever call this function.

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,286 @@
...
+			f.defaultFingerprintConfig.FingerprintStrategy,
+			f.defaultFingerprintConfig.Count,
+			f.defaultFingerprintConfig.CountToSkip,
+			f.defaultFingerprintConfig.MaxBytes)
+		return computeFingerprint(file.Path, f.defaultFingerprintConfig)
+	}
+
+	// Convert from config.FingerprintConfig to types.FingerprintConfig
+	fingerprintConfig := &types.FingerprintConfig{
+		FingerprintStrategy: types.FingerprintStrategy(configFingerprintConfig.FingerprintStrategy),
```
ddrthall, line 115: This cast shouldn't be necessary once the typing is consolidated

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+
+	log.Debugf("Computing fingerprint for %s with config: strategy=%s, count=%d, countToSkip=%d, maxBytes=%d",
+		filePath,
+		fingerprintConfig.FingerprintStrategy,
+		fingerprintConfig.Count,
+		fingerprintConfig.CountToSkip,
+		fingerprintConfig.MaxBytes)
+
+	fpFile, err := os.Open(filePath)
```
ddrthall: os.Open isn't necessarily friendly across operating systems, look into filesystem.OpenShared

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,286 @@
...
+		log.Debugf("Falling back to default line-based strategy for %s", filePath)
+		return computeFingerPrintByLines(fpFile, filePath, defaultLinesConfig)
+	}
+}
+
+// computeFingerPrintByBytes computes fingerprint using byte-based approach for a given file path
+func computeFingerPrintByBytes(fpFile *os.File, filePath string, fingerprintConfig *types.FingerprintConfig) *types.Fingerprint {
+	bytesToSkip := fingerprintConfig.CountToSkip
+	maxBytes := fingerprintConfig.Count
+	if fingerprintConfig.FingerprintStrategy == "line_checksum" {
```
ddrthall: We'll never want computeFingerPrintByBytes to convert an invalid config within itself. If we want extra surety that we've rigged the code up correctly we could make this an error case

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,286 @@
...
+	// Read up to maxBytes for hashing
+	buffer := make([]byte, maxBytes)
+	bytesRead, err := io.ReadFull(fpFile, buffer)
+	if err != nil && err != io.EOF && err != io.ErrUnexpectedEOF {
+		log.Warnf("Failed to read bytes for fingerprint %q: %v", filePath, err)
+		return newInvalidFingerprint(fingerprintConfig)
+	}
+
+	// Trim buffer to actual bytes read
+	buffer = buffer[:bytesRead]
```
ddrthall: Under what circumstances will bytesRead != maxBytes and is this how we want to handle that?

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,286 @@
...
+	maxBytes := fingerprintConfig.MaxBytes
+
+	// Create a LimitedReader to respect maxBytes constraint
+	limitedReader := io.LimitReader(fpFile, int64(maxBytes))
+
+	// Create scanner for line-by-line reading
+	scanner := bufio.NewScanner(limitedReader)
+
+	// Skip the configured number of lines
+	for i := 0; i < linesToSkip; i++ {
```
ddrthall: There's a lot of shared logic between the linesToSkip and the linesRead loops. It's quite likely they can be safely conjoined to something akin to
```
for i :=; i < linesToSkip + linesToRead; i++ {
     if Scan() {
          if enoughLinesHaveBeenSkipped {
                append buffer + newLine
          } 
     } else {
          handle error states
     }
}
```

pkg/logs/types/fingerprint.go
```diff
@@ -0,0 +1,73 @@
...
+
+const (
+	// FingerprintStrategyLineChecksum computes a checksum based on line content
+	FingerprintStrategyLineChecksum FingerprintStrategy = "line_checksum"
+	// FingerprintStrategyByteChecksum computes a checksum based on byte content
+	FingerprintStrategyByteChecksum FingerprintStrategy = "byte_checksum"
+)
+
+// Fingerprinter interface defines the methods for fingerprinting files
+type Fingerprinter interface {
```
ddrthall: As mentioned above, right now we don't need this interface. If we do, it should be fully featured such that we don't need to cast it to a real struct for full functionality.

pkg/logs/tailers/file/fingerprint.go
lucastemb: Yeah, looking at it this might be a relic of earlier iterations of the byte_checksum approach and may no longer apply. Going to remove this line.

pkg/logs/tailers/file/fingerprint.go
lucastemb: Going to remove that check for `line_checksum` as well since it no longer applies to how we are currently doing things. The new approach is that if we have to fall back to `byte_based` fingerprinting from `line_based` we will just be using the `defaultBytesConfig` anyways.

pkg/logs/launchers/file/launcher.go
```diff
@@ -212,20 +258,40 @@ func (s *Launcher) scan() {
...
 		}
+
+		// Check if we have stored info for this file from a previous rotation
+		oldInfo, hasOldInfo := lastIterationOldInfo[scanKey]
+		var fingerprint *types.Fingerprint
+		if s.fingerprinter.ShouldFileFingerprint(file) {
+			// Check if this specific file should be fingerprinted
+			fingerprint = s.fingerprinter.ComputeFingerprint(file)
+			// Skip files with invalid fingerprints (Value == 0)
+			//TODO: Put the old stuff back in
```
ddrthall: TODO flag

pkg/logs/launchers/file/launcher.go
ddrthall: Heads up: only one of the two items was compressed, can safely jettison IsFingerprintingEnabled

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,268 @@
...
+	// Read up to maxBytes for hashing
+	buffer := make([]byte, maxBytes)
+	bytesRead, err := io.ReadFull(fpFile, buffer)
+	if err != nil && err != io.EOF && err != io.ErrUnexpectedEOF {
+		log.Warnf("Failed to read bytes for fingerprint %q: %v", filePath, err)
+		return newInvalidFingerprint(fingerprintConfig)
+	}
+
+	// Check if we have enough bytes to create a meaningful fingerprint
+	if bytesRead == 0 || bytesRead < maxBytes {
```
ddrthall, line 191: Nit: I'm assuming that this check is in place for a clearer error message, if so the bytesRead == 0 is superfluous

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	// Handle scanner errors
+	if err := scanner.Err(); err != nil {
+		log.Warnf("Error while reading file for fingerprint %q: %v", filePath, err)
+		return newInvalidFingerprint(fingerprintConfig)
+	}
+
+	// Check if we need to fall back due to byte limits
+	if linesRead == 0 && limitedReader.(*io.LimitedReader).N == 0 {
```
ddrthall: If statement is slightly off, if we ever reach N == 0 without a full compliment of lines we should fall back

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,268 @@
...
+		pos, err := fpFile.Seek(0, io.SeekStart)
+		if pos != 0 || err != nil {
+			log.Warnf("Error %s occurred while trying to reset file offset", err)
+		}
+		return computeFingerPrintByBytes(fpFile, filePath, defaultBytesConfig)
+	}
+
+	// Check if we have enough data for fingerprinting
+	// We need either enough lines OR enough bytes to create a meaningful fingerprint
+	bytesRead := maxBytes - int(limitedReader.(*io.LimitedReader).N)
```
ddrthall: The bytes checks here are likely safe to remove

pkg/logs/launchers/file/launcher.go
```diff
@@ -212,20 +258,39 @@ func (s *Launcher) scan() {
...
+
+		// Check if we have stored info for this file from a previous rotation
+		oldInfo, hasOldInfo := lastIterationOldInfo[scanKey]
+		var fingerprint *types.Fingerprint
+		if s.fingerprinter.ShouldFileFingerprint(file) {
+			// Check if this specific file should be fingerprinted
+			fingerprint = s.fingerprinter.ComputeFingerprint(file)
+			// Skip files with invalid fingerprints (Value == 0)
+			if fingerprint != nil && !fingerprint.ValidFingerprint() {
+				continue
```
ddrthall, line 269: TODO flag needed to be implemented: we should persist oldInfo back into s.oldInfoMap if we think we still need it

pkg/logs/launchers/file/launcher.go
```diff
@@ -33,6 +35,22 @@ import (
...
+// resolveFingerprintConfig resolves the global fingerprint configuration and converts it to the appropriate type
+func resolveFingerprintConfig() *types.FingerprintConfig {
+	fingerprintConfig, err := config.GlobalFingerprintConfig(pkgconfigsetup.Datadog())
+	if err != nil {
+		log.Warnf("Failed to get global fingerprint config: %v, using default", err)
+		return nil
+	}
+	// Convert from config.FingerprintConfig to types.FingerprintConfig
+	return &types.FingerprintConfig{
+		FingerprintStrategy: types.FingerprintStrategy(fingerprintConfig.FingerprintStrategy),
```
ddrthall: Nit: Shouldn't need this cast either

pkg/logs/launchers/file/launcher.go
lucastemb: Whoops, I thought I implemented it and forgot to remove the comment. 🤦 😅

pkg/logs/launchers/file/launcher.go
```diff
@@ -33,6 +35,22 @@ import (
 // DefaultSleepDuration represents the amount of time the tailer waits before reading new data when no data is received
 const DefaultSleepDuration = 1 * time.Second
 
+// resolveFingerprintConfig resolves the global fingerprint configuration and converts it to the appropriate type
+func resolveFingerprintConfig() *types.FingerprintConfig {
+	fingerprintConfig, err := config.GlobalFingerprintConfig(pkgconfigsetup.Datadog())
+	if err != nil {
+		log.Warnf("Failed to get global fingerprint config: %v, using default", err)
+		return nil
```
ddrthall: The callers of this function will segfault on a nil return, and the error is misleading - it implies that we will fall back to a "default", but my understanding is that the global config is itself the default. 

On a quick scan of the codebase there didn't seem to be a clean way to exit with an error once we've committed to forming the launcher. Unless you see something better, I would recommend pulling the global config at the same level as the GlobalProcessingRules in agent.go and passing it down to the launcher

pkg/logs/tailers/file/fingerprint.go
```diff
@@ -0,0 +1,263 @@
...
+				buffer = append(buffer, line...)
+				linesRead++
+			}
+		} else {
+			/// Check if we need to fall back due to byte limits
+			if limitedReader.(*io.LimitedReader).N == 0 {
+				log.Warnf("Scanner stopped with no bytes remaining, falling back to byte-based fingerprint for %q", filePath)
+				pos, err := fpFile.Seek(0, io.SeekStart)
+				if pos != 0 || err != nil {
+					log.Warnf("Error %s occurred while trying to reset file offset", err)
```
ddrthall, line 232: We'll want to handle this the same as an arbitrary error reading the file - once we encounter an io error that's not eof we should immediately back off until we can read the filesystem cleanly again.

pkg/logs/tailers/file/rotate_nix.go
```diff
@@ -52,3 +52,22 @@ func (t *Tailer) DidRotate() (bool, error) {
...
+func (t *Tailer) DidRotateViaFingerprint(fingerprinter *Fingerprinter) (bool, error) {
+	newFingerprint := fingerprinter.ComputeFingerprint(t.file)
+
+	// If the original fingerprint is nil, we can't detect rotation
+	if t.fingerprint == nil {
+		return false, nil
+	}
+
+	// If fingerprints are different, it means the file was rotated.
+	// This is also true if the new fingerprint is invalid (Value=0), which means the file was truncated.
```
ddrthall, line 75: We likely need to extend compute fingerprint to uplift IO errors such that they can be distinguished from "too small to fingerprint" invalid fingerprints - the former is not a rotation, and the latter is

pkg/logs/launchers/file/position.go
```diff
@@ -11,23 +11,46 @@ import (
...
+	fingerprintsAlign := true
+
+	if fingerprinter.IsFingerprintingEnabled() && filePath != "" {
+		prevFingerprint := registry.GetFingerprint(identifier)
+		if prevFingerprint != nil {
+			newFingerprint, err := fingerprinter.ComputeFingerprintFromConfig(filePath, prevFingerprint.Config)
+			if err != nil {
+				log.Warnf("Failed to compute fingerprint for file %s: %v", filePath, err)
+				// If fingerprint computation fails, assume fingerprints don't align to be safe
+				fingerprintsAlign = false
```
ddrthall: I think fingerprintsAlign = true is actually the safer option, or at least it's the option that guarantees the user will see behavior at least as good as the one they have available right now. We're generally more likely to have the agent come back up pointed to the same file compared to a rotated file

pkg/logs/tailers/file/rotate_windows.go
```diff
@@ -53,10 +53,14 @@ func (t *Tailer) DidRotate() (bool, error) {
...
-	newFingerprint := fingerprinter.ComputeFingerprint(t.file)
+	newFingerprint, err := fingerprinter.ComputeFingerprint(t.file)
 
+	// If computing the fingerprint led to an error there was likely an IO issue, handle this appropriately below.
+	if err != nil {
+		return false, err
+	}
 	// If the original fingerprint is nil, we can't detect rotation
-	if t.fingerprint == nil {
+	if t.fingerprint == nil || err != nil {
```
ddrthall: nit: duplicate err check

ddrthall (approved):

L3n41c (approved):

PR #39676, opened 2025-08-08
[AGENTRUN-866] Add ability to run Rust-based checks through shared libraries

Enzu83:
### What does this PR do?

⚠️ This new feature is experimental ⚠️

This PR introduces a new way of running checks, through shared libraries. They are Rust-based and loaded at runtime by a new checks loader. You can read the documentation [here](https://datadoghq.atlassian.net/wiki/spaces/ARUN/pages/5479301643/Running+shared+library+checks+in+the+Agent).

The Rust-based checks API is minimal for now, only the submit functions are available. The API can be expanded quite easily by adding new fields to the C structure `aggregator_t`.

Also, this PR moves the Go submit functions (like `SubmitMetric`) from the collector Python package to a new package (named `aggregator`). That way both Python and shared library checks can use callbacks from this package. These Go submit functions were in the Python package just because Python checks were the only ones using them, but their scope are larger than just Python checks

### Motivation

Provide a new way of writing checks to improve Agent performances and to rely a bit less on the Python runtime.

Few unit tests to test the new checks loader and the shared library checks implementation (for the Go part).
An e2e test for Linux and Windows to load and run a simple shared library check that submits one metric.

### Additional Notes

The Rust part of this feature (where shared libraries are compiled from) is on this PR:
- https://github.com/DataDog/datadog-agent/pull/42351


pkg/collector/rustchecks/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[workspace]
+# checks list
+members = ["core", "checks/*"]
+resolver = "2"
+
+# optimizations to reduce lib sizes by ~45% (from 3.8MB to 2.1MB for `httpcheck`)
+# more here: https://github.com/DoumanAsh/xxhash-rust/issues/10#issuecomment-1121860260
+# build time takes few seconds though
+[profile.release]
+opt-level = "z"     # Optimize for size.
```
fabbing: Did `s` perform less good than `z`?
The documentation says:
```z: optimize for binary size, but more aggressively. Often results in larger binaries than s```

File: pkg/collector/rustchecks/Cargo.toml [hunk 2]
```diff
@@ -0,0 +1,14 @@
...
+lto = true          # Enable Link Time Optimization
+codegen-units = 1   # Reduce number of codegen units to increase optimizations.
+panic = "abort"     # Abort on panic
```
fabbing: This won't be viable in production!

pkg/collector/rustchecks/Cargo.toml
Enzu83: I didn't try many optimization methods to be fair, I left it in this state long ago and I didn't pay much attention if there was better methods.
I'm opened to any recommendations about optimization :)

pkg/collector/rustchecks/Cargo.toml
Enzu83: Yeah you're right, some of them need to be changed/removed

pkg/collector/rustchecks/Cargo.toml
Enzu83: Building with `s` resulted in slightly larger sizes (+0.1MB) for `http_check`

hush-hush (approved):
Looks good for the file own by agent-configuration.

NouemanKHAL (approved):
LGTM for `pkg/cli/subcommands/check/command.go`

pkg/collector/sharedlibrary/aggregator.go
```diff
@@ -0,0 +1,182 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+// this file is identical to pkg/collector/python/aggregator.go besides the C imports and the function names
```
pgimalac: I think you could just use `SubmitMetric` defined for Python, possibly by going through some `unsafe.Pointer` / `void*` ? As long as the code and arguments are exactly the same
That would avoid having to duplicate it
Also maybe using some `extern` directives for the C part

pkg/collector/sharedlibrary/ffi/library_loader.go
```diff
@@ -0,0 +1,164 @@
...
+func getLibExtension() string {
+	switch runtime.GOOS {
+	case "linux", "freebsd":
+		return ".so"
+	case "darwin":
+		return ".dylib"
+	case "windows":
+		return ".dll"
+	default:
+		return ".so"
```
pgimalac: I'm not sure if it's a case we want to handle (it's a code path we'll never hit), but I do think it's correct (everything else is most likely Unix-based)

pkg/collector/sharedlibrary/library_loader.go
```diff
@@ -0,0 +1,164 @@
...
+/*
+#include <stdlib.h>
+
+#include "ffi.h"
+
+void SubmitMetricSo(char *, metric_type_t, char *, double, char **, char *, bool);
+void SubmitServiceCheckSo(char *, char *, int, char **, char *, char *);
+void SubmitEventSo(char *, event_t *);
+void SubmitHistogramBucketSo(char *, char *, long long, float, float, int, char *, char **, bool);
+void SubmitEventPlatformEventSo(char *, char *, int, char *);
```
pgimalac: As mentioned in a previous comment I think you could use `extern void SubmitMetric(char *, metric_type_t, char *, double, char **, char *, bool);` and reuse the function defined in the Python package

pkg/collector/sharedlibrary/library_loader.go
```diff
@@ -0,0 +1,164 @@
...
+	version *C.version_function_t
+}
+
+// libraryLoader is an interface to load/close checks' shared libraries and call their symbols
+type libraryLoader interface {
+	Load(name string) (libraryHandles, error)
+	Close(libHandle unsafe.Pointer) error
+	Run(runPtr *C.run_function_t, checkID string, initConfig string, instanceConfig string) error
+	Version(versionPtr *C.version_function_t) (string, error)
+}
```
pgimalac: I guess the interface is defined in order to allow mocking ?

pkg/collector/sharedlibrary/library_loader.go
```diff
@@ -0,0 +1,164 @@
...
+func (l *sharedLibraryLoader) Load(name string) (libraryHandles, error) {
+	var cErr *C.char
+
+	// the prefix "libdatadog-agent-" is required to avoid possible name conflicts with other shared libraries in the include path
+	libPath := path.Join(l.folderPath, "libdatadog-agent-"+name+getLibExtension())
+
+	cLibPath := C.CString(libPath)
+	defer C.free(unsafe.Pointer(cLibPath))
+
+	cLibHandles := C.load_shared_library(cLibPath, &cErr)
```
pgimalac: Nit: I think defining the variable (`cErr`) where it's used make the code simpler to read and understand

pkg/collector/sharedlibrary/ffi/library_loader.go
```diff
@@ -0,0 +1,164 @@
...
+	aggregator *C.aggregator_t
+}
+
+// Load looks for a shared library with the corresponding name and check if it has a `Run` symbol.
+// If that's the case, then the method will return handles for both.
+func (l *sharedLibraryLoader) Load(name string) (libraryHandles, error) {
+	var cErr *C.char
+
+	// the prefix "libdatadog-agent-" is required to avoid possible name conflicts with other shared libraries in the include path
+	libPath := path.Join(l.folderPath, "libdatadog-agent-"+name+getLibExtension())
```
pgimalac: Libraries often have a version number in their file name, is that something we should handle ?
Eg. `libdatadog-agent-rtloader.0.1.0.dylib`

pgimalac: I guess we could have a naming conflict with rtloader itself, named `libdatadog-agent-rtloader.dylib` 😄

pgimalac: Although to be fair most often there is a symlink from the name without the version, eg. we have `libdatadog-agent-rtloader.dylib` -> `libdatadog-agent-rtloader.2.dylib` -> `libdatadog-agent-rtloader.0.1.0.dylib`

pkg/collector/sharedlibrary/library_loader.go
```diff
@@ -0,0 +1,164 @@
...
+type libraryHandles struct {
+	lib     unsafe.Pointer
+	run     *C.run_function_t
+	version *C.version_function_t
+}
+
+// libraryLoader is an interface to load/close checks' shared libraries and call their symbols
+type libraryLoader interface {
+	Load(name string) (libraryHandles, error)
+	Close(libHandle unsafe.Pointer) error
```
pgimalac: I'm wondering if `Close` should take the whole `libraryHandles` (to be symmetrical to `Load` in some way)

pgimalac: Also possibly `Close` shouldn't be part of the loader, but just be a method of `libraryHandles` ?

pkg/collector/sharedlibrary/ffi/ffi.c
```diff
@@ -0,0 +1,137 @@
...
+    if (dlsym_error) {
+        *error = strdup(dlsym_error);
+    }
+}
+#endif
+
+void run_shared_library(run_function_t *run_ptr, char *check_id, char *init_config, char *instance_config, aggregator_t *aggregator, const char **error) {
+    // verify `Run` pointer
+    if (!run_ptr) {
+        *error = strdup("pointer to 'Run' symbol of the shared library is NULL");
```
pgimalac, line 149: It's a shame to `strdup` constant strings just so that we can free them later, if we really wanted to avoid allocations we could do something `errno` -like with a static list of known errors
I don't think it's necessary for now, the code is simpler as is and it's not the happy path

pkg/collector/sharedlibrary/ffi/ffi.c
```diff
@@ -0,0 +1,137 @@
...
+#include <stdio.h>
+#include <string.h>
+
+#ifdef _WIN32
+#    include <Windows.h>
+#else
+#    include <dlfcn.h>
+#endif
+
+#ifdef _WIN32
```
pgimalac, line 12: Would be best if we can refactor to avoid having everything defined twice
Maybe add some small helpers, eg. "loading a lib", "finding a symbol", "freeing the lib", and just duplicate the implementation of those
The bigger functions would be simpler to read

pkg/collector/sharedlibrary/ffi/ffi.h
pgimalac, line 1: I think this is at least partly duplicated from `rtloader/include/rtloader_types.h`, could we just reuse that ?

pkg/collector/sharedlibrary/check.go
```diff
@@ -0,0 +1,185 @@
...
+	"github.com/DataDog/datadog-agent/pkg/collector/check/defaults"
+	checkid "github.com/DataDog/datadog-agent/pkg/collector/check/id"
+	"github.com/DataDog/datadog-agent/pkg/collector/check/stats"
+	"github.com/DataDog/datadog-agent/pkg/util/log"
+)
+
+// SharedLibraryCheck is the definition of shared library checks
+//
+//nolint:revive
+type SharedLibraryCheck struct {
```
pgimalac: I think the lint is that the typename stutters, which is true, could we just name it `Check` (and use it as `sharedlibrary.Check` ?

pgimalac: In general if you're in package X you don't need every exported function/type to mention X, because the package name will make that clear

pkg/collector/sharedlibrary/check.go
```diff
@@ -0,0 +1,185 @@
...
+
+// Cancel closes the associated shared library and prevents the check from running
+func (c *SharedLibraryCheck) Cancel() {
+	err := c.libraryLoader.Close(c.libHandles.lib)
+	if err != nil {
+		log.Errorf("Cancel failed: %s", err)
+	}
+
+	c.cancelled = true
+}
```
pgimalac: Do we need any synchronization here or are we guaranteed that it won't be called twice, or be called while it's running, etc ?

pkg/collector/sharedlibrary/check.go
```diff
@@ -0,0 +1,185 @@
...
+
+// String representation (for debug and logging)
+func (c *SharedLibraryCheck) String() string {
+	return c.name
+}
+
+// Version is always an empty string
+func (c *SharedLibraryCheck) Version() string {
+	return c.version
+}
```
pgimalac: I think version is actually not always an empty string, since you search for a `Version` symbol in the lib

pkg/collector/sharedlibrary/ffi/library_loader.go
pgimalac: Do we need to do some validation on the name to avoid reading outside the expected directory ? Eg. that it doesn't contains path separators or `..` ?
Potentially be careful with symlinks too, as that could point outside of the directory

pkg/collector/sharedlibrary/ffi.c
```diff
@@ -0,0 +1,137 @@
...
+}
+
+#else
+
+handles_t load_shared_library(const char *lib_path, const char **error) {
+    handles_t lib_handles;
+    char *dlsym_error = NULL;
+
+    // load the library
+    lib_handles.lib = dlopen(lib_path, RTLD_LAZY | RTLD_GLOBAL);
```
pgimalac: Can you explain why you use `RTLD_LAZY` and `RTLD_GLOBAL` ?
I think if there is a linking error we want to see it at load time, so not sure about `RTLD_LAZY`, and I don't believe we want `RTLD_GLOBAL` at all, it would mean loading multiple libraries could conflict ?

pgimalac: I think we should specify somewhere the behavior of dlopen and dlclose when loading multiple times the same object:
```
If the same shared object is opened again with dlopen(), the same
object handle is returned.  The dynamic linker maintains reference
counts for object handles, so a dynamically loaded shared object
is not deallocated until dlclose() has been called on it as many
times as dlopen() has succeeded on it.
```

In particular it means that any global state of the shared library will be shared among instances, which is a big footgun.

pkg/collector/sharedlibrary/library_loader.go
```diff
@@ -0,0 +1,164 @@
...
+	return nil
+}
+
+func (ml *mockSharedLibraryLoader) Run(_ *C.run_function_t, _ string, _ string, _ string) error {
+	return nil
+}
+
+func (ml *mockSharedLibraryLoader) Version(_ *C.version_function_t) (string, error) {
+	return "mock_version", nil
+}
```
pgimalac: The mock should be defined a separate file with `test` build tag

Depending on how it's used, you can use `mockery` to autogenerate an implementation for a given interface, which allows asserting expectation on calls
Current implementation is more a no-op loader than a mock really

pkg/config/setup/config.go
```diff
@@ -1210,6 +1210,10 @@ func InitConfig(config pkgconfigmodel.Setup) {
 
 	// Agent Workload Filtering config
 	config.BindEnvAndSetDefault("cel_workload_exclude", []interface{}{})
+
+	// Shared library check
+	config.BindEnvAndSetDefault("shared_library_check.enabled", true)
```
pgimalac: Not sure if we want to enable it by default ?

pkg/collector/sharedlibrary/init.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/DataDog/datadog-agent/pkg/collector/check"
+	"github.com/DataDog/datadog-agent/pkg/collector/loaders"
+	pkgconfigsetup "github.com/DataDog/datadog-agent/pkg/config/setup"
+	"github.com/DataDog/datadog-agent/pkg/util/log"
+	"github.com/DataDog/datadog-agent/pkg/util/option"
+)
+
+// init adds the shared library loader to the scheduler
+func init() {
+	if pkgconfigsetup.Datadog().GetBool("shared_library_check.enabled") {
```
pgimalac: You can make an early stop instead, and reduce the indentation level by one after

```go
	if !pkgconfigsetup.Datadog().GetBool("shared_library_check.enabled") {
		return
	}
```

pkg/collector/sharedlibrary/init.go
```diff
@@ -0,0 +1,40 @@
...
+			priority := 40
+			return loader, priority, err
+		}
+
+		loaders.RegisterLoader(factory)
+
+		log.Infof("Shared library checks are enabled. Looking for shared libraries in %q.", libFolderPath)
+		log.Warn("Shared library checks are still experimental. Be careful when using it, breaking changes can be made.")
+	} else {
+		log.Info("Shared library checks are disabled.")
```
pgimalac: I don't think we need to log if it's disabled, considering it'll be disabled for most users at the beginning

pkg/collector/sharedlibrary/init.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/DataDog/datadog-agent/pkg/aggregator/sender"
+	"github.com/DataDog/datadog-agent/pkg/collector/check"
+	"github.com/DataDog/datadog-agent/pkg/collector/loaders"
+	pkgconfigsetup "github.com/DataDog/datadog-agent/pkg/config/setup"
+	"github.com/DataDog/datadog-agent/pkg/util/log"
+	"github.com/DataDog/datadog-agent/pkg/util/option"
+)
+
+// init adds the shared library loader to the scheduler
+func init() {
```
pgimalac: I know we do this in an `init` function for Python and Go check, but ideally this should be added in the `run` / `check` subcommands of the agent, could you try that ?
In particular it would be nice to make sure it runs after the logger is initialized

pkg/collector/sharedlibrary/inter.go
```diff
@@ -0,0 +1,42 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+// this file is identical to pkg/collector/python/inter.go
```
pgimalac: We should probably move this to `pkg/util` then

pkg/collector/sharedlibrary/inter.go
pgimalac, line 1: A `unique` package has been added in Go 1.23 so maybe this could be replaced

pkg/collector/sharedlibrary/check_context.go
pgimalac, line 1: Why do we need this ? We don't have anything similar in Python that I know of
I'm not a big fan of this global state

pkg/collector/sharedlibrary/helpers.go
```diff
@@ -0,0 +1,67 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+// the content of this file is a part of pkg/collector/python/helpers.go
```
pgimalac: Same comment as previously, we should be able to reuse the functions defined in the other package ?

pkg/collector/sharedlibrary/loader.go
```diff
@@ -0,0 +1,78 @@
...
+type SharedLibraryCheckLoader struct {
+	logReceiver option.Option[integrations.Component]
+	loader      libraryLoader
+}
+
+// NewSharedLibraryCheckLoader creates the checks loader
+func NewSharedLibraryCheckLoader(senderManager sender.SenderManager, logReceiver option.Option[integrations.Component], tagger tagger.Component, filter workloadfilter.Component, loader libraryLoader) (*SharedLibraryCheckLoader, error) {
+	initializeCheckContext(senderManager, logReceiver, tagger, filter)
+	return &SharedLibraryCheckLoader{
+		logReceiver: logReceiver,
```
pgimalac: Do we need to store `logReceiver` ? I didn't see it used anywhere ?

pkg/collector/sharedlibrary/aggregator.go
Enzu83: Since I plan now to move the submit functions out of the Python package with [this PR](https://github.com/DataDog/datadog-agent/pull/42433/files), I can remove all the copied files from Python. This will largely simplify the PR.

About using external functions in Cgo, you can call every exported Go functions so there's nothing more to do to support that.

pkg/collector/sharedlibrary/ffi/library_loader.go
Enzu83: It could happen if the shared library checks folder is the same as the folder where RTLoader shlibs are (but we'll already have name conflicts beforehand if we do that)

pkg/config/setup/config.go
Enzu83: No we don't, by default it should be disabled. I was about to commit the change.

pkg/collector/sharedlibrary/init.go
Enzu83: Since it's experimental, there's indeed no need to indicate that the feature is disabled

pkg/collector/sharedlibrary/library_loader.go
Enzu83: Same answer as previous comment, it's indeed not a problem to reuse exported Go functions from the Python package

pkg/collector/sharedlibrary/init.go
Enzu83: i removed the `init` function and add the logic directly in the commands, it caused some issues with the e2e test because the shared library checks loader wasn't initialized at startup (the flag to enable it is on `false` by default)

pkg/collector/sharedlibrary/loader.go
Enzu83: it's not used so it can indeed be removed

pkg/collector/sharedlibrary/library_loader.go
Enzu83: i agree that it's nice to have the whole struct for `Close`, in case we'd want to call some symbols before closing the lib

Enzu83: I'm not sure to understand why `Close` should be located elsewhere. the `libraryLoader` exposes functions to handle shared libraries so closing shared libraries is done here

pkg/collector/sharedlibrary/check.go
Enzu83: yes, i didn't update the comment

pkg/collector/sharedlibrary/ffi.c
Enzu83: yes, it's better to use `RTLD_NOW` since we want to be sure that the library loaded has no problems at load time instead at the check execution time.
`RTLD_GLOBAL` is not ideal indeed in this case, `RTLD_LOCAL` is better to avoid conflicts.

pkg/collector/sharedlibrary/library_loader.go
pgimalac: For example consider the `File` API of Go, you call `os.Open` to get a `File`, but then you don't call `os.Close` to close it, `Close` is a method of the `File` type directly, which makes sense because that way all the logic is within the type

That's subjective, I won't push for it, it was just a thought

pkg/collector/sharedlibrary/ffi/library_loader.go
pgimalac: Ah yeah rtloader is in `embedded/lib`, but checks will be in `checks.d`, so no risk there I guess

pkg/collector/sharedlibrary/library_loader.go
Enzu83: I see now, we could implement something similar.
But I like the idea of using only one component for interacting with shared libraries (since it involves Cgo and C code).

pkg/collector/sharedlibrary/ffi/library_loader.go
pgimalac: Maybe just ensure that the file is owned by `dd-agent` ?

pkg/collector/sharedlibrary/check_context.go
pgimalac: My bad it exists in Python

pkg/collector/sharedlibrary/check.go
Enzu83: The synchronization should be guaranteed by the scheduler/runner so I don't know if we should do more here

pkg/collector/sharedlibrary/ffi/library_loader.go
Enzu83: As a first approach, I'd prefer to look only for the exact name (even if it's a symlink) to keep it simple.

pkg/collector/sharedlibrary/ffi/ffi.c
Enzu83: Once we'll have a version that is working well and we identified the relevant errors, we could replace the error strings by error codes. For a first prototype, it's easier to keep the strings

pkg/collector/sharedlibrary/ffi/library_loader.go
Enzu83: Just in case, I'll keep the default case but it should indeed not be relevant

pkg/collector/sharedlibrary/ffi/library_loader.go
Enzu83: Since they're custom checks with shared libraries created by customers, I don't think we should have restrictions on which libraries can be used for checks

pducolin (approved):

aiuto (approved):
For agent-build

pkg/collector/aggregator/inter.go
pgimalac, line 1: This is duplicated from pkg/collector/python/inter.go, could you factor both implementations into a unique package, and use it in both ?

pkg/collector/aggregator/check_context.go
```diff
@@ -19,21 +17,21 @@ import (
...
 // per dependency used inside SubmitMetric like methods.
-type checkContext struct {
+type CheckContext struct {
 	senderManager sender.SenderManager
-	logReceiver   option.Option[integrations.Component]
-	tagger        tagger.Component
-	filter        workloadfilter.FilterBundle
+	LogReceiver   option.Option[integrations.Component]
+	Tagger        tagger.Component
+	Filter        workloadfilter.FilterBundle
```
pgimalac: I think adding accessors for those rather than exporting them would avoid the risk of changing the value of the field

pkg/collector/aggregator/check_context.go
```diff
@@ -61,6 +60,9 @@ func initializeCheckContext(senderManager sender.SenderManager, logReceiver opti
 	checkContextMutex.Unlock()
 }
 
+// releaseCheckContext is only used in test files
+//
+//nolint:unused
```
pgimalac: In that case can you move it to a file with a `//go:build test` directive ?

pkg/collector/aggregator/helpers.go
pgimalac, line 1: Ideally this shouldn't be duplicated but Go doesn't play well with C types so it's either that or using `unsafe.Pointer` everywhere

pkg/collector/sharedlibrary/sharedlibraryimpl/loader.go
```diff
@@ -0,0 +1,75 @@
...
+// SharedLibraryCheckLoaderName is the name of the Shared Library loader
+const SharedLibraryCheckLoaderName string = "sharedlibrary"
+
+// CheckLoader is a specific loader for checks living in this package
+type CheckLoader struct {
+	loader ffi.LibraryLoader
+}
+
+// NewSharedLibraryCheckLoader creates the checks loader
+func NewSharedLibraryCheckLoader(_ sender.SenderManager, _ option.Option[integrations.Component], _ tagger.Component, _ workloadfilter.Component, loader ffi.LibraryLoader) (*CheckLoader, error) {
```
pgimalac: I believe this can be unexported

test/new-e2e/tests/agent-runtimes/checks/shared-library/files/libdatadog-agent-example.dll
pgimalac, line 1: I'm not a fan of checking the .dll and .so in git, ideally we should build them on the fly or pull them from eg. s3

pgimalac: In particular if the source code generating those shared libraries is not in this PR 😄

test/new-e2e/tests/agent-runtimes/checks/shared-library/common.go
```diff
@@ -0,0 +1,72 @@
...
+	data := checkutils.ParseJSONOutput(v.T(), []byte(check))
+	metrics := data[0].Aggregator.Metrics
+
+	// only one metric should have been emitted
+	assert.Equal(v.T(), len(metrics), 1)
+
+	// check metric info
+	metric := metrics[0]
+	assert.Equal(v.T(), "hello.gauge", metric.Metric)
+	assert.Equal(v.T(), metric.Points[0][1], 1.0)
```
pgimalac: Would be nice to test more of the API, in particular service checks and events

pgimalac (approved):
LGTM, mostly nitpicks !
Checking in the `dll` and `so` files for the e2e test is not great though, we should look for a workaround.

pkg/collector/aggregator/inter.go
Enzu83: Indeed no need to keep these instructions in the `python` package. They can come from `pkg/collector/aggregator`

pkg/collector/aggregator/check_context.go
Enzu83: Done. thanks!

test/new-e2e/tests/agent-runtimes/checks/shared-library/files/libdatadog-agent-example.dll
Enzu83: The best would be to have them compiled just for the e2e test, I'll look into that

pkg/collector/aggregator/helpers.go
Enzu83: I replaced the `**C.char` parameter by `unsafe.Pointer` and removed the duplication, I think it's better to not have the code duplicated even if we need to do an additional type cast

pkg/collector/sharedlibrary/ffi/library_loader.go
Enzu83: WIll add permissions verification in a following PR

test/new-e2e/tests/agent-runtimes/checks/shared-library/common.go
Enzu83: I expanded the test by checking for a service check and an event too!

test/new-e2e/tests/agent-runtimes/checks/shared-library/files/libdatadog-agent-example.dll
Enzu83: Will work on it in a following PR (likely need to have a new job for the Rust compilation that comes right before the e2e test job)

PR #39680, opened 2025-08-08
Add startup failure reason to remote-config status

maeve-sechrist:
### What does this PR do?
Adds to the Remote Configuration status the startup failure reason if the service fails to start up

### Motivation
If remote configuration fails during start up, the error is logged but currently the status message is not updated with the correct failure reason. This change provides transparency into why remote configuration failed to start up.

To test the changes, I set the remote configuration service to look for a database file that it would not have permission to read and confirmed the status message matched. I also started the agent normally and confirmed the remote configuration status message was as expected.

#### Failed Status
```
====================
Remote Configuration
====================


Startup Failure Reason: could not remove existing rc db (/tmp/remote-config.db): remove /tmp/remote-config.db: permission denied
```

#### Successful Status
```
====================
Remote Configuration
====================


Organization enabled: True
API Key: Authorized
Last error: None
```

### Additional Notes

Also includes a typo fix for a log message

### DIFF
--- a/pkg/config/remote/service/util.go
+++ b/pkg/config/remote/service/util.go
@@ -55,7 +55,7 @@ func recreate(path string, agentVersion string, apiKeyHash string, url string) (
 	if err == nil {
 		err := os.Remove(path)
 		if err != nil {
-			return nil, fmt.Errorf("could not remote existing rc db (%s): %v", path, err)
+			return nil, fmt.Errorf("could not remove existing rc db (%s): %v", path, err)
 		}
 	}
 	err = os.MkdirAll(filepath.Dir(path), 0700)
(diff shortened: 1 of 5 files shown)


pkg/config/remote/service/util.go
```diff
@@ -55,7 +55,7 @@ func recreate(path string, agentVersion string, apiKeyHash string, url string) (
 	if err == nil {
 		err := os.Remove(path)
 		if err != nil {
-			return nil, fmt.Errorf("could not remote existing rc db (%s): %v", path, err)
+			return nil, fmt.Errorf("could not remove existing rc db (%s): %v", path, err)
```
moth-a, line 58: 👏

moth-a (approved):
Awesome!! One note - I think the qa label should be `done`, not `no-changes`

PR #39681, opened 2025-08-08
pkg/dyninst: Fix package name in dyninst 

grantseltzer:
### What does this PR do?

I keep getting linter errors in my PRs that read as follows:

```
golangci-lint run  --timeout 25m0s --concurrency 4 --build-tags "ec2 grpcnotrace linux_bpf netcgo npm nvml test trivy trivy_no_javadb zlib zstd" --path-prefix "."  ././pkg/...
pkg/dyninst/doc.go:1: : found packages dyninst (decoder_error_handling_test.go) and dyninst_test (test_utilities.go) in pkg/dyninst (typecheck)
// Unless explicitly stated otherwise all files in this repository are licensed
section_end:1754675369:6902505e-977f-4d52-912c-a7933cfb28b5
section_start:1754675369:37552487-a409-455c-aa66-741e0b75e7be
Linter execution time...
Execution time summary:
- golangci-lint: 135.3s
```
([link](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1072057669))

I believe that this fixes it, but not sure if this breaks intention with how we're naming the packages.

More so for discussion: Should we get rid of the tests in the top level dyninst altogether and put them in `pkg/dyninst/dyninsttest`? The package godoc in `pkg/dyninst/doc.go` doesn't really apply anymore either?

### DIFF
--- a/pkg/dyninst/main_test.go
+++ b/pkg/dyninst/main_test.go
@@ -5,7 +5,7 @@
 
 //go:build linux_bpf
 
-package dyninst
+package dyninst_test
 
 import (
 	"os"


ajwerner:
This change would be fine, but doesn't do anything and I do not think it relates at all to the problem you're trying to solve.

PR #39689, opened 2025-08-08
set service to manual mode in driver verifier tests

clarkb7:
### What does this PR do?
disable service auto-start during startstop-tests that enable the driver verifier.

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1714

tests flaking with error
```
Error:      	Received unexpected error:
        	            	sftp: "Bad message" (SSH_FX_BAD_MESSAGE)
        	Test:       	TestDriverVerifierOnServiceBehaviorAgentCommand/TestAgentStartsAllServices
        	Messages:   	should remove agent.log
```

I think this is due to the log files being in use. The test normally stops the services before this, but when the driver verifier is enabled the host is rebooted which creates a race between the stopservice call and the OS auto starting the services.

call `host.Remove("/ProgramData/Datadog/logs/agent.log")` while the Agent is running and observe the `SSH_FX_BAD_MESSAGE` error

### DIFF
--- a/test/new-e2e/tests/installer/windows/base_suite.go
+++ b/test/new-e2e/tests/installer/windows/base_suite.go
@@ -259,6 +259,18 @@ func (s *BaseSuite) BeforeTest(suiteName, testName string) {
 	s.installer = NewDatadogInstaller(s.Env(), s.CurrentAgentVersion().MSIPackage().URL, outputDir)
 	s.installScriptImpl = NewDatadogInstallScript(s.Env())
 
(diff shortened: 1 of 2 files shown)


clarkb7:
if we get https://github.com/DataDog/datadog-agent/pull/35679 merged then it improves the error message in this case

with sftp@v1.13.9 the error is `permission denied`

BaptisteFoy (approved):

lrbhatia (approved):

PR #39690, opened 2025-08-08
Add bugcheck data to Windows crash events

alexn-dd:
### What does this PR do?
This PR fills the bugcheck code field, as well as provide bugcheck arguments, as part of the Windows crash detection events.  

### Motivation
The bugcheck code was originally filled but no longer reported after past updates.
The bugcheck code will help better understand customer crashes and provide better diagnostics data.
https://datadoghq.atlassian.net/browse/WINA-1705

Manual testing:
- Launch system probe in a test VM.
- Drop a known kernel memory dump file into C:\Windows\MEMORY.DMP.
- Copy NamedPipeCmd.exe to the test VM.
- With an elevated command prompt, run the following command twice:
-- `NamedPipeCmd.exe -method GET -path /windows_crash_detection/check`
-- Twice is required due to the asynchronous processing of crash detection.
- A JSON output should be provided by the command, which should include the bugcheck code from the kernel dump.

### DIFF
--- a/pkg/collector/corechecks/system/wincrashdetect/probe/crashdump.cc
+++ b/pkg/collector/corechecks/system/wincrashdetect/probe/crashdump.cc
@@ -1,6 +1,8 @@
 #include "crashdump.h"
 #include "_cgo_export.h"
 
+#include <stdio.h>
+
 #define INITGUID
 #include "guiddef.h"
 DEFINE_GUID(IID_IDebugControl, 0x5182e668, 0x105e, 0x416e,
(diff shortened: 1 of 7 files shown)


pkg/collector/corechecks/system/wincrashdetect/probe/crashdump.cc
```diff
@@ -55,58 +55,83 @@ STDMETHODIMP StdioOutputCallbacks::Output(THIS_ IN ULONG, IN PCSTR Text) {
...
     }
+
+    hr = iControl->ReadBugCheckData(
+             &bugCheckInfo->code,
+             &bugCheckInfo->arg1,
+             &bugCheckInfo->arg2,
+             &bugCheckInfo->arg3,
+             &bugCheckInfo->arg4);
+    if (S_OK != hr) {
+        // OK to fail, this may not be a proper kernel dump, try get other info.
```
sar-shah: Should we log any errors here?

alexn-dd: There is no logging facility in cgo, at least for Windows.  At best people use `fprintf`, but given that system-probe runs as a service, that print is not visible.

alexn-dd: I can use `fprintf` here and in the future the Windows agent can setup redirection with `SetStdHandle `.

cswatt (approved):

pkg/collector/corechecks/system/wincrashdetect/probe/crashdump.cc
iglendd: As long as extended error is logged on the Go side we should be good IMHO

pkg/collector/corechecks/system/wincrashdetect/probe/crashdump.cc
```diff
@@ -55,58 +57,85 @@ STDMETHODIMP StdioOutputCallbacks::Output(THIS_ IN ULONG, IN PCSTR Text) {
...
     IDebugClient* iClient = NULL;
     IDebugControl* iControl = NULL;
     StdioOutputCallbacks iOutputCb(ctx);
     READ_CRASH_DUMP_ERROR ret = RCD_NONE;
-    HRESULT hr = DebugCreate(IID_IDebugClient, (void**)&iClient);
-    if(S_OK != hr) {
+    HRESULT hr = E_FAIL;
+
+    if (NULL == bugCheckInfo) {
+        *extendedError = E_FAIL;
```
lrbhatia: Check extendedError for possible NULL before dereferencing it.

lrbhatia (approved):

pkg/collector/corechecks/system/wincrashdetect/probe/crashdump.cc
alexn-dd: The change will log the extended error on the Go side.

PR #39691, opened 2025-08-09
network v2: cleanup of ethtool handling

paulcacheux:
### What does this PR do?

Related to https://github.com/DataDog/datadog-agent/pull/39679 that fixed an issue with closing the ethtool object and the related incident.

This PR cleans up the ethtool handling in the network v2 check around multiple dimensions:

#### ethtool lifetime:
in order to test this check, the function to get the driver info and stats are put into global variables ready to be overridden in tests. This is not great because it obfuscates the lifetime of the ethtool object (see the ptr/object/close dance that is on main currently). This PR refactors this to actually mock the fetching of the ethtool object itself. This allows the actual code to use the ethtool library directly, and the tests don't interact with ethtool at all and simply returns a mocked object directly

#### ethtool hoisting:
the ethtool object creation is hoisted up the loop, allowing to create a single ethtool object (which is a socket under the hood) once per check instead of once per interface

In addition to those points, the interface name extra copy is removed

### DIFF
--- a/pkg/collector/corechecks/net/networkv2/network.go
+++ b/pkg/collector/corechecks/net/networkv2/network.go
@@ -287,13 +287,13 @@ func handleEthtoolStats(sender sender.Sender, ethtoolObject ethtoolInterface, in
 	}
 
 	// Preparing the interface name and copy it into the request
-	ifaceBytes := []byte(interfaceIO.Name)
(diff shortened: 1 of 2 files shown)


jeremy-hanna (approved):
👍 thanks for cleaning this up!

PR #39391, opened 2025-07-30
WorkloadFilter Lazy Loading

gabedos:
### What does this PR do?

Migrates the workload filtering component to lazily initialization the filtering programs

### Motivation

Memory efficiency for unused filters and thus now will never be created.

### Describe how you validated your changes

Unit Tests and CI

### Possible Drawbacks / Trade-offs

Shifts and distributes cost (memory/cpu) related to initialization of filters to the first caller. For example, the first iteration of a container corecheck would be slightly slower, but every iteration after will have the filters compiled and ready to use.

### Additional Notes

N/A

PR #39694, opened 2025-08-10
Optimize WorkloadFilter Helper Initializations

gabedos:
### What does this PR do?

* Caches the results of filter utils like `GetAutodiscoveryFilters` and `GetContainerSharedMetricsFilters` which can be expensive querying the Datadog Config every time the constant values are used.
* Creates 1 source for Paused Container filters that are shared. These are expensive regexes and don't need to be duplicated. It was previously also embedded within the SBOM filter but are now using the global `ContainerPaused` program.

### Motivation

Optimize memory overhead of using the workloadFilter component.

Unit Tests and CI

N/A


comp/core/workloadfilter/catalog/container.go
```diff
@@ -130,24 +130,13 @@ func LegacyContainerSBOMProgram(config config.Component, logger log.Component) p
 	programName := "LegacyContainerSBOMProgram"
 	var initErrors []error
 
-	if !config.GetBool("sbom.enabled") && !config.GetBool("sbom.container_image.enabled") && !config.GetBool("sbom.container.enabled") {
```
gabedos, line 133: With lazy loaded introduced in this PR https://github.com/DataDog/datadog-agent/pull/39391, we no longer require manually preventing the creation of SBOM filters when they aren't used. This is already done by the structure of lazy loading itself.

comp/core/workloadfilter/def/utils.go
```diff
@@ -24,66 +26,87 @@ const (
...
+	// Cache for shared metric filters
+	containerSharedMetricFiltersCache [][]ContainerFilter
+	containerSharedMetricOnce         sync.Once
 
-	// TODO: Add config option for users to configure AD annotations to take lower priority
-	high := []ContainerFilter{ContainerADAnnotations}
-	low := []ContainerFilter{LegacyContainerGlobal}
+	// Cache for sbom filters
+	containerSBOMFiltersCache [][]ContainerFilter
+	containerSBOMOnce         sync.Once
```
clamoriniere, line 40: Could you explain the reasoning behind defining global variables and caching the results in this implementation?

An alternative approach I was considering would be to invoke the `config.Datadog().Get...()` methods once per call to [`newFilter()`](https://github.com/DataDog/datadog-agent/blob/daa795197efcf15458d1ffeb8e4adc70f94c5368/comp/core/workloadfilter/impl/filter.go#L97C6-L97C15), and encapsulate the results in a dedicated `filterConfigSettings` struct to hold all relevant config values extracted from the `config` package.

```golang
// filter is the implementation of the filter component.
type filter struct {
	config              filterConfigOptions
	log                 log.Component
	telemetry           coretelemetry.Component
	programFactoryStore map[workloadfilter.ResourceType]map[int]*filterFactory
}

type filterConfigOptions struct {
  container_include []string
  container_exclude []string

  container_include_metrics []string
  container_exclude_metrics []string

  //....
}

```

The main motivation for this would be to make it easier to support **configuration reloads** in the future. Avoiding global caching could simplify re-evaluating config state dynamically without having to manage additional invalidation logic.

Let me know if this makes sense, happy to clarify if anything is unclear!

gabedos:
Synced offline with Cedric:

Thoughts are to rely on the "component initialization" function to get the config option values and then have the function `GetContainerSharedMetricFilters()` as a function member of the WorkloadFilter component.

PR #39695, opened 2025-08-10
[USMON-1463] Allow CNM & USM to report host/container tags for configurable duration during startup

guyarb:
### What does this PR do?

Importing the logic of `expected_tags_duration` from the core agent to system-probe (CNM and USM).

Allowing both products to report host/container tags in our payload, to handle race in the backend pipelines.

### Motivation

Today, a if a new host is spawned, there's a race between the pipeline that processes the host and container tags, to the CNM/USM pipeline which tries to pull those tags from the other pipeline.

The issue is "known" in the agent, and handled using a mechaism to report host tags on the relevant metrics for a configurable duration after the startup of the agent.

We import this logic to USM/CNM, and enabling it by default.

### Additional Notes

Note - we don't use the core-agent configuration, as it is not enabled by default while the system-probe configuration is.

### DIFF
--- a/pkg/config/setup/system_probe.go
+++ b/pkg/config/setup/system_probe.go
@@ -312,6 +312,8 @@ func InitSystemProbeConfig(cfg pkgconfigmodel.Setup) {
 	cfg.BindEnv(join(netNS, "http_max_request_fragment"))
 	cfg.BindEnv(join(smNS, "http_max_request_fragment"))
 
+	cfg.BindEnvAndSetDefault(join(spNS, "expected_tags_duration"), 10*time.Minute, "DD_SYSTEM_PROBE_EXPECTED_TAGS_DURATION")
+
 	// list of DNS query types to be recorded
 	cfg.BindEnvAndSetDefault(join(netNS, "dns_recorded_query_types"), []string{})
 	// (temporary) enable submitting DNS stats by query type.
--- a/pkg/config/setup/system_probe.go
+++ b/pkg/config/setup/system_probe.go
@@ -312,7 +312,7 @@ func InitSystemProbeConfig(cfg pkgconfigmodel.Setup) {
 	cfg.BindEnv(join(netNS, "http_max_request_fragment"))
 	cfg.BindEnv(join(smNS, "http_max_request_fragment"))
 
-	cfg.BindEnvAndSetDefault(join(spNS, "expected_tags_duration"), 10*time.Minute, "DD_SYSTEM_PROBE_EXPECTED_TAGS_DURATION")
+	cfg.BindEnvAndSetDefault(join(spNS, "expected_tags_duration"), 5*time.Minute, "DD_SYSTEM_PROBE_EXPECTED_TAGS_DURATION")
 
 	// list of DNS query types to be recorded
 	cfg.BindEnvAndSetDefault(join(netNS, "dns_recorded_query_types"), []string{})
--- a/comp/process/connectionscheck/connectionscheckimpl/check.go
+++ b/comp/process/connectionscheck/connectionscheckimpl/check.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/DataDog/datadog-agent/comp/core/config"
 	"github.com/DataDog/datadog-agent/comp/core/sysprobeconfig"
+	tagger "github.com/DataDog/datadog-agent/comp/core/tagger/def"
 	workloadmeta "github.com/DataDog/datadog-agent/comp/core/workloadmeta/def"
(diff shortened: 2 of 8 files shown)


guyarb:
Waiting on https://github.com/DataDog/agent-payload/pull/400 to update the go.mods

gjulianm (approved):

kkhor-datadog (commented):
@DataDog/container-experiences shouldn't need to review as our only affected files are in `comp/process/connectionscheck/`, which CNM actually owns the functionality for. But since this is the connections check, Networks should review the changes, not us.

Marking as approved on our end, but please get a review from Networks as well - we can look into updating the CODEOWNERS to reflect the ownership separately.

kkhor-datadog (approved):

pkg/aggregator/host_tag_provider.go
```diff
@@ -18,23 +18,29 @@ import (
...
 
+// NewHostTagProvider creates a new HostTagProvider with the default expected tags duration from the configuration.
 func NewHostTagProvider() *HostTagProvider {
-	return newHostTagProviderWithClock(clock.New())
+	return NewHostTagProviderWithDuration(pkgconfigsetup.Datadog().GetDuration("expected_tags_duration"))
 }
 
-func newHostTagProviderWithClock(clock clock.Clock) *HostTagProvider {
+// NewHostTagProviderWithDuration creates a new HostTagProvider with a specified duration for host tags expiration.
+func NewHostTagProviderWithDuration(duration time.Duration) *HostTagProvider {
```
vickenty: Since it is no longer specific to the aggregator, would it make sense to move this to hosttags component/package?

guyarb: do you think it is in the scope of the current change?

vickenty: I think so, there won't be a better time to do the move. Amount of code is small and there aren't much changes to it at the same time.

guyarb: I'll make it in a different PR then

guyarb: https://github.com/DataDog/datadog-agent/pull/39800

pkg/config/setup/system_probe.go
```diff
@@ -312,6 +312,8 @@ func InitSystemProbeConfig(cfg pkgconfigmodel.Config) {
 	cfg.BindEnv(join(netNS, "http_max_request_fragment"))
 	cfg.BindEnv(join(smNS, "http_max_request_fragment"))
 
+	cfg.BindEnvAndSetDefault(join(spNS, "expected_tags_duration"), 10*time.Minute, "DD_SYSTEM_PROBE_EXPECTED_TAGS_DURATION")
```
akarpz: this seems like a long time? usually the race lasts only a few seconds

akarpz: additionally, what is the the cost of sending the tags from the agent? I assume there must be one if we only do it for a few minutes on startup... is it just customer network from the larger payload size? if the downsides are small, why not just do this forever and remove some backend complexity?

guyarb: > usually the race lasts only a few seconds

That's hard to say, as there's no metric to support that
AFAIK, everytime the core-agent products enable the feature, they do that with 10m or 15m
I'll set it for 5m by default, and in rare cases we can ask the customer to increase it.

> additionally, what is the the cost of sending the tags from the agent? I assume there must be one if we only do it for a few minutes on startup... is it just customer network from the larger payload size? if the downsides are small, why not just do this forever and remove some backend complexity?

The side effects are:
* larger payload of CNM and USM
* CPU consumption by process-agent (fetching the tags)

We should not rely on this as a robust solution and drop backend resolution, as we can spare redudnant calculations for the customer side (The information exists in the backend, it might just take a couple of minutes to fully arrive, parse and index). Also, the customer is able to add user-defined tags from the UI. Those tags won't be available in the agent side.

guyarb: re: payload size. AFAIK, CNM (connections check) does not measure the payload size, or it is not feasiable to filter it correctly to distinguish between agents with the feature and the agent without it.

DanielLavie (approved):

pkg/config/setup/system_probe.go
akarpz: > re: payload size. AFAIK, CNM (connections check) does not measure the payload size, or it is not feasiable to filter it correctly to distinguish between agents with the feature and the agent without it.

agreed... I think we only measure the # of connections in the payload. we should probably measure the total size as well to know the impact of changes like this, but I don't think we need to block this PR on it.

akarpz (approved):

PR #39707, opened 2025-08-11
service discovery: Split information about TCP and UDP ports

vitkyrka:
### What does this PR do?

It could be useful to know what kind of ports we're dealing with so
store this information instead of dropping it.  The payload already
allows for this to separated.

### Motivation

https://datadoghq.atlassian.net/browse/DSCVR-173

Unit test modified.


smeso-ddg (approved):

pkg/process/checks/process_linux.go
```diff
@@ -121,26 +122,40 @@ func mapWLMProcToProc(wlmProc *workloadmetacomp.Process, stats *procutil.Stats)
...
-// formatPorts converts a list of uin16 ports to a int32 PortInfo
-// TODO: because the service discovery response does not distinguish between tcp and udp currently, we currently send everything as TCP
-func formatPorts(ports []uint16) *model.PortInfo {
-	// if ports were not collected, we want to semantically indicate that no data was collected instead of
+// formatPorts converts separate TCP and UDP uint16 port lists to a int32 PortInfo
+func formatPorts(tcpPorts, udpPorts []uint16) *model.PortInfo {
+	// if both port lists are nil, we want to semantically indicate that no data was collected instead of
 	// returning no open ports
-	if ports == nil {
+	if tcpPorts == nil && udpPorts == nil {
```
daniel-taf: Just to make sure I understand correctly, if we collected only TCP ports, then the UDP ports slice will be `nil` and not `[]int32{}`.

daniel-taf (approved):

pkg/process/checks/process_fallback.go
```diff
@@ -32,7 +32,7 @@ func (p *ProcessCheck) processesByPID() (map[int32]*procutil.Process, error) {
 }
 
 // formatPorts is a stub for non-linux platforms
-func formatPorts(_ []uint16) *model.PortInfo {
+func formatPorts(_, _ []uint16) *model.PortInfo {
```
daniel-taf: missing `bool` parameter to match linux version.

adel121 (approved):

PR #39708, opened 2025-08-11
[POC][Agent-Health] Create component for agent health check management (WIP)

louis-cqrl:


comp/core/health-platform/README.md
```diff
@@ -0,0 +1,180 @@
+# Health Platform Component
+
+The Health Platform component provides a mechanism for collecting and reporting health information from the Datadog Agent to the backend. It includes hostname information, host ID from gopsutil, organization ID, and a list of issues with their details.
```
iglendd: Should you also add something along the following line "... to show customer in Datadog UI that that particular Agent is healthy and instructions to address it or notify user by other means"

comp/core/health-platform/README.md
```diff
@@ -0,0 +1,180 @@
...
+
+import (
+    "context"
+    "time"
+
+    "github.com/DataDog/datadog-agent/comp/core/health-platform"
+    "github.com/DataDog/datadog-agent/comp/core/health-platform/fx"
+)
+
+func example(healthPlatform healthplatform.Component) {
```
iglendd: I wonder if word platform is needed. Do we have many other components with word health?

comp/core/health-platform/SUMMARY.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Overview
+
+I have successfully created a new component called "health-platform" following the Datadog Agent's established architecture patterns. This component collects and reports health information from the host system, sending it to the Datadog backend with hostname, host ID (from gopsutil), organization ID, and a list of issues.
+
+## Files Created
+
+### 1. Protobuf Definitions
+
+- **`pkg/proto/datadog/health-platform/health-platform.proto`** - Protocol buffer schema defining Issue and HealthReport messages
```
iglendd: I wonder how we reconsoliate backend protobuf schema with client protobuf schema. What other Agent parts are doing about it.

comp/core/health-platform/SUMMARY.md
```diff
@@ -0,0 +1,183 @@
...
+# Organization ID (required)
+org_id: 12345
+
+# Alternative: API key org ID
+api_key_orgid: 12345
+
+# Reporting interval (default: 15 minutes)
+health_platform:
+  interval: 15m
+```
```
iglendd: README and SUMMARY have lot of duplicate content?

comp/core/health-platform/impl/health-platform.go
```diff
@@ -0,0 +1,319 @@
...
+	if len(issues) == 0 {
+		log.Debug("No issues to report")
+		return nil
+	}
+
+	// Get hostname (with nil check for testing)
+	var hostname string
+	var err error
+	if c.hostname != nil {
+		hostname, err = c.hostname.Get(ctx)
```
iglendd: Should host name be taken during init since it would not change after start, right?

louis-cqrl: IIRC the hostname is cached right after first iteration, so we are just fetching the cache by calling this method, however that's a good point it will avoid calling multiple time a function

comp/core/health-platform/SUMMARY.md
louis-cqrl: deleted one, i will rewrite the readme at the end

comp/core/health-platform/README.md
louis-cqrl: healthprobe, i would avoid to use only health since we may mistake with other health cmd in the agent.

(I saw you were talking about a renaming to `recommendation` that might also be a solution)

PR #39712, opened 2025-08-11
upgrade(fleet): Agent size improvements

BaptisteFoy:
### What does this PR do?
Reduces the Agent size:
- Removes `embedded/include` directory from the final build
- Empties useless `.installed_by_pkg.txt` file
  - It was used to avoid removing files tracked by the deb / rpm during deb / rpm removal... The files are removed anyway
- Compress some licenses

### Motivation
Make room for https://github.com/DataDog/datadog-agent/pull/39721


BaptisteFoy:

omnibus/config/software/datadog-agent-finalize.rb
```diff
@@ -110,7 +110,6 @@
             # have been setup by the installer, this way, on removal, we'll be able to delete only files
             # which have not been created by the package.
             command "echo '# DO NOT REMOVE/MODIFY - used by package removal tasks' > #{install_dir}/embedded/.installed_by_pkg.txt"
```
san-jos, line 112: **❓ question:** ‏Is it still needed

san-jos (approved):

omnibus/config/software/datadog-agent-finalize.rb
BaptisteFoy: yes

JSGette (approved):

omnibus/config/projects/agent.rb
```diff
@@ -297,7 +297,7 @@
 exclude 'bundler\/git'
 
 # Exclude headers that are not needed in the final package
-exclude "embedded/include/systemd"
+exclude "embedded/include"
```
alopezz, line 300: As discussed on Slack, we need to at least leave `Python.h` in for the moment, given that our last attempt at dropping headers showed that setting up the cacti integration require, and required [a revert](https://github.com/DataDog/datadog-agent/pull/20274).

This seems to come from the fact that[ we instruct users that want to set up the cacti integration to install the rrdtool python package](https://github.com/DataDog/integrations-core/tree/master/cacti#python-bindings). And that package contains a C extension that [requires Python.h](https://github.com/commx/python-rrdtool/blob/5d8c7bd59608f24cb1c9d240d0ce546b0a22c98b/rrdtoolmodule.c#L25).

alopezz (changes requested):
Blocking based on the risk of breaking an integration as discussed.

alopezz (changes requested):
Blocking based on the risk of breaking an integration as discussed.

PR #39714, opened 2025-08-11
[EBPF-739] gpu: reduce timeouts for querying system-probe

gjulianm:
### What does this PR do?

This PR updates the timeouts for the CheckClient used in the gpu check, so that we avoid issues with weighted metrics being reported at different intervals. Now the agent should emit metrics at the check interval consistently even if system-probe is not responding ocrrectly.

### Motivation

Long system-probe response times (usually at startup) can cause the check to be delayed and weighted metrics suffer from precision issues (https://datadoghq.atlassian.net/browse/EBPF-794).

CI tests passing.


jasondellaluce (approved):

dustmop (approved):

PR #39715, opened 2025-08-11
[CONTINT-4735]: Add pod termination duration metric to kubelet check

bitomaxsp:
### What does this PR do?
Add `kubernetes.pod.terminating.duration` metric to highlight the PODs that stuck beyond `deletionTimestamp` in Termination phase.
Metric is only emitted if `deletionTimestamp` is present and before Now().

### Motivation
Required in order to monitor POD that hang in Termination stage for a long time

1. Create Kind cluster: `kind create cluster --config config.yaml -n kind-4735`
config.yaml:
```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
  - role: control-plane
  - role: worker
  - role: worker
```
2. Build agent change in dev container: `inv agent.hacky-dev-image-build`
3. Load the image to the cluster: `kind load docker-image testload:latest -n kind-4735`
4. Deploy agent using helm: `helm install ddtest datadog/datadog -f agent-values.yaml`
5. Make a test deployment with POD spec containing a finilizer:
```yaml
spec:
  template:
    metadata:
      finalizers:
        - example.com/my-finalizer
```
6. Delete successfully deployed testload POD and expect metric to appear after grace peroid
Screenshot 2025-08-12 at 16 27 17

### DIFF
--- a/releasenotes/notes/add-pod-termination-duration-metric-8827f349dd3cc15b.yaml
+++ b/releasenotes/notes/add-pod-termination-duration-metric-8827f349dd3cc15b.yaml
@@ -9,5 +9,4 @@
 enhancements:
   - |
     Add the new metric `kubernetes.pod.terminating.duration` to kubelet check to track pods
-    that are stuck in the termination phase. The metric will be emitted after resource `deletionTimestamp`
-    is in the past relative to now.
+    that are stuck in the termination phase. If `deletionTimestamp` is set to a time in the future,`kubernetes.pod.terminating.duration` is only emitted when the current time reaches the time set in `deletionTimestamp`.
--- a/releasenotes/notes/add-pod-termination-duration-metric-8827f349dd3cc15b.yaml
+++ b/releasenotes/notes/add-pod-termination-duration-metric-8827f349dd3cc15b.yaml
@@ -8,6 +8,6 @@
 ---
 enhancements:
   - |
-    Add new metric `kubernetes.pod.terminating.duration` to kubelet check to track PODs
-    that are stuck in Termination phase. Metric will be emitted once resource `deletionTimestamp`
+    Add the new metric `kubernetes.pod.terminating.duration` to kubelet check to track pods
+    that are stuck in the termination phase. The metric will be emitted after resource `deletionTimestamp`
     is in the past relative to now.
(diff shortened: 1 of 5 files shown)


lavigne958 (approved):

sblumenthal (commented):
The change itself looks good, but please add a unit test to ensure that we are ensuring this metric is being emitted as expected

gabedos (approved):
Update to PodMetadata looks good

releasenotes/notes/add-pod-termination-duration-metric-8827f349dd3cc15b.yaml
```diff
@@ -0,0 +1,13 @@
...
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+enhancements:
+  - |
+    Add new metric `kubernetes.pod.terminating.duration` to kubelet check to track PODs
+    that are stuck in Termination phase. Metric will be emitted once resource `deletionTimestamp`
+    is in the past relative to now.
```
maycmlee: What does it mean for the resource to be "in the past relative to now"?
```suggestion
    Add the new metric `kubernetes.pod.terminating.duration` to kubelet check to track pods
    that are stuck in the termination phase. The metric will be emitted after resource `deletionTimestamp`
    is in the past relative to now.
```

maycmlee (commented):
Some small suggestions and a question

releasenotes/notes/add-pod-termination-duration-metric-8827f349dd3cc15b.yaml
bitomaxsp: Good that you asked. It seems I failed to explain it clearly. When `deletionTimestamp` set it is in the future. When  current time reaches that point only then metric is emitted.

sblumenthal (approved):

releasenotes/notes/add-pod-termination-duration-metric-8827f349dd3cc15b.yaml
```diff
@@ -0,0 +1,13 @@
...
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+enhancements:
+  - |
+    Add the new metric `kubernetes.pod.terminating.duration` to kubelet check to track pods
+    that are stuck in the termination phase. The metric will be emitted after resource `deletionTimestamp`
+    is in the past relative to now.
```
maycmlee: Thanks @bitomaxsp for the explanation. How about this?

```suggestion
    that are stuck in the termination phase. If `deletionTimestamp` is set to a time in the future,`kubernetes.pod.terminating.duration` is only emitted when the current time reaches the time set in `deletionTimestamp`.
```

maycmlee (commented):
Made a suggestions based on your explanation!

maycmlee (approved):

PR #39721, opened 2025-08-11
feat(fleet): Split installer daemon & installer binary

BaptisteFoy:
### What does this PR do?
This PR splits the installer daemon & binary for the Linux & Darwin builds.

### Motivation
We need the installer binary to be statically linked, so we can make the OCI agent path independent while still supporting remote upgrades.

E2E are enough


san-jos (approved):

Dockerfiles/agent/Dockerfile
```diff
@@ -159,6 +160,7 @@ RUN --mount=from=artifacts,target=/artifacts \
  && find /artifacts -maxdepth 1 -name "${DD_AGENT_ARTIFACT}" -exec tar xvf {} -C . \; \
  && rm -rf usr etc/init lib \
```
JSGette, line 161: Isn't it better to do `COPY --exclude` if we don't need certain directories? This way we won't have unnecessary files stored in intermediate layers. 
Details: https://docs.docker.com/reference/dockerfile/#copy---exclude

feel free to resolve

JSGette (approved):

Dockerfiles/agent/Dockerfile
BaptisteFoy: We're copying a tar.xz, we unfortunately can't exclude directory at copy

PR #39730, opened 2025-08-11
feat(trace-agent): [SVLS-7232] AAS Extension trace Function Tags

apiarian-datadog:
### What does this PR do?
Adds Tracer Payload function tags for the trace-agent running in the Azure App Services Extension.

### Motivation
We want to support proper function tags for all serverless environments.

- [x] Deploy to AAS and validate that the trace tags get applied correctly.


apiarian-datadog:
built on top of https://github.com/DataDog/datadog-agent/pull/39471

maycmlee (approved):

apiarian-datadog:
the earlier implementation for serverless-init: https://github.com/DataDog/datadog-agent/pull/38374

pgimalac (approved):
LGTM for Agent Runtimes

comp/trace/payload-modifier/def/component.go
```diff
@@ -0,0 +1,20 @@
...
+	pkgagent "github.com/DataDog/datadog-agent/pkg/trace/agent"
+)
+
+// team: agent-apm
+
+// Component provides trace payload modification functionality
+type Component interface {
+	// GetModifier returns the TracerPayloadModifier instance
+	// Returns nil if payload modification is not enabled
+	GetModifier() pkgagent.TracerPayloadModifier
```
truthbk: This feels like a bit of a strange interface for the `Component` to me? I would've probably expected the interface for the modifier itself to be the `Component` interface, as opposed to a Getter. Even as a wrapper, I would've maybe expected a fully wrapped interface for the PayloadModifier interface as opposed to introducing a getter.

Can you elaborate a little bit on why we've made this implementation decision?

truthbk (commented):
Thank you for creating a component out of the PayloadModifier. I think we're on the right track, but I have some questions around the ultimate decision of building a component for a mere `getter`. I think my expectation would've been to see the interface for the PayloadModifier exposed by the component. Not sure if perhaps doing that would've been problematic? Happy to talk about it.

comp/trace/payload-modifier/def/component.go
apiarian-datadog: almost certainly because i'm just dipping my toes into the agent codebase and am quite unfamiliar (and somewhat frustrated) with fx.

happy to implement this in a different way if you think there is one!

what would a fully wrapped interface look like? do you mean the component itself should implement the TracerPayloadModifier interface? and if  so how would this work with a nil component when the modifier is not needed?

apiarian-datadog: I've updated the code with your suggestions, I think =]

comp/trace/payload-modifier/def/component.go
```diff
@@ -0,0 +1,20 @@
...
+package payloadmodifier
+
+import (
+	pb "github.com/DataDog/datadog-agent/pkg/proto/pbgo/trace"
+)
+
+// team: agent-apm
+
+// Component provides trace payload modification functionality
+type Component interface {
```
truthbk, line 16: 🍰

truthbk: Thanks, yeah, this is exactly what I meant. 

I was going to suggest you maybe aliased `type TracerPayloadModifier interface` in [pkg/trace/agent](https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/agent/agent.go#L146-L150) to instead use this interface, but that may take you down a bit of a rabbit hole, could lead to some circular dependencies, etc. 

If you have the energy and time I would suggest that so we don't end up with duplicate interfaces that could later lead to confusion, but otherwise it's OK, I'm aware that trace-agent cleanup is not really your responsibility.

truthbk (commented):
Thanks for following-up on the suggestion, I think this more in-line with our goals and best practices for "componentization".

truthbk (approved):

ajgajg1134 (approved):
This looks good from APM-Agent side, thanks for making the changes here :)

jcstorms1 (approved):

PR #39731, opened 2025-08-11
[APM/Stats] Extract GRPC Status Code from `status.code` when span name is `grpc.{server,client}`

vascoccorreia:
### What does this PR do?

This PR adds additional logic to `getGrpcStatusCode` in `pkg/trace/stats/aggregation.go`, allowing it to conditionally use `status.code` when the span's name matches `grpc.server` or `grpc.client`.

### Motivation

Set the GRPC status code for GRPC spans stats with the value under `status.code`.

Added additional test cases to validate the change works as intended.


ajgajg1134 (approved):
This looks ok to me, I think in parallel to this we should begin to move tracers to a more consistent naming so someday we can deprecate and remove this code. (Also please add a brief release note)

ajgajg1134 (changes requested):
(marking requested changes to discuss if this the right place for this change)

songy23 (approved):

cswatt (approved):

jccarneiro-ddog (approved):

dumontg (approved):
LGTM! Thanks for taking care of this fix!

IbraheemA (approved):

PR #39732, opened 2025-08-11
Adds ongoing_duration metric for deployments

frank-spano:
### What does this PR do?
Adds kubernetes ongoing rollout duration metrics for deployment and statefulset rollouts. 

### Motivation
Business request

#### Step 1:
Start up minikube cluster and install datadog operater per usual

#### Step 2:
Deploy agent with CRD with kubernetes state core check enabled, and permissions for agent to be able to fetch controllerRevisions (need to discuss this).
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: ksm-core-config
  namespace: default
data:
  kubernetes_state_core.yaml: |
    init_config:
    instances:
      - collectors:
        - pods
        - nodes
        - jobs
        - deployments
        - statefulsets
        - replicasets
        #pod_collection_mode: node_kubelet
        telemetry: true
        min_collection_interval: 15
---
apiVersion: datadoghq.com/v2alpha1
kind: DatadogAgent
metadata:
  name: datadog
  namespace: default
spec:
  global:
    clusterName: frank-spano-dev-cluster
    site: datadoghq.com
    credentials:
      apiSecret:
        secretName: datadog-secret
        keyName: api-key
  features:
    kubeStateMetricsCore:
      enabled: true
  override:
    clusterAgent:
      image:
        name: frankspano569/cluster-agent:frankspanodev
        pullPolicy: Always
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: cluster-agent-controllerrevisions-reader
rules:
- apiGroups: ["apps"]
  resources: ["controllerrevisions"]
  verbs: ["get", "list", "watch"]
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: cluster-agent-controllerrevisions-binding
subjects:
- kind: ServiceAccount
  name: datadog-cluster-agent
  namespace: default
roleRef:
  kind: ClusterRole
  name: cluster-agent-controllerrevisions-reader
  apiGroup: rbac.authorization.k8s.io
```

#### Step 3: 
Deploy "slow" statefulset and deployment rollouts
```
kubectl apply -f slow-stateful.yaml
kubectl apply -f slow-deploy.yaml
```

The files
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test-deployment-1min
  namespace: default
spec:
  replicas: 3
  strategy:
    type: RollingUpdate
    rollingUpdate:
      maxUnavailable: 1
      maxSurge: 1
  selector:
    matchLabels:
      app: test-deployment-1min
  template:
    metadata:
      labels:
        app: test-deployment-1min
    spec:
      containers:
      - name: app
        image: nginx:1.20
        ports:
        - containerPort: 80
        resources:
          requests:
            memory: "16Mi"
            cpu: "10m"
          limits:
            memory: "32Mi"
            cpu: "20m"
        # Slow startup - about 20 seconds per pod
        readinessProbe:
          httpGet:
            path: /
            port: 80
          initialDelaySeconds: 15  # Wait 15 seconds before first check
          periodSeconds: 3         # Check every 3 seconds
          timeoutSeconds: 2
          successThreshold: 2      # Require 2 consecutive successes (6 seconds)
          failureThreshold: 3      # Allow 3 failures before giving up
        livenessProbe:
          httpGet:
            path: /
            port: 80
          initialDelaySeconds: 25
          periodSeconds: 10
      # Add init container to slow down startup
      initContainers:
      - name: slow-init
        image: busybox:1.35
        command: ['sh', '-c', 'echo "Starting init container..." && sleep 10 && echo "Init complete"']
        resources:
          requests:
            memory: "8Mi"
            cpu: "5m"
```
and
```
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: test-statefulset-1min
  namespace: default
spec:
  serviceName: test-statefulset-1min-service
  replicas: 3
  updateStrategy:
    type: RollingUpdate
    rollingUpdate:
      maxUnavailable: 1
  selector:
    matchLabels:
      app: test-statefulset-1min
  template:
    metadata:
      labels:
        app: test-statefulset-1min
    spec:
      containers:
      - name: app
        image: nginx:1.20
        ports:
        - containerPort: 80
        resources:
          requests:
            memory: "16Mi"
            cpu: "10m"
          limits:
            memory: "32Mi"
            cpu: "20m"
        # Slow startup - about 20 seconds per pod
        readinessProbe:
          httpGet:
            path: /
            port: 80
          initialDelaySeconds: 12  # Wait 12 seconds before first check
          periodSeconds: 3         # Check every 3 seconds
          timeoutSeconds: 2
          successThreshold: 2      # Require 2 consecutive successes (6 seconds)
          failureThreshold: 3      # Allow 3 failures before giving up
        livenessProbe:
          httpGet:
            path: /
            port: 80
          initialDelaySeconds: 25
          periodSeconds: 10
      # Add init container to slow down startup
      initContainers:
      - name: slow-init
        image: busybox:1.35
        command: ['sh', '-c', 'echo "Starting StatefulSet init container..." && sleep 8 && echo "StatefulSet init complete"']
        resources:
          requests:
            memory: "8Mi"
            cpu: "5m"
---
apiVersion: v1
kind: Service
metadata:
  name: test-statefulset-1min-service
  namespace: default
spec:
  clusterIP: None  # Headless service required for StatefulSet
  selector:
    app: test-statefulset-1min
  ports:
  - port: 80
    targetPort: 80
```

#### Step 4:
Verify we see metric in backend
Screenshot 2025-08-12 at 11 05 25 AM
Screenshot 2025-08-12 at 11 05 10 AM


releasenotes/notes/add-rollout-ongoing-duration-metric-82f1582f84e91562.yaml
```diff
@@ -0,0 +1,11 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+enhancements:
+  - |
+    Adds ongoing_duration guage metric for deployment and statefulset rollouts.
```
cswatt:
```suggestion
    Adds ongoing_duration gauge metric for deployment and statefulset rollouts.
```

cswatt (approved):
spelling error in relnote, but approving to avoid blocking

bitomaxsp (approved):

frank-spano:
@clamoriniere @L3n41c 
 Had a conversion with @sblumenthal and wanted to get your thoughts on some structural things -  or rather - where the business logic of the computation/correlation should occur. For context, in both the ticket and conversations - we decided that we'd use replicaSets and controllerRevisions to fetch the start times of deployments and statefulsets respectively. For digestibility I'll use deployments in my examples, but any implementation details could be extended to statefulsets. 
From my investigation it seemed like there were a few options as far as computing the rollout duration.

1. 
- Define a new metric family generator for deployments, which on deployment events would emit a metric (doesn't need to be a duration, just a flag that the deployment is ongoing) with tags for that deployment. In the transformer functions, which are called each time the check runs, query the apiserver for replicaSet info to determine the creation time.
- As of now there are no calls to the apiserver in any of the transformer functions.
- At first I was unsure if we'd have all the information we'd need to do the computation at the transformer level.

2.
- Define a new metric family generator for deployments, which on deployment events queries the apiserver for replicaSet info and emits the start time of the deployment. The transformer function then handles converting that into a duration.
- Downside of calling the apiserver for every deployment event.

The following two options would be "new" paradigms as far as the check goes

3.
- Aggregate data from two metric family generators / collectors? Not sure if this is possible, but similar to our deployment metric family generator, we could define one for replicaSets, and try to correlate them somehow. Maybe a wrapper generator.

4. 
- Run through all deployments on an interval. Currently, the collectors run all the time and store metrics, which are then collected and forwarded to transformers when the check is run. We could write a function to be called every time the check is run that collects all the deployments + replicaSets and computes the duration of any ongoing rollouts. Basically moving the business logic out of the generators and transformers into somewhere new.
- Doesn't currently match the pattern of the check.

To start with, I went with option 2. It seemed to be a reasonable first approach and I wanted to at least test that the computation was possible using the methods we discussed/ determine any impediments. It is working correctly in a variety of circumstances, but also only tested on a small personal cluster. It does make an apiserver call for every deployment event, so in the case of many deployments advancing at once, it would add a meaningful amount of traffic to the apiserver. 

The good news is we can correctly generate the metric, it's just a matter of deciding the best place to do it. 

Also wanted to bring to you attention that to query controllerRevisions I had to add rbac auth to the cluster agent. Is that acceptable/expected?

clamoriniere:
Hi @frank-spano,  

Thanks for the details.  

### On RBAC
> Also wanted to bring to your attention that to query controllerRevisions I had to add rbac auth to the cluster agent. Is that acceptable/expected?  

Yes, this is expected. You’ll need to create a PR on the public Helm chart and the Datadog Operator to add the RBAC by default.  

For backward compatibility, it’s important that the check does not crash or produce excessive log warnings if the RBAC is missing. This can happen in cases where users don’t deploy via our official solutions.  

### On the proposed solutions
I’m more in favor of **solution 3**: computing and storing Deployment information through a new family generator on top of the ReplicaSet. You could store this data in a local map keyed by the namespace/name of the associated Deployment, then use it when processing the Deployment object.  

I think it’s acceptable to keep the two sources of information not fully in sync, we already follow this approach for the `labelJoin` feature.

frank-spano:
Ready for another review. Went with option 3 - storing the start time based on the replicaset owned by the deployment. Currently the maps are globally accessible in the rollout_tracker, but could be moved into the KSM check if we wanted. Logging is still in the code - just wanted to get it in for a high level review sooner than later. 

I tested the following scenarios on a local minikube. Can provide detailed instructions + yamls for each. 

  Test 1 - Canceled Mid-Rollout:
  - Applied deployment that takes ~45 seconds to complete, then deleted after 10 seconds during active rollout
  - System correctly detected ongoing rollout, then cleaned up via staleness detection after 120 seconds
  - Confirmed deleted deployments stop being accessed and get properly removed from tracking maps

  Test 2 - Never-Completing Deployment:
  - Deployed with failing readiness probe that never becomes ready
  - System continuously tracked rollout duration for stuck deployment
  - Metrics kept being emitted as transformer updates access time, preventing staleness cleanup
  - Deployment remained tracked indefinitely until manually deleted

  Test 3 - Update During Mid-Rollout:
  - Started initial rollout, then applied v2 update after 8 seconds while first rollout ongoing
  - System correctly handled overlapping rollouts with separate duration tracking
  - Generation-based detection properly reset start time for new rollout version

  Test 4 - Bad Image (ImagePullBackOff):
  - Deployed with non-existent image causing ImagePullBackOff
  - System tracked rollout duration despite pods never starting successfully
  - Continued emitting ongoing rollout metrics for failed deployment state

  Test 5 - Rapid Successive Updates:
  - Applied 3 rapid image patches with 0.5-second intervals between updates
  - Successfully detected all generation changes (1→2, 2→3, 3→4) with proper start time resets
  - ReplicaSets created for each generation with accurate Kubernetes timestamps

  Combined Stress Test:
  - Ran all 5 deployment scenarios simultaneously
  - System handled concurrent tracking of multiple deployments without interference
  - Each deployment maintained independent rollout duration metrics
  - Proper cleanup occurred for completed/deleted deployments during concurrent operations

  Large Scale Test (30 Deployments):
  - Deployed 30 concurrent deployments simultaneously
  - System successfully tracked rollout durations for all 30 deployments independently
  - Memory management and cleanup mechanisms handled the increased load effectively
  - All deployments emitted separate metrics without performance degradation

  Timing Accuracy Verification:
  - Compared generation-based detection timestamps vs ReplicaSet creation timestamps
  - Found timing was essentially identical, validating hybrid approach accuracy
  - Confirmed ReplicaSets are created for every generation change, enabling precise rollout timing

pkg/collector/corechecks/cluster/ksm/kubernetes_state.go
```diff
@@ -627,29 +632,37 @@ func (k *KSMCheck) Run() error {
 	// If KSM is running in the node agent, and it's configured to collect only
 	// pods and from the node agent, we don't need to run leader election,
 	// because each node agent is responsible for collecting its own pods.
-	podsFromKubeletInNodeAgent := k.isRunningOnNodeAgent && k.instance.PodCollectionMode == nodeKubeletPodCollection
+	//podsFromKubeletInNodeAgent := k.isRunningOnNodeAgent && k.instance.PodCollectionMode == nodeKubeletPodCollection
```
bitomaxsp: Should this be commented or be removed?

pkg/collector/corechecks/cluster/ksm/kubernetes_state.go
```diff
@@ -537,6 +538,9 @@ func (k *KSMCheck) discoverCustomResources(c *apiserver.APIClient, collectors []
 		customresources.NewExtendedNodeFactory(c),
 		customresources.NewExtendedPodFactory(c),
 		customresources.NewVerticalPodAutoscalerFactory(c),
+		// customresources.NewStatefulSetRolloutFactory(c),
```
bitomaxsp: Is it commented out on purpose or by accident ?
nit: if on purpose, i'd say it should have a comment why it is disabled

pkg/collector/corechecks/cluster/ksm/kubernetes_state.go
frank-spano: Will uncomment, this and the below block was commented out for testing. I was having an issue with leader election unrelated to this PR.

pkg/collector/corechecks/cluster/ksm/kubernetes_state.go
frank-spano: Will add comment, the idea is to implement the statefulset factory once we have approval on the method of tracking deployments.

pkg/collector/corechecks/cluster/ksm/customresources/rollout_tracker.go
```diff
@@ -0,0 +1,372 @@
...
+
+	deploymentMap[key] = dep.DeepCopy()
+	deploymentAccessTime[key] = time.Now() // Track when we stored it
+
+	log.Infof("ROLLOUT-TEST: Deployment stored. Total deployments in map: %d", len(deploymentMap))
+
+	// Run periodic cleanup occasionally (less frequent than transformer calls)
+	go func() {
+		PeriodicCleanup()
+	}()
```
sblumenthal: I am not sure about kicking off the cleanup in this manner... if during the course of the check running, at one point, there are 1000 deployments progressing, this function would have been called 1000 times and would have spun up 1000 goroutines

Which is not a problem, until the check gets rescheduled, or the check runner is restarted, and then we have 1000 threads being spun up all at once when the check starts back up again, all of which are blocked on each other... I wonder if there is a cleaner way to implement this background processes, possibly as a goroutine which runs once, forever, on check start

pkg/collector/corechecks/cluster/ksm/customresources/rollout_tracker.go
```diff
@@ -0,0 +1,372 @@
...
+
+	// Check if this is a new rollout (new deployment OR generation changed)
+	existingDep, exists := deploymentMap[key]
+	if !exists {
+		deploymentStartTime[key] = time.Now()
+		log.Infof("ROLLOUT-TEST: New rollout detected for deployment %s/%s (generation: %d), setting start time to %s",
+			dep.Namespace, dep.Name, dep.Generation, time.Now().Format(time.RFC3339))
+	} else if existingDep.Generation != dep.Generation {
+		oldGeneration := existingDep.Generation
+		deploymentStartTime[key] = time.Now()
```
sblumenthal: Does it make sense to have `deploymentStartTime[key] = time.Now()`? This would theoretically make this metric brittle to the check being rebalanced or the runner restarting... are we instead able to parse the timestamp of the `metadata.managedFields` entry which correlates to `manager: kube-controller-manager` and `subresource: status`? Or possibly something else instead, and maybe only fallback to `time.Now()` as a last resort if there are no better times we could use?

pkg/collector/corechecks/cluster/ksm/kubernetes_state_defaults.go
```diff
@@ -35,6 +35,7 @@ func defaultMetricNamesMapper() map[string]string {
 		"kube_deployment_status_replicas_unavailable":                                              "deployment.replicas_unavailable",
 		"kube_deployment_status_replicas_updated":                                                  "deployment.replicas_updated",
 		"kube_deployment_status_condition":                                                         "deployment.condition",
+		"kube_deployment_ongoing_rollout_duration":                                                 "deployment.ongoing_rollout.duration",
```
sblumenthal: Are the 2 changes to this file still relevant? Seems like we wouldn't want both a name mapper AND a transformer defined for the metric?

pkg/collector/corechecks/cluster/ksm/kubernetes_state_transformers.go
```diff
@@ -573,3 +576,51 @@ func removeSecretTransformer(s sender.Sender, _ string, metric ksmstore.DDMetric
...
+	// Calculate actual rollout duration using stored ReplicaSet data
+	duration := customresources.GetDeploymentRolloutDurationFromMaps(namespace, deploymentName)
+
+	log.Infof("ROLLOUT-TRANS: Sending deployment rollout duration metric for %s/%s: %.2f seconds", namespace, deploymentName, duration)
+
+	s.Gauge(ksmMetricPrefix+"deployment.rollout_duration", duration, hostname, tags)
+}
+
+// statefulsetRolloutDurationTransformer calculates statefulset rollout duration
+func statefulsetRolloutDurationTransformer(s sender.Sender, _ string, metric ksmstore.DDMetric, hostname string, tags []string, _ time.Time) {
```
sblumenthal: if the plan is to implement this in a future PR, we should not leave this here, because it looks like it is properly wired up and so I am guessing this would generate real (not useful at this point) metrics if deployed?

sblumenthal (commented):
The approach itself seems reasonable, but there are a couple of points that I have raised (which I don't think should be a heavy lift?)

I ignored all of the commented code and logging for now under the assumption that it will be removed.

pkg/collector/corechecks/cluster/ksm/customresources/rollout_tracker.go
frank-spano: You are correct, this was supposed to be removed. It's called in the check as well.

pkg/collector/corechecks/cluster/ksm/customresources/rollout_tracker.go
frank-spano: I'll look into those options. This is already considered the fallback, because we collect the true start time from the corresponding replicaSet. When we go to parse the start time, we first look for the replica set data, and then fall back to this if for some reason it isn't available.

pkg/collector/corechecks/cluster/ksm/kubernetes_state_transformers.go
frank-spano: Will remove if I end up not including statefulsets, but this won't ever be called because we never add the metric generator to the check.

pkg/collector/corechecks/cluster/ksm/customresources/rollout_tracker.go
sblumenthal: Got it, in that case then it is probably good enough for now, if it becomes enough of a problem we can revisit

pkg/collector/corechecks/cluster/ksm/kubernetes_state_defaults.go
frank-spano: So would that mean removing the mapper and changing https://github.com/DataDog/datadog-agent/pull/39732/files/b6151665c3971aadae4ea854fee1933f267381bc#diff-c5fcb66e95c9b8a88fb7fb63ae33409c8f5ec1c3ca8c828d873cc6131f2894d7R606 to "deployment.ongoing_rollout.duration" if that's what I wanted the final name to be?

sblumenthal: Yup, that is what that would mean, since the transformer functions directly call the sender to emit metrics

pkg/collector/corechecks/cluster/ksm/customresources/rollout_tracker.go
```diff
@@ -0,0 +1,150 @@
...
+	replicaSetMap        map[string]*ReplicaSetInfo
+	rolloutMutex         sync.RWMutex
+)
+
+func init() {
+	deploymentMap = make(map[string]*appsv1.Deployment)
+	deploymentAccessTime = make(map[string]time.Time)
+	deploymentStartTime = make(map[string]time.Time)
+	replicaSetMap = make(map[string]*ReplicaSetInfo)
+}
```
sblumenthal, line 40: This should probably not be package level vars / not use `init()` like this, as it will only be initialized once and will be shared across all instances of this check on a given node.

In reality, I am not sure that there is a real risk of things getting weird when multiple checks run at once, and I think that the underlying informers should also be shared across checks on a given node, so it is PROBABLY fine, but I just wanted to mention it

pkg/collector/corechecks/cluster/ksm/kubernetes_state.go
```diff
@@ -656,17 +660,35 @@ func (k *KSMCheck) Run() error {
...
-			labelJoiner.insertFamilies(metrics)
+			var metricsStore *ksmstore.MetricsStore
+			if ms, ok := store.(*ksmstore.MetricsStore); ok {
+				metricsStore = ms
+			} else if rolloutStore, ok := store.(*ksmstore.RolloutMetricsStore); ok {
+				metricsStore = rolloutStore.MetricsStore
+			}
+
+			if metricsStore != nil {
+				metrics := metricsStore.Push(k.familyFilter, k.metricFilter)
```
sblumenthal, line 671: nit: not sure if it is "better" or not, but it would be cleaner to define our own interface that contains the function we expect on top of the existing interface, something like (in `pkg/kubestatemetrics/store/store.go`):
```
type Store interface {
	cache.Store
	Push(familyFilter FamilyAllow, metricFilter MetricAllow) map[string][]DDMetricsFam
}
```

Then update the references to `cache.Store` in the KSM check to instead be `store.Store`

pkg/kubestatemetrics/store/rollout_store.go
```diff
@@ -0,0 +1,47 @@
...
+	// Handle rollout cleanup based on resource type
+	switch s.resourceType {
+	case "deployments":
+		if dep, ok := obj.(*appsv1.Deployment); ok {
+			customresources.CleanupDeployment(dep.Namespace, dep.Name)
+		}
+	case "replicasets":
+		if rs, ok := obj.(*appsv1.ReplicaSet); ok {
+			customresources.CleanupDeletedReplicaSet(rs.Namespace, rs.Name)
+		}
```
sblumenthal, line 42: This works, so 🤷 , but it does seem strange to have this package reference the KSM check package (even if it is a subpackage within that KSM package), and could potentially lead to an cyclical import headache later on for someone else

sblumenthal (approved):
The change looks good to me, approving

I left a couple of comments, but nothing blocking... they can be handled in a follow-up PR (or when we implement the statefulset metrics)

pkg/collector/corechecks/cluster/ksm/kubernetes_state.go
sblumenthal: Yea, I don't know that this actually works the way I thought it would / at all

PR #39735, opened 2025-08-11
[NDM] Add core check for NCM

zoedt:
### What does this PR do?
Adds the agent corecheck/integration for NCM / Network Config Management feature owned by NDM

This PR adds:
* A new core check `network_config_management` + defaults around that
* Refine shared logic between the core check and traps/syslogs-based retrieval 
* Adds tests

You can refer to this [documentation](https://datadoghq.atlassian.net/wiki/spaces/II/pages/5367792210/NCM+Architecture+Overview) for details regarding the vision for the agent-based architecture.

**Core check flow**
1. Check is scheduled, configured according to the `conf.yaml` (example below in QA section)
2. For the device, retrieval (default 15m) during the check only grabs the running config (hardcoded for Cisco devices currently)
3. Finishes by submitting to EvP

Many tasks left open, including:
* Additional support for Telnet
* Refining SSH support / configurations
* Validation, parsing, refinement of config output post-retrieval (incl. sensitive data scrubbing, etc.)
* etc.

these will be upcoming PRs to address, as this initial contribution has grown large :^D

**VALIDATION OUTPUT**
```
cisco@qa-agent:~$ sudo -u dd-agent -- datadog-agent check network_config_management
{"namespace":"zoe_ncm_test","integration":"","configs":[{"device_id":"zoe_ncm_test:10.10.1.1","device_ip":"10.10.1.1","config_type":"running","timestamp":1755274594,"tags":["device_ip:10.10.1.1"],"content":"\r\n\r\n\r\nBuilding configuration...\r\n\r\n  \r\nCurrent configuration : 3144 bytes\r\n!\r\n! Last configuration change at 20:53:27 UTC Thu Aug 14 2025\r\n!\r\nversion 15.9\r\nservice timestamps debug datetime msec\r\nservice timestamps log datetime msec\r\nno service password-encryption\r\n!\r\nhostname qa-device\r\n!\r\nboot-start-marker\r\nboot-end-marker\r\n!\r\n!\r\n!\r\nno aaa new-model\r\n!\r\n!\r\n!\r\nmmi polling-interval 60\r\nno mmi auto-configure\r\nno mmi pvc\r\nmmi snmp-timeout 180\r\n!\r\n!\r\n!\r\n!\r\n!\r\n!\r\n!\r\n!\r\n!\r\n!\r\n!\r\nip ...EDITED FOR THE SAKE OF CONCISENESS"}],"collect_timestamp":1755274594}


  Running Checks
  ==============

    network_config_management
    -------------------------
      Instance ID: network_config_management:zoe_ncm_test:51fd8431b479a10a [OK]
      Configuration Source: file:/etc/datadog-agent/conf.d/network_config_management.d/conf.yaml
      Total Runs: 1
      Metric Samples: Last Run: 0, Total: 0
      Events: Last Run: 0, Total: 0
      ndmconfig: Last Run: 1, Total: 1
      Service Checks: Last Run: 0, Total: 0
      Average Execution Time : 1.089s
      Last Execution Date : 2025-08-15 16:16:34 UTC (1755274594000)
      Last Successful Execution Date : 2025-08-15 16:16:34 UTC (1755274594000)


  Metadata
  ========
    config.hash: network_config_management:zoe_ncm_test:51fd8431b479a10a
    config.provider: file
    config.source: /etc/datadog-agent/conf.d/network_config_management.d/conf.yaml
```

QA steps/how to validate
* Pull in the `ndm-tools/cml-qa` repo and follow instructions for setting up the CLI ([link](https://github.com/DataDog/ndm-tools/tree/main/cml-qa))
```
python cml_template_generator.py \
  --deb-url <INSERT THE DEB FROM CI/CD BUILD> \
  --api-key <INSERT YOUR KEY> \
  --site "datad0g.com" \
  --title "<YOUR NAME/TITLE> NCM Agent QA"
```
* Go the NDM hosted CML ([docs](https://datadoghq.atlassian.net/wiki/spaces/II/pages/5061640281/Getting+Started+with+Cisco+Modeling+Labs+CML))
* Press Import
* Pull in the YAML that gets created, start the lab
* <to fill in more details to prep the IOS device for SSH, etc.>
image

`/etc/datadog-agent/conf.d/network_config_management.d/conf.yaml`
```
init_config:
instances:
- ip_address: "10.10.1.1"
  namespace: "zoe_ncm_test"
  auth:
    username: "cisco"
    password: "cisco"
    ssh_ciphers: [aes256-ctr, aes192-ctr, aes128-ctr]
    ssh_key_exchanges: [diffie-hellman-group14-sha1, diffie-hellman-group-exchange-sha1]
    ssh_host_key_algorithms: [ssh-rsa]
```

`iosv-0/0` steps to enable SSH on network device
```
qa-device#enable
qa-device#conf t
Enter configuration commands, one per line.  End with CNTL/Z.
qa-device(config)#username cisco privilege 15 secret cisco
qa-device(config)#line vty 0 4
qa-device(config-line)#login local
qa-device(config-line)#transport input ssh
qa-device(config-line)#exit
qa-device(config)#ip domain-name lab.local
qa-device(config)#crypto key generate rsa modulus 2048
The name for the keys will be: qa-device.lab.local

% The key modulus size is 2048 bits
% Generating 2048 bit RSA keys, keys will be non-exportable...
[OK] (elapsed time was 3 seconds)

qa-device(config)#ip ssh version 2
qa-device(config)#end
qa-device#write memory
```


louis-cqrl (commented):
So nice to see a great PR with tests, I also see a bunch of `todo` in the code, can you create card in your backlog to ensure we do not loose track of them ?

zoedt:
> So nice to see a great PR with tests, I also see a bunch of `todo` in the code, can you create card in your backlog to ensure we do not loose track of them ?

@louis-cqrl thank you! as you requested, i set up official jira tickets for the todos throughout the code:
* https://datadoghq.atlassian.net/browse/NDMII-3592
* https://datadoghq.atlassian.net/browse/NDMII-3593
* https://datadoghq.atlassian.net/browse/NDMII-3594
* https://datadoghq.atlassian.net/browse/NDMII-3595
* https://datadoghq.atlassian.net/browse/NDMII-3596

these include some context along with definitions of done

after your review, i also added some commits to address the failing static quality checks (removing the new core check from iot agent builds to not increase the build size)

if you have any further questions or comments, they would be super appreciated !

louis-cqrl (approved):

pkg/collector/corechecks/networkconfigmanagement/networkconfigmanagement.go
```diff
@@ -0,0 +1,176 @@
...
+// Copyright 2025-present Datadog, Inc.
+
+//go:build ncm
+
+// Package networkconfigmanagement defines the agent core check for retrieving network device configurations
+package networkconfigmanagement
+
+import (
+	"fmt"
+	pkgconfigsetup "github.com/DataDog/datadog-agent/pkg/config/setup"
```
FlorianVeaux: nit: might want to re-sort your imports. Not sure why it wasn't picked up automatically though

pkg/collector/corechecks/networkconfigmanagement/networkconfigmanagement.go
```diff
@@ -0,0 +1,176 @@
...
+	c.deviceConfig = &deviceConfig
+
+	err = yaml.Unmarshal(rawInitConfig, &initConfig)
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal init config: %s", err)
+	}
+	c.initConfig = &initConfig
+
+	var namespace string
+	if c.deviceConfig.Namespace != "" {
```
FlorianVeaux: Do we want to allow per-device override of the namespace? I think we shouldn't diverge from the idea that each Agent has a single namespace.
If it's only for internal purpose maybe renaming in the yaml/struct to something that's clear is internal/testing only

pkg/networkconfigmanagement/config/config.go
```diff
@@ -0,0 +1,161 @@
...
+// NcmConfig is the processed config structure for Network Config Management (NCM) to be used by the component
+type NcmConfig struct {
+	Namespace string
+	Devices   map[string]DeviceConfig // map of device IP addresses to DeviceConfig
+}
+
+// InitConfig holds the initial configuration for the NCM component, including the namespace and check interval.
+type InitConfig struct {
+	Namespace     string `yaml:"namespace"`      // Namespace for the NCM devices where configs are retrieved from, to help match a device on DD
+	CheckInterval int    `yaml:"check_interval"` // Interval in seconds to check for config changes
```
FlorianVeaux: There is already `min_collection_interval` that is standard in the Agent for that purpose.
It's automatically working for python check, I don't remember for go check but I would suggest to use `min_collection_interval` as the name for the yaml entry (and understand how and if the Agent parses it automatically)

pkg/collector/corechecks/networkconfigmanagement/networkconfigmanagement.go
```diff
@@ -0,0 +1,176 @@
...
+	if err != nil {
+		return err
+	}
+	c.deviceConfig = &deviceConfig
+
+	err = yaml.Unmarshal(rawInitConfig, &initConfig)
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal init config: %s", err)
+	}
+	c.initConfig = &initConfig
```
FlorianVeaux: Do we want to do any kind of validation in initConfig too? For instance the min_check_interval isn't a negative number, or is lower than X (to avoid users configuring a check that runs too often)

pkg/collector/corechecks/networkconfigmanagement/networkconfigmanagement.go
```diff
@@ -0,0 +1,176 @@
...
+	core.CheckBase
+	deviceConfig *ncmconfig.DeviceConfig
+	initConfig   *ncmconfig.InitConfig
+	sender       *ncmsender.NCMSender
+	agentConfig  config.Component
+	remoteClient ncmremote.Client
+}
+
+// TimeNow useful for mocking
+var TimeNow = time.Now
```
FlorianVeaux: FYI: The Agent repo already leverages `github.com/benbjohnson/clock` which is a tiny time.Now() util useful for testing. It's fine like you're doing here but figured I would call this out if you're interested

pkg/networkconfigmanagement/config/config.go
```diff
@@ -0,0 +1,161 @@
...
+
+// AuthCredentials holds the authentication credentials to connect to a network device.
+type AuthCredentials struct { // auth_credentials
+	Username string `yaml:"username"`
+	Password string `yaml:"password"`
+	Port     string `yaml:"port"`
+	Protocol string `yaml:"remote"`
+	// SSH-specific configurations
+	// TODO: Move to separate struct if needed (e.g., SSHConfig)
+	SSHCiphers      []string `yaml:"ssh_ciphers,omitempty"`
```
FlorianVeaux: For here and following lines in both struct, why do we need `omitempty`?  It doesn't looke like we're marshalling this back into yaml

pkg/collector/corechecks/networkconfigmanagement/networkconfigmanagement.go
```diff
@@ -0,0 +1,176 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-present Datadog, Inc.
+
+//go:build ncm
```
FlorianVeaux, line 6: Why having a build tag for ncm ?

pkg/networkconfigmanagement/config/config.go
```diff
@@ -0,0 +1,161 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-present Datadog, Inc.
+
+//go:build ncm
+
+// Package config provides configuration structures and functions for the Network Config Management (NCM) core check and component
+package config
```
FlorianVeaux, line 9: I'm wondering what's the reason to put this in a toplevel package `pkg/networkconfigmanagement` rather than directly in the corecheck definition. All of these subpackages look really tied to the corecheck right? Do we anticipate to use then somewhere else?

pkg/networkconfigmanagement/config/config.go
FlorianVeaux: Do we need these three configs? What's the use cases for a user to update that?

pkg/networkconfigmanagement/config/config.go
FlorianVeaux: Ah right trigger-based pulls, makes sense

pkg/networkconfigmanagement/config/config.go
```diff
@@ -0,0 +1,161 @@
...
+	//SshKeyPassphrase string `yaml:"sshKeyPassphrase"` // passphrase for SSH key if needed
+	//Enable           bool   `yaml:"enable"`           // if true, will use enablePassword to enter privileged exec mode
+	//EnablePassword   string `yaml:"enable_password"`  // to be able to use privileged exec mode
+}
+
+// DeviceConfig holds the info to connect to a network device, including its IP address and authentication credentials.
+type DeviceConfig struct {
+	Namespace            string          `yaml:"namespace"`
+	IPAddress            string          `yaml:"ip_address"`                       // ip address of the network device, e.g., "10.0.0.1"
+	CollectStartupConfig bool            `yaml:"collect_startup_config,omitempty"` // whether to collect the startup config (default: false)
```
FlorianVeaux: Any reason to disable this?

pkg/networkconfigmanagement/config/config.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	cr := integration.ConfigCheckResponse{}
+	err = json.Unmarshal(res, &cr)
+	if err != nil {
+		return nil, err
+	}
+	// Iterate through the instances (devices) and parse the device configs
+	var deviceConfigs []DeviceConfig
+	for _, c := range cr.Configs {
+		if c.Config.Name == "network_config_management" { // Check name for NCM
```
FlorianVeaux:
```suggestion
		if c.Config.Name == networkconfigmanager.CheckName { // Check name for NCM
```
?

pkg/networkconfigmanagement/config/config.go
```diff
@@ -0,0 +1,161 @@
...
+
+// InitConfig holds the initial configuration for the NCM component, including the namespace and check interval.
+type InitConfig struct {
+	Namespace     string `yaml:"namespace"`      // Namespace for the NCM devices where configs are retrieved from, to help match a device on DD
+	CheckInterval int    `yaml:"check_interval"` // Interval in seconds to check for config changes
+}
+
+// GetNCMConfigsFromAgent retrieves the NCM configurations from the agent's config check endpoint (from core check)
+// TODO: tests for this to come when the component is refactored / we're working on the trigger-based approach for config changes
+func GetNCMConfigsFromAgent(client ipc.HTTPClient) (*NcmConfig, error) {
```
FlorianVeaux: General topic (without an easy answer), but I find the naming in the code confusing between configs (user-edited yaml files for the agent/integrations), and device configs

pkg/networkconfigmanagement/config/config.go
```diff
@@ -0,0 +1,161 @@
...
+			log.Warnf("Duplicate device IP address found in config: %s, skipping duplicate", d.IPAddress)
+			continue
+		}
+		if d.IPAddress != "" {
+			deviceMap[d.IPAddress] = d
+		} else {
+			log.Warnf("Device config missing IP address, skipping: %+v", d)
+		}
+	}
+	// TODO: deal with namespace + defaults?
```
FlorianVeaux: Should we do that now? I'm concerned we won't prioritize this very soon, we'll forget, and this will lead to issues in the future?

pkg/networkconfigmanagement/remote/test_utils.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	return results[0], nil
+}
+
+// retrieveConfiguration retrieves the configuration using the provided commands.
+func (t *MockSSHClient) retrieveConfiguration(commands []string) ([]string, error) {
+	var results []string
+
+	for _, cmd := range commands {
+		session, err := t.NewSession()
```
FlorianVeaux, line 75: This means we'll have two sessions, one for running one for statup config. Does that create any over head, and should we aim to keep a single longer-running session ?

pkg/networkconfigmanagement/report/payload.go
```diff
@@ -0,0 +1,61 @@
...
+	// STARTUP represents the startup configuration of a network device (the configuration that is loaded on boot)
+	STARTUP ConfigType = "startup"
+)
+
+// NCMPayload contains network devices configuration payload sent to EvP / backend
+type NCMPayload struct {
+	Namespace        string                   `json:"namespace"`
+	Integration      integrations.Integration `json:"integration"`
+	Configs          []NetworkDeviceConfig    `json:"configs"`
+	CollectTimestamp int64                    `json:"collect_timestamp"`
```
FlorianVeaux, line 34: What's the purpose of having one "collect_timestamp" and one timestamp per device?

FlorianVeaux (commented):
Great job with this PR 🎉 . It's very easy to review and the decisions, structure, abstractions are well-thought

I've left a few questions and possible improvements here and there but nothing major really

pkg/networkconfigmanagement/config/config.go
```diff
@@ -0,0 +1,161 @@
...
+type InitConfig struct {
+	Namespace     string `yaml:"namespace"`      // Namespace for the NCM devices where configs are retrieved from, to help match a device on DD
+	CheckInterval int    `yaml:"check_interval"` // Interval in seconds to check for config changes
+}
+
+// GetNCMConfigsFromAgent retrieves the NCM configurations from the agent's config check endpoint (from core check)
+// TODO: tests for this to come when the component is refactored / we're working on the trigger-based approach for config changes
+func GetNCMConfigsFromAgent(client ipc.HTTPClient) (*NcmConfig, error) {
+	// Call the agent's config check endpoint to retrieve the NCM configs (from core check)
+	endpoint, err := client.NewIPCEndpoint("/agent/config-check")
```
TCheruy, line 115: is this the only way to retrieve the check config? Could it be accessed through the config component?

zoedt: i don't think this is the only way - i used the device scan component as a reference !

i shouldn't have put this function in in this pr (since this is specific for the syslog/traps processing version of NCM) but i got on a roll 😳 

i'll play around with that - i think probably the config component should be enough but i meant to check in w/ whoever added the code why they opted for the IPC client !

as for dealing with it in the PR -- i can remove it for now or we can revisit it and/or i can add to the syslog/traps tickets that this needs to be addressed, what do you think would be good? i'm fine with keeping it since it's not currently in use, but lmk your thoughts !

TCheruy: i'm fine with keeping it! was just curious as it feels like we're making an http call to ourselves ahah, i'll check the device scan implementation it might contains clues

pkg/collector/corechecks/networkconfigmanagement/networkconfigmanagement.go
zoedt: oop i forgot to incl. details about that on the PR itself - but i asked this question, previously this PR was failing on static quality checks:
Slack thread: https://dd.slack.com/archives/G01DX2UE2HL/p1755523434703619

[Confluence doc I wrote up with context and the solution (using the build tag)](https://datadoghq.atlassian.net/wiki/spaces/II/pages/5423399346/Static+Quality+Check+Failures)

pkg/networkconfigmanagement/config/config.go
zoedt: yes! it's pre-emptive but should help for the future upon that implementation!

i was following what other NDM checks/features seem to put re-used logic (inside `/pkg`) but lmk if there's something else, seems like a lot of things change fast on the agent !!

pkg/networkconfigmanagement/config/config.go
zoedt: oop good call out ! updating this :-D

CelianR (approved):

pkg/networkconfigmanagement/config/config.go
zoedt: so when i was testing on CML for qa - i was running into issues upon connecting such as:
```
Unable to negotiate with 10.10.1.1 port 22: no matching key exchange method found. Their offer: diffie-hellman-group-exchange-sha1,diffie-hellman-group14-sha1
```

i'm seeing that the client was rejecting the client b/c the key exchange/ciphers/algo methods the router (the IOS device) was old/insecure to whatever the default method was - i thought if customers are willing to take something less secure than the default to proceed, then it would be fine 

wdyt? i'm not sure how common this is but i guess some customers may/may not have super up to date devices all the time so it may be ok to keep supporting?

FlorianVeaux: Might be important to have then indeed.
Is this list of Cipher names standardized somewhere or is it specific to the go ssh library? If the latter we should probably wrap that through some boolean configurations to make things easier for the user

pkg/networkconfigmanagement/config/config.go
zoedt: i was thinking in the case of a vendor where there isn't a `startup` config, but effectively only one `active` / `running` config that this wouldn't be a useful command to keep in mind

(example from[ this doc](https://datadoghq.atlassian.net/wiki/spaces/II/pages/5310218888/Reference+Types+of+Configs+Use+cases+flows) to clarify visually)

pkg/networkconfigmanagement/config/config.go
FlorianVeaux: So technically we could make this whole thing a component instead and just important the component where we need. And if so it could live inside `/comp/` or `/comp/ndm/` for instance?

pkg/networkconfigmanagement/remote/test_utils.go
zoedt: yes this was a journey to understand 😅 i also wanted to keep a single longer-running session, there were a few issues:

* `crypto/ssh` 's `SSHClient` has the equivalence of only 1 exec command per `Session`
  * this was difficult to find official documentation for this, but here's [a reddit thread](https://www.reddit.com/r/webdev/comments/1gv5fpv/is_it_golang_thing_to_allow_one_command_via_ssh/) that confirmed it for me after struggling when QAing at some points
  * there is the `Shell` that opens up an interactive session, but it involves doing additional command formatting and parsing (as noted by the redditors) and i thought it'd be another headache to deal with the PTY 😓 
  * instead of deferring, i close immediately after the command is run for the session, which should free up resources vs. using the `defer` which i think would wait until the full loop of commands is done before cleaning things up 
* in addition to this during QA, i found out some devices have an unmodifiable attribute of `MaxSessions` of 1
  * this is confusing, but the `Session` in this case is equivalent to a `Client` - for some reason the CML cisco device had this setting so, i had to introduce the `redial` functionality which would have to replace the client between commands additionally (i only redial if the issue suggestive of a `MaxSession` of 1 occurs
  * however, i do `Close()` the `SSHClient` before redialing/creating a new one, so there shouldn't be any issue with zombie clients running amok

pkg/networkconfigmanagement/report/payload.go
zoedt: i think we wanted to capture the timestamp where the config was directly saved/updated vs. when the agent collected the config itself

pkg/networkconfigmanagement/config/config.go
zoedt: i agree 😢 this was hard to come to a conclusion for names with the word `Config` becoming incredibly overloaded - i'll try to play around with a better way to differentiate configurations for the feature (branching into the yaml files and the device configurations) and even the configs from the actual devices 🤔

jszwedko (approved):

pkg/networkconfigmanagement/config/config.go
zoedt: would introduce a circular import - going to add a new var for the checkname here

pkg/networkconfigmanagement/config/config.go
zoedt: got it - i'll rename, i thought that the `Interval()` (see [here](https://github.com/DataDog/datadog-agent/pull/39735/files#diff-b1cb3ef088bc55b2d09d24b7668cf8b6aeaf3103415e40ceae0ee87220d62979R155)) functionality of the check helps deal w/ that but not certain if using this variable overrides it somewhere 🤔

pkg/collector/corechecks/networkconfigmanagement/networkconfigmanagement.go
zoedt: i can have a separate function, in the `Interval()` function in `networkconfigmanagement.go` i do a quick check for non-negative numbers but good call out to have nothing lower than X / if it's worth having as a separate fn, i can separate the logic outside of the `Interval()` method as well

pkg/networkconfigmanagement/config/config.go
zoedt: my understanding is that even if i'm not marshalling - using `omitempty` will just not surface this at all (typically, it will fill with a default value, in this case string is `""`, with `omitempty` then it would just be nil if the field was trying to be accessed

i think i just was utilizing it when i was looking thru struct options, i don't think this is a huge requirement, so i'm happy to remove it - to be more specific, i don't think that we need the definition between `unset` value vs. a truly intentional empty string

soberpeach (approved):

pkg/networkconfigmanagement/config/config.go
FlorianVeaux: Yeah I'm not 100% certain without looking at the whole code but IIRC, but anyway `min_collection_interval` is the standard naming across integrations so it's good to keep using it anyway

pkg/collector/corechecks/networkconfigmanagement/networkconfigmanagement.go
FlorianVeaux: I think it would be good to fail if anything in the init config is really invalid. Whether this happens in a validate function or in Interval is up to you 👍

pkg/networkconfigmanagement/config/config.go
FlorianVeaux: Yeah that's not how it works see https://go.dev/play/p/wpwwU3XsB2C
If you really want this field to be nil instead of an array, you would need to either create custom unmarshalling, use a pointer to slice (not sure how idiomatic this is), or create a custom type wrapping []string and having a pointer to it

FlorianVeaux: But in any case I would advocate for removing the `omitempty` because it implies a behavior that's not performed (marshalling)

pkg/collector/corechecks/networkconfigmanagement/networkconfigmanagement.go
FlorianVeaux: Makes sense thanks!

pkg/networkconfigmanagement/config/config.go
FlorianVeaux: Should that be something we detect automatically instead? Or try to collect both and if the startup config fails do not send it.
It seems weird request customers to enable this themselves

(Also here's another place where we should drop the omitempty)

pkg/networkconfigmanagement/report/payload.go
FlorianVeaux: Right, but for now we set all of them to time.Now() anyway, we do not parse or know when the config was saved by a user right?

pkg/networkconfigmanagement/remote/test_utils.go
FlorianVeaux: Makes sense thanks for the details !

pkg/networkconfigmanagement/report/payload.go
FlorianVeaux: Makes sense, but for now both timestamps are set to time.Now at time of collection right? So it's not the behavior you describe

pkg/networkconfigmanagement/config/config.go
FlorianVeaux: The reason we have this in the device scan component is because the code can be called from the CLI which is a new process. To retrieve the actual running config we need to contact the Agent in http.
For NCM you're always running inside the main agent process, so you should be able to use the config component instead of using IPC inside the same process

pkg/networkconfigmanagement/config/config.go
zoedt: yes, i was originally going for this with my original PR but after chatting w/ someone from the agent re: the config they mentioned usually `pkg` directory would be a fine place to store shared logic between two features in different parts of the agent

i can follow up with getting the official stance about how to organize such a thing in a case like this from other folks from the agent !

pkg/networkconfigmanagement/config/config.go
zoedt: yee - with the introduction of device profiles that could certainly be part of the profile / part of the detection

i was running into issues with running more than one CLI command before, so this was part of debugging as part of it - i can definitely clean it up such that it's not visible to a user

pkg/networkconfigmanagement/report/payload.go
zoedt: yes, i didn't add a todo but i was planning on taking some time to see how possible it is to parse it out of the configuration file itself - i think iirc cisco does but i haven't done the digging for the other vendors that we're looking to support

pkg/networkconfigmanagement/config/config.go
zoedt: ooo nice ty for the context florian that makes sense! not sure if we wanna make an agent CLI version of NCM for customers to pull configs on demand? 🤔 either way, can update such that we use the config component instead to begin with

pkg/networkconfigmanagement/config/config.go
zoedt: this is for a function that will be used only by the component (traps/syslog processing) so i got a little ahead of myself here, but i'll try to clean it up - it'll likely be done with some refactoring of the several config structs as previously talked about

pkg/networkconfigmanagement/config/config.go
zoedt: hmm, it does seem easier to get a list of booleans, it may just be really long - this is the [list](https://github.com/golang/crypto/blob/master/ssh/common.go#L113) of supported algorithms, key exchanges and ciphers

i may deal with this in another PR since it might get even longer if this sounds reasonable?

zoedt:
**all feedback regarding the below should be addressed with the above commits:**
* utilizing the `benbjohnson` clock
* removal of the `CollectStartupConfig`
* removal of `omitempty` for yaml representations for golang structs
* refactoring of "config" structs to be more representative of yaml vs. internal information for the feature
  * additionally, validation of init config
  * dealing with namespace + only having one per agent instance
  * rename to standard `min_collection_interval` and usage with `Interval()`
* timestamp to retrieve (hardcoded cisco ios) from config string

**leftover / follow-ups:**
* agent standards on where shared logic can live:
  * slack thread [here](https://dd.slack.com/archives/C06PBHLD4DQ/p1756301275642759)
  * given it seems like stateless logic being shared - we can keep it in `pkg` for now but can revisit a refactor/move which should be relatively straightforward if we wish to keep in `comp` when we do the syslog/traps processing
* ipc endpoint for configurations
  * realizing the `config.Component` gives easy access to `datadog.yaml` config, but for integration-specific, pending on a follow-up here on [slack](https://dd.slack.com/archives/C9WBCH0AF/p1756303290185349)
* defaults for SSH ciphers, algorithms, key exchanges, etc.
  * thinking to leave as a todo to clean up in a future PR, found the supported / referenced variables [here](https://github.com/golang/crypto/blob/master/ssh/common.go#L113)

pkg/networkconfigmanagement/config/config.go
FlorianVeaux: Yup sounds good, and thanks for all the details !

FlorianVeaux (approved):

PR #39736, opened 2025-08-11
Reapply: Global Datadog() can read/write config, not build it

dustmop:
### What does this PR do?

Reapplies https://github.com/DataDog/datadog-agent/pull/39395, which broke the static quality gates due to binary size increase. The fix is the [final commit](https://github.com/DataDog/datadog-agent/commit/5c6e1740d5ec89bf9dbac2d04283165b148ff266), see https://datadoghq.atlassian.net/wiki/spaces/ACFG/pages/5386798973/Datadog+global+accessor+PR+size+increase for details on what happened.

### Motivation

To more cleanly separate the building and usage of the config, remove all `model.Setup` methods from the interface that is returned by the global `Datadog()` accessor.

Introduce the `model.BuildableConfig` interface, which is able to both build and read/write the config. It is only meant to be used during initial construction.

Finally, introduce `config.RevertFinishedBackToBuilder()` which is an escape hatch for OTel, since they build on top of the already built config.

Unit tests already exist to cover these changes


duncanista (approved):

pkg/config/nodetreemodel/config.go
```diff
@@ -213,6 +213,14 @@ func (c *ntmConfig) SetTestOnlyDynamicSchema(allow bool) {
 	c.allowDynamicSchema.Store(allow)
 }
 
+// RevertFinishedBackToBuilder returns an interface that can build more on the current
+// config, instead of treating it as sealed
+// NOTE: Only used by OTel, no new uses please!
```
truthbk, line 218: Should we try to lint this to prevent new uses?

truthbk (approved):
Good from the OTel side 👍 

Just wondering if we should consider adding some klnd of lint check to prevent new uses of `RevertFinishedBackToBuilder` except for approved ones like OTel.

pimlu (approved):

davidor (approved):
:+1: for container-platform files (only 1)

lrbhatia (approved):

pkg/config/nodetreemodel/config.go
dustmop: Yes, I think that's a good idea. I'll file a card for it on our team's board.

charan-manjunath (approved):

rahulkaukuntla (approved):

paulcacheux (approved):

val06 (approved):

PR #39764, opened 2025-08-12
Fix python for awscli on macos

alopezz:
### What does this PR do?

It adds a shim to have aws be run with the python where it is installed.

### Motivation

#incident-41849

Originally trying to test something around https://github.com/DataDog/ci-platform-machine-images/pull/393, but that failed and I couldn't get it to work.

### Additional Notes

This is not a permanent fix, and should be immediately followed up with a more sensible approach to managing python-based utilities.

### DIFF
--- a/.gitlab/common/macos.yml
+++ b/.gitlab/common/macos.yml
@@ -66,18 +66,13 @@
   - rm -f dda.tar.gz
   - export PATH="$DDA_DIR:$PATH"
   - dda self dep sync -f legacy-tasks
-  # Install awscli
-  - export AWSCLI_DIR="$TMPDIR/awscli-${CI_JOB_ID}"
+  # Add a shim to make aws work
   - |
-    echo 'Installing AWS CLI...'
-    AWSCLI_VERSION=1.29.45
-    awscli_d="$(mktemp -d)"
-    pushd "$awscli_d"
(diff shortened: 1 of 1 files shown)


.gitlab/common/macos.yml
```diff
@@ -66,6 +66,19 @@
...
+  # Install awscli
+  - export AWSCLI_DIR="$TMPDIR/awscli-${CI_JOB_ID}"
+  - |
+    echo 'Installing AWS CLI...'
+    AWSCLI_VERSION=1.29.45
+    awscli_d="$(mktemp -d)"
+    pushd "$awscli_d"
+    curl -fsSL "https://s3.amazonaws.com/aws-cli/awscli-bundle-$AWSCLI_VERSION.zip" --output bundle.zip
+    unzip bundle.zip
+    sudo /usr/local/bin/python"$PYTHON_VERSION" awscli-bundle/install -i $AWSCLI_DIR/installation -b $AWSCLI_DIR/aws
```
rdesgroppes: I was wrong in my initial draft (now [fixed](https://github.com/DataDog/ci-platform-machine-images/pull/393/files#diff-4aba17db6f4285f53c180fa906edc6a71758c8ec3423dd703a4f63f28e05cf3cR113)): the PYTHON_VERSION here denotes the `venv` one, whereas we should use the `system` one:
```suggestion
    awscli-bundle/install -i $AWSCLI_DIR/installation -b $AWSCLI_DIR/aws
```
(also without `sudo` since your PR doesn't mutate a system path)

PR #39773, opened 2025-08-12
[SQG] Measure package size in build job

JSGette:
### What does this PR do?
_This is **PART I**, the implementation is not complete yet and not used productively. However to not spawn another PR containing dozens of changes it was decided to bring this work to main step by step._

This is a further development of infamous Static Quality Gates. This PR adds `after_script` to `packaging` jobs that triggers in place size measurement of the package's on wire and on disk size. In comparison to current implementation we do not wait until all artifacts are present and work with packages right in build/packaging job instead. As an output we produce a yaml report that contains the following information:
- Gate name
- Artifact path
- Current on wire size
- Current on disk size
- Max on wire size
- Max on disk size
- Timestamp
- Commit SHA
- Pipeline ID
- File inventory that has a list of all files within the package, their size and checksum

**Sequential PR will change the behavior of `static_quality_gates` job to take those reports (each packaging job produces one report) and compare current vs max size and file inventory of a PR against an ancestor.**

### Motivation
We want to provide as much information to the developers as possible and also ensure good performance. It is painful right now and mostly reactive work as developers start to respect Static Quality Gates only after they become red in `main`. This PR is a first step to improve transparency and visibility of the gates to simplify the process of size optimizations on developers' side.

It is now set as `after_script` so even if it fails it should have no impact on current pipeline operations, although `packaging` jobs will take in average 10 seconds longer.

Experimental code is also separated from the production one in `experimental_gates.py` and additional unit tests are added.

Configuration of gates is somewhat tricky. Now we need to add `GATE_NAME` variable to the job where we are supposed to measure the package, `!reference [.report_path]` should be added to `artifacts` section of the job. And the `GATE_NAME` should correspond to an entry in `test/static/static_quality_gates.yml`, otherwise measurement will fail unable to pull configuration (max values). 

I will think of better way later on.


ofek (approved):

.gitlab/packaging/deb.yml
```diff
@@ -131,22 +140,29 @@ installer_deb-arm64:
...
+    BASE_PACKAGE_DIR: $OMNIBUS_PACKAGE_DIR
+    DD_PROJECT: iot-agent
   cache:
     - !reference [.cache_omnibus_ruby_deps, cache]
 
 iot_agent_deb-x64:
   extends: [.package_iot_deb_common, .package_deb_x86]
   needs: ["iot-agent-x64"]
+  variables:
+    GATE_NAME: "static_quality_gate_iot_agent_deb_amd64"
```
JSGette: Rename to `STATIC_QUALITY_GATE_NAME`

.gitlab/packaging/deb.yml
```diff
@@ -10,12 +14,14 @@
...
     paths:
       - $OMNIBUS_PACKAGE_DIR
+      - !reference [.report_path]
   variables:
     OMNIBUS_PACKAGE_ARTIFACT_DIR: $OMNIBUS_PACKAGE_DIR
     KUBERNETES_CPU_REQUEST: 16
     KUBERNETES_MEMORY_REQUEST: "32Gi"
     KUBERNETES_MEMORY_LIMIT: "32Gi"
     PACKAGE_REQUIRED_FILES_LIST: "test/required_files/agent-deb.txt"
+    BASE_PACKAGE_DIR: $OMNIBUS_PACKAGE_DIR
```
JSGette: Rename to `STATIC_QUALITY_GATE_BASE_PACKAGE_DIR`

.gitlab/packaging/common.yml
```diff
@@ -0,0 +1,86 @@
...
+              "${BUCKET_BASE_PATH}/${REPORT_PREFIX}_size_report_${CI_PIPELINE_ID}_${CI_COMMIT_SHA:0:8}.yml"
+          else
+            echo "⚠️  No package found matching pattern: $PACKAGE_PATTERN"
+          fi
+        done
+      else
+        echo "ℹ️  Skipping package measurement (no GATE_NAME defined)"
+      fi
+
+.report_path: "**/*_size_report_*.yml"
```
JSGette: Rename to `static_quality_gate_report_path`

.gitlab/packaging/deb.yml
```diff
@@ -131,22 +140,29 @@ installer_deb-arm64:
...
       - $OMNIBUS_PACKAGE_DIR
+      - !reference [.report_path]
   variables:
     KUBERNETES_CPU_REQUEST: 16
     KUBERNETES_MEMORY_REQUEST: "32Gi"
     KUBERNETES_MEMORY_LIMIT: "32Gi"
     OMNIBUS_PACKAGE_ARTIFACT_DIR: $OMNIBUS_PACKAGE_DIR
     PACKAGE_REQUIRED_FILES_LIST: "test/required_files/iot-agent-deb.txt"
+    BASE_PACKAGE_DIR: $OMNIBUS_PACKAGE_DIR
+    DD_PROJECT: iot-agent
```
JSGette, line 151: Add `STATIC_QUALITY_GATE_ARCH`

.gitlab/packaging/common.yml
```diff
@@ -0,0 +1,86 @@
+---
```
JSGette, line 1: Use `!reference` in `after_script` instead of extending template

rdesgroppes: Thanks!

.gitlab/packaging/deb.yml
rdesgroppes: Or strip, as discussed offline

JSGette: Forgot to remove

PR #39774, opened 2025-08-12
USMON-1458: USM DirectConsumer

DanielLavie:
### What does this PR do?
This PR augment USM Kernel event handling by implementing a new `DirectConsumer` that uses the pkg/ebpf/perf.EventHandler instead of the legacy USM custom event batching mechanism. Key changes include:

  - New DirectConsumer[V] implementation: Processes individual events directly from eBPF programs using perf/ring buffers
  - KernelAdaptiveConsumer[V] wrapper: Automatically selects between DirectConsumer and BatchConsumer based on kernel version and configuration
  - HTTP protocol migration: Updated to use the new consumer interface with kernel-adaptive selection
  - New configuration options: Added direct_consumer_* configuration parameters for buffer sizing and performance tuning
  - Consumer interface refactoring: Renamed legacy Consumer to BatchConsumer and introduced unified Consumer[V] interface
  - eBPF constant integration: Added use_direct_consumer constant to control eBPF behavior based on selected consumer type

### Motivation
USM currently uses a custom event handling abstraction that manually constructs perf and ring buffers, delivers events via additional Go channels, and implements a custom map-page batching scheme. This approach has several limitations:

  1. Complex custom batching logic with known race conditions between socket-filter and uprobe flush paths
  2. Code complexity requiring custom batch maps, flush kprobes, offsetManager, and map-walking logic
  3. Limited telemetry requiring manual implementation for lost events, channel lengths, and buffer statistics

  The new pkg/ebpf/perf.EventHandler addresses these issues by providing:

  - Watermark-based batching enabling direct `bpf_perf_event_output`/`bpf_ringbuf_output` from socket filters and uprobes
  - Simplified code removing the need for custom batching infrastructure on modern kernels
  - Built-in telemetry with lost-event counts, channel lengths, and ring-buffer statistics

  This implementation follows the pattern already deployed in production by the network-tracer closed connection path (pkg/network/tracer/connection), providing a foundation for USM event handling.

* Existing UT pass with the DirectConsumer.
* Soak Test [results](https://dddev.datadoghq.com/dashboard/gqu-2wz-asy/usm-performance-evaluation-20?fromUser=true&refresh_mode=paused&tpl_var_base_agent-env%5B0%5D=usmon-1458-soak-base&tpl_var_client-service%5B0%5D=python-k6-client-http&tpl_var_compare_agent-env%5B0%5D=usmon-1458-soak-new&tpl_var_http.version%5B0%5D=http%2F1.1&tpl_var_kube_cluster_name%5B0%5D=usm3&tpl_var_server-service%5B0%5D=python-httpbin&from_ts=1759668284434&to_ts=1759732354000&live=false).
* Production Load Test [results](https://dddev.datadoghq.com/dashboard/gqu-2wz-asy/usm-performance-evaluation-20?fromUser=true&refresh_mode=paused&tpl_var_base_agent-env%5B0%5D=usmon-1458-production-base&tpl_var_client-service%5B0%5D=python-k6-client-http&tpl_var_compare_agent-env%5B0%5D=usmon-1458-production-new&tpl_var_http.version%5B0%5D=http%2F1.1&tpl_var_kube_cluster_name%5B0%5D=usm3&tpl_var_server-service%5B0%5D=python-httpbin&from_ts=1759676007610&to_ts=1759680445688&live=false).

### Additional Notes

The DirectConsumer mechanism is disabled by default and can be enabled via a configuration flag. We’ll be rolling it out incrementally to ensure stability and avoid performance regression.


rahulkaukuntla (approved):

pkg/config/setup/system_probe_usm.go
```diff
@@ -30,6 +30,10 @@ func initUSMSystemProbeConfig(cfg pkgconfigmodel.Setup) {
 	// By setting this value to 100, the channel will buffer up to ~400KB of data in the Go heap memory.
 	cfg.BindEnvAndSetDefault(join(smNS, "data_channel_size"), 100)
 	cfg.BindEnvAndSetDefault(join(smNS, "disable_map_preallocation"), true)
+	cfg.BindEnvAndSetDefault(join(smNS, "direct_consumer_buffer_wakeup_count"), 32)
+	cfg.BindEnvAndSetDefault(join(smNS, "direct_consumer_channel_size"), 1000)
+	cfg.BindEnvAndSetDefault(join(smNS, "direct_consumer_perf_buffer_size_per_cpu"), 65536) // 64KB per CPU
+	cfg.BindEnvAndSetDefault(join(smNS, "direct_consumer_ring_buffer_size_per_cpu"), 65536) // 64KB per CPU
```
guyarb: can we migrate it into a tree like structure?

pkg/network/config/config_linux.go
```diff
@@ -14,3 +14,8 @@ import (
 func (c *Config) RingBufferSupportedNPM() bool {
 	return (features.HaveMapType(cebpf.RingBuf) == nil) && c.NPMRingbuffersEnabled
 }
+
+// RingBufferSupportedUSM returns true if the kernel supports ring buffers and the config enables them
+func (c *Config) RingBufferSupportedUSM() bool {
+	return (features.HaveMapType(cebpf.RingBuf) == nil) && c.EnableUSMRingBuffers
+}
```
guyarb: Let's move it to `usm_config_linux.go`, so USM team will own it

pkg/network/config/usm_config.go
```diff
@@ -49,6 +49,18 @@ type USMConfig struct {
...
+	USMDirectConsumerBufferWakeupCount int
+
+	// USMDirectConsumerChannelSize specifies the channel buffer size multiplier for USM direct consumer.
+	USMDirectConsumerChannelSize int
+
+	// USMDirectConsumerPerfBufferSizePerCPU specifies the per-CPU perf buffer size for USM direct consumer.
+	USMDirectConsumerPerfBufferSizePerCPU int
+
+	// USMDirectConsumerRingBufferSizePerCPU specifies the per-CPU ring buffer size for USM direct consumer.
+	USMDirectConsumerRingBufferSizePerCPU int
```
guyarb: You're already in a USMConfig object, do we need the USM prefix in the fields name as well?

pkg/network/config/usm_config_test.go
```diff
@@ -136,6 +136,66 @@ func TestMaxUSMConcurrentRequests(t *testing.T) {
...
 }
 
+func TestUSMDirectBufferWakeupCount(t *testing.T) {
+	t.Run("default value", func(t *testing.T) {
+		mock.NewSystemProbe(t)
+		cfg := New()
+		assert.Equal(t, 32, cfg.USMDirectConsumerBufferWakeupCount)
+	})
+
+	t.Run("env override", func(t *testing.T) {
```
guyarb: what about yaml override?
same comment for all tests

pkg/network/config/usm_config_test.go
```diff
@@ -136,6 +136,66 @@ func TestMaxUSMConcurrentRequests(t *testing.T) {
...
+		mock.NewSystemProbe(t)
+		t.Setenv("DD_SERVICE_MONITORING_CONFIG_DIRECT_CONSUMER_BUFFER_WAKEUP_COUNT", "32")
```
guyarb: you're not overriding anything, the default is already `32`
we won't be able to catch breaking changes

pkg/network/ebpf/c/protocols/flush.h
```diff
@@ -27,7 +27,13 @@ http2 is supported only from kernel 5.2, therefore it does not have the kprobe v
...
-    http_batch_flush_with_telemetry(ctx);
+    __u64 use_direct_consumer = 0;
+    LOAD_CONSTANT("use_direct_consumer", use_direct_consumer);
+    
+    if (!use_direct_consumer) {
+        // Only flush when using BatchConsumer (kernel < 5.8)
+        http_batch_flush_with_telemetry(ctx);
+    }
     return 0;
 }
```
guyarb, line 32: I believe this implementation is redundant.
The user mode is the one that both sets the constant, and decides which probes to enable.
Can't just disable the probe?

pkg/network/ebpf/c/protocols/http/http.h
```diff
@@ -35,7 +35,35 @@ static __always_inline void http_begin_response(http_transaction_t *http, const
...
-static __always_inline void http_batch_enqueue_wrapper(conn_tuple_t *tuple, http_transaction_t *http) {
+static __always_inline __u64 get_ringbuf_flags(size_t data_size) {
+    __u64 ringbuffer_wakeup_size = 0;
+    LOAD_CONSTANT("ringbuffer_wakeup_size", ringbuffer_wakeup_size);
+    if (ringbuffer_wakeup_size == 0) {
+        return 0;
+    }
+    __u64 sz = bpf_ringbuf_query(&http_batch_events, DD_BPF_RB_AVAIL_DATA);
+    return (sz + data_size) >= ringbuffer_wakeup_size ? DD_BPF_RB_FORCE_WAKEUP : DD_BPF_RB_NO_WAKEUP;
+}
```
guyarb: As we will migrate all other protocols to the same mechanism, shouldn't we move this utility and similar others, into a dedicated file/directory, so we won't have to include `http.h` file in other implementations?

pkg/network/protocols/events/batch_consumer.go
```diff
@@ -0,0 +1,272 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
```
guyarb:
```suggestion
// Copyright 2025-present Datadog, Inc.
```

pkg/network/protocols/modifier.go
```diff
@@ -0,0 +1,21 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
```
guyarb: 2025

pkg/network/ebpf/c/protocols/flush.h
```diff
@@ -27,7 +27,13 @@ http2 is supported only from kernel 5.2, therefore it does not have the kprobe v
 
 SEC("tracepoint/net/netif_receive_skb")
 int tracepoint__net__netif_receive_skb_http(void *ctx) {
-    http_batch_flush_with_telemetry(ctx);
+    __u64 use_direct_consumer = 0;
```
usamasaqib: It would be better if we do not load and attach the tracepoint altogether when batching is disabled. Then `use_direct_consumer` will not be needed here, and we will not pay for a tracepoint hit.

pkg/network/ebpf/c/protocols/http/http.h
```diff
@@ -35,7 +35,35 @@ static __always_inline void http_begin_response(http_transaction_t *http, const
...
 }
 
-static __always_inline void http_batch_enqueue_wrapper(conn_tuple_t *tuple, http_transaction_t *http) {
+static __always_inline __u64 get_ringbuf_flags(size_t data_size) {
+    __u64 ringbuffer_wakeup_size = 0;
+    LOAD_CONSTANT("ringbuffer_wakeup_size", ringbuffer_wakeup_size);
+    if (ringbuffer_wakeup_size == 0) {
+        return 0;
+    }
+    __u64 sz = bpf_ringbuf_query(&http_batch_events, DD_BPF_RB_AVAIL_DATA);
```
usamasaqib: Please add a comment saying that `sz` is the amount of data waiting to be consumed.

pkg/network/ebpf/c/protocols/http/http.h
```diff
@@ -35,7 +35,35 @@ static __always_inline void http_begin_response(http_transaction_t *http, const
...
+
+    long perf_ret;
+    if (ringbuffers_enabled) {
+        perf_ret = bpf_ringbuf_output_with_telemetry(&http_batch_events, event, sizeof(http_event_t), get_ringbuf_flags(sizeof(http_event_t)));
+    } else {
+        u32 cpu = bpf_get_smp_processor_id();
+        perf_ret = bpf_perf_event_output_with_telemetry(ctx, &http_batch_events, cpu, event, sizeof(http_event_t));
+    }
+
+    if (perf_ret < 0) {
```
usamasaqib: Do we need this, given that we have telemetry above?

pkg/network/protocols/events/batch_consumer.go
guyarb, line 1: was this code copied/moved from somewhere?

pkg/network/protocols/events/batch_consumer.go
```diff
@@ -0,0 +1,272 @@
...
+	// termination
+	eventLoopWG sync.WaitGroup
+	stopped     bool
+
+	// telemetry
+	metricGroup                                        *telemetry.MetricGroup
+	eventsCount                                        *telemetry.Counter
+	failedFlushesCount                                 *telemetry.Counter
+	kernelDropsCount                                   *telemetry.Counter
+	lengthExceededEventCount, negativeLengthEventCount *telemetry.Counter
```
guyarb, line 55: nit: split fields to 2 separate lines

pkg/network/protocols/events/direct_consumer.go
```diff
@@ -0,0 +1,225 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
```
guyarb:
```suggestion
// Copyright 2025-present Datadog, Inc.
```

pkg/network/protocols/http/protocol.go
```diff
@@ -125,11 +126,26 @@ func newHTTPProtocol(mgr *manager.Manager, cfg *config.Config) (protocols.Protoc
...
+		return nil, err
+	}
+
+	return p, nil
+}
+
+// Modifiers implements the ModifierProvider interface
+func (p *protocol) Modifiers() []ddebpf.Modifier {
+	if p.consumer == nil {
+		return []ddebpf.Modifier{}
```
guyarb:
```suggestion
		return nil
```

pkg/network/protocols/http/protocol.go
```diff
@@ -263,3 +302,39 @@ func (p *protocol) GetStats() (*protocols.ProtocolStats, func()) {
...
 	return true
 }
+
+// createAdaptiveConsumer creates the appropriate consumer based on configuration and kernel version
+// and determines which callback method to use internally
+func (p *protocol) createAdaptiveConsumer() error {
+	kernelVersion, err := kernel.HostVersion()
+	if err != nil {
+		return err
+	}
```
guyarb: fetch it near line 330

pkg/network/protocols/events/direct_consumer.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	// Set up perf mode and channel size similar to initClosedConnEventHandler
+	// For DirectConsumer, we want kernel-level batching for performance
+	// but individual event processing in userspace (unlike BatchConsumer)
+	perfMode := perf.WakeupEvents(config.USMDirectConsumerBufferWakeupCount) // Wait for N events before wakeup
+	chanSize := config.USMDirectConsumerChannelSize * config.USMDirectConsumerBufferWakeupCount
+
+	// Calculate total buffer sizes from per-CPU values
+	totalPerfBufferSize := config.USMDirectConsumerPerfBufferSizePerCPU * numCPUs
```
usamasaqib: For perf buffer the size is supposed to be specified per-cpu. Right now it seems you are overallocating.

pkg/network/protocols/events/direct_consumer.go
```diff
@@ -0,0 +1,225 @@
...
+
+	mapName := eventMapName(proto)
+	eventHandler, err := perf.NewEventHandler(
+		mapName,
+		handler,
+		mode,
+		perf.SendTelemetry(config.InternalTelemetryEnabled),
+		perf.RingBufferEnabledConstantName("ringbuffers_enabled"),
+		perf.RingBufferWakeupSize("ringbuffer_wakeup_size",
+			uint64(config.USMDirectConsumerBufferWakeupCount*(eventSize+unix.BPF_RINGBUF_HDR_SZ))),
```
usamasaqib: Is `USMDirectConsumerBufferWakeupCount` supposed to be the total number of events recorded before issuing wakeup? If so, then the setup of perf buffers above is incorrect, since the wakeup_events field is per-cpu buffer.

If this is supposed to be the total number of events per-cpu, then this line is incorrect, since ring buffers will cause more userspace wakeups due to smaller wakeup interval.

I would suggest to interpret `USMDirectConsumerBufferWakeupCount` as the per-cpu wakeup size, and adjust for the ring buffer case accordingly.

pkg/network/protocols/events/direct_consumer.go
DanielLavie: Not sure I understood the comment, can you please clarify?

usamasaqib: Perf buffers allocate a buffer per-cpu. When setting up a perf buffer we specify the size of the per-cpu buffer and the underlying loader sets up `numCPU` number of `perfBufferSize` sized buffers.

In your code you will end of with `numCPU` number of `config.USMDirectConsumerPerfBufferSizePerCPU * numCPUs` sized buffers.

In other words, `totalPerfBufferSize := config.USMDirectConsumerPerfBufferSizePerCPU` is the correction.

pkg/network/protocols/events/batch_consumer.go
DanielLavie: Yes, it was in the `consumer.go` file

pkg/network/protocols/events/batch_consumer.go
DanielLavie: This is unrelated to this PR, can be done in a different one

pkg/network/ebpf/c/protocols/http/http.h
DanielLavie: Created protocols/direct_consumer.h with USM_DIRECT_CONSUMER_INIT macro to generate protocol-specific DirectConsumer utility functions. This follows the same pattern as events.h's USM_EVENTS_INIT macro for BatchConsumer

pkg/network/protocols/events/direct_consumer.go
DanielLavie: `USMDirectConsumerBufferWakeupCount` defaults to 32 (which matches current behavior and looked good in our load test). But our test nodes have 64 CPUs, so to match that I’d need to set the flag to 0.5, which doesn’t make sense for something called a “Count".

matthewleese (approved):

pkg/network/ebpf/c/protocols/direct_consumer.h
```diff
@@ -0,0 +1,46 @@
...
+        if (ringbuffers_enabled) {                                                                          \
+            bpf_ringbuf_output_with_telemetry(&map_name, event, sizeof(event_type),                        \
+                                               name##_get_ringbuf_flags(sizeof(event_type)));               \
+        } else {                                                                                            \
+            u32 cpu = bpf_get_smp_processor_id();                                                           \
+            bpf_perf_event_output_with_telemetry(ctx, &map_name, cpu, event, sizeof(event_type));          \
+        }                                                                                                   \
+    }
+
+#endif // __USM_DIRECT_CONSUMER_H
```
guyarb: super nit - can you fix the indentation of `\` and add a new line at the end of the file?

pkg/network/protocols/http/protocol.go
```diff
@@ -163,14 +163,20 @@ func (p *protocol) ConfigureOptions(opts *manager.Options) {
...
+	if !p.useDirectConsumer {
+		netifProbeID := manager.ProbeIdentificationPair{
+			EBPFFuncName: netifProbe,
+			UID:          eventStream,
+		}
+		if usmconfig.ShouldUseNetifReceiveSKBCoreKprobe() {
+			netifProbeID.EBPFFuncName = netifProbe414
+		}
+		opts.ActivatedProbes = append(opts.ActivatedProbes, &manager.ProbeSelector{ProbeIdentificationPair: netifProbeID})
 	}
```
guyarb, line 181: on else, we should add the probe to `options.ExcludedFunctions`

guyarb: although it won't be activated either way, the exclusion ensures it won't be loaded/verified as well

pkg/network/protocols/http/protocol.go
```diff
@@ -327,6 +328,10 @@ func (p *protocol) createAdaptiveConsumer() error {
 			log.Debugf("HTTP monitoring: using direct consumer (requested via configuration)")
 		} else {
 			// Fall back to BatchConsumer on unsupported kernels
+			kernelVersion, err := kernel.HostVersion()
+			if err != nil {
+				return err
+			}
```
guyarb, line 339: no need to return error here, just tweak the warn log
although I doubt we will ever see such an error, I'd still lean into avoiding from such paths in the code

guyarb (approved):
Left a couple of small comments

pkg/network/protocols/events/direct_consumer.go
usamasaqib: Currently the wakeup rate between perf buffers and ring buffers would be different, with ring buffers causing more wakeup than perf buffers in most cases. I was suggesting to scale the wakeup count by the number of CPUs to account for this difference, especially on larger machines.

As an illustration, assuming equal distribution of events across CPUs, with `USMDirectConsumerBufferWakeupCount = 32` and `numCPUs = 64`, **in the worst case**, a wakeup when using perf-buffers would occur after `1985 (31*64 + 1 <- trigger wakeup)` events as compared to `32` when using ring buffers.

pkg/network/protocols/http/protocol.go
DanielLavie: Done

pkg/network/protocols/events/direct_consumer.go
DanielLavie: btw I believe the connection tracer has the same issue in `initClosedConnEventHandler`

DanielLavie: btw I think the `initClosedConnEventHandler` has the same issue

DanielLavie: Done

pkg/network/ebpf/c/protocols/http/http.h
```diff
@@ -12,6 +12,10 @@
 #include "protocols/http/maps.h"
 #include "protocols/http/usm-events.h"
 #include "protocols/tls/https.h"
+#include "protocols/direct_consumer.h"
+
+// Initialize DirectConsumer utilities for HTTP protocol
+USM_DIRECT_CONSUMER_INIT(http, http_event_t, http_batch_events)
```
guyarb: should be in `usm-events.h`

DanielLavie: Done

PR #39775, opened 2025-08-12
Allow rc_kubernetes_deploy job to fail

kacper-murzyn:
### What does this PR do?

This PR sets `allow_failure: true` on `rc_kubernetes_deploy` job.

### Motivation

The last job in the child pipeline created by rc_kubernetes_deploy is manual at the moment. We will clean that soon after enough verification, but for the time being we can allow this to fail.


sabrina-datadog (approved):

.gitlab/internal_kubernetes_deploy/rc_kubernetes_deploy.yml
```diff
@@ -25,6 +25,7 @@ rc_kubernetes_deploy:
       optional: true
   image: registry.ddbuild.io/ci/datadog-agent-buildimages/linux$CI_IMAGE_LINUX_SUFFIX:$CI_IMAGE_LINUX
   tags: ["arch:amd64"]
+  allow_failure: true
```
spencergilbert, line 29: > We will clean that soon after enough verification, but for the time being we can allow this to fail.

Can we leave a TODO comment with a @/user?

PR #39779, opened 2025-08-12
CONTINT-4744 expose init container metrics.

lavigne958:
### What does this PR do?

Expose the bellow init container metrics:

- `kube_pod_init_container_status_read`
  - 1 to 1 direct mapping
  - no changes are made from the source metric
- `kube_pod_init_container_status_running`
  - 1 to 1 direct mapping
  - no changes are made from the source metric
- `kube_pod_init_container_status_restarts_total`
  - was already defined and already exposed, nothing to do here
- `kube_pod_init_container_resource_limits`
  - this is generated from a custom resource generator, in order to expose a unified value.
  - this is transformed to expose a metric per allowed resource ('memory','cpu')
- `kube_pod_init_container_resource_requests`
  - this is generated from a custom resource generator, in order to expose a unified value.
  - this is transformed to expose a metric per allowed resource ('memory','cpu')
- `kube_pod_init_container_status_waiting_reason`
  - this is transformed in order to expose a single counter for all waiting reasons.
- `kube_pod_init_container_status_terminated_reason`
  - this is transformed in order to expose a single counter for all termination reasons.

it has been discussed that refactoring the current code to handle init+standard containers is at first glance the simplest way to add init container metrics.

To achieve this a new enum-like type has been added, 2 constants have been created: `Init=initcontainer` and `Standard=container`.

The function `containerResourceOwnerGenerator` contains the common code for standard containers and init containers

The calling function `customResourceOwnerGenerator` will take the decision of taking Resource* metrics for init containers *only* if they are actually sidecars.
We won't collect Resource* metrics for short lived init containers as they will always die before the standard containers start.

ran the agent, check each metrics in the metrics explorer
added basic unit tests.

closes: CONTINT-4744

### DIFF
--- a/pkg/collector/corechecks/cluster/ksm/kubernetes_state_transformers.go
+++ b/pkg/collector/corechecks/cluster/ksm/kubernetes_state_transformers.go
@@ -20,6 +20,7 @@ import (
 	"github.com/samber/lo"
 
 	"github.com/DataDog/datadog-agent/pkg/aggregator/sender"
+	crs "github.com/DataDog/datadog-agent/pkg/collector/corechecks/cluster/ksm/customresources"
 	ksmstore "github.com/DataDog/datadog-agent/pkg/kubestatemetrics/store"
 	"github.com/DataDog/datadog-agent/pkg/metrics/servicecheck"
 	"github.com/DataDog/datadog-agent/pkg/util/log"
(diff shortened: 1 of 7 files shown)


estherk15 (approved):

lavigne958:
squashed commits + rebase onton main

pkg/collector/corechecks/cluster/ksm/kubernetes_state_transformers.go
```diff
@@ -274,29 +287,47 @@ var allowedTerminatedReasons = map[string]struct{}{
...
-// containerTerminatedReasonTransformer validates the container waiting reasons for metric kube_pod_container_status_terminated_reason
 func containerTerminatedReasonTransformer(s sender.Sender, _ string, metric ksmstore.DDMetric, hostname string, tags []string, _ time.Time) {
+	submitContainerTerminatedReasonTransformer(s, metric, hostname, tags, crs.Standard)
+}
+
+func initcontainerTerminatedReasonTransformer(s sender.Sender, _ string, metric ksmstore.DDMetric, hostname string, tags []string, _ time.Time) {
+	submitContainerTerminatedReasonTransformer(s, metric, hostname, tags, crs.Init)
+}
+
+// containerTerminatedReasonTransformer validates the container waiting reasons for metric kube_pod_container_status_terminated_reason
```
justin-lesko: nit: comment does not match function name

justin-lesko (approved):

pkg/collector/corechecks/cluster/ksm/kubernetes_state_transformers.go
lavigne958: good catch thanks.

PR #39784, opened 2025-08-12
dogstatsd: Migrate string interner to Go's unique.Handle

scottopell:
## What
Replace DogStatsD's custom string interner with Go 1.23's `unique.Handle`.

## Why
The current implementation uses a manual map-based cache with size limits and periodic resets. Go 1.23's `unique.Handle` provides automatic string interning with GC-managed memory, eliminating the need for manual cache management.

## Changes

**New architecture:**
- `stringInternerInterface` abstracts the two implementations
- Runtime selection via `dogstatsd_string_interner_use_unique` config (default: false)
- Legacy implementation (`intern.go`) preserves existing map-based logic
- Unique implementation (`intern_unique.go`) uses `unique.Handle[string]`

**Configuration:**
- Added `dogstatsd_string_interner_use_unique` setting (default: false)
- No build tags required - works across Go versions

**Telemetry:**
- Legacy: all metrics (`hits`, `miss`, `resets`, `entries`, `bytes`, `str_bytes`)
- Unique: limited to `hits`, `miss`, `str_bytes` (cache internals not accessible)

## Testing
Both implementations have full test coverage with implementation-specific test files.

## Rollout
Opt-in via config to enable safe testing and gradual adoption.


scottopell:
* **#39784** <a href="https://app.graphite.dev/github/pr/DataDog/datadog-agent/39784?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/datadog-agent/39784?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

comp/dogstatsd/server/intern.go
```diff
@@ -56,20 +59,39 @@ func (i *stringInterner) LoadOrStore(key []byte) string {
...
 	i.strings[s] = s
 
-	i.telemetry.Miss(len(s))
+	if i.telemetry.enabled {
+		i.telemetry.miss.Inc()
+		i.telemetry.size.Inc()
+		i.telemetry.bytes.Add(float64(len(s)))
+		i.telemetry.globaltlmSIRStrBytes.Observe(float64(len(s)))
+		i.curBytes += len(s)
+	}
```
tobz: I'm confused by some of the telemetry changes.

It looks like we still have helper functions like `Miss`, but aren't using them. We also moved stuff like tracking the current number of interned bytes to a field  on the interner itself instead of just leaving it on the telemetry class... 🤔

tobz (changes requested):
Overall, this looks straightforward and worth experimenting with.

Left one specific comment. Only other thing I think would just be adding at least one superficial test for the unique-based interner implementation to assert that it, in fact, both strings returned for the same input are in fact the same string... I guess by comparing their pointers for equality or whatever?

comp/dogstatsd/server/intern.go
scottopell: thanks, that was accidental AI slop, removed now.

rahulkaukuntla (approved):

PR #39786, opened 2025-08-12
Add secret-generic-connector.exe to Windows Image

dustmop:
### What does this PR do?

Add secret-generic-connector.exe to the Windows image

### Motivation

Support the new secret backend type on Windows


rahulkaukuntla (approved):

pkg/util/defaultpaths/path_windows.go
```diff
@@ -8,18 +8,22 @@ package defaultpaths
...
+	// NOTE: Do NOT calculate paths relative to the executable.
+	//       The agent executables are not all installed in the same path and this can
+	//       lead to incorrect path calculations when this package is imported by
+	//       other executables.
+	// installPath is the default install path for the Agent. The install path is
+	// customizable at install time, so we update it from the registry in init().
+	installPath = `C:\Program Files\Datadog\Datadog Agent`
 	// PyChecksPath holds the path to the python checks from integrations-core shipped with the agent
-	PyChecksPath = filepath.Join(_here, "..", "checks.d")
+	PyChecksPath = filepath.Join(installPath, "checks.d")
```
pgimalac: Why do we set default values here if they're overridden in the `init` function just after ?

pgimalac: I think we should have a const with the default value, and then set those variables to a "runtime" value

clarkb7: i think it depends on if you need the default value or if you only ever need the correct value. I don't have a strong opinion either way.
However If the default const was exported I'd be concerned about it being accidentally used instead of the runtime updated/correct value.

dustmop: I tweaked the code a bit so that the defaultInstallPath is only used if the registry can't read the installation path, instead of assigning it to the global var from the beginning.

pgimalac (approved):

rahulkaukuntla:
Rebasing might fix the failing [windows_msi_and_bosh_zip_x64-a7-fips](https://gitlab.ddbuild.io/datadog/datadog-agent/builds/1086033899) job, as this pr was just merged into main: https://github.com/DataDog/datadog-agent/pull/39997

pkg/util/defaultpaths/path_windows.go
```diff
@@ -51,29 +59,35 @@ func init() {
...
+		// if the key isn't there, we might be running a standalone binary that wasn't installed through MSI
+		log.Debugf("Failed to open registry key: %s", err)
 	} else {
 		defer k.Close()
-		s, _, err = k.GetStringValue("InstallPath")
+		s, _, err := k.GetStringValue("InstallPath")
 		if err != nil {
 			log.Warnf("Installpath not found in registry: %s", err)
 		}
+		if s != "" {
```
rdesgroppes:
```suggestion
		} else if s != "" {
```
maybe?

rdesgroppes (approved):
LGTM

clarkb7 (approved):

.gitlab/container_build/docker_windows.yml
```diff
@@ -26,6 +26,7 @@
     # Omnibus uses a slightly different path than the actual install
     # TODO: This overwrites an extra agent.exe, we should also remove it from the zip
     - mv "Datadog Agent\bin\agent\agent.exe" "Datadog Agent\bin\agent.exe" -Force
+    - mv "Datadog Agent\bin\agent\secret-generic-connector.exe" "Datadog Agent\bin\secret-generic-connector.exe" -Force
```
clarkb7:
```suggestion
    # TODO(AGENTCFG-XXX): SGC is not supported in FIPS mode
    - if (Test-Path "Datadog Agent\bin\agent\secret-generic-connector.exe") { mv "Datadog Agent\bin\agent\secret-generic-connector.exe" "Datadog Agent\bin\secret-generic-connector.exe" -Force }
```

FlorentClarret (approved):

PR #39790, opened 2025-08-12
Add the Agent 3rd party deps available from the Bazel central repository

aiuto:
Add the dependencies identified in the PoC which are available via the bazel central repository, and do not require a custom BUILD.bazel file.

Along with this, provide a smoke test target to build them all.


deps/repos.MODULE.bazel
```diff
@@ -1 +1,23 @@
...
+
+# Defies a "liblua" target for the lib
+bazel_dep(name = "lua", version = "5.4.6")
+
+# Defines a "ncurses" target for the lib.
+# There are also "form", "menu", "panel" lib targets and "make_hash" binary
+bazel_dep(name = "ncurses", version = "6.4.20221231.bcr.2")
+
+# Defines a "sqlite3" target for the lib
+bazel_dep(name = "sqlite3", version = "3.46.1")
```
alopezz, line 13: Can be left for later, but bear in mind that some of these may require updates (https://github.com/DataDog/datadog-agent/blob/2c99235401284f868a82a72f72015a83ba47b9a6/omnibus/config/software/libsqlite3.rb#L2C18-L2C24)

```suggestion
bazel_dep(name = "sqlite3", version = "3.50.4")
```

deps/repos.MODULE.bazel
```diff
@@ -1 +1,23 @@
...
+
+# Defines a "libyaml" target for the lib
+bazel_dep(name = "libyaml", version = "0.2.5")
+
+# Defies a "liblua" target for the lib
+bazel_dep(name = "lua", version = "5.4.6")
+
+# Defines a "ncurses" target for the lib.
+# There are also "form", "menu", "panel" lib targets and "make_hash" binary
+bazel_dep(name = "ncurses", version = "6.4.20221231.bcr.2")
```
alopezz: We dropped ncurses at [one point](https://github.com/DataDog/datadog-agent/commit/e8c7b552372c626b180f1fed9508deb25b51c59a) so we can get rid of this.

```suggestion
```

alopezz (approved):
Looks good to me with the ncurses removal.

deps/repos.MODULE.bazel
aiuto: Removed.

deps/repos.MODULE.bazel
aiuto: Let's deal later. sqlite3 has many configure options. I think we need a private BUILD file instead of what is in the BCR.

aiuto: agette said the same. Let's deal with this later, since sqlite probably needs more work.

deps/repos.MODULE.bazel
aiuto: removed

PR #39780, opened 2025-08-12
Override system `awscli` dependency in the macOS runners

ofek:
### Motivation

This uses the Python environment managed by `dda` in order to resolve the newly encountered errors. Other attempts:

- https://github.com/DataDog/datadog-agent/pull/39769
- https://github.com/DataDog/datadog-agent/pull/39764

PR #39791, opened 2025-08-12
Fix PATH propagation on macOS jobs

ofek:
### Motivation

https://github.com/DataDog/datadog-agent/pull/39780#discussion_r2271378108

Wasn't caught because the exit code of some commands are ignored.


rdesgroppes (approved):
LG

PR #39792, opened 2025-08-12
Increase the file scan frequency for logs-agent based quality gates to smooth out expected RSS usage

ddrthall:

### DIFF
--- a/test/regression/cases/quality_gate_idle_all_features/datadog-agent/datadog.yaml
+++ b/test/regression/cases/quality_gate_idle_all_features/datadog-agent/datadog.yaml
@@ -14,6 +14,10 @@ telemetry.checks: '*'
 
 logs_enabled: true
 logs_config:
+  # Reducing the scan period to 5 seconds in order to smooth out the RSS fluctuations that occur during file rotation.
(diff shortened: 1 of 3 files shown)


test/regression/cases/quality_gate_idle_all_features/datadog-agent/datadog.yaml
```diff
@@ -14,6 +14,10 @@ telemetry.checks: '*'
 
 logs_enabled: true
 logs_config:
+  # Reducing the scan period to 5 seconds in order to smooth out the RSS fluctuations that occur during file rotation.
+  # This will allow tighter memory bounds on the quality gates that are more likely to catch code-caused memory regressions, 
+  # at the cost of deviating slightly from the expected production base behavior.
+  file_scan_period: 5
```
blt, line 20: The quality gate is really meant to demonstrate 'realistic' behavior, so I'd argue that we shouldn't be fiddling with the file_scan_period unless we intend to adjust that by default in the Agent. That is, the PSS -- not RSS -- spike we see in nightlies is what a user would see in practice.

PR #38320, opened 2025-06-30
[AGENTRUN-516] Introduce helper to run go build command

pgimalac:
### What does this PR do?
Introduce a common helper to run `go build` commands in invoke tasks.

### Motivation
Make it easier to introduce changes to all built Go artifacts, eg. passing a given environment variable, adding arguments by default, parsing output to detect issues, etc.
This also avoids code duplication.

### Describe how you validated your changes

CI is enough, no functional changes are expected.

### Additional Notes

This is a first step to ease checking for dead code elimination in Go binaries (and avoid copy pasting associated code for every binary): https://github.com/DataDog/datadog-agent/pull/38366/s

PR #39793, opened 2025-08-12
Properly adjust ownership of Go build artifacts

ofek:
### What does this PR do?

This changes the owner of binaries produced by `go build` commands based on the `HOST_UID` and `HOST_GID` environment variables.

### Motivation

Allow for builds within containers when using a bind mount, see https://github.com/DataDog/datadog-agent-dev/pull/164

### Additional Notes

I took the liberty of adding a few extra tests for the core behavior of the `go_build` utility as there were no explicit tests, as mentioned [here](https://github.com/DataDog/datadog-agent/pull/38320).

### DIFF
--- a/tasks/libs/common/go.py
+++ b/tasks/libs/common/go.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import os
+import sys
 from pathlib import Path
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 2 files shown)


tasks/libs/common/go.py
```diff
@@ -66,4 +67,14 @@ def go_build(
...
-    return ctx.run(cmd, env=env)
+    result = ctx.run(cmd, env=env)
+    if sys.platform == "win32" or result.exited != 0 or bin_path is None:
+        return result
+
+    if os.path.exists(bin_path):
+        uid = os.environ.get("HOST_UID", "-1")
+        gid = os.environ.get("HOST_GID", "-1")
+        if uid != "-1" and gid != "-1":
+            os.chown(bin_path, int(uid), int(gid))
```
KevinFairise2, line 78: I understand that is required so that it works properly in dev environment, but what is the impact of that change on the artifact that we ship to customer?

ofek: Ownership occurs at the level of the filesystem and are not persisted unlike permissions which do modify files themselves.

KevinFairise2 (approved):

PR #39719, opened 2025-08-11
Install the binary distributions of dda on macOS runners

ofek:
### Motivation

Builds started failing this morning due to a change on the runners, likely related to `pyenv`.

PR #39796, opened 2025-08-13
Install the binary distributions of dda on macOS runners (#39719)

KevinFairise2:
### What does this PR do?
Backport  #39719 and #39780

Note the PR also bump the version off dda used in MacOS job. This is required for the fix to work. Otherwise the targetted version is a pretty old one used in the current 7.68.x buildimage

___

### Motivation

Builds started failing this morning due to a runner restart?


.gitlab/common/macos.yml
```diff
@@ -17,71 +17,36 @@
...
-      echo "Find python version matching $PYTHON_VERSION"
-      PYTHON_VERSION_ESCAPED="${PYTHON_VERSION//./\\.}"
-      versionmatches="$(ls "$(pyenv root)/versions" | grep -E "^${PYTHON_VERSION_ESCAPED}\\.[0-9]+$")"
-      if [ -n "$versionmatches" ]; then
-        # 2.a.
-        echo "Found already existing python version for $PYTHON_VERSION"
+    # Perform installation only if the directory does not exist
+    if [ ! -d "$DDA_DIR" ]; then
+      # Get the commit from the build image variable in the format `vPIPELINE_ID-COMMIT`
+      export DDA_VERSION="v0.24.0"
```
rdesgroppes, line 35: hardcoded here? Doesn't look like a strict backport of #39719

KevinFairise2: Yep, was trying to bump dda version to see if it helps.
If I do a strict backport, given that the buildimages used in 7.68.x is pretty old, it will target a old version of dda that will not work with the fix.
Other option would be to globally bump the buildimages version but I feel like it is even worse to bump all the buildimages on that branch

PR #39798, opened 2025-08-13
(fleet) add CAP_NET_BIND_SERVICE to the agent by default

arbll:
### What does this PR do ?

This PR adds CAP_NET_BIND_SERVICE to the agent units by default allowing it to bind to ports <=1024 by default.

### Motivation

We are currently asking customers to manually do `setcap 'cap_net_bind_service=+ep'` on their agent binary in many situations (logs collection, [20 integrations](https://github.com/search?q=repo%3ADataDog%2Fintegrations-core+setcap+language%3AMarkdown&type=code&l=Markdown), NDM traps). This is leading to quite a few support tickets, especially with SNMP traps. This setup also **breaks on upgrades** making it very tricky to maintain for customers.

Alternatively we could recommend the systemd overide instead of `setcap` but given that CAP_NET_BIND_SERVICE isn't a particularly sensitive capability I believe it makes sense to give it to the agent and simplify the onboarding of logs/snmp/integrations.


BaptisteFoy (approved):

releasenotes/notes/ambiant-capability-cap-net-bind-service-af418640739780b0.yaml
```diff
@@ -0,0 +1,14 @@
+---
+enhancements:
+  - |
+    On linux the agent is now able to bind to ports below 1024 by default,
+    simplifying common logs, SNMP and integrations configurations.
+    This is done by setting `AmbientCapabilities=CAP_NET_BIND_SERVICE` in the agent systemd units.
```
maycmlee:
```suggestion
    On Linux, the Agent is now able to bind to ports below 1024 by default,
    simplifying common logs, SNMP, and integrations configurations.
    This is done by setting `AmbientCapabilities=CAP_NET_BIND_SERVICE` in the Agent systemd units.
```

releasenotes/notes/ambiant-capability-cap-net-bind-service-af418640739780b0.yaml
```diff
@@ -0,0 +1,14 @@
...
+
+    If you wish to disable this, you can do so by creating a systemd drop-in file
```
maycmlee:
```suggestion
    If you want to disable this, create a systemd drop-in file
```

maycmlee (commented):
Just a couple of small suggestions.

maycmlee (approved):

PR #39800, opened 2025-08-13
HostTagsProvider: Configurable duration and new pakcage

guyarb:
### What does this PR do?

Allows HostTagProvider to get a custom duration

Moves HostTagProvider to `comp/metadata/host/hostimpl/hosttags` package

### Motivation

Extracted from https://github.com/DataDog/datadog-agent/pull/39695


vickenty (approved):

louis-cqrl (approved):

.github/CODEOWNERS
```diff
@@ -384,6 +384,7 @@
 /pkg/fips                               @DataDog/agent-runtimes
 /pkg/gohai                              @DataDog/agent-configuration
 /pkg/gpu/                               @DataDog/ebpf-platform
+/pkg/hosttags/                          @DataDog/agent-metric-pipelines
```
vickenty: We don't own host tags, so I'm not sure this is correct. I think #agent-runtimes would make more sense here.

vickenty (changes requested):

.github/CODEOWNERS
guyarb: Removed the change, it will be by default agent-runtimes

vickenty (approved):

PR #39801, opened 2025-08-13
[CWS] remove dig import from cws-instrumentation

paulcacheux:
### What does this PR do?

The `cws-instrumentation` binary doesn't use fx, this means that we don't need the extra error display offered by `cmd/internal/runcmd` and cobra's default is good enough. This allows to remove the dig dependency from this binary.

### DIFF
--- a/cmd/cws-instrumentation/main_linux.go
+++ b/cmd/cws-instrumentation/main_linux.go
@@ -11,11 +11,13 @@ import (
 
 	"github.com/DataDog/datadog-agent/cmd/cws-instrumentation/command"
 	"github.com/DataDog/datadog-agent/cmd/cws-instrumentation/subcommands"
-
-	"github.com/DataDog/datadog-agent/cmd/internal/runcmd"
 )
 
 func main() {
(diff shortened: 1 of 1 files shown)


cmd/cws-instrumentation/main_linux.go
```diff
@@ -11,11 +11,13 @@ import (
...
 )
 
 func main() {
 	rootCmd := command.MakeCommand(subcommands.CWSInjectorSubcommands())
-	os.Exit(runcmd.Run(rootCmd))
+
+	if err := rootCmd.Execute(); err != nil {
+		// the error has already been printed
+		os.Exit(1)
+	}
```
spikat: should we keep trying to exit with the relevant exit code? it could be use by control planes to handle errors

paulcacheux: before my change it was -1 on failure, 0 on success (https://github.com/DataDog/datadog-agent/blob/9b6988a83f619f648053c76a0364f83b72ca35f7/cmd/internal/runcmd/runcmd.go#L22). But you are right that we probably should keep the same number, let me fix this

spikat (approved):

PR #39830, opened 2025-08-13
[AGNTLOG-272] Fix Agent Removing "Integration" Type From Docker Labels

soberpeach:
### What does this PR do?

This PR makes it so the agent no longer removes "type: integration" from config providers. This was causing a bug where integrations configs configured via docker labels were being overridden, and therefore files were not being created for integration logs.

### Motivation

See above

Verfied using the following setup
`docker-compose.yml`:
```yaml
services:
  datadog-agent:
    image: agent:latest
    container_name: datadog-agent
    environment:
      - DD_API_KEY=${DD_API_KEY}
      - DD_SITE=${DD_SITE}
      - DD_LOGS_ENABLED=true
      - DD_LOG_LEVEL=debug
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock:ro
      - /proc/:/host/proc/:ro
      - /sys/fs/cgroup/:/host/sys/fs/cgroup:ro
      - ./custom_check.py:/etc/datadog-agent/checks.d/custom_check.py:ro
    depends_on:
      - data-generator

  data-generator:
    build:
      context: .
      dockerfile: Dockerfile.datagen
    container_name: my-data-generator
    ports:
      - "5005:5005"
    labels:
      com.datadoghq.ad.checks: >
        {
          "custom_check": {
            "instances": [
              {
                "metrics_url": "http://%%host%%:5005/metrics",
                "tags": ["env:dev", "app:data-generator"]
              }
            ],
            "logs": [
              {
                "type": "integration",
                "source": "python",
               "service": "data-generator-service"
              }
            ]
          }
        }
```
Where the agent image was built using `dda inv hacy-dev-image-build`.
`custom_check.py`
```pytyhon
# import requests


class CustomCheck(AgentCheck):
    def check(self, instance):
        log = {"message": "hello world"}
        self.send_log(log)
```
`data_generator.py`:
```python
from flask import Flask, jsonify
import random
import time

app = Flask(__name__)

start_time = time.time()

@app.route('/metrics')
def get_metrics():
    """
    Exposes custom metrics in JSON format.
    """
    # Simulate some application metrics
    requests_processed = int((time.time() - start_time) * 10) + random.randint(1, 100)
    active_connections = random.randint(5, 20)
    error_rate = random.uniform(0.01, 0.05)
    
    # Simulate an application status
    statuses = ["OK", "DEGRADED", "ERROR"]
    current_status = random.choices(statuses, weights=[90, 8, 2], k=1)[0]

    data = {
        'requests_processed': requests_processed,
        'active_connections': active_connections,
        'error_rate': error_rate,
        'status': current_status,
        'uptime_seconds': time.time() - start_time
    }
    
    return jsonify(data)

if __name__ == '__main__':
    # Running on 0.0.0.0 makes it accessible from other containers
    app.run(host='0.0.0.0', port=5005)


```
`Dockerfile.datagen`:
```dockerfile
# Use a lightweight Python base image
FROM python:3.9-slim

# Set the working directory in the container
WORKDIR /app

# Install Flask
RUN pip install Flask

# Copy the Flask application into the container
COPY data_generator.py .

# Expose the port the app runs on
EXPOSE 5005

# Command to run the application
CMD ["python", "data_generator.py"]
```
There's also a new unit test covering this use case

### DIFF
--- a/releasenotes/notes/fix-checks-from-labels-26b10d9d8952eefb.yaml
+++ b/releasenotes/notes/fix-checks-from-labels-26b10d9d8952eefb.yaml
@@ -8,5 +8,6 @@
 ---
 fixes:
   - |
-    Checks configured from Docker labels or k8s annotations are no
-    longer overridden in the scheduler.
+    Fix bug causing integrations configured via docker labels or K8s
+    annotations to be overwritten in the agent so logs from integrations
+    now work when configured using via docker labels or K8s annotations.
--- a/releasenotes/notes/fix-checks-from-labels-26b10d9d8952eefb.yaml
+++ b/releasenotes/notes/fix-checks-from-labels-26b10d9d8952eefb.yaml
@@ -8,5 +8,5 @@
 ---
 fixes:
   - |
-    Checks configured from docker labels / k8s annotations are no
+    Checks configured from Docker labels or k8s annotations are no
     longer overridden in the scheduler.
--- a/releasenotes/notes/fix-checks-from-labels-26b10d9d8952eefb.yaml
+++ b/releasenotes/notes/fix-checks-from-labels-26b10d9d8952eefb.yaml
@@ -1,0 +1,12 @@
+# Each section from every release note are combined when the
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+fixes:
+  - |
+    Checks configured from docker labels / k8s annotations are no
+    longer overridden in the scheduler.
--- a/pkg/collector/python/loader.go
+++ b/pkg/collector/python/loader.go
@@ -249,6 +249,7 @@ func (cl *PythonCheckLoader) Load(senderManager sender.SenderManager, config int
 	}
 
 	if v, ok := cl.logReceiver.Get(); ok {
+		log.Debugf("Registering integration in loader: %s", c.ID())
 		v.RegisterIntegration(string(c.id), config)
 	}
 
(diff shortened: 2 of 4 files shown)


piochelepiotr (approved):
Thank you for making the fix!

ddrthall (approved):

releasenotes/notes/fix-checks-from-labels-26b10d9d8952eefb.yaml
```diff
@@ -0,0 +1,12 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+fixes:
+  - |
+    Checks configured from docker labels / k8s annotations are no
```
brett0000FF:
```suggestion
    Checks configured from Docker labels or k8s annotations are no
```

brett0000FF (approved):

releasenotes/notes/fix-checks-from-labels-26b10d9d8952eefb.yaml
```diff
@@ -0,0 +1,12 @@
...
+    longer overridden in the scheduler.
```
jszwedko: These release notes should be understandable by customers so I'd describe the fix in terms of the impact rather than the implementation (i.e. what behavior are we fixing?). For example, customers are unlikely to know what the "scheduler" is or what the effect of "no longer overridden" is.

soberpeach: Makes sense 👍, updating it

jeremy-hanna (approved):

PR #39843, opened 2025-08-14
[AGENTRUN-629] Moving IsCLCRunner to pkg/config/helpers

misteriaud:
### What does this PR do?
This PR does not change any behavior/feature. It's only refactoring.
This PR move the helper function `IsCLCRunner()` from `pkg/config/setup` to `pkg/config/helpers`.

### Motivation
I need to use `IsCLCRunner()` function in a component, but importing `pkg/config/setup` package in Agent components is forbidden.

No code change.


dustmop (commented):
This change itself seems okay, but I see from the CI that the static quality gate is failing for the IOT agent. I'm guessing this is because the IOT agent is using IsCLCRunner somewhere, so this change causes it to pull in all of `pkg/config/utils`. `pkg/config/utils` is unfortunately not very slim, so the other parts of this package are adding unnecessary bloat. I would suggest instead moving `IsCLCRunner` to a brand new package such as `pkg/config/helpers`, exact name is up to you.

misteriaud:
Thanks @dustmop ! Indeed I was wondering why this refactor had this impact on the IOT binary size 😅
I'll move it to its own package 👍

sblumenthal (approved):

robertjli (approved):

gabedos (approved):
Change looks good! Could you also make @DataDog/container-platform at least a co-codeowner for the `config/helpers/clcrunner.go`? Thanks!

dustmop (approved):
Thanks for the fixes!

kangyili (approved):

PR #39844, opened 2025-08-14
[SQG] Disable flaky iot rpm gate

JSGette:
### What does this PR do?
It disables `static_quality_gate_iot_agent_rpm_arm64` due to its flakiness. Even though Gitlab reports that the artifact's size is below the threshold the quality gate reports higher value every now and then. Retrigger of the SQG job doesn't help, the artifact has to be rebuilt to overcome the issue.


alopezz (approved):
It will come back :)
I'll note that it's hard to make increase the size of this package without increasing that of the corresponding `deb` variant, so the risk of this temporary disablement is low.

PR #39847, opened 2025-08-14
[AGENTRUN-632] Fix SEGFAULT happening during JMXFetch checks

misteriaud:
### What does this PR do?

Fixes a nil pointer dereference panic in the `agent check` command when running JMX checks with debug logging enabled. Replaces `jmxloggerimpl.NewDisabledParams()` with `jmxloggerimpl.NewCliParams("")` to properly initialize the JMXLogger component.

### Motivation

When running `agent check` on JMX integrations with `DD_LOG_LEVEL=debug`, the command would crash with a panic:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x7bd3b9]

goroutine 256 [running]:
github.com/DataDog/datadog-agent/pkg/util/log.(*loggerPointer).error(0x40?, {0xc0006e9dc0, 0x1d})
        github.com/DataDog/datadog-agent/pkg/util/log/log.go:516 +0xb9
github.com/DataDog/datadog-agent/pkg/util/log.logWithError(0x4, 0xc000732fa0, 0xc000a5deb8, 0x1, {0xc000dbf6b0, 0x1, 0x1})
        github.com/DataDog/datadog-agent/pkg/util/log/log.go:298 +0x11d
github.com/DataDog/datadog-agent/pkg/util/log.JMXError({0xc000dbf6b0, 0x1, 0x1})
        github.com/DataDog/datadog-agent/pkg/util/log/log.go:926 +0x99
github.com/DataDog/datadog-agent/comp/agent/jmxlogger/jmxloggerimpl.logger.JMXError(...)
        github.com/DataDog/datadog-agent/comp/agent/jmxlogger/jmxloggerimpl/jmxlogger.go:85
```

The root cause was that `NewDisabledParams()` creates a disabled logger that doesn't properly initialize the underlying logger. When JMXFetch attempts to log any messages during execution, it triggers a nil pointer dereference.

This issue only manifested when:
- Running JMX checks via `agent check` (not `agent jmx`)
- With logging enabled (any level except "off")
- When JMXFetch actually needs to log something

**Manual Testing:**
1. **Before fix**: Ran `docker exec -e DD_LOG_LEVEL=debug dd_kafka_py3.12-2.8 agent check kafka` - resulted in panic
2. **After fix**: Same command now executes successfully and produces expected JSON output
3. **Regression test**: Verified that `DD_LOG_LEVEL=off` continues to work as before

### Additional Notes

- The `agent check` command for JMX integrations was deprecated over 3 years ago in favor of the `agent jmx` command, making the impact very narrow
- We can safely use an empty string for the log file parameter of `jmxloggerimpl.NewCliParams` as this disables file logging while maintaining proper logger initialization


pgimalac (commented):
Should the proper fix be to update `comp/agent/jmxlogger/jmxloggerimpl/jmxlogger.go` so that when `deps.Params.disabled` is true we initialize a logger with OFF level ?
Otherwise I think any use of `jmxloggerimpl.NewDisabledParams()` would lead to this crash ?

misteriaud:
@pgimalac indeed, since no other place use a disabled jmxLogger, I removed the option. This is easier!

pgimalac (approved):

pkg/cli/subcommands/check/command.go
```diff
@@ -202,7 +202,7 @@ func MakeCommand(globalParamsGetter func() GlobalParams) *cobra.Command {
 				fx.Supply(option.None[integrations.Component]()),
 
 				getPlatformModules(),
-				jmxloggerimpl.Module(jmxloggerimpl.NewDisabledParams()),
+				jmxloggerimpl.Module(jmxloggerimpl.NewCliParams("")),
```
pgimalac, line 205: Is the created logger OFF by default ? (ie. is this equivalent to the previous behavior ?)

vickenty: Just in case, there is also code above that sets env var from a cli argument, might make sense to pass it here (not sure).

vickenty (approved):

pkg/cli/subcommands/check/command.go
misteriaud: Yes the logger is still set to "off" by default (using `LogParams: log.ForOneShot(globalParams.LoggerName, "off", true),` a bit above in the file).
And then, the `log_level` value is forwarded to JMXFetch, so if "off" is used, JMXFetch should not print anything.

misteriaud: > Just in case, there is also code above that sets env var from a cli argument, might make sense to pass it here (not sure).

`jmxloggerimpl.NewCliParams` receives the log filepath as argument, which env variable did you had in mind ?

vickenty: I misunderstood the parameter function, sorry for the noise.

GustavoCaso (commented):
Amazing work. 

Looking at the changes it seems that having a separate implementation of a noop jmx logger, could help with this use case. 

It seems that jmxlogger it selves deals with some globals and that might be more work to remove, but having a noop implementation it would make the code clearer that the check subcommands does not use the jmx logger component

misteriaud:
Thanks @GustavoCaso for the feedback ! I think we still want to have a logger for JMX in this command, having a noop would just hide the problem don't you think ?

PR #39851, opened 2025-08-14
Retry SSH commands on failure

usamasaqib:
### What does this PR do?
This PR tries to address SSH connection instability issues in KMT by retry SSH commands on failure

### Additional Notes

QA automatically performed when jobs run successfully in CI.


test/new-e2e/system-probe/connector/main.go
```diff
@@ -150,9 +154,16 @@ func run() (err error) {
...
-		return fmt.Errorf("connect: %s", err)
+
+	if err := backoff.Retry(func() error {
+		if err := communicator.Connect(ctx); err != nil {
+			failType = failConnect
+			return fmt.Errorf("connect: %s", err)
+		}
+
+		return nil
+	}, backoff.WithMaxRetries(backoff.NewConstantBackOff(sshRetryInterval), sshMaxRetries)); err != nil {
```
val06, line 165: do we want the constant backoff here? the 2s interval can be too little in some edge cases, no?

usamasaqib: Given we retry 10 times the total interval would be 20 seconds. So we give some time for the VM to become responsive for the conncetion.

> can be too little in some edge cases, no

Note that this is just retrying the ssh connection, nothing else. Retrying is not the appropriate solution for long lived commands which the `connector` is mainly used for.

val06 (approved):

PR #39854, opened 2025-08-14
Using the same owner for e2e than the code for the Agent CLI

hush-hush:
### What does this PR do?

Using the same owner for e2e than the code for the Agent CLI.


pgimalac (approved):
Thanks !

PR #39856, opened 2025-08-14
Skip binary validation when we use the internal SGC

hush-hush:
### What does this PR do?

Skip binary validation when we use the internal SGC.

This will require manual QA from Agent-Configuration.

### DIFF
--- a/comp/core/secrets/impl/fetch_secret.go
+++ b/comp/core/secrets/impl/fetch_secret.go
@@ -48,8 +48,10 @@ func (r *secretResolver) execCommand(inputPayload string) ([]byte, error) {
 	}
 	defer done()
 
-	if err := checkRights(cmd.Path, r.commandAllowGroupExec); err != nil {
-		return nil, err
+	if !r.embeddedBackendPermissiveRights {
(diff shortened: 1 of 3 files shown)


comp/core/secrets/impl/fetch_secret.go
```diff
@@ -48,8 +48,10 @@ func (r *secretResolver) execCommand(inputPayload string) ([]byte, error) {
 	}
 	defer done()
 
-	if err := checkRights(cmd.Path, r.commandAllowGroupExec); err != nil {
-		return nil, err
+	if !r.embeddedBackendUsed {
```
dustmop: Do we want to add the same skip for [this](https://github.com/DataDog/datadog-agent/blob/64b105145ac8c925c5060bd92624f9f903d7e30e/comp/core/secrets/impl/status.go#L47), maybe making the message "permission check skipped"? It would allow the `agent status` command to accurately report the status.

hush-hush: ho yes, good catch.

dustmop (approved):

PR #39866, opened 2025-08-14
Upload code coverage after e2e test execution

KevinFairise2:
### What does this PR do?

Update the test suite so that code coverage is generated per test suite.
Add an after_script test to retrieve that coverage folder, convert it into `textfmt` that is compatible with Codecov.
Upload the coverage to Codecov so that it can be visualized.
The visualization can be done using the flags, for example for **new-e2e-agent-configuration**: https://app.codecov.io/gh/DataDog/datadog-agent/tree/kfairise%2Ftest-modifications

### Motivation

We now compute the coverage in order to use it to dynamically know which test should be executed, it costs nothing to upload it on codecov so we have a way to easily visualize what parts of the code are covered by e2e tests

### DIFF
--- a/tasks/coverage.py
+++ b/tasks/coverage.py
@@ -98,18 +98,6 @@ powershell.exe -executionpolicy Bypass -file {GO_COV_TEST_PATH}.ps1 %*"""
                     os.remove(f)
 
 
-@task
-def convert_coverage_folder_to_txt(
-    ctx: Context,
-    coverage_folder: str,
-    coverage_file: str,
-):
-    """
-    Convert the coverage folder to a txt file.
-    """
-    ctx.run(f"go tool covdata textfmt -i={coverage_folder} -o={coverage_file}")
-
-
 @task(iterable=['extra_tag'])
 def upload_to_codecov(
     ctx: Context,
--- a/tasks/coverage.py
+++ b/tasks/coverage.py
@@ -1,3 +1,4 @@
+import json
 import os
 import pathlib
 import platform
@@ -9,6 +10,7 @@ from invoke import Context, task
 from invoke.exceptions import Exit
 
 from tasks.libs.common.color import Color, color_message
+from tasks.libs.common.coverage import upload_codecov
 from tasks.libs.common.git import get_commit_sha, get_main_parent_commit
 from tasks.libs.common.utils import get_distro, gitlab_section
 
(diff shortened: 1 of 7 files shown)


tasks/libs/common/coverage.py
```diff
@@ -0,0 +1,18 @@
+import platform
+
+from invoke import Context
+
+from tasks.libs.common.utils import get_distro
+
+
+def upload_codecov(ctx: Context, coverage_file: str, extra_tag: list[str]):
```
rdesgroppes, line 8: isn't it a dup of `upload_to_codecov` in `/tasks/coverage.py`?

chouetz: No, it's extracting a part of it for re-use, in `upload_to_codecov` and in `process_e2e_coverage_folders`

tasks/unit_tests/libs/common/coverage_test.py
chouetz, line 1: 👏🏽

tasks/unit_tests/libs/common/coverage_test.py
```diff
@@ -0,0 +1,202 @@
...
+        upload_codecov(self.mock_context, "coverage.out", ["", "valid-tag", ""])
+
+        # Verify the correct command was executed (empty strings should be included)
+        self.mock_context.run.assert_called_once_with(
+            "codecov -f coverage.out -F ubuntu -F  -F valid-tag -F ", warn=True, timeout=2 * 60
+        )
+
+
+if __name__ == '__main__':
+    unittest.main()
```
chouetz:
```
```
As this is called with the `unittest` module, you don't need to make it an executable

tasks/coverage.py
```diff
@@ -121,19 +125,85 @@ def upload_to_codecov(
...
+        coverage_folder = folder / "coverage"
+        metadata_file = folder / "metadata.json"
+
+        if not coverage_folder.exists():
+            print(color_message(f"No coverage folder found in {folder.name}, skipping", Color.ORANGE))
+            continue
+
+        if not metadata_file.exists():
+            print(color_message(f"No metadata.json found in {folder.name}, skipping", Color.ORANGE))
+            continue
```
chouetz, line 174: maybe all this defensive coding could be handled with a single `try/except` for readability?

chouetz (approved):
Few minor remarks

tasks/libs/common/coverage.py
rdesgroppes: Oh yes, sorry. Perfect, then, thanks!

rdesgroppes (approved):

tasks/coverage.py
KevinFairise2: Hmm I am not sure to see how I could make it more readable with a single `try/except` 
It would be harder to know which one is missing and the error about non-existant coverage folder would be raised by the `go tool covdata` command that is called through `ctx.run` which does not make it easy to know it failed because the file is not there

tasks/unit_tests/libs/common/coverage_test.py
KevinFairise2: removed

PR #39871, opened 2025-08-14
[NDINT-290] Collect Interface Metrics from Versa Analytics

ken-schneider:
### What does this PR do?
Updates the behavior of `collect_interface_metrics` to collect interfaces from Versa Analytics rather than the director. The previous director based metrics are now `collect_director_interface_metrics`.

Analytics exposes the following for interfaces:
`RxUtil` -> Inbound Utilization Percent
`TxUtil` -> Outbound Utilization Percent
`VolumeRx` -> Inbound Volume
`VolumeTx` -> Outbound Volume
`Volume` -> Total Volume
`BandwidthRx` -> Inbound Bandwidth (Throughput)
`BandwidthTx` -> Outbound Bandwidth (Throughput)
`Bandwidth` -> Total Bandwidth (Throughput)

These metrics are tagged with `site`, `access_circuit`, and `interface`.

I've also played around with the idea of creating interface metadata from this call, however, since `status` is not exposed in this table and I haven't found one where it is exposed, I've commented out the part of `versa.go` where it could be enabled.

### Motivation
The previous interface metrics were missing Utilization. Additionally, they required us to make a call per appliance to gather the metrics. The updated Analytics call should be able to return all interfaces across appliances in a paginated fashion.

Tested against our own internal Versa environment


pkg/collector/corechecks/network-devices/versa/client/client.go
```diff
@@ -627,6 +627,34 @@ func (client *Client) GetDIAMetrics(tenant string) ([]DIAMetrics, error) {
...
+	if tenant == "" {
+		return nil, fmt.Errorf("tenant cannot be empty")
+	}
+
+	return getPaginatedAnalytics(
+		client,
+		tenant,
+		"SDWAN",
+		client.lookback,
+		"intfUtil(site,accCkt,intf)",
```
zoedt, line 641: great find !!

zoedt (approved):
this overall lgtm !

it's unfortunate constraint w/ director vs. analytics, but as long as we document it before the TODO you mentioned gets put into motion - i think it's ok to live with until we have the resources/time to move forward !

also just remembering now we're approaching final metrics in the check - i forget if we need to define anything to help w/ dashboarding / monitors, etc. with a `metadata.csv`?

ken-schneider:
Good call out, this is going to go in integrations-core I believe. AFAIK there's no way to do this in this repo today.

PR #39874, opened 2025-08-15
chore(fleet): Make OCI agent path-independent

BaptisteFoy:
### What does this PR do?
Adds a relative rpath to the OCI agent. This works around the security-execution mode issues because of:
- https://github.com/DataDog/datadog-agent/pull/39721 statically building the installer, fixing `seteid` issues when the daemon calls the installer
- https://github.com/DataDog/datadog-agent/pull/39798 adding the `CAP_NET_BIND_SERVICE`, thus removing the only use case of customers

E2Es are enough


coignetp (approved):

rdesgroppes (approved):
LGTM

PR #39877, opened 2025-08-15
Refactor `RemoteAgentRegistry` configstream code to a new `configstream` component

rayz:
### What does this PR do?
This pr creates a new `configstream` component that is `Subscribe`d to via gRPC. The implementation logic was copied over from the `RemoteAgentRegistry` with a few changes since subscribing to configuration updates no longer relies on `RemoteAgentDetails`.

### Motivation
We would like to move the existing config stream gRPC to live on the `SecureAgent` server

This allows us to establish the config stream before a registration is done in the `RemoteAgentRegistry`.

Tested with ADP ([pr](https://github.com/DataDog/saluki/pull/869)) 

This should have no effect on remote agents since ADP is the only remote agent right now.


.github/CODEOWNERS
```diff
@@ -327,6 +327,7 @@
 /comp/checks/winregistry @DataDog/windows-products
 /comp/core/autodiscovery @DataDog/container-platform
 /comp/core/config @DataDog/agent-configuration
+/comp/core/configstream @DataDog/agent-metric-pipelines
```
hush-hush: Should we, `agent-configuration`, co-own this ?

comp/core/configstream/impl/configstream.go
```diff
@@ -0,0 +1,279 @@
...
+				SequenceId: int32(sequenceID),
+				Settings:   settings,
+			},
+		},
+	}
+
+	return snapshot, sequenceID, nil
+}
+
+func sanitizeValue(value interface{}) (interface{}, error) {
```
hush-hush, line 274: Could we add a comment on why we need this ?

comp/core/configstream/mock/mock.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"testing"
+
+	configstream "github.com/DataDog/datadog-agent/comp/core/configstream/def"
+)
+
+// Mock returns a mock for configstream component.
+func Mock(_ *testing.T) configstream.Component {
+	return nil
+}
```
hush-hush: Would this not panic if someone where to call `Mock(t).Subscribe()` ?

hush-hush (changes requested):

.github/CODEOWNERS
tobz: Yes, definitely.

comp/core/configstream/mock/mock.go
rayz: Thanks for the catch. I've updated the mock to have an implementation.

comp/core/configstream/impl/configstream.go
```diff
@@ -0,0 +1,284 @@
...
+		if currentSequenceID > sub.lastSequenceID+1 {
+			// Lazily create the snapshot, only if it's needed for the first time in this update cycle.
+			if snapshot == nil {
+				snapshot, snapshotSeqID, snapshotErr = cs.createConfigSnapshot()
+				if snapshotErr != nil {
+					cs.log.Errorf("Failed to create resynchronization snapshot, all out-of-sync subscribers will remain so until the next update: %v", snapshotErr)
+					continue // A snapshot could not be created, but other subscribers may still be able to process the incremental update.
+				}
+			}
+			cs.log.Infof("Discontinuity detected for subscriber '%s'. Last seen ID: %d, current ID: %d. Resynchronizing with a snapshot.", id, sub.lastSequenceID, currentSequenceID)
```
tobz: I'd _probably_ make this a warning.

My thought here is that we expect this to be rare, so it'd be good to surface this up at the warning level so we don't miss it.

tobz (approved):
Overall, this looks good to me. 👍🏻

hush-hush (approved):

comp/core/configstream/impl/configstream.go
```diff
@@ -0,0 +1,284 @@
...
+		case id := <-cs.unsubscribeChan:
+			cs.removeSubscriber(id)
+		case update := <-updatesChan:
+			cs.handleConfigUpdate(update)
+		case <-cs.stopChan:
+			cs.m.Lock()
+			for _, sub := range cs.subscribers {
+				close(sub.ch)
+			}
+			cs.subscribers = make(map[string]*subscription)
```
misteriaud, line 148: **❓ question:** ‏Why is it needed to recreate a an empty map before stopping the component ?

misteriaud (approved):
Great work ! 🎉  I just left a small question but everything looks good to me !

comp/core/configstream/impl/configstream.go
rayz: It's probably not needed but I added it to clear the map just in case.

pkg/config/setup/config.go
```diff
@@ -1111,7 +1111,6 @@ func InitConfig(config pkgconfigmodel.Setup) {
 	config.BindEnvAndSetDefault("remote_agent_registry.idle_timeout", time.Duration(30*time.Second))
 	config.BindEnvAndSetDefault("remote_agent_registry.query_timeout", time.Duration(3*time.Second))
 	config.BindEnvAndSetDefault("remote_agent_registry.recommended_refresh_interval", time.Duration(10*time.Second))
-	config.BindEnvAndSetDefault("remote_agent_registry.config_stream_retry_interval", time.Duration(1*time.Second))
```
GustavoCaso, line 1161: Are we not adding the same interval to the new component? Previously, we tried until the initial snapshot was successfully sent 😄 

Now it seems that we error

rayz: We no longer have a separate go routine for each subscriber (could be done but there is added complexity making sure a closed stream isn't sent on after an unsubscribe), so re-syncing on the next update is best vs a sleep interval that could potentially block other subscribers.

GustavoCaso: > so re-syncing on the next update is best vs a sleep interval that could potentially block other subscribers.

But if we fail to send an event we error:
```go
if err := stream.Send(event); err != nil {
    log.Warnf("Failed to send config event to client: %v", err)
    return err
}
```

Where is the re-syncing happening? Does the remote agent need to reconnect?

rayz: Sorry, to be clear re-sync here means getting sent a snapshot if the subscribers sequenceID is not up to date (in `handleConfigUpdate`).  If the stream is closed for whatever reason, the remote-agent / whoever is using `configStream` would have to resubscribe.

I've hardcoded a `1` second sleep for now on on `StreamConfigEvents`. (Would have to add a method to the interface to get the retry interval otherwise).

GustavoCaso: Should we add testing that ensures that we respect the retry logic? 

> Would have to add a method to the interface to get the retry interval otherwise

You could pass the configuration component to the server `func NewServer(comp configstream.Component) *Server {` and use that configured value instead of hardcoding it 😄

GustavoCaso (changes requested):
LGTM. Still I would add testing around the retry (if possible) and avoid hardcoding the sleep time 😄

comp/core/configstream/server/server_test.go
```diff
@@ -0,0 +1,145 @@
...
+	cfg.Set("config_stream.sleep_interval", 10*time.Millisecond, model.SourceAgentRuntime)
+
+	comp := &mockComp{}
+	stream := &mockStream{ctx: ctx}
+
+	server := NewServer(cfg, comp)
+	eventsCh := make(chan *pb.ConfigEvent, 1)
+
+	return server, comp, stream, eventsCh
+}
```
GustavoCaso, line 68: Missing new line 😄

GustavoCaso (approved):
LGTM 🎉 

Awesome work. Thanks for applying the suggested changes

PR #39878, opened 2025-08-15
[NDINT-362] OAuth Implementation

ken-schneider:
### What does this PR do?
Implements OAuth for Versa Director API calls allowing the user to choose the auth type and supply a client ID/secret for OAuth.

This does not use OAuth for Analytics calls. Since we can't test that with a managed cluster, I'm going to separate that into another PR, but I think it should be possible for folks that manage their own cluster.

This also cleans up a lot of the client configuration and abstracts it away from the main `versa.go` file so it needs to know less about the underlying client implementation.

The user can now specify an authentication method in their config:
```yaml
instances:
- director_endpoint: <director_endpoint>
  ...
  auth_method: oauth # default is basic
  client_id: <oauth_client_id>
  client_secret: <oauth_client_secret>
```

The options are `basic` and `oauth` with a default of `basic`. Depending on early feedback on the performance of OAuth, that should be updated to be the default in a future PR.

### Motivation
Enable users of the Versa integration to choose OAuth instead of Basic auth

Tested locally against our own Versa cluster


zoedt (approved):
🎉 lgtm !

PR #39890, opened 2025-08-15
Enable Test Singing For Windows Driver Testing

jack0x2:
### What does this PR do?
This PR checks `release.json`, if the windows drivers are test signed, it enables test-signing on all of the windows hosts. This allows us to kick off agent pipelines with test-signed windows drivers.

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1731

Adds new e2e functionality, see test pipeline here: [Test Fixed DDNPM](https://github.com/DataDog/datadog-agent/pull/39891)

### DIFF
--- a/test/new-e2e/pkg/provisioners/aws/host/windows/host.go
+++ b/test/new-e2e/pkg/provisioners/aws/host/windows/host.go
@@ -8,6 +8,7 @@ package winawshost
 
 import (
 	"fmt"
+	sysos "os"
 
 	"github.com/DataDog/test-infra-definitions/components/activedirectory"
 	"github.com/DataDog/test-infra-definitions/components/datadog/agent"
@@ -26,6 +27,7 @@ import (
 	"github.com/DataDog/datadog-agent/test/new-e2e/pkg/utils/optional"
 	"github.com/DataDog/datadog-agent/test/new-e2e/tests/windows/components/defender"
 	"github.com/DataDog/datadog-agent/test/new-e2e/tests/windows/components/fipsmode"
+	"github.com/DataDog/datadog-agent/test/new-e2e/tests/windows/components/testsigning"
 )
 
 const (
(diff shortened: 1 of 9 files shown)


test/new-e2e/pkg/provisioners/aws/host/windows/host.go
```diff
@@ -282,12 +307,17 @@ func getProvisionerParams(opts ...ProvisionerOption) *ProvisionerParams {
...
 		// We ALWAYS need to make a deep copy of `params`, as the provisioner can be called multiple times.
 		// and it's easy to forget about it, leading to hard to debug issues.
-		params := getProvisionerParams(opts...)
-		return Run(ctx, env, params)
+		params := GetProvisionerParams(opts...)
+
+		awsEnv, err := aws.NewEnvironment(ctx)
+		if err != nil {
+			return err
+		}
```
L3n41c, line 319: Just out of curiosity, why did we have to extract the `aws.NewEnvironment(ctx)` from the `Run(…)` function?

L3n41c (approved):

test/new-e2e/pkg/provisioners/aws/host/windows/host.go
clarkb7: looks like it is for the test below, which needs to add another component to the environment. It calls this `Run()` method to initialize the windows host env, and uses a single awsEnv for each. I have no idea if pulumi requires the same awsenv be used for all resources though.

clarkb7 (approved):

test/new-e2e/pkg/provisioners/aws/host/windows/host.go
jack0x2: Yeah it fails if you pass a new one in with pulumi stack name conflict.

PR #39892, opened 2025-08-15
Proof of concept of user customizable agent devcontainer setup.

aiuto:
### What does this PR do?

`We introduce the notion of `$HOME/.devcontainer/agent_overrides.json`.

If it exists, we load it and add to the defaults done by `inv devcontainer.setup`.
It's useful enough just to make it easy to permanently get some things you want mounted always appear in the container.
The bit about appending the createCommand is dubious.

Sample:
- mounts $HOME/ws and $HOME/SETUP into the devcontainer
- shows running an additional setup. Typically, that would be to rsync dotfiles from SETUP into $HOME in the container.
```
 $HOME/.devcontainer/agent_overrides.json
{
    "mounts": [
        "source=${localEnv:HOME}/ws,target=/home/datadog/ws,type=bind,consistency=cached",
        "source=${localEnv:HOME}/SETUP,target=/home/datadog/SETUP,type=bind,consistency=cached"
    ],
    "onCreate": [
        "echo Hello Tony"
    ]
}
```

### Motivation

Trying to get variations of the dev container suitable for different development tasks.

But there is so much more we could do. What I would also like to have is naming from the command line
```
agent_devcontainer create -name dev_normal
agent_devcontainer create -name dev_with_foo --extramount $HOME/foo:/home/datadog/foo
agent_devcontainer create -name dev_cursor --extramount $HOME/my_prompts:/home/datadog/my_prompts
```

Since that is the goal, I don't like the fixed path to `agent_overrides.json`. That should also be a command line parameter.  But I haven't figure out how to add command line args to invoke tasks yet.

### Meta comments

It feels like this command should be external to the agent code because it is not bound to a particular agent branch, it is more tied to the current container image we build.  And when doing that, we should generate devcontainer.json outside of the github workspace, and use the devcontainer facilities to set up the mount correctly.  That would make it cleaner to have several docker images, each mapping in a single github/datadog-agent repo.


chouetz (approved):
LGTM, you just need to lint your code :)

PR #39893, opened 2025-08-15
[OTAGENT-480] Add host metadata pusher to DDOT

songy23:
### What does this PR do?
Add host metadata pusher to DDOT

### Motivation
this is required to support Gateway in DDOT: https://docs.google.com/document/d/18LJGW1HZdFcu-kCuwX11-KapWRLGktIu_xXocp0XIYE/edit?pli=1&tab=t.0

new & modified tests


janine-c (approved):

comp/otelcol/otlp/components/exporter/datadogexporter/factory.go
```diff
@@ -40,6 +41,10 @@ type factory struct {
 	setupErr               error
 	onceSetupTraceAgentCmp sync.Once
 
+	onceReporter sync.Once
+	reporter     *inframetadata.Reporter
+	reporterErr  error
```
liustanley, line 46: Does `reporterErr` need to be on the factory struct or can we just make a local variable in `Reporter()`?

comp/otelcol/otlp/components/exporter/datadogexporter/factory.go
```diff
@@ -174,11 +179,18 @@ func (f *factory) createTracesExporter(
 	}
 	f.mclientwrapper.SetDelegate(otelmclient)
 
+	if cfg.HostMetadata.Enabled {
+		_, err = f.Reporter(set, cfg.HostMetadata.ReporterPeriod)
+		if err != nil {
+			return nil, fmt.Errorf("failed to build host metadata reporter: %w", err)
+		}
+	}
```
liustanley: Could we abstract this logic to a function if it's not too difficult?

liustanley (approved):
LGTM, just two small comments

comp/otelcol/otlp/components/exporter/datadogexporter/factory.go
songy23: Yes it needs to be stored so that `f.Reporter` can be safely called multiple times but return the same result. The idea is whenever the first one of [traces, metrics, logs] exporter gets created, we build the reporter (exactly once), and later when other exporters are created, they reuse the reporter

comp/otelcol/otlp/components/exporter/datadogexporter/factory.go
songy23: sure can simplify it a bit

PR #39894, opened 2025-08-15
dyninst/symdb: teach CLI to upload symbols to backend

andreimatei:
This should be useful for manual testing, but also for dogfooding where
we don't have a new-enough agent or dd-trace-go.

The upload comes from either a local file or a container image. For the
latter, you can now give the tool the name of an image from
registry.ddbuild.io and it will look for a file with the same name
inside the image's /usr/local/bin. This should make it easy to upload
symbols for services running in production by clicking through the
Datadog UI to find these image names.


hush-hush (approved):
Looks good for agent-configuration.

pkg/dyninst/uploader/config.go
```diff
@@ -84,3 +88,11 @@ func WithMaxBufferDuration(d time.Duration) Option {
 		c.batcherConfig.maxBufferDuration = d
 	}
 }
+
+// WithHeader adds a header key-value pair to all upload HTTP requests. The
+// option can be used repeatedly to add multiple headers.
+func WithHeader(key, val string) Option {
+	return func(c *config) {
+		c.headers = append(c.headers, [2]string{key, val})
+	}
```
GreenMatan: nit: it seems this function isn't use anywhere (?)

GreenMatan (approved):
LGTM

pkg/dyninst/uploader/config.go
andreimatei: removed

PR #39535, opened 2025-08-05
Remove usage of the old deb_* images to use the unified linux one

KevinFairise2:
### What does this PR do?

Remove the remaining usages of the old deb_* images to fully migrate to the new unified image for Linux jobs

PR #39896, opened 2025-08-16
Package IoT Agent using the standard Linux image

ofek:
### Motivation

Follow-up to https://github.com/DataDog/datadog-agent/pull/39535

Commit [`267d341` (#39535)](https://github.com/DataDog/datadog-agent/pull/39535/commits/267d3418f3fb824e55dcc7ff9ff5cbbdc57b14b2) retained the armhf image by merely added back the environment variable referencing the image version. So the build was failing due to a missing reference rather than incompatibility, as was previously thought, because the new image was never actually tested.

### Additional Notes

Depends on this update to the image https://github.com/DataDog/datadog-agent-buildimages/pull/964


CelianR (approved):

alopezz (changes requested):
I don't think this works, because it builds the [wrong artifact](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1083319725#L206) (arm64 instead of armhf). Also seen from the fact that the static quality gates job [fails to find the expected artifact](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1083320043#L491).

Now the name of the package is just due to how omnibus auto-detects architecture. I described this on the "Additional Notes" on the description of #39840.

> This architecture tag in package names is something that Omnibus decides based on the environment.
> - For deb, [it relies on dpkg](https://github.com/DataDog/omnibus-ruby/blob/c760785c93689c06da59e5f38875cdeda39438c4/lib/omnibus/packagers/deb.rb#L619). That's why deb packages were being tagged correctly.
> - For rpm, [it relies on ohai](https://github.com/DataDog/omnibus-ruby/blob/c760785c93689c06da59e5f38875cdeda39438c4/lib/omnibus/packagers/rpm.rb#L719C12-L723), which ultimately relies on uname -m, based on the source code [here](https://github.com/chef/ohai/blob/7462c000c9b11851945dd388956490103dda7e9b/lib/ohai/plugins/kernel.rb#L32). Therefore, this fakearmv7l trick is really only targeting this currently.

So we either have to:
- Modify omnibus to be able to force the architecture and sidestepping autodetection.
- Find some means to make autodetection point to the architecture we want.

.gitlab/packaging/deb.yml
```diff
@@ -166,14 +166,19 @@ iot_agent_deb-arm64:
...
   needs: ["iot-agent-armhf"]
   variables:
     PACKAGE_ARCH: armhf
     DD_PKG_ARCH: "arm64"
     FORCED_PACKAGE_COMPRESSION_LEVEL: 5
     STATIC_QUALITY_GATE_NAME: "static_quality_gate_iot_agent_deb_armhf"
+  script:
+    - !reference [.cache_omnibus_ruby_deps, setup]
+    - !reference [.setup_deb_signing_key]
+    - PATH="/opt/dpkg-armhf/bin:$PATH" dda inv -- -e omnibus.build --base-dir $OMNIBUS_BASE_DIR --skip-deps --flavor iot
```
alopezz, line 181: Adding a comment on top of this line to explain why we need to set PATH like this here could help future readers.

alopezz (approved):

.gitlab/packaging/deb.yml
ofek, line 181:
```suggestion
    # Use custom built dpkg for armhf packaging since we're on armv8/arm64
    - PATH="/opt/dpkg-armhf/bin:$PATH" dda inv -- -e omnibus.build --base-dir $OMNIBUS_BASE_DIR --skip-deps --flavor iot
```

ofek: Added, thanks!

PR #39915, opened 2025-08-18
RLS-20: Enable K8s SSI gradual rollout support in DCA

erikayasuda:
### What does this PR do?
The goal of this PR is to enable gradual rollout support for K8s SSI users. The breakdown of the changes is as follows:

**The `ImageResolver`**
The `ImageResolver` will be initialized in the `start` command for DCA, and use the existing remote config client (if remote config is enabled) to reference `K8S_INJECTION_DD`. `K8S_INJECTION_DD` is a DD managed config that is split up by repository (`apm-inject` and `dd-lib-[lang]-init`), and for each repo mutable tags (`latest`, `3`, `3.1`, etc...) are mapped to image digests (see [here](https://github.com/DataDog/rc-employee-configurations/tree/main/configs/K8S_INJECTION_DD/generated_versions)). The `ImageResolver` will use this mapping to determine which version of the tracer library and the injector to use (image reference with specific image digest like `gcr.io/datadoghq/dd-lib-python-init@sha123`). If remote config is not enabled, it will default to the current behavior (image reference with a mutable tag like `gcr.io/datadoghq/dd-lib-python-init:3`). DD will control the updates and deployments of `K8S_INJECTION_DD`, and gradually release the new mappings to additional subsets of organizations.

imageresolver

**Configuring DCA to support gradual rollout**
- **TL;DR** - We need to initialize an `ImageResolver` in the `start` command for the DCA, and pass the `ImageResolver` down to the `MutatorCore` as well as the `language.libInfoWithResolver(...)` method.

**Mutating the pod spec**
- **TL;DR** - We need to pass the `ImageResolver` from the `MutatorCore` to the `language.libInfoWithResolver(...)` method again. Since the `MutatorCore` already has its `ImageResolver` at this point, it can directly reference it when calling `MutatorCore.newInjector(...)` when initializing a new injector instance.

### Motivation
Currently, new images are released to SSI customers all at once. This means we risk causing significant impact on customers with faulty releases, and we have no way to slowly monitor a release and how it does on customer applications. This change will make the gradual rollout possible, as well as give Datadog the ability to halt a release from spreading if it's deemed problematic.

- Local E2E testing via `injector-dev`
- Added new unit tests for `ImageResolver` functionality
- Updated pre-existing tests pertaining to image references for the apm-injector and our tracer library images.

- One known trade-off with this gradual rollout plan is that this will inevitably lengthen the release process for SSI artifacts. We will need to let rollout phases happen over a non-trivial amount of time to ensure we have enough signal to proceed with the rollout (over a few days to a week), whereas current state it is immediate (same day). This is an acceptable trade-off, as it provides us the confidence to have more customers use our releases, and allows us to be more proactive instead of reactive to possible widespread issues.

### Additional Notes

- We will be monitoring injection success/error rates as this goes along. While success rates are deemed healthy, the rollout will continue. If at any point we have concerns, the `K8S_INJECTION_DD` rollout will be stopped and new organizations and new pods will not be able to pull the faulty image. Important to call out here that this is just "stopping the bleeding" and will not proactively revert any image versions on existing pods.


pkg/clusteragent/admission/mutate/autoinstrumentation/tag_resolver.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// NewTagResolver creates a new TagResolver instance.
+func NewTagResolver(rcClient *rcclient.Client) *TagResolver {
+	tr := &TagResolver{
+		rcClient:    rcClient,
+		tagMappings: make(map[string]map[string]ResolvedImage),
+	}
+	if rcClient != nil {
+		rcClient.Subscribe(state.ProductGradualRollout, tr.processUpdate)
```
erikayasuda: Double check if this triggers on changes only, or if it will always run callback on poll cadence

cmd/cluster-agent/subcommands/start/command.go
```diff
@@ -395,6 +396,9 @@ func start(log log.Component,
 		if config.GetBool("autoscaling.workload.enabled") {
 			products = append(products, state.ProductContainerAutoscalingSettings, state.ProductContainerAutoscalingValues)
 		}
+		if config.GetBool("admission_controller.auto_instrumentation.enabled") {
```
betterengineering: Note: This works well! It controls if the webhook receives requests or not. There is another one that enables pod mutation that you should be aware of:
```
apm_config.instrumentation.enabled
```

betterengineering: To test this, you can add this environment variable:
```
DD_ADMISSION_CONTROLLER_AUTO_INSTRUMENTATION_ENABLED
```
You can set that via this config: https://docs.datadoghq.com/containers/cluster_agent/commands/?tab=datadogoperator#cluster-agent-environment-variables

pkg/clusteragent/admission/controllers/webhook/controller_base.go
```diff
@@ -182,8 +184,8 @@ func generateTagsFromLabelsWebhook(wmeta workloadmeta.Component, datadogConfig c
 	return tagsfromlabels.NewWebhook(wmeta, datadogConfig, mutator), nil
 }
 
-func generateAutoInstrumentationWebhook(wmeta workloadmeta.Component, datadogConfig config.Component) (*autoinstrumentation.Webhook, error) {
-	config, err := autoinstrumentation.NewConfig(datadogConfig)
+func generateAutoInstrumentationWebhook(wmeta workloadmeta.Component, datadogConfig config.Component, rcClient *rcclient.Client) (*autoinstrumentation.Webhook, error) {
+	config, err := autoinstrumentation.NewConfigWithRemoteConfigClient(datadogConfig, rcClient)
```
betterengineering: Instead of passing this into config, it might make more sense to pass it directly into `NewMutatorWithFilter`. The rc client is a dependency and is not config

pkg/clusteragent/admission/mutate/autoinstrumentation/tag_resolver.go
```diff
@@ -0,0 +1,177 @@
...
+// not just the changed ones. This is the correct behavior for remote config:
+//
+// - Remote config maintains a complete state of all active configurations
+// - When processUpdate is called, it receives the complete current state via GetConfigs()
+// - If a repository is not in the update, it means it's no longer active
+// - Therefore, replacing the entire cache ensures we stay in sync with remote config
+func (tr *TagResolver) processUpdate(update map[string]state.RawConfig, applyStateCallback func(string, state.ApplyStatus)) {
+	/*
+		Example of processed cache:
+		{
```
betterengineering: Might be good to put this in a unit test! You can use `testdata/` to store the file and load it up in your test. See `target_mutator_test` or `config_test` for good examples of this

pkg/clusteragent/admission/mutate/autoinstrumentation/tag_resolver.go
```diff
@@ -0,0 +1,177 @@
...
+				"images": {
+					"latest": {
+						"digest": "sha256:abc123...",
+						"canonical_version": "2.1.3"
+					}
+				}
+			}
+		}
+	*/
+	tr.mu.Lock()
```
betterengineering: You could move these locks down to when you replace the cache object

pkg/clusteragent/admission/mutate/autoinstrumentation/tag_resolver.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	resolved, exists := repoCache[tag]
+	if !exists {
+		log.Debugf("TagResolver: No mapping found for %s:%s", repositoryName, tag)
+		return ""
+	}
+
+	log.Debugf("TagResolver: Resolved %s:%s -> %s", repositoryName, tag, resolved.FullImageRef)
+	return resolved.FullImageRef
```
betterengineering: This is a good change! It's worth noting that the customer will now see a hash instead of the version. In terms of safety that is a good trade off. It might be worth adding an annotation for the canonical version.

pkg/clusteragent/admission/mutate/autoinstrumentation/target_mutator.go
```diff
@@ -56,6 +56,11 @@ func NewTargetMutator(config *Config, wmeta workloadmeta.Component) (*TargetMuta
 		disabledNamespacesMap[ns] = true
 	}
 
+	var tagResolver *TagResolver
+	if config.rcClient != nil {
+		tagResolver = NewTagResolver(config.rcClient)
```
betterengineering: Nit - instead of using config, this probably should be passed as an argument to New

cmd/cluster-agent/subcommands/start/command.go
```diff
@@ -362,6 +362,7 @@ func start(log log.Component,
 		DatadogClient:          dc,
 	}
 
+	// ERIKA: This does NOT include the admission controller?
```
betterengineering: To develop this change, you should checkout [injector-dev](https://github.com/DataDog/injector-dev)! You can do something like this:

```bash
injector-dev apply -f scenario.yaml --build
```
<details>
  <summary>scenario.yaml</summary>


```yaml
helm:
  apps:
    - name: python-injection
      namespace: application
      values:
        env:
          - name: DD_TRACE_DEBUG
            value: "true"
            value: "true"
        image:
          repository: registry.ddbuild.io/ci/injector-dev/python
          tag: 2cd78ded
        podLabels:
          language: python
          tags.datadoghq.com/env: local
        service:
          port: "8080"
  versions:
    agent: 7.69.1
    cluster_agent:
      version: 7.69.1
      build: {}
    injector:
      version: 0.44.0
  config:
    datadog:
      site: "datad0g.com"
      apm:
        instrumentation:
          enabled: true
          targets:
            - name: python
              podSelector:
                matchLabels:
                  language: python
              ddTraceVersions:
                python: default
```

</details>

betterengineering (commented):
This looks great! Great work @erikayasuda! We could use some unit tests, which you mentioned you have locally

pkg/clusteragent/admission/mutate/autoinstrumentation/config.go
```diff
@@ -90,6 +94,12 @@ func excludedContainerNamesContainerFilter(c *corev1.Container) bool {
 
 // NewConfig creates a new Config from the datadog config. It returns an error if the configuration is invalid.
 func NewConfig(datadogConfig config.Component) (*Config, error) {
+	return NewConfigWithRemoteConfigClient(datadogConfig, nil)
+}
+
+// NewConfigWithRemoteConfigClient creates a new Config from the datadog config with a remote config client.
+// It returns an error if the configuration is invalid.
+func NewConfigWithRemoteConfigClient(datadogConfig config.Component, rcClient *rcclient.Client) (*Config, error) {
```
stanistan: One thing that this makes me think of is that we might not explicitly want to depend on `rcClient` everywhere and our actual dependency is the `TagResolver`, so instead we pass that into the configuration?

pkg/clusteragent/admission/mutate/autoinstrumentation/namespace_mutator.go
```diff
@@ -189,17 +220,18 @@ func (m *mutatorCore) injectTracers(pod *corev1.Pod, config extractedPodLibInfo)
 	}
 
 	for _, lib := range config.libs {
+		resolvedImage := m.resolveLibraryImage(lib)
```
stanistan: I think this should happen in `extractLibInfo`, basically, where we define the `libs` in the first place and not _after_ we've defined the libs. I think that's why you have the "This feels gross comment" :) 

https://github.com/DataDog/datadog-agent/blob/a2198e5607e233d2116d8f7f07671053663f36a1/pkg/clusteragent/admission/mutate/autoinstrumentation/namespace_mutator.go#L374

stanistan: The other thing to consider in `extractLibInfo` is that you could have `libInfo` itself take a `TagResolver` to configure itself when its being constructed.

pkg/clusteragent/admission/mutate/autoinstrumentation/namespace_mutator.go
```diff
@@ -304,9 +336,14 @@ func (m *mutatorCore) newInitContainerMutators(
 
 // newInjector creates an injector instance for this pod.
 func (m *mutatorCore) newInjector(pod *corev1.Pod, startTime time.Time, lopts libRequirementOptions) *injector {
+	imageTag := m.config.Instrumentation.InjectorImageTag
+	if m.tagResolver != nil {
```
stanistan: Two notes:

1. We should consider _always_ providing a `resolver` so that we don't have to do `nil` checks and we can create a `noop` and use an interface to always be able to use it
2. I think there's a bug here since it might be that the `InjectorImageTag` is just a tag like `"0"` and i'm not sure where the resolver is checking which _repo_ this image is for, only passing in the tag and the registry, we might be missing the repo name in the registry here?

pkg/clusteragent/admission/mutate/autoinstrumentation/tag_resolver.go
```diff
@@ -0,0 +1,177 @@
...
+import (
+	"encoding/json"
+	"sync"
+
+	rcclient "github.com/DataDog/datadog-agent/pkg/config/remote/client"
+	"github.com/DataDog/datadog-agent/pkg/remoteconfig/state"
+	"github.com/DataDog/datadog-agent/pkg/util/log"
+)
+
+// TagResolver is a component that resolves image tags from a remote configuration.
```
stanistan: Per another comment, I think it would be good to have an interface and have an implementation for a `noopTagResolver` that always returns the tag directly, and another that has one for an `rcclient` if its there.

The other thing I'm seeing here is that sometimes you're doing `nil` checks, but `NewTagResolver` never returns `nil`.

pkg/clusteragent/admission/mutate/autoinstrumentation/tag_resolver.go
```diff
@@ -0,0 +1,177 @@
...
+				Digest:           imageInfo.Digest,
+				CanonicalVersion: imageInfo.CanonicalVersion,
+			}
+		}
+		newCache[repo.RepositoryName] = tagMap
+
+		applyStateCallback(configKey, state.ApplyStatus{
+			State: state.ApplyStateAcknowledged,
+		})
+	}
```
stanistan: You're rebuilding the full cache every time so you can make the lock call smaller.

Build the `newCache` then do the `Lock` and `Unlock` and reassign, you're not using the full old cache at all here right?

pkg/clusteragent/admission/mutate/autoinstrumentation/tag_resolver.go
```diff
@@ -0,0 +1,177 @@
...
+	tr.tagMappings = newCache
+}
+
+// ResolveImageTag resolves an image tag from a repository name and a tag.
+func (tr *TagResolver) ResolveImageTag(repositoryName, tag string) string {
+	tr.mu.RLock()
+	defer tr.mu.RUnlock()
+
+	if len(tr.tagMappings) == 0 {
+		log.Debugf("TagResolver: Cache empty, using original tag %s for %s", tag, repositoryName)
```
stanistan: You don't need to prefix your logs with `TagResolver` all of our logs have the filename in them, so `tag_resolver` will be in the logs :)

pkg/clusteragent/admission/mutate/autoinstrumentation/tag_resolver.go
```diff
@@ -0,0 +1,177 @@
...
+
+		applyStateCallback(configKey, state.ApplyStatus{
+			State: state.ApplyStateAcknowledged,
+		})
+	}
+	tr.tagMappings = newCache
+}
+
+// ResolveImageTag resolves an image tag from a repository name and a tag.
+func (tr *TagResolver) ResolveImageTag(repositoryName, tag string) string {
```
stanistan: The thing that's a little interesting here is that you can put in a `tag` so the image that comes in is `repo: foo, tag: bar`, so `foo:bar`, the thing that comes out is the full image ref, i'm not exactly sure the structure, with an `@sha...` or is does it include the repo name?

We should consider changing this to be an `ImageResolver` so that we know we're going to get back the _full image_ here and we can give it a full image name, and parse it out as needed and it either _returns_ a `ResolvedImage` or not.

```go
//
func (ir *imageResolver) Resolve(repo, tag string) (string, bool) {
 // ...
}
```

When you're not changing anything and want to return `""` it's better to do `"", false` so the caller knows that nothing has changed.

stanistan (commented):
Thank you for the PR! 

I obviously have a lot of comments in this review as I read through it, this is a great start! At a high-level, there are a few changes I would recommend in this PR.

1. Construct the `Resolver` outside of the configuration object and pass it in so `autoinstrumentation.Config` doesn't need to know its getting config from rc or static or otherwise, fewer dependencies that way and easier to test.
2. Update the `TagResolver` to be an `ImageResolver`, I think it will end up being a cleaner implementation because we're swapping a `:tag` suffix to `@sha...` suffix on a container image.
3.  Represent it via an interface so when there's no `rcclient` we can have a simple implementation
4. We should consider moving this `defaultLibInfo` stuff into the ImageResolver https://github.com/DataDog/datadog-agent/blob/main/pkg/clusteragent/admission/mutate/autoinstrumentation/auto_instrumentation.go#L176-L185 as what you're doing is a better encapsulation of this stuff and this is where we do all the [defaults](https://github.com/DataDog/datadog-agent/blob/main/pkg/clusteragent/admission/mutate/autoinstrumentation/auto_instrumentation_test.go#L40-L63) as is.

sjain313:
heads up - You can rebase on top of latest master to get rid of any unexpected CI failures 
https://dd.slack.com/archives/C06PBHLD4DQ/p1755866084196919

https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1097317175 might be related to it so thought of giving a heads up.

releasenotes-dca/notes/add-k8s-gradual-rollout-support-315926f2d3c1b6d3.yaml
```diff
@@ -0,0 +1,13 @@
...
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+features:
+  - |
+    Introducing support for gradual rollout for K8s SSI deployments.
+    This will allow the injector and tracer library releases to be rolled out by subsets of customers,
+    and prevent the rollout to more customers if there are issues detected.
```
estherk15: Optional suggestion!

```suggestion
    Added support for gradual rollout in K8s SSI deployments.
This enables injector and tracer library releases to subsets of customers, with the ability to pause if issues are detected.
```

estherk15 (approved):

pkg/clusteragent/admission/mutate/autoinstrumentation/image_resolver.go
```diff
@@ -0,0 +1,267 @@
...
+				State: state.ApplyStateError,
+				Error: err.Error(),
+			})
+		} else {
+			applyStateCallback(configKey, state.ApplyStatus{
+				State: state.ApplyStateAcknowledged,
+			})
+		}
+	}
+	r.updateCacheFromParsedConfigs(validConfigs)
```
mellon85: The ACKing should be the last part of this function to avoid an ACK and then maybe a bug in the updateCacheFromParsedConfigs code crashes the agent

pkg/clusteragent/admission/mutate/autoinstrumentation/image_resolver.go
```diff
@@ -0,0 +1,267 @@
...
+	log.Debugf("Resolved %s:%s -> %s", repository, tag, resolved)
+	return &resolved, true
+}
+
+func isDatadoghqRegistry(registry string) bool {
+	var datadoghqRegistries = []string{
+		"gcr.io/datadoghq",
+	}
+	return slices.Contains(datadoghqRegistries, registry)
+}
```
adel121: Nit: could be simplified as `DatadogGCRRegistry == registry`

adel121: Also, is this the only supported registry ?

pkg/clusteragent/admission/mutate/autoinstrumentation/image_resolver.go
```diff
@@ -0,0 +1,267 @@
...
+			log.Debugf("Still waiting for initial remote config, retrying in %v", backoff)
+			backoff = min(backoff*2, maxBackoff)
+		}
+	}
+}
+
+// Resolve resolves a registry, repository, and tag to a digest-based reference.
+// Input: "gcr.io/datadoghq", "dd-lib-python-init", "v3"
+// Output: "gcr.io/datadoghq/dd-lib-python-init@sha256:abc123...", true
+// If resolution fails or is not available, it returns the original image reference.
```
adel121: The implementation below returns nil in this case. 

Can you please update the comment or the below implementation ?

pkg/clusteragent/admission/mutate/autoinstrumentation/image_resolver.go
erikayasuda: I'm confirming this with Injection Platform. If it's just this single registry, then I'll simplify this to be your suggestion!

pkg/clusteragent/admission/mutate/autoinstrumentation/image_resolver.go
erikayasuda: Ah OK, so even if I have a mix of `ApplyStateError` and `ApplyStateAcknowledged`, I should do all of them after everything has been processed?

pkg/clusteragent/admission/mutate/autoinstrumentation/image_resolver.go
erikayasuda: Fixed the comment, thanks for catching that

pkg/clusteragent/admission/mutate/autoinstrumentation/image_resolver.go
mellon85: yep, to be sure to return the appropriate state, worst come to worst (crash), the message will stay Unacked we know something is off

erikayasuda: Gotcha, thanks for clarifying. I'll update it to reflect that

joepeeples (approved):

pkg/clusteragent/admission/mutate/autoinstrumentation/image_resolver.go
```diff
@@ -0,0 +1,268 @@
...
+func newRemoteConfigImageResolver(rcClient RemoteConfigClient) ImageResolver {
+	resolver := &remoteConfigImageResolver{
+		rcClient:      rcClient,
+		imageMappings: make(map[string]map[string]ResolvedImage),
+	}
+
+	rcClient.Subscribe(state.ProductGradualRollout, resolver.processUpdate)
+	log.Debugf("Subscribed to %s", state.ProductGradualRollout)
+
+	if err := resolver.waitForInitialConfig(5 * time.Second); err != nil {
```
betterengineering: nit - this timeout could be configurable.

pkg/clusteragent/admission/mutate/autoinstrumentation/image_resolver.go
```diff
@@ -0,0 +1,268 @@
...
+		log.Debugf("No mapping found for %s:%s", repository, tag)
+		return nil, false
+	}
+
+	log.Debugf("Resolved %s:%s -> %s", repository, tag, resolved)
+	return &resolved, true
+}
+
+func isDatadoghqRegistry(registry string) bool {
+	var datadoghqRegistries = []string{
```
betterengineering: nit: this could be a map.

pkg/clusteragent/admission/mutate/autoinstrumentation/image_resolver.go
```diff
@@ -0,0 +1,268 @@
...
+		"gcr.io/datadoghq",
```
betterengineering: Here is the full list of registries: https://docs.datadoghq.com/containers/guide/changing_container_registry/

pkg/clusteragent/admission/mutate/autoinstrumentation/image_resolver.go
```diff
@@ -0,0 +1,268 @@
...
+			errors[configKey] = fmt.Errorf("missing repository_name or repository_url in config %s", configKey)
+			continue
+		}
+		validConfigs[configKey] = repo
+	}
+	return validConfigs, errors
+}
+
+func (r *remoteConfigImageResolver) updateCacheFromParsedConfigs(validConfigs map[string]RepositoryConfig) {
+	r.mu.Lock()
```
betterengineering: nit - this lock can be moved down to surround `r.imageMappings`

pkg/clusteragent/admission/mutate/autoinstrumentation/language_versions.go
```diff
@@ -48,6 +51,23 @@ func (l language) libInfo(ctrName, image string) libInfo {
 	}
 }
 
+// DEV: Will attempt to resolve, defaults to legacy if unable
+func (l language) libInfoWithResolver(ctrName, registry string, version string, imageResolver ImageResolver) libInfo {
+	resolvedImage, ok := imageResolver.Resolve(registry, fmt.Sprintf("dd-lib-%s-init", l), version)
+	var image string
+	if !ok {
+		log.Warnf("failed to resolve image %s/%s:%s, using fallback", registry, fmt.Sprintf("dd-lib-%s-init", l), version)
```
betterengineering: In the case there is a noop resolver, this will create a lot of warnings. We likely will need a way to only log if it's supposed to be using RC and is not able to.

pkg/clusteragent/admission/mutate/autoinstrumentation/testdata/image_resolver_multi_repo.json
```diff
@@ -0,0 +1,53 @@
+{
+  "employee/K8S_INJECTION_DD/1.dd-lib-python-init.json/config":
```
betterengineering, line 2: Just because I'm curious - what does employee mean? 👀

releasenotes-dca/notes/add-k8s-gradual-rollout-support-315926f2d3c1b6d3.yaml
```diff
@@ -0,0 +1,12 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+features:
+  - |
+    Added support for gradual rollout in K8s SSI deployments.
```
betterengineering, line 11: We likely need to highlight more to customers what they should expect to see from this change. Most notably, will this be enabled by default? Do they need remote config enabled or not? That sort of thing. Customers do read these, and they provide us something to point at if they get an unexpected result.

betterengineering (approved):
This looks great! Thanks so much for putting this together @erikayasuda! Approving because this looks good as is. The things we should follow up on:
* The release note should have more context
* The full list of datadog registries should be added
* The warn log line in `libInfoWithResolver` might be noisy when a noop resolver is used

mellon85 (approved):

pkg/clusteragent/admission/mutate/autoinstrumentation/language_versions.go
erikayasuda: Discussed offline: Converting this to `Debug` level for now

pkg/clusteragent/admission/mutate/autoinstrumentation/image_resolver.go
erikayasuda: Agreed! Will update this in a follow up PR

releasenotes-dca/notes/add-k8s-gradual-rollout-support-315926f2d3c1b6d3.yaml
erikayasuda: Updated the release note, let me know what you think!

pkg/clusteragent/admission/mutate/autoinstrumentation/testdata/image_resolver_multi_repo.json
erikayasuda: This is the format that I copied from the actual RC config updates. I think it's just to distinguish between employee managed configs and customer managed configs

pkg/clusteragent/admission/mutate/autoinstrumentation/language_versions.go
stanistan: I don't know if I would warn here every single time we're using a fallback, I think that means that if someone is using a private registry the logs will be full of warnings and that's not great

pkg/clusteragent/admission/mutate/autoinstrumentation/image_resolver.go
```diff
@@ -0,0 +1,271 @@
...
+	errors := make(map[string]error)
+	for configKey, rawConfig := range configs {
+		var repo RepositoryConfig
+		if err := json.Unmarshal(rawConfig.Config, &repo); err != nil {
+			errors[configKey] = fmt.Errorf("failed to unmarshal: %w", err)
+			continue
+		}
+
+		if repo.RepositoryName == "" || repo.RepositoryURL == "" {
+			errors[configKey] = fmt.Errorf("missing repository_name or repository_url in config %s", configKey)
```
stanistan, line 173: Given these are grouped by config_key and then also logged by config key by the caller, it's going to be pretty verbose.

---

~You can also consider using [`multierror`](https://github.com/DataDog/datadog-agent/blob/a79bec82cae278364ad29ee07cbbeeaba3f07db7/pkg/clusteragent/autoscaling/workload/config_retriever_values.go#L18) here.~  nevermind, I see these are also used for reporting back to RC

pkg/clusteragent/admission/mutate/autoinstrumentation/language_versions.go
erikayasuda: I think I'll just remove it then. I have pretty thorough logging in `ImageResolver.Resolve()` so it's probably redundant anyways

pkg/clusteragent/admission/mutate/autoinstrumentation/image_resolver.go
```diff
@@ -0,0 +1,271 @@
...
+	resolver := &remoteConfigImageResolver{
+		rcClient:      rcClient,
+		imageMappings: make(map[string]map[string]ResolvedImage),
+	}
+
+	rcClient.Subscribe(state.ProductGradualRollout, resolver.processUpdate)
+	log.Debugf("Subscribed to %s", state.ProductGradualRollout)
+
+	if err := resolver.waitForInitialConfig(5 * time.Second); err != nil {
+		log.Debugf("Failed to load initial image resolution config: %v. Image resolution will be disabled.", err)
```
stanistan: This should be a warning if this is expected to succeed.

stanistan (commented):
This looks good to me so far! I think somewhere in the implementation, we lost the changes to the `injector.go`, was that on purpose / scoping?

pkg/clusteragent/admission/mutate/autoinstrumentation/image_resolver.go
```diff
@@ -0,0 +1,271 @@
...
+func newRemoteConfigImageResolver(rcClient RemoteConfigClient) ImageResolver {
+	resolver := &remoteConfigImageResolver{
+		rcClient:      rcClient,
+		imageMappings: make(map[string]map[string]ResolvedImage),
+	}
+
+	rcClient.Subscribe(state.ProductGradualRollout, resolver.processUpdate)
+	log.Debugf("Subscribed to %s", state.ProductGradualRollout)
+
+	if err := resolver.waitForInitialConfig(5 * time.Second); err != nil {
```
stanistan: TODO in a follow up: we should have timing as a metric that we add here

Reasoning is that this is blocking and will delay the initialization of the mutating wehboook. It'd be good to report it so that we can use it for troubleshooting and introspection later.

cmd/cluster-agent/subcommands/start/command.go
```diff
@@ -506,6 +510,7 @@ func start(log log.Component,
 		} else {
 			log.Info("Auto instrumentation patcher is disabled")
 		}
+		imageResolver := autoinstrumentation.NewImageResolver(rcClient)
```
stanistan, line 513: Because this is called here, this will block everything after this call for up to 5 seconds. 

@adel121 should we consider running the `init` check in this function as a goroutine?

adel121: > should we consider running the init check in this function as a goroutine?

If we run it as an init check, how would the autoinstrumentation webhook know when the image resolver has finished the initialisation process, and how would it receive the new value?

erikayasuda: I was thinking we could just run the `waitForInitialConfig(...)` block in `newRemoteConfigImageResolver(...)` as a goroutine, but let the rest of the `ImageResolver` initialize. We handle cases where there is an empty cache/`imageMapping`, and it just defaults to using the image reference with the mutable tag (current behavior), so it should be OK if the cache isn't fully initialized yet.

adel121: It is what I also had in mind. 

But are you okay with some pods getting mutable tags while others get resolved tags ?

erikayasuda: I was chatting with Stan yesterday and we wanted to make sure we got Container Platform input on the 5 second block before going with this. If you think the tags being inconsistent would be an issue for customers, so the 5 second max wait to initialize the cache is worth that trade-off of consistency, then I'm comfortable keeping the current implementation. If the (max) 5 second block on initializing the other components is more of an issue, then we could go with the goroutine + possible inconsistency with tags while the cache is initializing.

adel121 (approved):

pkg/clusteragent/admission/mutate/autoinstrumentation/injector.go
```diff
@@ -178,6 +179,15 @@ func injectorWithImageName(name string) injectorOption {
 
 func injectorWithImageTag(tag string) injectorOption {
 	return func(i *injector) {
+		if i.imageResolver == nil {
+			panic("injectorWithImageTag called without imageResolver")
```
betterengineering: We shouldn't panic here. It's potentially dangerous as it would crash our webhook or potentially the full cluster agent. Instead you could check if it's not nil and nest the image resolver code in it

releasenotes-dca/notes/add-k8s-gradual-rollout-support-315926f2d3c1b6d3.yaml
```diff
@@ -0,0 +1,17 @@
...
+    This feature is enabled if the following conditions are met:
+    - ``DD_ADMISSION_CONTROLLER_AUTO_INSTRUMENTATION_ENABLED=true`` and/or ``DD_APM_INSTRUMENTATION_ENABLED=true``
+    - Remote config is enabled in the organization
+    - Using the `gcr.io/datadoghq` registry
```
betterengineering, line 15: Nit - it's all datadog repositories now

PR #39934, opened 2025-08-18
CXP-2433 [agent] Windows process test flake fix

gengnamstyle:
### What does this PR do?
- re-excutes windows defender scan process to increase cpu usage to remove flaky test

### Motivation
- manual process check becomes flaky when cpu usage is 0 and the json output omits zero values

- run test manually and ci passed
- I will watch the logs of this test to verify that the flake as actually disappeared

- run e2e test manually and ci

### DIFF
--- a/test/new-e2e/tests/process/windows_test.go
+++ b/test/new-e2e/tests/process/windows_test.go
@@ -246,6 +246,8 @@ func (s *windowsTestSuite) TestUnprotectedProcessCheckIO() {
 
 func (s *windowsTestSuite) TestManualProcessCheck() {
 	flake.Mark(s.T())
+	// test can be flaky due to missing CPU stats when cpu usage is extremely low (json output omits 0 values), so we want to re-run a full scan to ensure we have CPU stats
+	s.Env().RemoteHost.MustExecute("Start-MpScan -ScanType FullScan -AsJob")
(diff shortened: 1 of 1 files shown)


test/new-e2e/tests/process/windows_test.go
```diff
@@ -246,6 +246,8 @@ func (s *windowsTestSuite) TestUnprotectedProcessCheckIO() {
 
 func (s *windowsTestSuite) TestManualProcessCheck() {
 	flake.Mark(s.T())
+	// test can be flaky due to missing CPU stats when cpu usage is extremely low (json output omits 0 values), so we want to re-run a full scan to ensure we have CPU stats
+	s.Env().RemoteHost.MustExecute("Start-MpScan -ScanType FullScan -AsJob")
```
kkhor-datadog, line 250: If we need to do this for more tests, we might consider putting this in the test-level setup in the future. Currently I see this is in the suite-level setup.

kkhor-datadog (approved):

test/new-e2e/tests/process/windows_test.go
gengnamstyle: yeah! I agree. That's where I found the command.

PR #39935, opened 2025-08-18
APM: Enable configuring CSA flush interval

ajgajg1134:
### What does this PR do?
Add a new configuration for the ClientStatsAggregator(CSA) in the trace-agent so we can flush at different intervals

### Motivation
We're investigating some cases internally where stats payload flushing can take longer than 1 second, this can cause the CSA to back up in processing payloads and eventually tracers that are using Client Side Stats will timeout sending these stats payloads.

Ran locally and ensured stats payloads are handled correctly even with a longer flush interval

Ideally no one ever needs to use this configuration, but it will be useful here as we investigate this issue


songy23 (approved):

pkg/config/setup/apm.go
```diff
@@ -146,6 +146,7 @@ func setupAPM(config pkgconfigmodel.Setup) {
 	config.BindEnv("apm_config.connection_limit", "DD_APM_CONNECTION_LIMIT", "DD_CONNECTION_LIMIT")
 	config.BindEnv("apm_config.connection_reset_interval", "DD_APM_CONNECTION_RESET_INTERVAL")
 	config.BindEnv("apm_config.max_sender_retries", "DD_APM_MAX_SENDER_RETRIES")
+	config.BindEnv("apm_config.client_stats_flush_interval", "DD_APM_CLIENT_STATS_FLUSH_INTERVAL")
```
dustmop: Instead of having logic to set the default when `!IsConfigured`, how about doing this?

```suggestion
	config.BindEnvAndSetDefault("apm_config.client_stats_flush_interval", 1, "DD_APM_CLIENT_STATS_FLUSH_INTERVAL")
```

cswatt (approved):

ichinaski (approved):

pkg/config/setup/apm.go
ajgajg1134: That makes sense to me!

dustmop (approved):

PR #39936, opened 2025-08-18
Add `error.type` to fleet telemetry metadata for Error Tracking

mrafi97:
### What does this PR do?
This PR adds `error.type` metadata to spans collected by Fleet automation telemetry in order to utilize Error Tracking.

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1681

Updated an existing unit test to ensure the `error.type` metadata is added.

Results from building the installer and manually inducing errors:

Before adding error type:
image

After adding error type:
image

In Error Tracking:
image


pkg/fleet/installer/telemetry/span.go
```diff
@@ -148,3 +151,10 @@ func getSpanIDsFromContext(ctx context.Context) (spanIDs, bool) {
 func setSpanIDsInContext(ctx context.Context, span *Span) context.Context {
 	return context.WithValue(ctx, spanKey, spanIDs{traceID: span.span.TraceID, spanID: span.span.SpanID})
 }
+
+func getRootErrType(err error) string {
+	for u := errors.Unwrap(err); u != nil; u = errors.Unwrap(u) {
+		err = u
+	}
+	return reflect.TypeOf(err).String()
```
alexn-dd, line 163: Does `TypeOf handle nil?

pkg/fleet/installer/telemetry/span.go
```diff
@@ -148,3 +151,10 @@ func getSpanIDsFromContext(ctx context.Context) (spanIDs, bool) {
 func setSpanIDsInContext(ctx context.Context, span *Span) context.Context {
 	return context.WithValue(ctx, spanKey, spanIDs{traceID: span.span.TraceID, spanID: span.span.SpanID})
 }
+
+func getRootErrType(err error) string {
```
alexn-dd: For consistency, I suggest fully spelling it out: getRootErrorType

alexn-dd (approved):

pkg/fleet/installer/telemetry/span.go
BaptisteFoy: Looks like it does; but we may want to avoid setting the tag if the error is nil, to follow what the tracer does: https://github.com/DataDog/dd-trace-go/blob/6d45b474e927c1ffd07e455b2ee705448355dd45/ddtrace/tracer/span.go#L432

BaptisteFoy (approved):
Same comments as Alex; but else LGTM. Thanks! 🚀

pkg/fleet/installer/telemetry/span.go
mrafi97: Earlier in the code, we do check to see if err is nil before setting the error metadata so I thought it was fine to leave as is:
```
	if err != nil {
		s.span.Error = 1
		s.span.Meta = map[string]string{
			"error.message": err.Error(),
			"error.stack":   string(debug.Stack()),
			"error.type":    getRootErrorType(err),
		}
```

I can check for nil as well within the function as a sanity check

PR #39937, opened 2025-08-18
[AGENTRUN-642] remove global jmx logger

GustavoCaso:
### What does this PR do?

After fixing a SEGFAULT because of a nil pointer https://github.com/DataDog/datadog-agent/pull/39847, we decided to make the jmxLogger not use a global pointer. That way we no longer have to worry about the SEGFAULT problem before. 

### Motivation

Refactor, cleanup and remove use of global variables

Added unit tests

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -65,9 +65,6 @@ issues:
       linters: [staticcheck]
     # Treat this list as a TODO for fixing issues with pkgconfigusage custom linter
     # DO NOT ADD NEW ENTRIES
-    - path: comp/agent/jmxlogger/jmxloggerimpl/jmxlogger.go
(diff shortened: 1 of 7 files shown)


misteriaud (approved):
Nice work ! That's a great cleanup 🎉 
Maybe you should move the removed `TestJMXLog` to the component, to keep the coverage.

GustavoCaso:
> Maybe you should move the removed TestJMXLog to the component, to keep the coverage.

Absolutely. Great remark 😄

misteriaud (approved):

tobz (approved):

ofek (approved):

PR #39939, opened 2025-08-18
DSMON-930: Fix bug in kafka consumer check unscheduling logic

piochelepiotr:
### What does this PR do?

The data streams controller listens to remote configuration, and can modify running kafka checks.
However, when configurations are modified (tags are added in k8s, secrets are resolved), the config changes it sends to the configuration provider are supposed to be unresolved configurations. However, `GetAllConfigs` is returning resolved configurations. This is causing unscheduling to fail (because the unschedule function was comparing the resolved integration to unresolved ones, and was failing to find the right integration to unschedule).

This is leading to duplicate kafka consumer integrations running in parallel, causing resource exhaustion, and unwanted load on Kafka.


gitstevenpham (approved):

janine-c (approved):

rahulkaukuntla (approved):

.github/CODEOWNERS
```diff
@@ -605,7 +605,6 @@
 /pkg/ebpf/map_cleaner*.go               @DataDog/universal-service-monitoring
 /pkg/compliance/                        @DataDog/agent-cspm
 /pkg/databasemonitoring                 @DataDog/database-monitoring
-/pkg/datastreams                        @DataDog/data-streams-monitoring
```
davidor, line 608: Can you add your team as owners or co-owners of `comp/core/autodiscovery/providers/datastreams/`?

davidor (approved):

PR #39944, opened 2025-08-18
Cross-org telemetry for DogStatsD metrics type mixture

blt:
### What does this PR do?

This commit introduces COATs metrics for the metrics type mix we get in DogStatsD traffic. We understand that dogstatsd performance is heavily dictated by the metrics mixture an instance of dogstatsd receives. The goal of this change is to record this measure to improve fidelity of performance experiments.

I've introduced a new metric in the server `metric_type_count` that is an absolute counters, distinct from our existing `context_by_mtype` metric in that if a gauge is sent for the same context 1,000 times context_by_mtype will be 1 whereas metric_type_count will be 1,000. Metric types are distinguished by the `metric_type` tag. 

I have, with the same reasoning, exposed the `context_by_mtype` metric via COATs. 

Confirmed collected by the Regression Detector, see [this panel](https://app.datadoghq.com/dashboard/ykh-ua8-vcu/SMP-Regression-Detector-Metrics?fromUser=false&refresh_mode=paused&tpl_var_experiment%5B0%5D=quality_gate_metrics_logs&tpl_var_run-id%5B0%5D=14f1df1d-98c0-4a21-a6ab-78c65515dbac&view=spans&from_ts=1755552364000&to_ts=1755552964000&live=false&tile_focus=3140684982653796). 

### Additional Notes

This also serves as a correctness check on our Regression Detector experiments: if the metrics mix recorded by DogStatsD is distinct from the experiment configuration, that's a problem. This will make it much, much more feasible to model workloads in Regression Detector experiments.

### DIFF
--- a/comp/dogstatsd/server/server.go
+++ b/comp/dogstatsd/server/server.go
@@ -183,7 +183,13 @@ type server struct {
 	packetsTelemetry        *packets.TelemetryStore
 	stringInternerTelemetry *stringInternerTelemetry
 	// Counter for absolute metric types
-	tlmMetricTypes telemetry.Counter
+	tlmMetricTypes            telemetry.Counter
+	tlmMetricTypeGauge        telemetry.SimpleCounter
+	tlmMetricTypeCounter      telemetry.SimpleCounter
+	tlmMetricTypeDistribution telemetry.SimpleCounter
+	tlmMetricTypeHistogram    telemetry.SimpleCounter
+	tlmMetricTypeSet          telemetry.SimpleCounter
+	tlmMetricTypeTiming       telemetry.SimpleCounter
 }
 
 func initTelemetry() {
@@ -346,6 +352,15 @@ func newServerCompat(cfg model.ReaderWriter, log log.Component, hostname hostnam
 	// per-context but absolute.
 	s.tlmMetricTypes = telemetrycomp.NewCounter("dogstatsd", "metric_type_count",
 		[]string{"metric_type"}, "Count of metrics processed by dogstatsd by type")
+	// Cache counter/tag instances to avoid interior hashmap lookups. If
+	// SimpleCounter ever uses relaxed atomics this will also be improved
+	// over a simple Inc("tag") on ARM, as Inc implies a mutex cycle.
+	s.tlmMetricTypeGauge = s.tlmMetricTypes.WithValues("gauge")
+	s.tlmMetricTypeCounter = s.tlmMetricTypes.WithValues("counter")
+	s.tlmMetricTypeDistribution = s.tlmMetricTypes.WithValues("distribution")
+	s.tlmMetricTypeHistogram = s.tlmMetricTypes.WithValues("histogram")
+	s.tlmMetricTypeSet = s.tlmMetricTypes.WithValues("set")
+	s.tlmMetricTypeTiming = s.tlmMetricTypes.WithValues("timing")
 
 	s.listernersTelemetry = listeners.NewTelemetryStore(getBuckets(cfg, log, "telemetry.dogstatsd.listeners_latency_buckets"), telemetrycomp)
 	s.packetsTelemetry = packets.NewTelemetryStore(getBuckets(cfg, log, "telemetry.dogstatsd.listeners_channel_latency_buckets"), telemetrycomp)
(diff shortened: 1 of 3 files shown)


blt:
* **#39944** <a href="https://app.graphite.dev/github/pr/DataDog/datadog-agent/39944?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/datadog-agent/39944?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jeremy-hanna (approved):

tobz (approved):

comp/dogstatsd/server/server.go
```diff
@@ -850,6 +857,21 @@ func (s *server) parseMetricMessage(metricSamples []metrics.MetricSample, parser
 		return metricSamples, err
 	}
 
+	switch sample.metricType {
+	case gaugeType:
+		s.tlmMetricTypes.Inc("gauge")
```
vickenty: `Inc` with a value means a mutex lock and a hashmap lookup. This used to become a bottleneck in the past. Would it be possible to create `SimpleCounter` for each type and perform `Inc` on those (like [here](https://github.com/DataDog/datadog-agent/blob/81b36c3fd2c2d495340c30b1cb5042d30aca4aec/comp/dogstatsd/server/server.go#L820-L821))?

releasenotes/notes/Add-DogStatsD-metrics-to-count-metric-types.-4d74421bcd2222e6.yaml
```diff
@@ -0,0 +1,11 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+enhancements:
+  - |
+    DogStatsD emits a new metric `metric_type_count` that tracks gauge, counters etc totals. Exposed via COATs as well as contexts_by_mtype.
```
vickenty: Changelog is not markdown, but RST and uses double backticks for verbatim text.

However, do we need to highlight COAT changes to users?

```suggestion
    DogStatsD emits a new metric ``metric_type_count`` that tracks gauge, counters etc totals. Exposed via COATs as well as ``contexts_by_mtype``.
```

jszwedko: > However, do we need to highlight COAT changes to users?

➕ My understanding is that the changelog is user-facing and so we should describe changes in terms of the user-facing impact. In this case, I think the second sentence would not be understandable to users. We could drop it or rewrite it to say that the metric is published as internal telemetry for future improvements.

blt: I'll drop the second sentence.

comp/dogstatsd/server/server.go
blt: Sure thing. I like that the SimpleCounter approach makes the cardinality explicit. 

You did nerd snipe me some. I got curious and went through the `SimpleCounter` et al implementation. I believe it's strongly consistent atomics all the way through. If we did some relaxed adds we'd be ahead on ARM. We're mutex-equivalent on both x86 and ARM right now. SimpleCounter does come out ahead on caching the hashmap lookups.

releasenotes/notes/Add-DogStatsD-metrics-to-count-metric-types.-4d74421bcd2222e6.yaml
blt: Resolved in [57fcdec](https://github.com/DataDog/datadog-agent/pull/39944/commits/57fcdec4681b7e3ba5bb973abb762b93c6e3a041).

comp/dogstatsd/server/server.go
blt: Done in [db31536](https://github.com/DataDog/datadog-agent/pull/39944/commits/db31536181092f5cb94a7cfae87835164d11a763).

jszwedko (approved):

iglendd (approved):
Looks good

releasenotes/notes/Add-DogStatsD-metrics-to-count-metric-types.-4d74421bcd2222e6.yaml
```diff
@@ -0,0 +1,11 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+enhancements:
+  - |
+    DogStatsD emits a new metric ``metric_type_count`` that tracks gauge, counters etc totals.
```
janine-c:
```suggestion
    DogStatsD emits a new metric, ``metric_type_count``, that tracks total gauges, counters, etc.
```

janine-c (approved):

PR #39945, opened 2025-08-18
WIP: extension of unique-handles through DSD

scottopell:
fork/extension of #39159 to see if I can get SMP results


comp/core/secrets/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/DataDog/datadog-agent/comp/core/secrets
+
+go 1.23.0
```
scottopell, line 3: I know this file is wrong, but I couldn't figure out how to get my build working without this...

PR #39949, opened 2025-08-18
[OTAGENT-480] small fix on logs agent exporter

songy23:
### What does this PR do?
Fix an incorrect use of logger in logs agent exporter and stores a ref to config in exporter

### Motivation
working towards https://github.com/DataDog/datadog-agent/pull/39893

### DIFF
--- a/comp/otelcol/otlp/components/exporter/logsagentexporter/logs_exporter.go
+++ b/comp/otelcol/otlp/components/exporter/logsagentexporter/logs_exporter.go
@@ -30,7 +30,7 @@ type Exporter struct {
 	logSource        *sources.LogSource
 	translator       *logsmapping.Translator
 	gatewaysUsage    otel.GatewayUsage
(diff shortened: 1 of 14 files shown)


comp/otelcol/otlp/components/exporter/logsagentexporter/logs_exporter.go
```diff
@@ -107,7 +108,7 @@ func (e *Exporter) ConsumeLogs(ctx context.Context, ld plog.Logs) (err error) {
 
 		content, err := ddLog.MarshalJSON()
 		if err != nil {
-			logger.Error("Error parsing log: " + err.Error())
```
songy23, line 110: This used to use `"github.com/stormcat24/protodep/pkg/logger".logger` I believe that's simply a typo. We should use collector's logger

tim-liakh-datadog (approved):

PR #39962, opened 2025-08-19
POC using rules_multitools to use prebuilt binaries.

aiuto:
Use rules_multitools to bring in a prebuilt buildifier for linux, macos, and windows.
This is an inital PoC.

Try it:
```
bazel run @multitool//tools/buildifier -- tools/BUILD.bazel
```

I have some concerns about rules_multitool, so we'll probably do our
own fork of it to overcome those, but this is good enough for now.

Note to reviewer:  The extra pinned dependencies in MODULE.bazel are to retain control over what versions we use, and not allow deps like rules_multitool to make the choice for us.

Next step: Add the tool to read the releases for a package and generate the lockfiles.


JSGette (approved):

KevinFairise2 (approved):

ofek:
Can we please incorporate https://github.com/buildbuddy-io/bazel_env.bzl so that we have a mechanism to add the tools to PATH? Example: https://github.com/buildbuddy-io/bazel_env.bzl/tree/main/examples

After you add that, I plan to use what is here in the exact same way for the build and developer environment images. Over there I will adjust PATH using its [output mechanism](https://github.com/buildbuddy-io/bazel_env.bzl#without-direnv-eg-in-ci).

aiuto:
I absolutely hate the magic of bazel-env.   It makes things magically surprising rather than concrete.
I could not live with that on the command line.

PR #39967, opened 2025-08-19
add OTEL agent as Windows service

guohdd:
### What does this PR do?
This PR makes Datadog OTEL agent a Windows service.

### Motivation
DDOT needs to be packaged and delivered on the Windows platform.

The Datadog OTEL agent as Windows service has been verified locally with following steps:
- build OTEL agent
- place otel-agent.exe at: "%ProgramData%\Datadog\Installer\packages\datadog-agent-ddot\stable\embedded\bin\"
- place otel-config.yaml at "%ProgramData%\Datadog\"
- make sure datadog.yaml has 'otelcollector' enabled.
- create datadog-otel-agent service with dependency of datadogagent, e.g.: sc.exe create datadog-otel-agent binpath= "C:\ProgramData\Datadog\Installer\packages\datadog-agent-ddot\stable\embedded\bin\otel-agent.exe" start= demand DisplayName= "Datadog OpenTelemetry Agent" depend= datadogagent
- start/stop/restart datadog agent service

The above tests will be replaced by automation once the OCI installer for Windows is added and built.

The OTEL executable and config file will be installed by agent-installer eventually, within the scope of this PR we manually place the OTEL files at the installation location.


janine-c (approved):

cmd/otel-agent/windows_resources/version.h
```diff
@@ -0,0 +1,6 @@
+#define RC_FILE_VERSION MAJ_VER,MIN_VER,PATCH_VER,0
+
+#define STRINGIFY(x) #x
+#define TO_STRING(x) STRINGIFY(x)
+
+#define FILE_VERSION_STRING TO_STRING(MAJ_VER.MIN_VER.PATCH_VER.0)
```
chouetz: 🥜 
```suggestion
#define FILE_VERSION_STRING TO_STRING(MAJ_VER.MIN_VER.PATCH_VER.0)

```

tasks/otel_agent.py
```diff
@@ -54,6 +56,21 @@ def build(ctx, byoc=False):
     ldflags = get_version_ldflags(ctx, major_version='7')
     ldflags += f' -X github.com/DataDog/datadog-agent/cmd/otel-agent/command.BYOC={byoc}'
 
+    # generate windows resources
+    # Shall we not hardcode the major_version? Options could be:
```
chouetz, line 64: I don't think there is a real issue at hardcoding the major version.

chouetz:
The `lint_component` job is failing, can you have a look? Thanks in advance

dustmop (approved):

tasks/otel_agent.py
guohdd: OK, I was thinking that we would need to update the major version number in more than one place when major version changes in the future. An option is to resolve the major version number from tags. Anyway, this is not something problematic and I am OK with following the existing pattern.

cmd/otel-agent/subcommands/controlsvc/command.go
```diff
@@ -0,0 +1,51 @@
...
+// Commands returns a slice of subcommands for the 'otel-agent' command.
+func Commands(_ *subcommands.GlobalParams) []*cobra.Command {
+	return []*cobra.Command{
+		{
+			Use:     "start-service",
+			Aliases: []string{"startservice"},
+			Short:   "starts the otel-agent within the service control manager",
+			Long:    ``,
+			RunE: func(_ *cobra.Command, _ []string) error {
+				return fxutil.OneShot(controlsvc.StartService)
```
clarkb7: this is a regular function it doesn't have component deps, you can call it directly,
```suggestion
				return controlsvc.StartService()
```

cmd/otel-agent/main_windows.go
```diff
@@ -0,0 +1,69 @@
...
+func main() {
+	flavor.SetFlavor(flavor.OTelAgent)
+
+	// if command line arguments are supplied, even in a non interactive session,
+	// then just execute that.  Used when the service is executing the executable,
+	// for instance to trigger a restart.
+	if len(os.Args) == 1 && servicemain.RunningAsWindowsService() {
+		servicemain.Run(&service{})
+		return
+	}
```
clarkb7: let's move this block into the cobra `run` subcommand. That way we can process command line args.

cmd/otel-agent/subcommands/run/service_windows.go
```diff
@@ -0,0 +1,69 @@
...
+		params := &cliParams{
+			GlobalParams: &subcommands.GlobalParams{
+				ConfPaths:    []string{confURL},
+				CoreConfPath: coreCfgPath,
+				ConfigName:   "datadog-otel",
+				LoggerName:   loggerName,
+				BYOC:         false,
+			},
+			pidfilePath: "", // No pidfile for service mode
+		}
```
clarkb7: see other comment about moving the service call into the `run` subcommand so we can use the command line args and don't have to create this structure manually.

cmd/otel-agent/subcommands/run/service_windows.go
```diff
@@ -0,0 +1,69 @@
...
+
+		// Run the otel-agent command with the service context
+		err := runOTelAgentCommand(ctx, params)
```
clarkb7: I don't see where our context is referenced to enable shutdown. Does the service shutdown immediately or does it hang for 15 seconds and then log an error?
see [trace-agent](https://github.com/DataDog/datadog-agent/blob/7734318d01b84821a4f91a79ac3e1ecf759a1126/comp/trace/agent/impl/agent_windows.go#L18-L28) for example

songy23 (approved):

cmd/otel-agent/main_windows.go
guohdd: So the suggestion here is that we want to support args in service mode instead of getting parameters from config file, and the explicit arguments override the configurations from config files, correct? Currently the otel service is launched from main, which follows the patterns of other services.

cmd/otel-agent/subcommands/run/service_windows.go
guohdd: Yes the service shuts down immediately with a 15s timer. My understanding is the collector gets our context and the collector returns on shutdown. (reference: cmd\otel-agent\subcommands\run\command.go)

cmd/otel-agent/main_windows.go
```diff
@@ -0,0 +1,23 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
```
clarkb7: we don't need this new file anymore, it's the same as `main.go`

cmd/otel-agent/subcommands/controlsvc/command.go
```diff
@@ -0,0 +1,51 @@
...
+				return controlsvc.StartService()
+			},
+		},
+		{
+			Use:     "stop-service",
+			Aliases: []string{"stopservice"},
+			Short:   "stops the otel-agent within the service control manager",
+			Long:    ``,
+			RunE: func(_ *cobra.Command, _ []string) error {
+				return fxutil.OneShot(controlsvc.StopService)
```
clarkb7: ditto https://github.com/DataDog/datadog-agent/pull/39967#discussion_r2291142547

cmd/otel-agent/subcommands/controlsvc/command.go
```diff
@@ -0,0 +1,51 @@
...
+			},
+		},
+		{
+			Use:     "restart-service",
+			Aliases: []string{"restartservice"},
+			Short:   "restarts the otel-agent within the service control manager",
+			Long:    ``,
+			RunE: func(_ *cobra.Command, _ []string) error {
+				return fxutil.OneShot(controlsvc.RestartService)
```
clarkb7: ditto https://github.com/DataDog/datadog-agent/pull/39967#discussion_r2291142547

cmd/otel-agent/subcommands/controlsvc/command_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+	// restart-service → OneShot(winctrl.RestartService)
+	fxutil.TestOneShotSubcommand(
+		t,
+		cmds,
+		[]string{"restart-service"},
+		winctrl.RestartService,
+		func() {},
+	)
+}
```
clarkb7: ditto https://github.com/DataDog/datadog-agent/pull/39967#discussion_r2291142547

cmd/otel-agent/subcommands/run/command_windows.go
```diff
@@ -0,0 +1,36 @@
...
+		RunE: func(_ *cobra.Command, _ []string) error {
+			params.GlobalParams = globalConfGetter()
+			if servicemain.RunningAsWindowsService() {
+				servicemain.Run(&service{cliParams: params})
+				return nil
+			}
+			return runOTelAgentCommand(context.Background(), params)
+		},
+	}
+	cmd.Flags().StringVarP(&params.pidfilePath, "pidfile", "p", "", "path to the pidfile")
```
clarkb7: we discussed not needing the pidfile on Windows, did you find some new information that makes it required?

cmd/otel-agent/subcommands/run/service_windows.go
```diff
@@ -0,0 +1,52 @@
...
+			ddRoot = filepath.Join(ddRoot, "Datadog")
+		}
+		if len(params.ConfPaths) == 0 {
+			cfg := filepath.Join(ddRoot, "otel-config.yaml")
+			params.ConfPaths = []string{"file:" + filepath.ToSlash(cfg)}
+		}
+		if params.CoreConfPath == "" {
+			params.CoreConfPath = filepath.Join(ddRoot, "datadog.yaml")
+		}
+	}
```
clarkb7: the other callers to `runOTelAgentCommand` don't have this block, why is it here only?

releasenotes/notes/ddot-windows-service-fa509e103be48f56.yaml
```diff
@@ -0,0 +1,15 @@
...
+#
+# Each section note must be formatted as reStructuredText.
+---
+features:
+  - |
+    Added Datadog OTEL agent as Windows service
+other:
+  - |
+    This is part of the effort to add Datadog OpenTelemetry agent as a
+    Windows service.
```
clarkb7: the sections in release notes are separated when combined into the final release notes, so this `other` section won't have any context. You should either remove the `other` section or merge it, for example

```suggestion
features:
  - |
    Added Datadog OTEL agent as Windows service

    This is part of the effort to add Datadog OpenTelemetry agent as a
    Windows service. 
```

cmd/otel-agent/subcommands/run/service_windows.go
guohdd: Since the two paths are critical, I'd like to ensure they have the necessary values. But you have a point, I will look further into how things are handled if a path is not provided in this case.

cmd/otel-agent/subcommands/run/command_windows.go
guohdd: good catch, will remove.

clarkb7 (approved):

chouetz (approved):

PR #39978, opened 2025-08-19
Implement ddot-BYOC workflow tool for deb and rpm packages

alopezz:
### What does this PR do?

Extends the existing `Dockerfile.agent-otel` used for providing BYOC on containers (as documented [here](https://docs.datadoghq.com/opentelemetry/setup/ddot_collector/custom_components/)) to make it able to produce Linux _deb_ and _rpm_ files.

### Motivation

[ABLD-102](https://datadoghq.atlassian.net/browse/ABLD-102) (and proposal described on that ticket).

I validated this locally with some basic smoke testing, and added a couple of build jobs to CI to confirm that the artifacts are built (this is not very thorough testing but it confirms the mechanism essentially works).

- Most of these are currently documented on the proposal doc linked in the Jira card.
- The `--output` option (and possibly some of the things added to the Dockerfile) require buildkit. I don't see a good reason why a user with access to docker would have a problem with this.
- Using `nfpm` instead of omnibus here means less opportunities to reuse certain things (notably around version specification), but trying to make this fit into omnibus looks like more work than it's worth.

### Additional Notes

The expected way to use this is visible in the added CI jobs, looking like this:

```
      docker build \
        --target artifact \
        --output type=local,dest=./ \
        --build-arg AGENT_BRANCH=$CI_COMMIT_REF_NAME \
        --build-arg PACKAGE_TYPE=rpm \
        --build-arg AGENT_VERSION=$AGENT_VERSION \
        -f Dockerfile.agent-otel .
```

There are two main pieces of work left for follow-up PRs:
- Adding dependency related metadata to the packages to have it match what the regular `ddot` package does as well as to make it conflict with ddot itself.
-  Support for selecting the base ubuntu version for the builder, such that the compatibility for the resulting binaries can be adjusted.

[ABLD-102]: https://datadoghq.atlassian.net/browse/ABLD-102?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Dockerfiles/agent-ddot/Dockerfile.agent-otel
```diff
@@ -76,6 +77,48 @@ RUN . venv/bin/activate && dda inv collector.generate
...
+  nfpm package --config nfpm.yaml --target /dist --packager "${PACKAGE_TYPE}"
+else
+  cp /workspace/datadog-agent-${AGENT_BRANCH}/bin/otel-agent/otel-agent /dist/
+fi
+RUNSCRIPT
+
+# Create a stage for extracting the built artifact (binary or packages)
+FROM scratch AS artifact
+ARG AGENT_BRANCH
+COPY --from=builder /dist/* /ddot-byoc/
```
liustanley, line 121: Does this also need to be wrapped in a `if [ -n "$PACKAGE_TYPE" ]`?

liustanley (approved):
LGTM

Dockerfiles/agent-ddot/Dockerfile.agent-otel
alopezz: Not really. If you look at
https://github.com/DataDog/datadog-agent/blob/fe4223b339cc29e3ac93114f1a6ff7139265f445/Dockerfiles/agent-ddot/Dockerfile.agent-otel#L113-L114

When that variable is not defined we simply output the binary.

.gitlab/integration_test/otel.yml
```diff
@@ -62,3 +62,62 @@ datadog_otel_components_ocb_build:
...
+    - |
+      DEB_PATH=ddot-byoc/datadog-agent-ddot-byoc_${AGENT_VERSION}_amd64.deb
+      if [ ! -f "$DEB_PATH" ]; then
+        echo "ERROR: Expected DEB package not found in output directory!" >&2
+        exit 1
+      fi
+      dpkg -c "$DEB_PATH" | grep /opt/datadog-agent/embedded/bin/otel-agent || { echo "ERROR: otel-agent binary not found in DEB package"; exit 1; }
+  rules:
+    - !reference [.except_mergequeue]
+    - when: on_success
```
chouetz: 🥜 
```suggestion
    - when: on_success

```

chouetz (approved):

PR #39985, opened 2025-08-19
[SQG] Measure docker image size in build job

JSGette:
### What does this PR do?
_This is **PART II**, the implementation is not complete yet and not used productively. However to not spawn another PR containing dozens of changes it was decided to bring this work to main step by step._

This PR adds `- !reference [.static_quality_gate_docker]` to the end of the script block of `.docker_build_job_definition` to analyze produced image. To do so we had to introduce pretty drastic changes to our docker build process:

- Change CI Runners type to `docker-in-docker`. Plane CI Runners aren't able to perform any docker commands, therefore, in place analysis was not possible.
- Flatten images in Dockerfiles. `crane` is no longer needed. Main issue with `crane`is that it is unable to work with local registry. So `crane flatten` would download the image from the remote registry, flatten it and upload again, that is wasteful and also led to wrong analysis results as locally present image wasn't flattened.

It is also possible to run analysis locally. Example command, it is important to set the correct gate name that exists in `test/static/static_quality_gates.yml`:
```
dda inv quality-gates.measure-image-local \
    --image-ref "registry.ddbuild.io/ci/datadog-agent/dogstatsd:v74596342-e431809e-amd64" \
    --gate-name "static_quality_gate_docker_cluster_agent_amd64" \
    --build-job-name "local" \
    --output-path "report.yml" \
    --max-files 10000 \
    --no-checksums \
    --debug
```

### Motivation
We want to provide as much information to the developers as possible and also ensure good performance. It is painful right now and mostly reactive work as developers start to respect Static Quality Gates only after they become red in main.


tasks/static_quality_gates/experimental_gates.py
```diff
@@ -37,19 +65,74 @@ def __post_init__(self):
...
+        if not self.image_ref:
+            raise ValueError("image_ref cannot be empty")
+        if not self.architecture:
+            raise ValueError("architecture cannot be empty")
+        if not self.os:
+            raise ValueError("os cannot be empty")
+        if self.config_size < 0:
+            raise ValueError("config_size must be strictly non-negative")
+        if self.manifest_size < 0:
+            raise ValueError("manifest_size must be strictly non-negative")
```
rdesgroppes:
```suggestion
            raise ValueError("config_size must be positive")
        if self.manifest_size < 0:
            raise ValueError("manifest_size must be positive")
```

tasks/static_quality_gates/experimental_gates.py
```diff
@@ -66,41 +149,61 @@ class InPlaceArtifactReport:
     file_inventory: list[FileInfo]
 
     # Metadata
+    # TODO: integrate GateMetricHandler metadata
```
rdesgroppes:
```suggestion
    # TODO(whom-to-contact): integrate GateMetricHandler metadata
```

tasks/static_quality_gates/experimental_gates.py
```diff
@@ -66,41 +149,61 @@ class InPlaceArtifactReport:
...
         if not self.artifact_path:
             raise ValueError("artifact_path cannot be empty")
         if not self.gate_name:
             raise ValueError("gate_name cannot be empty")
         if self.on_wire_size < 0:
-            raise ValueError("on_wire_size must be non-negative")
+            raise ValueError("on_wire_size must be strictly non-negative")
         if self.on_disk_size < 0:
-            raise ValueError("on_disk_size must be non-negative")
+            raise ValueError("on_disk_size must be strictly non-negative")
```
rdesgroppes:
```suggestion
            raise ValueError("on_wire_size must be positive")
        if self.on_disk_size < 0:
            raise ValueError("on_disk_size must be positive")
```

tasks/static_quality_gates/experimental_gates.py
```diff
@@ -118,148 +221,65 @@ def _load_config(self) -> dict[str, Any]:
...
 
-        except Exception as e:
-            raise RuntimeError(f"Failed to extract and analyze package {package_path}: {e}") from e
+        try:
+            sha256_hash = hashlib.sha256()
+            with open(file_path, "rb") as f:
+                # Read in chunks to handle large files
+                for chunk in iter(lambda: f.read(4096), b""):
+                    sha256_hash.update(chunk)
+            return f"sha256:{sha256_hash.hexdigest()}"
```
rdesgroppes:
```suggestion
            with open(file_path, "rb") as f:
                sha256_hash = hashlib.file_digest(f, "sha256")
            return f"sha256:{sha256_hash.hexdigest()}"
```

rdesgroppes (commented):
intermediate review due to:
Image

tasks/unit_tests/experimental_gates_tests.py
```diff
@@ -181,9 +186,9 @@ def test_measure_package_success(self, mock_exists):
         mock_exists.return_value = True
 
         # Mock the optimized extraction and analysis method
-        with patch.object(self.measurer, '_extract_and_analyze_package') as mock_extract_analyze:
+        with patch.object(self.measurer._measurer.processor, 'measure_artifact') as mock_measure_artifact:
             # Create mock measurement
-            mock_measurement = ArtifactMeasurement(
+            mock_measurement = ArtifactMeasurement(  # noqa: F841
```
rdesgroppes:
```suggestion
            _ = ArtifactMeasurement(
```
Should do the trick?

rdesgroppes (approved):
Thank you!

KevinFairise2 (approved):
That looks good to me!
Out of curiosity what will be the end state? I see that we introduce ways to compute the quality gate directly in the build job itself, do we plan to drop the dedicated `static_quality_gate` job and to have each build job fail when the size of the artifact goes above the gate?

JSGette:
> That looks good to me! Out of curiosity what will be the end state? I see that we introduce ways to compute the quality gate directly in the build job itself, do we plan to drop the dedicated `static_quality_gate` job and to have each build job fail when the size of the artifact goes above the gate?

@KevinFairise2 the plan is exactly as you described. We want to move SQG measurements to the builds jobs to save time and ensure correctness. SQG job however will remain, but instead of pulling and manipulating with the artifacts it will rely on YAML reports that contain all necessary information including full file inventory. So SQG will just check size values against max values set in the config and compare file inventory with an ancestor. This way we not only improve the accuracy, but also integrate comparison directly into reports. This will allow developers to see immediately where size increase is coming from. 

Next PR will focus on reports comparison (they are already uploaded to S3 bucket and Gitlab Artifacts), I will also add a new section to the report that is published in PR. And the last PR should remove the old implementation entirely. I didn't want to bring everything at once as it would be too big of a change. This way I can already see if it's reliable enough or not.

L3n41c (approved):

L3n41c (approved):

GeorgeHahn (changes requested):
The `docker_build_agent7_full` step (and related `_arm64` step) does not currently set a target. That, along with the Dockerfile change here, appears to change the container that is built and pushed. This leads to regression detector runs failing with the error `Docker responded with status code 400: no command specified`.

GeorgeHahn (approved):

PR #39987, opened 2025-08-19
(fleet) ensure python scripts are run with umask 022 during install

arbll:
This PR is a followup to https://github.com/DataDog/datadog-agent/pull/37685. While this first PR addressed permission issues with the files created from the script we missed __pycache__, a directory created by the python runtime and apparently also created with 0777 by default if no umask are set.

This PR addresses the issue by spawning a child bash shell that will force umask to 022. Technically processes can be started with a custom umask but this isn't exposed in golang https://github.com/golang/go/issues/56016.

### Motivation

Some security products warn on world writeable dir in `/opt`.


arbll:

BaptisteFoy (approved):

PR #39993, opened 2025-08-19
[incident-42110] Fix gitlab ci compare to

CelianR:
### What does this PR do?

[incident-42110]

This triggers a pipeline by the API since the pipeline is not created anymore when the branch is created.
Unit tests have been updated.

Example [job](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1088881467).

### DIFF
--- a/tasks/pipeline.py
+++ b/tasks/pipeline.py
@@ -856,15 +856,12 @@ def compare_to_itself(ctx):
     try:
         max_attempts = 18
         compare_to_pipeline = None
-        branch_created = False
         for attempt in range(max_attempts):
             print(f"[{datetime.now()}] Waiting 10s for the branch to be created {attempt + 1}/{max_attempts}")
             time.sleep(10)
             if agent.branches.get(new_branch, raise_exception=False):
-                branch_created = True
                 break
-
-        if not branch_created:
+        else:
             print(f"{color_message('ERROR', Color.RED)}: Branch {new_branch} not created", file=sys.stderr)
             raise RuntimeError(f"No branch found for {new_branch}")
 
--- a/tasks/pipeline.py
+++ b/tasks/pipeline.py
@@ -859,7 +859,7 @@ def compare_to_itself(ctx):
         for attempt in range(max_attempts):
             print(f"[{datetime.now()}] Waiting 10s for the branch to be created {attempt + 1}/{max_attempts}")
             time.sleep(10)
-            if agent.branches.get(new_branch):
+            if agent.branches.get(new_branch, raise_exception=False):
                 branch_created = True
                 break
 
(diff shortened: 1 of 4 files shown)


tasks/pipeline.py
```diff
@@ -851,36 +851,28 @@ def compare_to_itself(ctx):
...
-            for pipeline in agent.pipelines.list(ref=new_branch, get_all=True):
-                cancel_pipeline(pipeline)
-            print("Cleaning up git")
-            ctx.run(f"git checkout {current_branch}", hide=True)
-            ctx.run(f"git branch -D {new_branch}", hide=True)
-            ctx.run(f"git push origin :{new_branch}", hide=True)
-            raise RuntimeError(f"No pipeline found for {new_branch}")
+
     try:
+        max_attempts = 18
```
CelianR, line 857: This is still 3m but it should be faster

KevinFairise2 (approved):
The changes looks good, but can we make sure that the job we are fixing is actually executed on that branch 🙏

tasks/pipeline.py
```diff
@@ -851,36 +852,28 @@ def compare_to_itself(ctx):
...
+        if not branch_created:
+            print(f"{color_message('ERROR', Color.RED)}: Branch {new_branch} not created", file=sys.stderr)
+            raise RuntimeError(f"No branch found for {new_branch}")
+
+        print('Branch created, triggering the pipeline')
+
+        # Trigger the pipeline on the last commit of this branch
+        compare_to_pipeline = agent.pipelines.create({'ref': new_branch})
+        print(f"Pipeline created: {compare_to_pipeline.web_url}")
+
```
alopezz: You may want to (optionally) use `for...else` here (https://docs.python.org/3/tutorial/controlflow.html#else-clauses-on-loops):

```suggestion
        for attempt in range(max_attempts):
            print(f"[{datetime.now()}] Waiting 10s for the branch to be created {attempt + 1}/{max_attempts}")
            time.sleep(10)
            if agent.branches.get(new_branch, raise_exception=False):
                break
        else:
            print(f"{color_message('ERROR', Color.RED)}: Branch {new_branch} not created", file=sys.stderr)
            raise RuntimeError(f"No branch found for {new_branch}")

        print('Branch created, triggering the pipeline')

        # Trigger the pipeline on the last commit of this branch
        compare_to_pipeline = agent.pipelines.create({'ref': new_branch})
        print(f"Pipeline created: {compare_to_pipeline.web_url}")

```

alopezz (approved):

tasks/pipeline.py
CelianR: Indeed, nice use case of for else 👌

PR #39997, opened 2025-08-19
[AGENTCFG-377] Only add the secret-generic-connectory side binary to the non-fips agents

rahulkaukuntla:
### What does this PR do?

We discovered that the secret-generic-connectory binary (which is shipped along with the agent) isn't FIPS-encrypted, and so, for now, we will need to avoid shipping it with fips agents.


rdesgroppes (approved):
LG

PR #39999, opened 2025-08-19
Remove unused Windows main for iot-agent

pgimalac:
### What does this PR do?
Remove some logic to build the iot agent on Windows and other binaries in `agent-binaries` in omnibus.

### Motivation
The Windows "agent-binaires" package is made up of a mix of the Windows iot-agent and dogstatsd packages, it is only used by https://github.com/DataDog/datadog-aas-extension/blob/38350ce512c5aae56d379643fc2734bd83a77a8e/node/scripts/build-packages.sh#L43-L44, but actually only the trace-agent and dogstatsd are used.
The Windows iot-agent build is not used for anything else, so a lot of the logic can be removed.

### Additional Notes

Ideally we should stop using the iot-agent software definitely for that, now it's only building the trace-agent, which really doesn't makes sense, but it would be a bigger change.


dustmop (approved):

rdesgroppes (approved):
LG

rdesgroppes (approved):
LGTM, thanks!

PR #40003, opened 2025-08-19
Move diff.go-deps task to go-deps.diff

pgimalac:
### What does this PR do?
Move the `diff.go-deps` task from `diff.py` to `go-deps.py`, and rename it `go-deps.diff`.

### Motivation
Fix ownership, also even just for naming it makes more sense to have it in `go-deps.py`.


GustavoCaso (approved):
LGTM

tasks/diff.py
```diff
@@ -4,256 +4,18 @@
...
-def patch_summary(diff):
-    add_count, remove_count = 0, 0
-    for line in diff.splitlines():
-        if line.startswith("+"):
-            add_count += 1
-        elif line.startswith("-"):
-            remove_count += 1
-    return add_count, remove_count
+def go_deps(ctx):
+    raise Exit("This task has been renamed `go-deps.diff`")
```
GustavoCaso, line 18: Since the task `diff.go-deps` was primarily used in CI, we could probably remove this fallback code 😄 
That way, we do not have to worry about removing this empty task in the future

pgimalac: Yeah I considered it too, but I actually run that task locally so I'd rather make it explicit in case someone else also does 😄

GustavoCaso: I would argue that not many people run that task locally, but if we want to keep it no problemo, remember to remove this empty task in the future 😄

KevinFairise2 (approved):

PR #40004, opened 2025-08-19
Update go mod version to 1.24.0 in otel modules

pgimalac:
### What does this PR do?
Update the `go` directive in `go.mod` files to 1.24.0, for modules that are imported from OTel.

### Motivation
Go 1.25 was released, and OTel updated their version to 1.24, so we can do it too !
It allows using Go 1.24 features / syntaxes in those modules.


louis-cqrl (approved):

pgimalac:
Replaced by https://github.com/DataDog/datadog-agent/pull/40230/s

PR #40005, opened 2025-08-19
Update windows images by default in go update workflow

pgimalac:
### What does this PR do?
Update the `go-update` workflow to update Windows buildimages by default.
Also add a `update_windows_images` on the buildimages update workflow.

### Motivation
The workflows didn't update Windows buildimages, which made Windows job fail, and required pushing another commit.

### DIFF
--- a/tasks/update_go.py
+++ b/tasks/update_go.py
@@ -80,7 +80,7 @@ def update_go(
 
     if image_tag:
         try:
-            update_gitlab_config(".gitlab-ci.yml", image_tag, test=test)
+            update_gitlab_config(".gitlab-ci.yml", image_tag, test=test, windows=True)
         except RuntimeError as e:
             if warn:
                 print(color_message(f"WARNING: {str(e)}", "orange"))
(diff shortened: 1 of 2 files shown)


tasks/update_go.py
```diff
@@ -80,7 +80,7 @@ def update_go(
 
     if image_tag:
         try:
-            update_gitlab_config(".gitlab-ci.yml", image_tag, test=test)
+            update_gitlab_config(".gitlab-ci.yml", image_tag, test=test, windows=True)
```
rdesgroppes, line 83: ~is the task Windows-specific?~
Shouldn't the flag be named `update_windows_images`?

louis-cqrl (approved):

tasks/update_go.py
rdesgroppes: Scratch that too. There doesn't seem to be a mapping between those flags.

rdesgroppes (approved):

PR #40028, opened 2025-08-19
Minor dev docs cleanup

ofek:
### Motivation

- The outdated document was already removed in https://github.com/DataDog/datadog-agent/commit/7c66a0185e0481c1d32c3badd886fc0ec4d45750 but was erroneously brought back by a bot in https://github.com/DataDog/datadog-agent/commit/6705c4e80c18b133e23edca92c76a4c5e19fbf43.
- Reflect the recent [change](https://github.com/DataDog/datadog-agent-dev/pull/174) of command name for consistency.


Ishirui (approved):
Huh, really weird that the bot brought it back... Should we take this as an indication that something is broken with the workflow creating these PRs ? In this case it was only docs, so no real impact, but it would be much more dangerous for this to happen with actual source code.

FlorentClarret:
> Huh, really weird that the bot brought it back... Should we take this as an indication that something is broken with the workflow creating these PRs ? In this case it was only docs, so no real impact, but it would be much more dangerous for this to happen with actual source code.

For posterity, from my slack message: 

> It think they got a conflict because of the change in the first commit and they mistakenly brought it back during the resolution

Also we still have references to this file here and there https://github.com/search?q=repo%3ADataDog%2Fdatadog-agent%20agent_dev_env.md&type=code

Ishirui:
> > Huh, really weird that the bot brought it back... Should we take this as an indication that something is broken with the workflow creating these PRs ? In this case it was only docs, so no real impact, but it would be much more dangerous for this to happen with actual source code.
> 
> For posterity, from my slack message:
> 
> > It think they got a conflict because of the change in the first commit and they mistakenly brought it back during the resolution
> 
> Also we still have references to this file here and there https://github.com/search?q=repo%3ADataDog%2Fdatadog-agent%20agent_dev_env.md&type=code

Tysm for checking Florent, since this was only re-deleting an already-deleted file I thought all references would already have been cleaned up. I'll retract my approval for now, I'd rather we ensure those refs are deleted first :+1:

Ishirui (changes requested):
References to deleted `agent_dev_env.md` file still exist in the repo, thanks @FlorentClarret for checking !

Ishirui (approved):
I think @Enzu83 is on PTO though, someone else from @DataDog/agent-runtimes will need to review

louis-cqrl (approved):

PR #40034, opened 2025-08-20
(fleet) use slog for installer logs, tee them to telemetry

arbll:
### What does this PR do?

This PR:
1. Converts all logging usage in the installer CLI to slog.
2. Adds a telemetry wrapper that forwards the logs to telemetry

### Motivation

- More debugging info for the customer (those logs will be visible in their deployment screen in the future)
- Get some data for situations where the installer CLI gets stuck (today we don't get the trace since it never completes)
- Generally more telemetry information

Note: I used claude code to run the logging migration, be careful with the review ;)


arbll:
will revisit in separate PR

PR #40040, opened 2025-08-20
[POC] Agnostic Checks API

GustavoCaso:
### What does this PR do?

**This code is meant to be used for exploration and validation** 

Exploratory work that aims to implement the Agonostic API described in this RFC https://docs.google.com/document/d/10a2gPqkut0L4H5BfEEooa0okXMDEYTvQdc5enWxBR70/edit

The idea is to remove the coupling between the host application (The Agent) and the integrations. Currently, the Python integrations have access to the internals of the Agent via FFI calls.

This new API, instead of exposing the internals of the Agent uses a defined contract of the payload it expect each integration to return:

```json
{
  "metrics": [
    {
      "type": "gauge",
      "name": "system.disk.total",
      "value": 42,
      "tags": ["device:/dev/sda"]
    }
  ]
}
```

By having an agnostic API for executing checks, we can execute checks in different forms, in this PoC we explore two new integrations loaders:

- shared library loader -> `pkg/collector/dynamiccheck`
- shell script loader  -> `pkg/collector/shellscript`

The code includes a shared library written in Rust using flatbuffers and a shell script. Both integrations read the configuration and return values that the Agent can parse and perform the necessary actions.

### How to try it (Only works on MacOS for this POC)

Inside your `dev/dist` folder, create these files:

```
├── conf.d
│   ├── dummy.d
│   │   └── conf.yaml.default
│   └── rustdisk.d
│       └── conf.yaml.default
└── scripts
    └── dummy.sh
```

For this example, the `rustdisk.d/conf.yaml.default` includes:

```yaml
init_config:
  file_system_global_exclude:
    - sda
instances:
  - min_disk_size: 10 
    all_partitions: true
```

The `dummy.d/conf.yaml.default` includes:

```yaml
init_config:
instances:
  - service_name: web-api 
    threshold: 85
```

The `dummy.sh` is: (ensure the file is executable)

```bash
#!/bin/bash

# Example shell script check that reads YAML configs from environment variables
# and outputs JSON metrics

# Read the YAML configurations from environment variables
INIT_CONFIG="${INIT_CONFIG:-}"
INSTANCE_CONFIG="${INSTANCE_CONFIG:-}"

# Example function to parse YAML (requires yq or similar tool)
# For this example, we'll simulate reading some config values
get_config_value() {
    local config="$1"
    local key="$2"
    # Simple grep-based parsing (in real scenarios, use proper YAML parser)
    echo "$config" | grep -E "^${key}:" | sed "s/${key}:[[:space:]]*//"
}

# Example: Parse some configuration values
# From init config (for demonstration - could be used for scheduling)

# From instance config  
service_name=$(get_config_value "$INSTANCE_CONFIG" "service_name" | tr -d '"' || echo "unknown")
threshold=$(get_config_value "$INSTANCE_CONFIG" "threshold" | tr -d '"' || echo "80")

# Simulate collecting some metrics
cpu_usage=$(echo "scale=2; $RANDOM % 100 + ($RANDOM % 100) / 100" | bc 2>/dev/null || echo "75.5")
memory_usage=$(echo "scale=2; $RANDOM % 100 + ($RANDOM % 100) / 100" | bc 2>/dev/null || echo "45.2")
disk_usage=$(echo "scale=2; $RANDOM % 100 + ($RANDOM % 100) / 100" | bc 2>/dev/null || echo "65.8")

# Output JSON payload
cat <<EOF
{
  "metrics": [
    {
      "type": "gauge",
      "name": "custom.cpu.usage",
      "value": ${cpu_usage},
      "tags": ["service:${service_name}", "source:shellscript"]
    },
    {
      "type": "gauge", 
      "name": "custom.memory.usage",
      "value": ${memory_usage},
      "tags": ["service:${service_name}", "source:shellscript"]
    },
    {
      "type": "gauge",
      "name": "custom.disk.usage", 
      "value": ${disk_usage},
      "tags": ["service:${service_name}", "source:shellscript"]
    },
    {
      "type": "count",
      "name": "custom.check.runs",
      "value": 1,
      "tags": ["service:${service_name}", "threshold:${threshold}"]
    }
  ]
}
EOF
```

From the folder `pkg/collector/dynamicchecks/rustdisk` run `cargo build --profile release` then move the final shared object from `pkg/collector/dynamicchecks/rustdisk/librustdisk.dylib` to `dev/lib`

Build the agent using `inv -e agent.build` and execute using `DD_LOG_LEVEL=debug ./bin/agent/agent run --cfgpath ./bin/agent/dist`


pkg/collector/agonsticapi/check.go
```diff
@@ -0,0 +1,289 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <dlfcn.h>
+#include "loader.h"
+
+extern void close_library(void *handle, const char **error)
+{
+	// Load the shared library
+	if (!handle) {
+		*error = strdup("Error loading library");
```
nathan-b: This error message feels a bit misleading.

pkg/collector/agonsticapi/check.go
```diff
@@ -0,0 +1,289 @@
...
+		return;
+	}
+
+	int result = dlclose(handle);
+	if (result > 0) {
+	 	*error = strdup("Error closing the library");
```
nathan-b: What about including errno or strerror so we can diagnose what the problem was?

pkg/collector/agonsticapi/check.go
```diff
@@ -0,0 +1,289 @@
...
+extern Result* run_agnostic_check(void *handle, const char **error)
+{
+	// Load the shared library
+	if (!handle) {
+		*error = strdup("Error loading library");
+		return NULL;
+	}
+
+	// Clear any previous dlerror
+  dlerror();
```
nathan-b: Nit: Weird indentation here and elsewhere (lines 43, 57, 72, etc...)

pkg/collector/agonsticapi/check.go
```diff
@@ -0,0 +1,289 @@
...
+			snprintf(formatted_error, len, "Error loading symbol 'Run': %s", dl_error);
+			*error = formatted_error;
+			return NULL;
+	}
+
+
+	// Call the function and get the result
+  Result *result = run();
+
+	return result;
```
nathan-b: Why not just `return run();`?

pkg/collector/agonsticapi/check.go
```diff
@@ -0,0 +1,289 @@
...
+	// Check for errors
+	char *dl_error = dlerror();
+	if (dl_error) {
+		size_t len = strlen("Error loading symbol 'FreeResult': ") + strlen(dl_error) + 1;
+		char *formatted_error = malloc(len);
+		snprintf(formatted_error, len, "Error loading symbol 'FreeResult': %s", dl_error);
+		*error = formatted_error;
+		return;
+	}
+
```
nathan-b: Nit: unnecessary whitespace

pkg/collector/agonsticapi/check.go
```diff
@@ -0,0 +1,289 @@
...
+			sender.Historate(name, value, hostn, tags)
+		}
+	}
+
+	return nil
+}
+
+func (c *agnosticCheck) Stop() {}
+
+func (c *agnosticCheck) Cancel() {
```
nathan-b: Is the agent expected to call `Cancel` on each dynamically loaded check on shutdown, or do we just exit and let the OS figure it out? If the former, renaming the API to something more intuitive (`Unload` or `Close` would make more sense to me) might be helpful.

nathan-b (commented):
I left a few comments. I still don't know much, so treat all of them as optional :)

pkg/collector/agonsticapi/check.go
GustavoCaso: Sharing some context: 

Checks are scheduled and unscheduled based on configuration changes. Those changes are monitored and propagated via the Autodiscovery component. 

The autodiscovery component has many providers (file system provider, k8s resources provider, etc..)

Then we have the scheduler, which the Autodiscovery schedules the event into. 

We have a [check scheduler](https://github.com/DataDog/datadog-agent/blob/4859845b7f340b05e3bec2aeddf53ef3d19d4c4f/pkg/collector/scheduler.go#L67) that is responsible for handling the events. 

We configure the scheduler for the Autodiscovery component [here](https://github.com/DataDog/datadog-agent/blob/441f994ebbb0dee24e513f6e8bfece6b7e7dd7a5/cmd/agent/subcommands/run/command.go#L663). 
Under the hood, it talks with the collector component to manage checks.

When unscheduling a check we call [`StopCheck`](https://github.com/DataDog/datadog-agent/blob/4859845b7f340b05e3bec2aeddf53ef3d19d4c4f/pkg/collector/scheduler.go#L114) logic [here](https://github.com/DataDog/datadog-agent/blob/f7f8e9ab4e7ae961c0c1c0ba1a2d1de6cb024a1a/comp/collector/collector/collectorimpl/collector.go#L240)

We then first can [`Stop`](https://github.com/DataDog/datadog-agent/blob/cbc65dc8c8e403204a371689c27148af95a4c25a/pkg/collector/runner/runner.go#L271) on the check and then [`Cancel`](https://github.com/DataDog/datadog-agent/blob/f7f8e9ab4e7ae961c0c1c0ba1a2d1de6cb024a1a/comp/collector/collector/collectorimpl/collector.go#L298) 

When the agent stops the collector, [stop](https://github.com/DataDog/datadog-agent/blob/f7f8e9ab4e7ae961c0c1c0ba1a2d1de6cb024a1a/comp/collector/collector/collectorimpl/collector.go#L194) all checks, but do not cancel them 😢  I guess it leaves the OS to figure out that it can reclaim some memory.


Sorry for the wall of text, but I hope it helps to understand a bit more about checks

As for: 
> If the former, renaming the API to something more intuitive (Unload or Close would make more sense to me) might be helpful.

The check Interface is already [defined](https://github.com/DataDog/datadog-agent/blob/cbc65dc8c8e403204a371689c27148af95a4c25a/pkg/collector/check/check.go#L21). We could rename it, but it might be too much change for just the dynamically loaded ones.

pkg/collector/agonsticapi/loader.h
```diff
@@ -0,0 +1,13 @@
...
+typedef struct
+{
+    const char *data;
+    int len;
+} Result;
+
+extern void *open_library(char *library, const char **error);
+extern void close_library(void *handle, const char **error);
+extern Result *run_agnostic_check(void *handle, const char **error);
+extern void free_result(void *handle, Result *result, const char **error);
```
Enzu83: `loader.h` seems to be only used in `check.go`, is there a reason to have these declarations in a separate file?

pkg/collector/agonsticapi/check.go
```diff
@@ -0,0 +1,283 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+package agonsticapi
```
Enzu83: Nit: small typo in the package name

pkg/collector/agonsticapi/loader.go
```diff
@@ -0,0 +1,90 @@
...
+// Copyright 2016-present Datadog, Inc.
+
+package agonsticapi
+
+/*
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <dlfcn.h>
+#include "loader.h"
```
Enzu83: Is `loader.h` used in `loader.go`?

pkg/collector/agonsticapi/loader.go
```diff
@@ -0,0 +1,90 @@
...
+	tagger "github.com/DataDog/datadog-agent/comp/core/tagger/def"
+	integrations "github.com/DataDog/datadog-agent/comp/logs/integrations/def"
+	"github.com/DataDog/datadog-agent/pkg/aggregator/sender"
+	"github.com/DataDog/datadog-agent/pkg/collector/check"
+	"github.com/DataDog/datadog-agent/pkg/util/option"
+)
+
+const loaderName string = "agnosticapi"
+
+type agonsticAPILoader struct {
```
Enzu83: Nit: small typo here too

pkg/collector/dynamicchecks/rustdisk/src/lib.rs
```diff
@@ -0,0 +1,89 @@
...
+use crate::payload_generated::integrations::PayloadArgs;
+
+#[repr(C)]
+pub struct Result {
+    data: *const u8,
+    len: c_int,
+}
+
+#[unsafe(no_mangle)]
+pub extern "C" fn Run() -> *mut Result {
```
Enzu83: How would you handle passing parameters to checks?

pkg/collector/agonsticapi/check.go
```diff
@@ -0,0 +1,283 @@
...
+	bufferData := C.GoBytes(unsafe.Pointer(result.data), result.len)
+
+	// Deserialize the FlatBuffer payload
+	payload := Integrations.GetRootAsPayload(bufferData, 0)
+
+	for i := 0; i < payload.MetricsLength(); i++ {
+		var metric Integrations.Metric
+		if payload.Metrics(&metric, i) {
+			name := string(metric.Name())
+			metrciType := string(metric.Type())
```
Enzu83: Nit: small typo

pkg/collector/agonsticapi/check.go
```diff
@@ -0,0 +1,283 @@
...
+			case "count":
+				sender.Count(name, value, hostn, tags)
+			case "monotonic_count":
+				sender.MonotonicCountWithFlushFirstValue(name, value, hostn, tags, false)
+			case "counter":
+				sender.Counter(name, value, hostn, tags)
+			case "histogram":
+				sender.Histogram(name, value, hostn, tags)
+			case "historate":
+				sender.Historate(name, value, hostn, tags)
```
Enzu83: Should we handle unknown metric types in the `switch` and write logs for that?

Enzu83 (commented):
Overall very good! No problem with the current implementation.

I'm mostly wondering about the `agnosticapi` package structure and how to handle additional features.

Also left some (useful?) nit typo comments.

pkg/collector/agonsticapi/loader.h
GustavoCaso: purely documentation 😄

pkg/collector/agonsticapi/check.go
GustavoCaso: Of course, this is a POC, so we are not looking for production-ready code, just to get a sense of what the code would look like.

If we decide to go this direction, absolutely, we should handle unknown metrics

pkg/collector/agonsticapi/check.go
```diff
@@ -0,0 +1,283 @@
...
+func (c *agnosticCheck) Run() error {
+	var cErr *C.char
+
+	result := C.run_agnostic_check(c.libHandle, &cErr)
+
+	if cErr != nil {
+		errorString := C.GoString(cErr)
+		defer C.free(unsafe.Pointer(cErr))
+
+		errMsg := fmt.Sprintf("failed to execute `run`  function on library %s: err %s", c.name, errorString)
```
sgnn7:
```suggestion
		errMsg := fmt.Sprintf("failed to execute `Run()` function on library %s: err %s", c.name, errorString)
```

pkg/collector/agonsticapi/check.go
```diff
@@ -0,0 +1,283 @@
...
+		return errors.New(errMsg)
+	}
+
+	if result == nil {
+		return fmt.Errorf("library %s execution did not returned any result", c.name)
```
sgnn7:
```suggestion
		return fmt.Errorf("library %s execution did not return any result", c.name)
```

sgnn7: Is this a true error condition? Would be good to codify this in the spec (and possibly the protocol format)

pkg/collector/agonsticapi/check.go
```diff
@@ -0,0 +1,283 @@
...
+	// Deserialize the FlatBuffer payload
+	payload := Integrations.GetRootAsPayload(bufferData, 0)
+
+	for i := 0; i < payload.MetricsLength(); i++ {
+		var metric Integrations.Metric
+		if payload.Metrics(&metric, i) {
+			name := string(metric.Name())
+			metrciType := string(metric.Type())
+			value := metric.Value()
+			hostn, _ := hostname.Get(context.TODO())
```
sgnn7: Resolving the hostname for each metric seems suboptimal

pkg/collector/agonsticapi/check.go
```diff
@@ -0,0 +1,283 @@
...
+			tags := []string{}
+
+			for i := range metric.TagsLength() {
+				tags = append(tags, string(metric.Tags(i)))
+			}
+
+			// We need to figure out what tags we need to fetch from the Agent
+			extraTags, err := c.tagger.GlobalTags(types.LowCardinality)
+			if err != nil {
+				fmt.Printf("failed to get global tags: %s\n", err.Error())
```
sgnn7: More of a question than a comment: should we drop the metric if we can't global tags?

pkg/collector/agonsticapi/check.go
```diff
@@ -0,0 +1,283 @@
...
+			value := metric.Value()
+			hostn, _ := hostname.Get(context.TODO())
+			tags := []string{}
+
+			for i := range metric.TagsLength() {
+				tags = append(tags, string(metric.Tags(i)))
+			}
+
+			// We need to figure out what tags we need to fetch from the Agent
+			extraTags, err := c.tagger.GlobalTags(types.LowCardinality)
```
sgnn7: Fetching global tags for each metric seems suboptimal

pkg/collector/agonsticapi/check.go
```diff
@@ -0,0 +1,283 @@
...
+				tags = append(tags, extraTags...)
+			}
+
+			fmt.Printf("Metric: %s = %f (type: %s, tags: %v)\n", name, value, metrciType, tags)
+			sender, err := c.sendermanager.GetSender(c.id)
+			if err != nil {
+				return fmt.Errorf("failed to get sender manager for shared library check %s: %v", c.name, err)
+			}
+
+			switch metrciType {
```
sgnn7:
```suggestion
			switch metricType {
```

pkg/collector/agonsticapi/check.go
```diff
@@ -0,0 +1,283 @@
...
+			case "gauge":
+				sender.Gauge(name, value, hostn, tags)
+			case "rate":
+				sender.Rate(name, value, hostn, tags)
+			case "count":
+				sender.Count(name, value, hostn, tags)
+			case "monotonic_count":
```
sgnn7: Style preference: Is there another type of `monotonic` value? If not, just `monotonic` should be fine

pkg/collector/agonsticapi/check.go
```diff
@@ -0,0 +1,283 @@
...
+func (c *agnosticCheck) Cancel() {
+	var cErr *C.char
+
+	C.close_library(c.libHandle, &cErr)
+
+	if cErr != nil {
+		errorString := C.GoString(cErr)
+		defer C.free(unsafe.Pointer(cErr))
+
+		fmt.Printf("Warning: failed to cancel the shared library check %s: %s\n", c.name, errorString)
```
sgnn7: nit: Should include the library name

pkg/collector/agonsticapi/loader.go
```diff
@@ -0,0 +1,90 @@
...
+
+func (sl *agonsticAPILoader) String() string {
+	return "Agnostic API Loader"
+}
+
+// Load returns a Shared Library check
+func (sl *agonsticAPILoader) Load(senderManager sender.SenderManager, config integration.Config, instance integration.Data) (check.Check, error) {
+	var cErr *C.char
+
+	name := "lib" + config.Name + ".dylib"
```
sgnn7: Not for a PoC but we will need to extend things to non-Darwin systems at some point.

pkg/collector/dynamicchecks/rustdisk/src/lib.rs
```diff
@@ -0,0 +1,89 @@
...
+    builder.finish(payload, None);
+    let buf = builder.finished_data();
+
+    let buf_vec = buf.to_vec();
+    let result = Result {
+        data: buf_vec.as_ptr(),
+        len: buf_vec.len() as c_int,
+    };
+
+    // Keep the Vec alive by leaking it - Go will need to access this data
```
sgnn7: Seems a bit of an odd pattern but I'm not familiar with best practices in CGo <-> Rust interactions

sgnn7 (commented):
Left some comments - definitely a plausible path forward :)

pkg/collector/dynamicchecks/rustdisk/src/lib.rs
GustavoCaso: I added some code purely poc to handle configuration https://github.com/DataDog/datadog-agent/pull/40040/commits/d93f0ff6eaec0ebe1db31240bdef56e014aa7d62

The gist is: We store the file as a flatbuffer for each instance, and the shared libraries read

pkg/collector/dynamicchecks/rustdisk/src/lib.rs
Enzu83: Leaking the Vec is the way to give its "ownership" to Cgo, supposing you'll eventually deallocate it later by calling the `Free` function.

Here's a [link](https://stackoverflow.com/questions/39224904/how-to-expose-a-rust-vect-to-ffi) about the same use case.

pkg/collector/dynamicchecks/rustdisk/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn FreeResult(result: *mut Result) {
+    if result.is_null() {
+        return;
+    }
+
+    unsafe {
+        // Take ownership of the Result to properly drop the Vec<u8>
+        let _ = Box::from_raw(result);
+        // The Vec<u8> will be automatically dropped when the Box goes out of scope
```
tobz: This isn't working how you think it is. 😅 

The `Vec<u8>` is indeed kept alive by calling `forget` it, but holding a pointer to the underlying allocation in a boxed `Result` doesn't do anything to deallocate it when you deallocate the `Box<Result`>`. It just stays leaked.

You're also allocating separately just for the `Result` struct, which is somewhat suboptimal.

I think you'd be better off if you actually exposed a method to allocating the result container ahead of time, and then required passing a pointer to it into `Run`. This would let you change `Result` to fully mirror `Vec<u8>` (including the capacity, which you need to reconstitute `Vec<u8>` from raw parts), skipping the second allocation and making it so `FreeResult` actually frees the underlying allocation.

PR #40043, opened 2025-08-20
[SBOM] store SBOM compressed in the workloadmeta

paulcacheux:
### What does this PR do?

Currently we store full SBOMs for all container image metadata in the workloadmeta store. This is used both by the flare provider (that creates a `sbom.json` file in the flare) and also by the periodic refresh that will send the SBOM for all images periodically.

Now SBOMs can be big (a few megabytes), and we expect them to get bigger when we enable more scanners (like the apps and libraries scanners). This is especially impactful on nodes with a lot of images thanks to binpacking (internally we call this nodeless), and can result in agent OOM kills.

To improve this state, this PR converts the workloadmeta store to store SBOMs in a proto-encoded + compressed format. SBOMs are decompressed on demand when they are needed (which should be pretty rare).

### DIFF
--- a/comp/core/workloadmeta/collectors/sbomutil/compress.go
+++ b/comp/core/workloadmeta/collectors/sbomutil/compress.go
@@ -1,0 +1,79 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2023-present Datadog, Inc.
+
+package sbomutil
+
+import (
+	"bytes"
+	"compress/gzip"
+	"io"
+
+	"github.com/DataDog/agent-payload/v5/cyclonedx_v1_4"
+	workloadmeta "github.com/DataDog/datadog-agent/comp/core/workloadmeta/def"
+	"google.golang.org/protobuf/proto"
+)
+
+// CompressSBOM converts a workloadmeta.SBOM into a workloadmeta.CompressedSBOM.
+func CompressSBOM(sbom *workloadmeta.SBOM) (*workloadmeta.CompressedSBOM, error) {
+	if sbom == nil {
+		return nil, nil
+	}
+
+	uncompressedBom, err := proto.Marshal(sbom.CycloneDXBOM)
+	if err != nil {
+		return nil, err
+	}
+
+	var compressedBom bytes.Buffer
+	writer := gzip.NewWriter(&compressedBom)
+	if _, err := writer.Write(uncompressedBom); err != nil {
+		return nil, err
+	}
+	if err := writer.Close(); err != nil {
+		return nil, err
+	}
+
+	return &workloadmeta.CompressedSBOM{
+		Bom:                compressedBom.Bytes(),
+		GenerationTime:     sbom.GenerationTime,
+		GenerationDuration: sbom.GenerationDuration,
+		Status:             sbom.Status,
+		Error:              sbom.Error,
+	}, nil
+}
+
+// UncompressSBOM converts a workloadmeta.CompressedSBOM into a workloadmeta.SBOM.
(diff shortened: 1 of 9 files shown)


spikat (approved):

safchain (approved):

comp/core/workloadmeta/collectors/sbomutil/compress.go
```diff
@@ -0,0 +1,79 @@
...
+	"compress/gzip"
+	"io"
+
+	"github.com/DataDog/agent-payload/v5/cyclonedx_v1_4"
+	workloadmeta "github.com/DataDog/datadog-agent/comp/core/workloadmeta/def"
+	"google.golang.org/protobuf/proto"
+)
+
+// CompressSBOM converts a workloadmeta.SBOM into a workloadmeta.CompressedSBOM.
+func CompressSBOM(sbom *workloadmeta.SBOM) (*workloadmeta.CompressedSBOM, error) {
```
adel121, line 19: This function marshals the payload, and also compresses it. 

Do you think this can also cause extra CPU at compression time? 

I am not sure about the throughput at which SBOM payloads are received so not sure how to judge.

paulcacheux: The scanning process is already a CPU intensive process, in addition to the conversion between 3 formats of SBOMs we do after it. This marshalling + compression ends up in the noise from my testing.

adel121:
The idea makes sense to me. 👍 

But I have some concerns:

1. Storing compressed payloads in workloadmeta means that we never have direct access to filterable data. In other words. a workloadmeta subscriber can't choose to subscribe to SBOM payloads satisfying a given filter because it is not possible to apply filter functions on compressed payloads. 

2. If multiple subscribers consume the compressed payloads from workloadmeta, how/when are we going to decompress? If I am not mistaken, the current code will result in notifying the subscribers with the compressed data. It is the responsibility of the subscriber to decompress. Which might mean we will be decompressing the same payload multiple times. 

3. Compressing the payload indeed reduces the memory footprint, but this is only true during the time the payloads are compressed, right? Once the flare is generated, we will need to decompress all payloads, so at some point we will be increasing memory. I understand that this will be a temporary memory increase, but how does this solve the OOM issue? I think the agent container will still suffer from OOM in case the peak memory is high enough, even if it is for a short period of time. 

But I am not very familiar with SBOM use cases so I am not even sure these use case exists (or may exist in the forseeable future).

paulcacheux:
> Storing compressed payloads in workloadmeta means that we never have direct access to filterable data. In other words. a workloadmeta subscriber can't choose to subscribe to SBOM payloads satisfying a given filter because it is not possible to apply filter functions on compressed payloads.

I'm not sure this is a real concern today, AFAIK there is no wm subscriber that is filtering SBOMs or even interested in SBOM themselves

> If multiple subscribers consume the compressed payloads from workloadmeta, how/when are we going to decompress? If I am not mistaken, the current code will result in notifying the subscribers with the compressed data. It is the responsibility of the subscriber to decompress. Which might mean we will be decompressing the same payload multiple times.

Yes, with this PR this would be the result. Similar to the first question, there is no wm subscriber that is interested by SBOMs currently.

> Compressing the payload indeed reduces the memory footprint, but this is only true during the time the payloads are compressed, right? Once the flare is generated, we will need to decompress all payloads, so at some point we will be increasing memory. I understand that this will be a temporary memory increase, but how does this solve the OOM issue? I think the agent container will still suffer from OOM in case the peak memory is high enough, even if it is for a short period of time.

Yes this is true. By this means that the peak of memory use is scoped to a short slice of time as opposed to continuously as is the case today. This also offers more improvements opportunities that I want to tackle next:
- for the flare: the SBOMs are uncompressed one by one. We could dump this do disk after each to limit the memory used for example
- for the periodic refresh: similarly the SBOMs are uncompressed and sent again to the backend one by one, so we could do this slowly to not

adel121 (approved):

PR #40047, opened 2025-08-20
[CWS-4782] Add eBPFLess support for setsockopt

theop-dd:
### What does this PR do?
Add eBPFLess support for setsockopt
### Motivation
Have setsockopt detection possible on Fargate.

### DIFF
--- a/pkg/security/ptracer/network_handlers.go
+++ b/pkg/security/ptracer/network_handlers.go
@@ -11,9 +11,10 @@ package ptracer
 import (
 	"encoding/binary"
 	"errors"
-	"golang.org/x/sys/unix"
 	"net"
 	"syscall"
+
+	"golang.org/x/sys/unix"
 
 	"github.com/DataDog/datadog-agent/pkg/security/proto/ebpfless"
 )
@@ -49,6 +50,12 @@ func registerNetworkHandlers(handlers map[int]syscallHandler) []string {
 			Func:       handleSocket,
 			ShouldSend: nil,
 			RetFunc:    handleSocketRet,
+		},
+		{
+			ID:         syscallID{ID: SetsockoptNr, Name: "setsockopt"},
+			Func:       handleSetsockopt,
+			ShouldSend: shouldSendSetsockopt,
+			RetFunc:    nil,
 		},
 	}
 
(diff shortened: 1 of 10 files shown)


pkg/security/ptracer/network_handlers.go
```diff
@@ -274,3 +275,78 @@ func shouldSendAccept(msg *ebpfless.SyscallMsg) bool {
...
+		FilterLen:      fLen,
+		SocketFamily:   socketInfo.AddressFamily,
+		SocketProtocol: socketInfo.Protocol,
+		SocketType:     socketInfo.SocketType,
+	}
+	return nil
+}
+
+func shouldSendSetsockopt(msg *ebpfless.SyscallMsg) bool {
+	return msg.Retval >= 0 || msg.Retval == -int64(syscall.EACCES) || msg.Retval == -int64(syscall.EPERM) || msg.Retval == -int64(syscall.EINVAL)
```
spikat: Not sure we want to send the msg when syscall returns EINVAL. If not, we may use the generic `isAcceptedRetval` func.

pkg/security/ptracer/network_handlers.go
```diff
@@ -274,3 +275,78 @@ func shouldSendAccept(msg *ebpfless.SyscallMsg) bool {
...
+		return errors.New("short read from remote process")
+	}
+	return nil
+}
+
+func handleSetsockopt(tracer *Tracer, process *Process, msg *ebpfless.SyscallMsg, regs syscall.PtraceRegs, _ bool) error {
+	fd := int32(tracer.ReadArgUint32(regs, 0))
+	socketInfo, ok := process.FdToSocket[fd]
+	if !ok {
+		return errors.New("unable to find socket")
```
spikat: Even without the socket info, we still can get the level/optname and bpffilter (if any), I guess we should send the event with incomplete fields in this case, it still have value.
Also, did you look if there is a way to fallback on proc to determine the socket type from the fd?

pkg/security/ptracer/network_handlers.go
```diff
@@ -274,3 +275,78 @@ func shouldSendAccept(msg *ebpfless.SyscallMsg) bool {
...
+	level := tracer.ReadArgUint32(regs, 1)
+	optname := tracer.ReadArgUint32(regs, 2)
+
+	// Read the sock_fprog header from argument 3 (optval)
+	// Expected size on 64-bit: 16 bytes (2 bytes length, 6 bytes padding, 8 bytes pointer)
+	hdr, err := tracer.ReadArgData(process.Pid, regs, 3, 16)
+	if err != nil {
+		return err
+	}
+	fLen := binary.NativeEndian.Uint16(hdr[0:2])
```
spikat, line 345: Shouldn't you also check the `optlen` ?
Also, could we skip trying to look at `optval` if the `optname` is not about setting a bpf filter?

pkg/security/ptracer/network_handlers.go
theop-dd: yes you are probably right

pkg/security/ptracer/network_handlers.go
theop-dd: Last I checked I have not found anything that could link the fd to the socket type directly. Maybe I missed something tho

pkg/security/ptracer/network_handlers.go
theop-dd: for optname yes I agree, What do you mean by checking optlen ?

spikat (approved):

PR #40061, opened 2025-08-20
Log Permissions error doc link

kelvin-lal:
### What does this PR do?
1. Adding an additional doc link when a permissions issue is evaluated for log files within the status.log
2. If statement to only trigger the relative doc link when a permission issue is present within the error message
### Motivation
- Agent Log improvement project
<!--
Validate your changes before merge, ensuring that:
- PR was tested in DDA environment with test log file with incorrect permissions
- The permissions error was triggered after this fact and output the doc link within the status.log

- N/A
### Additional Notes
- Keeping the initial error code int he else statement in case there is a different error other than permissions


jszwedko (approved):

kelvin-lal:
Redoing this pr separately

PR #40064, opened 2025-08-20
feat(ssi): allow configuring the library storage medium

betterengineering:
### What does this PR do?
This change allows customers to configure the storage medium for the  library storage `emptyDir`. The default will remain a file backed directory, and users can optionally use memory instead.

### Motivation
As part of Single Step Instrumentation in Kubernetes, we utilize init containers and an `emptyDir` to copy tracing libraries into place for the application to use. Historically, this has utilized disk space.

There is a customer who has a storage constrained environment that is looking to use Single Step Instrumentation. We want to allow them to use a memory backed `emptyDir` instead of a disk backed directory. This will allow them to consume memory instead of disk for their deployment.

The additional unit tests cover the feature from a configuration standpoint. Additionally, I've used `injector-dev` to run the following:

```bash
injector-dev apply -f storage-medium.yaml --build
```

<details>
  <summary>Test Scenario</summary>

```yaml
helm:
  apps:
    - name: java-app
      namespace: application
      values:
        env:
          - name: DD_TRACE_DEBUG
            value: "true"
            value: "true"
        image:
          repository: registry.ddbuild.io/ci/injector-dev/java
          tag: 95586dc4
        podLabels:
          language: "java"
          tags.datadoghq.com/env: local
        service:
          port: "8080"
    - name: java-app-with-resources
      namespace: application
      values:
        env:
          - name: DD_TRACE_DEBUG
            value: "true"
            value: "true"
        image:
          repository: registry.ddbuild.io/ci/injector-dev/java
          tag: 95586dc4
        podLabels:
          language: "java"
          tags.datadoghq.com/env: local
        service:
          port: "8080"
        resources:
          limits:
            memory: 500Mi
            cpu: 1
          requests:
            memory: 500Mi
            cpu: 1
    - name: java-app-constrained
      namespace: application
      values:
        env:
          - name: DD_TRACE_DEBUG
            value: "true"
            value: "true"
        image:
          repository: registry.ddbuild.io/ci/injector-dev/java
          tag: 95586dc4
        podLabels:
          language: "java"
          tags.datadoghq.com/env: local
        service:
          port: "8080"
        resources:
          limits:
            memory: 100Mi
            cpu: 100m
          requests:
            memory: 100Mi
            cpu: 100m
  versions:
    agent: 7.69.1
    cluster_agent:
      version: 7.69.1
      build: {}
    injector:
      version: 0.44.0
  config:
    clusterAgent:
      env:
        - name: "DD_ADMISSION_CONTROLLER_AUTO_INSTRUMENTATION_VOLUME_EMPTY_DIR_STORAGE_MEDIUM"
          value: "memory"
    datadog:
      apm:
        instrumentation:
          enabled: true
          targets:
            - name: java
              podSelector:
                matchLabels:
                  language: "java"
              ddTraceVersions:
                java: "1.52.1"
```

</details>

It's worth noting here and in our documentation that memory backed emptyDir will count against the memory limit of the pod. Some languages utilize the full limit of memory to configure heap sizes and will not account for the additional overhead of the tracing libraries. In these cases, 

### Additional Notes

To enable the memory backed storage location, administrators should add the [Cluster Agent environment variable](https://docs.datadoghq.com/containers/cluster_agent/commands/?tab=datadogoperator#cluster-agent-environment-variables):

```yaml
- name: "DD_ADMISSION_CONTROLLER_AUTO_INSTRUMENTATION_VOLUME_EMPTY_DIR_STORAGE_MEDIUM"
   value: "memory" 
```

### DIFF
--- a/pkg/clusteragent/admission/mutate/autoinstrumentation/config.go
+++ b/pkg/clusteragent/admission/mutate/autoinstrumentation/config.go
@@ -10,6 +10,7 @@ package autoinstrumentation
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
@@ -69,6 +70,9 @@ type Config struct {
 	// We don't expose this option to the user, and [[instrumentationV1]]
 	// is deprecated and slated for removal.
 	version version
+
+	// libraryStorageMedium is the storage medium to use for the library storage volume emptyDir.
+	libraryStorageMedium corev1.StorageMedium
 }
 
 var excludedContainerNames = map[string]bool{
@@ -102,6 +106,11 @@ func NewConfig(datadogConfig config.Component) (*Config, error) {
 		return nil, err
 	}
 
+	libraryStorageMedium, err := parseLibraryStorageMedium(datadogConfig)
+	if err != nil {
+		return nil, err
+	}
+
 	defaultResourceRequirements, err := initDefaultResources(datadogConfig)
 	if err != nil {
 		return nil, fmt.Errorf("unable to parse init-container's resources from configuration: %w", err)
@@ -120,6 +129,7 @@ func NewConfig(datadogConfig config.Component) (*Config, error) {
 		profilingClientLibraryMutator: profilingClientLibraryConfigMutators(datadogConfig),
 		containerFilter:               excludedContainerNamesContainerFilter,
 		version:                       version,
+		libraryStorageMedium:          libraryStorageMedium,
 	}, nil
 }
 
@@ -362,6 +372,24 @@ func getOptionalBoolValue(datadogConfig config.Component, key string) *bool {
 	}
 
 	return value
+}
+
+func parseLibraryStorageMedium(datadogConfig config.Component) (corev1.StorageMedium, error) {
+	confKey := "admission_controller.auto_instrumentation.volume.empty_dir_storage_medium"
+	if !datadogConfig.IsSet(confKey) {
+		return corev1.StorageMediumDefault, nil
+	}
+
+	confValue := datadogConfig.GetString(confKey)
+
+	switch strings.ToLower(confValue) {
+	case "", "default":
+		return corev1.StorageMediumDefault, nil
+	case "memory":
+		return corev1.StorageMediumMemory, nil
+	default:
+		return corev1.StorageMediumDefault, fmt.Errorf("invalid library storage medium: %s", confValue)
+	}
 }
 
 // getOptionalBoolValue returns a pointer to a bool corresponding to the config value if the key is set in the config
--- a/pkg/clusteragent/admission/mutate/autoinstrumentation/injector.go
+++ b/pkg/clusteragent/admission/mutate/autoinstrumentation/injector.go
@@ -12,8 +12,9 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/DataDog/datadog-agent/pkg/util/log"
 	corev1 "k8s.io/api/core/v1"
+
+	"github.com/DataDog/datadog-agent/pkg/util/log"
 )
 
 const (
@@ -29,48 +30,6 @@ func injectorFilePath(name string) string {
 	return injectPackageDir + "/stable/inject/" + name
 }
 
-var sourceVolume = volume{
-	Volume: corev1.Volume{
-		Name: volumeName,
-		VolumeSource: corev1.VolumeSource{
-			EmptyDir: &corev1.EmptyDirVolumeSource{},
-		},
-	},
-}
-
-var v1VolumeMount = sourceVolume.mount(corev1.VolumeMount{
-	MountPath: mountPath,
-})
-
-var v2VolumeMountInjector = sourceVolume.mount(corev1.VolumeMount{
-	MountPath: asAbs(injectPackageDir),
-	SubPath:   injectPackageDir,
-})
-
-var v2VolumeMountLibrary = sourceVolume.mount(corev1.VolumeMount{
-	MountPath: asAbs(libraryPackagesDir),
-	SubPath:   libraryPackagesDir,
-})
-
-var etcVolume = volume{
-	Volume: corev1.Volume{
-		Name: "datadog-auto-instrumentation-etc",
-		VolumeSource: corev1.VolumeSource{
-			EmptyDir: &corev1.EmptyDirVolumeSource{},
-		},
-	},
-}
-
-var volumeMountETCDPreloadInitContainer = etcVolume.mount(corev1.VolumeMount{
-	MountPath: "/datadog-etc",
-})
-
-var volumeMountETCDPreloadAppContainer = etcVolume.mount(corev1.VolumeMount{
-	MountPath: "/etc/ld.so.preload",
-	SubPath:   "ld.so.preload",
-	ReadOnly:  true,
-})
-
 type injector struct {
 	image      string
 	registry   string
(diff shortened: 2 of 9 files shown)


pkg/clusteragent/admission/mutate/autoinstrumentation/config.go
```diff
@@ -373,6 +384,28 @@ func getOptionalBoolValue(datadogConfig config.Component, key string) *bool {
 	return value
 }
 
+func setStorageMedium(medium corev1.StorageMedium) {
```
betterengineering: This certainly has a smell to it. The alternative though is to conduct a pretty large refactor to remove global volume variables and pass the medium through composition.

pkg/config/setup/config.go
```diff
@@ -818,6 +818,7 @@ func InitConfig(config pkgconfigmodel.Setup) {
 	config.BindEnv("admission_controller.auto_instrumentation.init_resources.cpu")
 	config.BindEnv("admission_controller.auto_instrumentation.init_resources.memory")
 	config.BindEnv("admission_controller.auto_instrumentation.init_security_context")
+	config.BindEnv("admission_controller.auto_instrumentation.library_storage_medium")
```
stanistan: All of the other flags for this are `init_...` so this is odd being _different_, and also this doesn't apply to the init container but the volume. 

Given we might end up different actual storage drivers, volume sources. Should this be more explicit in its name?

```
auto_instrumentation.volume.empty_dir_storage_medium`
```

I know that's a lot longer, but it's much more explicit to what this does

pkg/clusteragent/admission/mutate/autoinstrumentation/injector.go
```diff
@@ -29,11 +30,15 @@ func injectorFilePath(name string) string {
 	return injectPackageDir + "/stable/inject/" + name
 }
 
+// Note: The storage medium is set in the config via setStorageMedium. This is a global variable
+// that should be set through composition.
```
stanistan: I would move the function that sets the global variable `medium` here so its local to what it changes.

pkg/clusteragent/admission/mutate/autoinstrumentation/config.go
```diff
@@ -110,6 +114,12 @@ func NewConfig(datadogConfig config.Component) (*Config, error) {
 		return nil, err
 	}
 
+	libraryStorageMedium, err := parseLibraryStorageMedium(datadogConfig)
+	if err != nil {
+		return nil, err
+	}
+	setStorageMedium(libraryStorageMedium)
```
stanistan: Feel free to ignore for this:

---

The one thing I can think of to make this _kind of mutator-y_ if you were feeling the vibe for the future, is to have a `libraryStorageVolumeOption`. 

This has a bit more of a pattern for this to change in the future.

```go
type libraryStorageVolumeOption func(*corev1.Volume)

func libraryStorageOptionEmptyDirMedium(/*medium*/) libraryStorageOptionVolume {
   return func(v *corev1.Volume) { 
      // the stuff 
   }
}
```

releasenotes-dca/notes/add-library-storage-medium-config-option-1b8e03acf0b2ac3d.yaml
```diff
@@ -0,0 +1,13 @@
+---
+features:
+  - |
+    A new config option has been added to control the library storage medium for the auto-instrumentation webhook. This
+    will allow uses to utilize a memory backed emptyDir for environments that are storage constrained. The default
```
maycmlee:
```suggestion
    A new configuration option has been added to control the library storage medium for the auto-instrumentation webhook. This
    will allow you to utilize a memory backed emptyDir for environments that are storage constrained. The default
```

releasenotes-dca/notes/add-library-storage-medium-config-option-1b8e03acf0b2ac3d.yaml
```diff
@@ -0,0 +1,13 @@
...
+    behavior is to use a disk backed emptyDir for library storage. To enable memory backed emptyDir, set the following
+    config option for the cluster agent:
```
maycmlee:
```suggestion
    configuration option for the cluster agent:
```

maycmlee (approved):
A couple of small suggestions, but approving to not block merge

pkg/clusteragent/admission/mutate/autoinstrumentation/injector.go
betterengineering: I updated so we didn't need globals 👍

knusbaum (approved):

pkg/config/setup/config.go
betterengineering: Updated!

pkg/clusteragent/admission/mutate/autoinstrumentation/injector.go
```diff
@@ -29,64 +30,82 @@ func injectorFilePath(name string) string {
...
+	return sourceVolume(medium).mount(corev1.VolumeMount{
+		MountPath: mountPath,
+	})
+}
 
-var v2VolumeMountInjector = sourceVolume.mount(corev1.VolumeMount{
-	MountPath: asAbs(injectPackageDir),
-	SubPath:   injectPackageDir,
-})
+func v2VolumeMountInjector(medium corev1.StorageMedium) volumeMount {
```
stanistan, line 61: While in terms of functions, you want these to be the right mount, and you need the `medium` to create the volume, you don't need a `medium` in general for a volume mount, so it's a bit of a smell in terms of passing that around for all of the volume mounts.

betterengineering: I'm getting conflicting feedback on this one. @knusbaum didn't love overriding the global, so we updated to this approach. The benefit being we're passing a config option around instead of mutating global state. To be honest, I don't love either approach but I also don't love the idea of a heavy overhaul for this change that we want to make quickly and be confident.

stanistan: The one alternative I can see here is _completely removing the global_ and setting the `libVolume` mutator as a property on the configuration given this config?

But I think we should see if this solves the problem for users _first_ before we can consider refactoring it to the way it _needs to be_.

stanistan (approved):

adel121:
Small question. 

Not sure if I am missing something, but did we consider that this might result in OOM instead of disk exhaustion ? 

Also, did we consider language detection ? It might be a quick win here as it will probably reduce the number of libraries installed.

pkg/clusteragent/admission/mutate/autoinstrumentation/config.go
```diff
@@ -373,6 +383,24 @@ func getOptionalBoolValue(datadogConfig config.Component, key string) *bool {
 	return value
 }
 
+func parseLibraryStorageMedium(datadogConfig config.Component) (corev1.StorageMedium, error) {
+	confKey := "admission_controller.auto_instrumentation.volume.empty_dir_storage_medium"
+	if !datadogConfig.IsSet(confKey) {
```
dustmop, line 379: I would recommend you use `IsConfigured` here instead:
```suggestion
	if !datadogConfig.IsConfigured(confKey) {
```
The different is that `IsSet` returns true if the config setting has *any* value from *any* source, including a default value. On the other hand, `IsConfigured` returns true if the setting has any non-default value assigned to it. That would let you replace [this declaration](https://github.com/DataDog/datadog-agent/pull/40064/files#diff-a7b6619528f155e64cefe12cc4ba753eb7688c87fdf9d96e1052a12eeb519df5R821) with a default value, which is recommended.

dustmop (approved):
Approving, with one optional suggestion.

pkg/clusteragent/admission/mutate/autoinstrumentation/injector.go
knusbaum: Yes, I agree.

We can avoid constructing these everywhere and passing arguments around by constructing them and making them properties of the volumes or libInfo or something.

Right now, though this is the least ugly, least fragile implementation. We plan to do some refactoring on this code in the near term.

betterengineering:
FYI - I've updated the base of this PR to merge it into `ssi-memory-backed-storage` so we have a feature branch for this customer.

PR #40065, opened 2025-08-20
RC-1858 Remote-Config transactional store to have exclusive boltDB ownership

aya-shimizu24:
What does this PR do?
---------------------
The core changes are that I am _removing_ the `boltDB` reference from the CoreAgentService's `Service`.
Any interaction with the boltDB should be done via the transactionalStore

I also had to update how the tests create a core agent, it would previously:
`create RPC/HTTP service` -> `create uptane client` -> `create transactional store` -> `open boltdb`
then replace the uptane client with a mocked client
effect was that boltdb would never get closed
solution: create a uptaneFactory for tests, so `create RPC/HTTP service` -> `create mocked uptane client`

Motivation
----------
We've had error logs when clients try to connect to the agent service after it's shut down, so first this can potentially limit spots where the client may be directly calling the service is shut down (so we can stop client from accessing earlier)
Encapsulates boltDB within the transactional store to enforce safety & reduce error risk.

Testing Guidelines
------------------

Prod Release Checklist
----------------------

- [ ] Changes merged to and tested on staging before production
- [ ] I have a plan to verify that these changes are working as expected after merge
- [ ] I have a plan for how to revert these changes if things are not working as expected

Additional Notes
----------------


pkg/config/remote/service/service.go
```diff
@@ -651,8 +649,8 @@ func (s *CoreAgentService) Stop() error {
 	if s.stopConfigPoller != nil {
 		close(s.stopConfigPoller)
 	}
-
-	return s.db.Close()
+	// close boltDB via the transactional store
+	return s.uptane.TransactionalStore().Close()
```
mellon85: can't we make add a Close and Recreate to the uptane thing and make it close the underlying store by itself?
It should allow us to avoid exposing the TransactionalStore struct from the uptane module

pkg/config/remote/service/service.go
```diff
@@ -1227,20 +1216,19 @@ type HTTPClient struct {
...
 func NewHTTPClient(runPath, site, apiKey, agentVersion string) (*HTTPClient, error) {
 	dbPath := path.Join(runPath, "remote-config-cdn.db")
-	db, err := openCacheDB(dbPath, agentVersion, apiKey, site)
+	transactionalStore, err := uptane.NewTransactionalStore(dbPath, agentVersion, apiKey, site)
 	if err != nil {
 		return nil, err
 	}
 
-	uptaneCDNClient, err := uptane.NewCDNClient(db, site, apiKey)
+	uptaneCDNClient, err := uptane.NewCDNClient(transactionalStore, apiKey, site)
```
mellon85: Can we push the initialization of the store inside the CDN client to avoid exposing it?

BTW as it's currently unused we might want to schedule it for removal

pkg/config/remote/service/service.go
```diff
@@ -406,6 +407,11 @@ func WithAgentPollLoopDisabled() func(s *options) {
 	}
 }
 
+// WithUptaneFactory creates a mock version for testing
+func WithUptaneFactory(f func(md *uptane.Metadata) (coreAgentUptaneClient, error)) Option {
```
mellon85: if it's for testing only and to be used only in this package i'll avoid exposing it

```suggestion
func withUptaneFactory(f func(md *uptane.Metadata) (coreAgentUptaneClient, error)) Option {
```

mellon85 (approved):

pkg/config/remote/service/service.go
aya-shimizu24: good call - thank you!

aya-shimizu24:
macOS testing ✅  started up agent & fetched configs then restarted the agent, saw files get recreated and then fetched configs again. 
Screenshot 2025-10-15 at 3 55 35 PM

aya-shimizu24:
windows testing ✅ started up agent & fetched configs then restarted the agent, saw files get recreated and then fetched configs again. (when I fetched the files the second time we know they're different because of the version numbers)
Screenshot 2025-10-16 at 2 36 19 PM
Screenshot 2025-10-16 at 2 36 34 PM

PR #40067, opened 2025-08-20
quote and escape MSI property values

clarkb7:
### What does this PR do?
Fix quoting of MSI property values during remote updates on Windows

Replaces usage of `WithAdditionalArgs`, which passes arguments through as-is, with a new helper `WithProperties` that ensures the property values are properly formatted. Some properties were being escaped (`PROJECTLOCATION`) and others were not (`DDAGENTUSER_NAME`). The `WithProperties` helper will help ensure consistency.

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1742
Fix remote updates on systems where agent username contains whitespace, for example `NT AUTHORITY\SYSTEM`.

add new unit tests and e2e test

### Additional Notes

When remote updates fails it will hang, so we won't actually see an APM error span. The command line is ill formatted, so `msiexec.exe` pops a message box with help context. It does this even though we run it with the silent arg `/qn` in a non-graphical session... There does not appear to be a way to avoid this behavior with `msiexec.exe`. We could consider using the MSI API instead of running `msiexec.exe`.

### DIFF
--- a/releasenotes/notes/fix-remote-update-msi-property-quotes-089cdddf879937e3.yaml
+++ b/releasenotes/notes/fix-remote-update-msi-property-quotes-089cdddf879937e3.yaml
@@ -1,6 +1,6 @@
 ---
 fixes:
   - |
-    Remote Updates on Windows now quotes and escapes MSI property values.
-    This fixes an issue that causes remote updates to fail when the Agent username
+    Remote updates on Windows now quotes and escapes MSI property values.
+    This fixes an issue that caused remote updates to fail when the Agent username
     contains whitespace, for example ``DDAGENTUSER_NAME=NT AUTHORITY\SYSTEM``.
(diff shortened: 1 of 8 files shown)


releasenotes/notes/fix-remote-update-msi-property-quotes-089cdddf879937e3.yaml
```diff
@@ -0,0 +1,6 @@
+---
+fixes:
+  - |
+    Remote Updates on Windows now quotes and escapes MSI property values.
```
maycmlee:
```suggestion
    Remote updates on Windows now quotes and escapes MSI property values.
```

releasenotes/notes/fix-remote-update-msi-property-quotes-089cdddf879937e3.yaml
```diff
@@ -0,0 +1,6 @@
...
+    This fixes an issue that causes remote updates to fail when the Agent username
```
maycmlee:
```suggestion
    This fixes an issue that caused remote updates to fail when the Agent username
```

maycmlee (approved):
Small suggestions, but approving to not block merge.

BaptisteFoy (approved):

iglendd (approved):

PR #40074, opened 2025-08-20
fix(serverless-init): set serverless log hostname to empty string

duncanpharvey:
### What does this PR do?

Passes the `hostnameinterface` component to the Serverless Logs Agent so the hostname on logs is set to an empty string.

### Motivation

https://datadoghq.atlassian.net/browse/SVLS-7460

Currently the hostname is being set to `unknown`. In Datadog Orgs where a host with the name `unknown` already exists and has tags on it, logs with a matching hostname will inherit these tags, potentially resulting in scenarios where a log has multiple tags with the same key but different values.

Added a serverless-init log test to validate that the hostname component has an empty string set for a build of serverless-init.

Also deployed manually to Google Cloud Run and Azure Container Apps using both the sidecar and in process approach to validate the logs have hostname set as an empty string.

### Additional Notes

A regression was introduced in `v1.2.1` of the serverless-init image that caused the hostname to be set to `unknown`. It's unclear what specific change caused the regression as no Github tags were created for these older versions.

There is already a provider hostname configuration for Serverless. By passing the `hostnameinterface` component to the Serverless Logs Agent we can set the empty string hostname on the provider.
https://github.com/DataDog/datadog-agent/blob/cc85b5170958191c3385c517cc9d9763654fdb12/pkg/util/hostname/providers_serverless.go#L19

Ultimately the hostname is passed to the logs processor where it is read and inserted into the log message. This is where the hostname was set to `unknown` prior to this change.
https://github.com/DataDog/datadog-agent/blob/35ec44794e8b6f90d8fb5c5c208b086f62dadaa0/pkg/logs/processor/processor.go#L199-L219

Since the signature of `SetupLogAgent` is updated to include the hostname component `cmd/serverless` needs to be updated as well. However Lambda Functions have special handling in the logs processor which means this change will not affect the hostname on Lambda Function logs.
https://github.com/DataDog/datadog-agent/blob/42a25bbcf8055ea5c9be056f5bf539754d147df7/cmd/serverless-init/log/log.go#L59

### DIFF
--- a/.gitlab/source_test/linux.yml
+++ b/.gitlab/source_test/linux.yml
@@ -164,6 +164,17 @@ tests_serverless-init:
     - !reference [.retrieve_linux_go_deps]
     - go test -tags "serverless otlp test" ./cmd/serverless-init/.
 
+tests_serverless-init-logs:
+  extends: .linux_x64
+  stage: source_test
+  needs: ["go_deps"]
+  rules:
+    - !reference [.except_disable_unit_tests]
+    - !reference [.except_mergequeue]
+    - when: on_success
+  script:
+    - !reference [.retrieve_linux_go_deps]
+    - go test -tags "serverless otlp test" ./cmd/serverless-init/.
 
 # Check consistency of go.mod file with project imports
 go_mod_tidy_check:
--- a/cmd/serverless-init/log/log.go
+++ b/cmd/serverless-init/log/log.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/DataDog/datadog-agent/comp/core/hostname"
 	tagger "github.com/DataDog/datadog-agent/comp/core/tagger/def"
 	logsAgent "github.com/DataDog/datadog-agent/comp/logs/agent"
 	logConfig "github.com/DataDog/datadog-agent/comp/logs/agent/config"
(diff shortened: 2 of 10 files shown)


jszwedko (approved):
Approving for @DataDog/agent-log-pipelines owned files. Generally this approach makes sense to me to avoid having `host` set to `unknown`.

duncanista (approved):

jcstorms1 (approved):

ofek (approved):

ylee-figma:
Hello! I was curious when this version would be released? I hit this issue recently and I've verified using 1.1.4 does indeed not cause tag inheritance behavior as described. I'd also prefer to not use a quite outdated version of serverless-init (1.1.4 looks like it's 2 years old). Thanks!

ylee-figma:
Looks like it's released here: https://github.com/DataDog/datadog-agent/releases/tag/serverless-init-1.7.9 but the image isn't published.

duncanpharvey:
Hi @ylee-figma! serverless-init-1.7.9 was released this morning to Docker, ECR, and GAR.

SamuelMS:
@duncanpharvey Is there a way to specifically _set_ hostname, rather than it always being blank or unknown, e.g. when following [datadog's Azure Container Apps guide](https://docs.datadoghq.com/serverless/guide/aca_serverless_init/?tab=python) for serverless-init? We've got concrete hostnames in this case and would want to use the per-replica variant instead (e.g. `CONTAINER_APP_REPLICA_NAME` in the [ACA docs](https://learn.microsoft.com/en-us/azure/container-apps/environment-variables?tabs=portal#apps)).

duncanpharvey:
Hi @SamuelMS!

A Github issue was recently opened for this and we created a Feature Request on our side. However we don't currently have any estimates for when that feature request may be addressed.

https://github.com/DataDog/datadog-agent/issues/39536

SamuelMS:
> Hi @SamuelMS!
> 
> A Github issue was recently opened for this and we created a Feature Request on our side. However we don't currently have any estimates for when that feature request may be addressed.
> 
> #39536

Thanks for the reply; I'll drop a message in #39536 as well to share this same context, since this is a meaningful improvement for ACA customers.

PR #40076, opened 2025-08-20
Add more prebuilt tools from bazelbuild/build_tools and jq

aiuto:
Add the rest of the prebuilt tools from `bazelbuild/buildtools` and https://github.com/jqlang/jq.

Don't do this by hand.... there lies madness.

- Look at https://github.com/bazelbuild/buildtools/releases/tag/v8.2.1
- That is available as JSON. https://api.github.com/repos/bazelbuild/buildtools/releases
- Write a tool to read that JSON and turn it into the struct we need.  
  - the tool in experimental right now so the URL is redacted, you can figure out where it lies.
  - it needs some work

```
bazel run @multitool//tools/buildifier -- --help
bazel run @multitool//tools/jq -- --help
```

Things to look at in review:
- Keeping the pinning files different for each tool set is nice. It makes update clean.
- note that we include some extra fields, like the size and published_at. It's cheap to save and it helps with human understanding.
- the update to bazel_features was needed because adding the second lock file hit the case where that was needed. I did not investigate why.


JSGette (approved):

rdesgroppes (approved):
LG, thanks!

PR #39900, opened 2025-08-18
Update onboarding documentation

ofek:
### Motivation

We have an initiative to improve the documentation for Agent development itself. The most frequent complaint is that dev docs are scattered amongst various resources and Confluence has poor search functionality. There is appetite for a documentation "hub" of sorts and previous [efforts](https://github.com/DataDog/datadog-agent/pull/23305) were de-prioritized.

Now, we are resuming work on making the Agent's [developer docs site](https://datadoghq.dev/datadog-agent/) the central portal for Agent development. It will target both internal and external contributors.

### What does this PR do?

This focuses on improving the onboarding/set up documentation. You can see the changes by checking out this branch and running the `dda run docs serve` command.

- Split the single `/setup/` page into 3 separate pages:
    - `/setup/required/` - This streamlines onboarding by only discussing what is required to use the new developer environments. This is all most folks will need.
    - `/setup/optional/` - This discusses non-essential functionality that may be useful outside of the developer environments like tab completion and pre-commit hooks.
    - `/setup/manual/` - This discusses how to set up your local machine for building and testing the Agent. This does not support Windows (same as before) and we should discourage such setups in favor of the developer environments.
- Added a tutorial for working with developer environments
- Added a reference page that documents the developer environment images inspired by their [`README.md`](https://github.com/DataDog/datadog-agent-buildimages/blob/9e7c1dc30c27cfecdafcc109305c31686d48d7b9/dev-envs/README.md) (which will be modified to refer to this new page after merging)
- Added a reference page that documents the build images
- Added how-to pages for:
    - Building standalone agent binaries (`dda inv agent.build`)
    - Running unit tests (`dda inv test`)
    - Performing static analysis (`dda inv linter.*`)
- Migrated over most of the `README.md` file as how-to pages, allowing it to just reference the site (the only part that wasn't migrated was a short blurb about JMX which will happen in a separate PR when its dedicated page gets migrated)
- Removed the legacy `docs/dev/linters.md` page, now discussed by the static analysis how-to
- Removed the legacy `docs/dev/agent_dev_env.md` page as it was a duplicate of the manual setup instructions
- Pages now support the `<<<PYTHON_VERSION>>>` placeholder (like the existing one for Go) which uses the `.python-version` file at the root so docs never get out of sync
- Removed 2 obsolete badges from the `README.md` and site index pages

### Additional Notes

- I am reluctant to add anything else to this PR and would like to limit the review to only what has changed or what other legacy pages are out of date and can be removed.
[...]

PR #40080, opened 2025-08-21
Migrate docs for Go best practices

ofek:
### Motivation

https://github.com/DataDog/datadog-agent/pull/39900

### Additional Notes

I did some editorializing and light restructuring but otherwise didn't alter the content much. You can see the changes by checking out this branch and running the `dda run docs serve` command.

I would like to know if the part about Go import grouping static analysis being infeasible for us to check is actually true. I don't think it is, and I don't like saying that in any case.


chouetz (approved):

docs/public/guidelines/languages/go.md
```diff
@@ -0,0 +1,222 @@
...
+
+## Imports
+
+The imports defined in the `imports ( ... )` block of each Go file should be separated into the following sections, in order.
+
+1. Standard library packages (e.g. `fmt`, `net/http`)
+1. External packages (e.g. `github.com/stretchr/testify/assert`, `github.com/DataDog/datadog-agent/pkg/util/log`)
+1. [Internal](#public-apis) packages (e.g. `github.com/DataDog/datadog-agent/<parent>/internal`)
+
+This is not verified by our [static analysis](../../how-to/test/static-analysis.md) as doing so is very inefficient. Instead, configure your editor to keep imports properly sorted.
```
GustavoCaso: I wouldn't mention the part that is inefficient, as when having it set in the editor and saving the file, it does not take much time. 

What about something like: 

> This is not verified by our [static analysis](../../how-to/test/static-analysis.md) during CI. Instead, we suggest that you configure your editor to keep imports properly sorted.


Also, we should mention configuring the editor to format on save 😄

docs/public/guidelines/languages/go.md
```diff
@@ -0,0 +1,222 @@
...
+1. When refactoring a struct into its own package, try moving it into its own `internal` directory if possible (see example below).
+1. If the directory you are editing already has `internal` directories, try to move code to these `internal` directories instead of creating new ones.
+1. When creating new `internal` directories, try to use the deepest possible `internal` directory to limit the packages that can import yours. For example, if making `a/b/c/d` internal, consider moving it to `a/b/c/internal/d` instead of `a/internal/b/c/d`. With the first path, code from `a/b` won't be able to access the `d` package, while it could with the second path.
+
+/// details | Example
+Sometimes one wants to hide private fields of a struct from other code in the same package to enforce a particular code invariant. In this case, the struct should be moved to a different folder within the same package **making this folder `internal`**.
+
+Consider a module named `example.com` where you want to move `exampleStruct` from the `a` package to a subfolder to hide its private fields from `a`'s code.
```
GustavoCaso: I'm not sure if it is common to have a suffix such as `.com` in a Go package name. I would remove the suffix

docs/public/guidelines/languages/go.md
```diff
@@ -0,0 +1,222 @@
...
+
+//// tab | :octicons-file-code-16: example.com/a/code.go
+```go
+package a
+
+import (
+    "example.com/a/internal/b"
+)
+
+func doSomethingWithExampleStruct(e b.exampleStruct) {
```
GustavoCaso: To keep the example consistent, to access the struct, we should use `b.ExampleStruct`

docs/public/guidelines/languages/go.md
```diff
@@ -0,0 +1,222 @@
...
+
+Before the refactor, the code will look like this:
+
+//// tab | :octicons-file-code-16: example.com/a/code.go
+```go
+package a
+
+type exampleStruct struct {
+    // some fields
+}
```
GustavoCaso: To make the example clearer we could add some private a public fields 😄 

```go
type exampleStruct struct {
    // Public
    Foo string
    // Private
    bar string
}
```

GustavoCaso:
Great cleanup 👏 

BTW, there is a guide on how to set up VS Code for the Agent. We probably can embed some of its content here 😄 
https://datadoghq.atlassian.net/wiki/spaces/agent/pages/3164667957/VSCode+local+setup+for+datadog-agent

GustavoCaso (approved):
LGTM

PR #40083, opened 2025-08-21
fix(generate ci): Prevent failure on config parsing

chouetz:
### What does this PR do?
- Add a new condition on the parsing of the ci configuration
- Fix a warning raised by `mypy` on f-string usage

### Motivation
In #39773, there is a new part of the gitlab configuration, `.static_quality_gate_package`, which contains `variable` but is a string. This new part is causing a failure in the [parsing condition](https://github.com/DataDog/datadog-agent/blob/main/tasks/testwasher.py#L239-L240) that was used in the generation of [testwasher](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1091225533) and [fips](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1091225815) pipelines.
So I added a new condition to prevent this failure.

Executed locally the following commands:
```
dda inv -- -e gitlab.compute-gitlab-ci-config --before-file artifacts/before.gitlab-ci.yml --after-file artifacts/after.gitlab-ci.yml --diff-file artifacts/diff.gitlab-ci.yml
dda inv -- -e testwasher.generate-flake-finder-pipeline
dda inv -- -e fips.generate-fips-e2e-pipeline
```

The condition is still fragile. I suggest we don't have big scripts like in `.static_quality_gate_package` to reduce the possibility of configuration parsing errors

### DIFF
--- a/tasks/omnibus.py
+++ b/tasks/omnibus.py
@@ -36,7 +36,7 @@ def omnibus_run_task(ctx, task, target_project, base_dir, env, log_level="info",
         if host_distribution:
             overrides.append(f"--override=host_distribution:{host_distribution}")
 
-        omnibus = f"bundle exec {"omnibus.bat" if sys.platform == "win32" else "omnibus"}"
+        omnibus = f"bundle exec {'omnibus.bat' if sys.platform == 'win32' else 'omnibus'}"
         cmd = "{omnibus} {task} {project_name} --log-level={log_level} {overrides}"
         args = {
             "omnibus": omnibus,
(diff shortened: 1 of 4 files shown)


tasks/omnibus.py
```diff
@@ -36,7 +36,7 @@ def omnibus_run_task(ctx, task, target_project, base_dir, env, log_level="info",
         if host_distribution:
             overrides.append(f"--override=host_distribution:{host_distribution}")
 
-        omnibus = f"bundle exec {"omnibus.bat" if sys.platform == "win32" else "omnibus"}"
+        omnibus = f"bundle exec {'omnibus.bat' if sys.platform == 'win32' else 'omnibus'}"
```
chouetz, line 39: I modified this to have my pre-commit hook happy.

rdesgroppes: I'm the culprit

JSGette (approved):

Ishirui (approved):

PR #40088, opened 2025-08-21
Mark TestKindSuite/*/agent_pods_are_ready_and_not_restarting as flaky

alopezz:
### What does this PR do?

Marks `agent_pods_are_ready_and_not_restarting as flaky` tests under `TestKindSuite` as flaky.

### Motivation

Tests are flaking, exclusively on 7.69.x:

- https://app.datadoghq.com/s/yB5yjZ/jd2-pe6-5q2 (visualizes fail/pass ratio on 7.69.x)
- https://app.datadoghq.com/s/yB5yjZ/g98-z9u-sc3 (shows that only 7.69.x is affected)
- https://app.datadoghq.com/s/yB5yjZ/npu-et7-b3w (shows that both Test00 and TestZZ are failing, hence the marker placement).

### DIFF
--- a/test/new-e2e/tests/containers/k8s_test.go
+++ b/test/new-e2e/tests/containers/k8s_test.go
@@ -20,6 +20,7 @@ import (
 	"gopkg.in/zorkian/go-datadog-api.v2"
 
 	"github.com/DataDog/datadog-agent/pkg/util/pointer"
+	"github.com/DataDog/datadog-agent/pkg/util/testutil/flake"
 	"github.com/DataDog/datadog-agent/test/fakeintake/aggregator"
 	fakeintake "github.com/DataDog/datadog-agent/test/fakeintake/client"
 	"github.com/DataDog/datadog-agent/test/new-e2e/pkg/environments"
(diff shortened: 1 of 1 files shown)


davidor:
I have not checked the details, but maybe this is related to something I discussed with @KevinFairise2. If the one restarting is the cluster-agent and the 7.69.x branch uses a recent version of test-infra-definitions that deploys two replicas (edit: just confirmed that this is the case), it might be affected by the bug that was fixed in 7.70: https://github.com/DataDog/datadog-agent/pull/38694

alopezz:
> I have not checked the details, but maybe this is related to something I discussed with @KevinFairise2. If the one restarting is the cluster-agent and the 7.69.x branch uses a recent version of test-infra-definitions that deploys two replicas (edit: just confirmed that this is the case), it might be affected by the bug that was fixed in 7.70: #38694

Thanks for the context, that does line up. I suspect that marking this as flaky in this branch (which otherwise shouldn't see significant changes) still has a smaller blast radius than updating test-infra-definitions.

PR #40091, opened 2025-08-21
test(agent): Deduplicate files using symlinks

BaptisteFoy:
### What does this PR do?
This PR removes duplicated Agent files to save space, replacing duplicates by symlinks to the original. It does so by adding an omnibus method that's called after the Agent package has been built.

### Motivation
Reducing the Agent package size!

CI

As there are many unknowns in the duplicated files that could cause some issues (e.g. libraries accessing other files relatively to their own path), so to keep most of the wins and still be safe this was only applied to shared libraries (e.g. .so files).


coignetp (approved):

tasks/omnibus.py
```diff
@@ -621,3 +623,58 @@ def rpath_edit(ctx, install_path, target_rpath_dd_folder, platform="linux"):
...
+                hasher.update(chunk)
+        return hasher.hexdigest()
+
+    def find_duplicates(root_dir):
+        """Finds and returns duplicates as a map of hash -> list of files with that hash."""
+        hash_to_files = defaultdict(list)
+        for dirpath, _, filenames in os.walk(root_dir):
+            for name in filenames:
+                full_path = os.path.join(dirpath, name)
+                if not os.path.islink(full_path):  # Skip symlinks
```
rdesgroppes: minor (unlikely to happen): consider also skipping sockets, fifos, device files, etc.
```suggestion
                if os.path.isfile(full_path) and not os.path.islink(full_path):  # Skip symlinks
```

tasks/omnibus.py
```diff
@@ -621,3 +623,58 @@ def rpath_edit(ctx, install_path, target_rpath_dd_folder, platform="linux"):
...
+                        file_hash = hash_file(full_path)
+                        hash_to_files[file_hash].append(full_path)
+                    except Exception as e:
+                        print(f"Error hashing {full_path}: {e}")
+        return {h: paths for h, paths in hash_to_files.items() if len(paths) > 1}
+
+    def replace_with_symlinks(duplicates):
+        """Replaces all duplicates with symlinks to the first original."""
+        for files in duplicates.values():
+            files.sort()
```
rdesgroppes: minor again: you might prefer the oldest mtime or shortest path so the “canonical” file is favored, for instance:
```suggestion
            files.sort(key=lambda p: (len(p), p))
```

tasks/omnibus.py
```diff
@@ -621,3 +623,58 @@ def rpath_edit(ctx, install_path, target_rpath_dd_folder, platform="linux"):
...
+            original = files[0]
+            for dup in files[1:]:
+                try:
+                    os.remove(dup)
+                    rel_path = os.path.relpath(original, os.path.dirname(dup))
+                    if not rel_path.startswith("."):
+                        rel_path = f"./{rel_path}"
```
rdesgroppes: why the post-processing? `relpath` gives the relative path. Using `./` as a proxy for “relative” is supported by POSIX filesystems but rather [unwanted](https://www.man7.org/linux/man-pages/man1/dh_link.1.html#:~:text=relative%0A%20%20%20%20%20%20%20links%20with%20as%20short%20a%20path%20as%20possible) (and unsupported under Windows, but I understand the feature isn't expected to be used from there, right?)
```suggestion
```

BaptisteFoy: Oops, I added that when trying to fix an E2E test, but the problem was actually in the test itself. Fixed in [49b1903](https://github.com/DataDog/datadog-agent/pull/40091/commits/49b1903947b9b365c8db325ca33e3494d19ec2a3)

tasks/omnibus.py
BaptisteFoy: Good catch yes! Thanks! Fixed in [49b1903](https://github.com/DataDog/datadog-agent/pull/40091/commits/49b1903947b9b365c8db325ca33e3494d19ec2a3)

tasks/omnibus.py
BaptisteFoy: Fixed in [49b1903](https://github.com/DataDog/datadog-agent/pull/40091/commits/49b1903947b9b365c8db325ca33e3494d19ec2a3)

tasks/omnibus.py
```diff
@@ -621,3 +623,58 @@ def rpath_edit(ctx, install_path, target_rpath_dd_folder, platform="linux"):
...
+        return
+
+    print(f"Scanning for duplicates in: {root}")
+    duplicates = find_duplicates(root)
+
+    if not duplicates:
+        return
+
+    print(f"Found {sum(len(v) - 1 for v in duplicates.values())} duplicate files.")
+    replace_with_symlinks(duplicates)
```
rdesgroppes, line 687: Something else just came to my mind: what about scripts (and maybe also binaries) that are resolving paths relatively to their _real_ location? (i.e. after `readlink`/`realpath`)
For instance: (currently shipped under `/opt/datadog-agent/embedded/...`)
- https://github.com/DataDog/dd-trace-py/blob/77dc3aea4c3fd3871942b54c18e2d995b505decd/ddtrace/internal/datadog/profiling/cmake/tools/fetch_infer.sh#L19-L20
- https://github.com/openstack/openstacksdk/blob/f1e2760a6f040e710d77a7ed30250c3de2e28dde/openstack/config/defaults.py#L21
- https://github.com/mongodb/mongo-python-driver/blob/e08284bdca4e35121e618443741aae1d25153fc8/pymongo/daemon.py#L31
- ...

tasks/omnibus.py
rdesgroppes: Cool! - resolving the conversation.

tasks/omnibus.py
rdesgroppes: Thanks!

tasks/omnibus.py
BaptisteFoy: This is a good point, and while we could exclude files that contain `realpath` or `readlink` there's no real fix for binary files that I can think of. Though my gut feeling would be that such files are very unlikely to be duplicates of other.

The only idea I have here would be to play whack-a-mole by maintaining a list of files that can't be removed... what do you think?

rdesgroppes: This is a bit of a dilemma for me because, on the one hand, I would like that your initiative bears fruits, which will probably require us to take extra care with resolutions relative to real files, and on the other hand, I think we definitely need to review our construction process that has been leading to duplication in what we deliver...

BaptisteFoy: That makes sense. I would like to merge PR in the short term to offset the size increase of https://github.com/DataDog/datadog-agent/pull/39721; but I think it's a sensible long-term goal to review the process and reduce the number of duplicates

BaptisteFoy:

tasks/omnibus.py
```diff
@@ -635,11 +635,17 @@ def rpath_edit(ctx, install_path, target_rpath_dd_folder, platform="linux"):
...
     def hash_file(filepath, chunk_size=65536):
         """Returns the SHA-256 hash of the file's contents."""
         hasher = hashlib.sha256()
         with open(filepath, 'rb') as f:
-            while chunk := f.read(chunk_size):
+            while True:
+                chunk = f.read(chunk_size)
+                if not chunk:
+                    break
                 hasher.update(chunk)
```
rdesgroppes: I guess it's possible to leverage `file_digest` instead of having to write your own loop. Something like:
```py
            return hashlib.file_digest(f, "sha256").hexdigest()
```

BaptisteFoy:

tasks/omnibus.py
```diff
@@ -624,3 +626,62 @@ def rpath_edit(ctx, install_path, target_rpath_dd_folder, platform="linux"):
         if install_path in binary_rpath:
             new_rpath = os.path.relpath(target_rpath_dd_folder, os.path.dirname(file))
             _patch_binary_rpath(ctx, new_rpath, install_path, binary_rpath, platform, file)
+
+
+@task
+def deduplicate_files(ctx, directory):
+    # Match: .so, .so.0, .so.0.1, .so.0.1.2, ...
+    LIB_PATTERN = re.compile(r"\.so(?:\.\d+)*$")
```
rdesgroppes: If we'd like to extend it to:
- macOS and Windows shared libs
- Python shared libs
... we may consider extend it to sth like:
```suggestion
    LIB_PATTERN = re.compile(r"\.so(?:\.\d+)*$")
    LIB_PATTERN = re.compile(r"\.(bundle|dll|dylib|pyd|so(?:\.\d+)*)$"
```
Otherwise looks good to me, thanks!

rdesgroppes (approved):

tasks/omnibus.py
rdesgroppes: `LIB_PATTERN` addresses the concern - resolving discussion.

PR #40094, opened 2025-08-21
Support UST tags for Process entities

vitkyrka:
### What does this PR do?

Support the Unified Service Tagging service, env and version tags for
Process enties in the tagger based on the standard UST environment
variables (DD_SERVICE, DD_ENV, DD_VERSION) collected by the service
discovery code.

Note that the DDService field currently collected by the SD code is not
the same as the DD_SERVICE environment variable since it can come from
other sources such as command line parsing, and it also goes through
normalization.  For now, leave that field alone, and add a separate
field with the raw environment variable value for the UST service
environment variable, similar to what we're adding for env and version.

### Motivation

https://datadoghq.atlassian.net/browse/DSCVR-182

Tests added.


comp/core/workloadmeta/def/types.go
```diff
@@ -1459,9 +1459,14 @@ type Service struct {
 	// TracerMetadata contains APM tracer metadata
 	TracerMetadata []tracermetadata.TracerMetadata
 
-	// DDService is the value from DD_SERVICE environment variable
+	// DDService is a service name based on either the DD_SERVICE environment
+	// variable or other framework-specific tracer configuration of service
+	// names
```
stanistan: If you're distinguishing between `DDService` and `UST.Service` I think this isn't enough documentation to make that clear to other folks reading this source. 

Nit: Also would UST above `DDService` so that's next to `DDServiceInjected`

comp/core/workloadmeta/def/types.go
```diff
@@ -1548,6 +1560,9 @@ func (p Process) String(verbose bool) string {
 			_, _ = fmt.Fprintln(&sb, "Service Tracer Metadata:", p.Service.TracerMetadata)
 			_, _ = fmt.Fprintln(&sb, "Service DD Service:", p.Service.DDService)
 			_, _ = fmt.Fprintln(&sb, "Service DD Service Injected:", p.Service.DDServiceInjected)
+			_, _ = fmt.Fprintln(&sb, "Service UST Service:", p.Service.UST.Service)
```
stanistan: Consider putting this formatting logic in a `UST` print block you can call, UST can/should probably be reused by other workloadmeta types in the future :)

comp/core/workloadmeta/def/types.go
vitkyrka: I plan to drop `DDService` and just use `UST.Service` in a follow-up PR due to https://datadoghq.atlassian.net/browse/DSCVR-183. Also, I have a PR for dropping `DDServiceInjected`: https://github.com/DataDog/datadog-agent/pull/40157.

comp/core/tagger/collectors/workloadmeta_test.go
```diff
@@ -2783,3 +2783,135 @@ func assertTagInfoListEqual(t *testing.T, expectedUpdates []*types.TagInfo, upda
...
+		versionFromDD     = "1.2.3"
+	)
+
+	collector := &WorkloadMetaCollector{}
+
+	tests := []struct {
+		name            string
+		process         *workloadmeta.Process
+		expectedTagInfo *types.TagInfo
+		expectNil       bool
```
Yumasi: Nit: why a separate `expectNil` and not just put nil in `expectedTagInfo`?

Yumasi (approved):

daniel-taf (approved):

comp/core/tagger/collectors/workloadmeta_test.go
vitkyrka: Fixed

comp/core/workloadmeta/def/types.go
vitkyrka: I've expanded the comment and moved the fields

L3n41c (approved):

comp/core/tagger/collectors/workloadmeta_extract.go
```diff
@@ -257,6 +257,45 @@ func (c *WorkloadMetaCollector) handleContainer(ev workloadmeta.Event) []*types.
...
+	if process.Service.UST.Version != "" {
+		tagList.AddStandard(tags.Version, process.Service.UST.Version)
+	}
+
+	low, orch, high, standard := tagList.Compute()
+	if len(low)+len(orch)+len(high)+len(standard) == 0 {
+		return nil
+	}
+
+	processSource := string(workloadmeta.SourceServiceDiscovery)
```
davidor: I think all the other sources are defined as consts in `comp/core/tagger/collectors/workloadmeta_main.go`

comp/core/workloadmeta/def/types.go
```diff
@@ -1553,9 +1553,18 @@ type Service struct {
...
+	// environment variable, the DD_TAGS environment variable, or other
+	// parsing of framework-specific tracer configuration of service names, such
+	// as Java tracer properties on the command line.
+	//
+	// The UST.Service field on the other hand contains the raw value of the
+	// DD_SERVICE environment variable.
 	DDService string
 
+	// UST contains Unified Service Tagging environment variables
+	UST UST
```
davidor, line 1566: Should this be added to the protobufs?
The process entity was added recently in https://github.com/DataDog/datadog-agent/pull/40522

comp/core/tagger/collectors/workloadmeta_extract.go
vitkyrka: I changed it to use `processSource` from there. (It's not the same value, but I don't think it should matter.)

comp/core/workloadmeta/def/types.go
vitkyrka: Fixed.

davidor (approved):

PR #40099, opened 2025-08-21
[workloadmeta/kubelet] Stop relying on the pod watcher

davidor:
### What does this PR do?

This PR changes the kubelet workloadmeta collector to query the Kubelet directly instead of using the PodWatcher.

The first commit moves the tolerate-unready annotation to the AD code. Similar to what was done in this PR that needed to be reverted: https://github.com/DataDog/datadog-agent/pull/36253

  The PodWatcher has caused several issues in the past. For example, subtle bugs like missing tags under specific conditions. Some of the reasons are:
  - Custom update logic: It uses its own logic to detect when containers or pods have changed, which differs from workloadmeta's approach (`reflect.DeepEqual`).
  - Custom "ready" logic: It overrides the pod's ready field with its own logic.
  - Automatic deletion of containers: It deletes containers of pods that went unready (by its own definition) after 25 seconds, even when the Kubelet still returns them.

With the changes in this PR, instead of relying on the PodWatcher, the kubelet workloadmeta collector now:
- Queries the Kubelet directly.
- Uses workloadmeta's standard logic to detect updates.
- Reports exactly what the Kubelet returns without custom modifications.

Any custom logic that was in the PodWatcher should be moved to the specific components that need it, not applied globally in workloadmeta where it affects all subscribers. As an example, this PR moves the "tolerate-unready" annotation logic from the PodWatcher to the autodiscovery component where it belongs.

To reduce risk, this PR adds a new configuration option to fallback to the old behavior. Set `kubelet_use_pod_watcher: true` to use the pod watcher.

Workloadmeta was the only user of the PodWatcher. The other one was the KSM check, but it was already migrated in a previous PR: https://github.com/DataDog/datadog-agent/pull/39657

Unit tests for the new code plus existing e2e tests.
Also, manually verified that some of the components that could be impacted still work as expected: autodiscovery, logs, etc.

### DIFF
--- a/comp/core/workloadmeta/collectors/internal/kubelet/kubelet_test.go
+++ b/comp/core/workloadmeta/collectors/internal/kubelet/kubelet_test.go
@@ -3,23 +3,26 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2016-present Datadog, Inc.
 
-//go:build kubelet && test
+//go:build kubelet
 
 package kubelet
 
 import (
+	"maps"
+	"slices"
+	"sort"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 
+	workloadmeta "github.com/DataDog/datadog-agent/comp/core/workloadmeta/def"
 	"github.com/DataDog/datadog-agent/pkg/util/kubernetes"
 	"github.com/DataDog/datadog-agent/pkg/util/kubernetes/kubelet"
 	"github.com/DataDog/datadog-agent/pkg/util/pointer"
-
-	workloadmeta "github.com/DataDog/datadog-agent/comp/core/workloadmeta/def"
 )
 
 func TestPodParser(t *testing.T) {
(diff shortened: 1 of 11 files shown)


hush-hush (approved):

wdhif (approved):

pkg/util/kubernetes/kubelet/podwatcher_test.go
```diff
@@ -113,7 +113,7 @@ func (suite *PodwatcherTestSuite) TestPodWatcherComputeChangesInConditions() {
 	require.Nil(suite.T(), err)
 	require.Len(suite.T(), changes, 2)
 	assert.Equal(suite.T(), "nginx", changes[0].Spec.Containers[0].Name)
-	require.True(suite.T(), IsPodReady(changes[0]))
+	require.False(suite.T(), IsPodReady(changes[0]))
```
L3n41c, line 116: Is this change expected?

comp/core/workloadmeta/collectors/internal/kubelet/kubelet_test.go
```diff
@@ -385,3 +388,130 @@ func TestPodParser(t *testing.T) {
...
+		assert.Equal(t, workloadmeta.EventTypeUnset, event.Type)
+		assert.Equal(t, workloadmeta.SourceNodeOrchestrator, event.Source)
+		assert.Equal(t, expectedIDs[i], event.Entity.GetID().ID)
+
+		// Pods contain a FinishedAt field that's not set for containers
+		if pod, ok := event.Entity.(*workloadmeta.KubernetesPod); ok {
+			assert.NotZero(t, pod.FinishedAt)
+		}
+	}
+}
```
L3n41c, line 517: I’m wondering if we could also use `assert.ElementsMatch` for validating `podEvents` and `containerEvents` instead of having to define this helper function.

The main benefit would be the clarity of the error messages in case of test failure.
With this helper function, we might get an error message providing no more clue than:
```
len(events) != len(expectedIDs)
```

It would look something like:
```go
for _, event := range append(podEvents, containerEvents...) {
	assert.Equal(t, workloadmeta.EventTypeUnset, event.Type)
	assert.Equal(t, workloadmeta.SourceNodeOrchestrator, event.Source)
}

getID := func(event workloadmeta.CollectorEvent, _ int) string {
	return pod.Entity.GetID().ID
}
assert.ElementsMatch(t, test.expectedExpiredPodUIDs, lo.Map(podEvents, getID))
assert.ElementsMatch(t, test.expectedExpiredContainerIDs, lo.Map(containerEvents, getID))
```

L3n41c (approved):

pkg/util/kubernetes/kubelet/podwatcher_test.go
davidor: Yes. This was true because pods with the tolerate-unready annotation were considered ready by this part of the code. You can see that this was also changed on this PR that needed to be reverted https://github.com/DataDog/datadog-agent/pull/36253 and is more or less the same that I added in the first commit of this one.

PR #40101, opened 2025-08-21
[CWS-5005] Detect when exec is accessing a file through a symlink

theop-dd:
### What does this PR do?
This PR allows us to detect when an exec is accessing a file through a symlink.
We added a hook that provides a flag. When the flag is set to 1, the event is manually edited to add common symlink paths in order to match the rule.
### Motivation
The following rule was not working on Ubuntu due to alternatives symlinks : 

`exec.file.name == "nc"
`
In fact 

```
type nc
nc is hashed (/usr/bin/nc)
```
```
ls -al /usr/bin/nc
lrwxrwxrwx 1 root root 20 janv. 10  2024 /usr/bin
```

We added a test that reproduce the example presented above.

### DIFF
--- a/pkg/security/tests/process_test.go
+++ b/pkg/security/tests/process_test.go
@@ -2453,3 +2453,43 @@ func TestProcessFilelessExecution(t *testing.T) {
 		})
 	}
 }
+
+func TestSymLinkResolution(t *testing.T) {
+	SkipIfNotAvailable(t)
+
+	if testEnvironment == DockerEnvironment {
+		t.Skip("skipping in docker, not sharing the same pid ns and doesn't have a container ID")
+	}
+
+	ncPath, err := exec.Command("which", "nc").Output()
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Logf("nc path resolved by shell: %s", ncPath)
+
+	ruleDefs := []*rules.RuleDefinition{
+		{
+			ID:         "symlink_nc_exec",
+			Expression: `exec.file.name == "nc"`,
+		},
+	}
+
+	test, err := newTestModule(t, nil, ruleDefs)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer test.Close()
+
+	t.Run("exec nc via symlink", func(t *testing.T) {
+		test.WaitSignal(t, func() error {
+			cmd := exec.Command("nc")
+			cmd.Stdout = io.Discard
+			cmd.Stderr = io.Discard
+			_ = cmd.Run()
+			return nil
+		}, func(ev *model.Event, rule *rules.Rule) {
+			assertTriggeredRule(t, rule, "symlink_nc_exec")
+		})
+		assert.NoError(t, err)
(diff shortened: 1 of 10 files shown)


paulcacheux (approved):

pkg/security/tests/process_test.go
```diff
@@ -2453,3 +2453,43 @@ func TestProcessFilelessExecution(t *testing.T) {
 		})
 	}
 }
+
+func TestSymLinkResolution(t *testing.T) {
+	SkipIfNotAvailable(t)
+
+	if testEnvironment == DockerEnvironment {
```
safchain: this should be ok on docker env no ?

theop-dd: oh yes, you are right, let's try

pkg/security/tests/process_test.go
```diff
@@ -2453,3 +2453,43 @@ func TestProcessFilelessExecution(t *testing.T) {
...
+		t.Skip("skipping in docker, not sharing the same pid ns and doesn't have a container ID")
+	}
+
+	ncPath, err := exec.Command("which", "nc").Output()
```
safchain: I would suggest to create symlink and to use it, nc may not be a symlink in all the distro

pkg/security/tests/process_test.go
```diff
@@ -2453,3 +2454,43 @@ func TestProcessFilelessExecution(t *testing.T) {
...
+		require.NoError(t, err)
+
+		test.WaitSignal(t, func() error {
+			cmd := exec.Command(tmpLink)
+			cmd.Stdout = io.Discard
+			cmd.Stderr = io.Discard
+			_ = cmd.Run()
+			return nil
+		}, func(_ *model.Event, rule *rules.Rule) {
+			assertTriggeredRule(t, rule, "symlink_true_exec")
```
safchain, line 2486: I think you should check the state of the IsThroughSymLink field otherwise it seems that this test is not testing the fix.

pkg/security/tests/process_test.go
theop-dd: You are completely right, it was failing on some distros

pkg/security/tests/process_test.go
theop-dd: If we assume that the rule is not triggered without the fix, the test is actually testing. But I agree that checking the field would be more accurate

safchain: The rule triggers with or without the fix. The fix is for when the symlink is specified in the expression. Here the targeted file is used in the expression and that is this file which is actually accessed so the rule will match anyway ?

safchain (approved):

pkg/security/tests/process_test.go
theop-dd: Without the fix the rule was not triggered (because the rule is using the symlink as name). But now, it will be triggered. Anyway, I added the assert to check he flag, it will be better

PR #40102, opened 2025-08-21
fix(fleet): Dynamically search the agent install path, fallback if not found

BaptisteFoy:
### What does this PR do?
This is a follow-up for https://github.com/DataDog/datadog-agent/pull/39595 that does the following:
1. Instead of hardcoding the relative path; we go up directory by directory from the binary in search of a .install_root file that indicates where to stop
2. If we can't find that .install_root file, we fallback to the hardcoded path

### Motivation
Fix dev setups & cluster agent

Manual QA, unit tests

### DIFF
--- a/pkg/config/setup/config_nix.go
+++ b/pkg/config/setup/config_nix.go
@@ -60,18 +60,11 @@ const (
 // in time
 func osinit() {
 	// Agent binary
-	if installPathOverride, ok := os.LookupEnv("DD_TEST_INSTALL_PATH_OVERRIDE"); ok {
-		// Some tests use the InstallPath variable to fetch dependencies. These tests' executable.Folder()
-		// are relative to the test files, but dependencies are always put in the same place during the tests'
-		// executions. This environment variable allows us to override the install path during the tests.
-		InstallPath = installPathOverride
-	} else {
-		_here, err := executable.Folder() // {InstallPath}/bin/agent OR {InstallPath}/embedded/bin
-		if err != nil {
-			panic(fmt.Sprintf("Failed to get executable path: %v", err))
-		}
-		InstallPath = filepath.Join(_here, "..", "..")
+	_here, err := executable.Folder()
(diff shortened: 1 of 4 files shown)


dustmop (approved):

pkg/config/setup/config_nix.go
```diff
@@ -60,18 +60,11 @@ const (
...
-		_here, err := executable.Folder() // {InstallPath}/bin/agent OR {InstallPath}/embedded/bin
-		if err != nil {
-			panic(fmt.Sprintf("Failed to get executable path: %v", err))
-		}
-		InstallPath = filepath.Join(_here, "..", "..")
+	_here, err := executable.Folder()
+	if err != nil {
+		panic(fmt.Sprintf("Failed to get executable path: %v", err))
 	}
+	InstallPath = getInstallPathFromExecutable(_here)
```
rdesgroppes, line 67: out of curiosity (still have a lot to learn...): did we try leveraging `runtime.Caller(0)`? (https://pkg.go.dev/runtime#Caller)

BaptisteFoy: We use the similar `os.Executable` (`executable.Folder` is a wrapper around it) to get the binary path, we just need to iterate the path to find the Agent install path. 

For instance, the Agent binary path is usually `/opt/datadog-agent/bin/agent/agent`. `executable.Folder` returns `/opt/datadog-agent/bin/agent`. The expected install path is `/opt/datadog-agent` so we have go back up until we find `/opt/datadog-agent/.install_root`. Unfortunately (and as tested before) we can't use a fixed relative path like `../..`

rdesgroppes: Merci for enlightening me, @BaptisteFoy!

Stanislas167 (approved):

chouetz (approved):

gjulianm (approved):

PR #40108, opened 2025-08-21
fakeintake expose option to command line

lavigne958:
### What does this PR do?

The fakeintake server can use some options to be configured, expose these options on the command line to run the fakeintake server.

Add option to select the data retention period.

Add option to choose between "memory" and "sql" store.

In case of "sql" store, add option to choose the path for the sqlite database.
This file path can only be set using env variable. Set the value using the env variable already provided. In order to preserve current usage, if the env variable is set too, the env variable will be used.

### Motivation

these changes will allow us to use the fake intake with some local disk storage while running end2end tests. 

these changes will allow us to increase/decrease the retention period of data to match the dd-agent period (like when sending host tags to the fake intake, this is done every 30 minutes)

- start the fake intake using the new options
  - change the store to `sql`
  - change the retention period (you'll see a log when data are cleared from store)
  - change the path for the sql driver

- none, it's very useful 🙃

### DIFF
--- a/test/fakeintake/server/server.go
+++ b/test/fakeintake/server/server.go
@@ -80,6 +80,8 @@ type Server struct {
 	storeDriver string
 	store       serverstore.Store
 
+	sqliteDbPath string
+
 	responseOverridesMutex    sync.RWMutex
 	responseOverridesByMethod map[string]map[string]httpResponse
 }
@@ -108,7 +110,7 @@ func NewServer(options ...Option) *Server {
 		opt(fi)
 	}
 
-	fi.store = serverstore.NewStore(fi.storeDriver)
+	fi.store = serverstore.NewStore(fi.storeDriver, fi.sqliteDbPath)
 	registry := prometheus.NewRegistry()
 
 	storeMetrics := fi.store.GetInternalMetrics()
@@ -229,6 +231,13 @@ func WithDDDevForward() Option {
 			fi.apiKey = apiKey
 		}
 		fi.dddevForward = true
+	}
+}
+
+// WithSqlitePath sets the sqlite file path to store the received data.
+func WithSqlitePath(path string) Option {
+	return func(fi *Server) {
+		fi.sqliteDbPath = path
 	}
 }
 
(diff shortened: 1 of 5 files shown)


test/fakeintake/server/server.go
```diff
@@ -232,6 +232,20 @@ func WithDDDevForward() Option {
...
+// withSqlitePath sets the sqlite file path to store the received data.
+func WithSqlitePath(path string) Option {
+	return func(fi *Server) {
+		// sql store uses env variable to chose stlite filepath
+		_, exits := os.LookupEnv(serverstore.SqliteDbPathEnv)
+
+		// env variable takes over the command line argument
+		// only set the command line value if env is not set
+		if !exits {
+			os.Setenv(serverstore.SqliteDbPathEnv, path)
```
KevinFairise2: I am not a huge fan of using environment variable to pass parameters to the store, that looks quite hacky. Can we find a way to have it passed to the newSQLStore() directly?

lavigne958: hello sure ! I thought I stick to what is in-place.
I'll update it tomorrow with a new option in the server, that then will be passed to the SQLstore.

lavigne958: I updated the logic, let me know if this is better.

test/fakeintake/server/serverstore/db.go
```diff
@@ -46,16 +48,21 @@ type sqlMetrics struct {
...
-		}
-		p = f.Name()
+func newSQLStore(sqliteDbPath string) *sqlStore {
+	envOverrideDbPath := os.Getenv(SqliteDbPathEnv)
+
+	if envOverrideDbPath != "" {
+		sqliteDbPath = envOverrideDbPath
+	} else if sqliteDbPath == "" {
+		sqliteDbPath = defaultDBPath
+	}
```
lavigne958: This could be futher improved with something like:

```golang
var path

if env != "" {
 path = env
} else if userPath != "" {
  path = userPath
} else {
  path = default
}
```

to make it clearer in the logical order of override.

KevinFairise2 (approved):

PR #40110, opened 2025-08-21
[ACIX-973] Create `dda info owners jobs` local command

Ishirui:
### What does this PR do?

Create a new local `dda` command, called via `dda info owners jobs`, replicating the functionality of the `owners.find-jobowners` invoke task.

### Motivation

Improving devx + proof of concept for migration of invoke tasks to `dda` using local commands, and especially how they integrate within an already existing command group
See: DataDog/datadog-agent-dev#167

Running the commands locally with a few examples.
Automated tests for `dda` local commands are harder to implement for the moment.

### DIFF
--- a/.dda/extend/commands/info/owners/jobs/__init__.py
+++ b/.dda/extend/commands/info/owners/jobs/__init__.py
@@ -1,0 +1,54 @@
+# SPDX-FileCopyrightText: 2025-present Datadog, Inc. <dev@datadoghq.com>
+#
+# SPDX-License-Identifier: MIT
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+import click
+from dda.cli.base import dynamic_command, pass_app
+from dda.utils.fs import Path
+
+if TYPE_CHECKING:
+    from dda.cli.application import Application
+
+
+@dynamic_command(short_help="Find owners of GitLab CI jobs", features=["codeowners"])
+@click.argument(
+    "jobs",
+    required=True,
+    nargs=-1,
+)
+@click.option(
+    "--config",
+    "-c",
+    "config_filepath",
+    type=click.Path(exists=True, dir_okay=False, path_type=Path),
+    help="Path to JOBOWNERS file",
+    default=".gitlab/JOBOWNERS",
+)
+# TODO(@agent-devx): Make this respect any --non-interactive flag or other way to detect CI environment
+@click.option(
+    "--json",
+    is_flag=True,
+    help="Format the output as JSON",
+)
+@pass_app
+def cmd(app: Application, jobs: tuple[str, ...], *, config_filepath: Path, json: bool) -> None:
+    """
+    Gets the owners for the specified GitLab CI jobs.
+    """
+    import codeowners
+
+    with config_filepath.open(encoding="utf-8") as f:
+        owners = codeowners.CodeOwners(f.read())
+
+    res = {job: [owner[1] for owner in owners.of(job)] for job in jobs}
+
+    if json:
+        from json import dumps
+
+        app.output(dumps(res))
+    else:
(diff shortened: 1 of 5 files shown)


Ishirui:
@ofek do we have, as of today, a good way to implement unit tests for `dda` local commands, or is that not available yet ?

.dda/extend/commands/info/owners/jobs/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+)
+@click.option(
+    "--config",
+    "-c",
+    "config_filepath",
+    type=click.Path(exists=True, dir_okay=False, path_type=Path),
+    help="Path to JOBOWNERS file",
+    default=".gitlab/JOBOWNERS",
+)
+# TODO(@agent-devx): Make this respect any --non-interactive flag or other way to detect CI environment
```
CelianR, line 30: I think that `app` could be used for that right? IIRC the interactive flag is handled there

CelianR (approved):

.dda/extend/commands/info/owners/jobs/__init__.py
Ishirui: I don't think so - IIRC the only thing the `--non-interactive` flag does as of today is tweak how `click` outputs stuff to the terminal. We don't have any custom logic in `dda` referring to this flag... but we definitely should, especially as we start to use `dda`-native things more and more in CI (I'd really like to avoid the same kind of problems we had with the changelog incidents a few weeks ago 😢)

ofek:
We don't yet have a way to test local commands but I'll try to think of a nice UX soon.

.dda/extend/commands/info/owners/jobs/__init__.py
```diff
@@ -0,0 +1,53 @@
+# SPDX-FileCopyrightText: 2025-present Datadog, Inc. <dev@datadoghq.com>
+#
+# SPDX-License-Identifier: MIT
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+import click
+from dda.cli.base import dynamic_command, pass_app
```
ofek, line 9:
```suggestion
import click

from dda.cli.base import dynamic_command, pass_app
```

.dda/extend/commands/info/owners/jobs/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+from dda.utils.fs import Path
+
+if TYPE_CHECKING:
+    from dda.cli.application import Application
+
+
+@dynamic_command(short_help="Find owners of GitLab CI jobs", features=["codeowners"])
```
ofek, line 16: Nit: what are your thoughts about the future possibility of making this generic over CI e.g. this would show workflows for GitHub Actions.

.dda/extend/commands/info/owners/jobs/__init__.py
Ishirui: That gets reverted by the pre-commit hooks :sweat_smile:

.dda/extend/commands/info/owners/jobs/__init__.py
Ishirui: Sure, no reason why we couldn't - the JOBOWNERS file just makes a mapping between string patterns and owners, it does not actually matter where the strings come from (gitlabci files, gha files etc.)

.dda/extend/commands/info/owners/jobs/__init__.py
ofek: 🙈

ofek: Yeah, not important but we should eventually configure that as a first party library.

PR #40119, opened 2025-08-21
add Windows DDOT OCI packaging

guohdd:
### What does this PR do?
This PR adds OCI packaging for the Datadog Opentelemetry (DDOT) on Windows platform

### Motivation
Datadog OpenTelemetry solution (DDOT) needs to be packaged and delivered on the Windows platform.

The solution has been validated by building the DDOT OCI package using the PowerShell script Generate-OCIPackage.ps1, and running installation on Windows VM.

The solution was further verified with full CI pipelines and by inspecting artifact ddot_win_oci.zip generated by the CI pipelines.

This PR is part of the overall Datadog DDOT effort. The scope of this PR is OCI packaging for Windows platform with basic installation validation.


.gitlab/package_build/windows.yml
```diff
@@ -145,3 +145,45 @@ windows_zip_ddot_x64:
...
+      -e API_KEY_ORG2=${API_KEY_ORG2}
+      -e CI_JOB_TOKEN=${CI_JOB_TOKEN}
+      -e PIP_INDEX_URL=${PIP_INDEX_URL}
+      ${WINBUILDIMAGE}
+      powershell -NoProfile -C 'Set-Location C:\mnt; . .\tasks\winbuildscripts\common.ps1; Enable-DevEnv; Expand-ModCache -modcache modcache; Install-Deps; dda self telemetry disable; $env:GOFLAGS=''-x -v''; dda inv otel-agent.build; if (-not (Test-Path C:\mnt\bin\otel-agent\otel-agent.exe)) { Write-Error "otel-agent.exe not built"; exit 1 }; $ver=(dda inv -- agent.version --url-safe)+"-1"; powershell -File tasks\winbuildscripts\Generate-OCIPackage.ps1 -package datadog-agent-ddot -version $ver -omnibusOutput C:\mnt\omnibus\pkg\pipeline-$env:CI_PIPELINE_ID -CleanupStaging'
+    - Get-ChildItem omnibus\pkg\pipeline-$CI_PIPELINE_ID
+  artifacts:
+    expire_in: 2 weeks
+    paths:
+      - omnibus/pkg/pipeline-$CI_PIPELINE_ID
```
chouetz: 🥜 
```suggestion
      - omnibus/pkg/pipeline-$CI_PIPELINE_ID

```

.gitlab/package_build/windows.yml
```diff
@@ -145,3 +145,45 @@ windows_zip_ddot_x64:
...
+      -e WINDOWS_BUILDER=true
+      -e GOMODCACHE="c:\modcache"
+      -e AWS_NETWORKING=true
+      -e SIGN_WINDOWS_DD_WCS=true
+      -e S3_OMNIBUS_CACHE_BUCKET="$S3_OMNIBUS_CACHE_BUCKET"
+      -e API_KEY_ORG2=${API_KEY_ORG2}
+      -e CI_JOB_TOKEN=${CI_JOB_TOKEN}
+      -e PIP_INDEX_URL=${PIP_INDEX_URL}
+      ${WINBUILDIMAGE}
+      powershell -NoProfile -C 'Set-Location C:\mnt; . .\tasks\winbuildscripts\common.ps1; Enable-DevEnv; Expand-ModCache -modcache modcache; Install-Deps; dda self telemetry disable; $env:GOFLAGS=''-x -v''; dda inv otel-agent.build; if (-not (Test-Path C:\mnt\bin\otel-agent\otel-agent.exe)) { Write-Error "otel-agent.exe not built"; exit 1 }; $ver=(dda inv -- agent.version --url-safe)+"-1"; powershell -File tasks\winbuildscripts\Generate-OCIPackage.ps1 -package datadog-agent-ddot -version $ver -omnibusOutput C:\mnt\omnibus\pkg\pipeline-$env:CI_PIPELINE_ID -CleanupStaging'
```
chouetz: Nit: like the [other similar tasks](https://github.com/DataDog/datadog-agent/pull/40119/files?w=1#diff-b92bec50361406d6295308cb1dcd9f199ee2c341990ac30fe021f9a2c00d8c9dR129) could it be possible to have a single script execution here (rather than a multiline script). It would ease readability, further updates and reproducibility (if you want to execute this locally for an investigation)

releasenotes/notes/add_windows_ddot_oci_package-f79c162decc758db.yaml
```diff
@@ -0,0 +1,11 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+features:
+  - |
+    Add Windows DDOT OCI package to support Datadog OpenTelemetry
```
chouetz:
```suggestion
    Add Windows DDOT OCI package to support Datadog OpenTelemetry

```

chouetz (approved):
Minor suggestions

.gitlab/package_build/windows.yml
alopezz: Why are we rebuilding the otel-agent here? This job declares `windows_zip_ddot_x64` in its `needs`, that means it already gets the pre-built zip with the contents that should go into the same package.

However, the `windows_zip_ddot_x64` was in fact pretty much just a placeholder (I don't think we need the zip for anything). So it makes more sense to have the OCI package build replace it. Also, I think this would follow what we already do for the main Agent windows packaging jobs, which build both the MSI and the OCI package as part of the same job and out of the same base zip produced by the Omnibus build.

That should also remove the need for this complex powershell one-liner and most of the custom logic added to the Generate-OCIPackage.ps1 script. The only practical difference that I see here is that instead of going msi -> oci like the existing scripts do we now need a zip -> oci one.

alopezz (changes requested):

releasenotes/notes/add_windows_ddot_oci_package-f79c162decc758db.yaml
brett0000FF:
```suggestion
    Add Windows DDOT OCI package to support Datadog OpenTelemetry.
    
```

brett0000FF (approved):

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
```diff
@@ -0,0 +1,246 @@
...
+	// 1) Write otel-config.yaml with API key/site substitutions
+	if err = writeOTelConfigWindows(); err != nil {
+		return fmt.Errorf("could not write otel-config.yaml: %w", err)
+	}
+	// 2) Enable otelcollector in datadog.yaml
+	if err = enableOtelCollectorConfigWindows(ctx.Context); err != nil {
+		return fmt.Errorf("failed to enable otelcollector: %w", err)
+	}
+	// 3) Restart main Agent services to pick up config changes
+	if err = windowssvc.NewWinServiceManager().RestartAgentServices(ctx.Context); err != nil {
```
alexn-dd, line 61: If the Otel collector needs to restart services after install, is it possible to perform this config as a pre-install or install step (if available)?

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+// preRemoveDatadogAgentDdot performs pre-removal steps for the DDOT package on Windows
+// All the steps are allowed to fail
+func preRemoveDatadogAgentDdot(ctx HookContext) error {
+	_ = stopServiceIfExists(otelServiceName)
+	_ = deleteServiceIfExists(otelServiceName)
+	if !ctx.Upgrade {
+		// Remove config and revert datadog.yaml changes
+		_ = os.Remove(filepath.Join(paths.DatadogDataDir, "otel-config.yaml"))
```
alexn-dd: The agent or system may suffer from catastrophic failure during removal and potentially corrupt the configs.  How is such failure handled?

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
```diff
@@ -0,0 +1,246 @@
...
+		return err
+	}
+	var cfg map[string]any
+	if err := yaml.Unmarshal(data, &cfg); err != nil {
+		return err
+	}
+	apiKey, _ := cfg["api_key"].(string)
+	site, _ := cfg["site"].(string)
+
+	exampleData, err := os.ReadFile(example)
```
alexn-dd: Nit: I recommend the term "template" rather than "example."

.gitlab/package_build/windows.yml
guohdd: Great point.

guohdd: Thanks for the clarification. I wasn't very sure about the purpose of windows_zip_ddot_x64. I will have the OCI package build replace it.

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
guohdd: Great question. This is done in postinstalll due to the availability of the ddot files. If there's a concern that restarting all services is too much overhead, it may be possible to start services based on config changes. However, if the core agent service needs to pick up any change, that will trigger full restart.

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
guohdd: Excellent question. Yeah, files could get corrupted due to unforeseen conditions. We could consider a more robust config backup and atomic restore here. However it'd be better to have a general solution that all services could use. Please let me know if we already have a robust way to do it.

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
guohdd: Here we try to follow the naming of config files. But you have a great point, I will rename the variables for better clarification.

.gitlab/package_build/windows.yml
clarkb7: I'd prefer having separate build and package jobs. It's how most of the Agent pipeline is structured right now, except for the Windows job, which I'd like to split up to match the rest.

Having a zip artifact makes things like this easier: https://github.com/DataDog/datadog-agent/pull/38018

It would also be nice if we didn't need to spin up a Windows runner and a docker container just to run `datadog-package`. it's a cross-platform tool and the Windows runners are a lot slower. We should be able to create the OCI package just like https://github.com/DataDog/datadog-agent/blob/42fd6c48323bb6f8f8a462345bc6a414f174bf81/.gitlab/packaging/oci.yml#L104-L111

This would also help keep the logic for creating the OCI package the same / close across platforms.

tasks/winbuildscripts/Generate-OCIPackage.ps1
```diff
@@ -2,7 +2,12 @@ Param(
     [Parameter(Mandatory=$true)]
     [string] $package,
     [string] $version,
-    [string] $omnibusOutput = "$(Get-Location)\omnibus\pkg\"
+    [string] $omnibusOutput = "$(Get-Location)\omnibus\pkg\",
```
clarkb7: This script seems to have outlived it's usefulness, there's a lot of package specific cases in it now and it's looking hard to manage. The script is currently
- downloading the datadog-package tool
- creating the unique layout for each package from their unique artifacts
- running the datadog-package tool with unique args for each package
- sometimes copying the output to a different location

Let's try to keep this script as is for the Agent, and for DDOT re-use the job in `.gitlab/packaging/oci.yml`. Preferably that script would not be in .gitlab yaml and could be an invoke task so it is easier to run and also cross-platform, but we can handle that in a separate PR.

tasks/winbuildscripts/Build-OmnibusTarget.ps1
```diff
@@ -69,13 +69,20 @@ Invoke-BuildScript `
...
     # Show the contents of the output package directories for debugging purposes
-    Get-ChildItem -Path C:\omnibus-ruby\pkg\
-    Get-ChildItem -Path "C:\opt\datadog-agent\bin\agent\"
-    Get-ChildItem -Path ".\omnibus\pkg\"
+    if (Test-Path 'C:\omnibus-ruby\pkg\') { Get-ChildItem -Path 'C:\omnibus-ruby\pkg\' }
+    if (Test-Path '.\omnibus\pkg\') { Get-ChildItem -Path '.\omnibus\pkg\' }
 
     if ($BuildOutOfSource) {
-        # Copy the resulting package to the mnt directory
+        # Copy the resulting package to the mnt directory when relevant (non-ddot targets)
```
clarkb7: Why is DDOT a special case here? Don't we want to save the build output in the job artifacts?

test/new-e2e/tests/installer/unix/package_definitions.go
```diff
@@ -68,6 +68,8 @@ var PackagesConfig = []TestPackageConfig{
 	{Name: "datadog-installer", Version: fmt.Sprintf("pipeline-%v", os.Getenv("E2E_PIPELINE_ID")), Registry: "installtesting.datad0g.com.internal.dda-testing.com"},
 	{Name: "datadog-agent", Alias: "agent-package", Version: fmt.Sprintf("pipeline-%v", os.Getenv("E2E_PIPELINE_ID")), Registry: "installtesting.datad0g.com.internal.dda-testing.com"},
 	{Name: "datadog-ddot", Alias: "ddot-package", Version: fmt.Sprintf("pipeline-%v", os.Getenv("E2E_PIPELINE_ID")), Registry: "installtesting.datad0g.com.internal.dda-testing.com"},
+	// Windows/Linux DDOT OCI package name aligned with packaging jobs
+	{Name: "datadog-agent-ddot", Alias: "ddot-package", Version: fmt.Sprintf("pipeline-%v", os.Getenv("E2E_PIPELINE_ID")), Registry: "installtesting.datad0g.com.internal.dda-testing.com"},
```
clarkb7: Why are there two entries for ddot? Shouldn't we use the same name for all platforms?

test/new-e2e/tests/installer/windows/suites/ddot-package/install_test.go
```diff
@@ -0,0 +1,70 @@
...
+		s.Env().RemoteHost,
+		windowsAgent.WithPackage(s.CurrentAgentVersion().MSIPackage()),
+		windowsAgent.WithInstallLogFile(filepath.Join(s.SessionOutputDir(), "install-agent-msi.log")),
+		windowsAgent.WithZeroAPIKey(),
+	)
+	s.Require().NoErrorf(err, "failed to install Agent with MSI")
+}
+
+func (s *testDDOTInstallSuite) uninstallAgentIfPresent() {
+	_ = windowsAgent.UninstallAgent(s.Env().RemoteHost, filepath.Join(s.SessionOutputDir(), "uninstall-agent-msi.log"))
```
clarkb7, line 68: if you're trying to clear previous state you should use `Purge()` so it removes all packages
```
s.Installer().Purge()
```

.gitlab/package_build/windows.yml
guohdd: Sure I will look into it.

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
```diff
@@ -0,0 +1,246 @@
...
+		// Restart core agent to pick up reverted config
+		if err := windowssvc.NewWinServiceManager().RestartAgentServices(ctx.Context); err != nil {
+			log.Warnf("failed to restart agent services: %s", err)
+		}
+	}
+	return nil
+}
+
+// writeOTelConfigWindows creates otel-config.yaml by substituting API key and site values from datadog.yaml
+func writeOTelConfigWindows() error {
```
clarkb7, line 143: can we share some of this with `datadog_agent_ddot_linux` ?

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
```diff
@@ -0,0 +1,246 @@
...
+		content = strings.ReplaceAll(content, "${env:DD_API_KEY}", apiKey)
+	}
+	if site != "" {
+		content = strings.ReplaceAll(content, "${env:DD_SITE}", site)
+	}
+	return os.WriteFile(out, []byte(content), 0o600)
+}
+
+// enableOtelCollectorConfigWindows adds otelcollector.enabled and agent_ipc defaults to datadog.yaml
+func enableOtelCollectorConfigWindows(_ context.Context) error {
```
clarkb7, line 159: can we share some of this with datadog_agent_ddot_linux ?

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
```diff
@@ -0,0 +1,246 @@
...
+	existing["agent_ipc"] = map[string]any{"port": 5009, "config_refresh_interval": 60}
+	updated, err := yaml.Marshal(existing)
+	if err != nil {
+		return err
+	}
+	return os.WriteFile(ddYaml, updated, 0o600)
+}
+
+// disableOtelCollectorConfigWindows removes otelcollector and agent_ipc from datadog.yaml
+func disableOtelCollectorConfigWindows() error {
```
clarkb7, line 165: can we share some of this with datadog_agent_ddot_linux ?

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+	defer m.Disconnect()
+	s, err := m.OpenService(name)
+	if err != nil {
+		return nil
+	}
+	defer s.Close()
+	_, _ = s.Control(svc.Stop)
+	return nil
+}
```
clarkb7, line 228: this doesn't wait for the service to stop, use `winutil.StopService(serviceName)` instead

in a future PR maybe we can modify and use https://github.com/DataDog/datadog-agent/blob/70fe41543968eb9bc086590ff23706104b153548/pkg/fleet/installer/packages/service/windows/interface.go#L15-L22

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
```diff
@@ -0,0 +1,246 @@
...
+	s, err := m.OpenService(otelServiceName)
+	if err == nil {
+		defer s.Close()
+		// No-op: service exists; could update config here if needed
+		return nil
+	}
+	s, err = m.CreateService(otelServiceName, bin, mgr.Config{
+		DisplayName:      "Datadog Distribution of OpenTelemetry Collector",
+		Description:      "Datadog OpenTelemetry Collector",
+		StartType:        mgr.StartAutomatic,
```
clarkb7: I think we want this to be Manual start, I thought the core Agent is responsible for starting it if it is enabled in the configuration.

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
clarkb7: should this be done after you create the DDOT service? right now I think the Agent is going to try to start a non-existing service.

tasks/winbuildscripts/Generate-OCIPackage.ps1
julien-lebot: I don’t think this script has outlived its usefulness, having the packaging steps in `Generate-OCIPackage.ps1` avoids duplicating Windows-specific logic in GitLab YAML and makes local repro much easier.

That said, we can simplify it so it’s easier to maintain: drop the manifest-parsing and require the caller to pass `-Version`, assume `datadog-package` is already on `PATH` etc... That would keep it as a thin wrapper around `datadog-package` with minimal branching.

Once the Windows path follows the same `datadog-package` pattern as Linux, we can remove the script entirely.

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
julien-lebot: The `datadog-otel-agent` service isn’t one of the dependent services the Agent tries to start 
(https://github.com/DataDog/datadog-agent/blob/main/cmd/agent/subcommands/run/dependent_services_windows.go#L28). 
Even if it didn’t exist, it would only log a warning and not crash 
(https://github.com/DataDog/datadog-agent/blob/main/cmd/agent/subcommands/run/dependent_services.go#L43).

@guohdd could you share the reasoning for restarting the Agent services before installing the DDOT one?

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
julien-lebot: Today DDOT is created as an independent Windows service and started automatically, and the Agent does not list it among its dependent services. That’s fine for first install because the hook restarts Agent before starting DDOT, but on boot DDOT could start before the Agent.

Two options depending on the intended contract:
- If DDOT requires Agent IPC at start, we should encode that: make `datadog-otel-agent` depend on `datadogagent` (SCM `DependOnService`) or at least set `DelayedAutoStart` so it starts after the initial wave.
- If DDOT tolerates Agent absence and retries, then Automatic is OK and we can document that DDOT is resilient to starting before the Agent.

Do we want DDOT to be supervised by the Agent eventually (i.e., treated as a dependent component)? If yes, we can switch it to `Manual` and have the Agent manage its lifetime; otherwise I’d add a `DependOnService` to make boot ordering explicit.

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
```diff
@@ -0,0 +1,246 @@
...
+		ServiceStartName: "", // LocalSystem
```
julien-lebot, line 209: **✏️ suggestion:** ‏ Could you add a note on why we’re creating the service under `LocalSystem`? 
My instinct would be to pick a lower-privilege account such as `NetworkService`, so it would be helpful to know if DDOT requires `LocalSystem` privileges (e.g. for file system or IPC access). Documenting the rationale here would make the intent clearer for future maintainers.

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
guohdd: Good questions. The purpose of restarting the main agent is to pick up the config changes for DDOT so that the corresponding IPC endpoint could be brought up. DDOT service is started later depending on the IPC channel (via which the main agent sends things like config/state.

tasks/winbuildscripts/Build-OmnibusTarget.ps1
guohdd: The implementation does the DDOT packaging inside the same container and does not need to to output the job artifacts. However, since we will go with splitting the build and packaging, the DDOT will not be a special case here.

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
clarkb7: currently planning to have it run as `ddagentuser`, just using `LocalSystem` here for simplicity of initial commit, see https://datadoghq.atlassian.net/browse/WINA-1619

Can you link the card in a TODO comment?
```
// TODO(WINA-1619): Change service user to ddagentuser
```

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
clarkb7: dependent service changes in this PR: https://github.com/DataDog/datadog-agent/pull/39967

guohdd:
> I see this PR removes `Generate-OCIPackage.ps1` and its integration ([f659991](https://github.com/DataDog/datadog-agent/commit/f65999112f90f5c09c52e32abe9fc30f64c296cf)), but that breaks Windows OCI generation for the Agent packages. The E2E test failures suggest the fleet installer can't find the expected Windows OCI packages.
> 
> Are you planning to:
> 
> 1. Restore the script (per Branden's earlier guidance), or
> 2. Complete the migration to GitLab CI-based OCI generation for agent packages?
> 
> If option 2, we'd need an additional job similar to `ddot_win_oci` for the main Agent package, and update the `agent_oci` job to merge Windows OCI package alongside Linux ones.

Yes the plan is to add a job like agent_win_oci a little later.

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
julien-lebot: Thank you @clarkb7 🙇 I was missing that context

.gitlab/packaging/oci.yml
```diff
@@ -75,6 +75,32 @@
...
+    - MSI="$(ls -1 "${PIPE_DIR}"/datadog-agent-*x86_64.msi | head -n1)"
+    - '[ -n "${MSI}" ]'
+    - STAGING="$(mktemp -d)"; cp "${MSI}" "${STAGING}/"
+    - BASE="$(basename "${MSI}")"
+    - VERSION="$(echo "${BASE}" | sed -E 's/^datadog-agent-(.+)-x86_64\.msi$/\1/')"
+    - datadog-package create --version "${VERSION}" --package datadog-agent --os windows --arch amd64 --archive --archive-path "${PIPE_DIR}/datadog-agent-${VERSION}-windows-amd64.oci.tar" "${STAGING}/"
+  artifacts:
+    expire_in: 2 weeks
+    paths:
+      - omnibus/pkg/pipeline-$CI_PIPELINE_ID
```
alopezz: I'm not sure I understand why we would be making this change here (as well as dropping the Generate-OCIPackage.ps1 script and all those related changes). It doesn't seem to be necessary to the addition of the ddot OCI package.

.gitlab/packaging/oci.yml
```diff
@@ -101,11 +128,44 @@ installer_oci:
...
+ddot_win_oci:
+  extends: .package_oci
+  stage: packaging
+  needs:
+    - job: "windows_zip_ddot_x64"
+      artifacts: true
+  rules:
+    - !reference [.except_mergequeue]
+    - when: on_success
+  script:
```
alopezz: Instead of repeating all this logic, perhaps we could better either:
- Make `.package_oci`'s logic more generic such that we can reuse its `script` as is, or
- Extract common logic into a separate script and reuse that.

clarkb7: agreed. i said the same here

https://github.com/DataDog/datadog-agent/pull/40119#discussion_r2319948370

tasks/winbuildscripts/Generate-OCIPackage.ps1
guohdd: This has been re-worked so build and packaging are separate jobs. The Windows OCI package is done based on job windows_zip_ddot_x64 by Linux runner, so no need to spin a separate Windows runner for the packaging.

test/new-e2e/tests/installer/unix/package_definitions.go
guohdd: Good catch. This is now consolidated.

test/new-e2e/tests/installer/windows/suites/ddot-package/install_test.go
guohdd: Added.

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
guohdd: Removed dependency at SCM level and OTEL service is set for manual start. Dependency is set in [PR39967](https://github.com/DataDog/datadog-agent/pull/39967) so core agent could start/stop the service.

.gitlab/packaging/oci.yml
guohdd: Great point. Made consolidation.

.gitlab/packaging/oci.yml
guohdd: Possibly I misunderstood earlier suggestions. The Powershell script is removed because the build and packaging are done in separate jobs. The packaging is done in Linux runner.

.gitlab/packaging/oci.yml
```diff
@@ -75,6 +75,55 @@
     paths:
       - ${OMNIBUS_PACKAGE_DIR}
 
+.package_oci_win:
+  extends: .package_oci
+  script:
+    - set -euo pipefail
```
alopezz: All gitlab jobs run with `set -eo pipefail` by default so these are technically not necessary here.

```suggestion
    - set -u
```

.gitlab/packaging/oci.yml
```diff
@@ -75,6 +75,55 @@
...
+    - PIPE_DIR="$OMNIBUS_PACKAGE_DIR/pipeline-$CI_PIPELINE_ID"
+    - ARTIFACT="$(ls -1 "${PIPE_DIR}"/${WIN_ARTIFACT_PATTERN} | head -n1)"
+    - '[ -n "${ARTIFACT}" ]'
+    - |
+      case "${WIN_SOURCE_TYPE}" in
+        msi)
+          STAGING="$(mktemp -d)"; cp "${ARTIFACT}" "${STAGING}/"
+          BASE="$(basename "${ARTIFACT}")"
+          VERSION="$(echo "${BASE}" | sed -E "${WIN_VERSION_SED}")"
```
alopezz: Shouldn't we be using the `PACKAGE_VERSION` variable defined in the `before_script` instead?

.gitlab/packaging/oci.yml
```diff
@@ -75,6 +75,55 @@
...
+          SRC_DIR="${WIN_INPUT}"
+          EXTRA_FLAGS=""
+          if [ -d "${WIN_INPUT}/etc/datadog-agent" ]; then
+            EXTRA_FLAGS="--configs ${WIN_INPUT}/etc/datadog-agent"
+          fi
+          ;;
+        *)
+          echo "Unknown WIN_SOURCE_TYPE: ${WIN_SOURCE_TYPE}"; exit 1
+          ;;
+      esac
```
alopezz, line 110: I feel like this could be greatly simplified because there's a lot more in common than what this makes it look like. Particularly if we use `PACKAGE_VERSION` instead of parsing it from the package. Something like:

```suggestion
      SRC_DIR="$(mktemp -d)"
      EXTRA_FLAGS=""
      case "${WIN_SOURCE_TYPE}" in
        msi)
          cp "${ARTIFACT}" "${SRC_DIR}"
          ;;
        zip)
          unzip -q "${ARTIFACT}" -d "${SRC_DIR}"
          if [ -d "${SRC_DIR}/etc/datadog-agent" ]; then
            EXTRA_FLAGS="--configs ${WIN_INPUT}/etc/datadog-agent"
          fi
          ;;
        *)
          echo "Unknown WIN_SOURCE_TYPE: ${WIN_SOURCE_TYPE}"; exit 1
          ;;
      esac
```

.gitlab/packaging/oci.yml
```diff
@@ -75,6 +75,55 @@
     paths:
       - ${OMNIBUS_PACKAGE_DIR}
 
+.package_oci_win:
+  extends: .package_oci
+  script:
+    - set -euo pipefail
+    - PIPE_DIR="$OMNIBUS_PACKAGE_DIR/pipeline-$CI_PIPELINE_ID"
+    - ARTIFACT="$(ls -1 "${PIPE_DIR}"/${WIN_ARTIFACT_PATTERN} | head -n1)"
+    - '[ -n "${ARTIFACT}" ]'
```
alopezz: Suggestion: rather than allowing several files to match the pattern and arbitrarily choose one, we should probably assert that there's really only a single file and otherwise fail, something like:

```suggestion
    - ARTIFACT="$(ls ${PIPE_DIR}"/${WIN_ARTIFACT_PATTERN})" || { echo "No suitable input artifact found"; exit 1; } 
    - [ "$(echo ${ARTIFACT} | wc -w)" = 1 ] || { echo "Expected to only one candidate input artifact"; exit 1; }
```

.gitlab/packaging/oci.yml
```diff
@@ -85,6 +134,7 @@ agent_oci:
       "go_tools_deps",
       "installer-arm64-oci",
       "installer-amd64-oci",
+      "agent_win_oci"
```
alopezz, line 137: I think it should be possible to drop `"windows_msi_and_bosh_zip_x64-a7",` from this list of `needs` at this point since the oci package is provided by the new job.

.gitlab/packaging/oci.yml
guohdd: Yeah, I am not sure if Windows artifact naming always matches PACKAGE_VERSION exactly. My intention was to make the OCI package version match the MSI/ZIP that was built in a separate job. Please share more insight if possible. Meanwhile it's a good idea and I will look into it.

.gitlab/packaging/oci.yml
guohdd: The intention was to ensure job "windows_msi_and_bosh_zip_x64-a7" is done before we do the top level packaging. But you made a good point, I will look into it if it's really necessary to have "wind block.ws_msi_and_bosh_zip_x64-a7" in the "needs" block.

.gitlab/packaging/oci.yml
guohdd: Made the change per suggestion.

.gitlab/packaging/oci.yml
guohdd: Made changes per suggestion.

.gitlab/packaging/oci.yml
guohdd: made changes per suggestion.

.gitlab/packaging/oci.yml
guohdd: Made changes per suggestion

.gitlab/package_build/windows.yml
guohdd: Made changes to separate build and packaging per suggestion.

.gitlab/packaging/oci.yml
```diff
@@ -75,6 +75,50 @@
...
+          ;;
+      esac
+    - datadog-package create --version "${PACKAGE_VERSION}" --package "${OCI_PRODUCT}" --os windows --arch amd64 --archive --archive-path "${PIPE_DIR}/${OCI_PRODUCT}-${PACKAGE_VERSION}-windows-amd64.oci.tar" ${EXTRA_FLAGS} "${SRC_DIR}/"
+    - ls -l "${PIPE_DIR}"
+  artifacts:
+    expire_in: 2 weeks
+    paths:
+      - ${OMNIBUS_PACKAGE_DIR}/pipeline-$CI_PIPELINE_ID
+
+agent_win_oci:
```
clarkb7, line 118: didn't have to change how the Agent OCI was built, it's unrelated to the original intent of the PR. But since it's already done we might as well leave it. It's better to try to keep PR changes isolated. without these changes you wouldn't have hit the earlier merge conflict and test failure in `winbuild.py`

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
```diff
@@ -0,0 +1,366 @@
...
+func postInstallDatadogAgentDdot(ctx HookContext) (err error) {
+	// 1) Write otel-config.yaml with API key/site substitutions
+	if err = writeOTelConfigWindows(); err != nil {
+		return fmt.Errorf("could not write otel-config.yaml: %w", err)
+	}
+	// 2) Enable otelcollector in datadog.yaml
+	if err = enableOtelCollectorConfigWindows(ctx.Context); err != nil {
+		return fmt.Errorf("failed to enable otelcollector: %w", err)
+	}
+	// 3) Restart main Agent services to pick up config changes
```
clarkb7, line 60: Why do we need to restart the Agent service? And why are we doing so before creating the DDOT service? I see the Linux package is doing the same, so lets leave it for now and discuss further with the DDOT and fleet teams.

Some more context/thoughts:
- We might not want package code to touch the Agent services. The caller [(setup)](https://github.com/DataDog/datadog-agent/blob/0874460d9bf6bd9d04224e5f091c67d00aa69a8e/pkg/fleet/installer/setup/common/setup.go#L150-L153) already restarts the services after running all the package install hooks. We don't support remote install yet, but I imagine we'd do the same there, and not have the Agent restart managed by subpackages.
- there's also some future work to support installing but not starting the services https://datadoghq.atlassian.net/browse/WINA-1270

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
```diff
@@ -0,0 +1,366 @@
...
+	if err = windowssvc.NewWinServiceManager().RestartAgentServices(ctx.Context); err != nil {
+		return fmt.Errorf("failed to restart agent services: %w", err)
+	}
+	// 4) Ensure DDOT service exists/updated, then start it (best-effort)
+	if err = ensureDDOTService(); err != nil {
+		return fmt.Errorf("failed to install ddot service: %w", err)
+	}
+	// Optional: allow E2E packaging tests to skip starting the service for installation test
+	if skipDDOTServiceStart {
```
clarkb7: do we need this or is it a leftover iteration/debugging helper?

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
```diff
@@ -0,0 +1,366 @@
...
+	}
+	if err = waitForServiceRunning(otelServiceName, 30*time.Second); err != nil {
+		log.Warnf("DDOT: service %q did not reach Running state: %s", otelServiceName, err)
+		return nil
+	}
+	return nil
+}
+
+// waitForServiceRunning waits until the given Windows service reaches the Running state or times out
+func waitForServiceRunning(name string, timeout time.Duration) error {
```
clarkb7: if the service stops then this will unnecessarily wait for the full timeout. `StartService` puts the service into `SERVICE_START_PENDING` state, so we can wait for the state to change and then check if it is `RUNNING` (success) or else fail. We might be able to re-use an existing helper, or a similar one, see [waitForPendingStateChange](https://github.com/DataDog/datadog-agent/blob/a3e7933b3c2e866ced6161c09e1aee80f5ac613b/pkg/util/winutil/service.go#L358)

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
```diff
@@ -0,0 +1,366 @@
...
+	// Optional: allow E2E packaging tests to skip starting the service for installation test
+	if skipDDOTServiceStart {
+		log.Warnf("DDOT: skipping service start due to skipDDOTServiceStart=%t", skipDDOTServiceStart)
+		return nil
+	}
+	// Start DDOT only when core Agent is running and credentials exist
+	if !isServiceRunning(coreAgentService) {
+		log.Warnf("DDOT: skipping service start (core Agent not running)")
+		return nil
+	}
```
clarkb7: because of the earlier agent restart this might be racy and fail if the Agent hasn't yet transitioned from `SERVICE_START_PENDING` to `SERVICE_RUNNING`

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
```diff
@@ -0,0 +1,366 @@
...
+
+// preRemoveDatadogAgentDdot performs pre-removal steps for the DDOT package on Windows
+// All the steps are allowed to fail
+func preRemoveDatadogAgentDdot(ctx HookContext) error {
+	_ = stopServiceIfExists(otelServiceName)
+	_ = deleteServiceIfExists(otelServiceName)
+
+	// Best effort remove otel-agent.exe from the package’s stable path
+	bin := filepath.Join(paths.PackagesPath, agentDDOTPackage, "stable", "embedded", "bin", "otel-agent.exe")
+	_ = os.Remove(bin)
```
clarkb7: why do you delete the binary here? It should be deleted by the [repository manager.](https://github.com/DataDog/datadog-agent/blob/a3e7933b3c2e866ced6161c09e1aee80f5ac613b/pkg/fleet/installer/installer.go#L648-L651)

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
```diff
@@ -0,0 +1,366 @@
...
+	}
+	var existing map[string]any
+	if err := yaml.Unmarshal(data, &existing); err != nil {
+		return err
+	}
+	if existing == nil {
+		existing = map[string]any{}
+	}
+	existing["otelcollector"] = map[string]any{"enabled": true}
+	existing["agent_ipc"] = map[string]any{"port": 5009, "config_refresh_interval": 60}
```
clarkb7: lets ask DDOT team why we need to set the `agent_ipc` option

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
guohdd: Made the function shared.

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
guohdd: done.

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
guohdd: done

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
guohdd: great suggestion. Made changes.

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
guohdd: good point that we might not want package code to touch the Agent services. Restarting the main service ensures the needed IPC channel is in place for the OTEL service. We will discuss further with DDOT team.

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
guohdd: This is a leftover and I think it might be useful so didn't remove it. It can definitely be removed.

.gitlab/packaging/oci.yml
```diff
@@ -75,6 +75,50 @@
...
       - ${OMNIBUS_PACKAGE_DIR}
 
+.package_oci_win:
+  extends: .package_oci
+  script:
+    - set -u pipefail
+    - PIPE_DIR="$OMNIBUS_PACKAGE_DIR/pipeline-$CI_PIPELINE_ID"
+    - ARTIFACTS=$(ls -1 "${PIPE_DIR}"/${WIN_ARTIFACT_PATTERN} 2>/dev/null)
+    - '[ -n "$ARTIFACTS" ] || { echo "No suitable input artifact found"; exit 1; }'
+    - '[ "$(printf ''%s\n'' "$ARTIFACTS" | wc -l)" = 1 ] || { echo "Expected exactly one candidate input artifact"; exit 1; }'
```
alopezz: A recommendation for avoiding all the escaping here (which somewhat affects the readability) would be to use multiline with `- |`, even if for a single line (though here you may also group all the validations together, since the error message will tell you which failed anyway).

I'd also prefer plan `wc -w` over combining `ls -1` with `wc -l` (as in my [previous comment](https://github.com/DataDog/datadog-agent/pull/40119#discussion_r2353204183)).
```suggestion
    - |
      [ "$(echo "$ARTIFACTS" | wc -w)" = 1 ] || { echo "Expected exactly one candidate input artifact"; exit 1; }
```

.gitlab/packaging/oci.yml
```diff
@@ -75,6 +75,50 @@
...
+    - 'ARTIFACT=$(printf ''%s'' "$ARTIFACTS")'
```
alopezz: I wonder what this is needed for? We could just name the variable ARTIFACT from the start.

.gitlab/packaging/oci.yml
alopezz: But `agent_win_oci` already depends on `windows_msi_and_bosh_zip_x64-a7`, so all necessary dependencies will be satisfied. From the point of view of this specific job we don't want the msi produced by that job. It also means we unnecessarily retrieve artifacts from that job, which is not really used here.

alopezz (commented):
Almost there as far as we're concerned. The main thing that I'd like to clear up before approving is the unnecessary dependency thread.

BaptisteFoy (approved):
LGTM for Fleet

khewonc (approved):
LGTM for agent onboarding file

.gitlab/packaging/oci.yml
guohdd: Great suggestions. Made changes per code review. Regarding using "wc -w", I wanted to cover the case that space(s) would be part of the file name. However, if we are sure it will never happen, I can use "wc -w".

alopezz: On second thought, I think you're probably right about `ls -1 | wc -l`.

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
guohdd: Good point. We do not need it here.

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
guohdd: Good point, changes made to avoid race condition.

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
guohdd: Great catch. Changes made to address this.

.gitlab/packaging/oci.yml
guohdd: Ok, got it. Thanks for the clarification. Change was made per your suggestion.

pkg/fleet/installer/packages/datadog_agent_ddot_windows.go
guohdd: the flag was removed

alopezz (approved):
Thanks for your patience!

clarkb7 (approved):

PR #40133, opened 2025-08-22
Remove runtime installation of shellcheck in CI

ofek:
### Motivation

The tool was added to the build images in https://github.com/DataDog/datadog-agent-buildimages/pull/954 and we started using a version of the images containing that change in https://github.com/DataDog/datadog-agent/pull/39896

### DIFF
--- a/.gitlab/lint/technical_linters.yml
+++ b/.gitlab/lint/technical_linters.yml
@@ -17,7 +17,6 @@ lint_licenses:
 lint_shell:
   extends: .lint
   script:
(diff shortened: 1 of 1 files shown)


Ishirui (approved):
Should we also remove the invoke task at the same time ? I don't think anyone uses it locally

ofek:
I would be reluctant to remove them right away without a replacement for installing/updating tools (no timeline yet on that as the details are fuzzy currently).

PR #40137, opened 2025-08-22
[CWS] Fix cgroup cache consistency and add cgroupfs fallbacks

spikat:
### What does this PR do?

This PR aims to fix some cgroups issues we found with our resolver cache:
* On cgroup_write events the cgroup cache was not updated accordingly (which led to a state where we spammed cgroup created/deleted notifications)
* The `CGroupID` path was inconsistent and contained the cgroupfs mount path if resolved during the snapshot
* PIDs can be assigned in multiple cgroups

This PR fixes those issues and add fallbacks and guards to ensure the cache consistency. From my tests I know that we really need those because we still miss (or run into a race?) some pid cgroup transitions.

It also fixes the cgroup scope for variables.

TODO:
* see if we can remove the `containersWorkload` cache to keep only one and simplify this way the code.

### DIFF
--- a/pkg/security/resolvers/cgroup/resolver.go
+++ b/pkg/security/resolvers/cgroup/resolver.go
@@ -11,6 +11,7 @@ package cgroup
 import (
 	"context"
 	"fmt"
+	"slices"
 	"sync"
 
 	"github.com/hashicorp/golang-lru/v2/simplelru"
@@ -21,7 +22,6 @@ import (
 	cgroupModel "github.com/DataDog/datadog-agent/pkg/security/resolvers/cgroup/model"
 	"github.com/DataDog/datadog-agent/pkg/security/secl/containerutils"
 	"github.com/DataDog/datadog-agent/pkg/security/secl/model"
-	"github.com/DataDog/datadog-agent/pkg/security/seclog"
 	"github.com/DataDog/datadog-agent/pkg/security/utils"
 )
 
@@ -45,8 +45,8 @@ const (
 type ResolverInterface interface {
 	Start(context.Context)
 	AddPID(*model.ProcessCacheEntry)
-	GetWorkload(containerutils.ContainerID) (*cgroupModel.CacheEntry, bool)
 	DelPID(uint32)
+	GetWorkload(containerutils.ContainerID) (*cgroupModel.CacheEntry, bool)
 	Len() int
 	RegisterListener(Event, utils.Listener[*cgroupModel.CacheEntry]) error
 }
@@ -55,6 +55,7 @@ type ResolverInterface interface {
 type Resolver struct {
 	*utils.Notifier[Event, *cgroupModel.CacheEntry]
 	sync.Mutex
+	cgroupFS           *utils.CGroupFS
 	statsdClient       statsd.ClientInterface
 	cgroups            *simplelru.LRU[model.PathKey, *model.CGroupContext]
 	hostWorkloads      *simplelru.LRU[containerutils.CGroupID, *cgroupModel.CacheEntry]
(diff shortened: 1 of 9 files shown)


pkg/security/resolvers/cgroup/resolver.go
```diff
@@ -145,6 +192,42 @@ func (cr *Resolver) AddPID(process *model.ProcessCacheEntry) {
...
+	defer cr.Unlock()
+
+	found := false
+
+	for _, cgroup := range cr.hostWorkloads.Values() {
+		if cgroup.CGroupFile == process.CGroup.CGroupFile {
+			cgroup.PIDs[process.Pid] = true
+			found = true
+		} else if _, exist := cgroup.PIDs[process.Pid]; exist {
+			delete(cgroup.PIDs, process.Pid)
```
safchain, line 213: not sure to understand why we delete here if the pid is present.

spikat: Otherwise the same pid could be present on multiple cache cgroups.
It can happen when we miss a pid cgroup transition.

pkg/security/rules/engine_linux.go
```diff
@@ -31,13 +32,12 @@ func (e *RuleEngine) GetSECLVariables() map[string]*api.SECLVariableState {
...
-			if containerWorkloads == nil {
-				continue
-			}
 
-			for _, cgce := range containerWorkloads.Values() {
+			cgr.Iterate(func(cgce *cgroupModel.CacheEntry) {
 				cgce.RLock()
+				if cgce.ContainerContext.ContainerID == "" {
+					return
+				}
```
YoannGh, line 41: Can we move this after the next line to make sure `cgce.RUnlock()` is called?

spikat: Nice catch, it's fixed :)

pkg/security/resolvers/process/resolver_test.go
```diff
@@ -830,15 +830,15 @@ func TestIsExecExecSnapshot(t *testing.T) {
 	child3 := newFakeExecEvent(3, 4, 769, resolver)
 
 	// X(pid:3)
-	resolver.insertEntry(parent.ProcessCacheEntry, model.ProcessCacheEntryFromSnapshot)
+	resolver.insertEntry(parent.ProcessCacheEntry, model.ProcessCacheEntryFromSnapshot, true)
```
safchain, line 833: nit : I think it should be false ? this is a fork

pkg/security/resolvers/process/resolver_test.go
```diff
@@ -830,15 +830,15 @@ func TestIsExecExecSnapshot(t *testing.T) {
...
 	assert.Equal(t, parent.ProcessCacheEntry, resolver.entryCache[parent.ProcessCacheEntry.Pid])
 	assert.Equal(t, 1, len(resolver.entryCache))
 
 	// X(pid:3)
 	//    |
 	// X(pid:4)
 	resolver.setAncestor(child.ProcessCacheEntry)
-	resolver.insertEntry(child.ProcessCacheEntry, model.ProcessCacheEntryFromSnapshot)
+	resolver.insertEntry(child.ProcessCacheEntry, model.ProcessCacheEntryFromSnapshot, true)
```
safchain, line 841: same here ?

safchain (approved):

pkg/security/resolvers/process/resolver_test.go
spikat: it should be true on forks (and false for execs) :)

YoannGh (approved):

PR #40146, opened 2025-08-22
feat(ssi): add memory limit buffer for emptyDir

betterengineering:
**⚠️ Note - this change is being merged to a feature branch that we don't expect to merge into main ⚠️**

### What does this PR do?
This change adds a buffer container that requests the additional memory required for the `emptyDir` when the limit is set.

### Motivation
We are building a feature for a customer that allows them to use a memory backed `emptyDir`. As part of that change, we have added the ability to set the limit to control the size of the `emptyDir`. The problem is, the additional memory used for the tmpfs is not accounted for anywhere in the pod resources. With this change, we will account for the additional memory used by the tmpfs so that the pod will not be OOMKilled.

I've been using injector-dev:
```bash
injector-dev apply -f storage-constraints.yaml --build
```

<details>
  <summary>storage-constraints.yaml</summary>

```yaml
operator:
  apps:
  - name: python
    namespace: application
    values:
      env:
      - name: DD_TRACE_DEBUG
        value: "true"
        value: "true"
      image:
        repository: registry.ddbuild.io/ci/injector-dev/python
        tag: 2cd78ded
      podLabels:
        language: python
        tags.datadoghq.com/env: local
      service:
        port: "8080"
  versions:
    operator: 1.17.0
    cluster_agent:
      build: {}
    agent: 7.69.1
    injector: 0.44.0
  config:
    apiVersion: datadoghq.com/v2alpha1
    kind: DatadogAgent
    metadata:
      name: datadog
    spec:
      override:
        clusterAgent:
          # image:
          #   name: "datadog/cluster-agent-dev:mark-spicer-inplat-772-add-library-storage-medium"
          env:
            - name: "DD_ADMISSION_CONTROLLER_AUTO_INSTRUMENTATION_VOLUME_EMPTY_DIR_STORAGE_MEDIUM"
              value: "memory"
            - name: "DD_ADMISSION_CONTROLLER_AUTO_INSTRUMENTATION_VOLUME_EMPTY_DIR_STORAGE_LIMIT"
              value: "100Mi"
      features:
        apm:
          instrumentation:
            enabled: true
            targets:
            - ddTraceVersions:
                python: default
              name: python
              podSelector:
                matchLabels:
                  language: python
```

</details>

I tested that with the storage limit, we get the buffer container with limits set. Without the storage limit, we don't get the buffer container. I also tested with multiple SDKs to ensure that only one container was added.

I technically am not checking the type of storage before setting the limit. I set `10m` for CPU request/limits to ensure there is a QoS class associated, but with 100 pods on a node, that would account for an additional core. 

### Additional Notes

I've chosen to use the injector for the image. That feels safe given they will already need that image to be available.

### DIFF
--- a/pkg/clusteragent/admission/mutate/autoinstrumentation/mutators.go
+++ b/pkg/clusteragent/admission/mutate/autoinstrumentation/mutators.go
@@ -105,6 +105,26 @@ func (i initContainer) mutatePod(pod *corev1.Pod) error {
 	return nil
 }
 
+type sleepContainer struct {
+	corev1.Container
+	Prepend bool
+}
+
+var _ podMutator = (*sleepContainer)(nil)
+
+func (s sleepContainer) mutatePod(pod *corev1.Pod) error {
+	container := s.Container
+	for idx, c := range pod.Spec.Containers {
+		if c.Name == container.Name {
+			pod.Spec.Containers[idx] = container
+			return nil
+		}
+	}
+
+	pod.Spec.Containers = appendOrPrepend(container, pod.Spec.Containers, s.Prepend)
+	return nil
+}
+
 // volume is a podMutator which adds the volume to a pod.
 //
 // It will only add the volume one time based on the volume name.
--- a/pkg/clusteragent/admission/mutate/autoinstrumentation/namespace_mutator.go
+++ b/pkg/clusteragent/admission/mutate/autoinstrumentation/namespace_mutator.go
@@ -15,6 +15,7 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/client-go/dynamic"
 
 	"github.com/DataDog/datadog-agent/comp/core/config"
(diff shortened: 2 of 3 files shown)


releasenotes-dca/notes/add-memory-limit-buffer-482a6fbb7cf14ebc.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    When the Single Step Instrumentation webhook is configured to use a library storage limit, we now add a buffer
+    container to the Pod to request the amount of memory needed for the emptyDir. This will allow Kubernetes to properly
+    account for the emptyDir usage in the Pod's memory limit.
```
estherk15:
```suggestion
    When the Single Step Instrumentation webhook is configured with a library storage limit, the Pod now includes a
    buffer container that requests memory for the `emptyDir`. This will allow Kubernetes to properly
    account for the `emptyDir` usage in the Pod's memory limit.
```

estherk15 (approved):

pkg/clusteragent/admission/mutate/autoinstrumentation/mutators.go
```diff
@@ -105,6 +105,26 @@ func (i initContainer) mutatePod(pod *corev1.Pod) error {
...
+var _ podMutator = (*sleepContainer)(nil)
+
+func (s sleepContainer) mutatePod(pod *corev1.Pod) error {
+	container := s.Container
+	for idx, c := range pod.Spec.Containers {
+		if c.Name == container.Name {
+			pod.Spec.Containers[idx] = container
+			return nil
+		}
+	}
```
knusbaum, line 122: This just replaces the sleep container if there's already one there?

Will that ever happen? I don't mind it exactly, I just find it spooky we might be calling mutate on a pod with a sleep container more than once.

knusbaum (approved):

pkg/clusteragent/admission/mutate/autoinstrumentation/mutators.go
betterengineering: Yes, this was copy paste from other mutators. I assumed there was a good reason for it and kept it here in the off chance the mutator was called multiple times.

pkg/clusteragent/admission/mutate/autoinstrumentation/namespace_mutator.go
```diff
@@ -130,6 +131,29 @@ func (m *mutatorCore) mutatePodContainers(pod *corev1.Pod, cm containerMutator)
 	return mutatePodContainers(pod, filteredContainerMutator(m.config.containerFilter, cm))
 }
 
+func (m *mutatorCore) genSleepContainer(image string) podMutator {
```
adel121, line 134:
```suggestion
func (m *mutatorCore) getSleepContainer(image string) podMutator {
```

pkg/clusteragent/admission/mutate/autoinstrumentation/namespace_mutator.go
```diff
@@ -171,6 +196,10 @@ func (m *mutatorCore) injectTracers(pod *corev1.Pod, config extractedPodLibInfo)
 		}
 	)
 
+	if m.config.libraryStorageLimit != nil {
+		podMutators = append(podMutators, m.genSleepContainer(injector.image))
```
adel121, line 200:
```suggestion
        podMutators = append(podMutators, m.getSleepContainer(injector.image))
```

adel121 (commented):
Looks good. 

Small nit.

PR #40153, opened 2025-08-22
[ABLD-135] Add initial `bazel` jobs

rdesgroppes:
### What does this PR do?
Now `bazelisk` is available on all our CI executors ([macOS](https://github.com/DataDog/ci-platform-machine-images/pull/395) runners, [Linux](https://github.com/DataDog/datadog-agent-buildimages/pull/951) & [Windows](https://github.com/DataDog/datadog-agent-buildimages/pull/953) containers), this change secures an initial part of our `bazel` setup while providing reusable job templates to ease caching in CI.

Practically, it adds a handful of jobs focusing on **building `bazel` dependencies** in the corresponding GitLab `deps_build` stage.

Overall, it consists in:
1. verifying **`bazelisk` properly bootstraps `bazel` across all platforms**, (primary scope of the PR)
2. extracting initial `bazel:`-prefixed CI job templates and dogfooding them, (i.e. by building deps)
3. binding `bazelisk`/`bazel` caches to GitLab caching capabilities (runner-based as of now): installed binaries, "repository cache", "repo contents cache", and "disk cache" are all saved/restored correctly to/from GitLab while honoring OS/architecture boundaries,
4. marking in-workspace cache in both `.bazelignore` and `.gitignore`,
5. adjusting code/job ownership accordingly.

### Motivation
Securing some initial `bazel` configuration in CI :
- ensures all platforms are kept in the radar over next iterations,
- prevents regressions over next iterations,
- establishes foundations for future jobs.

- on Windows, compilation errors made be descope dependencies being built to **only `bzip2`**. This will be of course addressed in a subsequent PR,
- GitLab caching is still runner-based for the time being, which can be revisited later. (like everything)

### Additional Notes
Main addressed challenges:
- cache location conflicts[^1]: externalized "repo contents cache" through `tools/bazel*` wrappers:
  - bazelbuild/bazel#26384
  - bazelbuild/bazel#26522
  - bazelbuild/bazel#26773
  - bazelbuild/bazel#26802
- macOS runners:
  - `bzip2` dependency: fetch from a reachable source:
    - [x] DataDog/datadog-agent#40219
- Windows:
  - `bazel` spawn strategy: fallback to a permissive strategy since `sandboxed` is unsupported
    - [x] DataDog/datadog-agent#40328
  - `tools/bazel` wrapper: fallback to batch (`tools/bazel.bat`), since `bash` is discouraged by `bazel` in this case and `tools/bazel.ps1` poses detection problems,
  - long paths were supported in containers but not on runners:
    - [x] DataDog/ci-platform-machine-images#429
  - recursive symlinks: use `robocopy` instead of `copy`/`move`/`xcopy`.
  
[^1]:
> ERROR: The repo contents cache [/path/to/datadog-agent/.cache/repo_contents] is inside the workspace [/path/to/datadog-agent]. This can cause spurious failures. Disable the repo contents cache with `--repo_contents_cache=`, or specify `--repo_contents_cache=<path outside the workspace>`.


.github/CODEOWNERS
```diff
@@ -749,6 +750,7 @@
 /test/regression/cases/docker_containers*     @DataDog/single-machine-performance @DataDog/container-integrations
 
 /tools/                                 @DataDog/agent-devx
+/tools/bazel*                           @DataDog/agent-build
```
CelianR, line 754: NIT: Could you also add `.bazelignore`?

CelianR (approved):

.github/CODEOWNERS
rdesgroppes: Good point, please hold the line

rdesgroppes: @CelianR it happens to be already covered by:
https://github.com/DataDog/datadog-agent/blob/e70c1bec0e5d63a36e51af79a3c235ab44ee4236/.github/CODEOWNERS#L27

CelianR: Oops, nice catch 😄 LGTM then!

.gitlab/bazel/include.yaml
```diff
@@ -0,0 +1,50 @@
...
+        prefix: bazel-$CI_RUNNER_DESCRIPTION
+        files: [ .bazelversion ]
+      paths:
+        - .cache/bazelisk
+        - .cache/bazel/install
+      policy: pull-push
+      when: on_success
+    - key: bazel-$CI_RUNNER_DESCRIPTION
+      paths:
+        - .cache/bazel/cache
```
JSGette: Do we really want to have this? CI Infra team is working on persistent runners this sprint so we should be able to reuse directories on the workers, sharing disk cache via Gitlab will actually add to the Runner's startup time.

https://datadoghq.atlassian.net/browse/CIEXE-1222

.gitlab/deps_build/deps_build.yml
```diff
@@ -122,3 +122,41 @@ build_processed_btfhub_archive:
     - dda inv -- -e system-probe.process-btfhub-archive --branch $BTFHUB_ARCHIVE_BRANCH
```
JSGette: I would rather have a separate yaml file to include it to keep the current `deps_build.yml` clean. Something like `deps_bazel_build.yml`

.gitlab/deps_build/deps_build.yml
```diff
@@ -122,3 +122,41 @@ build_processed_btfhub_archive:
...
+  tags: [ "macos:ventura-arm64", "specific:true" ]
+  script:
+    - bazel build //deps/...
+
+bazel:build-deps:windows-amd64:
+  extends: [ .bazel:build-deps, .bazel:ext:windows ]
+  variables:
+    ARCH: x64
+    SCRIPT: |-
+      bazel build @bzip2//...
```
JSGette: Why is the target different for `Windows`? I would at least write a comment or add a todo hinting that Windows support isn't complete yet

tools/bazel
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Check `bazelisk` properly bootstraps `bazel` or fail with instructions
+if [ -z "${BAZEL_REAL:-}" ] || [ "${BAZELISK_SKIP_WRAPPER:-}" != "true" ]; then
+  cat >&2 "$(dirname "$0")/bazelisk.md"
+  exit 1
+fi
+
+# Not in CI: simply execute `bazel` - done
```
JSGette, line 10: Why?

JSGette (approved):
It's a banger!

.gitlab/bazel/include.yaml
rdesgroppes: > Do we really want to have this?

Yes.
tl;dr: I approach my PRs with a “stage objective” in mind.

This probably comes from my habit of working in environments where it’s not acceptable to deliver intermediate results that aren’t self-contained, even if we expect them to be replaced soon. In my experience, “soon” often stretches out when we rely on other teams who naturally have different priorities.

That’s why I prefer to avoid situations where, even for a few days, the CI experience is degraded (for example, downloading dependencies on every single run). To me, it’s worth having this temporary stage in place while waiting for the final solution.

.gitlab/deps_build/deps_build.yml
rdesgroppes: ✅ Done in second commit.
> to keep the current `deps_build.yml` clean.

... not only the said file is kept clean, but the extraction also allows to bother fewer reviewers.

.gitlab/deps_build/deps_build.yml
rdesgroppes: > ### Possible Drawbacks / Trade-offs
> - on Windows, compilation errors made be descope dependencies being built to **only `bzip2`**. This will be of course addressed in a subsequent PR,

rdesgroppes: ✅ Answered in 3rd commit, which adds the following message: (better than a comment to raise awareness)
> 🟡 TODO(regis): compilation errors remain to be addressed - limiting to a working subset for the time being

PR #40162, opened 2025-08-22
Streamline the PR template.

aiuto:
Streamline the PR template.

- Drop the advise to read CONTRIBUTING.md. That is standard practice for any OSS project.
   - Everyone who has touched github should know that.
   - If the rare case that someone is completely new to OSS development, and has obviously not read any guidelines, we should be able to graciously point them at the rules.
- Push mandates like adding tests and having them pass into the contributing guidelines.
   - Update the contributing guidelines to say Draft first, then move to Ready to Review after the tests pass
   - Remove sections that are often left blank.


aiuto:
This conflicts with https://github.com/DataDog/datadog-agent/pull/40160

It is roughly the same but goes further.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,35 +1,9 @@
...
-* Contributors are encouraged to read our [CONTRIBUTING](/CONTRIBUTING.md) documentation.
-* Both Contributor and Reviewer Checklists are available at https://datadoghq.dev/datadog-agent/guidelines/contributing/#pull-requests.
-* The pull request:
-  * Should only fix one issue or add one feature at a time.
-  * Must update the test suite for the relevant functionality.
-  * Should pass all status checks before being reviewed or merged.
-* Commit titles should be prefixed with general area of pull request's change.
-* Please fill the below sections if possible with relevant information or links.
--->
+This is just a template. Please delete all unneeded sections and add others if appropriate.
```
chouetz, line 1:
```suggestion
This is just a template. Can you please:
- Check first https://github.com/DataDog/datadog-agent/blob/main/docs/public/guidelines/contributing.md for more details.
- Delete all unneeded sections and add others if appropriate.
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,35 +1,9 @@
...
+
 ### What does this PR do?
 
 ### Motivation
```
chouetz, line 5: I prefer to have the `why` first, as the `what` should already be covered (partially) by the PR title. This `what` section could be optional to me, only when it requires additional information to describe the change (when this is more complex than an atomic change I suppose), wdyt?
```suggestion
### Motivation

### What changes did you make?
```

docs/public/guidelines/contributing.md
```diff
@@ -18,14 +18,15 @@ In order to ease/speed up our review, here are some items you can check/improve
 
 /// tab | Contributor Checklist
 - [X] Have a proper commit history (we advise you to rebase if needed) with clear commit messages.
+- [X] Sign your commits.
 - [X] Write tests for the code you wrote.
 - [X] Preferably make sure that all tests pass locally.
-- [X] Summarize your PR with an explanatory title and a message describing your changes, cross-referencing any related bugs/PRs.
+- [X] Summarize your PR with an explanatory title and a message describing your changes, cross-referencing any related bugs/PRs.  Commit titles should be prefixed with general area of pull request's change.
```
chouetz:
```suggestion
- [X] Summarize your PR with an explanatory title and a message describing your changes, cross-referencing any related bugs/PRs.  Commit titles should be prefixed with general area of pull request's change. You can use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) for the PR title.
```

docs/public/guidelines/contributing.md
```diff
@@ -89,6 +92,18 @@ The goals ordered by priority are:
 
 You must open the PR when the code is reviewable or you must set the PR as draft if you want to share code before it's ready for actual reviews.
 
+#### Write a good PR description
+
+Reviewers and future maintainers will not see your individual commit history, so the PR description should incorporate everything that they will need.
+
+- A description of what is changed.
+- Often, a reason why the change is made. Pointing to an issue is usually a good reason.
```
chouetz: Here as well I think the `why` is the first thing that should be in the description, mostly because what is changed can be inferred from the code change. And in the case of atomic changes this is often redundant with the PR title. WDYT?

docs/public/guidelines/contributing.md
```diff
@@ -89,6 +92,18 @@ The goals ordered by priority are:
 
 You must open the PR when the code is reviewable or you must set the PR as draft if you want to share code before it's ready for actual reviews.
 
+#### Write a good PR description
+
+Reviewers and future maintainers will not see your individual commit history, so the PR description should incorporate everything that they will need.
```
chouetz:
```suggestion
Reviewers and future maintainers will not see your individual commit history, so the PR description (which is included in the [merge commit](https://github.com/DataDog/datadog-agent/commit/350038eda272012da77904f83a921b99be87ae69)) should incorporate everything that they will need.
```

chouetz (approved):
Few comments, most probably minor, so feel free to update or merge as is, thanks.

chouetz:
> This conflicts with #40160
> 
> It is roughly the same but goes further.

I've done comments on this one and closed the duplicate, thanks

.github/PULL_REQUEST_TEMPLATE.md
aiuto: I kept the order that was in the doc.
I've always used what, because that is what I want first when looking at the commit comment later.  Why is usually just a link to a design or issue. 
For a big change, you settle the why once, and may have 10 changes that get their incrementally, and the why is just that identical pointer.
So I think of what as mandatory and why as optional.

.github/PULL_REQUEST_TEMPLATE.md
aiuto: CONTRIBUTING is implied. See the PR description for why I think the link is noise.

docs/public/guidelines/contributing.md
aiuto: I think we are coming at the value of descriptions from different points of view.
I equally weight two things. 
1. make it easy for the reviewer
2. provide a meaningful commit description so that a future maintainer can see what happened to a file over time from the git log.

We also may be thinking of "why" in 2 different ways. Sometimes you make a change in a particular way and you want to explain why you used that implementation over another. That "why" is part of the "what" section. 

The other "why" is the motivation for making the change at all.  In daily work I find this mostly reduces to "more work on project X".  When a team is aligned on a feature, and sprinting on it, most "why"s should point to the same RFC.

docs/public/guidelines/contributing.md
aiuto: I don't like either one. How about
```
Reviewers and future maintainers only see the PR description, not your individual commits, so the description should incorporate everything that they will need.  The merge commit will may include PR descriptions from multiple PRs, it so the description should tie back to the changed code in some way.  For example `run mdformat on iot-agent team docs` rather than just `mdformat the docs`.

Comment (chouetz) [reply docs/public/guidelines/contributing.md]: I'm fine with this one, I put it as a suggestion with minor updates
```suggestion
Reviewers and future maintainers only see the PR description, not your individual commit history, so the description should incorporate everything that they will need. The merge commit may include PR descriptions from multiple PRs, so the description should tie back to the changed code in some way. For example `run mdformat on iot-agent team docs` rather than just `mdformat the docs`.```

File: docs/public/guidelines/contributing.md (hunk 4 above)
Comment (chouetz) [reply docs/public/guidelines/contributing.md]:
```suggestion
- A description of what is changed.
- A reason why the change is made. Pointing to an issue is usually a good reason.
```
Would you agree to just dismiss the `often`? I just want to avoid a situation where you investigate something broken, and you can clearly understand the change, but if you miss the motivation there is a risk of regression: as context evolved, this reason for change is not obvious and might be something you are not aware anymore. This is not necessarily self-explained in the code or comments and I think it deserves to be in the PR description.

.github/PULL_REQUEST_TEMPLATE.md
chouetz: I'm ok to keep the order, though I don't think this is optional: if you don't have a motivation you don't do a code change 😄. If you agree I would just like it not to become optional (see below).

pgimalac: FWIW I have a strong preference for "What" then "Why"
The web-ui repo does it the other way around and I'm always super confused when I try to write the PR description (and often end up explaining `What` the change is in the `Why` section because otherwise it's difficult to explain why I'm making it) 😄

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,35 +1,9 @@
...
-
-If you want additional validation by a second person, you can ask reviewers to do it. Describe how to set up an environment for manual tests in the PR description. Manual validation is expected to happen on every commit before merge.
-
-Any manual validation step should then map to an automated test. Manual validation should not substitute automation, minus exceptions not supported by test tooling yet.
--->
+### Describe how you validated your changes (if not by through tests)
 
 ### Possible Drawbacks / Trade-offs
-
-### Additional Notes
```
pgimalac, line 29: I actually regularly write things in the additional notes, I think it's useful and often provides very interesting information about the change (but I think we could drop the comment)

I've almost never seen anyone use the "possible drawbacks" section though

chouetz: The `drawbacks/trade-offs` could be part of the `Additional Notes`, so I think we could keep this section instead of the previous one.

docs/public/guidelines/contributing.md
aiuto: SGTM

.github/PULL_REQUEST_TEMPLATE.md
aiuto: In the long run, it is all just "whatever makes sense for this PR".
I think we can hone this in future PRs.  The most important thing here is to reduce a giant pile of boilerplate that no one removes.

PR #40165, opened 2025-08-22
dyninst/testprogs: update dd-trace-go, stop replacing agent module

andreimatei:
The go.mod for the test programs was hacky -- it was replacing the remoteconfig module coming from the agent repo in order to support the remote config product related to SymDB. There's now a version of dd-trace-go that pulls in an update remoteconfig module, so this replacement is no longer necessary.


ajwerner (approved):
LGTM

PR #40166, opened 2025-08-22
 [AGNTLOG-243] Add Filtering to MRF Feature Followup

soberpeach:
### What does this PR do?

Allows the agent to send a filtered subset of logs to a backup different backup endpoint when MRF is triggered. Allowed logs services are set through the `multi_region_failover.logs_service_allowlist ` variable. 

The change only works when using the `http_sender`, and works in the following manner:
1. If MRF has been triggered, the `processor` will begin using the `multi_region_failover.logs_service_allowlist ` to filter logs by services. Allowed logs will be tagged and sent to the batcher.
2. The batcher will batch normal logs and MRF logs separately then send them each to the `sender/worker`. The MRF payloads will be tagged with an `isMRF` tag.
3. The sender/worker will send all logs to every destination, but will drop non-MRF payloads to non-MRF destinations. 

There is a bit of design change to how `batch_strategy` works now. It is now more of coordinator of individual `batches`: when it receives a message, it will forward it to the correct `batch` (creating it in the process if it has not been created already). This architecture allows some more flexibility in the future if we want to add more batches for various other things, such as sending logs from different sources to different endpoints, etc.

### Motivation

Needed by MRF

1. Add the following configuration to the `datadog.yaml`
```yaml
multi_region_failover:
  enabled: true
  site: 127.0.0.1:9999
  api_key: api_key
  failover_logs: true
  logs_service_allowlist:
    - mrf-service

logs_enabled: true
log_level: 'debug'
```
2. Have the logs agent tail two log files, one who's logs should be filtered and one who's shouldn't be:
```yaml
logs:
  - type: file
    path: "/var/log/myapp.log"
    service: "mrf-service"
    source: "mrf-source"
```
```yaml
logs:
  - type: file
    path: "/var/log/another_app.log"
    service: "non-mrf-service"
    source: "non-mrf-source"
```
and write logs to both files with the one-liners:
```bash
while true; do echo "$(date) - INFO: Application running" >> /var/log/myapp.log; sleep 1; done
```
```bash
while true; do echo "ANOTHER LOG" >> /var/log/another_app.log; sleep 1; done
```
3. Run the agent and check for the log: `Dropping non-MRF payload to MRF destination: https://agent-http-intake.logs.mrf.127.0.0.1:9999/api/v2/logs`

While it's not possible to tell exactly in which logs payloads have been dropped to the MRF destination while running the agent normally, if you debug the agent, you can see which payloads are making it through (payloads tagged with `IsMRF`).

Slight increase in memory as a result of holding two batches in the batcher as opposed to one.


brett0000FF (approved):

pkg/logs/sender/batch_strategy.go
```diff
@@ -138,133 +112,35 @@ func (s *batchStrategy) Start() {
 					// inputChan has been closed, no more payloads are expected
 					return
 				}
-				s.processMessage(m, s.outputChan)
+
+				if m.IsMRFAllow {
+					s.getBatch("mrf").processMessage(m, s.outputChan)
```
ddrthall: Nit: The biggest advantage to using a dictionary is that it allows us to do things like
s.getBatch(msg.DestinationSignature()).process....

We don't necessarily need to do that in the current MR, given we're still in the boolean state, but if we keep the current setup then I would recommend at making the "mrf" and "main" keys consts

pkg/config/setup/multi_region_failover.go
```diff
@@ -16,6 +16,7 @@ func setupMultiRegionFailover(config pkgconfigmodel.Setup) {
 	config.BindEnv("multi_region_failover.site")
 	config.BindEnv("multi_region_failover.dd_url")
 	config.BindEnv("multi_region_failover.metric_allowlist")
+	config.BindEnv("multi_region_failover.logs_service_allowlist")
```
aqian01: NIT: different naming then PR description

`multi_region_failover.logs_service_allowlist` vs `multi_region_failover.logs_allowlist`

pkg/logs/sender/batch_strategy.go
soberpeach: Ah okay makes sense, I'll change those to constants. But yeah, I think this change gets us most of the way there to a day where there are basically any number of endpoints.

pkg/config/setup/multi_region_failover.go
soberpeach: Good catch, changed it 👍

rahulkaukuntla (approved):

pkg/logs/sender/batch.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+	b.sendMessages(messagesMetadata, outputChan)
+}
+
+func (b *batch) sendMessages(messagesMetadata []*message.MessageMetadata, outputChan chan *message.Payload) {
+	defer b.resetBatch()
+
+	if err := b.compressor.Close(); err != nil {
+		log.Warn("Encoding failed - dropping payload", err)
+		// s.utilization.Stop()
```
ddrthall: I think this comment and the one on line 176 need to be added back in, I'm not seeing a replacement utilization stop line.

ddrthall (approved):

pkg/logs/sender/batch.go
soberpeach: Ah thanks for the catch, adding it back in now

PR #40170, opened 2025-08-22
Forbid calling SetWithoutSource with a struct, or outside test code

dustmop:
### What does this PR do?

Disallow calling SetWithoutSource unless in test code. This method should only be called from tests.

Disallow calling SetWithoutSource with a struct shaped parameter. Only plain data types, those that can be obtained by parsing yaml, should be assigned to the config's data model.

### Motivation

Stop relying on viper's implicit conversion of types. The in-memory config should have a consistent data model.

Unit tests

### DIFF
--- a/pkg/aggregator/demultiplexer_serverless.go
+++ b/pkg/aggregator/demultiplexer_serverless.go
@@ -58,7 +58,7 @@ func InitAndStartServerlessDemultiplexer(keysPerDomain map[string][]utils.APIKey
 	}
 	h, _ := hostname.Get(context.Background())
 	config := pkgconfigsetup.Datadog()
-	config.SetWithoutSource("serializer_compressor_kind", "none")
+	config.Set("serializer_compressor_kind", model.SourceAgentRuntime)
 	serializer := serializer.NewSerializer(forwarder, nil, selector.FromConfig(config), pkgconfigsetup.Datadog(), logger, h)
 	metricSamplePool := metrics.NewMetricSamplePool(MetricSamplePoolBatchSize, utils.IsTelemetryEnabled(pkgconfigsetup.Datadog()))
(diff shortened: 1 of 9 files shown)


remeh:
Since you are relying on the `test` build tag, can't you just have `SetWithoutSource` defined with the `test` build tag instead.
That way, the flavors without the `test` build tag won't compile at all if they use `SetWithoutSource`?

pkg/aggregator/demultiplexer_serverless.go
```diff
@@ -58,7 +58,7 @@ func InitAndStartServerlessDemultiplexer(keysPerDomain map[string][]utils.APIKey
 	}
 	h, _ := hostname.Get(context.Background())
 	config := pkgconfigsetup.Datadog()
-	config.SetWithoutSource("serializer_compressor_kind", "none")
+	config.Set("serializer_compressor_kind", model.SourceAgentRuntime)
```
vickenty: Looks like the value has gone missing. What does this set `serializer_compressor_kind` to?

dustmop: That was a mistake, caused by trying one fix then switching to another and mis-merging the result. It's fixed now.

dustmop:
> Since you are relying on the `test` build tag, can't you just have `SetWithoutSource` defined with the `test` build tag instead. That way, the flavors without the `test` build tag won't compile at all if they use `SetWithoutSource`?

Unfortunately this won't work. Most calls to this method are going through the Reader interface, sometimes but not always via the mock, so the `SetWithoutSource` method needs to be defined on that interface. Fixing that would be a larger change.

rahulkaukuntla (approved):

robertjli (approved):

vickenty (approved):

PR #40171, opened 2025-08-22
Fix README check example to work with new check API

nathan-b:
The "hello world" custom check example in the documentation does not actually work, as there are two main problems:

* The check class requires a constructor, which must in turn call the constructor for the AgentCheck class
* The YAML config for the new custom check needs an `instances` section which defines at least one instance

### What does this PR do?
Fix some documentation.

### Motivation
Our documentation should lead users down a path that works.

I followed the steps in the existing documentation. They did not work. I iterated until I got something working, then modified the documentation to match what I did to make it work.

### DIFF
--- a/docs/dev/checks/README.md
+++ b/docs/dev/checks/README.md
@@ -61,6 +61,10 @@ inherits from `AgentCheck` and implements the `check` method:
 from datadog_checks.checks import AgentCheck
 
 class MyCheck(AgentCheck):
+    def __init__(self, name, init_config, instances):
+        super().__init__(name, init_config, instances)
+        # Read config, set up instances, initialize checks
+        # ...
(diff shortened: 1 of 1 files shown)


louis-cqrl (approved):

GustavoCaso (approved):
LGTM. Curious to see what error shows where there is not `instances` section in the configuration 😄

nathan-b:
@GustavoCaso the error is:
```
CORE | WARN | (comp/core/autodiscovery/providers/config_reader.go:292 in collectEntry) | /media/psf/go/src/github.com/DataDog/datadog-agent/bin/agent/dist/conf.d/sigcheck.yaml is not a valid config file: Configuration file contains no valid instances
```

PR #39943, opened 2025-08-18
feat(otel-2754): add RUM <> OTLP mappings to pkg/opentelemetry-mapping-go

dd-jasminesun:
### What does this PR do?
Add RUM <> OTLP mappings that are already in opentelemtry-mapping-go upstream repo to pkg/opentelemetry-mapping-go in datadog-agent.

### Describe how you validated your changes
Screenshot 2025-08-18 at 3 26 52 PM

### Additional Notes
The upstream library PRs are here:
- https://github.com/DataDog/opentelemetry-mapping-go/pull/747
- https://github.com/DataDog/opentelemetry-mapping-go/pull/745
- https://github.com/DataDog/opentelemetry-mapping-go/pull/762

PR #40172, opened 2025-08-22
add module replace linter and gitlab test

jackgopack4:
### What does this PR do?
Updates all go.mod with modules.yml replace statements and attempts to add Gitlab and pre-commit checks to avoid this issue next time
### Motivation
failure to update go.mods after #39943 and no CI blocking merge


tasks/modules.py
```diff
@@ -363,3 +363,34 @@ def add_all_replace(ctx: Context):
...
+    result = ctx.run("git diff --exit-code **/go.mod", warn=True)
+
+    if result.exited is None or result.exited > 0:
+        # Files were modified, which means replace rules were missing
+        ctx.run("git diff --name-only **/go.mod", hide=False)
+        raise Exit(
+            code=1,
+            message=color_message(
+                "ERROR: Some go.mod files are missing replace rules. "
+                "Please run 'dda inv modules.add-all-replace' to fix them.",
```
nathan-b: Sorry for the probably dumb question, but in this case does the user actually need to run `dda inv modules.add-all-replace`, or did the previous call to `add_all_replace` already do that?

jackgopack4: no you're right; it should say something like "add changed files and try again"

jackgopack4: actually I've thought about it more and since this check will primarily be applied on CI, it should still advise the user to commit the changes from `dda inv modules.add-all-replace`

nathan-b (approved):
Based on my understanding this seems good from an agent runtimes perspective. Approving on behalf of that team.

ajgajg1134 (approved):

ofek (approved):
LGTM, though I wish there was a way to check without modifying files/the state of Git.

davidor (approved):

PR #40181, opened 2025-08-25
[ACIX-936] Migrate gitlab PATs

CelianR:
### What does this PR do?

> [!NOTE]
> Buildimages are also bumped in this PR.

This migrates Gitlab PATs, now we use a custom API to generate short lived Gitlab PATs.


.gitlab/trigger_release/agent.yml
```diff
@@ -16,7 +16,7 @@
   script:
     - !reference [.setup_dd_pkg]
     - RELEASE_VERSION="$(dda inv agent.version --url-safe --omnibus-format --pipeline-id $PARENT_PIPELINE_ID)-1" || exit $?; export RELEASE_VERSION
-    - GITLAB_TOKEN="$($CI_PROJECT_DIR/tools/ci/fetch_secret.sh $GITLAB_TOKEN write_api)" || exit $?; export GITLAB_TOKEN
+    - GITLAB_TOKEN="$(dda inv -- auth.gitlab)" || exit $?; export GITLAB_TOKEN
```
CelianR: For this one I suppose we need the token for `dd-pkg`

gjulianm (approved):
LGTM for ebpf-platform file (1)

.gitlab/trigger_release/agent.yml
spencergilbert: IIRC it's required to `--follow` the triggered downstream pipeline. Will the new token get the same permissions as the old one?

tasks/libs/ciproviders/gitlab_api.py
```diff
@@ -40,62 +41,38 @@
...
-                    return output.strip()
-            except subprocess.CalledProcessError:
-                print("GITLAB_BOT_TOKEN not found in keychain...")
-                pass
-        print("Please make sure that the GITLAB_BOT_TOKEN is set or that the GITLAB_BOT_TOKEN keychain entry is set.")
-        raise Exit(code=1)
-    return os.environ["GITLAB_BOT_TOKEN"]
+    if verbose:
+        # Prints also the scopes + the expiration date
+        print('Got Gitlab token:', res.json())
```
spencergilbert: Could this be considered a "leaked" token?

CelianR: Yes, it is useful to debug. Do you prefer printing everything except the token?

spencergilbert: 🤷 just thinking of the risk - if they're short lived, it's not _as_ risky. It's probably better to not print the token, if that's easy enough.

CelianR: Done ✅

spencergilbert (approved):

tbavelier (approved):
LGTM for @DataDog/agent-onboarding owned single file.

L3n41c (approved):

rdesgroppes (approved):
LG

tasks/auth.py
```diff
@@ -18,6 +19,13 @@ def datadog_infra(ctx, audience, datacenter="us1.ddbuild.io"):
     print(token)
 
 
+@task
+def gitlab(ctx, repo='datadog-agent', verbose=False):
+    """Get a gitlab token."""
+
+    print(get_gitlab_token(ctx, repo, verbose))
```
chouetz:
```suggestion
    if running_in_ci():
        raise Exit(message='This task is meant to be run locally, not in CI', code=1)
    
    if "GITLAB_TOKEN" in os.environ:
        print(os.environ["GITLAB_TOKEN"])
    else:
        print(get_gitlab_token(ctx, repo, verbose))
```
To be consistent with the GITHUB task

chouetz (changes requested):
I think we need to prevent the capacity to get a token in the CI

tasks/pipeline.py
```diff
@@ -342,15 +338,9 @@ def trigger_child_pipeline(_, git_ref, project_name, variable=None, follow=True,
...
-        if follow:
-            raise Exit("GITLAB_TOKEN variable needed to follow child pipelines.", 1)
-        else:
-            # The Gitlab lib requires `GITLAB_TOKEN` to be
-            # set, but trigger_pipeline doesn't use it
-            os.environ["GITLAB_TOKEN"] = os.environ['CI_JOB_TOKEN']
-
-    repo = get_gitlab_repo(project_name)
+    # Use the CI_JOB_TOKEN which is passed from gitlab
+    token = None if follow else os.environ['CI_JOB_TOKEN']
```
chouetz, line 342: Could we have this line in the `get_gitlab_repo` method (or even in the `get_gitlab_api` one). As such the token could be removed as argument of all methods except the closest one were we create the API context, and we either get the token from environ if it exists, or get one using the BTI API, wdyt?

chouetz (approved):
One minor suggestion, thanks in advance

tasks/pipeline.py
CelianR: I already tried once but this token has different permissions than the main one and it breaks some jobs if we use this token when this env var is set

PR #39946, opened 2025-08-18
enable ci vars build

jack0x2:
### What does this PR do?
This PR enables the passing of the driver versions through gitlab vars to the pipeline. This enables us to kick off the agent pipeline with testsigned drivers from the driver repo. 

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1741

### Describe how you validated your changes

Only adds variable names. Tested with following kicked off pipeline in driver:
https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/74141002

In which all of the windows build and tests successfully ran.

PR #40183, opened 2025-08-25
Adjust Omnibus env variable pass-through logic

alopezz:
### What does this PR do?

- Reuse existing logic for overriding windows driver variables if present in the environment.
- ~~Skip empty variables from the environment on the pass-through logic.~~ (can't be done, as we rely on [this](https://github.com/DataDog/datadog-agent-buildimages/blob/475a6a2b0bbb2e09d01b8e36fafcb63bc5b1a666/linux/Dockerfile#L218) empty `PKG_CONFIG_LIBDIR` variable).

### Motivation

Follow-up to #39946.

CI (and a look at the logs) should be enough.

### DIFF
--- a/tasks/omnibus.py
+++ b/tasks/omnibus.py
@@ -90,27 +90,19 @@ def get_omnibus_env(
 ):
     env = load_dependencies(ctx)
 
-    windows_only_vars = [
-        'WINDOWS_DDNPM_DRIVER',
-        'WINDOWS_DDNPM_VERSION',
-        'WINDOWS_DDNPM_SHASUM',
-        'WINDOWS_DDPROCMON_DRIVER',
-        'WINDOWS_DDPROCMON_VERSION',
-        'WINDOWS_DDPROCMON_SHASUM',
-    ]
-    # Discard windows variables when not on Windows
+    # Discard windows variables when not on Windows (so that they're not used in the cache key either)
     if sys.platform != 'win32':
+        windows_only_vars = [
+            'WINDOWS_DDNPM_DRIVER',
+            'WINDOWS_DDNPM_VERSION',
+            'WINDOWS_DDNPM_SHASUM',
+            'WINDOWS_DDPROCMON_DRIVER',
+            'WINDOWS_DDPROCMON_VERSION',
+            'WINDOWS_DDPROCMON_SHASUM',
+        ]
         for var in windows_only_vars:
             if var in env:
                 del env[var]
-
-    else:
-        # if any of windows vars set in env keep them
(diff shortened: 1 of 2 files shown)


tasks/omnibus.py
```diff
@@ -90,28 +90,20 @@ def get_omnibus_env(
...
+    # Discard windows variables when not on Windows (so that they're not used in the cache key either)
     if sys.platform != 'win32':
+        windows_only_vars = [
+            'WINDOWS_DDNPM_DRIVER',
+            'WINDOWS_DDNPM_VERSION',
+            'WINDOWS_DDNPM_SHASUM',
+            'WINDOWS_DDPROCMON_DRIVER',
+            'WINDOWS_DDPROCMON_VERSION',
+            'WINDOWS_DDPROCMON_SHASUM',
+        ]
```
rdesgroppes, line 102: Does it mean these environment variables will also be present for non-Windows jobs?

alopezz: They're populated by `load_dependencies` here from [release.json](https://github.com/DataDog/datadog-agent/blob/c764e4a7876814bc47105158cea5652640add247/release.json#L8-L14), regardless of OS:
https://github.com/DataDog/datadog-agent/blob/c764e4a7876814bc47105158cea5652640add247/tasks/omnibus.py#L91

tasks/omnibus.py
```diff
@@ -90,28 +90,20 @@ def get_omnibus_env(
...
             if var in env:
                 del env[var]
 
-    else:
-        # if any of windows vars set in env keep them
-        for key in windows_only_vars:
-            value = os.environ.get(key)
-            # Only overrides the env var if the value is a non-empty string.
-            if value:
-                env[key] = value
```
rdesgroppes, line 113: but the ability to override will be lost, no?

alopezz: Doesn't the override happen here?
https://github.com/DataDog/datadog-agent/blob/c5014096332e747cf64873364be1c9abff8b7931/tasks/omnibus.py#L198
As it's the last env manipulation here:
https://github.com/DataDog/datadog-agent/blob/c5014096332e747cf64873364be1c9abff8b7931/tasks/omnibus.py#L183-L184

alopezz: The only difference in behavior is if the variable is set to an empty string on the environment, which would probably simply break the build (which seems fine to me?).

rdesgroppes: I think so too.

rdesgroppes (approved):

PR #39898, opened 2025-08-17
usm,cnm: Change ownership of connections check

guyarb:
### What does this PR do?

Changes the ownership of connections check

### Motivation

The check belongs with CNM and USM teams rather than container-experiences

https://github.com/DataDog/datadog-agent/pull/39695#pullrequestreview-3106302263

PR #40184, opened 2025-08-25
fix(invoke-unit-tests): Do not use the regular CODEOWNERS file

chouetz:
### What does this PR do?
Prevent the `invoke-unit-tests` from relying on the regular CODEOWNERS file

### Motivation
Whenever an owner is changed, [it can have an impact on the UT](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1095803050). We don't want to create failures on tests on such changes.

- The tests are covered by UT, and each method can import its own CODEOWNER file.
- Will prevent the failure raised by https://github.com/DataDog/datadog-agent/pull/39898

### DIFF
--- a/tasks/unit_tests/junit_tests.py
+++ b/tasks/unit_tests/junit_tests.py
@@ -48,14 +48,14 @@ class TestSplitJUnitXML(unittest.TestCase):
 
     def test_without_split(self):
         xml_file = Path("./tasks/unit_tests/testdata/secret.tar.gz/bedroom-rspec-win2016-azure-x86_64.xml")
-        owners = read_owners(".github/CODEOWNERS")
+        owners = read_owners("./tasks/unit_tests/testdata/TEST_JUNIT_CODEOWNERS")
         self.assertEqual(junit.split_junitxml(xml_file.parent, xml_file, owners, {}, {}), 1)
         generated_folder = xml_file.parent / "windows-products_base"
         self.assertTrue(generated_folder.exists())
 
     def test_with_split(self):
         xml_file = Path("./tasks/unit_tests/testdata/secret.tar.gz/-go-src-datadog-agent-junit-out-base.xml")
-        owners = read_owners(".github/CODEOWNERS")
+        owners = read_owners("./tasks/unit_tests/testdata/TEST_JUNIT_CODEOWNERS")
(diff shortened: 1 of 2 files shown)


tasks/unit_tests/junit_tests.py
```diff
@@ -130,7 +131,7 @@ def test_success(self, mock_which, mock_check_call):
             for k in tmp_dir_vars:
                 self.assertIn(k, env)
             last_tmp_dir = env[tmp_dir_vars[-1]]
-            self.assertDictEqual({k: env[k] for k in tmp_dir_vars}, {k: last_tmp_dir for k in tmp_dir_vars})
+            self.assertDictEqual({k: env[k] for k in tmp_dir_vars}, dict.fromkeys(tmp_dir_vars, last_tmp_dir))
```
chouetz, line 134: This was changed by `ruff`

tasks/unit_tests/testdata/TEST_CODEOWNERS
```diff
@@ -0,0 +1,777 @@
+# Package code owners
```
Ishirui: We should probably use a more descriptive filename for this, something like `test_junit_upload_CODEOWNERS`, as we already have a codeowners file used for testing [here](https://github.com/DataDog/datadog-agent/blob/main/tasks/unit_tests/testdata/codeowners.txt) and it would be nice to be able to differentiate them easily

tasks/unit_tests/testdata/TEST_CODEOWNERS
```diff
@@ -0,0 +1,777 @@
...
+/.gitlab/source_test/go_generate_check.yml           @DataDog/agent-security
+
+/.gitlab/package_build/                              @DataDog/agent-build
+/.gitlab/package_build/windows.yml                   @DataDog/agent-build @DataDog/windows-products
+/.gitlab/package_build/installer.yml                 @DataDog/agent-build @DataDog/fleet
+/.gitlab/packaging/                                  @DataDog/agent-build
+
+/.gitlab/benchmarks/benchmarks.yml                   @DataDog/agent-apm
+
+/.gitlab/functional_test/regression_detector.yml     @DataDog/single-machine-performance
```
Ishirui: Do we need to define any owners other than these ? IIUC only job ownership is checked, not ownership for other files

Ishirui (approved):
👌

PR #40203, opened 2025-08-25
[OTEL-2753] Add infraattributes to OTLP ingest pipelines

jade-guiton-dd:
### What does this PR do?

This PR adds the infraattributes processor to the metrics and logs OTLP ingest pipelines.

This means that all OTLP ingest pipelines now include the infraattributes processor, which means that the code in the serializer exporter which uses the Agent's tagger component to add container tags, and which is only used by OTLP ingest pipelines, is now fully redundant. This PR removes said code and tears out the whole `tagEnricher` API.

The infraattributes processor adds Datadog-convention tags to the resource attributes of the input telemetry. After checking our translation code for metrics ([link 1](https://github.com/DataDog/opentelemetry-mapping-go/blob/c5054d91ff21b834d855081a64fa44bf1a5d5f8a/pkg/otlp/metrics/metrics_translator.go#L818), [link 2](https://github.com/DataDog/opentelemetry-mapping-go/blob/c5054d91ff21b834d855081a64fa44bf1a5d5f8a/pkg/otlp/attributes/attributes.go#L230)), I believe these will get translated into metric tags systematically, preserving the previous behavior of the serializerexporter.

### Motivation

This is a follow-up to #39455, which expanded the functionality of the infraattributes processor to add external container ID detection from auxiliary data. The main goal of this PR is to make this new feature available in OTLP ingest pipelines, where it is most useful, since it has no alternative for external container ID detection.

This should be covered by our existing E2E tests.

This is a breaking change in the serializerexporter's API.

### Additional Notes

The trace agent's OTLP endpoint has [similar logic](https://github.com/DataDog/datadog-agent/blob/4ddbbd77409a843e01864f8b1182533aba570c23/pkg/trace/api/otlp.go#L318) which uses the tagger to set container tags, which is also redundant. However, it seems that APM payloads draw a distinction between regular span attributes and actual "container tags" which are used for correlation: the latter are stored in the special `_dd.tags.container` attribute, usually in `TracerPayload.Tags`.

Removing this logic from the trace agent would require making sure that the container tags added by the infraattributes processor are recognized as such, and get added to `_dd.tags.container`. This is currently not the case, because the container tags are added as resource attributes with names using Datadog conventions, and the trace agent uses the [ContainerTagsFromResourceAttributes](https://github.com/DataDog/opentelemetry-mapping-go/blob/c5054d91ff21b834d855081a64fa44bf1a5d5f8a/pkg/otlp/attributes/attributes.go#L269) function, which only recognizes OTel semantic conventions and `datadog.tag.*` attributes.

I've decided to leave this additional cleanup as future work in OTEL-2764.


domalessi (approved):

liustanley (approved):
LGTM

PR #40204, opened 2025-08-25
[ACIX-982] Generate local github token

CelianR:
### What does this PR do?

> [!NOTE]
> You might want to disable whitespace changes on review.

This removes `GITHUB_TOKEN` usage on local tasks.
This variable is still used by Github Actions so it is not completely deprecated but we now use `ddtool` when there is no `GITHUB_TOKEN` variable.


tasks/libs/ciproviders/github_api.py
```diff
@@ -519,52 +518,39 @@ def _chose_auth(self, public_repo):
...
-            except subprocess.CalledProcessError:
-                print("GITHUB_TOKEN not found in keychain...")
-                pass
-        raise Exit(
-            message="Please create a 'repo' access token at "
-            "https://github.com/settings/tokens and "
-            "add it as GITHUB_TOKEN in your keychain "
-            "or export it from your .bashrc or equivalent.",
-            code=1,
+        assert "GITHUB_APP_ID" in os.environ and "GITHUB_KEY_B64" in os.environ, (
```
ofek: Let's instead do `if exit_condition: … raise Exit(...)` as we do below.

tasks/libs/ciproviders/github_api.py
```diff
@@ -794,3 +780,25 @@ def ask_review_actor(pr):
...
             return event.actor.name or event.actor.login
+
+
+def generate_local_github_token(ctx):
+    """
+    Generates a github token locally.
+    """
+
+    # Try without login first, then with login
+    for login in False, True:
```
ofek: We don't gain much by doing this and readability takes a hit. Let's instead put the logic for the non-login generation inside the `try` and the login approach during exception handling.

tasks/auth.py
```diff
@@ -24,3 +27,15 @@ def datadog_infra(ctx, audience, datacenter="us1.ddbuild.io", verbose=False):
         print(token)
 
     return token
+
+
+@task
+def github(ctx):
+    """Get a github token."""
+
+    assert not running_in_ci(), 'This task is meant to be run locally, not in CI'
```
ofek: Let's use `Exit` as mentioned in a previous comment.

ofek (changes requested):

spencergilbert (approved):

ofek (approved):
Thanks!

PR #40218, opened 2025-08-25
Allow older Ubuntu versions to be used as base for building ddot-byoc

alopezz:
### What does this PR do?

It makes it possible to build custom ddot (byoc) targeting a broader compatibility base (defined by available version of glibc on the OS), by allowing the choice of the base version of Ubuntu used, via `--build-arg UBUNTU_VERSION=xx.xx`.

### Motivation

Allow more users to benefit from the ddot byoc workflow.

Added a job to the pipeline which runs the build on Ubuntu 20.04 and checks that the maximum glibc version required by the binary doesn't exceed what would be expected for that version of Ubuntu.

- Not very user-friendly, but provides a quick way to provide this compatibility.

### Additional Notes

This pins the `dda` version in use and switches to its "standalone" version such that we don't need to worry about the python version or python env management.

In practice, by dropping to Ubuntu 20.04, we currently already see the glibc requirement drop to 2.17 max, which makes it at least as compatible as the regular Agent.

We eventually plan to package our C/C++ toolchains that target the compatibility that the Agent supports, which would probably give us a better way to provide the necessary compatibility without the need to change the base image like done here.

### DIFF
--- a/Dockerfiles/agent-ddot/Dockerfile.agent-otel
+++ b/Dockerfiles/agent-ddot/Dockerfile.agent-otel
@@ -1,8 +1,9 @@
 ARG AGENT_REPO=datadog/agent-dev
 ARG AGENT_VERSION=nightly-full-main-jmx
 ARG AGENT_BRANCH=main
+ARG UBUNTU_VERSION=24.04
 # Use the Ubuntu Slim AMD64 base image
-FROM ubuntu:24.04 AS builder
+FROM ubuntu:${UBUNTU_VERSION} AS builder
 
 # Set environment variables
 ARG AGENT_REPO
@@ -21,9 +22,6 @@ RUN apt-get update && \
     software-properties-common \
     build-essential \
     git \
-    python3 \
-    python3-pip \
-    python3-venv \
     rpm \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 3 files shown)


chouetz (approved):

Dockerfiles/agent-ddot/Dockerfile.agent-otel
```diff
@@ -1,8 +1,9 @@
 ARG AGENT_REPO=datadog/agent-dev
 ARG AGENT_VERSION=nightly-full-main-jmx
 ARG AGENT_BRANCH=main
+ARG UBUNTU_VERSION=24.04
 # Use the Ubuntu Slim AMD64 base image
-FROM ubuntu:24.04 AS builder
+FROM ubuntu:${UBUNTU_VERSION} AS builder
```
liustanley, line 6: Do we have to add tests for Ubuntu versions being used here other than 24.04 and 20.04, or are these the only versions with functional differences? I'm also wondering if we should restrict the versions that can be used with this script.

liustanley (approved):
LGTM, just one question on the Ubuntu versions

Dockerfiles/agent-ddot/Dockerfile.agent-otel
alopezz: I don't think it's worth being exhaustive here. The further back we go in versions, the more risk there is of something on the dockerfile breaking (that's why I had to change how dda is installed, for instance).

W.r.t. restrictions, the fact that we're relying on the user running `docker build` directly here doesn't leave a lot of room for logic. We _could_ just use 20.04 all the time for the builder, but it's probably not good to default to an EOL version for everyone. I think we can always adjust this later, though, and we do have some plans to make it easier to use the same toolchain that the Agent uses for this, which could remove the need to specify this altogether.

PR #40221, opened 2025-08-25
Inject non-empty release overrides earlier

rdesgroppes:
### Motivation
So that we don't need to think twice where/whether overridden versions were injected from environment variables.

### DIFF
--- a/tasks/libs/releasing/version.py
+++ b/tasks/libs/releasing/version.py
@@ -371,6 +371,13 @@ def get_version_numeric_only(ctx, major_version='7'):
 
 
 def load_overridden_dependencies(_):
+    """
+    Load dependency versions from release.json with environment variable overrides.
+    WINDOWS_* dependencies are skipped on non-Windows platforms.
+
+    Returns:
+        dict: Environment dictionary with dependency versions as strings.
+    """
     from tasks.libs.releasing.json import load_release_json  # delayed to avoid circular imports
 
     release = load_release_json()
--- a/tasks/libs/releasing/version.py
+++ b/tasks/libs/releasing/version.py
@@ -370,14 +370,23 @@ def get_version_numeric_only(ctx, major_version='7'):
     return version
 
 
-def load_dependencies(_):
-    with open("release.json") as f:
-        versions = json.load(f)
-        if RELEASE_JSON_DEPENDENCIES not in versions:
-            raise Exception(f"Could not find '{RELEASE_JSON_DEPENDENCIES}' in release.json")
-        # windows runners don't accepts anything else than strings in the
-        # environment when running a subprocess.
-        return {str(k): str(v) for k, v in versions[RELEASE_JSON_DEPENDENCIES].items()}
+def load_overridden_dependencies(_):
+    from tasks.libs.releasing.json import load_release_json  # delayed to avoid circular imports
+
+    release = load_release_json()
+    if not (dependencies := release.get(RELEASE_JSON_DEPENDENCIES)):
+        raise Exit(f"Could not find {RELEASE_JSON_DEPENDENCIES!r} in release.json")
+    env = {}
+    for key, value in dependencies.items():
+        if key.startswith("WINDOWS_") and sys.platform != "win32":
+            print(f"Ignoring {key!r} on {sys.platform}", file=sys.stderr)
+            continue
+        if override := os.getenv(key):
+            print(f"Overriding {key!r}: {value!r} -> {override!r}", file=sys.stderr)
+            value = override
+        # windows runners don't accept anything else than strings in the environment when running a subprocess.
+        env[key] = str(value)
+    return env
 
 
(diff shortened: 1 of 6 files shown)


clarkb7 (approved):

clarkb7:
should you also update [_get_release_json_value](https://github.com/DataDog/datadog-agent/blob/e291c293d597a1bae0dbeee1b20a144d7c62e71f/tasks/libs/releasing/json.py#L267) ?
We use it in various e2e tests like
https://github.com/DataDog/datadog-agent/blob/e291c293d597a1bae0dbeee1b20a144d7c62e71f/.gitlab/e2e/e2e.yml#L56-L58

ofek (approved):

sabrina-datadog (approved):

tasks/libs/releasing/version.py
```diff
@@ -370,14 +370,23 @@ def get_version_numeric_only(ctx, major_version='7'):
...
 
-def load_dependencies(_):
-    with open("release.json") as f:
-        versions = json.load(f)
-        if RELEASE_JSON_DEPENDENCIES not in versions:
-            raise Exception(f"Could not find '{RELEASE_JSON_DEPENDENCIES}' in release.json")
-        # windows runners don't accepts anything else than strings in the
-        # environment when running a subprocess.
-        return {str(k): str(v) for k, v in versions[RELEASE_JSON_DEPENDENCIES].items()}
+def load_overridden_dependencies(_):
```
JSGette: Having a docstring is always nice, especially for newcomers

alopezz: Maybe we might take this chance to drop the unused argument, since we're modifying all calling sites anyway.

gjulianm (approved):

rdesgroppes:
> should you also update [_get_release_json_value](https://github.com/DataDog/datadog-agent/blob/e291c293d597a1bae0dbeee1b20a144d7c62e71f/tasks/libs/releasing/json.py#L267) ? We use it in various e2e tests like
> 
> https://github.com/DataDog/datadog-agent/blob/e291c293d597a1bae0dbeee1b20a144d7c62e71f/.gitlab/e2e/e2e.yml#L56-L58

@clarkb7 `_get_release_json_value` happens to be a general purpose helper.
I've tried to add a commit that addresses your request by exposing and using a specific helper for the very need, but it didn't work because the call happens in a Linux job, where `WINDOWS_*` keys are skipped, see:
https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1199320500/viewer#L282

L3n41c (approved):

tasks/libs/releasing/version.py
rdesgroppes: Done through resp. commits.

tasks/libs/releasing/version.py
```diff
@@ -370,14 +370,30 @@ def get_version_numeric_only(ctx, major_version='7'):
...
-        return {str(k): str(v) for k, v in versions[RELEASE_JSON_DEPENDENCIES].items()}
+def load_overridden_dependencies():
+    """
+    Load dependency versions from release.json with environment variable overrides.
+    WINDOWS_* dependencies are skipped on non-Windows platforms.
+
+    Returns:
+        dict: Environment dictionary with dependency versions as strings.
+    """
+    from tasks.libs.releasing.json import load_release_json  # delayed to avoid circular imports
```
alopezz: Perhaps this `load_overridden_dependencies` should be in a different module, given that its purpose is not really all that linked thematically to the rest of functions in this file, and we can take advantage of the fact that we need to update all call sites anyway.

tasks/libs/releasing/version.py
```diff
@@ -370,14 +370,30 @@ def get_version_numeric_only(ctx, major_version='7'):
...
+
+    release = load_release_json()
+    if not (dependencies := release.get(RELEASE_JSON_DEPENDENCIES)):
+        raise Exception(f"Could not find {RELEASE_JSON_DEPENDENCIES!r} in release.json")
+    env = {}
```
alopezz: Nit: I'd suggest more clear/consistent naming here, such as `final_dependencies` or some such (or renaming `dependencies` to`static_dependencies` or something like that). Alternatively the name of the function could be updated to reflect that this returns an "env".

alopezz (approved):
Both comments are optional.

tasks/libs/releasing/version.py
rdesgroppes: :heavy_check_mark: Moved to own module.

tasks/libs/releasing/version.py
rdesgroppes: @alopezz Addressed renaming suggestions as follows:
- variable `dependencies` -> `release_dependencies`
- variable `env` -> `effective_dependencies_env`
- function `load_overridden_dependencies` -> `get_effective_dependencies_env`

PR #40225, opened 2025-08-25
dyninst: ebpf and decoder support for interfaces

ajwerner:
This is the first part of [DEBUG-4043](https://datadoghq.atlassian.net/browse/DEBUG-4043). This just wires up the ebpf and decoding.

There are several followups after to resolve interface names and to augment irgen with the ability to resolve the implementations of interfaces with methods. 

### DIFF
--- a/pkg/dyninst/irgen/irgen.go
+++ b/pkg/dyninst/irgen/irgen.go
@@ -146,7 +146,7 @@ func generateIR(
 		cfg.maxDynamicTypeSize,
 		cfg.maxHashBucketsSize,
 	)
-	pendingSubprograms, err := processDwarf(interests, d, typeCatalog, objFile)
+	pendingSubprograms, goModuledataInfo, err := processDwarf(interests, d, typeCatalog, objFile)
 	if err != nil {
 		return nil, err
 	}
@@ -207,12 +207,13 @@ func generateIR(
 	slices.SortFunc(issues, ir.CompareProbeIDs)
 
 	return &ir.Program{
-		ID:          programID,
-		Subprograms: subprograms,
-		Probes:      probes,
-		Types:       typeCatalog.typesByID,
-		MaxTypeID:   typeCatalog.idAlloc.alloc,
-		Issues:      issues,
+		ID:               programID,
+		Subprograms:      subprograms,
+		Probes:           probes,
+		Types:            typeCatalog.typesByID,
+		MaxTypeID:        typeCatalog.idAlloc.alloc,
+		Issues:           issues,
+		GoModuledataInfo: goModuledataInfo,
 	}, nil
 }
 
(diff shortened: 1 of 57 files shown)


pkg/dyninst/irgen/irgen.go
```diff
@@ -1073,14 +1080,65 @@ func (v *unitChildVisitor) push(
...
+		// pointer to a base type.
+		name, ok, err := maybeGetAttr[string](entry, dwarf.AttrName)
+		if err != nil || !ok {
+			return nil, err
+		}
+		nameWithoutStar := name
+		if entry.Tag == dwarf.TagPointerType {
+			nameWithoutStar = name[1:]
+		}
+		if !onlyLowercaseLettersAndNumbers.MatchString(nameWithoutStar) {
```
piob-io: Let's comment on what this comment is achieving (i.e. we only want to include base/basic types by default...).

pkg/dyninst/loader/loader.go
```diff
@@ -280,8 +280,12 @@ func (l *Loader) loadData(
 	const throttlerMapName = "throttler_params"
 	const throttlerStateMapName = "throttler_buf"
 	const probeParamsMapName = "probe_params"
+	const goRuntimeTypeIDsMapName = "go_runtime_type_ids"
+	const goRuntimeTypesMapName = "go_runtime_types"
+	const notSingleEntry = false
```
piob-io: Maybe swap and rename:
const singleEntryMap = true
const regularMap = false
or
const forceSingleEntry = true
const notForceSingleEntry = false
?
It's all weird, and imperfect, but it's about forcing the array to just be a single entry. And false means we are not forcing it to be like that. It could still be single entry, if there is only one piece of data to store...

pkg/dyninst/loader/serialize.go
```diff
@@ -96,13 +117,36 @@ func serializeProgram(
...
+		typeID := uint64(t.GetID())
+		serialized.typeIDs[i] = typeID
 		serialized.typeInfos[i] = typeInfo{
 			Byte_len:   t.GetByteSize(),
 			Enqueue_pc: metadata.FunctionLoc[compiler.ProcessType{Type: t}],
 		}
+		if goRuntimeType, ok := t.GetGoRuntimeType(); ok {
+			grts.goRuntimeTypes = append(grts.goRuntimeTypes, uint64(goRuntimeType))
+			// If the type is a reference type, then the value underneath the
+			// pointer in the is the pointee of that reference type.
```
piob-io: Some words missing in the comment... or maybe too many words...

piob-io (approved):
Thank you!

pkg/dyninst/loader/serialize.go
ajwerner: Said more things.

pkg/dyninst/loader/loader.go
ajwerner: Done.

pkg/dyninst/irgen/irgen.go
ajwerner: Renamed it and added commentary.

ajwerner (commented):
Thanks for the review!

PR #40229, opened 2025-08-25
[CNM-4617] Improve errors when traceroute module is not enabled.

jmw51798:
### What does this PR do?
Improve error messages when traceroute module is not enabled.

Validated new error messages running agent in docker with traceroute not enabled in /etc/datadog-agent/system-probe.yaml.

1) non-existant or empty system-probe.yaml

2) valid system-probe.yaml, but with traceroute not enabled
```
ping:
  enabled: true
```


pkg/networkpath/traceroute/sysprobe.go
```diff
@@ -32,7 +33,9 @@ func getTraceroute(client *http.Client, clientID string, host string, port uint1
 	req.Header.Set("Accept", "application/json")
 	resp, err := client.Do(req)
 	if err != nil {
-		return nil, err
+		if strings.Contains(err.Error(), "sysprobe.sock: connect: no such file or directory") {
```
pimlu: I would recommend checking against `OpErr`, something like:

```go
    var opErr *net.OpError
    if errors.As(err, &opErr) && opErr.Op == "dial" {
        return nil, fmt.Errorf("failed to dial system-probe, please check that the traceroute module is enabled in the system-probe.yaml config file and that system-probe is running: %w", err)
    }
```

The reason for that is, some platforms such as Windows actually use localhost instead of unix domain sockets, and in rare cases sysprobe.sock can get renamed (there's a config to choose its location IIRC)

jmw51798: Makes sense, changed.  Also replaced the "return nil, err" after this if statement that I inadvertently removed.

pimlu (approved):

PR #40233, opened 2025-08-25
dyninst/decoder: fix memory leak in decoding

ajwerner:
### What does this PR do?

This change adds unit testing with a manually crafted message, and then asserts that we cleared the map.

### Motivation

I noticed that we had data items from previous probe calls when decoding which got me thinking we had a bug.

Testing.

### DIFF
--- a/pkg/dyninst/decode/decoder.go
+++ b/pkg/dyninst/decode/decoder.go
@@ -33,10 +33,10 @@ type Decoder struct {
 	program      *ir.Program
 	decoderTypes map[ir.TypeID]decoderType
 	probeEvents  map[ir.TypeID]probeEvent
+	stackFrames  map[uint64][]symbol.StackFrame
 
 	// These fields are initialized and reset for each message.
 	snapshotMessage   snapshotMessage
-	stackFrames       map[uint64][]symbol.StackFrame
 	dataItems         map[typeAndAddr]output.DataItem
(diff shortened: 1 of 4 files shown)


ajwerner:
Maybe this doesn't need to be backported. It only affects long-running programs and the leak will be pretty slow. Given folks tend not to use us for long-running things, it'll probably be okay.

pkg/dyninst/decode/decoder.go
```diff
@@ -91,15 +90,20 @@ func (d *Decoder) Decode(
...
 	probe, err = d.snapshotMessage.init(d, event, symbolicator)
 	if err != nil {
 		return nil, err
 	}
-	defer d.snapshotMessage.clear()
 	err = json.MarshalWrite(out, &d.snapshotMessage)
 	return probe, err
 }
 
+func (d *Decoder) reset() {
```
piob-io: resetCurrentMessage, resetForNextMessage? Something with messages? Because we are not resetting the whole decoder, we are only cleaning the "	// These fields are initialized and reset for each message." part. Minor point, since it's all private stuff.

piob-io (approved):
Nice catch!

Agreed, no need to backport.

pkg/dyninst/decode/decoder.go
ajwerner: Done.

PR #40236, opened 2025-08-26
dyninst/gotype,decode: parse Go's type info, adopt to resolve type names

ajwerner:
Continuation from #40225. Part of [DEBUG-4043](https://datadoghq.atlassian.net/browse/DEBUG-4043).

This PR primarily is focused on parsing type information from inside a Go binary. It somewhat unfortunately relies on hard-coded layouts of ABI structures. These layouts could be derived from DWARF. At least parts of them could also be validated by checking for internal consistency.

The second commit hooks this up to the decoder to resolve type names for unknown interface types.


pkg/dyninst/decode/decoder.go
```diff
@@ -27,6 +28,34 @@ type probeEvent struct {
 	probe *ir.Probe
 }
 
+// TypeNameResolver resolves type names from type IDs as communicated by the
+// probe regarding types in interfaces.
```
piob-io, line 32: "as represented in go process memory", "as stored inside interface boxes" or just "from go runtime type IDs"?
(probe could be doing whatever it wants, including re-mapping the ids)

pkg/dyninst/gotype/table.go
```diff
@@ -0,0 +1,140 @@
...
+	if rodata == nil {
+		return nil, fmt.Errorf(".rodata section not found")
+	}
+	if moduleData.Types < rodata.Addr {
+		return nil, fmt.Errorf(
+			"moduledata.Types offset is less than rodata address: %#x < %#x",
+			moduleData.Types, rodata.Addr,
+		)
+	}
+	if moduleData.ETypes < moduleData.Types {
```
piob-io, line 64: I understand the prior check, because we want to make sure types are within section we are looking at. But here, I think we should assume that ParseModuleData returns a self-consistent values, and this check is not required (nor any other check that would just compare between things that are all inside the moduleData).

pkg/dyninst/gotype/table.go
```diff
@@ -0,0 +1,140 @@
...
+
+// NewTable constructs a Table from an ELF file.
+func NewTable(obj object.File) (*Table, error) {
+	moduleData, err := object.ParseModuleData(obj)
+	if err != nil {
+		return nil, err
+	}
+
+	var roMap object.SectionData
+	rodata := obj.Section(".rodata")
```
piob-io: This is sometimes not sufficient, due to multiple rodata sections, we need to iterate over possibly many sections that have the same name .rodata. Add a todo, or fix.

piob-io (approved):
Neat!

pkg/dyninst/gotype/table.go
ajwerner: I disagree unless you're going to tell me that `ParseModuleData` ensures these invariants. We're not constructing these executables. I'd rather get this error than some weird error later due to mmapping an incorrect byte range.

pkg/dyninst/gotype/table.go
ajwerner: Done.

pkg/dyninst/decode/decoder.go
ajwerner: Done.

ajwerner (commented):
Thanks for the review!

PR #40240, opened 2025-08-26
[AGENTRUN-669] Fix "Last Execution Date" in GUI check status

Enzu83:
### What does this PR do?

This PR fixes the display of the "Last Execution Date" field in the GUI check status.
The issue was indirectly introduced by a previous PR (https://github.com/DataDog/datadog-agent/pull/39717) that changed the check stats map to use time.Time for its timestamp field. As a result, JSON marshaling automatically converted time.Time values to string dates, which broke the time parsing because it was expecting either an integer or a float.

Units test were also added to prevent future modifications from breaking this.

### Motivation

The GUI check status displayed an error for "Last Execution Date".
The error was `invalid time parameter string` for every check.

GUI check status display and unit tests.

### DIFF
--- a/comp/core/status/render_helpers.go
+++ b/comp/core/status/render_helpers.go
@@ -201,7 +201,7 @@ func parseUnixTime(value any) (time.Time, error) {
 		raw = int64(v)
 	// Case where the unix time is a time.Time and has been converted into a string date due to a JSON marshall
 	case string:
-		t, err := time.Parse(timeLayout, v)
+		t, err := time.Parse(time.RFC3339, v)
 		if err != nil {
 			return time.Time{}, fmt.Errorf("error while parsing time: %s", v)
 		}
--- a/comp/core/status/render_helpers.go
+++ b/comp/core/status/render_helpers.go
@@ -95,6 +95,7 @@ func TextFmap() pkgtexttemplate.FuncMap {
 }
 
 const timeFormat = "2006-01-02 15:04:05.999 MST"
+const timeLayout = "2006-01-02T15:04:05.999999-07:00"
 
 // RenderHTML reads, parse and execute template from embed.FS
 func RenderHTML(templateFS embed.FS, template string, buffer io.Writer, data any) error {
@@ -198,6 +199,14 @@ func parseUnixTime(value any) (time.Time, error) {
 		raw = v
 	case float64:
 		raw = int64(v)
+	// Case where the unix time is a time.Time that has converted into a string date due to a JSON marshall
+	case string:
+		t, err := time.Parse(timeLayout, v)
+		if err != nil {
+			return time.Time{}, fmt.Errorf("error while parsing time %q", v)
+		}
+
+		return t, nil
 	default:
 		return time.Time{}, fmt.Errorf("invalid time parameter %T", v)
 	}
--- a/comp/core/status/render_helpers.go
+++ b/comp/core/status/render_helpers.go
@@ -203,9 +203,8 @@ func parseUnixTime(value any) (time.Time, error) {
 	case string:
 		t, err := time.Parse(timeLayout, v)
 		if err != nil {
-			return time.Time{}, fmt.Errorf("error while parsing time %q", v)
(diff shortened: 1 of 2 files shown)


comp/core/status/render_helpers.go
```diff
@@ -198,6 +199,13 @@ func parseUnixTime(value any) (time.Time, error) {
 		raw = v
 	case float64:
 		raw = int64(v)
+	// Case where the unix time is a time.Time and has been converted into a string date due to a JSON marshall
+	case string:
+		t, err := time.Parse(timeLayout, v)
```
pgimalac: According to the documentation it uses format from RFC3339, so you could just use that ?https://pkg.go.dev/time#Time.MarshalJSON
```suggestion
		t, err := time.Parse(time.RFC3339, v)
```

Note that it's not quite the same format as you entered previously so maybe check that this works as expected first

comp/core/status/render_helpers_test.go
```diff
@@ -28,3 +29,40 @@ func TestMkHuman(t *testing.T) {
...
+		{int64(1756199835000000000), time.Unix(0, 1756199835000000000)}, // nanoseconds
+		{float64(1756201396.123), time.Unix(1756201396, 0)},
+		{"2025-08-26T11:43:16.000000+02:00", time.Unix(1756201396, 0)},
+	}
+
+	for _, tc := range cases {
+		output, err := parseUnixTime(tc.value)
+
+		assert.WithinDuration(t, tc.expectedOutput, output, 0)
+		assert.Nil(t, err)
```
pgimalac: There is a `NoError` helper which has a better output in case of issue, and you might want to only do the second assertion if there was no error (otherwise you'll get two errors in the output, which makes it harder to debug)
```suggestion
		if assert.NoError(t, err) {
			assert.WithinDuration(t, tc.expectedOutput, output, 0)
		}
```

comp/core/status/render_helpers_test.go
```diff
@@ -28,3 +29,40 @@ func TestMkHuman(t *testing.T) {
...
+		{false, "invalid time parameter bool"},
+		{"Tue Aug 26 11:43:16 CEST 2025", "error while parsing time: Tue Aug 26 11:43:16 CEST 2025"},
+	}
+
+	for _, tc := range cases {
+		output, err := parseUnixTime(tc.value)
+
+		assert.WithinDuration(t, time.Time{}, output, 0)
+		assert.NotNil(t, err)
+		assert.Equal(t, err.Error(), tc.expectedErrorMsg)
```
pgimalac: Some other useful helpers:
```suggestion
		assert.EqualError(t, err, tc.expectedErrorMsg)
		assert.Zero(t, output)
```

pgimalac (approved):

comp/core/status/render_helpers.go
Enzu83: RFC3339 format doesn't include microseconds but works perfectly fine, better using it than defining the date format. Thanks!

comp/core/status/render_helpers_test.go
```diff
@@ -28,3 +29,38 @@ func TestMkHuman(t *testing.T) {
...
+		{int64(1756201396), time.Unix(1756201396, 0)},
+		{int64(1756199835000000000), time.Unix(0, 1756199835000000000)}, // nanoseconds
+		{float64(1756201396.123), time.Unix(1756201396, 0)},
+		{"2025-08-26T11:43:16.000000+02:00", time.Unix(1756201396, 0)},
+	}
+
+	for _, tc := range cases {
+		output, err := parseUnixTime(tc.value)
+		if assert.NoError(t, err) {
+			assert.WithinDuration(t, tc.expectedOutput, output, 0)
```
pgimalac, line 47: This should be equivalent I guess
```suggestion
			assert.Equal(t, tc.expectedOutput, output)
```

pgimalac (approved):

dustmop (approved):
LGTM, thanks for the fix!

PR #40274, opened 2025-08-26
chore: disable Admission Controller tests on Windows

wdhif:
### What does this PR do?

Disable Admission Controller unit tests on Windows as this platform is not supported by the Cluster Agent.

### Motivation

This is done to avoid tests flakyness.

Changes are validated by unit tests.

N/A

### Additional Notes

N/A

### DIFF
--- a/pkg/clusteragent/admission/controllers/webhook/controller_base_test.go
+++ b/pkg/clusteragent/admission/controllers/webhook/controller_base_test.go
@@ -3,7 +3,7 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2016-present Datadog, Inc.
 
-//go:build kubeapiserver && !darwin
+//go:build kubeapiserver
 
 package webhook
 
--- a/pkg/clusteragent/admission/controllers/webhook/controller_v1_test.go
+++ b/pkg/clusteragent/admission/controllers/webhook/controller_v1_test.go
@@ -2,14 +2,15 @@
 // under the Apache License Version 2.0.
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2016-present Datadog, Inc.
-// Skip mac os test as cluster agent doesn't run on mac os
-//go:build kubeapiserver && !darwin
+
+//go:build kubeapiserver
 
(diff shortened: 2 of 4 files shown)


pkg/clusteragent/admission/controllers/webhook/controller_base_test.go
```diff
@@ -3,7 +3,8 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2016-present Datadog, Inc.
 
-//go:build kubeapiserver && !darwin
+// Skip macOS and Windows as Cluster Agent doesn't run on those platforms.
+//go:build kubeapiserver && !darwin && !windows
```
davidor: This makes it harder to work on these files locally on a Mac and Windows.
Can we disable this at the GitLab level instead?

wdhif: Another solution would be to use `runtime` and doing something like:
```
	if runtime.GOOS == "darwin" || runtime.GOOS == "windows" {
		t.Skip("Skipping on darwin and windows as cluster agent doesn't run on these OS")
	}
```

Checking with DevX to see about disabling the test directly on GitLab.

wdhif: Checking `GOOS` is also the recommended solution from DevX, I'll do the modification 👍

davidor: That's better because at least it can be compiled, but still the tests cannot be run.

davidor: If we don’t have a way to exclude tests in GitLab, we can keep the `runtime.GOOS` check for now. But I think it’s something worth looking into.

wdhif: Yes, I've made the modification to use `runtime.GOOS` as that was the one recommended by DevX, but I agree that I would prefer to have a solution to use the build tags but have them being disregarded in someway by IDEs for development purposes.

davidor (approved):

PR #40276, opened 2025-08-26
[Serverless-Init] Change Cloud Run Jobs origin tag; refactor logs source

nhulston:
### What does this PR do?

- Renames the origin for cloud run jobs to `cloudrunjobs` (for billing purposes)
- Adds the `origin:cloudrunjobs` tag to be used on the frontend
- Create a `GetDefaultLogsSource()` that will be used when creating the agent log config, because we want Cloud Run Jobs to have `origin:cloudrunjobs` but a logs source value of `cloudrun`
- Also, origin and source are not the same thing. The previous implementation was confusing and seems to imply that the two values are equivalent, when they are actually very different. This makes it clear that we are just falling back to `GetDefaultLogsSource()` when `DD_SOURCE` is not defined
- Although this is technically breaking (since we're changing the tag value), this should not be considered breaking because Cloud Run Jobs is not officially supported or released yet.

### Motivation

- Separate Cloud Run services and Cloud Run Jobs for billing purposes
- Expose the `origin` tag for the frontend

Built serverless-init locally and deployed it to a test Cloud Run Job.

As you can see, the tag is `origin:cloudrunjobs`, and the log source is `cloudrun`:
Screenshot 2025-08-27 at 1 36 51 PM

None

### DIFF
--- a/cmd/serverless-init/cloudservice/service.go
+++ b/cmd/serverless-init/cloudservice/service.go
@@ -61,7 +61,7 @@ func (l *LocalService) GetTags() map[string]string {
 
 // GetDefaultLogsSource is a default implementation that returns an empty logs source
 func (l *LocalService) GetDefaultLogsSource() string {
-	return ""
+	return "unknown"
 }
 
 // GetOrigin is a default implementation that returns a local empty origin
--- a/cmd/serverless-init/cloudservice/service.go
+++ b/cmd/serverless-init/cloudservice/service.go
@@ -17,6 +17,10 @@ type CloudService interface {
 	// the logs, traces, and metrics.
 	GetTags() map[string]string
 
+	// GetDefaultLogsSource returns the value that will be used for the logs source
+	// if `DD_SOURCE` is not set by the user.
+	GetDefaultLogsSource() string
+
 	// GetOrigin returns the value that will be used for the `origin` attribute for
 	// all logs, traces, and metrics.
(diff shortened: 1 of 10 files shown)


nina9753 (dismissed):

jcstorms1:
This will pull cloudrun jobs under cloudrun billing. Do we want to have a new origin for this to differentiate from normal cloudrun? Something like `origin:cloudrunjob`?

cmd/serverless-init/cloudservice/service.go
```diff
@@ -55,6 +59,11 @@ func (l *LocalService) GetTags() map[string]string {
 	return map[string]string{}
 }
 
+// GetDefaultLogsSource is a default implementation that returns an empty logs source
+func (l *LocalService) GetDefaultLogsSource() string {
+	return ""
```
nina9753: Can we maybe make the default log source `serverless`, as the fallback that was the log pipeline might still catch the logs? Also, has this been tested with an empty string value for source, causing any issues?

nhulston: Okay as discussed, I changed it to return "unknown" instead of an empty string

I agree that the default log source should be `serverless` instead of duplicated pipelines for `cloudrun`, `appservices`, etc -- will need some backend changes first, so this should be a future PR

nina9753 (approved):

PR #40283, opened 2025-08-26
[build] Fork rules_multitool to our own extension.

aiuto:
Import the core parts of rules_multitool but with appropriate modifications for our needs.

- Re-root things so this is a local extension, rather than a distinct module.
- remove the ability to use .netrc for authentication.
- remove the :cwd and :workspace_root variations.
  - There are equivalent workarounds using `--run_under="cd <path> &&"`
  - Using `bazel run` is usually not the best practice. Rules should use `$(location <tool>)`. If a tool is so common that people need to call it anywhere, at any time, then it should be in their path.
  - The code is left commented out. Ready to be enabled if a good case is presented.
- Remove the support for WORKSPACE.

Next steps:
- Change download structure so the exeuctable name is the tool, rather than `"executable"`.
- Stop passing the attributes of each binary around as json blobs to be encoded and decoded.
- Make `:path` feature that will print the full execution path.


bazel/multitool/extension.bzl
```diff
@@ -0,0 +1,280 @@
...
+load("//bazel/multitool/private:templates.bzl", "templates")
+
+_DEFAULT_HUB_NAME = "multitool"
+
+def file_extension(os):
+    if os == "windows":
+        return ".exe"
+    return ""
+
+def _download_extract_tool(rctx, tool_name, binary):
```
JSGette:
```suggestion
def _download_and_extract_tool(rctx, tool_name, binary):
```

bazel/multitool/extension.bzl
```diff
@@ -0,0 +1,280 @@
...
+                url = binary["url"],
+            ))
+        rctx.symlink(archive_file, target_executable)
+    else:
+        fail("Unknown 'kind' {kind}".format(kind = binary["kind"]))
+
+    templates.tool_tool(rctx, tool_name, "BUILD.bazel", {"{target_filename}": target_filename})
+
+def _tool_repo_impl(rctx):
+    _download_extract_tool(rctx, rctx.attr.tool_name, json.decode(rctx.attr.binary))
```
JSGette:
```suggestion
    _download_and_extract_tool(rctx, rctx.attr.tool_name, json.decode(rctx.attr.binary))
```

bazel/multitool/extension.bzl
```diff
@@ -0,0 +1,280 @@
...
+        archive_file = "{archive_path}/{file}".format(archive_path = archive_path, file = binary["file"])
+        if not rctx.path(archive_file).exists:
+            fail("{tool_name} ({os}, {cpu}): Cannot find {file} in archive from {url}".format(
+                tool_name = tool_name,
+                os = binary["os"],
+                cpu = binary["cpu"],
+                file = archive_file,
+                url = binary["url"],
+            ))
+        rctx.symlink(archive_file, target_executable)
```
JSGette, line 98: What does it do in case of Windows?

bazel/multitool/private/templates.bzl
```diff
@@ -0,0 +1,105 @@
...
+}
+
+def _render_tool(rctx, filename, substitutions = None):
+    rctx.template(
+        filename,
+        Label(_TOOL_TEMPLATE.format(filename = filename)),
+        substitutions = substitutions or {},
+    )
+
+def _render_tool_tool(rctx, tool_name, filename, substitutions = None):
```
JSGette, line 27: Name of this function is awesome

bazel/multitool/private/lockfile.bzl
```diff
@@ -0,0 +1,60 @@
+"Utilities for interacting with the multitool lockfile."
+
+def _check(condition, message):
+    "fails iff condition is False and emits message"
```
JSGette:
```suggestion
    "fails if condition is False and emits message"
```

bazel/multitool/private/templates.bzl
```diff
@@ -0,0 +1,105 @@
...
+def _render_hub_tool(rctx, tool_name, filename, substitutions = None):
+    rctx.template(
+        "tools/{tool_name}/{filename}".format(tool_name = tool_name, filename = filename),
+        Label(_HUB_TOOL_TEMPLATE.format(filename = filename)),
+        substitutions = {
+            "{name}": tool_name,
+        } | (substitutions or {}),
+    )
+
+def _renter_tool_labels(tools):
```
JSGette:
```suggestion
def _render_tool_labels(tools):
```

bazel/multitool/private/templates.bzl
```diff
@@ -0,0 +1,105 @@
...
+
+    return "\n".join([
+        _render_tool_repo(hub_name, tool_name, binary)
+        for tool_name, tool in tools.items()
+        for binary in tool["binaries"]
+    ])
+
+def _tools_subs(hub_name, tools):
+    return {
+        "{tool_labels}": _renter_tool_labels(tools),
```
JSGette:
```suggestion
        "{tool_labels}": _render_tool_labels(tools),
```

JSGette (approved):
All in all LGTM

bazel/multitool/private/hub_repo_tool_template/BUILD.bazel.template
```diff
@@ -0,0 +1,28 @@
...
+#    name = "cwd",
+#    tool = ":{name}",
+#    visibility = ["//visibility:public"],
+#)
+
+#workspace_root(
+#    name = "workspace_root",
+#    tool = ":{name}",
+#    visibility = ["//visibility:public"],
+#)
```
rdesgroppes, line 29: do we really want to merge commented-out code?

bazel/multitool/extension.bzl
```diff
@@ -0,0 +1,280 @@
+"""
+Multitool takes as input a JSON lockfile and emits the following repos:
```
rdesgroppes, line 2: `<rewrite>`
Sorry for the previous wordy version, which sounded like I was lecturing - here’s a more concise summary of my concerns about vendoring `multitool`:

### Why vendoring hurts me
* Maintenance drift: once copied, we diverge from upstream, and every bug fix, perf improvement, or Bazel bump becomes our burden.
* Security & compliance: miss upstream CVEs, complicate license attribution, and muddy provenance for SBOMs.
* Engineering cost: we spend time maintaining, testing, and documenting a fork instead of upstreaming or delivering product value.
* Community loss: cuts us off from upstream reviews, fixes, and goodwill - and creates morale drag with bespoke workflows.
* Governance risk: exceptions encourage more “just copy it” solutions, leading to repo bloat and inconsistent practices.

### Options I'd favor
1. Consume via `bzlmod` or `git_repository` with pinned versions and integrity checks.
2. Use minimal patches instead of a fork.
3. Contribute fixes upstream to reduce long-term maintenance.
4. Wrap behavior in a thin macro/rule instead of copying code.
5. If necessary, maintain a mirror (usual GitHub fork) to buy time for upstreaming.

### Decision check
Before any vendoring, I would ask: is the change truly "un-upstreamable", can a patch or wrapper solve it, do we have an explicit owner, and are we ready to accept the long-term debt?
`</rewrite>`

rdesgroppes (changes requested):

bazel/multitool/private/hub_repo_tool_template/BUILD.bazel.template
aiuto: I don't have a blanket objection to commented out code. It shows where you might go soon. As long as we know things are in flux.  I often use it for alternates where a developer might want to do something different during testing, and revert the change before commit.

bazel/multitool/extension.bzl
aiuto: We should talk about this rather than debate in code comments.
I do want a fork here, because I want different features than upstream, and I don't intend to track it.
I also want to not take the same dependencies as they do. 
I've lived with vendoring code for over 15 years.  I don't view it as an absolute evil.

bazel/multitool/private/lockfile.bzl
aiuto: done

ofek:
Out of curiosity, have we considered forking in a separate repository or do we have a requirement for committing everything related to Bazel in this repo?

bazel/multitool/extension.bzl
aiuto: Great question. IIUC, the only practical way to use bazel on windows is to enable symlinks.

ofek: If that is true, then this is a hard blocker for local usage of external Windows customers and I should probably bump the work on the Windows developer environment container to maximum priority.

aiuto: Why a hard blocker? There are requirements to build, such has a minimal OS version. Setting up the machine for build can be part of that.

While we are at it.... Case sensitive file systems!  But I am not going to die on that hill.

aiuto: Either way.... symlinks is a general bazel issue, separate from this PR.

aiuto:
> Out of curiosity, have we considered forking in a separate repository or do we have a requirement for committing everything related to Bazel in this repo?

It could be a separate repo, but I don't want to start making new public repos until I absolutely need to.
For now, while in development, while I am likely to make incompatible changes, to work in one repo and then move it out when it becomes stable.

bazel/multitool/extension.bzl
aiuto: I'm going to resolve this because I don't think the discussion applies.
This is more of "stealing an idea" than reusing someone else's project.

rdesgroppes: I obviously disagree. Especially: who's supposed to maintain the "borrowed" code?

aiuto: I don't want rules like that to be maintained by others. This is a piece where we are asserting that binaries we did not build are safe to use. That is the kind of case that justifies checked in code that we review.  The fewer places were we trust repository rules maintained by others, the happier I am.

PR #40286, opened 2025-08-26
dyninst/decode: fix always 0 address

ajwerner:
The address used to be always 0 because of erroneous use of a variable. This is now fixed and tested both in the unit tests and the integration tests.


grantseltzer (approved):
LGTM, thanks!

PR #40287, opened 2025-08-26
Bump `bazel` version to 8.4.0rc1

rdesgroppes:
### Motivation
The Bazel documentation now [explicitly states](https://bazel.build/remote/output-directories):
> If the environment variable `$XDG_CACHE_HOME` is set on either Linux or macOS, the value `${XDG_CACHE_HOME}/bazel` will override the default.

With `bazel` 8.3.1 and earlier versions, this was not the case on macOS, which just got fixed:
- https://github.com/bazelbuild/bazel/issues/16937
- https://github.com/bazelbuild/bazel/issues/25167
- https://github.com/bazelbuild/bazel/pull/26773
- https://github.com/bazelbuild/bazel/releases/tag/8.4.0rc1

The fact it's a release candidate is all but a problem at this stage of our Bazel integration.


JSGette:
I would actually wait until it's final, relying on RCs might cause more issues

rdesgroppes:
> I would actually wait until it's final, relying on RCs might cause more issues

Fair enough - closing.

PR #40292, opened 2025-08-26
dyninst/rcscrape: make debouncer dumber

andreimatei:
This is a minor refactor moving some of the responsibilities of the debouncer to the parent Scraper. Now, the state of the debouncer is simplified such that it only keeps track of "updates", without explicitly caring any longer about procmon updates for processes coming and going. Thus, it no longer needs to have state about tracked processes that have no pending updates -- now it only knows about processes that have pending updates.

The responsibility of dealing with procmon updates is lifted to the scraper, which already had to deal with the procmon.

### DIFF
--- a/pkg/dyninst/rcscrape/debouncer.go
+++ b/pkg/dyninst/rcscrape/debouncer.go
@@ -14,16 +14,16 @@ import (
 
 	"github.com/DataDog/datadog-agent/pkg/dyninst/actuator"
 	"github.com/DataDog/datadog-agent/pkg/dyninst/ir"
-	"github.com/DataDog/datadog-agent/pkg/dyninst/procmon"
 	"github.com/DataDog/datadog-agent/pkg/dyninst/rcjson"
 	"github.com/DataDog/datadog-agent/pkg/util/log"
 )
 
 // debouncer tracks messages from the dd-trace-go callback and coalesces them
-// into a single update for a process. We have to do this because there is
-// no sentinel message when the iteration through the configs begins or ends.
-// Instead, we rely on the idle period to determine when we've seen all the
-// configs for a process.
+// into a single update for a process. We have to do this because old tracer
+// versions didn't have a sentinel method we could probe to figure out when the
+// complete set of remote config configuration has been scraped (newer tracers
+// do have such a method, but we haven't started using it yet). Instead, we rely
+// on an idle period to determine when we've seen all the configs for a process.
 type debouncer struct {
 	idlePeriod time.Duration
 	processes  map[actuator.ProcessID]*debouncerProcess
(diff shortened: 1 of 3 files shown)


pkg/dyninst/rcscrape/debouncer.go
```diff
@@ -37,105 +37,72 @@ func makeDebouncer(idlePeriod time.Duration) debouncer {
...
 }
 
-func (c *debouncer) coalesceInFlight(now time.Time) []ProcessUpdate {
-	var updates []ProcessUpdate
-
+// getUpdates returns the state of processes that have pending updates but
+// have been quiesced for a bit (thereby ending a debounce window). Processes
+// for which an updated state is returned are removed from c.processes.
+func (c *debouncer) getUpdates(now time.Time) map[actuator.ProcessID]accumulatedState {
+	res := make(map[actuator.ProcessID]accumulatedState)
```
ajwerner: just make this a slice and put the process ID into accumulated state. You're not getting anything out of this being a map. If you insist on a map, at least lazily construct it.

pkg/dyninst/rcscrape/scraper.go
```diff
@@ -67,33 +78,118 @@ func NewScraper[A Actuator[T], T ActuatorTenant](
...
 func (s *Scraper) GetUpdates() []ProcessUpdate {
 	s.mu.Lock()
 	defer s.mu.Unlock()
-	return s.mu.debouncer.coalesceInFlight(time.Now())
+	updates := s.mu.debouncer.getUpdates(time.Now())
+	res := make([]ProcessUpdate, 0, len(updates))
+	for procID, u := range updates {
+		p, ok := s.mu.processes[procID]
+		if !ok {
+			log.Warnf("bug: debouncer update for untracked process: %v", procID)
```
ajwerner: I don't think this is definitely a bug because there isn't synchronization between the removal of the probes and the processing of the ringbuf, but, if this keeps happening, then it's a bug. I sort of liked that about the old implementation, but whatever.

pkg/dyninst/rcscrape/scraper.go
```diff
@@ -28,10 +30,18 @@ type Scraper struct {
...
 		sinks     map[ir.ProgramID]*scraperSink
+		processes map[actuator.ProcessID]*trackedProcess
 		debouncer debouncer
 	}
 }
 
+type trackedProcess struct {
+	pu procmon.ProcessUpdate
+	// The runtimeID reported by the tracer inside this process. Used for sanity
+	// checks.
```
ajwerner: I don't understand "Used for sanity checks".

ajwerner (approved):
Seems fine. I get that the debouncer being the primary store of Scraper state rubbed you the wrong way.

pkg/dyninst/rcscrape/debouncer.go
andreimatei: done

pkg/dyninst/rcscrape/scraper.go
andreimatei: discussed offline. Added a comment explaining the race in AddUpdate.

pkg/dyninst/rcscrape/scraper.go
andreimatei: removed; I had not understood the deal correctly

andreimatei (commented):
TFTR

PR #40300, opened 2025-08-27
[CWS] Export event.source to allow snapshot only rules

lebauce:
### What does this PR do?

Add `event.source` - that can either be "runtime" or "snapshot".

### Motivation

This allows to create rules that will only be evaluated on snapshot replay.

### DIFF
--- a/pkg/security/secl/model/model.go
+++ b/pkg/security/secl/model/model.go
@@ -149,6 +149,7 @@ type BaseEvent struct {
 	Service       string         `field:"event.service,handler:ResolveService,opts:skip_ad|gen_getters"` // SECLDoc[event.service] Definition:`Service associated with the event`
 	Hostname      string         `field:"event.hostname,handler:ResolveHostname"`                        // SECLDoc[event.hostname] Definition:`Hostname associated with the event`
 	RuleTags      []string       `field:"event.rule.tags"`                                               // SECLDoc[event.rule.tags] Definition:`Tags associated with the rule that's used to evaluate the event`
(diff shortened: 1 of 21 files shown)


pkg/security/secl/model/model.go
```diff
@@ -149,6 +149,7 @@ type BaseEvent struct {
 	Service       string         `field:"event.service,handler:ResolveService,opts:skip_ad|gen_getters"` // SECLDoc[event.service] Definition:`Service associated with the event`
 	Hostname      string         `field:"event.hostname,handler:ResolveHostname"`                        // SECLDoc[event.hostname] Definition:`Hostname associated with the event`
 	RuleTags      []string       `field:"event.rule.tags"`                                               // SECLDoc[event.rule.tags] Definition:`Tags associated with the rule that's used to evaluate the event`
+	Source        string         `field:"event.source,handler:ResolveSource"`                            // SECLDoc[event.source] Definition:`[Experimental] Source of the event`
```
spikat: can we add something like `can either be "runtime" or "snapshot"` to the doc line, otherwise the produced documentation looks confusing to me.
Like, we already have the `origin` field (which is also kind of bad naming tbh) for ebpf/ebpfless which is already confusing. Maybe we can also make it clear like adding also something like` can be ebpf or ebpfless` ?
I even would not be against renaming it for something like `mode` or `monitoring_mode`; as for source I'm not sure to find a better one, maybe `temporality` ?

rtrieu (approved):

spikat (approved):

PR #40303, opened 2025-08-27
[CWS] Add eBPFLess support for setrlimit and prlimit

theop-dd:
### What does this PR do?
Add eBPFLess for setrlimit and prlimit.

### Motivation
Have setrlimit and prlimit detected on Fargate.

Enabled existing tests for eBPFLess

### DIFF
--- a/pkg/security/probe/model_ebpfless.go
+++ b/pkg/security/probe/model_ebpfless.go
@@ -52,7 +52,6 @@ func NewEBPFLessModel() *model.Model {
 				!strings.HasPrefix(field, "bind.") &&
 				!strings.HasPrefix(field, "connect.") &&
 				!strings.HasPrefix(field, "setrlimit.") &&
-				!strings.HasPrefix(field, "prlimit64.") &&
 				!strings.HasPrefix(field, "setsockopt.") {
 				return rules.ErrEventTypeNotEnabled
 			}
--- a/pkg/security/probe/model_ebpfless.go
+++ b/pkg/security/probe/model_ebpfless.go
@@ -51,6 +51,8 @@ func NewEBPFLessModel() *model.Model {
 				!strings.HasPrefix(field, "accept.") &&
 				!strings.HasPrefix(field, "bind.") &&
 				!strings.HasPrefix(field, "connect.") &&
+				!strings.HasPrefix(field, "setrlimit.") &&
+				!strings.HasPrefix(field, "prlimit64.") &&
 				!strings.HasPrefix(field, "setsockopt.") {
 				return rules.ErrEventTypeNotEnabled
 			}
--- a/pkg/security/probe/probe_ebpfless.go
+++ b/pkg/security/probe/probe_ebpfless.go
@@ -344,19 +344,12 @@ func (p *EBPFLessProbe) handleSyscallMsg(cl *client, syscallMsg *ebpfless.Syscal
 		event.SetSockOpt.FilterLen = syscallMsg.Setsockopt.FilterLen
 		event.SetSockOpt.SizeToRead = uint32(syscallMsg.Setsockopt.FilterLen)
 	case ebpfless.SyscallTypeSetrlimit:
-		if syscallMsg.Setrlimit != nil {
-			event.Type = uint32(model.SetrlimitEventType)
-			event.Setrlimit.Resource = syscallMsg.Setrlimit.Resource
-			event.Setrlimit.RlimCur = syscallMsg.Setrlimit.CurLimit
-			event.Setrlimit.RlimMax = syscallMsg.Setrlimit.MaxLimit
-			event.Setrlimit.TargetPid = syscallMsg.Setrlimit.Pid
-		} else {
-			event.Setrlimit.Resource = 0
-			event.Setrlimit.RlimCur = 0
-			event.Setrlimit.RlimMax = 0
-			event.Setrlimit.TargetPid = 0
+		event.Type = uint32(model.SetrlimitEventType)
+		event.Setrlimit.Resource = syscallMsg.Setrlimit.Resource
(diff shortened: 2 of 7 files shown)


pkg/security/ptracer/process_handlers.go
```diff
@@ -444,6 +457,38 @@ func handleDeleteModule(tracer *Tracer, process *Process, msg *ebpfless.SyscallM
...
+	if err != nil {
+		fmt.Errorf("failed to read rlimit: %w", err)
+		return err
+	}
+	var rlimit syscall.Rlimit
+	if err = binary.Read(bytes.NewBuffer(rlimitBuf), binary.NativeEndian, &rlimit); err != nil {
+		return fmt.Errorf("failed to parse rlimit: %w", err)
+	}
+	pid := tracer.ReadArgUint32(regs, 0)
+	if pid == 0 {
```
paulcacheux: could we maybe extract the common parts out of the if ?

pkg/security/probe/probe_ebpfless.go
```diff
@@ -343,6 +343,29 @@ func (p *EBPFLessProbe) handleSyscallMsg(cl *client, syscallMsg *ebpfless.Syscal
 		event.SetSockOpt.RawFilter = syscallMsg.Setsockopt.Filter
 		event.SetSockOpt.FilterLen = syscallMsg.Setsockopt.FilterLen
 		event.SetSockOpt.SizeToRead = uint32(syscallMsg.Setsockopt.FilterLen)
+	case ebpfless.SyscallTypeSetrlimit:
+		if syscallMsg.Setrlimit != nil {
```
paulcacheux: I don't see this nil check being done in the other cases, is this even possible ?

pkg/security/probe/model_ebpfless.go
```diff
@@ -51,6 +51,8 @@ func NewEBPFLessModel() *model.Model {
 				!strings.HasPrefix(field, "accept.") &&
 				!strings.HasPrefix(field, "bind.") &&
 				!strings.HasPrefix(field, "connect.") &&
+				!strings.HasPrefix(field, "setrlimit.") &&
+				!strings.HasPrefix(field, "prlimit64.") &&
```
paulcacheux: I don't think we have fields starting with `prlimit64`. I believe setrlimit handles both syscalls

pkg/security/ptracer/process_handlers.go
```diff
@@ -444,6 +457,38 @@ func handleDeleteModule(tracer *Tracer, process *Process, msg *ebpfless.SyscallM
 	return nil
 }
 
+func handlePrlimit64(tracer *Tracer, process *Process, msg *ebpfless.SyscallMsg, regs syscall.PtraceRegs, _ bool) error {
+	msg.Type = ebpfless.SyscallTypeSetrlimit
+
+	rlimitBuf, err := tracer.ReadArgData(process.Pid, regs, 2, uint(unsafe.Sizeof(syscall.Rlimit{})))
+	if err != nil {
+		fmt.Errorf("failed to read rlimit: %w", err)
```
paulcacheux: this feels like a bug, maybe return this directly ?

pkg/security/ptracer/process_handlers.go
theop-dd: Yes you are right!

pkg/security/probe/probe_ebpfless.go
theop-dd: yes, I think I put it at the beginning for debug. I agree.

pkg/security/probe/model_ebpfless.go
theop-dd: Yes, it works without it, I've just tested it

pkg/security/probe/probe_ebpfless.go
theop-dd: After a double check, it happens that we have nil values, I'll investigate that

pkg/security/ptracer/process_handlers.go
```diff
@@ -444,6 +457,37 @@ func handleDeleteModule(tracer *Tracer, process *Process, msg *ebpfless.SyscallM
 	return nil
 }
 
+func handlePrlimit64(tracer *Tracer, process *Process, msg *ebpfless.SyscallMsg, regs syscall.PtraceRegs, _ bool) error {
+	rLimitPtr := tracer.ReadArgUint64(regs, 2)
+	if rLimitPtr == 0 {
+		// We don't write a new rlimit so we don't send the event
+		msg.Type = ebpfless.SyscallTypeUnknown
```
spikat: nit: this assignation is not needed (it's the default value)

pkg/security/ptracer/process_handlers.go
```diff
@@ -444,6 +457,37 @@ func handleDeleteModule(tracer *Tracer, process *Process, msg *ebpfless.SyscallM
...
+	}
+	msg.Type = ebpfless.SyscallTypeSetrlimit
+	rlimitBuf, err := tracer.ReadArgData(process.Pid, regs, 2, uint(unsafe.Sizeof(syscall.Rlimit{})))
+	if err != nil {
+		return fmt.Errorf("failed to read rlimit: %w", err)
+	}
+	var rlimit syscall.Rlimit
+	if err = binary.Read(bytes.NewBuffer(rlimitBuf), binary.NativeEndian, &rlimit); err != nil {
+		return fmt.Errorf("failed to parse rlimit: %w", err)
+	}
```
spikat: you must set the msg type only after no errors can be returned, otherwise a broken msg could be sent

theop-dd: Yes you are right thanks!

spikat (approved):

PR #39287, opened 2025-07-28
Add Windows zip package build of ddot

alopezz:
### What does this PR do?

Adds Windows (zip) package build for ddot.

### Motivation

[ABLD-95](https://datadoghq.atlassian.net/browse/ABLD-95)

### Describe how you validated your changes

Manually inspected zip files output by this job: https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1050320323

### Additional Notes

[ABLD-95]: https://datadoghq.atlassian.net/browse/ABLD-95?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #40304, opened 2025-08-27
Bump memory limit on windows zip-building jobs

alopezz:
### What does this PR do?

Bumps the memory limit passed to `docker run` for jobs that are based on the zip building template.

This was probably missed on https://github.com/DataDog/datadog-agent/pull/31489, which bumped such limits in other Windows jobs.

### Motivation

We've been seeing plenty (6-7% job failure rate) of memory errors and timeouts from the windows_zip_ddot_x64 since it was introduced on [this PR](https://github.com/DataDog/datadog-agent/pull/39287). The failures were more or less "hidden" thanks to retries (i.e. they seldom resulted in actual failed pipelines). This job relies on the template that this PR is changing.

Green CI is sufficient. I'll monitor this after merge to make sure the errors are indeed gone.

No significant risk, this is a limit so we shouldn't see an increase in overall memory usage just from this change.

### DIFF
--- a/.gitlab/package_build/windows.yml
+++ b/.gitlab/package_build/windows.yml
@@ -98,7 +98,7 @@ windows_msi_and_bosh_zip_x64-a7-fips:
     - mkdir omnibus\pkg
     - >
       docker run --rm
-      -m 8192M
+      -m 24576M
       -v "$(Get-Location):c:\mnt"
       -e CI_COMMIT_BRANCH=${CI_COMMIT_BRANCH}
       -e CI_PIPELINE_ID=${CI_PIPELINE_ID}


.gitlab/package_build/windows.yml
```diff
@@ -98,7 +98,7 @@ windows_msi_and_bosh_zip_x64-a7-fips:
     - mkdir omnibus\pkg
     - >
       docker run --rm
-      -m 8192M
+      -m 24576M
```
rdesgroppes, line 101: LG, but the fact this case was missed by #31489 calls for a single point of maintenance (new entry to Windows-specific `variables:`?)

rdesgroppes (approved):

rdesgroppes:
```diff
-The failures were more or less "hidden" thanks to retries
+The failures were more or less "hidden" because of retries
```
Job retries are bad, unless they're tailored to transient infrastructure issues. (okay-ish for docker registry, cache server, etc.)

.gitlab/package_build/windows.yml
alopezz: That's a good point. There doesn't seem to be a particularly good reason to not use the same limit everywhere by default. I'll look into this as a follow-up.

PR #40315, opened 2025-08-27
stop services at beginning of setup

clarkb7:
### What does this PR do?
stop services at beginning of setup

### Motivation
services should be stopped before modifying the configuration/install

existing e2e tests
also this code isn't publicly used on Windows yet


pkg/fleet/installer/setup/common/services_nix.go
```diff
@@ -34,3 +35,11 @@ func (s *Setup) restartServices(pkgs []packageWithVersion) error {
 	}
 	return nil
 }
+
+// stopServices stops the services that need to be stopped before running the installer
+func (s *Setup) stopServices(_ context.Context, _ []packageWithVersion) error {
+	// TODO: How to ignore error if service doesn't exist? (initial install case)
+	//       Should we also be re-using packages/service/service.go to support other
+	//       service managers?
+	return nil
```
BaptisteFoy: `packages/service/service.go` is probably the best path forward, though we already stop the agent in its preinst, so no need to do it here.

BaptisteFoy (approved):

PR #40319, opened 2025-08-27
setup default flavor with no args

clarkb7:
### What does this PR do?
setup default flavor with no args

Allows installing Agent with
```
datadog-installer.exe
```
instead of
```
datadog-installer.exe setup --flavor default
```

### Motivation
simpler default setup, will soon replace `Install-Datadog.ps1` PowerShell script

updated e2e test

```powershell
$env:DD_API_KEY="xxx"
$env:DD_SITE="datadoghq.com"
$env:DD_INSTALLER_REGISTRY_URL_AGENT_PACKAGE="installtesting.datad0g.com"
curl.exe -L 'https://s3.amazonaws.com/dd-agent-mstesting/pipelines/A7/74894883/datadog-installer-7.71.0-devel.git.382.d649c13.pipeline.74894883-1-x86_64.exe' -o datadog-installer.exe
.\datadog-installer.exe
```

### Additional notes
we don't want install to trigger if you accidentally double click the .exe file. thankfully we use cobra and it already handles blocking execution via "double click" using https://github.com/inconshreveable/mousetrap


lrbhatia (approved):

cmd/installer/command/command.go
```diff
@@ -104,5 +104,16 @@ Datadog Installer installs datadog-packages based on your commands.`,
...
+	// Run the default command when no subcommands are provided
+	// Intended as shortcut for running the executable as the install script.
+	// default: setup --flavor default
+	agentCmd.RunE = func(cmd *cobra.Command, args []string) error {
+		if len(args) == 0 {
+			cmd.SetArgs([]string{"setup", "--flavor", "default"})
+			return cmd.Execute()
+		}
+		return nil
+	}
```
BaptisteFoy: I don't believe this is idiomatic on Linux, I would expect the help message to come up instead -- should we make this behaviour happen on Windows only?

PR #40323, opened 2025-08-27
[ABLD-133] Stop using OS-specific temp dir for bazel disk cache

rdesgroppes:
### Motivation
On the one hand, `/tmp` is a problematic (ambiguous) path under Windows (`c:/tmp`, `c:\\tmp` and even `\\tmp` would all denote absolute paths, but would be uncommon - let alone wiped out automatically).

On the other hand, having the disk cache outside of a temporary/volatile (e.g., `tmpfs`) directory would allow to persist the cache:
- across reboots on plain developer machines,
- in GitLab CI, where it must be under `$CI_PROJECT_DIR`. (#40153)

In case someone asks: disk and remote caching work well together: `bazel` forms a combined cache, reading from and writing to both, which reduces remote load and speeds up builds, see: https://www.buildbuddy.io/blog/bazels-remote-caching-and-remote-execution-explained/#flags


alopezz (commented):
I think the original setup was meant to cooperate with docker by mounting a volume where the disk cache would be written to, surviving restarts of the docker container:

https://github.com/DataDog/datadog-agent/blob/69836952b0602a4cc44409c457441541c2f72002/bazel/.devcontainer/devcontainer.json#L5-L10

(that was just for context)

> keeping the default disk cache path

What is the default?

rdesgroppes:
> I think the original setup was meant to cooperate with docker by mounting a volume where the disk cache would be written to, surviving restarts of the docker container:

It certainly fits the devcontainer use case, but only for POSIX filesystems.

> What is the default?

None: no `cas` object is persisted unless a `--disk_cache` is specified.
By default, each workspace has a dedicated entry in the bazel cache where it stores its own artifacts. They call it `outputBase` (https://bazel.build/remote/output-directories#layout-diagram).

So we need to find a way that accommodates:
- local developments with devcontainers
- local developments without devcontainers
- GitLab CI where cachable paths have to be under `$CI_PROJECT_DIR`
- Windows, all use cases

alopezz (approved):

alopezz:
I guess that means that this was redundant all along with this thing that we did on the Dockerfile:
https://github.com/DataDog/datadog-agent/blob/69836952b0602a4cc44409c457441541c2f72002/bazel/.devcontainer/Dockerfile#L12-L13

rdesgroppes:
> I guess that means that this was redundant all along with this thing that we did on the Dockerfile:
> 
> https://github.com/DataDog/datadog-agent/blob/69836952b0602a4cc44409c457441541c2f72002/bazel/.devcontainer/Dockerfile#L12-L13

Not exactly:
- `--output_user_root` is a **startup option** that specifies the root `bazel` should obey for its out-of-workspace storage instead of its platform default (that is, unless `XDG_CACHE_HOME` is set: `~/.cache/bazel` on Linux, `/private/var/tmp/bazel` on Darwin, `$USERPROFILE\bazel` on Windows)
- `--disk_cache` is a **build option** that serves both as an _activation flag_ by enabling local caching of `cas` objects when set _while telling where_ to store these.

But they have something in common we need to keep in mind: the locations they designate are context-dependent, which encompasses:
- the OS, in that filesystems have different forms of paths (POSIX vs Windows) and conventions, (e.g., `/tmp` vs `/private/var/tmp` vs `$USERPROFILE\AppData\Local\Temp`)
- the use case: dev cycle on a regular laptop (plain paths), dev cycle in a container (mounted paths), CI execution in a container (mounted paths, with hard constraints for cacheables), CI execution on direct runners (plain paths, constraints for cacheables).

PR #40326, opened 2025-08-27
[AGENTRUN-673] Add mlx5_core network check metrics

jeremy-hanna:
### What does this PR do?

The mlx5_core NIC metrics were added to the [python network check](https://github.com/DataDog/integrations-core/pull/20481) after we had ported the ethtool parsing over so these were not being sent in our Azure staging clusters.

This PR adds support for parsing those stats and comprehensive unit tests for the ethtool metrics in general.

* Thorough unit tests
* Will deploy on staging to confirm


pkg/collector/corechecks/net/networkv2/network.go
```diff
@@ -424,12 +429,15 @@ func getEthtoolMetrics(driverName string, statsMap map[string]uint64) map[string
 				queueTag = "cpu:" + cpuNum
 				newKey = strings.Join(parts[1:], "_")
 				metricPrefix = ".cpu."
+				continueCase = false
```
jeremy-hanna: We never told the parsing to halt when it found a matching queue name format

pkg/collector/corechecks/net/networkv2/network.go
```diff
@@ -456,6 +497,14 @@ func getEthtoolMetrics(driverName string, statsMap map[string]uint64) map[string
 			}
 		}
 		if newKey != "" && queueTag != "" && metricPrefix != "" {
+			if queueTag != "global" {
+				// we already guard against parsing unsupported NICs
+				queueMetrics := ethtoolMetricNames[driverName]
+				// skip queues metrics we don't support for the NIC
+				if !slices.Contains(queueMetrics, newKey) {
```
jeremy-hanna, line 488: we were only filtering [global metrics](https://github.com/DataDog/datadog-agent/pull/40326/files#diff-e88657054ac54654934fe01fd1722599b882419fb419c62b9d4a8c861ffe9dabR492) but not queue metrics, this led to many additional metrics that the python check does not submit

pkg/collector/corechecks/net/networkv2/network.go
```diff
@@ -442,11 +450,44 @@ func getEthtoolMetrics(driverName string, statsMap map[string]uint64) map[string
...
+				//   tx0_bytes -> (queue:0, tx_bytes)
+				//   rx1_packets -> (queue:1, rx_packets)
+				parts := strings.Split(statName, "_")
+				queueIndex := -1
+				queueNum := -1
+				for i, part := range parts {
+					if !strings.HasPrefix(part, "rx") && !strings.HasPrefix(part, "tx") {
+						continue
+					}
+					if cur, err := strconv.Atoi(part[2:]); err == nil {
```
nathan-b: Nit: `cur` doesn't seem like the correct variable name here. Maybe `queue` or `num`?

pkg/collector/corechecks/net/networkv2/network.go
```diff
@@ -442,11 +450,44 @@ func getEthtoolMetrics(driverName string, statsMap map[string]uint64) map[string
...
+						queueIndex = i
+						queueNum = cur
+						break
+					}
+				}
+				if queueIndex == -1 {
+					continueCase = true
+				} else {
+					parts = append(append([]string{}, parts[queueIndex][:2]), parts[queueIndex+1:]...)
```
nathan-b: I don't understand the need for the extra `append` here (I'm not a golang expert so this could be a me issue...). Why is this preferable to 
```
parts = append([]string{parts[queueIndex][:2]}, parts[queueIndex+1:]...)
```
?

nathan-b (approved):
Seems generally reasonable to me, and I love all the extra testing! A couple of comments that you can take or ignore as you see fit :)

pkg/collector/corechecks/net/networkv2/network.go
jeremy-hanna: good call out, my brain went to the [insert slicetrick](https://go.dev/wiki/SliceTricks#insert) but really we can just modify the `part[queueIndex]` element instead of creating a new slice

jeremy-hanna: fixed in https://github.com/DataDog/datadog-agent/pull/40326/commits/f9301b7a8ddafbc920c7a5ed79a33d8368330b55

pkg/collector/corechecks/net/networkv2/network.go
jeremy-hanna: fixed in https://github.com/DataDog/datadog-agent/pull/40326/commits/f9301b7a8ddafbc920c7a5ed79a33d8368330b55

PR #40329, opened 2025-08-27
[ AGNTLOG-264 ] Enable lazy loading for the Logs Agent

ddrthall:
The Logs Agent currently defaults to a disabled state, which makes onboarding new users more difficult. An idle logs agent takes a small but real amount of system resources, and we would prefer to not change the default behavior completely for users who do not want the logs agent but have not explicitly altered the configuration to disable. Logs agent lazy loading is a middle ground - by default when the logs agent is neither enabled nor disabled, it will exist in a partially initiated state. In this state it will use only the bare minimum amount of resources to watch for viable log configs to be provided to the agent - when one is, the entire logs agent will startup and operate normally.

### What does this PR do?
If the logs agent detects that it should be lazily loaded, it will start up its schedulers and initialize the required FX dependencies. Once in that state, there are three pathways to enable the full agent:
1. New sources are added by the schedulers. This is the primary way the agent will start up, as both local configuration and autodiscovery pathways route through this functionality.
2. A new logs integration is detected and registered. The integrations.go file has been modified slightly to support operations performed before the integrations launcher has been started by the logs agent.
3. A pipeline provider is requested (via the otel agent). The agent is started rather than returning a nil pipeline due to the agent's presence in fx components - unless fully disabled, the otel agent may expect that the logs agent can safely provide a new pipeline.

There are additionally three exposed functions that will not trigger an agent startup when called
1. AddScheduler - this function modifies scheduler state, which will always be started even in lazy load status
2. GetSources - the sources entity is fully initiated alongside the agent fx dependencies and should be safe to manipulate even if the agent is not yet started
3. GetMessageReceiver - utilized by the stream logs flow, the message receiver will be nil if the agent is not yet started. Slight modifications have been made to callers in order to fully support this flow, which will correctly cause streamlogs to treat the logs agent as disabled when it is not yet started.

As a result of these changes, SMP regression tests have been modified to better represent the new Logs Agent defaults. The 'quality_gate_idle' test is designed to test the most minimal feature set and so now has the logs agent explicitly disabled rather than implicitly disabled. All tests that had `logs_enabled` set to true have also had that configuration option removed - the expectation is that the new lazy loading feature will provide equivalent functionality for all of these tests.

In addition to testing expansion, the two confguration-based enablement pathways were manually tested in both host and kubernetes deployments. Tests were structured with lazy loading enabled (no logs_enabled configuration field defined), and making sure the agent was disabled at startup but safely enabled and ingested logs once the enablement pathway being tested was called. Sample configurations are provided below.
In addition, stream logs was manually tested in a number of agent states.

#### New Source Additions:
Option 1: static configuration present at agent startup
have a conf.yaml set up in the relevant conf.d folder:
```
logs:
  - type: file
    path: "/Users/username/trial_log.txt"
    source: "test_source"
    service: "test-service"
```

Option 2: utilize annotations in a kubernetes environment:
```
     annotations:
        ad.datadoghq.com/app2.logs: |
          [{
            "source":"test_source",
            "service":"test-service",
          }]    
```

#### Logs Integration
define a custom python check configuration (e.g. conf.d/custom_checks/custom_checks.yaml):
```
init_config:
instances:
  - min_collection_interval: 1

logs:
  - type: integration
    service: custom_checks
    source: custom_checks
```

and a corresponding python file (e.g. checks.d/custom_check.py)
```
from datadog_checks.base.checks import AgentCheck

class HelloCheck(AgentCheck):
  def check(self, instance): 
    log = {"message": "hello world"}
    self.send_log(log)
```
Memory usage will increase in the default configuration on the order of hundreds of kilobytes. Both the fully enabled and the fully disabled configurations will experience no operational change.

Additionally, one of the primary risks for a change of this nature is that an unexpected interaction occurs between another piece of code and an unstarted logs agent that either disables or hangs the log agent startup behavior. A new metric `logs.hung_start` has been added to add visibility to any cases where the logs agent was asked to startup following a lazy initialization but failed/hung in the attempt.

### DIFF
--- a/comp/logs/agent/agentimpl/agent.go
+++ b/comp/logs/agent/agentimpl/agent.go
@@ -63,6 +63,8 @@ const (
 
 	// inventory setting name
 	logsTransport = "logs_transport"
+
+	expectedStartTimeout = 10 * time.Second
 )
 
 // Module defines the fx options for this component.
@@ -208,6 +210,14 @@ func newLogsAgent(deps dependencies) provides {
 
 func (a *logAgent) start(context.Context) error {
 	a.log.Info("Starting logs-agent...")
+
+	go func() {
+		time.Sleep(expectedStartTimeout)
+		if a.started.Load() != status.StatusRunning {
+			a.log.Warn("logs-agent did not start within the expected amount of time")
+			metrics.TlmLogsHungStart.Add(1)
+		}
+	}()
 
 	// setup the server config
 	endpoints, err := buildEndpoints(a.config)
--- a/comp/logs/agent/agentimpl/agent.go
+++ b/comp/logs/agent/agentimpl/agent.go
@@ -127,55 +127,82 @@ type logAgent struct {
 
 	// started is true if the logs agent is running
 	started *atomic.Uint32
+
+	// guaranteeStart is a sync.OnceValue that can be called to guarantee that the logs agent is started
+	guaranteeStart func() error
 }
 
 func newLogsAgent(deps dependencies) provides {
-	if deps.Config.GetBool("logs_enabled") || deps.Config.GetBool("log_enabled") {
-		if deps.Config.GetBool("log_enabled") {
-			deps.Log.Warn(`"log_enabled" is deprecated, use "logs_enabled" instead`)
-		}
+	logsAgentConfigured := deps.Config.IsConfigured("logs_enabled") || deps.Config.IsConfigured("log_enabled")
+	logsAgentEnabled := deps.Config.GetBool("logs_enabled") || deps.Config.GetBool("log_enabled")
 
-		integrationsLogs := integrationsimpl.NewLogsIntegration()
-
-		logsAgent := &logAgent{
-			log:                deps.Log,
-			config:             deps.Config,
-			inventoryAgent:     deps.InventoryAgent,
-			hostname:           deps.Hostname,
-			started:            atomic.NewUint32(status.StatusNotStarted),
-			auditor:            deps.Auditor,
-			sources:            sources.NewLogSources(),
-			services:           service.NewServices(),
-			tracker:            tailers.NewTailerTracker(),
-			flarecontroller:    flareController.NewFlareController(),
-			wmeta:              deps.WMeta,
-			schedulerProviders: deps.SchedulerProviders,
-			integrationsLogs:   integrationsLogs,
-			tagger:             deps.Tagger,
-			compression:        deps.Compression,
-		}
-		deps.Lc.Append(fx.Hook{
-			OnStart: logsAgent.start,
-			OnStop:  logsAgent.stop,
-		})
-
+	if logsAgentConfigured && !logsAgentEnabled {
+		deps.Log.Info("logs-agent disabled")
 		return provides{
-			Comp:           option.New[agent.Component](logsAgent),
+			Comp:           option.None[agent.Component](),
 			StatusProvider: statusComponent.NewInformationProvider(NewStatusProvider()),
-			FlareProvider:  flaretypes.NewProvider(logsAgent.flarecontroller.FillFlare),
-			LogsReciever:   option.New[integrations.Component](integrationsLogs),
-			APIStreamLogs: api.NewAgentEndpointProvider(streamLogsEvents(logsAgent),
-				"/stream-logs",
-				"POST",
-			),
+			LogsReciever:   option.None[integrations.Component](),
 		}
 	}
 
-	deps.Log.Info("logs-agent disabled")
+	if deps.Config.GetBool("log_enabled") {
+		deps.Log.Warn(`"log_enabled" is deprecated, use "logs_enabled" instead`)
+	}
+
+	integrationsLogs := integrationsimpl.NewLogsIntegration(deps.Log, deps.Config)
+	sources := sources.NewLogSources()
+	services := service.NewServices()
+
+	logsAgent := &logAgent{
+		log:                deps.Log,
+		config:             deps.Config,
+		inventoryAgent:     deps.InventoryAgent,
+		hostname:           deps.Hostname,
+		started:            atomic.NewUint32(status.StatusNotStarted),
+		auditor:            deps.Auditor,
+		sources:            sources,
+		services:           services,
+		tracker:            tailers.NewTailerTracker(),
+		flarecontroller:    flareController.NewFlareController(),
+		wmeta:              deps.WMeta,
+		schedulerProviders: deps.SchedulerProviders,
+		integrationsLogs:   integrationsLogs,
+		tagger:             deps.Tagger,
+		compression:        deps.Compression,
+		guaranteeStart:     func() error { return nil },
+	}
+
+	// Logs agent startup behavior is different depending on whether the logs agent is explicitly enabled.
+
+	startHook := logsAgent.start
+
+	if !logsAgentConfigured {
+		startHook = logsAgent.initializeLazyStart
+
+		guaranteeStart := sync.OnceValue(func() error {
+			logsAgent.log.Info("Starting logs-agent via lazy start")
+			return logsAgent.start(context.Background())
+		})
+
+		logsAgent.guaranteeStart = guaranteeStart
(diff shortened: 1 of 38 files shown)


comp/logs/agent/agentimpl/agent.go
```diff
@@ -127,61 +129,96 @@ type logAgent struct {
...
 	}
 }
 
 func (a *logAgent) start(context.Context) error {
 	a.log.Info("Starting logs-agent...")
 
+	go func() {
+		time.Sleep(expectedStartTimeout)
+		if a.started.Load() != status.StatusRunning {
+			a.log.Warn("logs-agent did not start within the expected amount of time")
```
gh123man, line 227: I assume this codepath is for debugging/diagnostics and doesn't have a functional impact on the logs agent startup? 
Just curious - are there any cases you have observed that would cause the logs agent to not start fast enough?

comp/logs/agent/agentimpl/agent.go
```diff
@@ -257,22 +294,35 @@ func (a *logAgent) startPipeline() {
...
-
-		a.log.Info("logs-agent started")
-		a.started.Store(status.StatusRunning)
 	})
 }
 
 func (a *logAgent) stop(context.Context) error {
+	if a.started.Load() == status.StatusNotStarted {
+		a.log.Info("logs-agent never started, stopping schedulers")
+		a.schedulers.Stop()
```
gh123man, line 336: Is it possible for the following order of operations to happen?

1. `a.started.Load() == status.StatusNotStarted` passes
2. A new source is scheduled (on another goroutine) causing the logs agent to lazy start
3. `a.schedulers.Stop()` is called stopping the scheduler 
4. now we have a dangling started logs agent that never stops because we skip the rest of the shutdown logic.

(I haven't fully traced the scheduler code to see if this is possible, but just wanted to check if you considered this case)

comp/logs/integrations/impl/integrations.go
```diff
@@ -3,50 +3,114 @@
...
-		Config:        config,
+	sources, err := ad.CreateSources(cfg)
+	if err != nil {
+		li.log.Errorf("Failed to create source for %q: %v", cfg.Name, err)
+		return
 	}
 
-	li.integrationChan <- integrationConfig
+	for _, source := range sources {
+		// TODO: integrations should only be allowed to have one IntegrationType config.
```
gh123man, line 72: NIT: Maybe we should annotate TODOs like this with a JIRA ticket ID so we don't lose track of it?

comp/logs/integrations/impl/integrations.go
```diff
@@ -3,50 +3,114 @@
...
+	log                log.Component
+	actionCallback     func() error
+	registrationList   map[string]bool
+	integrationTimeout time.Duration
 }
 
 // NewLogsIntegration creates a new integrations instance
-func NewLogsIntegration() *Logsintegration {
+func NewLogsIntegration(log log.Component, config configComponent.Component) integrations.Component {
+	integrationTimeout := time.Duration(config.GetInt("logs_config.integrations_logs_timeout")) * time.Second
```
gh123man: I know the agent config is super inconsistent about this, but I tend to prefer using `time.Duration` ex `GetDuration` for time based configs as it allows better config ergonomics. ex: `10s` or `1m` as config options. Let me know what you think.

comp/logs/integrations/impl/integrations.go
```diff
@@ -3,50 +3,114 @@
...
 	integrationLog := integrations.IntegrationLog{
 		Log:           log,
 		IntegrationID: integrationID,
 	}
 
-	li.logChan <- integrationLog
+	select {
+	case li.logChan <- integrationLog:
+	case <-time.After(li.integrationTimeout):
+		li.log.Warnf("Integration %s timed out sending a log", integrationID)
```
gh123man: Will this case only happen if the logs agent was not started or is not running when we expect it to be? If so maybe we should add some additional context to the log ex `Integration %s timed out sending a log, the logs agent may not be running`. 

I also worry that this log could be excessively spammy if it happens in practice as this will be triggered for every log line.

gh123man (commented):
Mostly just a few questions. 
PR description is excellent 👌

comp/logs/integrations/impl/integrations.go
ddrthall: Realistically this case should only happen in normal operation is if the launcher is falling behind the integrations log load to the point of data loss -  if the launcher/logs agent is not correctly started then registration could not have succeeded and we would exit out of send log early. It is also possible that the launcher itself is experiencing an error of some kind, but that would be unusual. My assumption was that it was better to drop a log than hang the entire check system indefinitely, though the delay on the checks would probably still be an issue. The log emission rate would be one every 10 seconds per hung log source, which definitely could get noisy at scale - we do want a warn (or even an error) log, however, given we have effectively lost data in this case. I can "batch" the warnings as a middle ground.

As for how often we would expect this in practice: the integration logs feature seems pretty firmly designed around the assumption that it will receive relatively little traffic, e.g. it doesn't batch log line writes to disk and its log channel isn't buffered - is that the actual expectation of the system? If we expect to be in danger of hitting this timeout then I can spin up an optimization ticket

comp/logs/integrations/impl/integrations.go
ddrthall: This comment was hoovered in raw from the launcher - is this still a desired feature?

comp/logs/agent/agentimpl/agent.go
ddrthall: Yup, this is from the Drawbacks/tradeoffs section of the PR description: primarily want to get a signal in our internal deploy that we found a race condition/hang that I wasn't able to locate during testing. I don't expect the current code to ever hit this

comp/logs/integrations/impl/integrations.go
gh123man: I think batching/throttling (the log line) and maybe even upgrading this to an error would be a good middle ground. 
It does feel like a "point of no return" scenario, and as you pointed out - we have to break *somewhere* (and ideally not cause everything else to come down with us). 

I mostly wanted to call this out because we have had a few cases of high frequency error logs come from this feature in the past.

comp/logs/integrations/impl/integrations.go
gh123man: Oh my bad. I didn't notice that. I'd defer to @soberpeach

releasenotes/notes/logs-agent-lazy-load-8c76a20c3a1ec6f7.yaml
```diff
@@ -0,0 +1,11 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+enhancements:
+  - |
+    In a default configuration with no `logs_enabled` set, the Logs Agent is now lazy loaded and won't start until the first log is received.
```
buraizu:
```suggestion
    In a default configuration without `logs_enabled` set, the Logs Agent is now lazy-loaded and doesn't start until the first log is received.
```

buraizu (approved):
Approving with some minor updates requested

comp/logs/agent/agentimpl/agent.go
```diff
@@ -127,61 +129,97 @@ type logAgent struct {
...
-			Comp:           option.New[agent.Component](logsAgent),
+			Comp:           option.None[agent.Component](),
 			StatusProvider: statusComponent.NewInformationProvider(NewStatusProvider()),
-			FlareProvider:  flaretypes.NewProvider(logsAgent.flarecontroller.FillFlare),
-			LogsReciever:   option.New[integrations.Component](integrationsLogs),
-			APIStreamLogs: api.NewAgentEndpointProvider(streamLogsEvents(logsAgent),
-				"/stream-logs",
-				"POST",
-			),
+			LogsReciever:   option.None[integrations.Component](),
```
goxberry: LogsReciever -> LogsReceiver

rahulkaukuntla (approved):

PR #40330, opened 2025-08-28
Migrate WorkloadFilter Util Selections

gabedos:
### What does this PR do?

Migrates the workload filtering utility selection methods to be part of the WorkloadFiltering component interface.

### Motivation

Efficiency gains to avoid re-calculating the filter selection. Note this is different from compiling the filter, but rather just selecting the filterIDs to be used in the Evaluation of inclusion/exclusion.

In the profiles below we can see that a significant amount of CPU time is spent on simply _choosing_ which filters to be used. This is because accessing values from the `config.Component` is expensive. Instead we are storing the pre-computed constant selection within the `workloadfilter.Component` and exposing it for clients to use.

image
 
Unit tests to verify that the same selections are being made across each client

Context: https://github.com/DataDog/datadog-agent/pull/39694 to move away from using functions to component methods.


pkg/collector/corechecks/sbom/processor.go
```diff
@@ -99,7 +99,7 @@ func newProcessor(workloadmetaStore workloadmeta.Component, filterStore workload
 		}),
 		workloadmetaStore:     workloadmetaStore,
 		filterStore:           filterStore,
-		selectedFilters:       workloadfilter.GetContainerSBOMFilters(),
+		selectedFilters:       filterStore.GetContainerSBOMFilters(),
```
paulcacheux, line 102: I really don't understand this code, why do we need to keep track of both the filter store and the different filters ? Why can't we simply get the filter at init time and use it during the check run ?

wdhif (approved):

pkg/collector/corechecks/sbom/processor.go
gabedos: Hey @paulcacheux,

I've discussed this with @wdhif and we agree that it would provide a simpler user experience to have the Getter methods like `GetContainerSBOMFilters` return the actual bundle of filters instead of just their IDs that are then used to re-query the filterStore for the final computation.

As we are close to the cutoff we would like to merge this PR in that optimizes the existing behavior to use less resources and will shortly thereafter refactor the interface accordingly.

For SBOM specifically this PR has no performance impact since the returned filterID selection has always been constant ([eg](https://github.com/DataDog/datadog-agent/pull/40330/files#diff-f62859a5b8f72d71a8c5a476a90888675d75bd1f9a9cc3bba5beeb6c13b2f584L96)), however, the SharedMetrics and Autodiscovery filter selections themselves have been dynamic requiring the use of the `Config` Component. ([eg](https://github.com/DataDog/datadog-agent/pull/40330/files#diff-f62859a5b8f72d71a8c5a476a90888675d75bd1f9a9cc3bba5beeb6c13b2f584L60-L87)).

paulcacheux: definitely go for it, my comment was for sure non blocking !

paulcacheux (approved):

sblumenthal (approved):

PR #40335, opened 2025-08-28
[EBPF] complexity: fix retrieval for stacked PRs

gjulianm:
### What does this PR do?

Adds code to fetch the base branch of the PR if it doesn't exist.

### Motivation

Fix comparison for stacked PRs

CI working correctly

### DIFF
--- a/tasks/ebpf.py
+++ b/tasks/ebpf.py
@@ -751,11 +751,18 @@ def generate_complexity_summary_for_pr(
         _tag_complexity_job(ctx, {"result": "skip", "reason": "no_branch_complexity_data"})
         return
 
+    remote = "origin"
+    base_branch_ref = f"{remote}/{base_branch}"
+    is_base_locally_available = ctx.run(f"git rev-parse --verify {base_branch_ref}", warn=True, hide=True)
+    if is_base_locally_available is None or not is_base_locally_available.ok:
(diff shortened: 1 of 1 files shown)


tasks/ebpf.py
```diff
@@ -751,11 +751,18 @@ def _try_delete_github_comment(msg: str):
         _tag_complexity_job(ctx, {"result": "skip", "reason": "no_branch_complexity_data"})
         return
 
+    remote = "origin"
+    base_branch_ref = f"{remote}/{base_branch}"
+    is_base_locally_available = ctx.run(f"git rev-parse --verify {base_branch_ref}", warn=True, hide=True)
+    if is_base_locally_available is None or not is_base_locally_available.ok:
+        print(f"Base branch {base_branch} is not locally available, fetching from origin")
+        ctx.run(f"git fetch {remote} {base_branch}")
```
mbertrone, line 759: We may want to handle the case the fetch command fails. WDYT?

mbertrone (approved):
Looks Good, just a minor nit.

tasks/ebpf.py
gjulianm: In that case the call will throw an exception and the job will fail automatically. I don't think we have extra fallback mechanisms

PR #40337, opened 2025-08-28
feat(Tagger.EnrichTags): detect container ID mismatch

wdhif:
### What does this PR do?
Adds a debug log and telemetry to detect when we have a mismatch in container IDs during tag enrichment.

### Motivation
This is done to help debug potential issues around Origin Detection in cases where we suspect PID reuse.

Changes are validated by unit tests and E2E tests.

N/A

### DIFF
--- a/comp/core/tagger/impl/tagger.go
+++ b/comp/core/tagger/impl/tagger.go
@@ -17,6 +17,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"slices"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


comp/core/tagger/impl/tagger.go
```diff
@@ -479,10 +486,18 @@ func (t *localTagger) EnrichTags(tb tagset.TagsAccumulator, originInfo taggertyp
...
 		if generatedContainerID != "" {
+			containerIDs = append(containerIDs, generatedContainerID)
 			if err := t.AccumulateTagsFor(types.NewEntityID(types.ContainerID, generatedContainerID), cardinality, tb); err != nil {
 				t.log.Tracef("Cannot get tags for entity %s: %s", generatedContainerID, err)
 			}
 		}
+
+		// Check for container ID mismatch
+		if len(containerIDs) > 1 &&
+			slices.ContainsFunc(containerIDs[1:], func(id string) bool { return id != containerIDs[0] }) {
```
tobz, line 497: I think this should be `id == containerIDs[0]`?

wdhif: No we should be good: https://pkg.go.dev/slices#ContainsFunc
If one of the elements satisfies the function (the ID is not equal to the first item in the slice), then we have a diff.

tobz: Sorry, yeah: we're looking for a _mismatch_. Doh!

tobz (approved):

gabedos (approved):
lgtm

PR #40345, opened 2025-08-28
[AGENTRUN-722][#incident-42451] Add scrubber key

louis-cqrl:
### What does this PR do?

This PR aims to add new key to the default scrubbe

### Motivation

Enhance Agent scrubber to scrub the following keys:
- consumer_secret
- token_secret
- consumer_key
- token_id

unit test

### DIFF
--- a/pkg/util/scrubber/default.go
+++ b/pkg/util/scrubber/default.go
@@ -157,7 +157,7 @@ func AddDefaultReplacers(scrubber *Scrubber) {
 		[]string{"consumer_key", "consumer_secret", "token_id", "token_secret"},
 		[]byte(`$1 "********"`),
 	)
-	oauthCredentialsReplacer.LastUpdated = parseVersion("7.70.0") // https://github.com/DataDog/datadog-agent/pull/XXXXX
+	oauthCredentialsReplacer.LastUpdated = parseVersion("7.70.0") // https://github.com/DataDog/datadog-agent/pull/40345
 
 	snmpReplacer := matchYAMLKey(
 		`(community_string|auth[Kk]ey|priv[Kk]ey|community|authentication_key|privacy_key|Authorization|authorization)`,
--- a/pkg/util/scrubber/default.go
+++ b/pkg/util/scrubber/default.go
@@ -151,13 +151,20 @@ func AddDefaultReplacers(scrubber *Scrubber) {
 	)
 	tokenReplacer.LastUpdated = defaultVersion
 
-	// OAuth credentials scrubbers for continuous_ai_netsuite and similar integrations
-	oauthCredentialsReplacer := matchYAMLKey(
-		`(consumer_key|consumer_secret|token_id|token_secret)`,
-		[]string{"consumer_key", "consumer_secret", "token_id", "token_secret"},
+	secretReplacer := matchYAMLKeyPart(
+		`secret`,
+		[]string{"secret", "consumer_secret", "token_secret"},
 		[]byte(`$1 "********"`),
 	)
-	oauthCredentialsReplacer.LastUpdated = parseVersion("7.70.0") // https://github.com/DataDog/datadog-agent/pull/40345
+	secretReplacer.LastUpdated = parseVersion("7.70.0") // https://github.com/DataDog/datadog-agent/pull/40345
+
+	// OAuth credentials scrubbers for continuous_ai_netsuite and similar integrations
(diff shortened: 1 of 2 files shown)


pkg/util/scrubber/default.go
```diff
@@ -150,6 +150,15 @@ func AddDefaultReplacers(scrubber *Scrubber) {
 		[]byte(`$1 "********"`),
 	)
 	tokenReplacer.LastUpdated = defaultVersion
+
+	// OAuth credentials scrubbers for continuous_ai_netsuite and similar integrations
+	oauthCredentialsReplacer := matchYAMLKey(
+		`(consumer_key|consumer_secret|token_id|token_secret)`,
+		[]string{"consumer_key", "consumer_secret", "token_id", "token_secret"},
```
hush-hush: Should we try to catch more things ?

I think anything `*secret*` and `*token*` might be good catch, wdyt ?

nathan-b:
Question: is `qa/no-code-change` an appropriate tag for this PR given that it does involve changing product code? I would expect `qa/done` since you implemented a unit test.

pkg/util/scrubber/default.go
louis-cqrl: fixed, discussion on slack

louis-cqrl:
> Question: is `qa/no-code-change` an appropriate tag for this PR given that it does involve changing product code? I would expect `qa/done` since you implemented a unit test.

good catch @nathan-b

pkg/util/scrubber/default_test.go
```diff
@@ -619,6 +619,60 @@ func TestAuthorization(t *testing.T) {
...
+  consumer_secret: my_consumer_secret
+  token_id: my_token_id
+  token_secret: my_token_secret`,
+		`oauth_config:
+  consumer_key: "********"
+  consumer_secret: "********"
+  token_id: "********"
+  token_secret: "********"`)
+
+	// Test that other keys are not affected
```
nathan-b: This test does not seem to be testing what this comment says it's testing (the comment says it's testing that other keys are not affected, but the test has every key being affected).

pkg/util/scrubber/default.go
```diff
@@ -150,6 +150,22 @@ func AddDefaultReplacers(scrubber *Scrubber) {
 		[]byte(`$1 "********"`),
 	)
 	tokenReplacer.LastUpdated = defaultVersion
+
+	secretReplacer := matchYAMLKeyPart(
+		`secret`,
+		[]string{"secret", "consumer_secret", "token_secret"},
```
nathan-b: I'm definitely not an expert, but based on my reading of the code it's useless to set two hints, one which is a substring of another.

See https://github.com/DataDog/datadog-agent/blob/5278a7b1596f8da984a45d1a703e80c10f3ea30b/pkg/util/scrubber/scrubber.go#L209: the second and third hints are not useful.

Maybe I'm missing something though? Feel free to correct me if my understanding is wrong 😄

hush-hush (approved):

pkg/util/scrubber/default.go
louis-cqrl: I'm entirely sure about what you are asking, when scrubbing txt file we are first looking if one hint is present within the current line, if yes then we apply the replacer if no then the we skip it to the next replacer.

hints aims to be a lookup where as the keys aims to fill up a pattern to be a replacer I would say ?

pkg/util/scrubber/default_test.go
louis-cqrl: yes, just removed the line not very useful

pkg/util/scrubber/default.go
nathan-b: Sorry I wasn't clear. What I don't understand is why you would have two hints, "secret" and "consumer_secret" when if the line contains the text "consumer_secret" it will match "secret" first and so we'll never get to "consumer_secret"?

louis-cqrl: I see now, yes you are right, I wanted a track on the keyword we want to follow directly, but yes let's put all the key we want to follow in the PR description, and remove them from the code, does it sounds good to you ?

nathan-b: Makes sense to me, thanks!

louis-cqrl: done

nathan-b (approved):
LGTM, thanks for answering my noob questions :)

PR #40348, opened 2025-08-28
[AGENTCFG-385] Fixing the secrets section of the status command

rahulkaukuntla:
### What does this PR do?

The malformed section is caused by the Secret's StatusProvider having an incomplete implementation. Other StatusProvider's have a Text() and HTML() method that render each type of output, but for the Secrets component they both render text.

As such, to fix, I added an actual HTML template implementation, and then changed the getDebugOutput() method to return a map which could be passed into the appropriate template to provide the correct rendering.

On top of unit tests, I manually checked that secrets section looked good:
Screenshot 2025-08-28 at 5 44 01 PM
Screenshot 2025-08-28 at 4 28 29 PM

However, we should confirm this via QA.


comp/core/secrets/impl/info_nix.go
```diff
@@ -8,44 +8,33 @@
...
-var permissionsDetailsTemplate string
-
-type permissionsDetails struct {
-	FileMode string
-	Owner    string
-	Group    string
-}
-
-func (r *secretResolver) getExecutablePermissions() (interface{}, error) {
+func (r *secretResolver) getExecutablePermissions() (map[string]string, error) {
```
dustmop: I don't think it's good to have the return value be a map with arbitrary fields, better to keep strong typing if we can. To avoid the problem of having different structs declared in os-specific source files, how about creating a file `permissions_details.go` and defining the struct in there. The fields should be the union of fields from the *_nix and *_windows versions, along with `isWindows`.

comp/core/secrets/impl/info_nix_test.go
```diff
@@ -63,16 +63,15 @@ func TestGetExecutablePermissionsSuccess(t *testing.T) {
...
-	assert.Equal(t, currentGroup, details.Group)
+	assert.Equal(t, "100700", res["FileMode"])
+	assert.Equal(t, currentUser, res["Owner"])
+	assert.Equal(t, currentGroup, res["Group"])
 }
 
 func TestDebugInfo(t *testing.T) {
 	tel := nooptelemetry.GetCompatComponent()
 	resolver := newEnabledSecretResolver(tel)
+	secretStatus := secretsStatus{resolver: resolver}
```
dustmop: Could you move this closer to where it is needed?

comp/core/secrets/impl/info_nix_test.go
```diff
@@ -87,39 +86,52 @@ func TestDebugInfo(t *testing.T) {
...
+	assert.Contains(t, output, "Owner: "+currentUser)
+	assert.Contains(t, output, "Group: "+currentGroup)
+	assert.Contains(t, output, "Number of secrets resolved: 3")
+	assert.Contains(t, output, "'secret_refresh_interval' is disabled")
 }
 
 func TestDebugInfoError(t *testing.T) {
 	tel := nooptelemetry.GetCompatComponent()
 	resolver := newEnabledSecretResolver(tel)
+	secretStatus := secretsStatus{resolver: resolver}
```
dustmop: Same here, please keep it near where it gets used.

comp/core/secrets/impl/secrets.go
```diff
@@ -647,58 +640,69 @@ type handlePlace struct {
...
-		fmt.Fprintf(w, "error parsing secret info template: %s\n", err)
-		return
-	}
+	stats["enabled"] = true
 
-	t, err = t.Parse(permissionsDetailsTemplate)
-	if err != nil {
-		fmt.Fprintf(w, "error parsing secret permissions details template: %s\n", err)
-		return
+	if r.backendCommand == "" {
```
dustmop, line 662: Not sure how this was working before, but if `backendCommand` is an empty string, we shouldn't return early. The customer have be using the SGC via the `backendType` setting.

comp/core/secrets/impl/secrets.go
```diff
@@ -145,16 +143,19 @@ func NewComponent(deps Requires) Provides {
...
+	}
 	fb.AddFile("secrets.log", buffer.Bytes())
 	fb.CopyFile(r.auditFilename)
 	return nil
 }
 
 func (r *secretResolver) writeDebugInfo(w http.ResponseWriter, _ *http.Request) {
-	r.getDebugInfo(w)
+	status := secretsStatus{resolver: r}
+	_ = status.Text(true, w)
```
dustmop: Why is this passing `true` instead of `false` like the other cases? What does this parameter do?

Also, the error value from this call should get checked. If a future developer messes up the template, `RenderText` will return an error and that error should get written to the `ResponseWriter`, preferably with an error status http code.

comp/core/secrets/impl/secrets.go
```diff
@@ -711,19 +715,16 @@ func (r *secretResolver) getDebugInfo(w io.Writer) {
 		for _, context := range contexts {
 			details = append(details, []string{context.origin, strings.Join(context.path, "/")})
```
dustmop, line 719: Definitely not needed for this PR, but a great fix in the future would be replacing this nested string slice with a meaningful struct instead.

comp/core/secrets/impl/info_nix.go
rahulkaukuntla: Good idea, I've implemented this in [24ce338](https://github.com/DataDog/datadog-agent/pull/40348/commits/24ce3381e7c3dcf2d4db4f4dcc6675bf66b5af67).

comp/core/secrets/impl/info_nix_test.go
rahulkaukuntla: Done.

comp/core/secrets/impl/info_nix_test.go
rahulkaukuntla: Done.

comp/core/secrets/impl/secrets.go
rahulkaukuntla: [The behavior for this](https://github.com/DataDog/datadog-agent/pull/40348/files#diff-752d5ac62b49d4afa2ea92ea4f59d4648a7f14edf35f6cf3fd6ea01e2eb90744L662-L664) has not been changed at all in this PR. It should be fine, as if `backendType` [is used](https://github.com/DataDog/datadog-agent/blob/main/comp/core/secrets/impl/secrets.go#L226), then `backendCommand` is set to the value of the embedded executable.

comp/core/secrets/impl/secrets.go
rahulkaukuntla: `secretsStatus` is an implementation of the [Provider interface](https://github.com/DataDog/datadog-agent/blob/0e6496c33c9c58875aafc29b7cc33139d210ba48/comp/core/status/component.go#L36-L45) in the secrets component. I have changed it to be `false` for consistency, but we don't use it all in the [Text method](https://github.com/DataDog/datadog-agent/pull/40348/files#diff-6b728079f6ec66f77985926aedc609e8cb21edeb4f2850ad176db677de50be09L82), so it shouldn't matter what the value is.

comp/core/secrets/impl/secrets.go
rahulkaukuntla: Nice, I'll keep that in mind for a future PR.

comp/core/secrets/impl/secrets.go
dustmop: Ah I missed that we assign to `backendCommand`, makes sense!

comp/core/secrets/impl/secrets.go
dustmop: I see, thanks for explaining, that sounds good.

dustmop (approved):
Thanks for the fixes, LGTM!

PR #40352, opened 2025-08-28
refactored collectors in gpu core-check

val06:

### What does this PR do?

consolidates some of the collectors into 2 main templates:

stateless collector
- just get 1..N metrics from an NVML API and return them
- no need to make samples and perform computations betweek them

sample collector
- use lastTimeStamp to store state between the runs

### Motivation

- simplify code base
- reduce the number of created collectors in runtime

there are a few collectors that weren't migrated as they don't fit any of the patterns:
- ebpf collector (based on SP metrics)
- fields collector (has a special state to determine how to trigger the nvml API
- gpm collector


pkg/collector/corechecks/gpu/nvidia/base.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/hashicorp/go-multierror"
+)
+
+// apiCallInfo represents a single NVML API call that can produce one or more metrics.
+// It supports both stateless collectors (ignore timestamp) and sampling collectors (use timestamp).
+type apiCallInfo struct {
+	Name     string                                                // Name of the API call for logging/debugging
+	TestFunc func(ddnvml.Device) error                             // Function to test if the API is supported on the device
+	CallFunc func(ddnvml.Device, uint64) ([]Metric, uint64, error) // Function to call the API and return metrics
+}
```
gjulianm, line 21: Can we describe the arguments and return values in this comment?

pkg/collector/corechecks/gpu/nvidia/sample.go
```diff
@@ -0,0 +1,207 @@
...
+					sumSmUtil = 100
+				}
+				deviceSmActive := float64(maxSmUtil+sumSmUtil) / 2.0
+
+				allMetrics = append(allMetrics,
+					Metric{Name: "sm_active", Value: deviceSmActive, Type: ddmetrics.GaugeType},
+					Metric{Name: "core.limit", Value: float64(device.GetDeviceInfo().CoreCount), Type: ddmetrics.GaugeType, Tags: allPidTags},
+				)
+
+				return allMetrics, currentTime, err
```
gjulianm: Maybe this could be moved to a separate function (like `processSamples` is) so that the code is not as nested

pkg/collector/corechecks/gpu/nvidia/base.go
```diff
@@ -0,0 +1,91 @@
...
+	"fmt"
+	ddnvml "github.com/DataDog/datadog-agent/pkg/gpu/safenvml"
+	"github.com/hashicorp/go-multierror"
+)
+
+// apiCallInfo represents a single NVML API call that can produce one or more metrics.
+// It supports both stateless collectors (ignore timestamp) and sampling collectors (use timestamp).
+type apiCallInfo struct {
+	Name     string                                                // Name of the API call for logging/debugging
+	TestFunc func(ddnvml.Device) error                             // Function to test if the API is supported on the device
```
gjulianm: Is this `TestFunc` necessary? From what I've seen, we probably could test based on `CallFunc` alone and we'd remove a lot of code.

val06: i think it is, the CallFunc can be heavy in some cases, or even require different input for execution (like providing non-zero timestamp for `GetProcessUtilization` case in the sample collector

gjulianm: But we only test for the specific unsupported/function not found error no? And the check for support gets called only once, so in terms of computation we are not spending that much. Using the same function would also have the advantage of ensuring that we are testing exactly the same function we are using, no room for mistakes if we change something.

gjulianm (approved):

PR #40355, opened 2025-08-28
fix(kafka): Use available bitnami images

chouetz:
### What does this PR do?
Use the `bitnami` images from the [bitnamilegacy](https://hub.docker.com/u/bitnamilegacy) registry.

### Motivation
Agent6 pipeline kmt jobs started to [fail](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1103480000).
```
Error: GET https://index.docker.io/v2/bitnami/kafka/manifests/3.4: MANIFEST_UNKNOWN: manifest unknown; unknown tag=3.4
```
On the `bitnami` [docker page](https://hub.docker.com/r/bitnami/kafka) they alert about progressive deprecation of their images from 28AUG25.
IIUC, only the `latest` tag will be available soon
```
Bitnami will begin deprecating support for non-hardened, Debian-based software images in its free tier and will gradually remove tags from the public catalog. As a result, community users will have access to a reduced number of hardened images. These images are published only under the “latest” tag and are intended for development purposes
```
As we want to keep using the same version we move to their copy on the [bitnamilegacy](https://hub.docker.com/u/bitnamilegacy) registry

CI should be green


guyarb (changes requested):
We shouldn't use latest images, but to pin versions to ensure tests do not break after a release of zookeeper and kafka

guyarb:
waiting on
https://github.com/DataDog/ami-builder/pull/238

guyarb (approved):

PR #40358, opened 2025-08-28
EKS test for container now use AL2023

KevinFairise2:

### What does this PR do?

Move containers test to new AL2023 nodes. It should remove some flakiness due to a bug in the AL2 AMI

Note: The cluster name is explicitly set in the Agent configuration, because the agent is not able to retrieve it properly with the new AL2023 nodes.
The cluster name detection is made by doing a call to EC2 API and it is failing with
```
EC2: GetNetworkID failed to get mac addresses: could not fetch token from IMDSv2: Put "http://169.254.169.254/latest/api/token": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```

### DIFF
--- a/test/new-e2e/pkg/provisioners/aws/kubernetes/eks.go
+++ b/test/new-e2e/pkg/provisioners/aws/kubernetes/eks.go
@@ -131,6 +131,7 @@ func EKSRunFunc(ctx *pulumi.Context, env *environments.Kubernetes, params *Provi
 			params.agentOptions = append(params.agentOptions, kubernetesagentparams.WithDeployWindows())
 		}
 
+		params.agentOptions = append(params.agentOptions, kubernetesagentparams.WithClusterName(cluster.ClusterName))
 		kubernetesAgent, err = helm.NewKubernetesAgent(&awsEnv, "eks", cluster.KubeProvider, params.agentOptions...)
 		if err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


test/new-e2e/pkg/provisioners/aws/kubernetes/eks.go
```diff
@@ -131,6 +131,7 @@ func EKSRunFunc(ctx *pulumi.Context, env *environments.Kubernetes, params *Provi
 			params.agentOptions = append(params.agentOptions, kubernetesagentparams.WithDeployWindows())
 		}
 
+		params.agentOptions = append(params.agentOptions, kubernetesagentparams.WithClusterName(cluster.ClusterName))
```
L3n41c, line 134: On `kind`, we already validate that the agent is able to use the cluster name provided in its config.
That’s why on EKS, we prefer to validate that the agent is properly able to figure the cluster name on its own without any specific config.

```suggestion
```

L3n41c (changes requested):
The agent is normally able to automatically discover the k8s cluster name.
So, putting it explicitly in the config would reduce the test coverage.

test/new-e2e/pkg/provisioners/aws/kubernetes/eks.go
sblumenthal: For context, [it looks like AL2023 explicitly sets the default hop limit to 1](https://docs.aws.amazon.com/linux/al2023/ug/imdsv2.html), which seems to break our ability to hit that api endpoint. I couldn’t figure out how to update it to 2 via pulumi in the couple of hours I spent playing around with it

Even still, short term this PR is preferable to what we currently have (a SIGNIFICANT number of jobs failing at startup / before the tests can run)

L3n41c (approved):

PR #40359, opened 2025-08-28
[EBPF] gpu: fix concurrent access to stream maps

gjulianm:
### What does this PR do?

Adds a RW mutex to the stream collection, as there were some errors in production due to concurrent access.

### Motivation

Avoid errors.

https://datadoghq.atlassian.net/browse/NWL-1576

Existing unit tests

### DIFF
--- a/pkg/gpu/stream_collection.go
+++ b/pkg/gpu/stream_collection.go
@@ -9,7 +9,7 @@ package gpu
 
 import (
 	"fmt"
-	"iter"
+	"sync"
 
 	"golang.org/x/sys/unix"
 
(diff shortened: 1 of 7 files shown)


pkg/gpu/stream_collection.go
```diff
@@ -40,11 +40,13 @@ type globalStreamKey struct {
 }
 
 type streamCollection struct {
-	streams       map[streamKey]*StreamHandler
-	globalStreams map[globalStreamKey]*StreamHandler
-	sysCtx        *systemContext
-	telemetry     *streamTelemetry
-	streamConfig  config.StreamConfig
+	streams            map[streamKey]*StreamHandler
+	streamsMutex       sync.RWMutex // streamsMutex protects concurrent read+write access to the streams map, not the streams themselves
```
val06: perhaps using [sync.Map](https://pkg.go.dev/sync#Map) could be more beneficial for the access pattern?

val06 (commented):
left comment

pkg/gpu/stream_collection.go
gjulianm: You're right, I thought sync.Map was generally slower but for this it seems better. I'll refactor.

val06 (approved):

PR #40361, opened 2025-08-28
[CIEXE-1103] Migrate datadog-agent arm jobs to dedicated runners only

sjain313:

### DIFF
--- a/.gitlab/binary_build/cluster_agent.yml
+++ b/.gitlab/binary_build/cluster_agent.yml
@@ -37,7 +37,7 @@ cluster_agent-build_amd64:
 cluster_agent-build_arm64:
   extends: .cluster_agent-build_common
   image: registry.ddbuild.io/ci/datadog-agent-buildimages/linux$CI_IMAGE_LINUX_SUFFIX:$CI_IMAGE_LINUX
-  tags: ["arch:arm64"]
+  tags: ["arch:arm64", "specific:true"]
 
 cluster_agent_fips-build_amd64:
   extends: cluster_agent-build_amd64
--- a/.gitlab/binary_build/cws_instrumentation.yml
+++ b/.gitlab/binary_build/cws_instrumentation.yml
@@ -28,7 +28,7 @@ cws_instrumentation-build_arm64:
     - !reference [.except_mergequeue]
     - when: on_success
   image: registry.ddbuild.io/ci/datadog-agent-buildimages/linux$CI_IMAGE_LINUX_SUFFIX:$CI_IMAGE_LINUX
-  tags: ["arch:arm64"]
+  tags: ["arch:arm64", "specific:true"]
   needs: ["go_mod_tidy_check", "go_deps"]
   variables:
     ARCH: arm64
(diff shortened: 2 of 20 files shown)


sjain313:
* **#40375** <a href="https://app.graphite.dev/github/pr/DataDog/datadog-agent/40375?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#40361** <a href="https://app.graphite.dev/github/pr/DataDog/datadog-agent/40361?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/datadog-agent/40361?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rdesgroppes (approved):
LG, thanks!

chagui (approved):
Eventually we should probably disable the instance runners for project that have their own shards.

.gitlab/binary_build/cluster_agent.yml
```diff
@@ -37,7 +37,7 @@ cluster_agent-build_amd64:
 cluster_agent-build_arm64:
   extends: .cluster_agent-build_common
   image: registry.ddbuild.io/ci/datadog-agent-buildimages/linux$CI_IMAGE_LINUX_SUFFIX:$CI_IMAGE_LINUX
-  tags: ["arch:arm64"]
+  tags: ["arch:arm64", "specific:true"]
```
KevinFairise2, line 40: Looks good to me.
QQ: When it is deployed on all the runner is there a way so that we automatically pick our dedicated runners with only `arch:arm64` tag ? Will we always need to also add `specific:true` ? We'd need to make sure we never forget it if that's the case

KevinFairise2 (approved):

ogaca-dd (approved):

paulcacheux (approved):

.gitlab/binary_build/cluster_agent.yml
sjain313: Hello
Once we disable the shared runners for this repo(only "arch:arm64" runner tags), any new test would need to have ["arch:arm64", "specific:true"]`otherwise they wouldn't find any runners and tests would fail. So I think it would be hard to miss the tags. I think this what @chagui meant by his suggestion above.

If we both disable the instance runners and have dedicated runners with only "arch:arm64" tag, we can get what you are asking but I would suggest to have both `["arch:arm64", "specific:true"]`  for your dedicated runner tags so its clear for code readability that we are using dedicated runners. Its a pattern that I see we have for other dedicated runners too.

FlorentClarret (approved):

sblumenthal (approved):

jorgetomtz (approved):

duncanista (approved):

BaptisteFoy (approved):

alexn-dd (approved):

gjulianm (approved):

rdesgroppes:
:clap: One of the most approved PR I've seen at DD!

PR #40364, opened 2025-08-28
CPU Management Tag

justin-lesko:
### What does this PR do?
Adds a tag `kube_requested_cpu_management` to kubernetes CPU metrics that indicates whether a container uses dedicated core(s). This tag has two potential values `none` or `static`.

In order to be `static` the following needs to be true:
1. CPU Requests == CPU Limits
2. CPU Requests/Limits are an integer i.e. a whole core or multiple cores
3. Kubelet configuration field `cpu-manager-policy` is `static`

**NOTE:** to compute this tag, this PR also adds the unparsed kubelet config into workloadmeta.
### Motivation
Add a new tag to kubernetes CPU metrics

I deployed a minikube cluster with the kubelet args overriden to set the appropriate `cpu-manager-policy` field 
```
minikube start -p contint-4586-static 
  --driver=docker \
  --container-runtime=containerd \                                                                                                                                                                                                                                                                   
  --cpus=4 --memory=4096 \
  --extra-config=kubelet.cpu-manager-policy=static \
  --extra-config=kubelet.reserved-cpus=0
```

Onto my test cluster, I deployed a pod with with the appropriate CPU requests/limits to warrant the `static` tag
```
apiVersion: v1
kind: Pod
metadata:
  name: guaranteed-qos-pod
spec:
  containers:
  - name: stress
    image: busybox
    command: ["sh","-c","sleep 3600"]
    resources:
      requests:
        cpu: "1"
        memory: "64Mi"
      limits:
        cpu: "1"
        memory: "64Mi"
```

Onto my test cluster I deployed a custom Datadog Agent build of my branch and observed this tag in the UI:
<img width="992" height="1107" alt="Screenshot 2025-08-29 at 1 39 34 PM" src="https://github.com/user-attachments/assets/4ac318a0-317e-4c11-9a28-6deb44396d82" />



releasenotes/notes/kube_cpu_management_tag-300164f56deac3af.yaml
```diff
@@ -0,0 +1,12 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+enhancements:
+  - |
+    Add a new tag kube_cpu_management on kubernetes CPU metrics
```
estherk15:
```suggestion
    Add a new tag `kube_cpu_management` on kubernetes CPU metrics 
```

estherk15 (approved):

comp/core/workloadmeta/collectors/internal/kubelet/kubelet.go
```diff
@@ -115,13 +122,45 @@ func (c *collector) Pull(ctx context.Context) error {
...
+
 func (c *collector) pullFromKubelet(ctx context.Context) error {
+	events := []workloadmeta.CollectorEvent{}
+
+	// Pull the kubelet config once at the first pull because its interval is long
+	c.kubeletConfigOnce.Do(func() {
+		configEvent, err := c.pullKubeletConfig(ctx)
+		if err == nil {
+			events = append(events, configEvent)
+		}
```
lavigne958: as we pull the config at init, should we set the `c.kubeletConfigLastExpire` as well ? so we don't do it twice in a row ?

comp/core/workloadmeta/def/types.go
```diff
@@ -1122,6 +1153,107 @@ func (m *KubernetesMetadata) String(verbose bool) string {
...
+// Kubelet is an Entity representing the kubelet, right now it only holds
+// the kubelet configuration
+type Kubelet struct {
+	EntityID
+	EntityMeta
+	Config KubeletConfig
+}
+
+// GetID implements Entity#GetID
+func (ku *Kubelet) GetID() EntityID { return ku.EntityID }
```
lavigne958: ni picking, puting this function on a single line makes it look like some type definition.

Can we put it on multi-line so it looks like all other functions/method ? Unless the linter complains about it of course :-|

pkg/collector/corechecks/containers/kubelet/provider/cadvisor/provider.go
```diff
@@ -167,6 +167,26 @@ func (p *Provider) processContainerMetric(metricType, metricName string, metricF
...
+
+			wmetaKubelet, _ := p.store.GetKubelet()
+			if wmetaKubelet != nil {
+				cpuManagerPolicy, _ := wmetaKubelet.GetCPUManagerPolicy()
+
+				// strip the leading `container_id://` because containerID is a formatted entity ID
+				id := strings.ReplaceAll(containerID, "container_id://", "")
+				container, _ := p.store.GetContainer(id)
+
+				var guaranteedWholeCore bool
```
lavigne958: should we set it to `false` explicitly just to make it obvious to read ? even though it `false` by default ?

lavigne958 (commented):
I noticed in `comp/core/workloadmeta/collectors/internal/kubelet/kubelet.go` we use the  value of `kubeletConfigLastExpire` to check if we should update the config.

its defautl value is 0, it's not set so anything when we start, so every time we start the call to `time.Since(c.kubeletConfigLastExpire)` will always return the time since EPOCH, which will always be > `kubeletConfigExpireFreq`.

I suggest we rely on this to pull the config, which will always trigger on the first run, and we can remove the use of `sync.Once`.

what do you think ?

comp/core/workloadmeta/collectors/internal/kubelet/kubelet.go
```diff
@@ -655,6 +703,14 @@ func extractResources(spec *kubelet.ContainerSpec) workloadmeta.ContainerResourc
 		resources.MemoryLimit = kubernetes.FormatMemoryRequests(memoryLimit)
 	}
 
+	if cpuLimit, found := spec.Resources.Limits[kubelet.ResourceCPU]; found {
+		if cpuReq, found := spec.Resources.Requests[kubelet.ResourceCPU]; found {
+			if cpuReq == cpuLimit && cpuReq.MilliValue()%1000 == 0 {
+				resources.GuaranteedWholeCore = pointer.Ptr(true)
```
vboulineau: This check is not accurate. You should rely instead on the `Pod status.qosClass == Guaranteed` (we're already parsing it in WLM).
I'm not sure we need to actually store something vs computing it at runtime.

comp/core/workloadmeta/def/types.go
```diff
@@ -1122,6 +1153,107 @@ func (m *KubernetesMetadata) String(verbose bool) string {
 
 var _ Entity = &KubernetesMetadata{}
 
+// KubeletConfig is the kubelet configuration, it is stored as a map
+// and not as a declarative struct
+type KubeletConfig map[string]interface{}
```
vboulineau: We should instead recreate a `struct` with only the fields we are interested in, it will save memory space at runtime (potentially make the JSON parsing faster with JSON parser v2, but probably irrelevant for something we poll every 20m)

pkg/collector/corechecks/containers/kubelet/provider/cadvisor/provider.go
```diff
@@ -167,6 +167,26 @@ func (p *Provider) processContainerMetric(metricType, metricName string, metricF
 		} else {
 			cID, _ := kubelet.KubeContainerIDToTaggerEntityID(containerID)
 			tags, _ = p.tagger.Tag(cID, types.HighCardinality)
+
+			wmetaKubelet, _ := p.store.GetKubelet()
+			if wmetaKubelet != nil {
+				cpuManagerPolicy, _ := wmetaKubelet.GetCPUManagerPolicy()
+
+				// strip the leading `container_id://` because containerID is a formatted entity ID
+				id := strings.ReplaceAll(containerID, "container_id://", "")
```
vboulineau: nit: Ideally we extract containerID and then build tagger ID

pkg/collector/corechecks/containers/kubelet/provider/cadvisor/provider.go
```diff
@@ -167,6 +167,26 @@ func (p *Provider) processContainerMetric(metricType, metricName string, metricF
...
+				container, _ := p.store.GetContainer(id)
+
+				var guaranteedWholeCore bool
+				if container.Resources.GuaranteedWholeCore != nil {
+					guaranteedWholeCore = bool(*container.Resources.GuaranteedWholeCore)
```
vboulineau: nit: `bool()` is not needed. `GuaranteedWholeCore` is already a `bool`

pkg/collector/corechecks/containers/kubelet/provider/cadvisor/provider.go
```diff
@@ -167,6 +167,26 @@ func (p *Provider) processContainerMetric(metricType, metricName string, metricF
...
+				}
+
+				if guaranteedWholeCore && cpuManagerPolicy == workloadmeta.CPUManagerPolicyStatic {
+					tags = utils.ConcatenateStringTags(tags, "kube_cpu_management:static")
```
vboulineau: I don't think the tag name is actually accurate. `kube_cpu_management` would make sense at node level (i.e. this node is configured with `static` CPU), but for a given container, I give the important information is the CPU limit type (is it none, CFS or CPUSet).

vboulineau (commented):
Wondering if we can make it more generic by retrieving the CPUSet for each container and then determine if there's an overlap between CPUSets used by container or not.
It will reflect the _actual_ state rather than the requested state (i.e. a `Burstable` workload can temporarily be running with dedicated cores if it's the only `Burstable` one on the machine and all others are guaranteed + integer)

estherk15 (approved):

lavigne958 (approved):

pkg/collector/corechecks/containers/kubelet/provider/pod/provider.go
```diff
@@ -147,6 +151,24 @@ func (p *Provider) generateContainerSpecMetrics(sender sender.Sender, pod *kubel
...
+				requestedWholeCores = *wmetaContainer.Resources.RequestedWholeCores
+			}
+			if pod.Status.QOSClass == "Guaranteed" &&
+				requestedWholeCores &&
+				cpuManagerPolicy == "static" {
+				tagList = utils.ConcatenateStringTags(tagList, tags.KubeRequestedCPUManagement+":static")
+			} else {
+				tagList = utils.ConcatenateStringTags(tagList, tags.KubeRequestedCPUManagement+":none")
+			}
+		}
```
sblumenthal: Does it make sense to pull this out from here as well as `pkg/collector/corechecks/containers/kubelet/provider/cadvisor/provider.go` and instead put it in `pkg/collector/corechecks/containers/kubelet/common/pod.go`? Seems like these 2 code blocks are identical

sblumenthal (approved):
One small note, otherwise looks good to me

comp/core/tagger/tags/tags.go
```diff
@@ -105,6 +105,9 @@ const (
 	KubeAppManagedBy = "kube_app_managed_by"
 	// KubeAutoscalerKind is the tag reflecting if a pod is managed by an Autoscaler
 	KubeAutoscalerKind = "kube_autoscaler_kind"
+	// KubeRequestedCPUManagement is the tag that describes a container's cpu management
+	// which is either static (dedicated) or none
+	KubeRequestedCPUManagement = "kube_requested_cpu_management"
```
gabedos, line 110: Just double checking that you mean for it to be `kube_requested_cpu_management`. You have it referenced as only `kube_cpu_management` everywhere else.

I personally like the shorter naming convention and not sure how much value the `requested` sub-string adds but slightly indifferent on this

gabedos: Synced offline: adding `requested` does add some significance.

pkg/collector/corechecks/containers/kubelet/provider/pod/provider.go
```diff
@@ -147,6 +151,24 @@ func (p *Provider) generateContainerSpecMetrics(sender sender.Sender, pod *kubel
 	tagList = utils.ConcatenateTags(tagList, p.config.Tags)
 
 	if container.Resources != nil { // Ephemeral containers do not have resources defined
+		wmetaKubelet, _ := p.store.GetKubelet()
+		if wmetaKubelet != nil {
+			cpuManagerPolicy := wmetaKubelet.ConfigDocument.KubeletConfig.CPUManagerPolicy
+			wmetaContainer, _ := p.store.GetContainer(containerID.GetID())
```
gabedos: nit: for extra safety to avoid any seg faults, we can add error handling here.

For eg. if only the kubelet collector is providing container information which loads every ~5 seconds, in a race, it might be possible that the check executes before workloadmeta processes the pods containers.

gabedos (approved):
lgtm

PR #40325, opened 2025-08-27
pass `/norestart` to msiexec

clarkb7:
### What does this PR do?
Pass `/norestart` to msiexec commands to prevent Windows from automatically rebooting the host

https://learn.microsoft.com/en-us/windows/win32/msi/standard-installer-command-line-options#norestart
https://learn.microsoft.com/en-us/windows/win32/msi/reboot

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1683

### Describe how you validated your changes (if not by through tests)
e2e tests will make sure there are no regressions, but a follow up PR will be required to gracefully handle the 3010 exit code

### Additional notes
see flags and exit codes in our [chocolatey installer](https://github.com/DataDog/datadog-agent/blob/c7a047e7155fede3903ecd4a1a1cb4092fe7fa2b/chocolatey/datadog-agent/tools/chocolateyinstall.ps1#L13)

PR #40370, opened 2025-08-28
handle non-zero success exit codes

clarkb7:
### What does this PR do?
Remote Agent updates on Windows now consider MSI exit codes 3010 and 1641 to indicate success

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1683
follow up to https://github.com/DataDog/datadog-agent/pull/40325

### DIFF
--- a/pkg/fleet/installer/msi/msiexec.go
+++ b/pkg/fleet/installer/msi/msiexec.go
@@ -347,6 +347,8 @@ func (m *Msiexec) processLogFile(logFile fs.File) ([]byte, error) {
 }
 
 // isRetryableExitCode returns true if the exit code indicates the msiexec operation should be retried
+//
+// https://learn.microsoft.com/en-us/windows/win32/msi/error-codes
 func isRetryableExitCode(err error) bool {
 	if err == nil {
 		return false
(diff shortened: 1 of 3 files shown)


hestonhoffman (approved):

pkg/fleet/installer/msi/msiexec.go
```diff
@@ -367,6 +369,30 @@ func isRetryableExitCode(err error) bool {
...
+	}
+
+	var exitError exitCodeError
+	if errors.As(err, &exitError) {
+		if exitError.ExitCode() == int(windows.ERROR_SUCCESS_REBOOT_REQUIRED) {
+			// 3010 - success but requires reboot
+			return true
+		} else if exitError.ExitCode() == int(windows.ERROR_SUCCESS_REBOOT_INITIATED) {
+			// 1641 - success but Windows will reboot the host
+			// this is unexpected now that we pass /norestart, msiexec should return 3010 instead
```
BaptisteFoy, line 388: Shouldn't we error on it then ..?

clarkb7: it's better to treat it as success here, as the Agent has either been successfully uninstalled or successfully installed, and callers expect that returning an error means the opposite of that.

If we want to try to handle the sudden reboot better, then we can return an error in this case to signal to the callers. But since using `/norestart` means we're not expecting a sudden reboot anymore it shouldn't be an issue.

We'll still have the error reported by the `msiexec` span, as noted below, so we can track occurances.

BaptisteFoy (approved):

PR #40374, opened 2025-08-28
[log] Fix memory leaking

XSAM:
### What does this PR do?

Resolves #40373

Use the example in #40373

No possible drawbacks, as this is the behavior in [datadogexporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/586b518c73c9a10f74e0faabfcd33e79d7458f8a/exporter/datadogexporter/factory.go#L73-L75), which runs on every single otel collector contrib.

### DIFF
--- a/pkg/util/log/log.go
+++ b/pkg/util/log/log.go
@@ -915,3 +915,8 @@ func CriticalStackDepth(depth int, v ...interface{}) error {
 		return logger.criticalStackDepth(s, depth)
 	}, true, v...)
 }
+
+// Initializes the logger to no-op by default
+func init() {
+	SetupLogger(Disabled(), "off")
+}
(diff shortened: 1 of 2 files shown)


pgimalac:
Hey @XSAM, thanks for opening a PR, we buffer logs on purpose as the datadog-agent can log before the logger is properly initialized, so just disabling it initially would lose those logs (which might not be an issue for `datadogexporter` though).

I opened a [PR](https://github.com/DataDog/datadog-agent/pull/40376) to simply remove the dependency of `pkg/obfuscate` to `pkg/util/log`, I think that should fix your issue, I'm discussing with owners of `pkg/obfuscate` to make sure the change is acceptable

PR #40376, opened 2025-08-29
Do not import log package in pkg/obfuscate

pgimalac:
### What does this PR do?
Remove the import to pkg/util/log in pkg/obfuscate.

There were only two uses in the package, they are updated to use a logger injected from the configuration.

### Motivation
Fix https://github.com/DataDog/datadog-agent/issues/40373.

CI

### Additional details
The only API change for the package is adding a `FullLogger` field in the configuration struct, which takes precedence over the previous `Logger` field when set (if we extended the `Logger` interface then it would have been a breaking change).


XSAM (approved):
Thanks @pgimalac for creating this PR. It looks great!

ichinaski (approved):

PR #40383, opened 2025-08-29
chore: [fa]add integration status to agent_check payload

san-jos:
### What does this PR do?
Add integration status to agent_check payload

>  Example: 
[
            "jmx",
            "jmx",
            "jmx:sample_jvm",
            "ERROR",
            "Unable to instantiate or initialize instance sample_jvm. Is the target JMX Server or JVM running? Failed to retrieve RMIServer stub: javax.naming.ServiceUnavailableException [Root exception is java.rmi.ConnectException: Connection refused to host: 127.0.0.1; nested exception is: \n\tjava.net.ConnectException: Connection refused]",
			"",
 			["service:sample-jvm","env:dev"]
],

### Motivation
Generate integration metrics with integration tags

Manual QA


comp/collector/collector/collectorimpl/agent_check_metadata.go
```diff
@@ -172,3 +204,77 @@ func (c *collectorImpl) collectMetadata(ctx context.Context) time.Duration {
...
+
+// collectTags is a fast-path extractor assuming `tags` is at the root of the YAML.
+// It decodes only the root-level `tags` field and returns as soon as it's found.
+func collectTags(config string) ([]string, error) {
+	if config == "" {
+		return nil, nil
+	}
+
+	dec := yaml.NewDecoder(strings.NewReader(config))
+	for {
```
coignetp: **🥜 nitpick:** Would it be possible to put the break condition here?‏

coignetp (approved):

vickenty (approved):

comp/collector/collector/collectorimpl/agent_check_metadata.go
pgimalac, line 1: We pretty much can't modify this payload, the format dates back from Agent 5 and should not change

If you want this information in REDAPL then you should check out https://github.com/DataDog/datadog-agent/tree/main/comp/metadata/inventorychecks instead

pgimalac (changes requested):

comp/collector/collector/collectorimpl/agent_check_metadata.go
```diff
@@ -144,14 +146,41 @@ func (c *collectorImpl) GetPayload(ctx context.Context) *Payload {
...
+
+	return cr
+}
+
+// GetPayload builds a payload of all the agentchecks metadata
+func (c *collectorImpl) GetPayload(ctx context.Context) *Payload {
+	hostnameData, _ := c.hostname.Get(ctx)
+
+	meta := hostMetadataUtils.GetMetaFromCache(ctx, c.config, c.hostname)
+	meta.Hostname = hostnameData
```
hush-hush: `GetMetaFromCache` doesn't return a copy but the value from the cache at the moment, we should not update it or make a copy of it somewhere.

Also, the hostname should already be set in the return struct, why override it ?

comp/metadata/inventorychecks/inventorychecksimpl/inventorychecks.go
```diff
@@ -52,11 +52,12 @@ type checksMetadata map[string][]metadata
...
-	Hostname     string                `json:"hostname"`
-	Timestamp    int64                 `json:"timestamp"`
-	Metadata     map[string][]metadata `json:"check_metadata"`
-	LogsMetadata map[string][]metadata `json:"logs_metadata"`
-	UUID         string                `json:"uuid"`
+	Hostname     string                   `json:"hostname"`
+	Timestamp    int64                    `json:"timestamp"`
+	Metadata     map[string][]metadata    `json:"check_metadata"`
+	LogsMetadata map[string][]metadata    `json:"logs_metadata"`
+	CheckResults []map[string]interface{} `json:"check_results"`
```
hush-hush: We should embed the result of each checks within `check_metadata` instead of having 2 different maps about the same thing.

Could you also update the README of the module which we use as documentation ?

hush-hush (changes requested):

comp/collector/collector/collectorimpl/agent_check_metadata.go
san-jos: I did not touch this part but I can check it out.

releasenotes/notes/inventory-checks-status-306a9aebe91b85ff.yaml
```diff
@@ -0,0 +1,11 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+enhancements:
+  - |
+    Agent integration inventory checks payload now include the checks statuses.
```
rtrieu:
```suggestion
    Agent integration inventory checks payload now includes the status of all checks.
```

rtrieu (approved):
Minor suggestion to comply with grammar rules.

hush-hush (approved):

comp/collector/collector/collectorimpl/agent_check_metadata.go
pgimalac, line 1: I think this would only get the tags which are explicitly in the instance config, but I don't think it will contains tags added by autodiscovery on the check (eg. container name, etc)

pgimalac (approved):
LGTM !

comp/collector/collector/collectorimpl/agent_check_metadata.go
pgimalac: (I'm not quite sure where those tags are added by autodiscovery)

PR #40385, opened 2025-08-29
[CWS] improve classification of icmp packet

safchain:
### What does this PR do?

Improve the classification of ICMP packet by parsing the ICMP header and report the type of ICMP packet in the events.


pkg/security/secl/model/model.go
```diff
@@ -101,6 +101,7 @@ type NetworkContext struct {
 	Destination      IPPortContext `field:"destination"`       // destination of the network packet
 	NetworkDirection uint32        `field:"network_direction"` // SECLDoc[network_direction] Definition:`Network direction of the network packet` Constants:`Network directions`
 	Size             uint32        `field:"size"`              // SECLDoc[size] Definition:`Size in bytes of the network packet`
+	Type             uint32        `field:"type"`              // SECLDoc[type] Definition:`Type of the network packet` Constants:`Network Protocol Types`
```
spikat, line 104: As it's only related to ICMP packets, should we put is as `omitempty` ?

pkg/security/ebpf/c/include/helpers/network/stats.h
```diff
@@ -113,6 +113,11 @@ __attribute__((always_inline)) void flush_pid_network_stats(u32 pid, void *ctx,
 }
 
 __attribute__((always_inline)) void count_pkt(struct __sk_buff *skb, struct packet_t *pkt) {
+    // only count TCP & UDP packets for now
+    if (pkt->ns_flow.flow.l4_protocol != IPPROTO_TCP && pkt->ns_flow.flow.l4_protocol != IPPROTO_UDP) {
+        return;
+    }
```
spikat, line 119: Not sure why we don't count icmp packets here. It may be interesting to detect data leak using icmp tunnels I guess

pkg/security/secl/model/model.go
safchain: For now yes only for ICMP. omitemty is for JSON marshal things not for SECL. I have put omitempty on the serializer side.

pkg/security/ebpf/c/include/helpers/network/stats.h
safchain: It is not supported by the network flow feature user space side. It generates errors if I let the packet being processed here. I'll address the network flow use case in a follow-up PR.

spikat (approved):

PR #40392, opened 2025-08-29
[CWS-5004] Add overlayfs path operator override

YoannGh:
### What does this PR do?

Add a new operator override that looks at whether a file is access through an overlay mount point when evaluating rules.

To do so this PR allows setting multiple operator overrides on the same field by chaining each operator using a logical OR.

One notable drawback of this solution is that some fields (e.g. `exec.file.path`) might end up being evaluated multiple times against the exact same value if none of the overrides apply to the event (no symlink, no overlayfs).     

### Motivation

The goal is to trigger rules such as `open.file.path == "/tmp/foo"` when the `/tmp/foo` container file is access through an overlay mount point (e.g. `touch /var/lib/docker/overlay2/0123456789abcdef/merged/tmp/foo`).


pkg/security/tests/overlayfs_test.go
theop-dd, line 1: Would adding another test with a different rule—e.g., `mkdir.file.path` or another syscall—be useful, or would it be redundant?

pkg/security/secl/model/oo_overlayfs_unix.go
```diff
@@ -0,0 +1,106 @@
...
+
+				path := ev.FieldHandlers.ResolveFilePath(ev, fileEvent)
+				fs := ev.FieldHandlers.ResolveFileFilesystem(ev, fileEvent)
+				if fs == "overlay" {
+					mountPoint := strings.TrimRight(fileEvent.MountPath, "/")
+					return strings.TrimPrefix(path, mountPoint)
+				}
+
+				// return the original path if not overlayfs
+				// this will cause the same path to be evaluated twice because of the OverlayFSPathname OR override
```
theop-dd: Just for my understanding, this will be only evaluated twice if the path does not match right? 
I assume that if the first path matches the Or operator will do an early return right?

safchain (approved):

pkg/security/tests/overlayfs_test.go
YoannGh: Good idea, I added a new test with 2cdb33511e82ead2f48e9b89d2a625dffa646dc7

pkg/security/secl/model/oo_overlayfs_unix.go
YoannGh: That's indeed only when the path doesn't match, I updated the comment to make that more explicit

theop-dd (approved):

PR #40394, opened 2025-08-29
Update and prevent `MODULE.bazel.lock` file from diverging again

rdesgroppes:
### What does this PR do?
1. Run a quick `bazel` command that fails if the `MODULE.bazel.lock` file gets out-of-sync with actually used dependencies,
2. Update `MODULE.bazel.lock` thanks to `bazel mod deps` as advised in the [error message](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1241267622#L40):
```sh
ERROR: The module extension '@@gcc_toolchain+//:internal.bzl%non_bazel_dependencies' does not exist in the lockfile
ERROR: The module extension '@@rules_fuzzing+//fuzzing/private:extensions.bzl%non_module_dependencies' does not exist in the lockfile
ERROR: The module extension '@@pybind11_bazel+//:python_configure.bzl%extension' does not exist in the lockfile
<root> (@_)
ERROR: Results may be incomplete as 3 extensions failed.
Running after script...
$ if [ $CI_JOB_STATUS = failed ]; then
💡 bazel mod deps
```
### Motivation
Faced differences unrelated to my change when invoking `bazel build //...` locally.
### Additional Notes
Currently, there's a problem with the `gcc_toolchain` repo:
```sh
$ bazel mod deps
ERROR: [redacted]/gcc_toolchain+/internal.bzl:43:17: Traceback (most recent call last):
	File "[redacted]/gcc_toolchain+/internal.bzl", line 117, column 9, in _non_bazel_dependencies_ext_impl
		_avl()
	File "[redacted]/gcc_toolchain+/internal.bzl", line 43, column 17, in _avl
		http_archive(
Error in repository_rule: no repository visible as '@' in the extension '@@gcc_toolchain+//:internal.bzl%non_bazel_dependencies', but referenced by label '@//:examples/avl/avl.BUILD.bazel' in attribute 'build_file' of http_archive 'avl'.
ERROR: error evaluating module extension @@gcc_toolchain+//:internal.bzl%non_bazel_dependencies
```
Hence the additional patch, pending:
- f0rmiga/gcc-toolchain#211

### DIFF
--- a/MODULE.bazel.lock
+++ b/MODULE.bazel.lock
@@ -348,6 +348,93 @@
         ]
       }
     },
+    "@@gcc_toolchain+//:internal.bzl%non_bazel_dependencies": {
+      "general": {
+        "bzlTransitiveDigest": "6NYXupGFeHdYYEnMvtPmKCZCQ8BnIUs/VUgIGhwmfhA=",
+        "usagesDigest": "sWe6gvmFpdWalNUJUWNZF23cgclfU+ijKXCf9JEbTjs=",
+        "recordedFileInputs": {},
+        "recordedDirentsInputs": {},
+        "envVariables": {},
+        "generatedRepoSpecs": {
+          "openssl": {
+            "repoRuleId": "@@bazel_tools//tools/build_defs/repo:http.bzl%http_archive",
+            "attributes": {
+              "build_file_content": "filegroup(\n    name = \"srcs\",\n    srcs = glob(\n        include = [\"**\"],\n        exclude = [\"**/* *\"],\n    ),\n    visibility = [\"//visibility:public\"],\n)\n",
+              "sha256": "40dceb51a4f6a5275bde0e6bf20ef4b91bfc32ed57c0552e2e8e15463372b17a",
+              "strip_prefix": "openssl-1.1.1n",
+              "url": "https://www.openssl.org/source/openssl-1.1.1n.tar.gz"
+            }
+          },
+          "lapack": {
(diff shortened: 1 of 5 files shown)


MODULE.bazel.lock
```diff
@@ -840,6 +840,2748 @@
           ]
         ]
       }
+    },
```
JSGette, line 956: Why do we have python references here if we didn't bring python to main yet?

rdesgroppes: TBH, I don't know.

aiuto: rules_pkg depends on python. And now our own tools to doo.

aiuto:
Do we really need this?
ISTM that every PR that changes a dep is now going to fail in CI first, then they will have to run the deps thing.
We will annoy people quickly if that happens.
I would much rather see CI success, and our own daily pipeline fail if they are out of sync.

rdesgroppes:
> Do we really need this?

Yes. See **Motivation**. The annoyance has been reported frequently enough.

> ISTM that every PR that changes a dep is now going to fail in CI first, then they will have to run the deps thing.

Exactly.

> We will annoy people quickly if that happens.

Nothing more, nothing less than for other lock files.

> I would much rather see CI success, and our own daily pipeline fail if they are out of sync.

~I'm not sure what you mean by own daily pipeline.~
**Update**: we agreed offline to turn it into a precommit hook once `bazel` becomes developer-facing.

JSGette (approved):

PR #40397, opened 2025-08-29
Move stringified json unmarshalling into UnmarshalKey

dustmop:
### What does this PR do?

This PR handles a repeated pattern used by certain config settings. A few different callsites allow the setting's value to be either structured data retrieved from yaml, or a json stringified version of the same data. Instead of making callsites handle this on their own, move the functionality into UnmarshalKey.

### Motivation

Reduce repetition, allow bug fixes to happen in only one place. These stringified values often occur when environment variables are used, and we currently don't have a better solution, such as automatically converting them at config build time.

Unit tests.


ddrthall (approved):

comp/logs/agent/config/config.go
```diff
@@ -78,16 +77,7 @@ var (
...
 	var rules []*ProcessingRule
-	var err error
-	raw := coreConfig.Get("logs_config.processing_rules")
-	if raw == nil {
-		return rules, nil
-	}
-	if s, ok := raw.(string); ok && s != "" {
-		err = json.Unmarshal([]byte(s), &rules)
-	} else {
-		err = structure.UnmarshalKey(coreConfig, "logs_config.processing_rules", &rules, structure.ConvertEmptyStringToNil)
```
hush-hush, line 89: We don't need `ConvertEmptyStringToNil` anymore ?

hush-hush (approved):
Added a question, feel free to merge after answering !

pkg/config/structure/unmarshal.go
```diff
@@ -15,6 +15,7 @@ import (
 	"unicode"
 	"unicode/utf8"
 
+	json "github.com/json-iterator/go"
```
paulcacheux: Why not use the stdlib one ? Is this that performance sensitive ?

paulcacheux: especially considering https://github.com/DataDog/datadog-agent/pull/40397#issuecomment-3237696761

dustmop: Seems like a mistake. I had heard it works better than encoding/json, see [comment here](https://github.com/DataDog/datadog-agent/blob/f561bb44bbbb41b2171fa1f4194e88ea89d8461b/comp/core/settings/settingsimpl/settingsimpl_test.go#L143-L146), but that seems unnecessary and the binary size increase means it's actually undesirable. Switched to use encoding/json instead.

comp/logs/agent/config/config.go
dustmop: Nope, it's made unnecessary by the new logic.

pkg/config/structure/unmarshal.go
dustmop: Switching off of `json-iterator` fixed the quality game, thanks for identifying that! Going to merge now.

PR #40403, opened 2025-08-29
[CXP-2599] Enable WLM for process collection by default on linux

daniel-taf:
### What does this PR do?
* Enable WLM for process collection by default on linux by default
* Fixes/updates any test associated with the change
* updates manual check so that it waits for WLM to be initialized, otherwise results are inconsistent.

### Motivation
The `process_config.process_collection.use_wlm` will be baselined in a subsequent release since WLM is required for the process check to get service data via the `process_collector`. Using a config first will give us an easy way to rollback in case of issues.

e2e and unit tests. This is also already deployed internally on all clusters without issue.


Yumasi (approved):

CelianR (approved):

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -195,7 +195,7 @@ func (c *collector) Start(ctx context.Context, store workloadmeta.Component) err
 		return errors.NewDisabled(componentName, "wlm process collection disabled")
 	}
 
-	if !c.isProcessCollectionEnabled() && !c.isServiceDiscoveryEnabled() {
+	if !c.isProcessCollectionEnabled() && !c.isServiceDiscoveryEnabled() && !c.isLanguageCollectionEnabled() {
```
gengnamstyle, line 198: curiosity - is language detection something that should always be optional? It seems like process collection + service discovery configs should be the primary gate of the collector and not also the language collection. But if language collection can also control this, it makes me think we can also default this to true in the future.

cmd/process-agent/subcommands/check/check.go
```diff
@@ -51,6 +51,26 @@ import (
 
 const defaultWaitInterval = time.Second
 
+func waitForWorkloadMeta(logger log.Component, wm workloadmeta.Component) {
```
gengnamstyle, line 54: nit: can you add a commment on why this was needed

gengnamstyle (approved):

cmd/process-agent/subcommands/check/check.go
daniel-taf: done

comp/core/workloadmeta/collectors/internal/process/process_collector.go
daniel-taf: The config is on by default in K8s environments where its primarily used. Not sure if it would be worthwhile to baseline it.

dustmop (approved):

zhuminyi (approved):

PR #40408, opened 2025-08-30
Add pquery.bzl tool

aiuto:
pquery is a tool to pretty print providers of a target. This is useful for rule development.

Usage:
     `bazel cquery --output=starlark --starlark:file=${PATH_TO_THIS_FILE} //path/to/pkg:target`

For fun, try it:
```
bazel cquery --output=starlark --starlark:file=bazel/tools/pquery.bzl //deps:smoketest
```

I did not include a wrapper to invoke this easily because it's too early. We could potentially share this with dd-source, but it is generally useful, even to other repos.  So perhaps it could be a standalone command with a self extracting .bzl file, so you could just have it in your path.  But then it would need an installer.

But also, every rule developer will most likely use this as a starting point to add a little custom formatting around their particular needs. So, like The Pirate's Code, this is more guidelines than rules.


bazel/tools/pquery.bzl
```diff
@@ -0,0 +1,300 @@
...
+        value = provider,
+        indent = 0,
+    ) for (name, provider) in target_providers.items()[::-1]]
+
+    # No way to infinite loop (`while` not supported, no `Infinity` symbol,
+    # `range(0, 1, -1)` just doesn't work), so instead we loop with the largest
+    # available number of times. Starlark requires the value be a 32-bit signed
+    # integer, so `2 ^ 31 - 1` is the best we can do.
+    result = ""
+    for i in range(_pow(2, 31) - 1):
```
JSGette: why not `len(stack)`?

JSGette: ah, because stack size changes after push

JSGette (approved):

PR #9485, opened 2021-10-12
Remove direct dependency to `k8s.io/kubernetes`

L3n41c:
### What does this PR do?

Remove direct dependency to `k8s.io/kubernetes` GO module.

### Motivation

`k8s.io/kubernetes` package is huge and pulls a ton of useless dependencies.
Those dependencies are triggering dependabot alerts which are false-positive because most of the offending code is unused.

### Additional Notes

In order to be able to remove the dependency to `k8s.io/kubernetes` package, I decided to copy the functions that we were using and that weren’t available elsewhere.
I put them in the `third_party` directory and used a file hierarchy which mimics the one of the upstream project. I copied the files with as few modifications as possible.

### Describe how to test your changes

Write here in detail how you have tested your changes
and instructions on how this should be tested in QA.

Describe or link instructions to set up environment 
for testing, if the process requires more than just
running the agent on one of the supported platforms.

### Checklist

- [ ] A [release note](https://github.com/DataDog/datadog-agent/blob/main/docs/dev/contributing.md#reno) has been added or the `changelog/no-changelog` label has been applied.
- [ ] The `need-change/operator` and `need-change/helm` labels has been applied if applicable.
- [ ] The appropriate `team/..` label has been applied, if known.
- [ ] If known, an appropriate milestone has been selected; otherwise the `Triage` milestone is set.
- [ ] The [config template](https://github.com/DataDog/datadog-agent/blob/main/pkg/config/config_template.yaml) has been updated if applicable.

Note: Adding GitHub labels is only possible for contributors with write access.

PR #40414, opened 2025-09-01
[util/containers/cri] Use k8s.io/cri-client mod

davidor:
### What does this PR do?

Removes the copied CRI client code from `internal/third_party/kubernetes/pkg/kubelet/cri/remote/fake/` and replaces it with the upstream `k8s.io/cri-client` module.

This code was copied to avoid depending on `k8s.io/kubernetes` (#9485). Since then, the CRI client has been extracted into its own separate module (`k8s.io/cri-client`).

This PR deletes the duplicate code and updates imports to use the upstream module instead.

CI


L3n41c (approved):
👍🏻

PR #40422, opened 2025-09-01
[dynamic tests]: Improve telemetry handling so that it is reusable in other classes

KevinFairise2:
### What does this PR do?

Introduce a telemetry handler in dynamic tests. And use it to upload the evaluation results.
The goal is to have that reusable component that can be used in other dynamic test classes later on.

### Motivation

Proper handling of the telemetry


CelianR (approved):
Good refactoring! However I'm not sure for the use case of the base class, can we have anything else than Datadog telemetry? 

The other thing I might add is a mock implementation when writing some integration tests

PR #40424, opened 2025-09-01
[dynamic tests]: Introduce a file based index to compare accuracy against package based index

KevinFairise2:
### What does this PR do?

Introduce and evaluate a new index that relies on code coverage, instead of considering only package, we now consider each file in the coverage report. The goal is to have a more precise index and to evaluate if this extra accuracy can make us skipping some test that we should not.

### Motivation

Running only relevant test, and trying to run as few of them as possible

### DIFF
--- a/tasks/dyntest.py
+++ b/tasks/dyntest.py
@@ -6,6 +6,7 @@ import os
 
 from invoke import Context, task
 
+from tasks.libs.common.color import Color, color_message
 from tasks.libs.common.git import get_modified_files
 from tasks.libs.dynamic_test.backend import S3Backend
 from tasks.libs.dynamic_test.evaluator import DatadogDynTestEvaluator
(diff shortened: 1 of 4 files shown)


tasks/dyntest.py
```diff
@@ -11,33 +11,56 @@
...
     changes = get_modified_files(ctx)
     results = evaluator.evaluate([os.path.dirname(change) for change in changes])
     evaluator.print_summary(results)
     evaluator.send_stats_to_datadog(results)
+
+    # File coverage indexer
+    executor = DynTestExecutor(ctx, uploader, IndexKind.FILE, commit_sha)
+    evaluator = DatadogDynTestEvaluator(ctx, IndexKind.FILE, executor, pipeline_id)
+    if not evaluator.initialize():
+        print("Failed to initialize index for file coverage")
```
CelianR: NIT: Maybe we can prefix this by a yellow `WARNING` (same above)

tasks/libs/dynamic_test/indexers/e2e.py
```diff
@@ -112,6 +121,167 @@ def _read_metadata(self, suite_folder: Path) -> dict[str, str]:
...
+                    file_path, has_coverage = self._parse_coverage_line(line)
+                    if not has_coverage or not file_path:
+                        continue
+
+                    # Extract relative file path and keep .go extension
+                    relative_file_path = self._extract_relative_path(file_path)
+                    if relative_file_path:
+                        covered.add(relative_file_path)
+        except Exception as e:
+            print(color_message(f"Error parsing coverage file {coverage_txt}: {e}", Color.ORANGE))
```
CelianR, line 250: Could you either return an empty set or add a try except inside the for? (Either we don't return or error or we skip only wrong lines)

CelianR (approved):

PR #40425, opened 2025-09-01
Remove various configurations and variables related to Agent 6

pgimalac:
### What does this PR do?
Remove various configurations and variables related to Agent 6:
- `python_version` config
- `DefaultPython` and `ForceDefaultPython` global variables
- `validatePython3` linter (only used in Agent 6)
- TraceMalloc with Python 2 check
- `agent-py2py3` config generation

### Motivation
They are not needed anymore since we always just have Python 3 in this codebase.

It simplifies some code, and will make it easier to run some tests from IDEs (less options to specify for the test command, in particular the global variables default value).

CI

### Additional Notes
The only part I really want to merge is removing `DefaultPython` and `ForceDefaultPython`, as they are causing issues when running tests from my IDE (I could just add the compile flag to set them if needed, but it's best to remove them since they're useless).


cmd/agent/subcommands/integrations/command.go
```diff
@@ -104,7 +104,7 @@ func Commands(globalParams *command.GlobalParams) []*cobra.Command {
 	integrationCmd.PersistentFlags().CountVarP(&cliParams.verbose, "verbose", "v", "enable verbose logging")
 	integrationCmd.PersistentFlags().BoolVarP(&cliParams.allowRoot, "allow-root", "r", false, "flag to enable root to install packages")
 	integrationCmd.PersistentFlags().BoolVarP(&cliParams.useSysPython, "use-sys-python", "p", false, "use system python instead [dev flag]")
-	integrationCmd.PersistentFlags().StringVarP(&cliParams.pythonMajorVersion, "python", "", "", "the version of Python to act upon (2 or 3). defaults to the python_version setting in datadog.yaml")
+	integrationCmd.PersistentFlags().StringVarP(&cliParams.pythonMajorVersion, "python", "", "", "the version of Python to act upon (2 or 3). Defaults to 3.")
```
louis-cqrl, line 107: since we do not support python2 anymore, should we just remove any python2 mentions ?

pgimalac: For that one specifically I'm not sure because it's part of the CLI...
I don't think anyone installs python 2 integrations with Agent 7, but I wanted to ask Agent Integrations first as that would be a visible change for users

louis-cqrl (approved):
LGTM for ARUN files

pkg/collector/python/loader.go
```diff
@@ -308,39 +308,8 @@ func reportPy3Warnings(checkName string, checkFilePath string) {
...
-
-		if strings.TrimSpace(pkgconfigsetup.Datadog().GetString("python_version")) == "3" {
-			// the linter used by validatePython3 doesn't work when run from python3
-			status = a7TagPython3
-			metricValue = 1.0
-		} else {
-			// validatePython3 is CPU and memory hungry, make sure we only run one instance of it
-			// at once to avoid CPU and mem usage spikes
-			linterLock.Lock()
-			warnings, err := validatePython3(checkFilePath)
```
sgnn7, line 324: @pgimalac Drive-by conceptual question: Should we have this as part of Diagnose tooling?

pgimalac: It's an interesting suggestion, but the linter helps transitioning from python2 to python3 so I'm not quite sure it's useful for users of Agent 7 (this code never runs in agent 7 btw), it would require adding an extra python dependency (`pylint`) to the package, and lastly it has been removed from Agent 7 because [it could not run on Python 3](https://github.com/pylint-dev/pylint/issues/1969#issuecomment-377127442) (seems to have been fixed in more recent versions though, but I haven't tested)

chouetz (approved):

julien-lebot (approved):
✅ for Windows product changes

adel121 (approved):

joepeeples (approved):

pkg/collector/python/loader.go
sgnn7: I suppose my curiosity is if we should include a regular linter (`pylint`) warnings/errors (not `py3k` specifically) in the diagnose command output to find issues with problematic Python code either in our integrations or custom ones. Fair point that the extra dep is definitely a barrier in terms of footprint constraints though.

dustmop (approved):

PR #40428, opened 2025-09-01
[USMON-1476] usm: redis: Add track_resources configuration option

guyarb:
### What does this PR do?

Adds a new configuration option `service_monitoring_config.redis.track_resources` that controls whether Redis monitoring tracks resource names (key names) or only command types. When disabled (default), the monitoring becomes more efficient by avoiding key name collection and storage.

### Motivation

This change provides users with the ability to optimize Redis monitoring performance by disabling resource tracking when detailed key-level metrics are not needed. This can reduce memory usage and processing overhead in high-throughput Redis environments while still maintaining command-level monitoring.

- Added comprehensive unit tests for the new configuration option covering YAML, environment variable, and default scenarios
- Enhanced existing Redis monitoring tests to cover both enabled and disabled resource tracking modes
- Verified that when `track_resources` is false, key names are not collected or reported
- Verified that when `track_resources` is true, behavior remains unchanged from current implementation
- All existing tests pass with the changes
- Built and tested system-probe component successfully

### Additional Notes

- Default value is `false` to optimize performance by default
- Implementation is done in atomic commits: Phase 1 adds configuration infrastructure, Phase 2 implements the conditional tracking
- Changes maintain backward compatibility - existing behavior is preserved when the option is enabled


amitslavin (approved):

pkg/config/setup/system_probe.go
```diff
@@ -265,6 +265,7 @@ func InitSystemProbeConfig(cfg pkgconfigmodel.Setup) {
 	cfg.BindEnvAndSetDefault(join(smNS, "enable_kafka_monitoring"), false)
 	cfg.BindEnv(join(smNS, "enable_postgres_monitoring"))
 	cfg.BindEnv(join(smNS, "enable_redis_monitoring"))
+	cfg.BindEnv(join(smNS, "redis", "track_resources"))
```
hush-hush: Please add a default for any new settings.

hush-hush (changes requested):

pkg/config/setup/system_probe.go
guyarb: the default for a boolean is false, why is it needed to be explicit?

gjulianm (approved):

pkg/config/setup/system_probe.go
hush-hush: Because we want to move toward a schema for the configuration and for this we need a type for every settings.

The way the config works right now is that we only enforce/convert settings to the right type when accessing them (with the GetXXX method). It's not great but that the way it is for now.

From the point of view of the config, we don't know how each setting will be queried and we don't validate the data from the YAML config file. In other word, it's not because you use `GetBool` that the configuration contains a bool.

guyarb: fixed

akarpz (approved):

hush-hush (approved):

PR #40429, opened 2025-09-01
process: service: make service heartbeats do partial updates

Yumasi:
### What does this PR do?

This PR make the process collector, when service collection is enabled, able to request partial Service updates from system-probe on heartbeats. This is done by separating the `pids` parameter of the requests made to SP between new pids and "heartbeat pids". The SP `/discovery/services` path will now perform full collection of service info on the new pids, as before, and collect only the ports and log files on pids that need a heartbeat.

### Motivation

DSCVR-188
In short, with the removal of service info caching in system-probe, the existing code was re-computing every fields on heartbeats, including service names, language, etc... which are expensive operations and unneeded.

This PR prevents the expensive updates and should reduce the number of allocations made by service updates.

Change covered by unit tests.

### DIFF
--- a/comp/core/workloadmeta/collectors/internal/process/process_collector.go
+++ b/comp/core/workloadmeta/collectors/internal/process/process_collector.go
@@ -310,14 +310,13 @@ func (c *collector) detectLanguages(processes []*procutil.Process) []*languagemo
 	return nil
 }
 
-// filterPidsToRequest filters PIDs to only request services for new or stale processes.
-// It returns a slice of pids to request (to be used as a request parameters), and
-// a map of pids to *model.Service to be filled up with the response received from
-// system-probe. This map is useful to know for which pids we have not received
-// service info and that needs to be handled by the retry mechanism.
-func (c *collector) filterPidsToRequest(alivePids core.PidSet, procs map[int32]*procutil.Process) ([]int32, map[int32]*model.Service) {
+// filterPidsToRequest filters PIDs to categorize them as new or needing heartbeat refresh.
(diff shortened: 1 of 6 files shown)


vitkyrka (approved):

comp/core/workloadmeta/collectors/internal/process/process_collector.go
```diff
@@ -432,17 +436,59 @@ func (c *collector) getProcessEntitiesFromServices(pids []int32, pidsToService m
...
+		if err != nil || existingProcess == nil || existingProcess.Service == nil {
+			log.Debugf("Heartbeat for pid %d but no existing service found, skipping", pid)
+			continue
+		}
+
+		c.pidHeartbeats[int32(service.PID)] = now
+
+		// For heartbeat updates, preserve static fields from existing service
+		// Since workloadmeta replaces entities from the same source instead of merging,
+		// we need to preserve static fields here
```
adel121, line 458: Question: 

Why would these static fields change in the first place? 

In other words, if they are static, we shouldn't have a problem if workloadmeta merges them.

Yumasi: Because in the payload we receive from system-probe for the heartbeats, all of these static fields will be set to nil, and since we use the service discovery source for heartbeats as well, workloadmeta won't merge them but rather overwrite the service data with nil values.

We could cache those fields in system-probe or the collector (and we actually were doing that with the old check), but now we prefer to have the data in only one place and limit deduplication of this data in memory. So we use workloadmeta as the cache instead.

daniel-taf (approved):

adel121 (approved):

PR #40445, opened 2025-09-02
[EBPF-808] gpu: use pools for spans and enriched kernel launches

gjulianm:
### What does this PR do?

Enables the use of pools for objects that are allocated in the heap in the consumer hot path. These were causing a high number of allocations during runtime, which caused GC pressure and high CPU usage.

To monitor leaks and the state of the pools, this PR also includes a `TypedPool` structure with telemetry.

### Motivation

Reduce CPU usage in prod.

Included tests to ensure that objects are released back to the pool properly in all cases.

Benchmarked in load-test environment, showing that for high workloads we have a CPU usage reduction


jeremy-hanna (approved):

pkg/util/sync/pool.go
```diff
@@ -64,3 +69,64 @@ func (t *TypedPool[K]) Get() *K {
 func (t *TypedPool[K]) Put(x *K) {
 	t.p.Put(x)
 }
+
+// typedPoolWithTelemetry is a TypedPool with telemetry counters
+type typedPoolWithTelemetry[K any] struct {
+	p  *TypedPool[K]
```
val06: why not embed?

pkg/util/sync/pool.go
```diff
@@ -64,3 +69,64 @@ func (t *TypedPool[K]) Get() *K {
...
+func newPoolTelemetry(tm telemetry.Component) *poolTelemetry {
+	return &poolTelemetry{
+		get:    tm.NewCounter("sync__pool", "get", []string{"module", "name"}, "Number of gets from the sync pool"),
+		put:    tm.NewCounter("sync__pool", "put", []string{"module", "name"}, "Number of puts to the sync pool"),
+		active: tm.NewGauge("sync__pool", "active", []string{"module", "name"}, "Number of active items in the sync pool"),
+	}
+}
+
+// globalPoolTelemetry is a memoized function to create a single poolTelemetry instance from the telemetry component
+var globalPoolTelemetry = funcs.MemoizeArgNoError(newPoolTelemetry)
```
val06, line 111: can't we just replace this with static initialization of the local unexported var?

pkg/gpu/stream.go
```diff
@@ -28,7 +31,7 @@ const noSmVersion uint32 = 0
 // kernel spans and memory allocations from them.
 type StreamHandler struct {
 	metadata           streamMetadata
-	kernelLaunches     []enrichedKernelLaunch
+	kernelLaunches     []*enrichedKernelLaunch
```
val06, line 34: why did you decide to store by pointer?

pkg/gpu/stream.go
```diff
@@ -131,6 +144,20 @@ type enrichedKernelLaunch struct {
 
 var errFatbinParsingDisabled = errors.New("fatbin parsing is disabled")
 
+var enrichedKernelLaunchPool ddsync.Pool[enrichedKernelLaunch]
+var kernelSpanPool ddsync.Pool[kernelSpan]
+var memorySpanPool ddsync.Pool[memorySpan]
```
val06: why do we have them as global vars and not part of some struct? feels like we are missing some layer that will hold those and the sync.Once declared below

pkg/gpu/stream.go
```diff
@@ -197,11 +223,12 @@ func (sh *StreamHandler) handleKernelLaunch(event *gpuebpf.CudaKernelLaunch) {
 
 // trySendToChannel attempts to send an item to a channel in a non-blocking way, if the channel is full
 // it will increment the rejectedSpans telemetry counter
-func trySendSpan[T any](sh *StreamHandler, ch chan T, item T) {
+func trySendSpan[T any](sh *StreamHandler, ch chan *T, item *T, pool ddsync.Pool[T]) {
```
val06, line 231: i don't think trySendSpan should be in charge of managing the pool, maybe better have some return value and leave that logic to the caller?

pkg/gpu/stream.go
```diff
@@ -172,10 +199,9 @@ func newStreamHandler(metadata streamMetadata, sysCtx *systemContext, config con
 func (sh *StreamHandler) handleKernelLaunch(event *gpuebpf.CudaKernelLaunch) {
 	sh.lastEventKtimeNs = event.Header.Ktime_ns
 
-	enrichedLaunch := &enrichedKernelLaunch{
-		CudaKernelLaunch: *event, // Copy events, as the memory can be overwritten in the ring buffer after the function returns
-		stream:           sh,
-	}
+	enrichedLaunch := enrichedKernelLaunchPool.Get()
+	enrichedLaunch.CudaKernelLaunch = *event // Copy events, as the memory can be overwritten in the ring buffer after the function returns
```
val06, line 208: what do you mean by "copy events" in this case?

val06 (commented):
left comments

pkg/gpu/stream.go
gjulianm: So that we can control the allocations in the pool, without the pointers the memory regions for this are controlled by the array. There was also a small mistake in handleKernelLaunch where enrichedKernelLaunch was being created as a pointer and then copied into the array, so by switching to use pointers all of the time we are consistent.

pkg/gpu/stream.go
gjulianm: Fixed

pkg/gpu/stream.go
gjulianm: Copying all of the fields to the "enrichedLaunch" object, as the event we get from eBPF is a pointer to buffer memory that can change afterwards

pkg/gpu/stream.go
gjulianm: I did it this way to avoid forgetting to do it and avoid copying the same "check, if error return to pool" logic everywhere. All of the "send span" code uses this helper so it made sense to me to add this here.

pkg/util/sync/pool.go
gjulianm: Unfortunately we need the telemetry component to be able to create the counters and gauges, so we can't do static initialization

val06 (approved):

PR #40455, opened 2025-09-02
fakeintake client parse host-tags payloads

lavigne958:
### What does this PR do?

Add necessary code in the `fakeintake` library to parse and deserialize the host-impl tags.

This requires some filtering, the route to read the tags contains other payloads that could match while calling `Unmarshall`.

The field `meta` is located in the payload we try to extract, but it's also located in some other struct that we don't want.

Use an inner filed allocated with a pointer to check for its presence and therefore validate if we are looking at the right payload.

Add necessary code to query the host-implementation tags from the command line.

Once we have a match return all payloads.

### Motivation

This allows us later to add some end2end tests for host-tags

1. deploy a test stack using `pulumi`
2. use the `fakeintakectl` to query the host tags
3. see the host-tags in the output

command:
```
./test/fakeintake/build/fakeintakectl --url "http://w.x.y.z:80" get host-infos
```

output:
```
[
  {
    "host-tags": {
      "system": [
        "arch:amd64",
        "cluster_name:gcp-gcp-alexandre-lavigne-gke-deb76db",
        "kube_cluster_name:gcp-gcp-alexandre-lavigne-gke-deb76db",
        "kube_node:gke-gcp-gcp-alexandre-la-default-pool-7a56dc29-8ggv",
        "orch_cluster_id:f340972a-1311-47b3-a6af-c6f83a7ece1c",
        "os:linux"
      ]
    },
    "internalHostname": "gke-gcp-gcp-alexandre-la-default-pool-7a56dc29-8ggv.c.datadog-agent-sandbox.internal"
  }
]
```

### DIFF
--- a/test/fakeintake/aggregator/hostAggregator.go
+++ b/test/fakeintake/aggregator/hostAggregator.go
@@ -1,0 +1,88 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-present Datadog, Inc.
+
+package aggregator
+
+import (
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"github.com/DataDog/datadog-agent/test/fakeintake/api"
+)
+
+// Host struct contains agents host-tags payload and attributes to fit Aggregator implementation
+//
+// Use a pointer to assign `HostTags` inner struct.
+// When we ParseHostPayload we receive variaous payload types.
+// We only want to keep those with host-tags
+// Using a pointer allows us to check if the pointer has been allocated
+// and therefore found the right payload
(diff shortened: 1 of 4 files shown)


lavigne958:
trying to pull the whole payload has a drawbacks !

1. if you try to importe the `hostimpl.Payload` structure, you will then import as dependency the whole datadog-agent.... we don't want such strong coupling between the fakeintake and the agent

2. if you copy the structure definition, you will need to copy *all* inner *private* structure defintion from the payload ! that a lot of structure definition to maintain, it's error prone for the future when any changes will happen.

*Suggestion*:

- we only want to test for `host-tags` validation
- only define the necessary structure to parse host-tags
- add tests to cover the content validation, so if anything changes on the datadog-agent payload side -> this test should fail

CelianR (approved):

PR #40460, opened 2025-09-02
Template snapshot support

grantseltzer:
### What does this PR do?

Add's template support for capturing top level parameters in log probes, including snapshots.

### Motivation

https://datadoghq.atlassian.net/browse/DEBUG-4426

Added test cases to integration tests and others.

### DIFF
--- a/pkg/dyninst/decode/decoder.go
+++ b/pkg/dyninst/decode/decoder.go
@@ -23,6 +23,7 @@ import (
 	"github.com/DataDog/datadog-agent/pkg/dyninst/gotype"
 	"github.com/DataDog/datadog-agent/pkg/dyninst/ir"
 	"github.com/DataDog/datadog-agent/pkg/dyninst/output"
+	"github.com/DataDog/datadog-agent/pkg/dyninst/rcjson"
 	"github.com/DataDog/datadog-agent/pkg/dyninst/symbol"
 	"github.com/DataDog/datadog-agent/pkg/util/log"
 )
@@ -188,8 +189,8 @@ type snapshotMessage struct {
 	Logger    logger           `json:"logger"`
 	Debugger  debuggerData     `json:"debugger"`
 	Timestamp int              `json:"timestamp"`
-
-	rootData []byte
+	Message   message          `json:"message"`
+	rootData  []byte
 }
 
 func (s *snapshotMessage) init(
@@ -277,7 +278,7 @@ func (s *snapshotMessage) init(
 			probe.GetID(), where,
 		)
 	}
-
+	s.Message.probe = probe
 	s.Logger.Version = probe.GetVersion()
 	s.Logger.ThreadID = int(header.Goid)
 	s.Debugger.Snapshot.Probe.ID = probe.GetID()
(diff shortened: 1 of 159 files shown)


pkg/dyninst/decode/decoder.go
```diff
@@ -247,15 +248,35 @@ func (s *snapshotMessage) init(
...
+func populateNeededValues(probe ir.ProbeDefinition) map[string]string {
+	segments := probe.GetSegments()
+	neededValues := make(map[string]string, 0)
+	for i := range segments {
+		switch seg := segments[i].(type) {
+		case rcjson.JSONSegment:
+			var parameterName string
+			// XXX: For now we only support parameters in templates
+			if json.Unmarshal(seg.JSON, &parameterName) != nil {
+				// TODO: surface error to evaluationErrors
```
grantseltzer: I'll do this as part of this PR

pkg/dyninst/decode/marshal.go
```diff
@@ -31,6 +33,32 @@ type debuggerData struct {
...
+	for i := range segments {
+		switch seg := segments[i].(type) {
+		case rcjson.StringSegment:
+			message += string(seg)
+		case rcjson.JSONSegment:
+			// TODO: seg.DSL right now only supports parameter names
+			// so we don't need to handle the case where the DSL is a JSON object
+			val, ok := m.argumentsData.neededValues[seg.DSL]
+			if !ok {
+				// TODO: handle error to evaluationErrors
```
grantseltzer: I'll do this as part of this PR

pkg/dyninst/decode/marshal.go
```diff
@@ -31,6 +33,32 @@ type debuggerData struct {
...
+
+func (m *message) MarshalJSONTo(enc *jsontext.Encoder) error {
+	segments := m.probe.GetSegments()
+	message := ""
+	for i := range segments {
+		switch seg := segments[i].(type) {
+		case rcjson.StringSegment:
+			message += string(seg)
+		case rcjson.JSONSegment:
+			// TODO: seg.DSL right now only supports parameter names
```
grantseltzer: I need to add  a test for this

ajwerner (commented):
My initial and rather high-level feedback is that I think you're pushing the task of analyzing the templates far too much into the decoder package. The `ir` should capture the types of the expressions expicitly in the events. 

Specifically, I'd look here: https://github.com/DataDog/datadog-agent/blob/38e3e8eb618625fe8aece797e80fc397779620c0/pkg/dyninst/ir/program.go#L137-L138.

I'd expect then the TemplateSegments will reference root expressions, or vice-versa. We'll then annotate the root expressions that are arguments (or again, could be the other way around where the subprogram references root expressions in an event, idk what's better). 

the tl;dr is perhaps there should be a whole PR or commit series that's just about ir and irgen changes and then one after that that adopts those changes in the decoder.

grantseltzer:
Closing this for now. I started thinking about what the ir changes would look like when we implement expressions and realized the way I have it here won't be extendable enough to handle it. I've opened #40901 for the sake of discussion at first.

PR #40463, opened 2025-09-02
Report successful first RC update

mellon85:
- Reworked the Remote Configuration poll loop
- The Remote Configuration Client doesn't log when it does a first update successfully but only if it fails which makes it harder to understand when it's in working state.
This can happen as the trace-agent can start faster then the core-agent.

The modified/added logs look like this

```
2025-09-03 10:09:38 CEST | TRACE | INFO | (pkg/config/remote/client/client.go:404 in pollLoop) | retrying the first update of remote-config state (rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp [::1]:5001: connect: connection refused"), consecutive failures: 1
2025-09-03 10:09:39 CEST | TRACE | INFO | (pkg/config/remote/client/client.go:404 in pollLoop) | retrying the first update of remote-config state (rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp [::1]:5001: connect: connection refused"), consecutive failures: 2
2025-09-03 10:09:40 CEST | TRACE | INFO | (pkg/config/remote/client/client.go:404 in pollLoop) | retrying the first update of remote-config state (rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp [::1]:5001: connect: connection refused"), consecutive failures: 3
2025-09-03 10:09:41 CEST | TRACE | INFO | (pkg/config/remote/client/client.go:404 in pollLoop) | retrying the first update of remote-config state (rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp [::1]:5001: connect: connection refused"), consecutive failures: 4
2025-09-03 10:09:42 CEST | TRACE | ERROR | (pkg/config/remote/client/client.go:402 in pollLoop) | retrying the first update of remote-config state (rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp [::1]:5001: connect: connection refused"), consecutive failures: 5
2025-09-03 10:09:46 CEST | TRACE | DEBUG | (pkg/config/remote/client/client.go:421 in pollLoop) | update successful: successful_first_run:false, consecutive failures:8
2025-09-03 10:09:46 CEST | TRACE | INFO | (pkg/config/remote/client/client.go:434 in pollLoop) | first update successful after 8 attempts
```

### DIFF
--- a/pkg/config/remote/client/client.go
+++ b/pkg/config/remote/client/client.go
@@ -39,6 +39,9 @@ const (
 	recoveryInterval      = 2
 
 	maxMessageSize = 1024 * 1024 * 110 // 110MB, current backend limit
+
+	// number of update failed update attempts before reporting an error log
+	consecutiveFailuresThreshold = 5
 )
 
 // ConfigFetcher defines the interface that an agent client uses to get config updates
@@ -398,7 +401,7 @@ func (c *Client) pollLoop() {
 					// as an Info log as the race is expected. If the error persists, we log with error logs
 					if logLimit.ShouldLog() {
 						message := "retrying the first update of remote-config state (%v), consecutive failures: %d"
-						if consecutiveFailures >= 5 {
+						if consecutiveFailures >= consecutiveFailuresThreshold {
 							log.Errorf(message, err, consecutiveFailures)
 						} else {
 							log.Infof(message, err, consecutiveFailures)
--- a/pkg/config/remote/client/client.go
+++ b/pkg/config/remote/client/client.go
@@ -371,35 +371,10 @@ func (c *Client) startFn() {
 // structure in startFn.
 func (c *Client) pollLoop() {
 	successfulFirstRun := false
-	// First run
-	err := c.update()
-	if err != nil {
-		if status.Code(err) == codes.Unimplemented {
-			// Remote Configuration is disabled as the server isn't initialized
-			//
-			// As this is not a transient error (that would be codes.Unavailable),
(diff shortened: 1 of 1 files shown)


pkg/config/remote/client/client.go
```diff
@@ -436,6 +436,10 @@ func (c *Client) pollLoop() {
 					log.Errorf("could not update remote-config state: %v", c.lastUpdateError)
 				}
 			} else {
+				if !successfulFirstRun {
+					log.Infof("first update successful")
```
domodwyer: Is this variable misnamed? "if not successful first run, log successful first update" ?

mellon85: I forgot to remove the `!`

mellon85: the variable is reporting if a first update *was* successful, at this point in the code it's still `false` even if it's the first time, it's updated at the end of the function.
It's in the `else` bracket, so the updated was ok

ameske: We unconditionally on line 442 debug log that it was successful right after this - so do we really need this check?

ameske: I think we should move it to be strictly next to where we set the variable to true after a first successful run, so that it's more clear we only want to log this for the first successful one. If it's here it's likely to cause a double take on every reading of the code since it is a little tricky.

przemko0 (approved):

aya-shimizu24 (approved):

PR #40472, opened 2025-09-02
[pkg/dyninst/dwarf] Don't fail hard when parsing loclists

piob-io:
We can warn, but pretend the variable is unavailable

At this point we should be ready to enable go1.25 testing, but that requires ebpf review, so sending a separate PR.

https://datadoghq.atlassian.net/browse/DEBUG-4385

### DIFF
--- a/pkg/dyninst/dwarf/loclist/parse.go
+++ b/pkg/dyninst/dwarf/loclist/parse.go
@@ -69,6 +69,13 @@ func ParseInstructions(data []byte, ptrSize uint8, totalByteSize uint32) ([]ir.P
 	pieces := []ir.Piece{}
 	instCnt := 0
 
+	// We expect the ops list to be either:
+	// - a single op
+	// - multiple ops each followed by at least one piece: op, piece, op, piece, piece, op, piece, ...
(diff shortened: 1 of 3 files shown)


pkg/dyninst/dwarf/loclist/parse.go
```diff
@@ -89,7 +89,11 @@ func ParseInstructions(data []byte, ptrSize uint8, totalByteSize uint32) ([]ir.P
 		}
 
 		if op != nil {
-			return nil, fmt.Errorf("unconsumed op: %v", op)
+			// TODO(piob): support dw_op_deref
```
andreimatei: consider putting a comment around here, maybe on top of the `for reader.Len() > 0` loop, saying that the structure we expect is op, then piece, then op, etc. And also say that there are either multiple pieces, each ending with op_piece, or a single piece corresponding to all the instructions.

andreimatei (approved):
LGTM

pkg/dyninst/dwarf/loclist/parse.go
piob-io: Moved the comment to be on the loop, and rephrased for clarity.

PR #40436, opened 2025-09-02
Final updates for release.json and Go modules for 7.70.0 release + preludes

FlorentClarret:

PR #40479, opened 2025-09-02
Add date argument to release.finish task

matt-dz:
### What does this PR do?

Adds a release date (format YYYY-MM-DD) argument to the `release.finish` task. The date is added to the generated prelude yaml files. If no date is provided, the default is the current date.

Usage: `dda -- inv -e release.finish 7.69.x --release-date 2025-09-03`.

### Motivation

Currently, when running the release.finish task, users must manually edit generated files for future release dates. This process is cumbersome and prone to errors.

https://datadoghq.atlassian.net/browse/BARX-1163


FlorentClarret (commented):
When we run the `finish` command, we generate two yaml files called "preludes". This is where we have a date. By default, this date is the current date and this is where we want to be able to provide another one. 

For example, [this PR](https://github.com/DataDog/datadog-agent/pull/40436) is generated by the `finish` command and contains two prelude file where you have the date. What. we'd like is to be able to override this date so we can set one in the future

PR #40482, opened 2025-09-03
dyninst/irgen: only add needed types based on reference limits

ajwerner:
Part of [DEBUG-4043](https://datadoghq.atlassian.net/browse/DEBUG-4043) to control the type graph explosion. Review as individual commits.

Fixes [DEBUG-4272](https://datadoghq.atlassian.net/browse/DEBUG-4272).

#### dyninst/testprogs: add deep type reference testing

This commit adds some deeply nested types to the complex scenario
in the sample testprog and rewrites the tests accordingly.

We see that the output only gets some of the chain.

#### dyninst/irgen: per-probe budgets in bfs

This reworks type expansion to use a BFS scheme that is respects
per-probe budgets. It also makes it straightforward to avoid expanding
non-parameter types that we do not need to expand.

#### dynisnt/ir,irgen: allow some pointers to not be resolved

This commit introduces the notion of an ir.UnresolvedPointeeType which
corresponds to a type we don't know anything about beyond its name, but
we might collect a pointer to it. Practically speaking we should never
be able to collect one of these because of pointer chasing limits in
bpf.

This new type is crucial to be able to have a well-formed type graph
without resolving everything.

This change doesn't perfectly model the minimal amount of type
resolution; it takes the maximum depth across all probes and then
resolves types to that depth.

For testing, this change now makes sure that as we increase the depth,
eventually we converge on the set of types we'd find if we have the
maximal depth we tend to use in our probes.

#### dyninst/irgen: chase the types more lazily

Before this change, whenever we'd add a type, we'd always recursively
add all the reachable types. This change refactors the code so that when
we add a type, if we get to a pointer, we just leave a placeholder for
the pointee that we can come back to later.

This change doesn't actually stop chasing anything based on depth, but
it sets the stage for that in a follow-up. Now to chase everything down
we keep processing types until we hit a fixed point with nothing left to
process.

#### dyninst/irgen: defer type creation

Before this change, we created types eagerly when we visited subprograms
and variables. This made it very hard to control the depth to which we
constructed the type graph. This refactor should have no functional
change, though it does rearrange the order in which we instantiate
types, and thus the type IDs.

#### dyninst/compiler: sort things

Sort the functions by name so there's a bit more stability
when the set of types changes.

#### dyninst/irprinter: sort types by name

We don't really care what IDs they get, do we? Or rather, when that
changes, let's not shuffle everything.

#### dyninst/irgen: do not reprocess map types

For whatever reason, we were revisiting map types which lead to duplicates
in the type graph.

[DEBUG-4272]: https://datadoghq.atlassian.net/browse/DEBUG-4272?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/dyninst/irgen/irgen.go
+++ b/pkg/dyninst/irgen/irgen.go
@@ -232,7 +232,7 @@ func generateIR(
 func materializePending(
 	loclistReader *loclist.Reader,
 	pointerSize uint8,
-	typeCatalog *typeCatalog,
+	tc *typeCatalog,
 	pending []*pendingSubprogram,
 ) ([]*ir.Subprogram, error) {
 	subprograms := make([]*ir.Subprogram, 0, len(pending))
@@ -255,7 +255,7 @@ func materializePending(
 			v, err := processVariable(
 				p.unit, die, die.Tag == dwarf.TagFormalParameter,
 				parseLocs, ranges,
-				loclistReader, pointerSize, typeCatalog,
+				loclistReader, pointerSize, tc,
 			)
 			if err != nil {
 				return nil, err
@@ -295,7 +295,7 @@ func materializePending(
 					v, err := processVariable(
 						p.unit, inlVar, inlVar.Tag == dwarf.TagFormalParameter,
 						true /* parseLocations */, ranges,
-						loclistReader, pointerSize, typeCatalog,
+						loclistReader, pointerSize, tc,
 					)
 					if err != nil {
 						return nil, err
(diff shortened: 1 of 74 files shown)


pkg/dyninst/irgen/irgen.go
```diff
@@ -275,6 +244,235 @@ func generateIR(
...
+
+	// Local helper to complete a just-added type ID.
+	ensureCompleted := func(id ir.TypeID) error {
+		return completeGoTypes(tc, id, id)
+	}
+
+	for q.Len() > 0 {
+		q.Swap(0, q.Len()-1)
+		wi := q.pop()
+		heap.Fix(q, 0)
```
piob-io: We are not far from implementing heap fully, I'd just do it, so that this logic is encapsulated and here we only need to call pop/push.

piob-io (approved):
Neat!

Let's test that map, slices and strings don't consume reference depth budget. (For strings it's only that if header managed to get under the budget, data is always captured).

pkg/dyninst/irgen/irgen.go
ajwerner: I realized that I can just no-op the Push and Pop methods and then the right things happen.

ajwerner:
Alright, added the testing for maps, slices, and strings. Thanks for the review!

PR #40485, opened 2025-09-03
system-probe: enable configsync component

paulcacheux:
### What does this PR do?

This PR adds a dependency on the configsync component for the `run` CLI command of the system-probe, allowing to sync some config options from the core agent.

The component is configured to force a first sync on start (with a 10 seconds timeout) so that we can be sure the sync is done before the config is read during the system-probe boot procedure.

Side note: configsync uses the new-ish agent_ipc mechanism that is disabled by default for now (but enabled internally).


mbertrone (approved):
LGTM!

mbertrone (approved):

PR #40486, opened 2025-09-03
[CWS] use cookie helper to resolver flow pid

safchain:
### What does this PR do?

Use the socket cookie helpers to improve the pid context resolution in flow related detection.

### DIFF
--- a/pkg/security/ebpf/c/include/helpers/network/pid_resolver.h
+++ b/pkg/security/ebpf/c/include/helpers/network/pid_resolver.h
@@ -43,13 +43,25 @@ __attribute__((always_inline)) void resolve_pid_from_flow_pid(struct packet_t *p
     pkt->pid = get_flow_pid(&pid_route);
 }
 
-__attribute__((always_inline)) void resolve_pid(struct packet_t *pkt) {
+__attribute__((always_inline)) void resolve_pid(struct __sk_buff *skb, struct packet_t *pkt) {
+    // direct pid from sched_cls
     u64 sched_cls_has_current_pid_tgid_helper = 0;
     LOAD_CONSTANT("sched_cls_has_current_pid_tgid_helper", sched_cls_has_current_pid_tgid_helper);
     if (sched_cls_has_current_pid_tgid_helper) {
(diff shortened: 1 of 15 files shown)


pkg/security/ebpf/probes/all.go
```diff
@@ -330,6 +332,13 @@ func AllMapSpecEditors(numCPU int, opts MapSpecEditorOpts, kv *kernel.Version) m
 		}
 	}
 
+	if opts.CgroupSocketEnabled {
+		editors["sock_cookie_pid"] = manager.MapSpecEditor{
+			MaxEntries: 40960,
```
spikat: do we need this such amount of entries? Is there a `0` to remove?

spikat (approved):

pkg/security/ebpf/probes/all.go
safchain: I clearly don't know what is a good number here, I found 4096 socket a bit too low. I'll try to evaluate the number of time we get flows without a PID and I'll adapt later.

pkg/security/ebpf/c/include/helpers/network/pid_resolver.h
```diff
@@ -43,13 +43,25 @@ __attribute__((always_inline)) void resolve_pid_from_flow_pid(struct packet_t *p
...
     }
+
+    // pid from socket cookie
+    if (pkt->pid == 0) {
+        u64 cookie = bpf_get_socket_cookie(skb);
+        u32 *pid = bpf_map_lookup_elem(&sock_cookie_pid, &cookie);
+        if (pid) {
+            pkt->pid = *pid;
+        }
+    }
```
YoannGh: What do you think about moving this resolver before the get_current_pid_tgid one?
I think get_current_pid_tgid is less reliable on the ingress path because it might not return a user process pid but the pid of the kworker running the tc hook instead

safchain: Yes, that's is probably better.

YoannGh (approved):

PR #40488, opened 2025-09-03
[CWS] Send original rule ID as tag when using group_id

lebauce:
### What does this PR do?

Add the original rule ID as a tag.

### Motivation

When using `group_id` for a rule, the original rule id is not sent in the event,
making it hard to match to the real agent rule.


pkg/security/module/server.go
```diff
@@ -414,6 +415,9 @@ func (a *APIServer) SendEvent(rule *rules.Rule, event events.Event, extTagsCb fu
 	eventTags := event.GetTags()
 
 	tags := []string{"rule_id:" + ruleID}
+	if ruleID != originalID {
+		tags = append(tags, "source_rule_id:"+originalID)
```
safchain: Is the tag enough or should it be part of the json as well ?

janine-c (approved):

safchain (approved):

PR #40490, opened 2025-09-03
[CWS-4569] Evict untouched nodes from profile after a given time

mftoure:
### What does this PR do?
This PR adds the mechanism that will evict nodes from profiles's activity tree if they have not been touched after a given time. It is part of the efforts to change the lifecycle of activity dumps and security profiles.


pkg/config/setup/system_probe_cws.go
```diff
@@ -62,7 +62,7 @@ func initCWSSystemProbeConfig(cfg pkgconfigmodel.Setup) {
 	cfg.BindEnvAndSetDefault("runtime_security_config.activity_dump.min_timeout", "10m")
 	cfg.BindEnvAndSetDefault("runtime_security_config.activity_dump.max_dump_size", 1750)
 	cfg.BindEnvAndSetDefault("runtime_security_config.activity_dump.traced_cgroups_count", 5)
-	cfg.BindEnvAndSetDefault("runtime_security_config.activity_dump.traced_event_types", []string{"exec", "open", "dns", "imds"})
+	cfg.BindEnvAndSetDefault("runtime_security_config.activity_dump.traced_event_types", []string{"exec", "open", "dns", "imds", "exit"})
```
safchain: I think if we have exec we should add by default exit

pkg/security/ebpf/c/include/hooks/exec.h
```diff
@@ -316,6 +316,15 @@ int __attribute__((always_inline)) handle_do_exit(ctx_t *ctx) {
             event.exit_code |= 0x80;
             bpf_map_delete_elem(&tasks_in_coredump, &pid_tgid);
         }
+
+        // Check if we should sample this event for activity dumps
+        struct activity_dump_config *config = lookup_or_delete_traced_pid(event.process.pid, bpf_ktime_get_ns(), NULL);
```
safchain: as all the exec are mark with the EVENT_FLAGS_ACTIVITY_DUMP_SAMPLE flag, I'll do the same for the exit, basically not checking the config

pkg/security/security_profile/activity_tree/process_node.go
```diff
@@ -550,3 +550,89 @@ func (pn *ProcessNode) EvictImageTag(imageTag string, DNSNames *utils.StringKeys
...
+		if socketNode.NodeBase.EvictBeforeTimestamp(before) > 0 {
+			if len(socketNode.Seen) == 0 {
+				pn.Sockets = append(pn.Sockets[:i], pn.Sockets[i+1:]...)
+				totalEvicted++
+			}
+		}
+	}
+
+	// Evict unused syscall nodes
+	if pn.Process.IsExecExec && !pn.Process.ExitTime.IsZero() {
```
safchain: Not sure to understand why we need to check the IsExecExec field

pkg/security/security_profile/activity_tree/process_node.go
```diff
@@ -550,3 +550,89 @@ func (pn *ProcessNode) EvictImageTag(imageTag string, DNSNames *utils.StringKeys
...
+	if !pn.Process.ExitTime.IsZero() {
+		evicted := pn.NodeBase.EvictBeforeTimestamp(before)
+		totalEvicted += evicted
+		if evicted > 0 && len(pn.Seen) == 0 {
+			totalEvicted++
+		}
+	}
+
+	// Evict unused file nodes
+	for path, fileNode := range pn.Files {
```
safchain: could we skip the rest of the Evictions if the process was evicted ?

pkg/security/security_profile/activity_tree/process_node.go
mftoure: sorry, autocompletion error

dustmop (approved):

pkg/security/config/config.go
```diff
@@ -635,6 +638,10 @@ func NewRuntimeSecurityConfig() (*RuntimeSecurityConfig, error) {
 		FileMetadataResolverEnabled: pkgconfigsetup.SystemProbe().GetBool("runtime_security_config.file_metadata_resolver.enabled"),
 	}
 
+	if slices.Contains(rsConfig.ActivityDumpTracedEventTypes, model.ExecEventType) {
+		rsConfig.ActivityDumpTracedEventTypes = append(rsConfig.ActivityDumpTracedEventTypes, model.ExitEventType)
+	}
```
spikat: Why should we add exit events here?

pkg/security/security_profile/activity_tree/activity_tree.go
```diff
@@ -958,3 +958,30 @@ func (at *ActivityTree) ExtractSyscalls(arch string) []string {
...
+func (at *ActivityTree) EvictUnusedNodes(before time.Time) int {
+	totalEvicted := 0
+
+	// Iterate through all process nodes and evict unused nodes
+	for i := len(at.ProcessNodes) - 1; i >= 0; i-- {
+		node := at.ProcessNodes[i]
+		if node == nil {
+			continue
+		}
+		// node.EvictBeforeTimestamp(before)
```
spikat: comment to be removed?

pkg/security/security_profile/activity_tree/activity_tree.go
```diff
@@ -958,3 +958,30 @@ func (at *ActivityTree) ExtractSyscalls(arch string) []string {
...
+		evicted := node.EvictUnusedNodes(before)
+		totalEvicted += evicted
+
+		// If the process node itself has no image tags left after eviction, remove it entirely
+		isRoot := len(at.ProcessNodes) == 1
+		if len(node.Seen) == 0 && !isRoot {
+			// Remove the node
+			at.ProcessNodes = append(at.ProcessNodes[:i], at.ProcessNodes[i+1:]...)
+			totalEvicted++
+		}
```
spikat, line 981: I'm not sure to understand the isRoot purpose here.

safchain (approved):

pkg/security/security_profile/manager.go
```diff
@@ -390,6 +391,14 @@ func (m *Manager) Start(ctx context.Context) {
 		silentWorkloadsTickerChan = make(chan time.Time)
 	}
 
+	if m.config.RuntimeSecurity.SecurityProfileEnabled && m.config.RuntimeSecurity.SecurityProfileNodeEvictionTimeout > 0 {
+		nodeEvictionTicker := time.NewTicker(m.config.RuntimeSecurity.SecurityProfileNodeEvictionTimeout)
```
spikat, line 395: I wonder if we need a different value here. Let's say we want to clean node older than 2h, we may want to proceed to the cleanup more frequently. WDYT ?

pkg/security/security_profile/manager.go
```diff
@@ -636,3 +647,32 @@ func perFormatStorageRequests(requests []config.StorageRequest) map[config.Stora
...
+	totalEvicted := 0
+
+	m.profilesLock.Lock()
+	defer m.profilesLock.Unlock()
+
+	for selector, profile := range m.profiles {
+		if profile == nil || profile.ActivityTree == nil {
+			continue
+		}
+
```
spikat, line 672: IMHO we should lock the profile here no?

pkg/security/security_profile/activity_tree/process_node.go
```diff
@@ -550,3 +550,89 @@ func (pn *ProcessNode) EvictImageTag(imageTag string, DNSNames *utils.StringKeys
...
 	return false
 }
+
+// EvictUnusedNodes evicts all child nodes that haven't been touched since the given timestamp
+// and returns the total number of nodes evicted
+func (pn *ProcessNode) EvictUnusedNodes(before time.Time) int {
+	totalEvicted := 0
+
+	// Only evict image tags from process node if the process has exited
+	if !pn.Process.ExitTime.IsZero() {
```
spikat: Not sure we should care about nodes with exit timestamps here. And if we want to I see some limitations:
- we'll need to clear this ts once we detect an already present exec
- we'll have a put a lot of logic when a policy is (re)loaded/saved to maintain consistency
Happy to chat about it offline

pkg/security/security_profile/activity_tree/process_node.go
```diff
@@ -550,3 +550,89 @@ func (pn *ProcessNode) EvictImageTag(imageTag string, DNSNames *utils.StringKeys
...
+		evicted := pn.NodeBase.EvictBeforeTimestamp(before)
+		totalEvicted += evicted
+		if evicted > 0 && len(pn.Seen) == 0 {
+			totalEvicted++
+			// we should return immediately because we don't need to evict other nodes if the process node itself should be removed
+			return totalEvicted
+		}
```
spikat: - no need to `totalEvicted++`, it's already done line 633.
- I wonder if instead of returning "only" totalEvicted we should also return the number of sub-nodes (dns, files etc). WDYT?

pkg/security/security_profile/activity_tree/process_node.go
```diff
@@ -550,3 +550,89 @@ func (pn *ProcessNode) EvictImageTag(imageTag string, DNSNames *utils.StringKeys
...
+
+	// Evict unused file nodes
+	for path, fileNode := range pn.Files {
+		if fileNode.NodeBase.EvictBeforeTimestamp(before) > 0 {
+			if len(fileNode.Seen) == 0 {
+				delete(pn.Files, path)
+				totalEvicted++
+			}
+		}
+	}
```
spikat: For every sub-node: I wonder if we should count the number of version evicted (not only if ALL the versions were evicted). I've no real opinion on this ATM, just curious about your thoughts :)

pkg/security/security_profile/activity_tree/process_node.go
spikat: Also I would make this check at the end of the function, and only perform it if all children were deleted. Otherwise we'll land on removing processes in the middle of a lineage with all its descendants.

pkg/security/security_profile/manager.go
```diff
@@ -636,3 +647,32 @@ func perFormatStorageRequests(requests []config.StorageRequest) map[config.Stora
...
+	evictionTime := time.Now().Add(-m.config.RuntimeSecurity.SecurityProfileNodeEvictionTimeout)
+	totalEvicted := 0
+
+	m.profilesLock.Lock()
+	defer m.profilesLock.Unlock()
+
+	for selector, profile := range m.profiles {
+		if profile == nil || profile.ActivityTree == nil {
+			continue
+		}
```
spikat, line 671: should we perform evictions only on loaded/active profiles?

pkg/security/tests/security_profile_test.go
```diff
@@ -2654,3 +2654,202 @@ func TestSecurityProfileSyscallDriftNoNewSyscall(t *testing.T) {
...
+				if len(nodes) != len(activities) {
+					t.Errorf("Expected %d process nodes found in profile, got %d", len(activities), len(nodes))
+					return false
+				}
+
+				return true
+			})
+
+		// Wait for eviction timeout + some buffer
+		time.Sleep(12 * time.Second)
```
spikat: just curious why we have to wait more than 5+1 sec here

pkg/security/config/config.go
mftoure: - we don't want to evict processes that are have not exited as they may make another action. 
- For that we check the exitTime which is not set if the exit event type is not traced

pkg/security/security_profile/activity_tree/activity_tree.go
mftoure: It's a sylvain's suggestion, It's a safeguard as we don't want to evict the root node because it would mean that all container processes are terminated

pkg/security/security_profile/manager.go
mftoure: what do you mean here ? the eviction period is settable on the config files.

pkg/security/security_profile/activity_tree/process_node.go
mftoure: same reply as above for the exit related question,

pkg/security/security_profile/manager.go
spikat: I mean we use the same config field `SecurityProfileNodeEvictionTimeout` for both:
- the ticker triggering the eviction
- the retention duration we want to keep nodes for

pkg/security/config/config.go
spikat: I though we decided to not rely on process exit right? Should we remove the exit from the traced event types?

mftoure: yep, forgot to remove it

pkg/security/security_profile/manager.go
spikat: I guess we could stay as this for now

pkg/security/security_profile/activity_tree/process_node.go
```diff
@@ -235,10 +237,15 @@ newSyscallLoop:
...
 		}
 		pn.Syscalls = append(pn.Syscalls, NewSyscallNode(int(newSyscall), e.ResolveEventTime(), imageTag, Runtime))
+		anySyscallNodeTagged = true
 		syscallMask[int(newSyscall)] = int(newSyscall)
 		stats.SyscallNodes++
 	}
 
+	if anySyscallNodeTagged {
+		pn.AppendImageTag(imageTag, e.ResolveEventTime())
+	}
```
spikat: If I understand correctly, we refresh the process tag anyway no? (except if the `e.Syscalls.Syscalls` list is empty but it should not happen I guess). Should we simplify this by directly frefreshing the process timestamp ?

spikat: In fact I saw that you refresh process tag timestamp for every eventype on multiple location, what about doing it only once ? like here: https://github.com/DataDog/datadog-agent/blob/5b1db21a790c0f4eee9e400e7f32f16944292020/pkg/security/security_profile/activity_tree/activity_tree.go#L397
ie: we don't really care if it's something new or not for the profile, in any case it's a valid activity related to a process that need it's tag to be update

pkg/security/security_profile/manager.go
```diff
@@ -636,3 +647,37 @@ func perFormatStorageRequests(requests []config.StorageRequest) map[config.Stora
...
+	m.profilesLock.Lock()
+	defer m.profilesLock.Unlock()
+
+	for selector, profile := range m.profiles {
+		if profile == nil {
+			continue
+		}
+		profile.Lock()
+		if profile.ActivityTree == nil {
+			continue
```
spikat, line 670: missing profile.unlock() here

pkg/security/security_profile/activity_tree/process_node.go
```diff
@@ -550,3 +581,119 @@ func (pn *ProcessNode) EvictImageTag(imageTag string, DNSNames *utils.StringKeys
...
+			pn.Children = append(pn.Children[:i], pn.Children[i+1:]...)
+			totalEvicted++
+		}
+	}
+
+	// Check if this node can be evicted (all children must be evictable)
+	canEvictThisNode := pn.canEvictNode(before)
+
+	// Only evict child nodes if this node itself can be evicted
+	if canEvictThisNode {
```
spikat: Why should we NOT evict process activity nodes (open, dns etc) only if the process itself can be evicted?
Also: as we refresh the process tag timestamp every time we insert an activity for it, if the process itself can be evicted, all of its node should be either already evicted or could be evited too.
IMHO we should:
* return directly if `pn.canEvictNode(before) == true`
* otherwise, try to evict all process activity nodes

pkg/security/security_profile/activity_tree/process_node.go
```diff
@@ -550,3 +581,119 @@ func (pn *ProcessNode) EvictImageTag(imageTag string, DNSNames *utils.StringKeys
...
+				if len(capabilityNode.Seen) == 0 {
+					pn.Capabilities = append(pn.Capabilities[:i], pn.Capabilities[i+1:]...)
+					totalEvicted++
+				}
+			}
+		}
+
+		// Finally, evict image tags from this process node
+		evicted := pn.NodeBase.EvictBeforeTimestamp(before)
+		totalEvicted += evicted
```
spikat: IMHO this should be almost first. It would simplify the `canEvictNode()` method by simply doing a:
`return len(pn.Children) == 0 && len(pn.Seen) == 0`
WDYT ?

spikat (approved):

PR #40491, opened 2025-09-03
[USMON-1477] USM Redis: Add conditional resource tracking support

guyarb:
### What does this PR do?

Implements conditional Redis resource (key) tracking in the eBPF monitoring system. When `service_monitoring_config.redis.track_resources` is disabled (default), the system uses optimized data structures that exclude key name buffers, significantly reducing memory usage.

### Motivation

Redis monitoring was previously always tracking resource names (keys), consuming ~270 bytes per in-flight transaction even when users didn't need key-level granularity. Since the cardinality of the keys is extremely high, and we didn't yet implement a quantization algorithm, we decided for now to make key monitoring as optional and disabled by default.

- **Memory Optimization**: `redis_transaction_t` reduced from ~270 bytes to ~20 bytes for in-flight storage
- **Dual Consumer Architecture**: Separate event consumers for minimal events vs keyed events based on configuration
- **Map Management**: Added conditional map sizing and cleanup for `redis_key_in_flight` map
- **Configuration**: Added runtime conditional logic using `is_redis_with_key_monitoring_enabled()` eBPF function

### Additional Notes

- **Memory Savings**: ~92% reduction in per-transaction memory usage by default (when `track_resources=false`)
- **Performance**: Key parsing completely skipped when not needed, reducing CPU overhead
- **Architecture**: Single probe function with runtime conditionals and separate event structures
- **Configuration**: New `service_monitoring_config.redis.track_resources` boolean flag (defaults to `false`)
- **Breaking Change**: Users who need key-level granularity must now explicitly set `track_resources: true`
- **Event Processing**: Dual consumer pattern handles both `redis` (minimal) and `redis_with_key` (full) event streams

### DIFF
--- a/pkg/network/ebpf/c/protocols/redis/types.h
+++ b/pkg/network/ebpf/c/protocols/redis/types.h
@@ -18,15 +18,19 @@ typedef enum {
     __MAX_REDIS_COMMAND
 } __attribute__ ((packed)) redis_command_t;
 
-// Redis in-flight transaction info
+// Represents redis key name.
 typedef struct {
     char buf[MAX_KEY_LEN];
+    __u16 len;
+    bool truncated;
+} redis_key_data_t;
+
+// Redis in-flight transaction info
+typedef struct {
     __u64 request_started;
     __u64 response_last_seen;
-    __u16 buf_len;
     redis_command_t command;
     __u8 tags;
-    bool truncated;
     bool is_error;
 } redis_transaction_t;
 
@@ -34,6 +38,7 @@ typedef struct {
 typedef struct {
     conn_tuple_t tuple;
     redis_transaction_t tx;
+    redis_key_data_t key;
 } redis_event_t;
 
 // Controls the number of Redis transactions read from userspace at a time.
--- a/pkg/network/ebpf/c/protocols/redis/types.h
+++ b/pkg/network/ebpf/c/protocols/redis/types.h
@@ -39,9 +39,9 @@ typedef struct {
     conn_tuple_t tuple;
     redis_transaction_t tx;
     redis_key_data_t key;
-} redis_event_t;
+} redis_with_key_event_t;
 
 // Controls the number of Redis transactions read from userspace at a time.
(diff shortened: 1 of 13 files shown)


pkg/network/ebpf/c/protocols/redis/types.h
```diff
@@ -18,22 +18,27 @@ typedef enum {
     __MAX_REDIS_COMMAND
 } __attribute__ ((packed)) redis_command_t;
 
-// Redis in-flight transaction info
+// Represents redis key name.
 typedef struct {
     char buf[MAX_KEY_LEN];
+    __u16 len;
```
amitslavin, line 24: shouldn't we keep it buf_len?

pkg/network/protocols/redis/protocol.go
```diff
@@ -171,6 +181,14 @@ func (p *protocol) DumpMaps(w io.Writer, mapName string, currentMap *ebpf.Map) {
 		for iter.Next(unsafe.Pointer(&key), unsafe.Pointer(&value)) {
 			spew.Fdump(w, key, value)
 		}
+	} else if mapName == keyInFlightMap {
```
amitslavin: consider using switch case instead?

amitslavin (changes requested):

amitslavin:
Can we run the load test to validate the change? I would also like to validate it once with key handling enabled and once without, in order to assess both the change itself and its impact

pkg/network/ebpf/c/protocols/redis/types.h
guyarb: we can, but now that we have a very compact struct, it seems like "redundant" naming

guyarb:
> Can we run the load test to validate the change? I would also like to validate it once with key handling enabled and once without, in order to assess both the change itself and its impact

sure, waiting for backend version to complete its deployment

guyarb:
WSS - 
image

cpu

image

amitslavin (approved):

PR #40498, opened 2025-09-03
Add agent specific gcc toolchains

JSGette:
### What does this PR do?
Introduces hermetic and downloadable gcc toolchains for Linux x86 and aarch64 for bazel.

### Motivation
To ensure compatibility with older versions of `libc` we are relying on cross-compilation toolchains built via [crosstool-ng ](https://github.com/crosstool-ng/crosstool-ng/). To take the next logical step we want to let bazel handle toolchains. We are reusing [gcc-toolchain](https://github.com/f0rmiga/gcc-toolchain) project that represents a fully hermetic toolchain that, however, requires certain adaptations to make it work with those created by `crosstool-ng`.

Used `toolchain_resolution_debug` flag and `--subcommands` to ensure we are calling the tools from our hermetic toolchains.


deps/repos.MODULE.bazel
```diff
@@ -25,10 +25,41 @@ http_archive = use_repo_rule("@bazel_tools//tools/build_defs/repo:http.bzl", "ht
 
 http_archive(
```
JSGette: Result of `buildifier`

bazel/patches/datadog_agent_toolchain.patch
```diff
@@ -0,0 +1,513 @@
...
+                         flags = sanitizer.cflags,
+diff --git a/toolchain/defs.bzl b/toolchain/defs.bzl
+index 00ef405..94d2c49 100644
+--- a/toolchain/defs.bzl
++++ b/toolchain/defs.bzl
+@@ -32,6 +32,7 @@ def _gcc_toolchain_impl(rctx):
+     absolute_toolchain_root = str(rctx.path("."))
+     execroot = paths.normalize(paths.join(absolute_toolchain_root, "..", ".."))
+     toolchain_root = paths.relativize(absolute_toolchain_root, execroot)
++    rctx.symlink("{}/{}-unknown-linux-gnu/bin".format(toolchain_root, rctx.attr.target_arch), "{}/xbin".format(toolchain_root))
```
rdesgroppes: I'm not a big fan of comments because they frequently paraphrase the next few lines of code (and later diverge from them), but I would love to have one here that explains the purpose of the symlink.

bazel/patches/datadog_agent_toolchain.patch
```diff
@@ -0,0 +1,513 @@
...
+     c_builtin_includes = [
+         include.format(
+@@ -71,7 +72,9 @@ def _gcc_toolchain_impl(rctx):
+         ] + ([
+             "%workspace%/{include_prefix}include",
+         ] if target_arch != ARCHS.x86_64 else []) + [
+-            "%workspace%/sysroot/usr/include",
++            #"%workspace%/sysroot/usr/include",
++        ] + [
++            "%workspace%/{include_prefix}sysroot/usr/include",
```
rdesgroppes:
```suggestion
+            "%workspace%/{include_prefix}sysroot/usr/include",
```

rdesgroppes (approved):
Looks very good to me:
- proper, comprehensive patch
- changes tailored to the very need
- in short: `bazel`-idiomatic

bazel/patches/datadog_agent_toolchain.patch
JSGette: That's not needed actually, removed

deps/repos.MODULE.bazel
rdesgroppes: rebasing the PR on top of `main` would shrink changes down - maybe with conflicts?

deps/repos.MODULE.bazel
```diff
@@ -25,10 +25,41 @@ http_archive = use_repo_rule("@bazel_tools//tools/build_defs/repo:http.bzl", "ht
...
+gcc_toolchains.toolchain(
+    name = "gcc_toolchain_aarch64",
+    binary_prefix = "aarch64-unknown-linux-gnu-",
+    gcc_version = "12.3.0",
+    target_arch = "aarch64",
+)
+use_repo(gcc_toolchains, "gcc_toolchain_x86_64")
+use_repo(gcc_toolchains, "gcc_toolchain_aarch64")
+
+register_toolchains(
```
JSGette, line 62: The only thing that I don't like about this is that we are unconditionally registering aarch64 toolchain that isn't a crosstool, it cannot be used on x86 system. There should be a way, though, to avoid that

JSGette: I want to benefit from bazel's laziness

rdesgroppes: To second this, from another perspective: contributors are likely interested in a small subset of available targets to build or test, but all of them are anyway declared. So, yes, the laziness of `bazel` is the feature we count on to prepare only needed dependencies (whether `dev_dependency = True` or `False`).

rdesgroppes: All that to say I would even remove the TODO...

rdesgroppes (approved):
Even greater than earlier!

bazel/patches/datadog_agent_toolchain.patch
```diff
@@ -0,0 +1,551 @@
...
+-    name = "_fortran_toolchain",
+-    cc_toolchain = ":_cc_toolchain",
+-)
+-
+ toolchain(
+     name = "cc_toolchain",
+     exec_compatible_with = [
+         "@platforms//os:linux",
+-        "@platforms//cpu:x86_64",
++        "@platforms//cpu:{target_arch}",
```
rdesgroppes, line 435: looks like this fixes a bug?

JSGette: not really, this is to ensure no one tries to do cross-compilation as `aarch64` toolchain won't work on x86 machine.

deps/repos.MODULE.bazel
JSGette: I would keep it, though. Yes, one extra toolchain is 100 Mb, but what if we add others (we will likely add more in the future). And now you have to pull something that you don't need and wait longer. Let TODO stay in case someone comes across and has a better idea

PR #40501, opened 2025-09-03
AGTMETRICS-362 add payloads builder v3

vickenty:
### What does this PR do?

Add prototype implementation of v3 intake protocol

Disabled by default, additional unit tests were added in #41406 to make sure existing functionality is unaffected.

### Additional Notes


pkg/config/setup/config.go
```diff
@@ -1483,6 +1483,7 @@ func serializer(config pkgconfigmodel.Setup) {
 	config.BindEnvAndSetDefault("serializer_use_events_marshaler_v2", true)
 
 	config.BindEnvAndSetDefault("use_v2_api.series", true)
+	config.BindEnvAndSetDefault("use_v3_api.series", false)
```
dustmop: Just curious, are these two flags mutually exclusive? Will a customer only ever be using v2 OR v3?

dustmop (approved):

pkg/config/setup/config.go
vickenty: In the end, when this is GA, most customers that only send to the intake would indeed pick one. In the context of this PR, I needed something to enable v3 to be able to test it. On a second thought, I should use something more placeholder-like here, to make sure we revisit this later.

pkg/serializer/internal/metrics/iterable_series_v3.go
```diff
@@ -0,0 +1,703 @@
...
+	"SourceTypeName",
+	"OriginInfo",
+}
+
+// Constants for type column
+const (
+	metricCount = 0x01
+	metricRate  = 0x02
+	metricGauge = 0x03
+	metricketch = 0x04
```
tobz: `metricketch` -> `metricSketch`

pkg/serializer/internal/metrics/iterable_series_v3.go
```diff
@@ -0,0 +1,703 @@
...
+	txn         *stream.ColumnTransaction
+	dict        *dictionaryBuilder
+
+	deltaName           deltaEncoder
+	deltaTags           deltaEncoder
+	deltaResources      deltaEncoder
+	deltaInterval       deltaEncoder
+	deltaTimestamp      deltaEncoder
+	deltaSourceTypeName deltaEncoder
+	deltaOriginInfo     deltaEncoder
```
tobz: I might add a suffix to the ones that are dealing with _indexes_, since at first I was confused on how we would delta encode a string. 😅 Something like `deltaNameDictIdx`, etc.

fabbing (approved):

pkg/serializer/internal/metrics/iterable_series_v3.go
```diff
@@ -0,0 +1,705 @@
...
+	panic("unknown APIMetricType")
+}
+
+func varintLen(v int) int {
+	n, rem := bits.Div(0, uint(bits.Len(uint(v))), 7)
+	if rem > 0 {
+		n++
+	}
+	return int(n)
+}
```
lukesteensen, line 701: Not sure if we'd actually hit this, but this returns a zero length for an input value of zero. I believe that would still need one byte to encode.

pkg/serializer/internal/metrics/iterable_series_v3.go
```diff
@@ -0,0 +1,705 @@
...
+
+func (db *dictionaryBuilder) appendTagsSlice(tags []string) {
+	for _, s := range tags {
+		db.tagsBuffer = append(db.tagsBuffer, db.tagsInterner.intern(istr(s)))
+	}
+}
+
+func (db *dictionaryBuilder) internTags1(prefixID int64, tags []string) int64 {
+	var hash1, hash2 uint64 = uint64(prefixID), 0
+	for _, s := range tags {
```
lukesteensen: Should we sort the input before hashing so it matches the serialized form?

pkg/serializer/internal/metrics/iterable_series_v3.go
```diff
@@ -0,0 +1,705 @@
...
+		hash1, hash2 = murmur3.SeedStringSum128(hash1, hash2, s)
+	}
+
+	key := hash1
+	if id, ok := db.tagsIndex[key]; ok {
+		return id
+	}
```
lukesteensen, line 576: Do we need to consider collisions here for correctness? I'm sure it's extremely unlikely, but I haven't done the math and we do process quite a lot of data 😅. If there's a way to have an equality-based fallback without costing too much, that might be a good idea.

pkg/serializer/internal/metrics/iterable_series_v3_test.go
```diff
@@ -0,0 +1,315 @@
...
+		15<<3 | 2, 4, 1, 1, 1, 1,
+		16<<3 | 2, 1, 8, 0xc2, 0xb8, 0xad, 0x8b, 0xd, 0, 0, 0,
+		17<<3 | 2, 1, 2, 4, 6,
+		19<<3 | 2, 1, 8, 0x1f, 0x85, 0xeb, 0x51, 0xb8, 0x1e, 0x9, 0x40,
+		23<<3 | 2, 1, 4, 0, 0, 2, 0,
+		24<<3 | 2, 1, 4, 2, 0, 2, 2,
+	}, ps[0].GetContent())
+}
+
+func BenchmarkPaylodsBuilderV3(b *testing.B) {
```
lukesteensen:
```suggestion
func BenchmarkPayloadsBuilderV3(b *testing.B) {
```

pkg/serializer/internal/metrics/iterable_series_v3_test.go
```diff
@@ -0,0 +1,315 @@
...
+	r.NoError(pb.writeSerie(series[2]))
+	r.NoError(pb.writeSerie(series[3]))
+	pb.finishPayload()
+	payloads := pb.transactionPayloads()
+	r.Len(payloads, 2)
+	r.Equal(3, payloads[0].GetPointCount())
+	r.Equal(9, payloads[1].GetPointCount())
+}
+
+func TestPaylodsBuilderV3_ReservedSpace(t *testing.T) {
```
lukesteensen:
```suggestion
func TestPayloadsBuilderV3_ReservedSpace(t *testing.T) {
```

pkg/serializer/internal/metrics/iterable_series_v3_test.go
```diff
@@ -0,0 +1,315 @@
...
+
+	pb, err := newPayloadsBuilderV3(500, 2_000, 10_000, noopimpl.New())
+	require.NoError(t, err)
+	for len(pb.payloads) == 0 {
+		require.NoError(t, pb.writeSerie(serie))
+	}
+	require.NoError(t, pb.finishPayload())
+}
+
+func TestPaylodsBuilderV3_Tags(t *testing.T) {
```
lukesteensen:
```suggestion
func TestPayloadsBuilderV3_Tags(t *testing.T) {
```

pkg/serializer/internal/stream/column.go
```diff
@@ -0,0 +1,233 @@
...
+func (cc *ColumnCompressor) AddItem(txn *ColumnTransaction) error {
+	if !cc.hasRoomForItem(txn) {
+		if cc.inputLength == 0 {
+			return ErrItemTooBig
+		}
+
+		err := cc.pack()
+		if err != nil {
+			return err
+		}
```
lukesteensen, line 161: I'm curious often we end up calling `pack` for a typical large payload and whether flushing the compressor more often has much impact on our final compression ratio. Given that we tend to see very high ratios, it seems like we would call it somewhat often but I have no idea whether that would have any meaningful impact. It seems like we keep a single compressor instance open the whole time so I imagine it can continue reusing dictionaries, etc.

lukesteensen (commented):
Added a few notes, but overall this looks very solid 👏

pkg/serializer/internal/metrics/iterable_series_v3.go
vickenty: Having an equality check would make me sleep better for sure. I had an earlier [version](https://github.com/DataDog/datadog-agent/blob/2d9d08e95ef4aa8dbe786294ad06a0adfbadff38/pkg/serializer/internal/metrics/iterable_series_v3.go#L577-L592) that used actual tags as the map key, and it had some measurable cost in cpu (each lookup need to do equality check) and allocations (to safely put a slice into the map it needs to be copied). In the end it seemed unnecessary complexity, since we already rely on hash uniqueness to deduplicate tags in the aggregator: if there are any collisions, they won't happen here.

vickenty: We could, I guess, use slice identity as the key with the contract that the slices should be made unique by the caller (if they want small payloads), that would be both safe and efficient, if more misuse-friendly API. Happy to discuss this more.

pkg/serializer/internal/metrics/iterable_series_v3.go
vickenty: Is there anything to gain from that? We shouldn't modify the slice without copying because it is owned by the aggregator and is on loan here, and could be accessed from other threads elsewhere.

pkg/serializer/internal/metrics/iterable_series_v3.go
vickenty: Good catch, fixed.

pkg/serializer/internal/metrics/iterable_series_v3.go
lukesteensen: The potential gain would be if we have slices of the same set of tags, but in different orders. Right now they'd get interned as different entries, which is just less space-efficient. Realistically it's probably a corner case that wouldn't have a significant impact.

pkg/serializer/internal/metrics/iterable_series_v3.go
lukesteensen: Good point that we already do essentially this same operation in the aggregator. From that perspective, I don't think we're adding any additional likelihood of a correctness issue, especially since here a collision would need to happen within a single payload. It does make me wonder if there's a good way to propagate and reuse the hashing we do for the aggregator's context key, but that's something we could explore later.

pkg/serializer/internal/metrics/iterable_series_v3.go
vickenty: We get tags sorted and deduplicated from the aggregator, so not sorting here wouldn't affect uniqueness of tags. But it seems to improve payload size slightly, and not too expensive, so might as well.

lukesteensen (approved):

goxberry (approved):

PR #40511, opened 2025-09-03
Update Datadog Agent Source List during MSI install

mrafi97:
### What does this PR do?
This PR adds a custom action at the end of a non fleet MSI install that updates the source list to use the path and package located in `C:\ProgramData\Datadog\Installer\packages\datadog-agent\stable`.

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1656

Fleet upgrades can fail when the MSI in `C:\Windows\Installer` or the MSI in the source list cannot be found due them being deleted. This PR should help prevent this as it updates the source list to use an MSI located in a Datadog directory.

Added e2e test that simulates a possible 1612 error environment and then downgrades the Agent.

Manually tested by doing an MSI install. We should see the following in the registry:
image

image


tools/windows/DatadogAgentInstaller/CustomActions/UpdateInstallSourceCustomAction.cs
```diff
@@ -0,0 +1,58 @@
...
+            var proc = session.RunCommand(installerPath, $"package-command datadog-agent updateRegistryInstallSource");
+
+            if (proc.ExitCode != 0)
+            {
+                session.Log($"install script exited with code: {proc.ExitCode}");
+                proc.Close();
+                return ActionResult.Failure;
+            }
+            proc.Close();
+            return ActionResult.Success;
```
julien-lebot: Use `using` scope block to automatically cleanup resources:
```suggestion
            using (var proc = session.RunCommand(installerPath, $"package-command datadog-agent updateRegistryInstallSource"))
            {
                if (proc.ExitCode != 0)
                {
                    session.Log($"install script exited with code: {proc.ExitCode}");
                    return ActionResult.Failure;
                }
                return ActionResult.Success;
            }
```

tools/windows/DatadogAgentInstaller/CustomActions/UpdateInstallSourceCustomAction.cs
```diff
@@ -0,0 +1,58 @@
...
+using Datadog.CustomActions.Native;
+using Datadog.CustomActions.Rollback;
+using Microsoft.Deployment.WindowsInstaller;
+using System;
+using System.Diagnostics;
+using System.IO;
+
+namespace Datadog.CustomActions
+{
+    public class UpdateInstallSourceCustomAction
```
julien-lebot, line 12: **✏️ suggestion:** ‏ We should have unit tests for all custom actions

pkg/fleet/installer/packages/datadog_agent_windows.go
```diff
@@ -798,6 +812,8 @@ func runDatadogAgentPackageCommand(ctx context.Context, command string) (err err
 		return preStopConfigExperimentDatadogAgentBackground(ctx)
 	case "postPromoteConfigExperimentBackground":
 		return postPromoteConfigExperimentDatadogAgentBackground(ctx)
+	case "updateRegistryInstallSource":
+		return updateRegistryInstallSource()
```
julien-lebot, line 820: **❓ question:** ‏ If this command is only called from the MSI, why is it in the installer and not the MSI directly?

mrafi97: I wanted to avoid adding additional Windows API calls to the MSI. I also thought a solution with Go would a be more manageable in the long run, for future cases where this command could be needed.

julien-lebot: Do we actually expect to call `updateRegistryInstallSource` anywhere outside MSI? Today the only caller I see is the MSI custom action. Given that, I’d prefer to keep the MSI API usage inside the MSI CA and drop the Go command. That reduces moving parts during install/repair/uninstall. If we do want reuse outside MSI (fleet remediation, post-install fixes), can we add a real non-MSI call-site or an e2e that exercises it? Otherwise it feels speculative.

clarkb7: The path to the MSI that we are adding to the source list only exists because the Go/fleet code is creating the `Installer/packages/datadog-agent/<version>/<msi>` path. Without that copy we have no need to update the source list. So makes sense to me to keep both sides in the Go code.

I think we should still take a look at WixSharp `ResilientPackage` though.

mrafi97: From my understanding the WixSharp ResilientPackage takes a path and will attempt to create a symbolic/hard link of the original MSI in that path. If that fails it will copy the MSI to the path. It will then update the SOURCELIST property with this new path.

This seems redundant to me though since we are already copying the MSI to `Installer/packages/datadog-agent/<version>/<msi>`, the install source just needs to be updated with its path and package name.

releasenotes/notes/fleet-automation-msi-update-install-source-edbfaccf24bece0f.yaml
```diff
@@ -0,0 +1,3 @@
+enhancements:
+  - |
+    When installing the Agent via MSI, the install source is now updated to the Installer stable package.
```
rtrieu:
```suggestion
    When installing the Agent through MSI, the install source is now updated to the Installer stable package.
```

rtrieu (approved):
Made one suggestion to replace "via" with "through" for translation purposes.

pkg/util/winutil/msi.go
```diff
@@ -0,0 +1,66 @@
...
+	procMsiSourceListForceResExW  = msi.NewProc("MsiSourceListForceResolutionExW")
+)
+
+// MsiSourceListAddSourceEx adds or reorders the set of sources of a patch or product in a specified context.
+// It can also create a source list for a patch that does not exist in the specified context.
+//
+// https://learn.microsoft.com/en-us/windows/win32/api/msi/nf-msi-msisourcelistaddsourceexw
+func MsiSourceListAddSourceEx(productCode string, context uint32, sourceType uint32, src string, index uint32) windows.Errno {
+	r1, _, _ := procMsiSourceListAddSourceExW.Call(
+		uintptr(unsafe.Pointer(windows.StringToUTF16Ptr(productCode))), // szProductCodeOrPatchCode
```
clarkb7: StringToUTF16Ptr is deprecated. Use UTF16PtrFromString instead.

pkg/util/winutil/msi.go
```diff
@@ -0,0 +1,66 @@
...
+	msi                           = windows.NewLazySystemDLL("msi.dll")
+	procMsiSourceListAddSourceExW = msi.NewProc("MsiSourceListAddSourceExW")
+	procMsiSourceListSetInfoW     = msi.NewProc("MsiSourceListSetInfoW")
+	procMsiSourceListForceResExW  = msi.NewProc("MsiSourceListForceResolutionExW")
+)
+
+// MsiSourceListAddSourceEx adds or reorders the set of sources of a patch or product in a specified context.
+// It can also create a source list for a patch that does not exist in the specified context.
+//
+// https://learn.microsoft.com/en-us/windows/win32/api/msi/nf-msi-msisourcelistaddsourceexw
```
clarkb7, line 27: thanks for adding msdn links!

tools/windows/DatadogAgentInstaller/CustomActions/UpdateInstallSourceCustomAction.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// <param name="s">The session object.</param>
+        /// <returns><see cref="ActionResult.Success"/></returns>
+        public static ActionResult UpdateInstallSource(ISession session)
+        {
+            var msiPath = session.Property("DATABASE");
+            // get installer path
+            var installerPath = session.Property("PROJECTLOCATION");
+            installerPath = Path.Combine(installerPath, "bin", "datadog-installer.exe");
+
+            // check if this is a fleet install
```
clarkb7, line 30:
```suggestion
            // check if this is a fleet install
            // fleet install runs the MSI from the package dir so the install source is already set
```

tools/windows/DatadogAgentInstaller/WixSetup/Datadog Agent/AgentCustomActions.cs
```diff
@@ -320,6 +322,21 @@ public AgentCustomActions()
                 .SetProperties(
                     "PROJECTLOCATION=[PROJECTLOCATION], FLEET_INSTALL=[FLEET_INSTALL], DATABASE=[DATABASE]");
 
+            UpdateInstallSource = new CustomAction<CustomActions>(
+                    new Id(nameof(UpdateInstallSource)),
+                    CustomActions.UpdateInstallSource,
+                    Return.check,
+                    When.Before,
+                    Step.InstallFinalize,
```
clarkb7, line 332:
```suggestion
                    // The built-in RegisterProduct action normally sets the install source,
                    // so our action must come after it to take effect.
                    When.Before,
                    Step.InstallFinalize,
```

test/new-e2e/tests/installer/windows/suites/agent-package/upgrade_test.go
```diff
@@ -779,3 +801,34 @@ func (s *testAgentUpgradeSuite) setExperimentMSIArgs(args []string) {
...
+			s.T().Logf("Removed MSI URL Install Source")
+		}
+
+	}
+
+	registryKey = "Registry::HKEY_CLASSES_ROOT\\Installer\\Products\\" + compressedProductCode + "\\SourceList\\Net"
+	err = windowscommon.SetTypedRegistryValue(s.Env().RemoteHost, registryKey, "2", "C:\\Windows\\FakePath", "ExpandString")
+	s.Require().NoError(err)
+	s.T().Logf("Set Fake MSI Install Source")
+
```
clarkb7, line 833: Why do we make these registry changes for this test?

pkg/fleet/installer/packages/datadog_agent_windows.go
```diff
@@ -782,6 +783,19 @@ func postPromoteConfigExperimentDatadogAgentBackground(ctx context.Context) erro
 	return nil
 }
 
+// updateRegistryInstallSource updates the install source and package name to the current stable MSI.
```
clarkb7, line 786:
```suggestion
// updateRegistryInstallSource updates the install source and package name to the current stable MSI.
//
// Called from the MSI to ensure the install source is our copy of the MSI and not the path run by the user.
// This helps ensure the MSI is available even when the original path is a temp dir, which is common
// with remote deployment scripts, or the Windows installer cache was removed for some reason.
```

clarkb7 (approved):

tools/windows/DatadogAgentInstaller/WixSetup/Datadog Agent/AgentCustomActions.cs
```diff
@@ -320,6 +322,21 @@ public AgentCustomActions()
                 .SetProperties(
                     "PROJECTLOCATION=[PROJECTLOCATION], FLEET_INSTALL=[FLEET_INSTALL], DATABASE=[DATABASE]");
 
+            UpdateInstallSource = new CustomAction<CustomActions>(
```
julien-lebot, line 325: We discussed this briefly offline - but I want to make sure it's captured here as well. What happens if the installer rolls back after modifying the source location?

test/new-e2e/tests/installer/windows/suites/agent-package/upgrade_test.go
mrafi97: My goal with the registry changes is to create a scenario where some registry sources are missing or pointing to directories that do not contain an MSI. Basically create a scenario where a 1612 could likely happen without removing the MSI in the Datadog Installer dir

tools/windows/DatadogAgentInstaller/WixSetup/Datadog Agent/AgentCustomActions.cs
mrafi97: I tested this and when a rollback occurs the registry keys are reverted to the same state as before the upgrade. 

I tested this by installing a current Agent version, and then upgrading with the new MSI and setting WIXFAILWHENDEFERRED=1. The install source registry keep their original values from before the rollback.

sblumenthal (approved):

CelianR (approved):

PR #40514, opened 2025-09-03
Deprecate python flag in integration subcommand

pgimalac:
### What does this PR do?
Deprecate the `python` flag of the `agent integration` command.

### Motivation
Simplify the code.
We don't support Python2 anymore.

CI

### Additional Notes
The flag still exists (so the command doesn't crash when used), but it won't show in the help message and a deprecation warning will be shown when used.


jeremy-hanna (approved):
👍

hestonhoffman (approved):

AAraKKe (approved):

PR #40516, opened 2025-09-03
feat(release): add release date parameter to the `release.finish` task

matt-dz:
### What does this PR do?

Adds a release date (format YYYY-MM-DD) parameter to the `release.finish` task. The date is added to the generated prelude yaml files. If no date is provided, the default is the current date.

Usage: `dda -- inv -e release.finish 7.69.x --release-date 2025-09-03`.

### Motivation

Currently, when running the release.finish task, users must manually edit generated files for future release dates. This process is cumbersome and prone to errors.

https://datadoghq.atlassian.net/browse/BARX-1163


sabrina-datadog (approved):

tasks/release.py
```diff
@@ -293,14 +293,34 @@ def tag_devel(ctx, release_branch, commit="HEAD", push=True, force=False):
 
 
 @task
-def finish(ctx, release_branch, upstream="origin"):
+def finish(ctx, release_branch, release_date=None, upstream="origin"):
```
FlorentClarret: nit - in general you should avoid adding parameters between two others. If someone relies on the order, either they're screwed or you'll have to update their call. In this case it does not change much but you you should add it at the end

tasks/release.py
```diff
@@ -293,14 +293,34 @@ def tag_devel(ctx, release_branch, commit="HEAD", push=True, force=False):
...
 
+    # Validate release_date (if provided)
+    try:
+        if release_date:
+            datetime.strptime(release_date, "%Y-%m-%d")
+    except ValueError as err:
+        raise Exit(
+            color_message(f"Invalid date `{release_date}`. Date should be valid and in format YYYY-MM-DD.", "red"),
+            code=1,
+        ) from err
```
FlorentClarret: It's not super relevant here, but I'd have the `try/except` block inside the `if` statement, it can't break if we do not reach the `strptime` call. I would put it inside the `if` and it's also easier to read because you know you should not care about it if you're not in the `if`

PR #40522, opened 2025-09-03
[CXP-2599] Add support for process entities in remote workloadmeta

daniel-taf:
### What does this PR do?
Adds support for process entities to be retrieved using remote workloadmeta

### Motivation
This will be needed by the process check in the process agent.

unit tests added in this PR, https://github.com/DataDog/datadog-agent/pull/40403 has e2e test coverage


pkg/proto/datadog/workloadmeta/workloadmeta.proto
```diff
@@ -175,11 +176,52 @@ message ECSTask {
...
+  string apmInstrumentation = 9;
+}
+
+message Language {
+  string name = 1;
+  string version = 2;
+}
+
+message Process {
+	WorkloadmetaEntityId entityId = 1;
```
wdhif: Seems like we have an issue here?

daniel-taf: what issue? I don't see it 😅

wdhif: Ah sorry, I meant the tabs are not the same as the other ones?

daniel-taf: Fixed

wdhif (approved):

gengnamstyle (approved):

PR #40523, opened 2025-09-03
CXP-2600 [agent] update wlm process entity logs

gengnamstyle:
### What does this PR do?
- adds additional process information to be printed in the `datadog-agent workload-list` output command for debugging

#### Default logging
- name (derived from `/proc/[pid]/status`, will be the main thread of the process)
- exe (path to executable)  

#### Verbose logging
- comm (similar to `name` but can be taken from any thread of the process)
- cwd (working dir)
- cmdline (cmdline can be long so this was initially designated to be verbose but since it's useful I can see the argument to put in default logging)
- uids (user ids)
- gids (group ids)

### Motivation
- the new process collector utilizes workload meta and logging this extra information can be helpful for debugging

- unit tests

### DIFF
--- a/comp/core/workloadmeta/def/types.go
+++ b/comp/core/workloadmeta/def/types.go
@@ -1545,6 +1545,7 @@ func (p Process) String(verbose bool) string {
 	_, _ = fmt.Fprintln(&sb, "PID:", p.EntityID.ID)
 	_, _ = fmt.Fprintln(&sb, "Name:", p.Name)
 	_, _ = fmt.Fprintln(&sb, "Exe:", p.Exe)
+	_, _ = fmt.Fprintln(&sb, "Cmdline:", sliceToString(p.Cmdline))
 	_, _ = fmt.Fprintln(&sb, "Namespace PID:", p.NsPid)
 	_, _ = fmt.Fprintln(&sb, "Container ID:", p.ContainerID)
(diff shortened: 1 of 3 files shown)


wdhif (approved):

buraizu (approved):

comp/core/workloadmeta/def/types.go
```diff
@@ -1543,13 +1543,25 @@ func (p Process) String(verbose bool) string {
...
 	_, _ = fmt.Fprintln(&sb, "Container ID:", p.ContainerID)
 	_, _ = fmt.Fprintln(&sb, "Creation time:", p.CreationTime)
 	if p.Language != nil {
 		_, _ = fmt.Fprintln(&sb, "Language:", p.Language.Name)
 	}
+
+	if verbose {
+		_, _ = fmt.Fprintln(&sb, "Comm:", p.Comm)
+		_, _ = fmt.Fprintln(&sb, "Cwd:", p.Cwd)
+		_, _ = fmt.Fprintln(&sb, "Cmdline:", sliceToString(p.Cmdline))
```
daniel-taf: I think Cmdline should likely be printed by default (without verbose) since its the most relevant for support escalations.

gengnamstyle: sure I'll move it to default

gengnamstyle: updated!

daniel-taf (approved):

PR #40529, opened 2025-09-03
Update Regression Detector lading version

blt:
### What does this PR do?

See release notes on https://github.com/DataDog/lading/pull/1485


goxberry (approved):

blt:
Ah, I believe there's a race here. Lading 0.28.0 is able to generate its containers much faster than previously. In the telemetry _from lading_ we see a container spike that will pass the `docker_container_*` check but the timer nature of the check means that the spike is missed.

sblumenthal (approved):

PR #40532, opened 2025-09-03
fix: [fa]stop using probe endpoint for connectivity checks

san-jos:
### What does this PR do?
Stop using GET /probe endpoints to perform connectivity checks.
Replaced by POST with empty payloads or removed.

### Motivation
Support question from a concerned customer
In the connectivity check, we use GET /probe endpoints. 
These endpoints are exposed by synthetics which is available depending on customer orgs setup and causes 403 when the endpoint is not available.

Manual QA


pkg/diagnose/connectivity/inventoryendpoint.go
```diff
@@ -117,7 +105,12 @@ func getAPIKey(cfg model.Reader, configPrefix string, defaultAPIKey string, altA
 	if !altAPIKey {
 		return defaultAPIKey
 	}
-	if apiKey := cfg.GetString(joinSuffix(configPrefix, ".") + "api_key"); apiKey != "" {
+
+	configAPIKey := joinSuffix(configPrefix, ".") + "api_key"
```
hush-hush, line 110: If the prefix is an empty string we would look for `.api_key` instead of `api_key`, no ?

pkg/diagnose/connectivity/inventoryendpoint.go
```diff
@@ -164,16 +157,12 @@ func getDomains(cfg model.Reader) []domain {
 
 func (e *endpointDescription) buildRoute(cfg model.Reader, domain domain) string {
 	baseURL := ""
-	if e.versioned {
-		baseURL, _ = utils.AddAgentVersionToDomain(domain.infraEndpoint, e.prefix)
+	urlOverrideKey := getURLOverrideKey(e.altURLOverrideKey, domain.isFailover)
+	schemedPrefix := fmt.Sprintf("https://%s", joinSuffix(e.prefix, "."))
```
hush-hush: Some domain seems to be configured without a prefix if I'm understanding the code correctly. If so we would looke for `http://.` as a schemaPrefix.

pkg/diagnose/connectivity/inventoryendpoint_test.go
```diff
@@ -88,16 +76,15 @@ func TestBuildRoute(t *testing.T) {
...
 				routePath:         tt.path,
-				versioned:         tt.versioned,
 				altURLOverrideKey: tt.urlOverrideKey,
 			}
 			version.AgentVersion = "6.0.0"
 			mockConfig := configmock.New(t)
-			mockConfig.SetWithoutSource("site", tt.domain.site)
-			mockConfig.SetWithoutSource("multi_region_failover.enabled", tt.domain.isFailover)
-			mockConfig.SetWithoutSource("multi_region_failover.site", tt.domain.site)
+			mockConfig.BindEnvAndSetDefault("site", tt.domain.site)
```
hush-hush: Please use `SetWithoutSource` when using the mock config. `BindEnvAndSetDefault` is meant to prepare the configuration to load YAML and env vars, we don't use them in our tests.

hush-hush (changes requested):

Hitsuji-M (approved):

hush-hush (approved):

PR #40539, opened 2025-09-03
dyninst: reduce memory usage for scraping and symbolicator

ajwerner:
This is fallout from #incident-42623. Along the way, we remove the need to parse the elf symbol table in order to load the gosym table. We also add some other optimizations to make it more efficient to search the gosym table by name. See individual commits.

#### dyninst/rcscrape: use gosym instead of elf symbols

```
name           old time/op    new time/op    delta
GenerateIR-14    4.07ms ±11%    1.95ms ± 7%  -52.00%  (p=0.000 n=10+10)

name           old alloc/op   new alloc/op   delta
GenerateIR-14    7.37MB ± 0%    0.01MB ± 0%  -99.82%  (p=0.000 n=8+8)

name           old allocs/op  new allocs/op  delta
GenerateIR-14     32.6k ± 0%      0.2k ± 0%  -99.40%  (p=0.000 n=10+10)
```

#### dyninst/object: add gosym parsing utilities

This code was getting repeated too much.

#### dyninst/gosym: remove dependency on GoVersion

Fixes https://datadoghq.atlassian.net/browse/DEBUG-4442.

#### dyninst/gosym/cli: adopt cobra, add list command

I wanted to look at the functions, this made it easier to do that.

#### dyninst/gosym: provide streaming iteration, fewer ptrs

It's nice to be able to iterate the functions without keeping
them all in RAM at the same time.

Along the way, I removed some allocations:

```
name                                      old time/op    new time/op    delta
ParseGoSymbolTable/ListFunctions-14          213µs ± 2%     141µs ± 2%  -34.00%  (p=0.000 n=10+10)
ParseGoSymbolTable/LocatePC-14               873ns ± 4%     763ns ± 3%  -12.66%  (p=0.000 n=9+10)

name                                      old alloc/op   new alloc/op   delta
ParseGoSymbolTable/ListFunctions-14          261kB ± 0%     203kB ± 0%  -22.22%  (p=0.002 n=8+10)
ParseGoSymbolTable/LocatePC-14                522B ± 0%      362B ± 0%  -30.70%  (p=0.000 n=10+10)

name                                      old allocs/op  new allocs/op  delta
ParseGoSymbolTable/ListFunctions-14          5.91k ± 0%     1.97k ± 0%  -66.72%  (p=0.000 n=10+10)
ParseGoSymbolTable/LocatePC-14                11.0 ± 0%       6.0 ± 0%  -45.45%  (p=0.000 n=10+10)
```

### DIFF
--- a/pkg/dyninst/gosym/symtab.go
+++ b/pkg/dyninst/gosym/symtab.go
@@ -16,6 +16,7 @@ import (
 	"fmt"
 	"iter"
 	"math"
+	"slices"
 
 	"github.com/DataDog/datadog-agent/pkg/dyninst/object"
 )
@@ -33,7 +34,7 @@ type GoFunction struct {
 	// The index of the function in the symbol table
 	idx uint32
 	// The internal function info
-	funcInfo *funcInfo
+	funcInfo funcInfo
 }
 
 // GoLocation represents a resolved source code location.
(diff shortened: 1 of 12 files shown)


pkg/dyninst/gosym/symtab.go
```diff
@@ -255,23 +267,14 @@ func parselineTable(data []byte, textRange, pcRange [2]uint64, goVersion *object
...
-	} else {
-		switch version {
-		case ver118:
-			wrapperFuncID = funcIDWrapperGo117ThroughGo121
-		case ver120:
-			wrapperFuncID = funcIDWrapperGo122Plus
-		default:
-			wrapperFuncID = defaultWrapperFuncID
-		}
+	var wrapperFuncID uint8
```
piob-io: I think it's gonna be misleading if the loop above fails to set correct wrapper id, and here we also set a wrong one. I'd remove this - so that if loop fails, we just don't mask any functions (not a horrible failure mode), rather than masking possible wrong one.

piob-io (approved):
Nice!

pkg/dyninst/gosym/symtab.go
ajwerner: Done.

ajwerner:
Thanks for the review!

PR #40542, opened 2025-09-03
[pkg/clusteragent/autoscaling] Load recommendation history from status and allow directional-only fallback

jennchenn:
### What does this PR do?

Fixes [CASCL-573](https://datadoghq.atlassian.net/browse/CASCL-573)
* Load recommendation history from new `LastRecommendations` field in CR 
* Allow local fallback to only be applied if it is for upscale / downscale / for both

### Motivation

Loading recommendation history allows us to more accurately determine the stabilized recommendation

Changes to fallback allow us to shorten the time to enable fallback by providing users the flexibility to set how they'd like fallback to be activated

1. Deploy these changes; set up DPA CR to set fallback direction
2. Verify that fallback is only enabled when the scaling direction matches the enabled direction
3. Restart the cluster agent - verify that DPAs in store is populated with data from the CR for recommendation history

### Additional Notes

[CASCL-573]: https://datadoghq.atlassian.net/browse/CASCL-573?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/clusteragent/autoscaling/workload/controller_horizontal.go
```diff
@@ -158,6 +158,14 @@ func (hr *horizontalController) computeScaleAction(
 		return nil, 0, errors.New("scaling disabled as current replicas is set to 0")
 	}
 
+	// Check if we are in fallback mode and scaling direction is disabled
```
vboulineau: The actual scale direction is computed later (line 231).

pkg/clusteragent/autoscaling/workload/model/pod_autoscaler.go
```diff
@@ -309,12 +309,11 @@ func (p *PodAutoscalerInternal) UpdateFromStatus(status *datadoghqcommon.Datadog
 		if len(status.Horizontal.LastActions) > 0 {
 			p.horizontalLastActions = status.Horizontal.LastActions
 
-			// TODO: Store the recommendations history as well, the last actions miss data
-			for _, recommendation := range status.Horizontal.LastActions {
-				if recommendation.RecommendedReplicas != nil {
-					p.horizontalLastRecommendations = addRecommendationToHistory(recommendation.Time.Time, p.horizontalRecommendationsRetention, p.horizontalLastRecommendations, HorizontalScalingValues{
-						Timestamp: recommendation.Time.Time,
-						Replicas:  *recommendation.RecommendedReplicas,
+			for _, recommendation := range status.Horizontal.LastRecommendations {
```
vboulineau, line 312: Needs to be out as technically not related to `len(status.Horizontal.LastActions) > 0`

vboulineau (approved):
Pending one change but LGTM otherwise

CelianR (approved):

pkg/clusteragent/autoscaling/workload/model/pod_autoscaler.go
```diff
@@ -34,6 +36,9 @@ const (
 	// statusRetainedActions is the number of horizontal actions kept in status
 	statusRetainedActions = 10
 
+	// statusRetainedRecommendations is the maximum number of horizontal recommendations kept in status
+	statusRetainedRecommendations = 20
```
vboulineau: Might be a bit low, it would mean we cover only 10 minutes, we should probably set that to 60 to accomodate 30 mins

pkg/clusteragent/autoscaling/workload/controller_horizontal.go
```diff
@@ -222,6 +222,15 @@ func (hr *horizontalController) computeScaleAction(
 	// Now that we have applied all modifications to targetDesiredReplicas, we can compute the scale direction
 	scaleDirection := common.GetScaleDirection(currentDesiredReplicas, targetDesiredReplicas)
 
+	// Check if we are in fallback mode and scaling direction is disabled
```
vboulineau, line 225: nit: Technically you only need to check that if `scaleDirection != NoScale`, as otherwise there's nothing to do and scaling is not actually disabled.

vboulineau (approved):

PR #40551, opened 2025-09-04
Introduce config-sync to deliver resolved secrets to system-probe

mbertrone:
### What does this PR do?
Introduce config-sync in System Probe, to allow for resolved secrets in `datadog.yaml` to be resolved and delivered to SP.

### Motivation
When we'll introduce direct communication from SP to our BE, resolving secrets using the secret_backend_command could be a security concern. Hence we are relying on config-sync mechanism already in place, but never introduced in SP.

Manual Testing

### DIFF
--- a/cmd/system-probe/subcommands/run/command.go
+++ b/cmd/system-probe/subcommands/run/command.go
@@ -29,6 +29,7 @@ import (
 	"github.com/DataDog/datadog-agent/comp/agent/autoexit"
 	"github.com/DataDog/datadog-agent/comp/agent/autoexit/autoexitimpl"
 	"github.com/DataDog/datadog-agent/comp/core/config"
+	configsyncimpl "github.com/DataDog/datadog-agent/comp/core/configsync/configsyncimpl"
 	healthprobe "github.com/DataDog/datadog-agent/comp/core/healthprobe/def"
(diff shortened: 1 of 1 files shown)


cmd/system-probe/subcommands/run/command.go
```diff
@@ -151,6 +152,7 @@ func Commands(globalParams *command.GlobalParams) []*cobra.Command {
 					}
 				}),
 				settingsimpl.Module(),
+				configsyncimpl.Module(configsyncimpl.NewDefaultParams()),
```
paulcacheux, line 155: From my testing in https://github.com/DataDog/datadog-agent/pull/40485, you actually need to enable the init sync config otherwise the system-probe loads the first config from file and then the business code may load the non yet synced config during init time.

mbertrone: You are right, without the `syncOnInit=true`, the initial configuration will not contain the resolved keys, hence those will be fetched/resolved only at the next sync. This may not be ideal, even if the next sync happen in a few secs and cws starts sending data.

On the other side, this will force SP to shutdown, if no running agent is found (but only if the `agent_ipc` is set in the config)

PR #40559, opened 2025-09-04
process_log: Add more source matches

vitkyrka:
Add explict source matching for some more common cases.

https://datadoghq.atlassian.net/wiki/spaces/DSCVR/pages/5342921002/

### DIFF
--- a/comp/core/autodiscovery/providers/process_log.go
+++ b/comp/core/autodiscovery/providers/process_log.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 	"sync"
 	"unicode/utf8"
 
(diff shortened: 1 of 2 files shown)


Yumasi (approved):

gh123man (commented):
I think this change deserves a changelog entry as IMO this would be considered an "enhancement". 

I think it would also be helpful to add a little more context in the PR description about what scenarios this impacts (for readers not familiar with the change)

vitkyrka:
Sorry, missed this comment.

> I think this change deserves a changelog entry as IMO this would be considered an "enhancement".

Possibly, but note that the process_log feature is not documented anywhere yet (and was in itself not announced in the changelog).

> I think it would also be helpful to add a little more context in the PR description about what scenarios this impacts (for readers not familiar with the change)

Agreed.

PR #40563, opened 2025-09-04
[corechecks/kubelet] Use workloadmeta as source for pods

davidor:
### What does this PR do?

This PR updates the Kubelet check to use workloadmeta instead of querying the Kubelet directly for pod info.
This only changes the pod provider of the check. The rest are already using workloadmeta or query the Kubelet to fetch info that's not pod related and it's not stored in workloadmeta.

This change is part of the effort to make workloadmeta the single source of truth for pod data. Benefits include:
- Consistent data across Agent components.
- Code easier to maintain and reason about.
- Improved performance by avoiding redundant data and extra calls to the Kubelet.

This is similar to the migration that was done for the AD prometheus provider (https://github.com/DataDog/datadog-agent/pull/39238) and the KSM check (https://github.com/DataDog/datadog-agent/pull/39657).

Relying mostly on existing tests plus manual checks to confirm the Kubelet check emits the same metrics.


pkg/collector/corechecks/containers/kubelet/testdata/pods_requests_limits.json
```diff
@@ -119,7 +119,7 @@
               },
               "requests":{
                 "cpu":"500m",
-                "ephemeral-storage":"500m",
+                "ephemeral-storage":"500Mi",
```
davidor, line 122: Note for reviewers: I changed this to make it more realistic, as `500m` for memory means 0.5 bytes.

L3n41c (approved):

davidor:
Rebased on top of main to fix conflicts

gabedos (approved):

PR #40564, opened 2025-09-04
[NDM]move ha disabled metadata log

jedupau:
### What does this PR do?

this pr moves an ha log at init of the component to avoid spamming customer logs
https://dd.slack.com/archives/C07RHJTEFRD/p1756979953567099

### DIFF
--- a/comp/metadata/haagent/impl/haagent.go
+++ b/comp/metadata/haagent/impl/haagent.go
@@ -52,7 +52,7 @@ func NewComponent(reqs Requires) (Provides, error) {
 	}
 
 	if !i.haAgent.Enabled() {
-		i.log.Infof("HA Agent Metadata unavailable as HA Agent is disabled")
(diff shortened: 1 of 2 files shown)


AlexandreYang (approved):

comp/metadata/haagent/impl/haagent.go
```diff
@@ -51,6 +51,10 @@ func NewComponent(reqs Requires) (Provides, error) {
 		haAgent:  reqs.HaAgent,
 	}
 
+	if !i.haAgent.Enabled() {
+		i.log.Infof("HA Agent Metadata unavailable as HA Agent is disabled")
```
AlexandreYang: I think we can even change it to debug log.

jedupau: eb2f78e27dc781a5064d5fba9cf1a9e235ff37f5

PR #40565, opened 2025-09-04
Minify javascript files

JSGette:
### What does this PR do?
Simply adds minified version of several javascript files and replaces originals in deliverables
to optimize their size.


louis-cqrl (approved):

comp/core/gui/guiimpl/views/private/js/codemirror.min.js
pgimalac, line 1: Should those files be created on the fly when building the agent (and not checked in git) ?

JSGette: Looking at original files they are pretty static, so why doing that each time if we can just commit them once?

pgimalac, line 1: Minifying removed the copyright in those files, I don't think we can do that

pgimalac: If they are fully static we should remove the existing files and keep only the minified version
Otherwise it's a big footgun

JSGette: Good point, added from the originals

JSGette: I would ask the owners what they think. If they say we should remove the originals - sure.

hush-hush: I agree that we need to generate them at build time. Maintaining minify JS is impossible and the next person that will modify the original file will not regenerate the min file.

JSGette: But we already have minified files (see other `js` files in this directory) baked into the repository

pgimalac: I think for library files that we don't edit it's not an issue, `javascript.js` however is the one that we edit

JSGette: If we need to bring the size down immediately I would propose to merge this as is. I will handle `javascript.js` on Monday

JSGette: I am replacing 3d party scripts with minified versions and leaving `javascript.js` out for now.

tasks/agent.py
```diff
@@ -352,7 +352,12 @@ def refresh_assets(_, build_tags, development=True, flavor=AgentFlavor.base.name
             os.path.join(dist_folder, "conf.d/process_agent.yaml.default"),
         )
 
-    shutil.copytree("./comp/core/gui/guiimpl/views/private", os.path.join(dist_folder, "views"), dirs_exist_ok=True)
+    # We only copy the minified versions of the javascript files to avoid having original javascript files in the distribution
```
pgimalac: I guess this comment can be removed

pgimalac (approved):

tasks/agent.py
JSGette: Reverted

hush-hush (approved):

PR #40569, opened 2025-09-04
(fleet) update user configuration

arbll:
This PR adds initial support to edit the configuration of the user (/etc/datadog-agent) instead of the fleet-specific layer (/etc/datadog-agent/managed).

It is backward compatible and hides this new mechanism behind a flag controlled at the level of the task sent by the backend.


pkg/fleet/installer/config/file.go
```diff
@@ -0,0 +1,64 @@
...
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+)
+
+// copyDirectory copies a directory from source to target.
+// It preserves the directory structure and file permissions.
+func copyDirectory(sourcePath, targetPath string) error {
+	return filepath.Walk(sourcePath, func(path string, info fs.FileInfo, err error) error {
```
coignetp: **🥜 nitpick:** ‏Let's make sure we don't provide a symlink as a sourcePath

pkg/fleet/installer/packages/embedded/templates/datadog-agent.service.tmpl
```diff
@@ -28,7 +28,7 @@ Restart=on-failure
 ExecStart={{.InstallDir}}/bin/agent/agent run -p {{.InstallDir}}/run/agent.pid
 {{- else}}
 Restart=no
-ExecStart=/usr/bin/timeout --kill-after=15s 3000s {{.InstallDir}}/bin/agent/agent run -p {{.InstallDir}}/run/agent.pid
+ExecStart=/usr/bin/timeout --kill-after=15s 3000s {{.InstallDir}}/bin/agent/agent run -c {{.EtcDir}} -p {{.InstallDir}}/run/agent.pid
```
coignetp, line 31: **❓ question:** ‏Isn't `-c` a config path and not a /etc dir path?

coignetp (approved):

pkg/fleet/daemon/daemon.go
```diff
@@ -70,9 +70,9 @@ type Daemon interface {
 	StartExperiment(ctx context.Context, url string) error
 	StopExperiment(ctx context.Context, pkg string) error
 	PromoteExperiment(ctx context.Context, pkg string) error
-	StartConfigExperiment(ctx context.Context, pkg string, operations config.Operations) error
-	StopConfigExperiment(ctx context.Context, pkg string) error
-	PromoteConfigExperiment(ctx context.Context, pkg string) error
+	StartConfigExperiment(ctx context.Context, pkg string, operations config.Operations, useLegacyFleetDir bool) error
+	StopConfigExperiment(ctx context.Context, pkg string, useLegacyFleetDir bool) error
+	PromoteConfigExperiment(ctx context.Context, pkg string, useLegacyFleetDir bool) error
```
BaptisteFoy: Feels weird to default to the new setup when upgrading locally but defaulting to legacy while remote. Do you have context on why we do that?

pkg/fleet/installer/packages/embedded/templates/datadog-agent.service.tmpl
BaptisteFoy: (answering to Paul) No: `agentCmd.PersistentFlags().StringVarP(&globalParams.ConfFilePath, "cfgpath", "c", "", "path to directory containing datadog.yaml")`

pkg/fleet/installer/packages/embedded/templates/main.go
```diff
@@ -135,7 +135,7 @@ var (
 	}
 	expDataDebRpm = systemdTemplateData{
 		InstallDir:       "/opt/datadog-agent",
-		EtcDir:           "/etc/datadog-agent",
+		EtcDir:           "/etc/datadog-agent-exp",
```
BaptisteFoy, line 138: ⚠️ Is this directory there by default? If it's not there by default & at least symlinked to /etc/datadog-agent then agent upgrades & legacy agent config will break.

pkg/fleet/installer/config/file.go
```diff
@@ -0,0 +1,64 @@
...
+	err = os.Rename(oldDir, filepath.Join(oldDir, ".bak"))
+	if err != nil {
+		return fmt.Errorf("could not rename old directory: %w", err)
+	}
+	defer func() {
+		if err != nil {
+			rollbackErr := os.Rename(filepath.Join(oldDir, ".bak"), oldDir)
+			if rollbackErr != nil {
+				err = fmt.Errorf("%w, rollback error: %w", err, rollbackErr)
+			}
```
BaptisteFoy, line 26: This means the .bak still exists. Do we want that?

arbll: yeah I think so, better keep the old config on disk at the minimum

PR #40570, opened 2025-09-04
Add support to take actions on Argo Rollouts objects in Autoscaling

vboulineau:
### What does this PR do?

Add support to take actions on `Rollout` objects in Autoscaling. There's not a lot of changes as mostly the Cluster Agent supports any object, _except_ for vertical, which what this PR addresses mostly.

### Motivation

Support Argo Rollouts

Currently you need custom code to "fake" feed the Autoscaling by updating the `postProcess` function in `pkg/clusteragent/autoscaling/workload/config_retriever_values.go` to inject fake data for Argo Rollout.
Once the backend has proper support it can be easily validated end-to-end by creating a `DPA` with a target to an Argo Rollout:
```
  kind: DatadogPodAutoscaler
  spec:
    owner: Local
    targetRef:
      apiVersion: argoproj.io/v1alpha1
      kind: Rollout
      name: rollout-canary
```

### Additional Notes

The `StrategicMergePatchType` is not supported for CRs, but for what we are doing (updating a `map`), there's no difference between `StrategicMergePatchType` and `MergePatchType (JSON Merge Patch)`

### DIFF
--- a/pkg/clusteragent/autoscaling/workload/controller.go
+++ b/pkg/clusteragent/autoscaling/workload/controller.go
@@ -454,10 +454,13 @@ func (c *Controller) validateAutoscaler(podAutoscalerInternal model.PodAutoscale
 
 	var resourceName string
 	switch owner := podAutoscalerInternal.Spec().TargetRef.Kind; owner {
-	case "Deployment":
+	case kubernetes.DeploymentKind:
 		resourceName = kubernetes.ParseDeploymentForPodName(clusterAgentPodName)
-	case "ReplicaSet":
+	case kubernetes.ReplicaSetKind:
 		resourceName = kubernetes.ParseReplicaSetForPodName(clusterAgentPodName)
+	default:
+		// We don't support other ways to deploy the Cluster Agent
+		return nil
 	}
 
 	clusterAgentNs := common.GetMyNamespace()
--- a/pkg/clusteragent/autoscaling/workload/pod_watcher.go
+++ b/pkg/clusteragent/autoscaling/workload/pod_watcher.go
@@ -147,7 +147,7 @@ func (pw *PodWatcherImpl) HandleEvent(event workloadmeta.Event) {
 }
 
 func (pw *PodWatcherImpl) handleSetEvent(pod *workloadmeta.KubernetesPod) {
-	podOwner, err := getNamespacedPodOwner(pod.Namespace, &pod.Owners[0])
+	podOwner, err := resolveNamespacedPodOwner(pod)
 	if err != nil {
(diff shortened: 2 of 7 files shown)


pkg/clusteragent/autoscaling/workload/controller.go
```diff
@@ -454,10 +454,13 @@ func (c *Controller) validateAutoscaler(podAutoscalerInternal model.PodAutoscale
...
 	switch owner := podAutoscalerInternal.Spec().TargetRef.Kind; owner {
-	case "Deployment":
+	case kubernetes.DeploymentKind:
 		resourceName = kubernetes.ParseDeploymentForPodName(clusterAgentPodName)
-	case "ReplicaSet":
+	case kubernetes.ReplicaSetKind:
 		resourceName = kubernetes.ParseReplicaSetForPodName(clusterAgentPodName)
+	default:
+		// We don't support other ways to deploy the Cluster Agent
+		return nil
```
jbartosik, line 463: Shouldn't we return an error?

pkg/clusteragent/autoscaling/workload/pod_watcher.go
```diff
@@ -220,22 +220,35 @@ func (pw *PodWatcherImpl) runPatcher(ctx context.Context) {
 	}
 }
 
-func getNamespacedPodOwner(ns string, owner *workloadmeta.KubernetesPodOwner) (NamespacedPodOwner, error) {
-	if owner.Kind == kubernetes.DeploymentKind {
+func resolveNamespacedPodOwner(pod *workloadmeta.KubernetesPod) (NamespacedPodOwner, error) {
+	if len(pod.Owners) == 0 || pod.Owners[0].Kind == kubernetes.DeploymentKind {
 		return NamespacedPodOwner{}, errDeploymentNotValidOwner
 	}
```
jbartosik, line 226: Don't we need to check for direct ownership, like with deployment?

pkg/clusteragent/autoscaling/workload/controller.go
vboulineau: No it's just the check to validate if the DCA itself is targeted by Autoscaling, we don't expect any other owner.

pkg/clusteragent/autoscaling/workload/pod_watcher.go
vboulineau: In theory yes, although being a 3rd party, we have no strong guarantee that things may not change in the future.
It's rare enough already for `Deployment` that it should be fine for Argo Rollout.

sblumenthal (approved):

PR #40582, opened 2025-09-04
[AGENTRUN] Fix NewHTTPSClient

misteriaud:
### What does this PR do?
This PR:
- rename function `NewHTTPSClient` to `NewSecureClient` in order to avoid confusion with `NewHTTPClient`.
- make use of the right `RequestOption` when using `NewSecureClient`

### Motivation
`NewHTTPSClient` was not using the provided `RequestOption`, which can lead to resources leak.

Added a unit test to assert the provided options are used during HTTP calls


paulcacheux (approved):

pkg/config/settings/http/client_test.go
```diff
@@ -0,0 +1,73 @@
...
+		args          args
+		shouldTimeout bool
+	}{
+		{
+			name: "without option [should work]",
+			args: args{
+				c:                 ipcComp.GetClient(),
+				baseURL:           srv.URL,
+				targetProcessName: "test-process",
+			},
```
hush-hush, line 47: Nit-pick but setting `shouldTimeout` to false manually make for more readable code than relying on implicit behavior.

hush-hush (approved):

pkg/config/settings/http/client_test.go
misteriaud: Done in 0ed9abe218e380eedf36eb7316ed7dc98c2687f3.

gabedos (approved):

Alos222 (approved):

PR #40586, opened 2025-09-04
feat(review): Automatically call for review when a PR becomes ready_for_review or review_requested

chouetz:
### What does this PR do?
- Move the `add_review` workflow to a dedicated file
- Trigger it on label addition or "pr_ready_for_review" or "reviewer_addition"
- Remove the condition from the task which would trigger only if the label is there.

### Motivation
More automation in the review process as discussed in [Tony's document](https://docs.google.com/document/d/1yquFoqOfDn_dqmSg3QKApdg7fFnlHPtYN1ZlKy-KYa4/edit?usp=sharing)
Also will enable auto-review for external contributors if thy convert their PR like this.
https://datadoghq.atlassian.net/browse/ACIX-1181

On the PR

### DIFF
--- a/.github/workflows/ask_review.yml
+++ b/.github/workflows/ask_review.yml
@@ -1,0 +1,27 @@
+---
+name: "Ask for code reviews"
+
+on:
+  pull_request:
+    types: [labeled, ready_for_review, review_requested]
+
+permissions: {}
+
+jobs:
+  ask-reviews:
+    if: github.triggering_actor != 'dd-devflow[bot]' && (github.event.action != 'labeled' || github.event.label.name == 'ask-review')
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          persist-credentials: false
+      - name: Install dda
+        uses: ./.github/actions/install-dda
+        with:
+          features: legacy-tasks
+      - name: Ask for code reviews
+        env:
+          SLACK_DATADOG_AGENT_BOT_TOKEN : ${{ secrets.SLACK_DATADOG_AGENT_BOT_TOKEN }}
+          PR_REQUESTED_TEAMS: ${{ toJSON(github.event.pull_request.requested_teams) }}
+        run: dda inv -- -e issue.ask-reviews -p ${{ github.event.pull_request.number }}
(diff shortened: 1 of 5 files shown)


.github/workflows/ask_review.yml
```diff
@@ -0,0 +1,26 @@
+---
+name: "Ask for code reviews"
+
+on:
+  pull_request:
+    types: [labeled, ready_for_review, review_requested]
```
chouetz, line 6: This is what changes

Ishirui (approved):

.github/workflows/ask_review.yml
```diff
@@ -0,0 +1,27 @@
...
+
+permissions: {}
+
+jobs:
+  ask-reviews:
+    if: github.triggering_actor != 'dd-devflow[bot]' && (github.event.action == 'labeled' && github.event.label.name == 'ask-review' || github.event.action != 'labeled')
```
pgimalac: I think this is equivalent ?
```suggestion
    if: github.triggering_actor != 'dd-devflow[bot]' && (github.event.action != 'labeled' || github.event.label.name == 'ask-review')
```

.github/workflows/ask_review.yml
```diff
@@ -0,0 +1,27 @@
+---
+name: "Ask for code reviews"
+
+on:
+  pull_request:
+    types: [labeled, ready_for_review, review_requested]
```
pgimalac, line 6: With `review_requested`, will it trigger twice if I add one reviewer, then later another ?
If so it could be quite spammy for the first team

chouetz: I've just updated the workflow so that it triggers only the added requested reviewer in this situation:
- We can have only one requested team by github event
- If I take the events in reverse order of creation and stop at the first, I'm ok

pgimalac: Ah yes if the event is associated to a single team then it should be fine !

KevinFairise2:
Should we have a label when we explicitly do not want the automatic message to be sent? I am not sure it would be useful but I feel that with that change we are going to send a lot more automatic messages to teams.
For instance sometimes when there are mutltiple code owner on a single file you do not always send the message to all of the code owners and only the one you feel is the more appropriate. It will no longer be possible after that change

chouetz:
> Should we have a label when we explicitly do not want the automatic message to be sent? I am not sure it would be useful but I feel that with that change we are going to send a lot more automatic messages to teams. For instance sometimes when there are mutltiple code owner on a single file you do not always send the message to all of the code owners and only the one you feel is the more appropriate. It will no longer be possible after that change

Yes if you think it's necessary I can add this to the logic of the test

rdesgroppes (approved):

PR #40589, opened 2025-09-04
dyninst/decode: encode ptrs to 0-sized types correctly

ajwerner:
### What does this PR do?

We weren't doing the right thing with pointers to 0-sized types.

### Motivation

I noticed it looked bad.

Added tests.


piob-io (approved):
Thanks!

PR #40591, opened 2025-09-04
[SLES-2438] Add a log file descriptor for AAS instance logging

jcstorms1:
### What does this PR do?
This PR adds a new environment variable `DD_AAS_INSTANCE_LOG_FILE_DESCRIPTOR` to narrow down the amount of tailed files when using the AAS sidecar with instance logging enabled.

### Motivation

When instance logging is enabled we tail all the files in path. This led to many open files in high load environments, which caused undesirable behavior. By adding the descriptor option we can allow users to focus on the files they care about most and limit issues with Azure's log rotation. Additionally, this is ideal for users who only want to tail the default application log `filename_default_docker.log` used by Azure App Services for application logging.

Unit Test and in an application

### DIFF
--- a/cmd/serverless-init/log/log.go
+++ b/cmd/serverless-init/log/log.go
@@ -27,7 +27,7 @@ import (
 const (
 	defaultFlushTimeout   = 5 * time.Second
 	defaultTailingPath    = "/home/LogFiles/*$COMPUTERNAME*.log"
-	modifiableTailingPath = "/home/LogFiles/*$COMPUTERNAME*.%s.log"
+	modifiableTailingPath = "/home/LogFiles/*$COMPUTERNAME*_%s.log"
 	logEnabledEnvVar      = "DD_LOGS_ENABLED"
 	envVarTailFilePath    = "DD_SERVERLESS_LOG_PATH"
 	aasInstanceTailing    = "DD_AAS_INSTANCE_LOGGING_ENABLED"
--- a/cmd/serverless-init/log/log.go
+++ b/cmd/serverless-init/log/log.go
@@ -27,11 +27,11 @@ import (
 const (
 	defaultFlushTimeout   = 5 * time.Second
 	defaultTailingPath    = "/home/LogFiles/*$COMPUTERNAME*.log"
-	modifiableTailingPath = "/home/LogFiles/*$COMPUTERNAME*_%s.log"
+	modifiableTailingPath = "/home/LogFiles/*$COMPUTERNAME*%s.log"
 	logEnabledEnvVar      = "DD_LOGS_ENABLED"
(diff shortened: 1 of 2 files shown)


nhulston (approved):
Nice

nina9753 (approved):

cmd/serverless-init/log/log.go
```diff
@@ -24,12 +25,15 @@ import (
...
 const (
-	defaultFlushTimeout = 5 * time.Second
-	logEnabledEnvVar    = "DD_LOGS_ENABLED"
-	envVarTailFilePath  = "DD_SERVERLESS_LOG_PATH"
-	aasInstanceTailing  = "DD_AAS_INSTANCE_LOGGING_ENABLED"
-	sourceEnvVar        = "DD_SOURCE"
-	sourceName          = "Datadog Agent"
+	defaultFlushTimeout   = 5 * time.Second
+	defaultTailingPath    = "/home/LogFiles/*$COMPUTERNAME*.log"
+	modifiableTailingPath = "/home/LogFiles/*$COMPUTERNAME*_%s.log"
```
apiarian-datadog: why add the underscore here before the `%s`?

cmd/serverless-init/log/log.go
```diff
@@ -24,12 +25,15 @@ import (
...
+	logEnabledEnvVar      = "DD_LOGS_ENABLED"
+	envVarTailFilePath    = "DD_SERVERLESS_LOG_PATH"
+	aasInstanceTailing    = "DD_AAS_INSTANCE_LOGGING_ENABLED"
+	aasLoggingSuffix      = "DD_AAS_INSTANCE_LOGGING_PATH_SUFFIX"
```
apiarian-datadog: nit: it's not really a suffix, it's an infix, right? since we aways have the `.log` at the end?

apiarian-datadog: nit: it's not really a suffix, it's an infix, right? since we aways have the `.log` at the end?

cmd/serverless-init/log/log.go
jcstorms1: Azure's system logs follow `timestamp_instanceID_default_docker.log`. I was thinking it would be easier to write the env var value without the `_`, but I guess it isn't that big of a deal.

apiarian-datadog: ah, so is the underscore a part of the computername, in a sense?

apiarian-datadog (approved):

cmd/serverless-init/log/log.go
jcstorms1: It is only on Azure's logs, not anything custom so it makes sense to remove it.

PR #40593, opened 2025-09-04
best effort to keep comments when updating YAMLs

clarkb7:
### What does this PR do?

forked from https://github.com/DataDog/datadog-agent/pull/38563

Fleet Automation now uses YAML replacement to update configuration, which is much safer than string replaces. The caveat of this is that it removed comments, ours and customers'. This PR fixes that as a best effort by serializing the YAML as `yaml.Node`. It supports the following cases:
- Comments before & after keys
- Inline comments (`key: value # comment`)

Due to YAML Marshal, indented comments will lose their indentation so I updated the `<name>.yaml.example` templates to avoid this.
For example, current template contains
```yaml
# scrubber:
#
  ## @param scrubber.additional_keys - list of strings - optional
  # additional_keys:
  #   - "sensitive_key_1"
  #   - "sensitive_key_2"
```
yaml.Node will change this to, which if uncommented is no longer valid
```yaml
# scrubber:
#
## @param scrubber.additional_keys - list of strings - optional
# additional_keys:
#   - "sensitive_key_1"
#   - "sensitive_key_2"
```
https://github.com/DataDog/datadog-agent/pull/38563 updates this to
```yaml
# scrubber:
#
##    @param scrubber.additional_keys - list of strings - optional
#    additional_keys:
#      - "sensitive_key_1"
#      - "sensitive_key_2"
```
but after uncommenting this block the comment and the fields are not at the same indent level
```yaml
scrubber:

#    @param scrubber.additional_keys - list of strings - optional
  additional_keys:
     - "sensitive_key_1"
     - "sensitive_key_2"
```
yaml.Node will change this to
```yaml
scrubber:

  #    @param scrubber.additional_keys - list of strings - optional
  additional_keys:
     - "sensitive_key_1"
     - "sensitive_key_2"
```

To keep indentation the same when both commented and uncommented, keep a single left most `#`. This whole block can be uncommented and keep its structure
```yaml
# scrubber:
#
#   # @param scrubber.additional_keys - list of strings - optional
#   additional_keys:
#     - "sensitive_key_1"
#     - "sensitive_key_2"
```

### Motivation
keep comments in yaml config

Manual QA using `bin/agent/dist/datadog.yaml` and the config IT uses on our laptops.

render template from main and branch, and then remove the first comment layer and compare diffs. should only be a few minor comment and indentation fixes
```bash
dda inv -- -e agent.generate-config --build-type="agent-py3" --output-file="datadog.yaml"
```
remove first layer of comments with
```bash
sed -Ei 's/^(\s*)# ?/\1/g' datadog.yaml.{main,branch}
```

 unit tests

### Additional Notes
Most of the changes are to fix the major caveat specified above, to move the comment indentation to __after__ the comment char (`#`).


janine-c (approved):

releasenotes/notes/fleet-setup-keep-yaml-comments-40d804ecd60e916c.yaml
```diff
@@ -0,0 +1,6 @@
+---
+enhancements:
+  - |
+    The Datadog Installer introduced in 7.65.0 now keeps YAML comments
```
julien-lebot: **❓Question :** Should this be updated to 7.72?

pkg/fleet/installer/setup/config/write.go
```diff
@@ -0,0 +1,126 @@
...
+// The values are merged as follows:
+// - If the value is a mapping, the nodes are merged recursively
+// - for other types, the src value overrides the dst value
+func mergeNodes(dst *yaml.Node, src *yaml.Node) {
+	if dst.Kind != yaml.MappingNode || src.Kind != yaml.MappingNode {
+		return
+	}
+
+	keyIndex := make(map[string]int)
+	for i := 0; i < len(dst.Content); i += 2 {
```
julien-lebot, line 93: **❓ question:** ‏ Why do we use +2?

clarkb7: `Content` has separate elements for the key and value, e.g.
```
api_key: xxx
```
has two list elements

releasenotes/notes/fleet-setup-keep-yaml-comments-40d804ecd60e916c.yaml
clarkb7: hmm I see the phrasing is a little weird.

I mean to say that starting with 7.65 the comments were being removed, and as of this release they will not be.

This was really only the case for Linux, though, since we haven't been using the `setup` subcommand on Windows.

cc @janine-c  for input

releasenotes/notes/fleet-setup-keep-yaml-comments-40d804ecd60e916c.yaml
```diff
@@ -0,0 +1,6 @@
...
+    when updating the Agent configuration during installation.
+    Some formatting changes can be expected.
```
hush-hush: We should either remove this line or better explain what we mean. This wording is not very explicit for users.

hush-hush (approved):
Looks good for files own by agent-configuration.

Have we tried uncommenting all the lines in the new template and check that the Agent can still start ?

julien-lebot (approved):
LGTM, let's fix the language in the release note before merging however

pkg/config/config_template.yaml
```diff
@@ -1,4 +1,4 @@
-{{ if .Common }}
+{{- if .Common -}}
```
clarkb7, line 1: @hush-hush 
> Have we tried uncommenting all the lines in the new template and check that the Agent can still start ?

It begins to start but fails in the end. I think that means it parsed the YAML without errors at least? I think there are just a lot of invalid configs, for example
```yaml
proxy:
  https: http://<USERNAME>:<PASSWORD>@<PROXY_SERVER_FOR_HTTPS>:<PORT>
  http: http://<USERNAME>:<PASSWORD>@<PROXY_SERVER_FOR_HTTP>:<PORT>
log_file: <AGENT_LOG_FILE_PATH>
```

in my case it crashes from a null-deref in the SNMP check due to a missing err check.
https://github.com/DataDog/datadog-agent/blob/35654a389911966ebe8430bf8338e18e67ad3ea1/pkg/snmp/devicededuper/device_deduper.go#L71-L72

releasenotes/notes/fleet-setup-keep-yaml-comments-40d804ecd60e916c.yaml
clarkb7: maybe something like?
```
While comment text is preserved, its format may not be. Changes to whitespace, indentation, or blank lines may occur.
```

janine-c: To address both comments, how about:
```suggestion
    The Datadog Installer now keeps YAML comments
    when updating the Agent configuration during installation.
    While comment text is preserved, its format may not be.
    Changes to whitespace, indentation, or blank lines may occur.
```
Is it important to mention that the Datadog Installer was introduced in 7.65? If it isn't, I think we can go ahead and omit that so we can focus on this change.

clarkb7: thanks!

PR #40595, opened 2025-09-04
Bump secret-generic-connector to 1.2.0

dustmop:
### What does this PR do?

Bump secret-generic-connector to 1.2.0.

### DIFF
--- a/omnibus/config/software/secret-generic-connector.rb
+++ b/omnibus/config/software/secret-generic-connector.rb
@@ -16,7 +16,7 @@
 require './lib/ostools.rb'
 
 name "secret-generic-connector"
-default_version "1.1.1"
+default_version "1.2.0"
 
 # Define URLs for each platform
 secret_generic_urls = {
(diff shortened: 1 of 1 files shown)


omnibus/config/software/secret-generic-connector.rb
```diff
@@ -38,16 +38,16 @@
 # Note: These should be updated for each version
 secret_generic_sha256 = {
   "linux" => {
-    "amd64" => "42499fc5bc2da362da00381cf5ea62afb67fd9f43dac1fdb9a84f1d72dce9af4",
-    "arm64" => "62e98bf8f631a1646cdffe8b808148512a6714a5a0c876063f2591cc090602f6"
+    "amd64" => "7127faea379dd6ed5731742eefafb6c7ed660dbd1b127d3dbcd8f41764b23836",
```
aiuto, line 41: FWIW. This separation of URL from SHA is a terrible pattern. We should fix that in a separate PR. Or... not fix it and just not do the same thing as we replace omnibus with Bazel.

aiuto (approved):

PR #40600, opened 2025-09-04
pkg/dyninst/decode: Add recover to Decode()

grantseltzer:
### What does this PR do?

Adds a catch all recover to Decode()

### Motivation

Avoid panics in decoding

### DIFF
--- a/pkg/dyninst/decode/decoder.go
+++ b/pkg/dyninst/decode/decoder.go
@@ -8,14 +8,15 @@
 package decode
 
 import (
-	"bytes"
 	"errors"
 	"fmt"
 	"io"
+	"runtime/debug"
 	"time"
 
 	"github.com/go-json-experiment/json"
 	"github.com/google/uuid"
+	pkgerrors "github.com/pkg/errors"
 
 	"github.com/DataDog/datadog-agent/pkg/dyninst/gotype"
 	"github.com/DataDog/datadog-agent/pkg/dyninst/ir"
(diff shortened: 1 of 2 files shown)


pkg/dyninst/decode/decoder.go
```diff
@@ -126,15 +128,26 @@ func (d *Decoder) Decode(
...
 		return nil, err
 	}
 	err = json.MarshalWrite(out, &d.snapshotMessage)
 	if err != nil {
-		t, ok := out.(*bytes.Buffer)
-		if ok {
+		if t, ok := out.(*bytes.Buffer); ok {
 			err = fmt.Errorf("error marshaling snapshot message: %w %q", err, t.String())
+		} else {
+			err = fmt.Errorf("error marshaling snapshot message: %w", err)
```
ajwerner: Let's get rid of printing out the buffer. I didn't mean to check it in.

ajwerner (approved):

pkg/dyninst/decode/decoder_test.go
```diff
@@ -736,3 +736,132 @@ func (r *noopTypeNameResolver) ResolveTypeName(
 ) (string, error) {
 	return fmt.Sprintf("type%#x", typeID), nil
 }
+
+// TestDecoderPanicRecovery tests that panics during processExpression are properly caught
```
ajwerner: This test is rather involved and not necessarily targeted. How about something that much more clearly intentionally induces a panic.

```
type panicDecoderType struct {
	decoderType
}

var _ decoderType = (*panicDecoderType)(nil)

func (t *panicDecoderType) encodeValueFields(*Decoder, *jsontext.Encoder, []byte) error {
	panic("boom")
}

func TestDecoderPanics(t *testing.T) {
	irProg := generateIrForProbes(t, "simple", "stringArg")
	decoder, err := NewDecoder(irProg, &noopTypeNameResolver{})
	caseIdx := slices.IndexFunc(cases, func(c testCase) bool {
		return c.probeName == "stringArg"
	})
	testCase := &cases[caseIdx]
	input := testCase.eventConstructor(t, irProg)
	var stringType ir.Type
	for _, t := range irProg.Types {
		if t.GetName() == "string" {
			stringType = t
			break
		}
	}
	require.NotNil(t, stringType)
	stringID := stringType.GetID()
	decoder.decoderTypes[stringID] = &panicDecoderType{decoder.decoderTypes[stringID]}
	_, err = decoder.Decode(Event{
		Event:       output.Event(input),
		ServiceName: "foo"},
		&noopSymbolicator{},
		bytes.NewBuffer(nil),
	)
	require.Error(t, err)
	require.Contains(t, err.Error(), "boom")
}
```

PR #40602, opened 2025-09-04
dyninst/symdb: use disk-based uncompression of binaries

andreimatei:
Before this patch, we were uncompressing files in memory for the SymDB upload. Switch to uncompressing them in the disk cache, to save on memory use.


ajwerner (approved):
Thank you!

PR #40606, opened 2025-09-05
dyninst/irgen: include type information for interface implementors

ajwerner:
### What does this PR do?

This PR concludes https://datadoghq.atlassian.net/browse/DEBUG-4043. Along the way it fixes some other rough edges in decoding related to interfaces. There's some details related to how the on-disk indexes are implemented that should be refactored in a follow-up, but it gets the job done and avoids using too much RAM.

See individual commits.


piob-io (approved):
Neat!

PR #40612, opened 2025-09-05
Fixed invalid result json parsing

CelianR:
### What does this PR do?

Fixes this [issue](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1113509369#L2439).


tasks/libs/testing/result_json.py
```diff
@@ -15,6 +15,7 @@ class ActionType(Enum):
     SKIP = "skip"  # noqa: F841
     PAUSE = "pause"  # noqa: F841
     CONT = "cont"  # noqa: F841
+    BUILD_OUTPUT = "build-output"  # noqa: F841
```
KevinFairise2, line 18: Do we need `"build-fail"` as well? I can see that one in the link you shared

CelianR: Nice catch!

CelianR: Updated

CelianR: I double checked, we needed only these two

KevinFairise2 (approved):

PR #40616, opened 2025-09-05
[ACIX-468] Remove duplicate rpm/deb unit tests

KevinFairise2:
### What does this PR do?

Running twice the unit test on rpm and deb OSes was judged not useful. Especially because the build is exactly the same for both rpm and deb builds, only the packaging is different.
The only thing that would make a test fail is a tunit est using the underlying OS to assert things, which would be a bad unit tests. Such things that do assertion on the OS on which it runs should be e2e or integration tests

### Motivation

Remove duplicated unit tests and reduce usage of an image we want to deprecate totally


ofek (approved):
🔥

chouetz (approved):
👏🏽

PR #40624, opened 2025-09-05
remove dependency on lint jobs from binary build jobs

paulcacheux:
### What does this PR do?

There is no actual dependency between the lint jobs and the dogstatsd/iot-agent binary build jobs. In addition this can prevent a PR failing some lints to progress forward and give additional errors in one pipeline (especially this blocks the static quality gates from running).

This PR removes the dependency link.


rdesgroppes (approved):
Right, indeed!

KevinFairise2 (approved):

PR #40626, opened 2025-09-05
fix: allow square brackets in evp proxy query strings

rarguelloF:
### What does this PR do?
LLM Observability (Experiments and Datasets) uses `[]` characters in some endpoints query strings.

For example, a query to search datasets by name would be:

`GET /api/unstable/llm-obs/v1/datasets?filter[name]=my-name`

When using non-agentless mode, we are getting 502s from the agent because square brackets are not allowed characters.

### Motivation
Allow LLM Observability to work using the agent.


gary-huang (approved):

ajgajg1134:
Are square brackets even allowed technically speaking? My understanding is that they are reserved characters (and are used in ipv6 addresses), can you get around this by percent encoding those special characters? 🤔

rarguelloF:
@ajgajg1134 correct, these are sent url-encoded as `%5B` and `%5D` respectively.

ichinaski (approved):

PR #40466, opened 2025-09-02
[CNM-4706] Update Network Path for datadog-traceroute schema update

AlexandreYang:
### What does this PR do?

[CNM-4706] Update Network Path for datadog-traceroute schema update
- Support Multi Traceroute schema
- Support E2E Probes schema

### Motivation

Network Tests project

### Describe how you validated your changes

- [x] Test by running the Agent locally
- [x] Test by running the Agent with Static Path & Dynamic Path on staging on medium & large clusters

### Additional Notes

Development:
```
inv test --targets ./pkg/networkpath/...,./comp/networkpath/npcollector/npcollectorimpl/...,./test/new-e2e/tests/netpath/network-path-integration,./pkg/collector/corechecks/networkpath/...
```

[CNM-4706]: https://datadoghq.atlassian.net/browse/CNM-4706?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #40630, opened 2025-09-05
[SYNTH-21261] Add a Synthetics Test Scheduler to run network tests

Fabrice-Matrat:
### What does this PR do?

Add a Synthetics test runner component that receive config for tests by remote config, run the test at a regular interval and send the result to the Synthetics track.

### Motivation

The goal is to be able to build, execute network tests inside the agent driven by Synthetics and  take advantage of data enrichment built by the network path team.


hush-hush (approved):

comp/syntheticstestscheduler/mock/mock.go
```diff
@@ -0,0 +1,21 @@
...
+package mock
+
+import (
+	"testing"
+
+	syntheticstestscheduler "github.com/DataDog/datadog-agent/comp/syntheticstestscheduler/def"
+)
+
+// Mock returns a mock for syntheticstestscheduler component.
+func Mock(t *testing.T) syntheticstestscheduler.Component {
```
hush-hush: We don't need a mock since your component doesn't have any public method in it's interface.

comp/syntheticstestscheduler/def/component.go
```diff
@@ -0,0 +1,14 @@
...
+
+// Package syntheticstestscheduler ... /* TODO: detailed doc comment for the component */
+package syntheticstestscheduler
+
+// team: /* TODO: add team name */
+
+// Component is the component type.
+type Component interface {
+    /* TODO: define Component interface */
+}
```
hush-hush: Nit: fix the TODOs

comp/syntheticstestscheduler/impl/component.go
```diff
@@ -0,0 +1,80 @@
...
+	RCListener              rctypes.ListenerProvider
+	SyntheticsTestScheduler *SyntheticsTestScheduler
+}
+
+// NewComponent creates a new syntheticstestscheduler component
+func NewComponent(reqs Requires) (Provides, error) {
+	configs := newSchedulerConfigs(reqs.AgentConfig)
+	if !configs.syntheticsSchedulerEnabled {
+		reqs.Logger.Debugf("Synthetics scheduler disabled")
+		return Provides{}, nil
```
hush-hush: I think that still register an empty interface for `Comp` and `RCListener` into the dependency tree. If anybody start using them they'll panic. Does returning an nil `RCListener` works ?

hush-hush (changes requested):

comp/syntheticstestscheduler/def/component.go
pgimalac, line 1: Could you update CODEOWNERS so that this component is owned by the proper team ?

comp/syntheticstestscheduler/mock/mock.go
Fabrice-Matrat: done

comp/syntheticstestscheduler/def/component.go
Fabrice-Matrat: done

comp/syntheticstestscheduler/impl/component.go
Fabrice-Matrat: done.

comp/syntheticstestscheduler/def/component.go
Fabrice-Matrat: done.

comp/syntheticstestscheduler/common/result.go
```diff
@@ -0,0 +1,172 @@
...
+type E2ETest struct {
+	PacketLoss float64 `json:"packet_loss"`
+	LatencyAvg float64 `json:"latency_avg"`
+	LatencyMin float64 `json:"latency_min"`
+	LatencyMax float64 `json:"latency_max"`
+	Jitter     float64 `json:"jitter"`
+}
+
+// NetpathResult represents the full result of a network path test.
+type NetpathResult struct {
```
AlexandreYang: For netpath payload, please use `pkg/networkpath/payload/pathevent.go` once this PR is merged: https://github.com/DataDog/datadog-agent/pull/40466

comp/syntheticstestscheduler/impl/component.go
```diff
@@ -0,0 +1,84 @@
...
+	epForwarder, ok := reqs.EpForwarder.Get()
+	if !ok {
+		return Provides{}, reqs.Logger.Errorf("error getting EpForwarder")
+	}
+
+	scheduler, err := newSyntheticsTestScheduler(
+		configs,
+		epForwarder,
+		reqs.Logger,
+		reqs.AgentConfig.GetString("config_id"),
```
AlexandreYang: Why do we get `config_id` from agent configs ? is this used for testing / development only?

comp/syntheticstestscheduler/impl/scheduler.go
```diff
@@ -0,0 +1,236 @@
...
+		}
+
+		newConfig[syntheticsTestCfg.PublicID] = syntheticsTestCfg
+		s.log.Debugf("Processed config %s: %v", configPath, syntheticsTestCfg)
+
+		applyStateCallback(configPath, state.ApplyStatus{State: state.ApplyStateAcknowledged})
+	}
+
+	s.updateRunningState(newConfig)
+	s.runningConfig = newConfig
```
AlexandreYang: Might be better to also run `s.runningConfig = newConfig` inside `updateRunningState()` to benefit from the mutex lock? for safety

comp/syntheticstestscheduler/impl/scheduler.go
```diff
@@ -0,0 +1,236 @@
...
+		} else {
+			current.cfg = newTestConfig
+		}
+	}
+
+	for pubID := range s.state.tests {
+		if _, exists := seen[pubID]; !exists {
+			delete(s.state.tests, pubID)
+		}
+	}
```
AlexandreYang, line 138: Are we sure this logic works? 

I mean, if users have many RC configs, they might *not arrive all in one RC update* right?
(actually, I'm not fully sure, but something to confirm with RC team if not already).

If so, this logic won't work, since the `seen` map here is specific to this case only.

comp/syntheticstestscheduler/impl/scheduler.go
```diff
@@ -0,0 +1,236 @@
...
+		return err
+	}
+	if err := persistentcache.Write(cacheKey, string(cacheValue)); err != nil {
+		return s.log.Errorf("couldn't write cache: %s", err)
+	}
+	return nil
+}
+
+// retrieveConfig reads scheduler config from persistent cache and initializes state.
+func (s *SyntheticsTestScheduler) retrieveConfig() error {
```
AlexandreYang: (nit) Maybe `loadConfigsFromCache` could be a more precise name?

comp/syntheticstestscheduler/impl/scheduler.go
```diff
@@ -0,0 +1,236 @@
...
+	cacheValue, err := persistentcache.Read(cacheKey)
+	if err != nil {
+		return s.log.Errorf("couldn't read from cache: %s", err)
+	}
+	if len(cacheValue) == 0 || cacheValue == "{}" {
```
AlexandreYang: hum. Why do we need to check for `cacheValue == "{}"` ? 
It should be handled naturally by subsequent code already?

comp/syntheticstestscheduler/impl/scheduler.go
```diff
@@ -0,0 +1,236 @@
...
+	s.state.mu.Lock()
+	defer s.state.mu.Unlock()
+
+	s.state.tests = make(map[string]*runningTestState, len(cfg))
+	now := s.TimeNowFn()
+	for id, testCfg := range cfg {
+		s.state.tests[id] = &runningTestState{
+			cfg:     testCfg,
+			lastRun: time.Time{},
+			nextRun: now.Add(time.Duration(testCfg.Interval) * time.Second),
```
AlexandreYang: Hum, aren't we supposed to persist in cache the nextRun and re-load it?

I mean, for example, if a test should run once every week.

- Let's say at t0 the test is run, we persist the data
- and t0+3 days, the agent restarts, at this point, we should still ideally run the test at t0+1week right? which seems not the case with the current impl.

comp/syntheticstestscheduler/impl/worker.go
```diff
@@ -0,0 +1,342 @@
...
+// SyntheticsTestCtx is the unit of work consumed by workers.
+type SyntheticsTestCtx struct {
+	nextRun time.Time
+	cfg     common.SyntheticsTestConfig
+}
+
+// updateTestState updates lastRun and nextRun for a running test.
+func (s *SyntheticsTestScheduler) updateTestState(flushTime time.Time, rt *runningTestState) {
+	rt.lastRun = flushTime
+	rt.nextRun = flushTime.Add(time.Duration(rt.cfg.Interval) * time.Second)
```
AlexandreYang: `flushTime` might vary when the agent is overloaded for example, shall we rely only on the nextRun/lastRun info?

For example, something like

```suggestion
	rt.nextRun = rt.nextRun.Add(time.Duration(rt.cfg.Interval) * time.Second)
	rt.lastRun = rt.nextRun
```

This way, there is less risk of "drifting" if current flush time is not precise.

comp/syntheticstestscheduler/impl/worker.go
```diff
@@ -0,0 +1,342 @@
...
+	}
+	path, err := tr.Run(context.TODO())
+	if err != nil {
+		return payload.NetworkPath{}, fmt.Errorf("run traceroute error: %s", err)
+	}
+	return path, nil
+}
+
+// networkPathToTestResult converts a WorkerResult into the public TestResult structure.
+func (s *SyntheticsTestScheduler) networkPathToTestResult(w *WorkerResult, testConfigID string) (*common.TestResult, error) {
```
AlexandreYang: Since the schema has changed a lot in datadog-traceroute, this code will need some update.

See example here: https://github.com/DataDog/datadog-agent/pull/40466

comp/syntheticstestscheduler/impl/worker.go
```diff
@@ -0,0 +1,342 @@
...
+	return &common.TestResult{
+		DD:     make(map[string]interface{}),
+		Result: result,
+		Test:   t,
+		V:      1,
+	}, nil
+}
+
+// generateRandomStringUInt63 returns a cryptographically random uint63 as decimal string.
+func generateRandomStringUInt63() (string, error) {
```
AlexandreYang: How safe is this impl? Is that similar/same compared to what s8s worker does?

Curious, how does it compare to using UUID4 ? why not use uuid4?

AlexandreYang (dismissed):
Did some high level review of the PR and left some comment.

The overall design look fine to me.


I can review again once the PR is more stable/finalised.

comp/syntheticstestscheduler/impl/component.go
Fabrice-Matrat: It's to populate TestConfigID in netpath result but might not be the right config id ?

comp/syntheticstestscheduler/impl/scheduler.go
Fabrice-Matrat: I asked the RC team and I got the following answer :
The agent clients get a map of all configurations that are valid. It's up to the client to know what to do
the ones were there before and are gone are the deleted one.

comp/syntheticstestscheduler/impl/worker.go
Fabrice-Matrat: https://github.com/DataDog/synthetics-worker/blob/prod/packages/lib/node-utils/src/helpers.ts#L30
I use the same algorythm than in the worker

comp/syntheticstestscheduler/impl/scheduler.go
Fabrice-Matrat: removed the running config in fact

comp/syntheticstestscheduler/impl/scheduler.go
Fabrice-Matrat: done

comp/syntheticstestscheduler/impl/scheduler.go
Fabrice-Matrat: done

comp/syntheticstestscheduler/impl/scheduler.go
Fabrice-Matrat: done

comp/syntheticstestscheduler/impl/worker.go
Fabrice-Matrat: done

comp/syntheticstestscheduler/def/component.go
pgimalac: I think you only edited the comment in the component, you should run `inv lint-components --fix` to actually update the codeowners file 
You might also need to add `synthetics-executing` to a list of known teams somewhere

comp/syntheticstestscheduler/impl/component.go
AlexandreYang: hum, the `test_config_id` should come from the backend via RC (from test propagator) right? cc @rafamelo12

comp/syntheticstestscheduler/impl/scheduler.go
AlexandreYang: Can we try to confirm and test it works (maybe later once we have configs in RC) when we have multiple configs for one agent?

comp/syntheticstestscheduler/impl/component.go
rafamelo12: What are we referring to as `test_config_id` here? Is it the `result_id` or other id?

AlexandreYang: This `reqs.AgentConfig.GetString("config_id")` doesn't seems correct to me, there is a concept of `config_id` in datadog.yaml config, but it's used by Fleet Automation to represent agent configs. Which seems to have nothing to do with synthetics test configs :)

rafamelo12: Indeed, the only two IDs we care about at s8s is `result_id` and `public_id` so we can probably disregard this one, unless this `config_id` is what we'll use to keep track of "loaded" RCs into our integration.

hush-hush (approved):
Looks good for agent-configuration files.

comp/syntheticstestscheduler/impl/component.go
Fabrice-Matrat: fixed

comp/syntheticstestscheduler/impl/worker.go
Fabrice-Matrat: done

comp/syntheticstestscheduler/common/data.go
```diff
@@ -0,0 +1,161 @@
...
+type TCPMethod string
+
+const (
+	// TCPMethodPreferSACK represents a preference to sack tcp method if available for network test.
+	TCPMethodPreferSACK TCPMethod = "prefer_sack"
+	// TCPMethodSYN represents a syn tcp method for network test.
+	TCPMethodSYN TCPMethod = "syn"
+	// TCPMethodSACK represents a sack tcp method for network test.
+	TCPMethodSACK TCPMethod = "sack"
+)
```
AlexandreYang: Shall we use the constants from https://github.com/DataDog/datadog-traceroute/blob/main/traceroute/traceroute.go#L20-L29 ?

comp/syntheticstestscheduler/common/data.go
```diff
@@ -0,0 +1,161 @@
...
+type SubType string
+
+const (
+	// SubTypeUDP represents a udp network test.
+	SubTypeUDP SubType = "udp"
+	// SubTypeTCP represents a tcp network test.
+	SubTypeTCP SubType = "tcp"
+	// SubTypeICMP represents a icmp network test.
+	SubTypeICMP SubType = "icmp"
+)
```
AlexandreYang: Shall we use the constants from: https://github.com/DataDog/datadog-traceroute/blob/main/traceroute/traceroute.go#L8-L15 

(the case is different though, maybe we should lower case the constants in datadog-traceroute for consistency)

comp/syntheticstestscheduler/common/data.go
```diff
@@ -0,0 +1,161 @@
...
+	} `json:"config"`
+
+	Interval int    `json:"tick_every"`
+	OrgID    int    `json:"orgID"`
+	MainDC   string `json:"mainDC"`
+	PublicID string `json:"publicID"`
+}
+
+// UnmarshalJSON is a Custom unmarshal for SyntheticsTestConfig
+func (c *SyntheticsTestConfig) UnmarshalJSON(data []byte) error {
```
AlexandreYang, line 84: Can we unit test this?

In general, when possible, if we have new/modified logic, it should be unit tested.

comp/syntheticstestscheduler/impl/config.go
```diff
@@ -0,0 +1,25 @@
...
+	"time"
+)
+
+type schedulerConfigs struct {
+	workers                    int
+	flushInterval              time.Duration
+	syntheticsSchedulerEnabled bool
+}
+
+func newSchedulerConfigs(agentConfig config.Component) *schedulerConfigs {
```
AlexandreYang, line 19: Can we unit test this?

comp/syntheticstestscheduler/impl/worker.go
```diff
@@ -0,0 +1,328 @@
...
+	triggeredAt      time.Time
+	startedAt        time.Time
+	finishedAt       time.Time
+	duration         time.Duration
+	tracerouteCfg    config.Config
+	hostname         string
+}
+
+// toNetpathConfig converts a SyntheticsTestConfig into a system-probe Config.
+func toNetpathConfig(c common.SyntheticsTestConfig) (config.Config, error) {
```
AlexandreYang, line 150: I think we should have a unit test for `toNetpathConfig()`, for the different code paths.

comp/syntheticstestscheduler/impl/worker.go
```diff
@@ -0,0 +1,328 @@
...
+	}
+	path, err := tr.Run(ctx)
+	if err != nil {
+		return payload.NetworkPath{}, fmt.Errorf("run traceroute error: %s", err)
+	}
+	return path, nil
+}
+
+// networkPathToTestResult converts a WorkerResult into the public TestResult structure.
+func (s *SyntheticsTestScheduler) networkPathToTestResult(w *WorkerResult) (*common.TestResult, error) {
```
AlexandreYang: I think `networkPathToTestResult()` should be unit tested.

comp/syntheticstestscheduler/impl/worker.go
```diff
@@ -0,0 +1,328 @@
...
+	return &common.TestResult{
+		DD:     make(map[string]interface{}),
+		Result: result,
+		Test:   t,
+		V:      1,
+	}, nil
+}
+
+// generateRandomStringUInt63 returns a cryptographically random uint63 as decimal string.
+func generateRandomStringUInt63() (string, error) {
```
AlexandreYang: Can we unit test `generateRandomStringUInt63()` ? maybe by mocking/faking the random value

Fabrice-Matrat: done

comp/syntheticstestscheduler/impl/worker.go
Fabrice-Matrat: done

comp/syntheticstestscheduler/impl/worker.go
Fabrice-Matrat: done

comp/syntheticstestscheduler/common/data.go
Fabrice-Matrat: done

comp/syntheticstestscheduler/impl/component.go
```diff
@@ -0,0 +1,90 @@
...
+		return Provides{
+			RCListener: rctypes.ListenerProvider{ListenerProvider: rctypes.RCListener{}},
+			Comp:       empty,
+		}, nil
+	}
+
+	forwarder := reqs.Forwarder
+	//if !ok {
+	//	return Provides{}, reqs.Logger.Errorf("error getting EpForwarder")
+	//}
```
AlexandreYang:
```suggestion
```
?

AlexandreYang (commented):
LGTM overall:

few things:

- let's make sure that the defaultforwarder is the best choice for our use case
- we should test it actually works with real RC data

vickenty (dismissed):

comp/syntheticstestscheduler/impl/component.go
Fabrice-Matrat: done

comp/syntheticstestscheduler/common/result.go
Fabrice-Matrat: done

comp/syntheticstestscheduler/impl/scheduler.go
```diff
@@ -0,0 +1,226 @@
...
+
+		newConfig[syntheticsTestCfg.PublicID] = syntheticsTestCfg
+		s.log.Debugf("Processed config %s: %v", configPath, syntheticsTestCfg)
+
+		applyStateCallback(configPath, state.ApplyStatus{State: state.ApplyStateAcknowledged})
+	}
+
+	if err := s.updateRunningState(newConfig); err != nil {
+		s.log.Warnf("unable to persist synthetics test config: %v", err)
+		// TODO: what path should I provide if it is global to the config
```
rafamelo12, line 119: Have we figured this out?

comp/syntheticstestscheduler/impl/scheduler.go
```diff
@@ -0,0 +1,226 @@
...
+	seen := map[string]struct{}{}
+	for _, newTestConfig := range newConfig {
+		pubID := newTestConfig.PublicID
+		seen[pubID] = struct{}{}
+		current, exists := s.state.tests[pubID]
+		if !exists {
+			s.state.tests[pubID] = &runningTestState{
+				cfg:     newTestConfig,
+				lastRun: time.Time{},
+				nextRun: s.TimeNowFn().Add(time.Duration(newTestConfig.Interval) * time.Second),
```
rafamelo12: Is this function timezone safe? I'm worried if we can have missed runs during a daylight time changes.

comp/syntheticstestscheduler/impl/syntheticstestscheduler_test.go
```diff
@@ -0,0 +1,628 @@
...
...6901488591,"testTriggeredAt":1756901488590,"assertions":null,"failure":null,"duration":1,"request":{"host":"example.com","port":443,"maxTtl":30,"timeout":5000},"netstats":{"packetsSent":0,"packetsReceived":0,"packetLossPercentage":0,"jitter":0,"latency":{"avg":0,"min":0,"max":0},"hops":{"avg":0,"min":0,"max":0}},"netpath":{"timestamp":0,"agent_version":"","namespace":"","test_config_id":"puf-9fm-c89","test_result_id":"4907739274636687553","pathtrace_id":"pathtrace-id-111-example.com","origin":"synthetics","protocol":"TCP","source":{"name":"test-hostname","display_name":"test-hostname","hostname":"test-hostname"},"destination":{"hostname":"example.com","port":443},"traceroute":{"runs":[{"run_id":"1","source":{"ip_address":"","port":0},"destination":{"ip_address":"","port":0},"hops":[{"ttl":0,"ip_address":"1.1.1.1","reachable":false},{"ttl":0,"ip_address":"1.1.1.2","reachable":false}]}],"hop_count":{"avg":0,"min":0,"max":0}},"e2e_probe":{"rtts":null,"packets_sent":0,"packets_received":0,"packet_loss_percentage":0,"jitter":0,"rtt":{"avg":0,"min":0,"max":0}}},"status":"passed"},"test":{"_internalId":"puf-9fm-c89","id":"puf-9fm-c89","subType":"TCP","type":"network","version":1},"v":1}`,
```
rafamelo12: `_internalId` should not exist on our payload, this will be enriched later at either the PL Bridge or Workloads

rafamelo12 (commented):
Just a few questions, overall LGTM

comp/syntheticstestscheduler/impl/scheduler.go
Fabrice-Matrat: done

comp/syntheticstestscheduler/impl/syntheticstestscheduler_test.go
Fabrice-Matrat: done

przemko0 (approved):

gh123man (approved):
ALP owned files LGTM

ddrthall (approved):

daniel-taf (approved):

ofek (approved):

rafamelo12 (approved):
LGTM

comp/syntheticstestscheduler/impl/scheduler.go
```diff
@@ -0,0 +1,226 @@
...
+		s.log.Debugf("Processed config %s: %v", configPath, syntheticsTestCfg)
+
+		applyStateCallback(configPath, state.ApplyStatus{State: state.ApplyStateAcknowledged})
+	}
+
+	if err := s.updateRunningState(newConfig); err != nil {
+		s.log.Warnf("unable to persist synthetics test config: %v", err)
+		// TODO: what path should I provide if it is global to the config
+		applyStateCallback("TODO", state.ApplyStatus{
+			State: state.ApplyStateError,
```
AlexandreYang: Could you address this TODO? I don't know how critical this is, could be a question for remote config team

comp/syntheticstestscheduler/impl/scheduler.go
AlexandreYang: @Fabrice-Matrat any news on this ? ☝️

comp/syntheticstestscheduler/impl/worker.go
```diff
@@ -0,0 +1,328 @@
...
+		}
+		if req.Timeout != nil {
+			cfg.Timeout = time.Duration(*req.Timeout) * time.Second
+		}
+		if req.SourceService != nil {
+			cfg.SourceService = *req.SourceService
+		}
+		if req.DestinationService != nil {
+			cfg.DestinationService = *req.DestinationService
+		}
```
AlexandreYang: 🤔 There is a lot of code duplication here, I think there is probably a way to factor that.

AlexandreYang (approved):

comp/syntheticstestscheduler/impl/scheduler.go
mellon85: you must put the key of the map itself, the `configPath`

cmd/agent/subcommands/run/command.go
```diff
@@ -532,6 +533,7 @@ func getSharedFxOption() fx.Option {
 		settingsimpl.Module(),
 		agenttelemetryfx.Module(),
 		networkpath.Bundle(),
+		syntheticsTestsfx.Module(),
```
pgimalac, line 536: I might be missing something but if you add a component without any other component depending on it, it won't be instanciated (`fx` computes the graph of components and only initializes the ones which are requested by another requested component)

cmd/process-agent/command/main_common.go
```diff
@@ -142,6 +143,9 @@ func runApp(ctx context.Context, globalParams *GlobalParams) error {
 		// Provide network path bundle
 		networkpath.Bundle(),
 
+		// Provide synthetics test scheduler module
+		syntheticsTestsfx.Module(),
```
pgimalac: Does this need to be in the process-agent ?
The process-agent is being removed (most of the features have been moved to the core-agent already)

comp/syntheticstestscheduler/common/data.go
```diff
@@ -0,0 +1,139 @@
...
+		return err
+	}
+
+	c.Version = tmp.Version
+	c.Type = tmp.Type
+	c.Subtype = tmp.Subtype
+	c.OrgID = tmp.OrgID
+	c.MainDC = tmp.MainDC
+	c.PublicID = tmp.PublicID
+	c.Config.Assertions = tmp.Config.Assertions
```
pgimalac, line 110: Any particular reason to unmarshal into a temp struct then copy the field one by one ? To avoid changing any field in `SyntheticsTestConfig` in case of error ?

comp/syntheticstestscheduler/common/data.go
```diff
@@ -0,0 +1,139 @@
...
+	case payload.ProtocolICMP:
+		var req ICMPConfigRequest
+		if err := json.Unmarshal(tmp.Config.Request, &req); err != nil {
+			return err
+		}
+		c.Config.Request = req
+	default:
+		// unknown subtype, return error
+		return fmt.Errorf("unknown subtype: %s", tmp.Subtype)
+	}
```
pgimalac, line 134: I think something like that should work while simplifying the code
```suggestion
	var req ConfigRequest
	switch payload.Protocol(tmp.Subtype) {
	case payload.ProtocolUDP:
		req = UDPConfigRequest{}
	case payload.ProtocolTCP:
		req = TCPConfigRequest{}
	case payload.ProtocolICMP:
		req = ICMPConfigRequest{}
	default:
		// unknown subtype, return error
		return fmt.Errorf("unknown subtype: %s", tmp.Subtype)
	}

	if err := json.Unmarshal(tmp.Config.Request, &req); err != nil {
		return err
	}
	c.Config.Request = req
```

comp/syntheticstestscheduler/common/result.go
```diff
@@ -0,0 +1,103 @@
...
+	Type    string `json:"type"`
+	Version int    `json:"version"`
+}
+
+// TestResult represents the full test execution result including metadata.
+type TestResult struct {
+	DD     map[string]interface{} `json:"_dd"`
+	Result Result                 `json:"result"`
+	Test   Test                   `json:"test"`
+	V      int                    `json:"v"`
```
pgimalac: What's `V` ?

comp/syntheticstestscheduler/common/result.go
```diff
@@ -0,0 +1,103 @@
...
+type APIError struct {
+	Code    APIErrorCode `json:"code"`
+	Message string       `json:"message"`
+}
+
+// APIFailure represents an API failure with a code and message.
+type APIFailure struct {
+	Code    APIFailureCode `json:"code"`
+	Message string         `json:"message"`
+}
```
pgimalac, line 85: Is `APIFailure` actually used somewhere ?

comp/syntheticstestscheduler/fx/fx.go
```diff
@@ -0,0 +1,25 @@
...
+
+// Module defines the fx options for this component
+func Module() fxutil.Module {
+	return fxutil.Component(
+		fxutil.ProvideComponentConstructor(
+			syntheticstestschedulerimpl.NewComponent,
+		),
+		fxutil.ProvideOptional[syntheticstestscheduler.Component](),
+	)
+}
```
pgimalac, line 34: As mentioned in another comment, `fx` will only initialize the components which are a dependency of a required component
A trick if your component is used for its side effects is to add an `fx.Invoke` taking your component as a parameter (it will make `fx` see your component as requested and initialize it).

See the `Module` function in `comp/core/configsync/configsyncimpl/module.go` for an example.

comp/syntheticstestscheduler/impl/scheduler.go
```diff
@@ -0,0 +1,179 @@
...
+
+// syntheticsTestScheduler is responsible for scheduling and executing synthetics tests.
+type syntheticsTestScheduler struct {
+	log                          log.Component
+	state                        runningState
+	cancel                       context.CancelFunc
+	running                      bool
+	workers                      int
+	workersDone                  chan struct{}
+	TimeNowFn                    func() time.Time
```
pgimalac: Any reason this field is exported ? (mostly curious as it's the only one)

comp/syntheticstestscheduler/impl/worker.go
```diff
@@ -0,0 +1,325 @@
...
+
+// flush enqueues tests whose nextRun is due.
+func (s *syntheticsTestScheduler) flush(flushTime time.Time) {
+	for id, rt := range s.state.tests {
+		if flushTime.After(rt.nextRun) || flushTime.Equal(rt.nextRun) {
+			s.log.Debugf("enqueuing test %s", id)
+			s.syntheticsTestProcessingChan <- SyntheticsTestCtx{
+				nextRun: flushTime,
+				cfg:     rt.cfg,
+			}
```
pgimalac, line 68: If the runners are slow this could block, is it an issue ?

comp/syntheticstestscheduler/impl/syntheticscomponent.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	scheduler := newSyntheticsTestScheduler(
+		configs,
+		epForwarder,
+		reqs.Logger,
+		reqs.HostnameService,
+		time.Now,
+		cancel)
```
pgimalac: It seems weird to pass the cancel to the constructor but not the context 😄
Why not just call `cancel` in the `OnStop` ?

Or alternatively why not let `start` create its own `Context` (and cancel) ? If the base context is `Background` anyway it wouldn't change anything

comp/syntheticstestscheduler/impl/worker.go
```diff
@@ -0,0 +1,325 @@
...
+	hostname         string
+}
+
+// toNetpathConfig converts a SyntheticsTestConfig into a system-probe Config.
+func toNetpathConfig(c common.SyntheticsTestConfig) (config.Config, error) {
+	var cfg config.Config
+
+	switch payload.Protocol(c.Subtype) {
+	case payload.ProtocolUDP:
+		req, ok := c.Config.Request.(common.UDPConfigRequest)
```
pgimalac: You can get the protocol by doing a type assertion on the `Request` field, so `SubType` is kind of redundant
Ie. 
```go
switch req := c.Config.Request.(type) {
case common.UDPConfigRequest:
	// logic
case common.TCPConfigRequest:
	// logic
```

comp/syntheticstestscheduler/impl/worker.go
pgimalac: Same comment, even just using an helper function to fill the fields of `NetworkConfigRequest` would be a good start

comp/syntheticstestscheduler/impl/worker.go
```diff
@@ -0,0 +1,325 @@
...
+	nextRun time.Time
+	cfg     common.SyntheticsTestConfig
+}
+
+// updateTestState updates lastRun and nextRun for a running test.
+func (s *syntheticsTestScheduler) updateTestState(rt *runningTestState) {
+	s.state.mu.Lock()
+	defer s.state.mu.Unlock()
+	rt.lastRun = rt.nextRun
+	rt.nextRun = rt.nextRun.Add(time.Duration(rt.cfg.Interval) * time.Second)
```
pgimalac, line 205: Why read-lock `s` to read/write `rt` ?

comp/syntheticstestscheduler/impl/worker.go
```diff
@@ -0,0 +1,325 @@
...
+
+	return &common.TestResult{
+		DD:     make(map[string]interface{}),
+		Result: result,
+		Test:   t,
+		V:      1,
+	}, nil
+}
+
+// generateRandomStringUInt63 returns a cryptographically random uint63 as decimal string.
```
pgimalac, line 300: > cryptographically random

It's only as good as the provided `randIntFn`, not sure the comment is very accurate

pgimalac: Why not just use a uuid library ?

comp/syntheticstestscheduler/common/result.go
pgimalac: And what's `DD` actually ?

pgimalac (commented):
Various nitpicks, nothing big but there are many so I'd rather have them addressed before approving

comp/syntheticstestscheduler/common/data.go
Fabrice-Matrat: that's for the translation to 		Request    ConfigRequest `json:"request"`

comp/syntheticstestscheduler/common/result.go
Fabrice-Matrat: that's the structure we have in synthetics type script code, so we send the same json result so it can be processed the same way _dd and seems mandatory but will be empty map for dd and V is 1.

comp/syntheticstestscheduler/common/result.go
Fabrice-Matrat: no yet but in next PR

comp/syntheticstestscheduler/impl/worker.go
Fabrice-Matrat: https://github.com/DataDog/synthetics-worker/blob/prod/packages/lib/node-utils/src/helpers.ts#L30

comp/syntheticstestscheduler/impl/worker.go
Fabrice-Matrat: rt is part of the state

comp/syntheticstestscheduler/impl/scheduler.go
Fabrice-Matrat: no

comp/syntheticstestscheduler/fx/fx.go
Fabrice-Matrat: thanks a lot, I was waiting for this trick :)

comp/syntheticstestscheduler/common/data.go
Fabrice-Matrat: &req is a pointer to the interface variable, not a pointer to the underlying struct.
json.Unmarshal cannot write JSON into an interface pointer. It needs a pointer to a concrete struct

comp/syntheticstestscheduler/impl/worker.go
Fabrice-Matrat: done

comp/syntheticstestscheduler/impl/syntheticscomponent.go
Fabrice-Matrat: done

cmd/process-agent/command/main_common.go
Fabrice-Matrat: done

comp/syntheticstestscheduler/impl/worker.go
Fabrice-Matrat: not really

comp/syntheticstestscheduler/impl/scheduler.go
Fabrice-Matrat: yep this is confirmed

comp/syntheticstestscheduler/common/data.go
pgimalac: I think you could implement the unmarshal method for `ConfigRequest` and it would quite nicely
But fair 👍

comp/syntheticstestscheduler/common/data.go
pgimalac: Ah I see, that's too bad

comp/syntheticstestscheduler/common/result.go
pgimalac: Ok
Those types are duplicated in the backend then ?

comp/syntheticstestscheduler/impl/worker.go
pgimalac: Nevermind I misread it, I thought it was read-locking

comp/syntheticstestscheduler/impl/worker.go
pgimalac: Ok fair haha

pgimalac (approved):
LGTM

comp/syntheticstestscheduler/common/result.go
Fabrice-Matrat: yep

PR #40632, opened 2025-09-05
[CWS-5081] Fix wrong kill status when process failed to be killed

theop-dd:
### What does this PR do?
**This PR fixes a bug.**
During a kill action, if the kill failed, the action was still flagged as "performed".

**Details of the bug fix:**

1) When a process needs to be killed but the kill action must be delayed (by the disarmer), it is placed in the KillQueue.

2) When the process is allowed to be killed, the kill is attempted on all processes in the queue.

3) The algorithm then assumed that every process in this map was successfully killed (and flagged the actions as "performed"), without checking the return value of the actual kill.

This PR introduces a mechanism to check whether the kill was successful.
If the kill fails, the process is removed from the queue and its status is properly updated to "error".

### DIFF
--- a/pkg/security/probe/process_killer.go
+++ b/pkg/security/probe/process_killer.go
@@ -116,25 +116,32 @@ func (p *ProcessKiller) getRuleStats(ruleID string) *processKillerStats {
 	return stats
 }
 
-func (p *ProcessKiller) updatePendingReportKillPerformed(now time.Time, killQueue *[]killContext, failedToBeKilled []uint32) {
+func (p *ProcessKiller) updatePendingReportKillPerformed(now time.Time, killQueue *[]killContext, failedToBeKilled []uint32, nbOfKilled int64) {
 	p.Lock()
 	defer p.Unlock()
 
-	// First we need to update the status of the processes that failed to be killed
 	for _, pid := range failedToBeKilled {
 		// First we remove them from the killQueue
 		*killQueue = slices.DeleteFunc(*killQueue, func(kc killContext) bool {
 			return kc.pid == int(pid)
 		})
 	}
-	// Then we update the status to KillActionStatusError
+	// Then we update the status
 	for _, report := range p.pendingReports {
 		report.Lock()
-		// if pid in failedToKillPids, update the status to KillActionStatusError
-		if slices.Contains(failedToBeKilled, report.Pid) {
-			report.Status = KillActionStatusError
-		} else if report.Status == KillActionStatusQueued {
-			if slices.ContainsFunc(*killQueue, func(kc killContext) bool {
+		if report.Status == KillActionStatusQueued {
+			if slices.Contains(failedToBeKilled, report.Pid) {
+				if nbOfKilled > 0 {
+					report.Status = KillActionStatusPartiallyPerformed
+					fmt.Printf("Partially failed to kill process %d\n", report.Pid)
+					report.KilledAt = now
+					continue
+				} else {
+					report.Status = KillActionStatusError
+					fmt.Printf("Failed to kill process %d\n", report.Pid)
+					continue
+				}
+			} else if slices.ContainsFunc(*killQueue, func(kc killContext) bool {
 				return kc.pid == int(report.Pid)
 			}) {
 				report.Status = KillActionStatusPerformed
@@ -176,8 +183,8 @@ func (p *ProcessKiller) killQueuedPidsAndGetNextAlarm(disarmer *ruleDisarmer, no
 	}
 
 	if now.After(disarmer.killQueueAlarm) {
-		failedToBeKilled := p.KillProcesses(false, disarmer.ruleID, disarmer.killSignal, disarmer.killQueue)
-		p.updatePendingReportKillPerformed(now, &disarmer.killQueue, failedToBeKilled)
+		failedToBeKilled, nbOfKilled := p.KillProcesses(false, disarmer.ruleID, disarmer.killSignal, disarmer.killQueue)
+		p.updatePendingReportKillPerformed(now, &disarmer.killQueue, failedToBeKilled, nbOfKilled)
 	} else {
 		if nextAlarm == nil || disarmer.killQueueAlarm.Before(*nextAlarm) {
 			return &disarmer.killQueueAlarm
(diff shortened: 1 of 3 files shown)


pkg/security/probe/process_killer.go
```diff
@@ -443,18 +443,38 @@ func (p *ProcessKiller) KillProcesses(killDirectly bool, ruleID string, sig int,
...
+		p.Lock()
+		for _, report := range p.pendingReports {
+			// if pid in failedToKillPids, update the status to KillActionStatusError
+			if slices.Contains(failedToKillPids, report.Pid) {
+				report.Status = KillActionStatusError
+			}
+		}
+		p.Unlock()
+	}
+	fmt.Printf("At that point, processesKilled: %d\n", processesKilled)
```
spikat: please remove your debug printfs

pkg/security/probe/process_killer.go
```diff
@@ -443,18 +443,38 @@ func (p *ProcessKiller) KillProcesses(killDirectly bool, ruleID string, sig int,
...
+		})
+		// Then we update the status to KillActionStatusError
+		p.Lock()
+		for _, report := range p.pendingReports {
+			// if pid in failedToKillPids, update the status to KillActionStatusError
+			if slices.Contains(failedToKillPids, report.Pid) {
+				report.Status = KillActionStatusError
+			}
+		}
+		p.Unlock()
```
spikat: I think this has to be done as e separate for loop

pkg/security/probe/process_killer.go
theop-dd: oh yes, I missed that one

spikat (approved):

pkg/security/probe/process_killer.go
```diff
@@ -438,18 +452,20 @@ func (p *ProcessKiller) KillAndReport(kill *rules.KillDefinition, rule *rules.Ru
...
-func (p *ProcessKiller) KillProcesses(killDirectly bool, ruleID string, sig int, kcs []killContext) bool {
+// KillProcesses kills the given list of processes, returns the list of pids that failed to be killed (nil if everything went well)
+func (p *ProcessKiller) KillProcesses(killDirectly bool, ruleID string, sig int, kcs []killContext) []uint32 {
+	var failedToKillPids []uint32
 	if !p.cfg.RuntimeSecurity.EnforcementEnabled {
-		return false
+		return failedToKillPids
 	}
-
 	var processesKilled int64
```
spikat, line 467: IMHO this variable should be deleted now

pkg/security/probe/process_killer.go
```diff
@@ -116,12 +116,24 @@ func (p *ProcessKiller) getRuleStats(ruleID string) *processKillerStats {
...
+			return kc.pid == int(pid)
+		})
+	}
+	// Then we update the status to KillActionStatusError
 	for _, report := range p.pendingReports {
-		if report.Status == KillActionStatusQueued {
+		report.Lock()
+		// if pid in failedToKillPids, update the status to KillActionStatusError
+		if slices.Contains(failedToBeKilled, report.Pid) {
+			report.Status = KillActionStatusError
```
spikat: I'll move those 2 lines in the `if report.Status == KillActionStatusQueued` bellow.

Also, we have multiple pid to kills only when trying to kill containers. If we succeed to kill some of them for me the kill is "performed" (that's fine to say that we killed what was still running and cannot kill what already exited). Or you can add a new "killPerformedPartially" state.

pkg/security/probe/process_killer.go
theop-dd: we still use it for the stats no?

pkg/security/probe/process_killer.go
theop-dd: Yes, I was thinking about a new state because the first issue was about getting a "performed" status when a process was not killed because of a fail in the path resolution for example. I will add that then, thanks

pkg/security/probe/process_killer.go
spikat: ah yes I didn't expend the whole func sry

pkg/security/probe/process_killer.go
```diff
@@ -424,10 +444,16 @@ func (p *ProcessKiller) KillAndReport(kill *rules.KillDefinition, rule *rules.Ru
...
+	if len(failedPids) == 0 {
 		report.KilledAt = now
 		report.Status = KillActionStatusPerformed
 	} else {
+		if nbOfKilled > 0 {
+			report.KilledAt = now
+			report.Status = KillActionStatusPartiallyPerformed
+			log.Warn("some processes failed to be killed in the container with PIDs : ", failedPids)
+		}
 		report.Status = KillActionStatusError
```
spikat: missing the `else` scope for KillActionStatusError here.

pkg/security/probe/process_killer.go
```diff
@@ -116,20 +116,40 @@ func (p *ProcessKiller) getRuleStats(ruleID string) *processKillerStats {
...
 	for _, report := range p.pendingReports {
-		if report.Status == KillActionStatusQueued {
-			if slices.ContainsFunc(*killQueue, func(kc killContext) bool {
-				return kc.pid == int(report.Pid)
-			}) {
-				report.Status = KillActionStatusPerformed
-				report.KilledAt = now
-				continue
+		report.Lock()
+		func() {
```
spikat: why a func here? If we can avoid it it's better IMHO

pkg/security/probe/process_killer.go
```diff
@@ -116,20 +116,40 @@ func (p *ProcessKiller) getRuleStats(ruleID string) *processKillerStats {
...
+func (p *ProcessKiller) updatePendingReportKillPerformed(now time.Time, killQueue *[]killContext, failedToBeKilled []uint32, nbOfKilled int64) {
 	p.Lock()
 	defer p.Unlock()
 
+	for _, pid := range failedToBeKilled {
+		// First we remove them from the killQueue
+		*killQueue = slices.DeleteFunc(*killQueue, func(kc killContext) bool {
+			return kc.pid == int(pid)
+		})
+	}
```
spikat: I'm not sure to understand why we need this first step

pkg/security/probe/process_killer.go
spikat: already fixed on last commit :)

pkg/security/probe/process_killer.go
theop-dd: this func was just a way of getting the unlock in a defer in a simple manner, we can delete it if you think that's better

paulcacheux: the idea of having a defer is great (because of unwind properties of defer), but instead of an inline function you can move this to an actual, top-level, function

PR #40642, opened 2025-09-05
Fix deadlock race condition in `TestAsyncInitialization/successful_async_initialization`

erikayasuda:
#### The Test Bug
`TestAsyncInitialization/successful_async_initialization` was flaking due to a race condition deadlock. This PR fixes the test to un-flake the test. This is not a bug in the ImageResolver feature, but a bug specifically in the way the test is configured.

Prior to this change, the following race condition would cause a deadlock:
1. `mockClient := newMockRCClient(...)` -> `blockGetConfigs`: `false` and `configsReady`: `make(chan struct{})`
2. `mockClient.setBlocking(true)` -> mutex lock **HOLD**, `blockGetConfigs`: `true`, mutex lock **RELEASE**
3. `resolver := newRemoteConfigImageResolverWithRetryConfig(...)`
   - `go func() waitForInitialConfig()` -> `rcClient.GetConfigs(...)` -> mutex lock **HOLD** until `configsReady` closed
4. `mockClient.setBlocking(false)` -> mutex lock **HOLD** ... ❌ cannot close `configsReady` because mutex lock by goroutine.

If step 3 (go routine) would get the mutex lock first, this would create a deadlock situation. If step 4 (test) gets the mutex lock first, then it would successfully close the `configsReady` channel, release the mutex lock, and then the goroutine would hold the mutex lock, do its work, then release the mutex.

#### The Fix
The fix for the above is to explicitly release the mutex hold in `GetConfig(...)` after we store the state of `m.blockGetConfigs` and `m.configsReady` in local variables first. Then we block `GetConfig(...)` on those local vars. This way, regardless of which step obtains the mutex first, it will be released properly.

### Motivation
Un-flake the `TestAsyncInitialization/successful_async_initialization` test for the ImageResolver to avoid blocking CI.

See example of failure from this [here](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/1113117038).

Reproduced CI issue locally with artificial delays. The above fix prevents this race condition consistently.


adel121 (approved):

erikayasuda:
Creating a backport for this fix. The flaky behavior seems to be a lot worse for some reason, and making other backport PRs a bit painful.

PR #40643, opened 2025-09-05
Reapply "[CXP-2599] Enable WLM for process collection by default on linux"

daniel-taf:
### What does this PR do?
The initial [PR](https://github.com/DataDog/datadog-agent/pull/40403) was reverted due to introducing flakes on the `TestLinuxTestSuite/TestManualProcessCheckWithIO` e2e test. Since the change resulted in the process-agent using remote WLM for process data, there is a race condition between the run of the check command on the process agent and whether the collectors actually ran on the core agent. The `workloadmeta.IsInitialized()` will only check if the collectors were started in the current runtime not across others, so we can't rely on it. To fix this, we introduce retries in the test since the race condition has more impact on agent startup.

### Motivation
Fix test flakes + original motivations from https://github.com/DataDog/datadog-agent/pull/40403
e2e tests + unit tests

e2e tests were run multiple times in the hopes of catching flakes. The problematic tests are also marked as flaky to prevent issues on main while we determine if the fix works.

### DIFF
--- a/test/new-e2e/tests/process/testing.go
+++ b/test/new-e2e/tests/process/testing.go
@@ -382,13 +382,7 @@ func assertManualContainerCheck(t require.TestingT, check string, expectedContai
 
 // assertManualProcessDiscoveryCheck asserts that the given process is collected and reported in
 // the output of the manual process_discovery check
-func assertManualProcessDiscoveryCheck(t *testing.T, check string, process string) {
-	defer func() {
-		if t.Failed() {
-			t.Logf("Check output:\n%s\n", check)
-		}
-	}()
-
+func assertManualProcessDiscoveryCheck(t require.TestingT, check string, process string) {
 	var checkOutput struct {
 		ProcessDiscoveries []*agentmodel.ProcessDiscovery `json:"processDiscoveries"`
 	}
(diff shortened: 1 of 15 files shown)


test/new-e2e/tests/process/testing.go
```diff
@@ -382,13 +382,7 @@ func assertManualContainerCheck(t require.TestingT, check string, expectedContai
 
 // assertManualProcessDiscoveryCheck asserts that the given process is collected and reported in
 // the output of the manual process_discovery check
-func assertManualProcessDiscoveryCheck(t *testing.T, check string, process string) {
-	defer func() {
-		if t.Failed() {
-			t.Logf("Check output:\n%s\n", check)
-		}
-	}()
```
daniel-taf, line 390: Removed since we don't have issues with this test anymore and there are better ways to investigate test failures now: https://datadoghq.atlassian.net/wiki/spaces/ADX/pages/3492283118/Troubleshoot+an+e2e+test+failing+on+the+CI.

gengnamstyle (approved):

dustmop (approved):

chouetz (approved):

Yumasi (approved):

zhuminyi (approved):

PR #40645, opened 2025-09-05
[CONTP-960] fix: Scrub secrets in datadog cluster agent 'clusterchecks' cmd

Mathew-Estafanous:
### What does this PR do?

Adds secrets scrubbing to the `agent clusterchecks` command by running all `integration.Data` objects through the scrubber util prior to writing to the output.

Unlike the regular agent, where the api performs scrubbing directly, this PR introduces cluster agent scrubbing **after** the config and endpoint API calls. I considered scrubbing in the API layer (as in the regular agent), but chose not to in order to avoid modifying the API interface with a one-off `scrub_secrets` type flag.

### Motivation

Customers have raised issues with the cluster agent exposing secrets through generated flares/clusterchecks. Sensitive information should be scrubbed with asterisks to prevent undesired exposure of secrets.

See [CONS-7565](https://datadoghq.atlassian.net/browse/CONS-7565) for customer issues.

Manually tested the change by configuring a cluster check on an nginx service using a 'password.' Validated that fix results in the secret being scrubbed vs the old agent which shows the secret in the outputs of flare/clusterchecks.

Steps:
1. Add a dummy service level check secrets (both `ENC[]` and not).

<details><summary>Dummy nginx example</summary>
<p>

**Service definition:**
```
apiVersion: v1
kind: Service
metadata:
  name: nginx-cluster-check-service
  namespace: default
  labels:
    app: nginx-cluster-check
  annotations:
    ad.datadoghq.com/service.checks: |
      {
        "http_check": {
          "init_config": {},
          "instances": [
            {
              "url":"http://%%host%%",
              "token_secret": "this_is_a_secret",
              "password":"ENC[k8s_secret@default/nginx/pwd]",
              "timeout": "ENC[k8s_secret@default/nginx-t/timeout]",
              "name":"My Nginx",
            }
          ]
        }
      }
spec:
  selector:
    app: nginx-cluster-check
  ports:
  - name: http
    port: 80
    targetPort: 80
    protocol: TCP
  type: ClusterIP
```

**Create secrets**
```
kubectl create secret generic nginx --from-literal=pwd=TOP_SECRET &&
kubectl create secret generic nginx-t --from-literal=timeout=1
```

</p>
</details>

2. Enable cluster agent checks

<details><summary>Helm chart</summary>
<p>

```
datadog:
   # rest of agent helm chart config
   clusterChecks:
      enabled: true
   secretBackend:
      command: "/readsecret_multiple_providers.sh"
```

</p>
</details> 

3. Confirm that "some_secret", "password", and "timeout" are all scrubbed from `agent clusterchecks`

<details><summary>Testing Results</summary>
<p>

**Current cluster agent:** 
Running `agent clusterchecks` / `agent flare` exposes the `TOP_SECRET_PASSWORD` in the nginx check config.
image

**Fixed cluster agent**
Running `agent clusterchecks`/ `agent flare` shows scrubbed password using asterisks (*) in place.

**agent clusterchecks**
image

**agent flare (clusterchecks.log)**
image

</p>
</details> 

### Additional Notes

[CONS-7565]: https://datadoghq.atlassian.net/browse/CONS-7565?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


releasenotes/notes/fix_clusterchecks_secrets_scrubbing-4caa0177b3c33851.yaml
```diff
@@ -0,0 +1,11 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+fixes:
+  - |
+    Cluster agent now scrubs secrets when generating flares and cluster check reports.
```
janine-c:
```suggestion
    Cluster Agent now scrubs secrets when generating flares and cluster check reports.
```

janine-c (approved):

janine-c (approved):

pkg/flare/config_check_test.go
```diff
@@ -212,6 +212,78 @@ func TestContainerExclusionRulesInfo(t *testing.T) {
...
+				Instances: []integration.Data{
+					integration.Data(`
+database:
+  host: db.example.com
+  username: dbuser
+  password: dbpass123
+  auth_token: token789
+api:
+  endpoint: https://api.example.com
+  api_key: abcdef1234567890abcdef1234567890
```
dustmop: Will this also get scrubbed? I would expect the scrubber to replace all sub the last 5 characters. If so, can we add this value to the `shouldNotContain` list?

pkg/flare/config_check_test.go
```diff
@@ -212,6 +212,78 @@ func TestContainerExclusionRulesInfo(t *testing.T) {
...
+  privKey: snmppriv
+`),
+				},
+				ClusterCheck: true,
+			},
+			shouldNotContain: []string{
+				"dbpass123",
+				"token789",
+				"apisecret",
+				"snmpcommunity",
```
dustmop: What is the reason that this field gets scrubbed?

Mathew-Estafanous: The [scrubber logic](https://github.com/DataDog/datadog-agent/blob/b5c4b6c7fcd47a9e9d938e5b8bdc62b6281273ca/pkg/util/scrubber/default.go#L171) treats the snmp `community_string` as a secret hence why I've included it in this list.

pkg/flare/config_check_test.go
Mathew-Estafanous: Yes it should. Thanks for catching. Adding now.

pkg/flare/config_check_test.go
dustmop: Ah, turns out it's the term used by snmp to refer to the basic authentication mechanism: https://blog.domotz.com/product-bytes/snmp-community-strings/. Today I learned

dustmop (approved):

pkg/flare/config_check_test.go
Mathew-Estafanous: TIL as well. 😄  thanks for the review.

Mathew-Estafanous:
Change of plans. Scrubbing logic will be added to the api layer with a `scrub` query flag in the `/clusterchecks` endpoint. @davidor highlighted that keys with non-secret names but use `ENC[]` will still have their secrets shown because it's is resolved and returned as a response by the API. Scrubbing after that point misses the fact that a key may have used `ENC[]` and has to rely only on the key names.

`/clusterchecks` default will return un-scrubbed configs so that the behaviour remains the same with other calls. Using an explicit `scrub = true` flag will return scrubbed configs.

davidor:
@Mathew-Estafanous thanks for fixing the ENC issue. I noticed something in the new screenshots that you shared. In the `agent clusterchecks` output, there’s now an empty configuration shown before the `http_check` one. That wasn’t present before your changes.

Mathew-Estafanous:
@davidor thanks for finding. Pushed a fix that resolves it and updated the PR to reflect that.

image

image

releasenotes/notes/fix_clusterchecks_secrets_scrubbing-4caa0177b3c33851.yaml
```diff
@@ -0,0 +1,11 @@
...
+# CHANGELOG.rst is rendered. So the text needs to be worded so that
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+fixes:
+  - |
+    Cluster Agent now scrubs secrets when generating flares and cluster check reports.
```
davidor: This only affects the cluster-agent, not the node-agent. When that's the case, release notes go in `releasenotes-dca/notes`.

Also, maybe we should be more specific here. flares where working as expected. The fix is for the `agent clusterchecks` command.

Mathew-Estafanous: moved the release note. thanks

Mathew-Estafanous: i think flares on the cluster agent didn't scrub everything. See this flare that I ran using the latest cluster agent. `token_secret` isn't scrubbed vs this fix where it is.

**latest cluster agent:**
image

**built cluster agent:**
image

davidor (approved):

PR #40653, opened 2025-09-05
add exponential backoff to the sending of the host metadata payload

s-alad:
### What does this PR do?

Front-loads host-payload attempts exponentially during the start to speed up node-level tag availability. Attempts are then stabilized at a normal collection interval. 

### Motivation

Customers noticed that they sometimes have to wait a considerable amount before node-level tags are available during fresh deployments due to the host payload only being sent every 30 minutes, so a miss early on could push tag availability back significantly.

1. Unit Tests to test proper intervals
2. Running the binary with custom configured interval times and grep-ing for the debug line showing when the next run should be
```
metadata_providers:
  - name: host
    interval: 1800
    early_interval: 600
```
```
dda inv agent.build && DD_LOG_LEVEL=debug stdbuf -oL -eL ./bin/agent/agent run -c ./dev/dist 2>&1 \
| grep -F --line-buffered --color=always 'Next host metadata collection scheduled'
```

### DIFF
--- a/comp/metadata/host/hostimpl/host.go
+++ b/comp/metadata/host/hostimpl/host.go
@@ -13,6 +13,7 @@ import (
 	"path/filepath"
 	"time"
 
+	"github.com/cenkalti/backoff"
 	"go.uber.org/fx"
 
 	api "github.com/DataDog/datadog-agent/comp/api/api/def"
@@ -48,9 +49,9 @@ type host struct {
 	resources    resources.Component
 	hostnameComp hostnameinterface.Component
 
-	hostname        string
-	collectInterval time.Duration
-	serializer      serializer.MetricSerializer
+	hostname      string
+	serializer    serializer.MetricSerializer
+	backoffPolicy *backoff.ExponentialBackOff
 }
 
 // Module defines the fx options for this component.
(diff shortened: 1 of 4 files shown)


iadjivon (approved):

comp/metadata/host/hostimpl/host.go
```diff
@@ -102,14 +103,26 @@ func newHostProvider(deps dependencies) provides {
 	}
 
 	hname, _ := deps.Hostname.Get(context.Background())
+
+	// exponential backoff for collection intervals which arrives at user's configured interval
+	bo := &backoff.ExponentialBackOff{
+		InitialInterval:     5 * time.Minute,
```
dustmop: I think we need to add another config option for the minimal. Right now there's only [an interval](https://github.com/DataDog/datadog-agent/blob/37212597f9e1a0c96de628a498614c3540cbb3b7/pkg/config/config_template.yaml#L2506), but we should add an `early_interval` for providers that use shorter interval earlier on. You'll need to add a field to [this struct](https://github.com/DataDog/datadog-agent/blob/37212597f9e1a0c96de628a498614c3540cbb3b7/pkg/config/utils/metadata_providers.go#L18) as well, that's where the value is pulled from the config.

The loop above retrieves `collectInterval` from the config, it will need to do the same for `earlyCollectInterval`, and should validate that `early <= collect`.

comp/metadata/host/hostimpl/host.go
```diff
@@ -126,10 +139,17 @@ func newHostProvider(deps dependencies) provides {
 
 func (h *host) collect(ctx context.Context) time.Duration {
 	payload := h.getPayload(ctx)
+
+	nextInterval := h.backoffPolicy.NextBackOff()
+	if nextInterval <= 0 || nextInterval > h.backoffPolicy.MaxInterval {
+		nextInterval = h.backoffPolicy.MaxInterval
+	}
+
```
dustmop, line 164: Please add a `log.Debug` line here that outputs what the `nextInterval` value is. That will make it easier to validate this behavior by running the agent and grep'ing the logs.

dustmop (changes requested):
We discussed a bit offline, I think this mostly looks good, but we need to do a bit more work to define how this works with the existing configuration settings for metadata collection.

comp/metadata/host/hostimpl/host.go
s-alad: should we keep the multiplier constant (it's currently 3) or should we force all intervals to have a certain amount of "retries" by doing some math

dustmop: I think keeping it simple for now is totally fine, let's just stick with the given multiplier.

dustmop (approved):
Great, thanks for the fixes. Looks good to me!

PR #40654, opened 2025-09-05
ignore late service does not exist

clarkb7:
### What does this PR do?
ignores "service not found" error when stopping services in the MSI

### Motivation
https://datadoghq.atlassian.net/browse/WINA-1776

added unit tests

### Additional Notes
Seems like it shouldn't occur because we only try to stop services that are returned to us from `ServiceController.GetServices()`, but we saw it happen in telemetry.

### DIFF
--- a/tools/windows/DatadogAgentInstaller/CustomActions/ServiceCustomAction.cs
+++ b/tools/windows/DatadogAgentInstaller/CustomActions/ServiceCustomAction.cs
@@ -50,6 +50,30 @@ namespace Datadog.CustomActions
             }
         }
 
+        public static bool IsServiceDoesNotExistError(Exception exception)
+        {
+            // Walk the exception chain to find a Win32Exception with ERROR_SERVICE_DOES_NOT_EXIST (1060)
+            var current = exception;
+            while (current != null)
+            {
+                if (current is System.ComponentModel.Win32Exception wex && wex.NativeErrorCode == 1060)
+                {
+                    return true;
+                }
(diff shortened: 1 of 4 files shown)


tools/windows/DatadogAgentInstaller/CustomActions/ServiceCustomAction.cs
```diff
@@ -50,6 +50,30 @@ public ServiceCustomAction(
...
+                if (current is System.ComponentModel.Win32Exception wex && wex.NativeErrorCode == 1060)
+                {
+                    return true;
+                }
+                // Fallback: some ServiceController operations surface as InvalidOperationException with a descriptive message
+                // System.ComponentModel.Win32Exception: The specified service does not exist as an installed service
+                if (current is InvalidOperationException ioe &&
+                    ioe.Message != null &&
+                    // string.Contains doesn't have a ignore-case option
+                    ioe.Message.IndexOf("does not exist", StringComparison.InvariantCultureIgnoreCase) >= 0)
```
alexn-dd, line 68: System error messages are most likely localized, but this is fine for the current scope.
May want to revisit this later.

alexn-dd (approved):

tools/windows/DatadogAgentInstaller/CustomActions/ServiceCustomAction.cs
clarkb7: yep, this is just meant to be a fallback. The primary handler is based on `wex.NativeErrorCode`

PR #40660, opened 2025-09-07
Extend the allowed duration of Bazel server inactivity

ofek:
### Motivation

I would like speedy Bazel interactions throughout intermittent working sessions to be the default customer experience.


.bazelrc
```diff
@@ -1,5 +1,8 @@
 # Do not edit this file without a review from @DataDog/agent-build
 
+# Keep the server alive for at most 8 hours of inactivity
+startup --max_idle_secs=28800
```
JSGette, line 4: Afair, bazel resets the timer each time you run a command, so setting this to higher values is not needed most of the time. I have no prior experience bumping this value, but considering how greedy bazel can be when it comes to RAM consumption it may actually make the experience worse.

I don't have a strong opinion on that, therefore, approved. Worst case we will set it back to default (which is 4 hours if I recall correctly)

JSGette (approved):

.bazelrc
rdesgroppes: I'm wondering what observation led to the change.

ofek: The default is 3 hours so a few times in the past few weeks I would run something, go to a meeting or two and then when I come back the first invocation is slow again.

rdesgroppes: Really?

PR #40666, opened 2025-09-08
[incident-42773] Fix cache errors on git merge-base operations (main)

CelianR:
### What does this PR do?

This fixes issues with S3 cache by fetching the target branch when doing `git merge-base` operations.

### DIFF
--- a/tasks/libs/common/git.py
+++ b/tasks/libs/common/git.py
@@ -148,7 +148,7 @@ def get_default_branch(major: int | None = None):
     return '6.53.x' if major is None and is_agent6(ctx) or major == 6 else 'main'
 
 
-def get_common_ancestor(ctx, branch, base=None, try_fetch=True) -> str:
+def get_common_ancestor(ctx, branch, base=None, try_fetch=True, hide=True) -> str:
     """
     Get the common ancestor between two branches.
 
@@ -165,15 +165,15 @@ def get_common_ancestor(ctx, branch, base=None, try_fetch=True) -> str:
     base = (base or get_default_branch()).removeprefix("origin/")
 
     try:
-        return ctx.run(f"git merge-base {branch} origin/{base}", hide=True).stdout.strip()
+        return ctx.run(f"git merge-base {branch} origin/{base}", hide=hide).stdout.strip()
     except Exception:
         if not try_fetch:
             raise
 
         # With S3 caching, it's possible that the base branch is not fetched
-        ctx.run(f"git fetch origin {base}")
+        ctx.run(f"git fetch origin {base}", hide=hide)
 
-        return ctx.run(f"git merge-base {branch} origin/{base}", hide=True).stdout.strip()
+        return ctx.run(f"git merge-base {branch} origin/{base}", hide=hide).stdout.strip()
 
 
 def check_uncommitted_changes(ctx):
@@ -202,7 +202,7 @@ def get_commit_sha(ctx, commit="HEAD", short=False) -> str:
 
 def get_main_parent_commit(ctx) -> str:
     """
-    Get the commit sha your current branch originated from
+    Get the commit sha from the LCA between main and the current branch
     """
     return get_common_ancestor(ctx, "HEAD", f'origin/{get_default_branch()}')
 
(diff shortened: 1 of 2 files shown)


chouetz (approved):

tasks/libs/common/git.py
```diff
@@ -148,10 +148,32 @@ def get_default_branch(major: int | None = None):
...
-def get_common_ancestor(ctx, branch, base=None) -> str:
-    base = base or f"origin/{get_default_branch()}"
+def get_common_ancestor(ctx, branch, base=None, try_fetch=True) -> str:
+    """
+    Get the common ancestor between two branches.
+
+    Args:
+        ctx: The invoke context.
+        branch: The branch to get the common ancestor with.
+        base: The base branch to get the common ancestor with.
```
pgimalac:
```suggestion
        base: The base branch to get the common ancestor with. Defaults to the default branch.
```

tasks/libs/common/git.py
```diff
@@ -148,10 +148,32 @@ def get_default_branch(major: int | None = None):
...
+
+    Returns:
+        The common ancestor between two branches.
+    """
+
+    base = (base or get_default_branch()).removeprefix("origin/")
+
+    try:
+        return ctx.run(f"git merge-base {branch} origin/{base}", hide=True).stdout.strip()
+    except Exception:
```
pgimalac, line 169: Are you sure it will raise an exception ? In a previous incident it just returned the wrong merge-base commit

tasks/libs/common/git.py
```diff
@@ -148,10 +148,32 @@ def get_default_branch(major: int | None = None):
...
+        try_fetch: Try to fetch the base branch if it's not found (to avoid S3 caching issues).
+
+    Returns:
+        The common ancestor between two branches.
+    """
+
+    base = (base or get_default_branch()).removeprefix("origin/")
+
+    try:
+        return ctx.run(f"git merge-base {branch} origin/{base}", hide=True).stdout.strip()
```
pgimalac: Could we add a parameter to decide whether to hide ?
We had an incident where merge-base returned the wrong commit, and not having the output made it annoying to check

tasks/libs/common/git.py
```diff
@@ -148,10 +148,32 @@ def get_default_branch(major: int | None = None):
...
+        ctx: The invoke context.
+        branch: The branch to get the common ancestor with.
+        base: The base branch to get the common ancestor with.
+        try_fetch: Try to fetch the base branch if it's not found (to avoid S3 caching issues).
+
+    Returns:
+        The common ancestor between two branches.
+    """
+
+    base = (base or get_default_branch()).removeprefix("origin/")
```
pgimalac, line 165: Should it default to the base branch instead of the default branch ? Main difference being on release branches

tasks/libs/common/git.py
pgimalac: Maybe base branch instead depending on the answer to my other comment

tasks/libs/common/git.py
```diff
@@ -182,7 +204,7 @@ def get_main_parent_commit(ctx) -> str:
     """
     Get the commit sha your current branch originated from
     """
-    return ctx.run(f"git merge-base HEAD origin/{get_default_branch()}", hide=True).stdout.strip()
+    return get_common_ancestor(ctx, "HEAD", f'origin/{get_default_branch()}')
```
pgimalac, line 207: Should this compare to the base branch instead of the default one ?

tasks/libs/common/git.py
CelianR: It works now with that on 7.71.x, could you link to me the incident to take a look?

tasks/libs/common/git.py
CelianR: It was on the default branch initially so I'm not changing the default behavior

tasks/libs/common/git.py
CelianR: I think that the name of the function is more explicit than the description, I'll update the description

CelianR: (I don't want to change the behavior on an incident fix)

pgimalac (approved):
I still think we should change the default to use base branch instead of default branch, but since it's the current behavior we can discuss that separately

PR #40669, opened 2025-09-08
fix(assign_issue): Add a missing dependency at runtime

chouetz:
### What does this PR do?
- Add `yq` at runtime during the `assign_issue` workflow

### Motivation
- `yq` is used by `dda` and the workflow is currently [failing](https://github.com/DataDog/datadog-agent/actions/runs/17498675682/job/49705999064)
- We install it at runtime because modification of the image is not [straightforward](https://datadoghq.atlassian.net/wiki/x/IQM0CQE)
- This is intended to be temporary. We should get yq in a better way in the future.

I need this to be merged before I can retrigger the workflow on an issue. The modification is quite simple.

### DIFF
--- a/.github/workflows/assign_issue.yml
+++ b/.github/workflows/assign_issue.yml
@@ -21,6 +21,11 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+    - name: Install yq
+      run: |
+        sudo add-apt-repository ppa:rmescandon/yq
+        sudo apt update
+        sudo apt install yq -y
     - name: Install dda
       uses: ./.github/actions/install-dda
       with:


.github/workflows/assign_issue.yml
```diff
@@ -21,6 +21,11 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+    - name: Install yq
+      run: |
+        sudo add-apt-repository ppa:rmescandon/yq
```
aiuto, line 26: From a supply chain point of view, I don't like adding new repositories. The AI overview says this is no longer maintained. https://www.google.com/search?q=rmescandon%2Fyq&rlz=1C5GCCM_en&oq=rmescandon%2Fyq&gs_lcrp=EgZjaHJvbWUyBggAEEUYOTIKCAEQABiABBiiBDIKCAIQABiABBiiBDIKCAMQABiiBBiJBTIHCAQQABjvBTIKCAUQABiiBBiJBTIGCAYQRRg60gEHODIxajBqN6gCCLACAfEFmYi0GaUmkeQ&sourceid=chrome&ie=UTF-8

This may be fine today, but it should be replaced in the future with something we have a better handle on. If we really need yq, I would rather vendor it into our own repo to build and distribute it.

For now, I'ld be OK with just a Jira ticket to replace it as a TODO comment here.

chouetz: Yes this workflow should be dismissed soon in favour of a different alternative. This is just an ugly workaround letting me the time to create the new one 😄

aiuto (approved):

CelianR (approved):

PR #40688, opened 2025-09-08
COAT: making relevant packages into modules for external use

truthbk:
### What does this PR do?

This PR converts the `agenttelemetetry` component into a module for external use. We need to do the same with a number of adjacent modules for the same reason.  

### Motivation

Use of COAT in OSS OTel Collector.

Successful tests and builds should be sufficient as no functional code changes are introduced.


go.work
```diff
@@ -7,6 +7,11 @@ godebug tlsmlkem=0
 
 use (
 	.
+	./comp/core/agenttelemetry/def
+	./comp/core/agenttelemetry/fx
+	./comp/core/agenttelemetry/impl
+	./pkg/util/hostinfo
+	./pkg/util/jsonquery
```
pgimalac: I think the file was previously sorted, could you re-sort it ?
EDIT: actually with the `./` it might currently be sorted, but to align with the rest it might be best to remove the `./`

pkg/util/hostinfo/hostinfo.go
```diff
@@ -0,0 +1,19 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+//
+// # Compatibility
+//
+// This module is exported and can be used outside of the datadog-agent
+// repository, but is not designed as a general-purpose logging system.  Its
+// API may change incompatibly.
```
pgimalac, line 10: Not sure if we need this in the header, we already have 150 modules without this 😄

pgimalac (approved):

iglendd (approved):
Good for my ownership

chouetz (approved):

hush-hush (approved):

pimlu (approved):

PR #40544, opened 2025-09-04
[USMON-1483] USM Redis: Migrate configuration to tree structure

guyarb:
### What does this PR do?

Migrates Redis monitoring configuration from flat structure to nested tree structure, changing `service_monitoring_config.enable_redis_monitoring` to `service_monitoring_config.redis.enabled`.

### Motivation

This is the first change in the effort to migrate protocol configurations into a tree structure, following the pattern established with the TLS tree migration (PR #19515). The nested structure improves organization and provides a namespace for protocol-specific configurations like `redis.track_resources`.

### Describe how you validated your changes

- **Configuration binding**: Updated `pkg/config/setup/system_probe.go` to use `BindEnvAndSetDefault` with nested path
- **Config usage**: Updated `pkg/network/config/config.go` to read from nested path  
- **Tests**: Updated existing `TestEnableRedisMonitoring` to use new configuration paths
- **Default value**: Maintains `false` default for Redis monitoring

### Additional Notes

- **No breaking change**: Redis monitoring was never released, so no backward compatibility needed
- **Tree structure**: Creates foundation for future protocol config migrations (HTTP, HTTP2, Kafka, Postgres)
- **Configuration paths**:
  - Old: `service_monitoring_config.enable_redis_monitoring`
  - New: `service_monitoring_config.redis.enabled`
- **Environment variables**:
  - Old: `DD_SERVICE_MONITORING_CONFIG_ENABLE_REDIS_MONITORING`
  - New: `DD_SERVICE_MONITORING_CONFIG_REDIS_ENABLED`

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #19515, opened 2023-09-18
usm: tls: Change enable_https_monitoring into new tls tree

guyarb:
### What does this PR do?

Unifies USM tls configuration under new tls tree.

Old:
```yaml
network_config:
  enable_https_monitoring: true

service_monitoring_config:
  enabled: true
  enable_go_tls_support: true
  java_tls:
    enabled: true
    debug: true
    other configuration values ...
```

New:
```yaml
service_monitoring_config:
  enabled: true
  tls:
    java:
      enabled: true
      other flags....
    go:
      enabled: true
      other flags....
    native:
      enabled: true
      other flags....
```

### Motivation
Making the configuration simpler for users.

### Additional Notes
Each change (native, go, java) was done in different commits, suggested to review each commit on its on.
For java & istio I didn't add an entry in the release note, as it is still an internal feature.

### Reviewer's Checklist

- [ ] If known, an appropriate milestone has been selected; otherwise the `Triage` milestone is set.
- [ ] Use the `major_change` label if your change either has a major impact on the code base, is impacting multiple teams or is changing important well-established internals of the Agent. This label will be use during QA to make sure each team pay extra attention to the changed behavior. For any customer facing change use a releasenote.
- [ ] A [release note](https://github.com/DataDog/datadog-agent/blob/main/docs/dev/contributing.md#reno) has been added or the `changelog/no-changelog` label has been applied.
- [ ] Changed code has automated tests for its functionality.
- [ ] Adequate QA/testing plan information is provided if the `qa/skip-qa` label is not applied.
- [ ] At least one `team/..` label has been applied, indicating the team(s) that should QA this change.
- [ ] If applicable, docs team has been notified or [an issue has been opened on the documentation repo](https://github.com/DataDog/documentation/issues/new).
- [ ] If applicable, the `need-change/operator` and `need-change/helm` labels have been applied.
- [ ] If applicable, the `k8s/<min-version>` label, indicating the lowest Kubernetes version compatible with this feature.
- [ ] If applicable, the [config template](https://github.com/DataDog/datadog-agent/blob/main/pkg/config/config_template.yaml) has been updated.

PR #40689, opened 2025-09-08
[USMON-1482] USM: postgres: Migrate configuration to tree structure

guyarb:
### What does this PR do?

Migrates Postgres monitoring configuration from flat structure to nested tree structure, changing `service_monitoring_config.enable_postgres_monitoring` to `service_monitoring_config.postgres.enabled`.

### Motivation

This continues the effort to migrate protocol configurations into a tree structure, following the pattern established with the TLS tree migration (PR #19515) and Redis migration (PR #40544). The nested structure improves organization and provides a namespace for protocol-specific configurations like `postgres.max_stats_buffered` and `postgres.max_telemetry_buffer`.

- **Configuration binding**: Updated `pkg/config/setup/system_probe.go` to use `BindEnvAndSetDefault` with nested path
- **Config usage**: Updated `pkg/network/config/usm_config.go` to read from nested paths  
- **Tests**: Updated existing `TestEnablePostgresMonitoring` and `TestMaxPostgres*` to use new configuration paths
- **Default values**: Maintains existing defaults for all Postgres configurations

### Additional Notes

- **Breaking change**: Postgres monitoring is already released, but this follows clean migration pattern like Redis
- **Tree structure**: Creates foundation for future protocol config migrations alongside existing Redis and TLS structures
- **Configuration paths**:
  - Old: `service_monitoring_config.enable_postgres_monitoring`
  - New: `service_monitoring_config.postgres.enabled`
  - Old: `service_monitoring_config.max_postgres_stats_buffered`
  - New: `service_monitoring_config.postgres.max_stats_buffered`
  - Old: `service_monitoring_config.max_postgres_telemetry_buffer`
  - New: `service_monitoring_config.postgres.max_telemetry_buffer`
- **Environment variables**:
  - Old: `DD_SERVICE_MONITORING_CONFIG_ENABLE_POSTGRES_MONITORING`
  - New: `DD_SERVICE_MONITORING_CONFIG_POSTGRES_ENABLED`
  - Old: `DD_SERVICE_MONITORING_CONFIG_MAX_POSTGRES_STATS_BUFFERED`
  - New: `DD_SERVICE_MONITORING_CONFIG_POSTGRES_MAX_STATS_BUFFERED`
  - Old: `DD_SERVICE_MONITORING_CONFIG_MAX_POSTGRES_TELEMETRY_BUFFER`
  - New: `DD_SERVICE_MONITORING_CONFIG_POSTGRES_MAX_TELEMETRY_BUFFER`


pkg/config/setup/system_probe.go
```diff
@@ -277,8 +277,8 @@ func InitSystemProbeConfig(cfg pkgconfigmodel.Setup) {
 	cfg.BindEnv(join(netNS, "max_http_stats_buffered"), "DD_SYSTEM_PROBE_NETWORK_MAX_HTTP_STATS_BUFFERED")
 	cfg.BindEnv(join(smNS, "max_http_stats_buffered"))
 	cfg.BindEnvAndSetDefault(join(smNS, "max_kafka_stats_buffered"), 100000)
-	cfg.BindEnv(join(smNS, "max_postgres_stats_buffered"))
-	cfg.BindEnvAndSetDefault(join(smNS, "max_postgres_telemetry_buffer"), 160)
+	cfg.BindEnv(join(smNS, "postgres", "max_stats_buffered"))
```
hush-hush: Since we're changing this, could we add a default for it ?

pkg/network/config/usm_config.go
```diff
@@ -191,9 +191,9 @@ func NewUSMConfig(cfg model.Config) *USMConfig {
 		MaxKafkaStatsBuffered: cfg.GetInt(sysconfig.FullKeyPath(smNS, "max_kafka_stats_buffered")),
 
 		// Postgres Protocol Configuration
-		EnablePostgresMonitoring:   cfg.GetBool(sysconfig.FullKeyPath(smNS, "enable_postgres_monitoring")),
-		MaxPostgresStatsBuffered:   cfg.GetInt(sysconfig.FullKeyPath(smNS, "max_postgres_stats_buffered")),
-		MaxPostgresTelemetryBuffer: cfg.GetInt(sysconfig.FullKeyPath(smNS, "max_postgres_telemetry_buffer")),
+		EnablePostgresMonitoring:   cfg.GetBool(sysconfig.FullKeyPath(smNS, "postgres", "enabled")),
+		MaxPostgresStatsBuffered:   cfg.GetInt(sysconfig.FullKeyPath(smNS, "postgres", "max_stats_buffered")),
+		MaxPostgresTelemetryBuffer: cfg.GetInt(sysconfig.FullKeyPath(smNS, "postgres", "max_telemetry_buffer")),
```
hush-hush, line 196: What about legacy configuration ? Are those setting already used by customer (documented in some way) ?

We don't want configuration to break upon upgrading the Agent (since we are only bumping the minor and not the major when releasing a new Agent). Usually for such change we add a small piece of logic that coalesce old and new setting with precedence for the old one.

hush-hush (changes requested):

pkg/network/config/usm_config.go
guyarb: No one is using it yet, that's why we feel comfortable breaking it

pkg/config/setup/system_probe.go
guyarb: The default is set in `adjust_usm`

guyarb: https://github.com/DataDog/datadog-agent/blob/ee14fe677ddc29219b38ae9a3e008e918cbb84fe/pkg/system-probe/config/adjust_usm.go#L73

hush-hush: Ho, we should not do that. `applyDefault` is basically re-implementing the entire config behavior for defaults but with the wrong source. We should remove it as soon as possible. Defaults are always overwritten by values from users.

A default is different that a setting enforce at runtime by the Agent. Also. this prevent us from creating a schema for the config in the future.

guyarb: I can agree with that, but it should be done in a separate PR to eliminate usages of applyDefault

hush-hush: I agree, can we just not use `applyDefault` for the newly added setting `postgres.max_stats_buffered` to move us one step closer to out goal ?

guyarb: Removed

amitslavin (approved):

hush-hush (approved):

PR #40695, opened 2025-09-08
pin go mod versions and update go mod name comp/core/ipc/impl

jackgopack4:
### What does this PR do?
pin gomods to fixed version when able and fix name of comp/core/ipc/impl (was previously named comp/core/ipc/ipcimpl in gomod but was not published properly)
### Motivation
using datadog-agent dependencies downstream (such as in OpenTelemetry Collector components)
`dda inv tidy` with no go.sum changes


comp/core/configsync/go.mod
```diff
@@ -4,14 +4,14 @@ go 1.24.0
 
 require (
 	github.com/DataDog/datadog-agent/comp/core/config v0.64.0-devel
-	github.com/DataDog/datadog-agent/comp/core/ipc/def v0.0.0-00010101000000-000000000000
-	github.com/DataDog/datadog-agent/comp/core/ipc/httphelpers v0.0.0-00010101000000-000000000000
-	github.com/DataDog/datadog-agent/comp/core/ipc/mock v0.0.0-00010101000000-000000000000
+	github.com/DataDog/datadog-agent/comp/core/ipc/def v0.69.3
+	github.com/DataDog/datadog-agent/comp/core/ipc/httphelpers v0.69.4
+	github.com/DataDog/datadog-agent/comp/core/ipc/mock v0.70.0
```
pgimalac: Any particular reason to pin different versions ?

pgimalac (approved):

comp/core/configsync/go.mod
jackgopack4: v0.70.0 wasn't published yet on pkg.go.dev, I will try a go get v0.70.0 for all of these and if it works, will do that instead

PR #40697, opened 2025-09-08
Add `image_resolution_attempts` telemetry for `ImageResolver.Resolve()`

erikayasuda:
### What does this PR do?
Adds `image_resolution_attempts` telemetry to the newly added `ImageResolver.Resolve()` function to track image resolution attempts, and whether they resulted in a successful image digest resolution, or if they defaulted to using the mutable tag.

### Motivation
This telemetry is necessary for the new K8s SSI gradual rollout feature in order to determine whether a rollout is working successfully or not. The goal is for this to be easier to display on a dashboard:

| `registry` | `repository` | `digest_resolution` | `outcome` | Notes |
|----------|----------|----------|----------|----------|
| `gcr.io/datadoghq`  | `dd-lib-python-init`  | `enabled`  | `sha256:abc123` | ✅ Was supposed to resolve, did resolve |
| `hub.docker.com/r/datadog`  | `dd-lib-java-init`  | `enabled`  | `v2` | ❌  Was supposed to resolve, but it did NOT resolve |
| `mycustomregistry.org`  | `dd-lib-php-init`  | `enabled`  | `v2` | ✅   Cannot resolve for custom registry, did NOT resolve |
| `gallery.ecr.aws/datadog`  | `dd-lib-rb-init`  | `disabled`  | `v1` | ✅   Was NOT supposed to resolve, did NOT resolve |
| `gcr.io/datadoghq`  | `dd-lib-dotnet-init`  | `disabled`  | `sha256:abc123` | ❌  Was NOT supposed to resolve, did resolve *this shouldn't be possible |

Ran a local app via `injector-dev` to verify that the telemetry counts for `apm-inject` and `dd-lib-python-init` were accurate.


iadjivon (approved):

releasenotes/notes/add-image-resolution-attempt-telemetry-a476599d78191e61.yaml
```diff
@@ -0,0 +1,5 @@
+---
+enhancements:
```
adel121: I don't think it is necessary to add a release note here because this change is directly related to the whole image tag resolution feature which already has a release note in the same agent release.

adel121 (approved):

releasenotes/notes/add-image-resolution-attempt-telemetry-a476599d78191e61.yaml
erikayasuda: OK makes sense, wasn't sure so I just opted to add one. I'll remove it!

pkg/clusteragent/admission/metrics/metrics.go
```diff
@@ -106,4 +118,9 @@ var (
 	PatchErrors = telemetry.NewCounterWithOpts("admission_webhooks", "patcher_errors",
 		[]string{}, "Number of patch errors.",
 		telemetry.Options{NoDoubleUnderscoreSep: true})
+
+	// Image resolution tracking for gradual rollout monitoring
+	ImageResolutionAttempts = telemetry.NewCounterWithOpts("admission_webhooks", "image_resolution_attempts",
```
stanistan, line 117: Would we want to add the actual `sha` or tag that's being resolved to? That way we can see which tags are being set.

erikayasuda: I think that makes more sense, I'll update accordingly. I also updated the PR description with what I think the data should help us understand with your suggestion. Let me know what you think!

PR #40699, opened 2025-09-08
add ec2_prioritize_instance_id_as_hostname option to template config

s-alad:
### What does this PR do?
Adds `ec2_prioritize_instance_id_as_hostname` to the config template as it's a feature that can be configured.
 
### Motivation
configuration options we want the customer to potentially use for the agent should be on the example datadog.yaml template file. 

No code was changed.

### DIFF
--- a/pkg/config/config_template.yaml
+++ b/pkg/config/config_template.yaml
@@ -450,10 +450,12 @@ api_key:
 
 ## @param ec2_prioritize_instance_id_as_hostname - boolean - optional - default: false
 ## @env DD_EC2_PRIORITIZE_INSTANCE_ID_AS_HOSTNAME - boolean - optional - default: false
-## Force the use of EC2 instance ID as hostname even when the hostname is not one of the
-## default EC2 hostnames. When set to true, this bypasses the standard hostname detection
-## logic and forces the EC2 instance ID to be used as the canonical hostname.
-## Only takes effect on EC2 instances.
+## On EC2, prefer the instance ID as the Agent hostname even when the OS hostname does not match
+## the known generic EC2 prefixes. Normal hostname detection still applies, and a hostname set in
+## datadog.yaml takes precedence. By default, the Agent uses the instance ID if the OS hostname
+## has a generic EC2 prefix to avoid merging multiple instances under one host. When this option
+## is true, the instance ID is used even if no prefix is detected. This is useful in cases such as
+## when custom images overwrite the system hostname, causing multiple instances to share the same hostname.
 #
 # ec2_prioritize_instance_id_as_hostname: false
 
--- a/pkg/config/config_template.yaml
+++ b/pkg/config/config_template.yaml
@@ -451,11 +451,9 @@ api_key:
 ## @param ec2_prioritize_instance_id_as_hostname - boolean - optional - default: false
 ## @env DD_EC2_PRIORITIZE_INSTANCE_ID_AS_HOSTNAME - boolean - optional - default: false
 ## On EC2, prefer the instance ID as the Agent hostname even when the OS hostname does not match
-## the known generic EC2 prefixes. Normal hostname detection still applies, and a hostname set in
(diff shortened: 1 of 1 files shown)


hestonhoffman (approved):

pkg/config/config_template.yaml
```diff
@@ -448,6 +448,15 @@ api_key:
 #
 # ec2_prefer_imdsv2: false
 
+## @param ec2_prioritize_instance_id_as_hostname - boolean - optional - default: false
+## @env DD_EC2_PRIORITIZE_INSTANCE_ID_AS_HOSTNAME - boolean - optional - default: false
+## Force the use of EC2 instance ID as hostname even when the hostname is not one of the
+## default EC2 hostnames. When set to true, this bypasses the standard hostname detection
+## logic and forces the EC2 instance ID to be used as the canonical hostname.
+## Only takes effect on EC2 instances.
```
hush-hush: This is not exactly true. We still go through the normal hostname detection. For example, a hostname configured in `datadog.yaml` will still take precedence.

The idea here is that EC2 instances, by default, use a generic hostname which is often the same across multiple instances. In such case we don't want to use the OS hostname as it would create merge conflict in the backend (merging multiple EC2 instances under a single host in the backend).

To solve this we check the system hostname has a specific prefix matching hostname from EC2 instances ( [here](https://github.com/DataDog/datadog-agent/blob/25b3df4250ccf281fd3fce3c1bfb28b6d8912898/pkg/util/hostname/common.go#L144) and [here](https://github.com/DataDog/datadog-agent/blob/25b3df4250ccf281fd3fce3c1bfb28b6d8912898/pkg/util/ec2/ec2.go#L22-L23) are the prefix we check for). If the prefix is found, we discard the OS hostname and try to fetch the instance ID.

`ec2_prioritize_instance_id_as_hostname` forces the use of the instance ID even if the system hostname doesn't have one of the generic prefix ([here](https://github.com/DataDog/datadog-agent/blob/25b3df4250ccf281fd3fce3c1bfb28b6d8912898/pkg/util/hostname/common.go#L144)). This is useful for customer that overwrite the system hostname in their image but still want the Agent to use the instance ID (classic example is customer using a custom image to spin VMs, which causes all their instance to have the same system hostname).

hush-hush (changes requested):

pkg/config/config_template.yaml
s-alad: thanks, just updated the description, let me know if it is better or if it should be more consice

pkg/config/config_template.yaml
```diff
@@ -448,6 +448,17 @@ api_key:
...
 # ec2_prefer_imdsv2: false
 
+## @param ec2_prioritize_instance_id_as_hostname - boolean - optional - default: false
+## @env DD_EC2_PRIORITIZE_INSTANCE_ID_AS_HOSTNAME - boolean - optional - default: false
+## On EC2, prefer the instance ID as the Agent hostname even when the OS hostname does not match
+## the known generic EC2 prefixes. Normal hostname detection still applies, and a hostname set in
+## datadog.yaml takes precedence. By default, the Agent uses the instance ID if the OS hostname
+## has a generic EC2 prefix to avoid merging multiple instances under one host. When this option
+## is true, the instance ID is used even if no prefix is detected. This is useful in cases such as
+## when custom images overwrite the system hostname, causing multiple instances to share the same hostname.
```
hush-hush: Small nit-pick to avoid repetition.

WDYT  of ?
```suggestion
## On EC2, prefer the instance ID as the Agent hostname even when the OS hostname does not match
## the known generic EC2 prefixes. By default, the Agent uses the instance ID if the OS hostname
## has a generic EC2 prefix to avoid merging multiple instances under one host. This is useful 
## when using custom images that share the same system hostname.
```

hush-hush (approved):

PR #40700, opened 2025-09-08
APM: add new efficient trace payload proto definitions

ajgajg1134:
### What does this PR do?
Add the initial proto definitions (currently unused) for the new indexed trace payload format aka "Efficient Trace Payload". 

### Motivation
These proto definitions are imported by a few external repos so getting them in here before the trace-agent itself uses them is useful for implementing tests
Code is currently unused, existing tests are sufficient here


felix-ganz (approved):
I've checked manually with our definition, looks good to me :)

pkg/proto/datadog/trace/idx/tracer_payload.proto
```diff
@@ -0,0 +1,51 @@
...
+	// samplingMechanism specifies the optional sampling mechanism (previously part of span tag _dd.p.dm)
+	uint32 samplingMechanism = 7;
+}
+
+// TracerPayload represents a payload the trace agent receives from tracers.
+message TracerPayload {
+	// strings specifies the array of strings referenced in this tracer payload, its chunks and spans.
+	repeated string strings = 1;
+	// containerIDRef specifies the string table ref of the ID of the container where the tracer is running on.
+	uint32 containerIDRef = 2;
```
ichinaski, line 32: Would it be worth adding the `containerTags` here as well, or do you intend to place them somewhere else? I like the idea of having a dedicated field for them.

ichinaski (approved):
Only remark is the potential addition of container tags, but leaving the PR approved.

pkg/proto/datadog/trace/idx/tracer_payload.proto
ajgajg1134: We certainly could, but I think it's also "ok" to put them in attributes here as I can imagine some use cases where container tags might not make sense or we want to refer to them with some other name for certain envs, e.g. serverless / fargate etc where they might have some other tracer_payload level metadata that is similar but different

ofek (approved):

PR #40701, opened 2025-09-08
add linter to disallow uses of RevertFinishedBackToBuilder

s-alad:
### What does this PR do?
linter to disallow uses of pkgconfigmodel.RevertFinishedBackToBuilder

### Motivation
OTel uses the method RevertFinishedBackToBuilder, because they build on top of the already finished configuration however this method should not be used by other calls. 

Using `RevertFinishedBackToBuilder` with and without `//nolint:forbidigo` when running the linter

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -627,7 +627,7 @@ linters-settings:
         pkg: ^github.com/DataDog/datadog-agent/pkg/config/setup$
         msg: use pkg/config/mock instead for tests or the config component
       - p: RevertFinishedBackToBuilder
-        msg: RevertFinishedBackToBuilder is deprecated and should not be used in any new code. Only existing OTel usage is allowed.
+        msg: RevertFinishedBackToBuilder is typically not needed, so using it is discouraged. Only existing OTel usage is allowed.
     analyze-types: true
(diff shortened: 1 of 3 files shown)


.golangci.yml
```diff
@@ -626,6 +626,8 @@ linters-settings:
       - p: ^setup.SetSystemProbe.*$
         pkg: ^github.com/DataDog/datadog-agent/pkg/config/setup$
         msg: use pkg/config/mock instead for tests or the config component
+      - p: RevertFinishedBackToBuilder
+        msg: RevertFinishedBackToBuilder is deprecated and should not be used in any new code. Only existing OTel usage is allowed.
```
dustmop: Technically this isn't deprecated, it's just discouraged. How about this:

```suggestion
        msg: RevertFinishedBackToBuilder is typically not needed, so using it is discouraged. Only existing OTel usage is allowed.
```

dustmop (approved):
LGTM! I have a small suggestion on how to change the language of the error, but aside from that this is good to merge.

ofek (approved):

liustanley (approved):

PR #40744, opened 2025-09-08
Add generic copies of OTLP attribute translation methods to mapping-go

IbraheemA:
### What does this PR do?

Add generic copies of OTLP attribute translation methods to mapping-go that can be used with all signal types via pcommon.Map (instead of just traces). These are copied from transform.go and otel_util.go, whose versions are now marked as deprecated. In a future PR, will replace these in the agent codebase.

### DIFF
--- a/pkg/opentelemetry-mapping-go/otlp/attributes/attributes.go
+++ b/pkg/opentelemetry-mapping-go/otlp/attributes/attributes.go
@@ -17,11 +17,27 @@ package attributes
 
 import (
 	"fmt"
+	"math"
+	"regexp"
+	"strconv"
 	"strings"
+	"time"
 
+	"github.com/DataDog/sketches-go/ddsketch"
+	"github.com/DataDog/sketches-go/ddsketch/mapping"
+	"github.com/DataDog/sketches-go/ddsketch/store"
+	"github.com/gogo/protobuf/proto"
 	"go.opentelemetry.io/collector/pdata/pcommon"
-	semconv127 "go.opentelemetry.io/otel/semconv/v1.27.0"
-	conventions "go.opentelemetry.io/otel/semconv/v1.6.1"
+	"go.opentelemetry.io/collector/pdata/pmetric"
+	"go.opentelemetry.io/collector/pdata/ptrace"
+	semconv1_12 "go.opentelemetry.io/otel/semconv/v1.12.0"
+	semconv1_17 "go.opentelemetry.io/otel/semconv/v1.17.0"
+	semconv1_27 "go.opentelemetry.io/otel/semconv/v1.27.0"
+	semconv1_6_1 "go.opentelemetry.io/otel/semconv/v1.6.1"
+
+	"github.com/DataDog/datadog-agent/pkg/opentelemetry-mapping-go/otlp/attributes/source"
+	"github.com/DataDog/datadog-agent/pkg/trace/log"
+	normalizeutil "github.com/DataDog/datadog-agent/pkg/util/normalize"
 )
 
 // customContainerTagPrefix defines the prefix for custom container tags.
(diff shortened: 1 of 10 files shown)


pkg/opentelemetry-mapping-go/otlp/attributes/attributes.test
dineshg13, line 1: remove the file ?

pkg/util/normalize/truncate.go
```diff
@@ -0,0 +1,37 @@
...
+		}
+		if len(s) >= 3 && utf8.Valid([]byte(s[len(s)-3:])) {
+			break
+		}
+		if len(s) >= 2 && utf8.Valid([]byte(s[len(s)-2:])) {
+			break
+		}
+		if len(s) >= 1 && utf8.Valid([]byte(s[len(s)-1:])) {
+			break
+		}
```
dineshg13, line 33: Can we rewrite this without repetitions

hestonhoffman (approved):

pkg/util/normalize/fuzz_test.go
```diff
@@ -0,0 +1,82 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+//go:build go1.18
```
pgimalac, line 6: Are we supposed to support Go < 1.18 ?

pkg/util/normalize/truncate.go
pgimalac, line 33: You could rewrite that with a for loop, breaking the outer loop

pkg/util/normalize/fuzz_test.go
pgimalac, line 1: Could you update codeowners for this package ? By default it would be owned by Agent Runtimes

pkg/util/normalize/normalize.go
pgimalac, line 1: If I'm not mistaken the whole file is copied from pkg/trace/traceutil/normalize/normalize.go, what's the point ? Why not just directly use that package ?

pkg/util/normalize/truncate.go
pgimalac, line 1: Same here, this whole file is pkg/trace/traceutil/normalize/truncate.go

pkg/opentelemetry-mapping-go/otlp/attributes/attributes.go
```diff
@@ -303,3 +339,638 @@ func ContainerTagFromAttributes(attr map[string]string) map[string]string {
...
+	if useDatadogNamespaceIfPresent {
+		cid := GetOTelAttrFromEitherMap(signalattrs, resourceattrs, true, KeyDatadogContainerID)
+		if cid != "" {
+			return cid
+		}
+	}
+	if ignoreMissingDatadogFields {
+		return ""
+	}
+	return GetOTelAttrFromEitherMap(signalattrs, resourceattrs, true, string(semconv1_27.ContainerIDKey), string(semconv1_27.K8SPodUIDKey))
```
jade-guiton-dd, line 861: Since we're refactoring things, we may want to consider removing the non-sensical `K8SPodUIDKey` fallback here

pkg/trace/traceutil/otel_util.go
```diff
@@ -305,6 +309,7 @@ func GetOTelService(span ptrace.Span, res pcommon.Resource, normalize bool) stri
 }
 
 // GetOTelResourceV1 returns the DD resource name based on OTel span and resource attributes.
+// Deprecated: Use GetOTelResource from pkg/opentelemetry-mapping-go/otlp/attributes instead.
```
jade-guiton-dd, line 312: The old behavior is deprecated _for users_, but I don't think we should recommend replacing calls to `GetOTelResourceV1` by the `GetOTelResource`

pkg/opentelemetry-mapping-go/otlp/attributes/attributes.go
```diff
@@ -303,3 +339,638 @@ func ContainerTagFromAttributes(attr map[string]string) map[string]string {
...
+	newSketch, err := ddsketch.LogCollapsingLowestDenseDDSketch(relativeAccuracy, maxNumBins)
+	if err != nil {
+		return nil, err
+	}
+
+	bucketCounts := dp.BucketCounts()
+	explicitBounds := dp.ExplicitBounds()
+
+	// Get scaling factor to convert unit to nanoseconds
+	scaleToNanos := getTimeUnitScaleToNanos(unit)
```
jade-guiton-dd, line 391: On top of maybe belonging to otlp/metrics, it seems these functions are specific to duration histograms; maybe the name should be more specific.

pkg/opentelemetry-mapping-go/otlp/attributes/attributes.go
```diff
@@ -303,3 +339,638 @@ func ContainerTagFromAttributes(attr map[string]string) map[string]string {
...
+		lowerBound = explicitBounds.At(idx - 1)
+	}
+	if idx < explicitBounds.Len() {
+		upperBound = explicitBounds.At(idx)
+	}
+	return
+}
+
+// createDDSketchFromHistogram creates a DDSketch from regular histogram data point
+func createDDSketchFromHistogram(dp pmetric.HistogramDataPoint, unit string) ([]byte, error) {
```
mackjmr, line 379: Why are we adding metric conversions in `otlp/attributes` and not `otlp/metrics` ?

PR #40746, opened 2025-09-08
[AGENTRUN-674] worker utilization warning

nathan-b:
### What does this PR do?
Adds a warning to the logs when worker utilization passes a specific threshold. Also display utilization data as part of `agent status`.

### Motivation
We have the data; might as well use the data

Unit testing
Manual testing

### Additional Notes


pkg/collector/runner/runner.go
```diff
@@ -285,3 +290,30 @@ func (r *Runner) StopCheck(id checkid.ID) error {
...
+	if err != nil {
+		log.Warnf("Error getting worker utilization data: %v", err)
+		return
+	}
+
+	averageUtilization := fmt.Sprintf("%.3f", overview.AverageUtilization)
+	log.Debugf("Average worker utilization: %v", averageUtilization)
+
+	if len(overview.WorkersOverThreshold) > 0 {
+		log.Warnf("Workers over utilization threshold: %v", overview.WorkersOverThreshold)
```
pgimalac, line 307: If this runs every minute we might need a log limiter to avoid printing the warning every single time

pkg/collector/runner/expvars/workers.go
```diff
@@ -12,11 +12,12 @@ import (
 )
 
 const (
-	// Top-level key for this expvar
-	workersExpvarKey = "Workers"
+	// WorkersExpvarKey - Top-level key for this expvar
+	WorkersExpvarKey = "Workers"
+	// InstancesExpvarKey - Nested key for the instances expvar
+	InstancesExpvarKey = "Instances"
```
pgimalac: Instead of exporting the keys, what do you think of adding Getter functions in the package ?
That way we don't leak the implementation outside of the package (that it's using expvar, with a key FOO within a map named BAR, using whatever type, etc)

pkg/collector/worker/utilization.go
pgimalac, line 1: Could we have this logic directly called by `SetWorkerStats` in the expvar package instead of the `Runner`, so that we check only when the stats are updated (+ eg. only if the last update was at least 1min ago)
It would also avoid having to extract data that way

nathan-b: I considered this approach, but I didn't go in that direction because it felt too much like blending separate concerns. I really want the expvar package to be focused on gathering and storing the expvar data without also as a side-effect monitoring worker utilization. I didn't think the overhead of getting data from the expvars was so bad that it was worth it.

A possible compromise would be having the expvar package take callbacks, such as `OnWorkerStatsUpdate` or similar, which could be registered inside a separate package. What do you think of that approach?

pkg/collector/runner/expvars/workers.go
nathan-b: This is an interesting request. My assumption was that the contents of expvars (which are, after all, global state) represent a kind of contract in itself. Providing a getter would imply that the structure of the expvar data is an internal implementation detail, when really that structure and those data are external state.

I'm open to being convinced otherwise :)

pkg/collector/runner/runner.go
nathan-b: Is there any sort of structured log limiting at Datadog, or is it ad-hoc? I could add "last log" and "cooldown duration" fields to the utilization class to implement limiting, or if there's already a solution within the log package I'm happy to do that instead 👍

nathan-b: I added a hand-crafted log limiter; if it turns out there's structured log limiting then we can revert that particular commit and use the global solution instead

pkg/collector/worker/utilization.go
nathan-b: I looked into this a little bit more. In addition to the objection I raised before (blending separate concerns), each worker would have to have its own `UtilizationMonitor` which separately tracks the logging cooldown, utilization threshold, etc. This would also mean that each worker that's over the threshold would log its own utilization warning, which would be scattered all over the log, rather than having a single log line that lists all the over-threshold workers.

I like your thinking here, but ultimately I don't believe the approach you suggest is the best one.

pkg/collector/runner/runner.go
pgimalac: There is `NewLogLimit` in `pkg/util/log/log_limit.go`, which should be good enough here:
```
// NewLogLimit creates a Limit where shouldLog will return
// true the first N times it is called, and will return true once every
// interval thereafter.
```

pkg/collector/runner/expvars/workers.go
pgimalac: That's a fair point !
Although the expvars are exposed to the outside, I don't consider them part of an API / "contract", we expose whatever variable we need to, but I don't think we'd be worried about removing one for example
So in that regard I would argue it's best to keep the details like actual underlying data types and path to access one variable internal to that package

pkg/collector/worker/utilization.go
pgimalac: Fair points, I'm fine with keeping it as is 👍

pkg/collector/runner/runner.go
nathan-b: Thanks for letting me know about this! I'll refactor the existing code to use the existing Limit code instead (maybe this would be a good usage of an AI agent...)

pkg/collector/runner/expvars/workers.go
nathan-b: I'm still on the fence about this, but I'm deferring to your greater experience with the agent. You can see commit 24af where I refactored it into using getters.

pkg/collector/runner/runner.go
nathan-b: The AI agent produced a mostly-useless test but otherwise did a good job.

nathan-b:
Just need to add the new stats to the status gui. Everything else should be complete.

Not sure if it would make sense to add a release note for this. Any thoughts?

pkg/config/config_template.yaml
```diff
@@ -774,6 +774,28 @@ api_key:
...
 
+## @param check_runner_utilization_threshold - float - optional - default: 0.95
+## @env DD_CHECK_RUNNER_UTILIZATION_THRESHOLD - float - optional - default: 0.95
+## The threshold above which workers are considered over-utilized. This is used by the
+## utilization monitor to identify workers that may need attention. The value should be
+## between 0.0 and 1.0, where 1.0 means 100% utilization.
+#
+# check_runner_utilization_threshold: 0.95
+
+## @param check_runner_utilization_monitor_interval - duration - optional - default: 60s
```
pgimalac: Not really sure whether we need `check_runner_utilization_monitor_interval` and `check_runner_utilization_warning_cooldown` in the template, we could consider the defaults to be sane enough that users won't need to change them

pkg/status/collector/status_templates/collector.tmpl
```diff
@@ -81,6 +81,27 @@
...
+    Total Workers: {{.Count}}
+    {{- if .AverageUtilization }}
+    Average Utilization: {{percent .AverageUtilization}}%
+    {{- end }}
+    {{- if .Instances }}
+
+    Individual Workers:
+      {{- range $workerName, $workerData := .Instances }}
+      {{$workerName}}: {{percent $workerData.Utilization}}%
+      {{- end }}
```
pgimalac: Small concern on having one line per worker when there might be dozens (or in some weird cases, hundreds)
Would be great if we found a way to show the information without taking that much visual space

pgimalac (approved):
Small nitpicks but overall LGTM !

iadjivon (approved):

pkg/status/collector/status_templates/collector.tmpl
nathan-b: I would guess that in the case where there are hundreds of workers, the average utilization starts to be what the user cares the most about. I think the top 25 plus the average would be sufficient, though I'm open to other ideas.

pkg/config/config_template.yaml
nathan-b: Fair, and probably accurate. I wasn't sure what the guidelines were around putting new config values in the template. I'll remove them, though also @hush-hush feel free to chime in if you have thoughts.

hush-hush: We want the template to include settings that customer will use. The template is here mostly to advertise settings. I agree with @pgimalac, we might not want those in the template.

hush-hush (changes requested):

pkg/status/collector/status_templates/collector.tmpl
nathan-b: Done!

pkg/config/config_template.yaml
nathan-b: Removed the new configs from the template. Thanks for the info!

pkg/status/collector/status.go
```diff
@@ -27,10 +28,82 @@ func GetStatusInfo() map[string]interface{} {
...
+				if workerCount > 0 {
+					avgUtilization := totalUtilization / float64(workerCount)
+					workerStats["AverageUtilization"] = avgUtilization
+
+					// Sort workers by utilization in descending order
+					sort.Slice(workers, func(i, j int) bool {
+						return workers[i].Utilization > workers[j].Utilization
+					})
+
+					// Keep only top 25 workers
```
pgimalac, line 80: **🥜 nitpick:** ‏I'm thinking we could also exclude workers with a zero utilisation, wdyt ?

pgimalac (approved):
LGTM !

hush-hush (approved):

pkg/status/collector/status.go
nathan-b: I'm wondering if customers might be concerned to not see workers they expected to see, though that might not be a big concern. Since the usual case would be a small number of workers, I think I will keep it as-is for right now. It's a reasonable idea, though, so if you feel strongly about it then I'm open to being convinced :)

releasenotes/notes/add-check-worker-utilization-stats-fd60866dd32527d8.yaml
```diff
@@ -0,0 +1,13 @@
...
+# it does not depend on any information only available in another
+# section. This may mean repeating some details, but each section
+# must be readable independently of the other.
+#
+# Each section note must be formatted as reStructuredText.
+---
+enhancements:
+  - |
+    Added tracking of check worker utilization.
+    Utilization stats can be seen using the ``agent status`` command.
```
iadjivon:
```suggestion
    Utilization stats can be seen using the `agent status` command.
```

very small nitpick

nathan-b: Ah, thank you...I read the documentation for reStructuredText but couldn't figure out how to render the RNs to validate that my formatting was correct 😓

iadjivon (approved):

pkg/status/collector/status.go
pgimalac: I definitely don't feel strongly about it
I was just thinking it's weird to show the "top" workers with some being 0 in the list

PR #40762, opened 2025-09-09
Update `ImageResolver` cache structure to use repo name as key

erikayasuda:
### What does this PR do?
Restructures the `ImageResolver` cache structure so that the keys are only the repository names (ex - `dd-lib-python-init`) and not the repository URL (ex - `gcr.io/dd-lib-python-init`). 

It also enables us to configure the default Datadog container registries via configuration. This is not public-facing, and will be used primarily for dogfooding the new feature on staging (given staging uses different container registries).

### Motivation
The repository URL field in the `K8S_INJECTION_DD` remote config data was not designed to consumed by DCA, but primarily for use by our internal `equilibrium` tooling. This meant that by using the repository URL for the key, it would be limited to only allowing gradual rollout for customers using `grc.io/datadoghq` as their registry. See [here](https://docs.datadoghq.com/tracing/trace_collection/automatic_instrumentation/single-step-apm/kubernetes/?tab=agentv764recommended#change-the-default-image-registry) for other valid Datadog registries.

- Updated existing tests
- Added new unit tests
- Local E2E testing with `injector-dev`


pkg/clusteragent/admission/mutate/autoinstrumentation/image_resolver.go
```diff
@@ -116,38 +122,46 @@ func (r *remoteConfigImageResolver) waitForInitialConfig() error {
...
 // Input: "gcr.io/datadoghq", "dd-lib-python-init", "v3"
-// Output: "gcr.io/datadoghq/dd-lib-python-init@sha256:abc123...", true
+// Output: "dd-lib-python-init@sha256:abc123...", true
 // If resolution fails or is not available, it returns nil.
 // Output: nil, false
 func (r *remoteConfigImageResolver) Resolve(registry string, repository string, tag string) (*ResolvedImage, bool) {
 	r.mu.RLock()
 	defer r.mu.RUnlock()
 
+	if !isDatadoghqRegistry(registry) {
```
TonyCTHsu: Thoughts: It is possible to push the logic to the caller and remove the `registry` argument for this function. Because the caller already have the value of registry, it can make the decision whether to invoke the `Resolve` method. 

If we decide to keep the dependency for this function, this input check should happen before the mutex lock.

cswatt (approved):

pkg/clusteragent/admission/mutate/autoinstrumentation/image_resolver.go
erikayasuda: I think we can keep it here for now (but will move to before the mutex lock), since the registry string is already being passed in here anyways. I agree that it could probably happen earlier since the registry should stay constant, unless the deployment gets updated. I'm down to make that change for 7.72 unless we foresee issues with this.

pkg/clusteragent/admission/mutate/autoinstrumentation/image_resolver.go
```diff
@@ -144,14 +153,19 @@ func (r *remoteConfigImageResolver) Resolve(registry string, repository string,
...
+	resolvedImage := newResolvedImage(registry, repository, resolved)
+	log.Debugf("Resolved %s/%s:%s -> %s", registry, repository, tag, resolvedImage.FullImageRef)
 	metrics.ImageResolutionAttempts.Inc(registry, repository, metrics.DigestResolutionEnabled, resolved.Digest)
-	return &resolved, true
+	return resolvedImage, true
+}
+
+func isDatadoghqRegistry(registry string) bool {
+	_, exists := datadoghqRegistries[registry]
+	return exists
```
erikayasuda, line 165: @betterengineering / @stanistan - Is it safe to just check for exact match?

stanistan: Yeah I think so

pkg/clusteragent/admission/mutate/autoinstrumentation/image_resolver.go
```diff
@@ -20,6 +20,12 @@ import (
 	"github.com/DataDog/datadog-agent/pkg/util/log"
 )
 
+var datadoghqRegistries = map[string]struct{}{
+	"gcr.io/datadoghq":       {},
+	"docker.io/datadog":      {},
+	"public.ecr.aws/datadog": {},
+}
+
```
adel121: Maybe it would be more useful to store these in the agent config instead of hardcoding them here so that we can tweak them in the future for testing / dogfooding. 

We don't have to document the config option for public users, but can still be useful for us to control via env vars.

adel121 (commented):
LGTM

Left a minor suggestion.

TonyCTHsu (approved):

pkg/clusteragent/admission/mutate/autoinstrumentation/image_resolver.go
erikayasuda: The config option is actually documented for public users [here](https://docs.datadoghq.com/tracing/trace_collection/automatic_instrumentation/single-step-apm/kubernetes/?tab=agentv764recommended#change-the-default-image-registry). Or are you suggesting we store the source of truth (the registries that we support for gradual rollout) as a config?

adel121: The one documented publicly allows users to decide which registry will be used in the injected init containers. 

If the one specified via the public config is not in the set STATICALLY defined here, no resolution will take place. 

This makes testing harder as we will be forced to use one of the hardcoded registries to trigger the resolution path in the code.

erikayasuda: Gotcha, yeah I think it is a bit annoying for testing. Would it be as simple as adding it the same way you linked:
```go
config.BindEnv("admission_controller.auto_instrumentation.container_registry")
```
But something like:
```go
config.BindEnv("admission_controller.auto_instrumentation.gradual_rollout_container_registries")
```
We won't document it publicly, and if empty, I can default to the 3 I've listed above? If it's a quick fix like this then I'm down to include it in this PR (since it would be really nice to have for dogfooding), but if it's more involved, I think I prefer to hold off and do it as a follow up.

adel121: I believe this should be a one-liner in config.go, and another line in auto instrumentation webhook code.

stanistan: Having this in a config would make it much easier for us to test gradual rollout in internal registries as well.

erikayasuda: Addressed in: https://github.com/DataDog/datadog-agent/pull/40855

rahulkaukuntla (approved):

adel121 (approved):

PR #40764, opened 2025-09-09
[NCM] Add device profiles for corecheck

zoedt:
### What does this PR do?

https://datadoghq.atlassian.net/browse/NDMII-3608

This PR adds initial logic for supporting profiles that define CLI commands for common actions we need to take to retrieve a configuration from a networking device.

You can refer to this [sub-RFC](https://datadoghq.atlassian.net/wiki/spaces/II/pages/5314904489/Sub-RFC+Ingestion+Generic+retrieval+of+configs) for more explicit details regarding device profiles for NCM.

This PR adds:
* new `profile` package, generalized for other YAML parsing needs (can be extracted outside of `pkg/networkconfigmanagement`
  * testing
    * adds yaml fixtures for testing
    * testing for profile logic additions
* update the existing core check to utilize the profiles + rudimentary matching
* allow for profiles to be a field to access within the remote client, to pull the commands necessary

**regarding exclusion of a `user_profiles` directory**
a directory for user-made profiles is not currently being supported to begin with - we hope to slowly provide

another todo is to support some "overrides" from the `DeviceInstance` if there are any use cases for custom command changes, but without as much maintenance as user profiles may be (as seen by SNMP profiles).

slack thread for context [here](https://dd.slack.com/archives/C08NTAS3A1E/p1756919426988199), screenshot for posterity below:
<details>
<summary>🖼️ screenshot mentioned above</summary>
image
</details>


pkg/collector/corechecks/networkconfigmanagement/networkconfigmanagement.go
```diff
@@ -153,3 +168,30 @@ func newCheck(agentConfig config.Component) check.Check {
 		agentConfig: agentConfig,
 	}
 }
+
+// FindMatchingProfile supports testing profiles until one is found with a successful command for the device
+func (c *Check) FindMatchingProfile() (*profile.NCMProfile, error) {
+	for profName, prof := range c.checkContext.ProfileMap {
+		session, err := c.remoteClient.NewSession()
```
NouhaManai96: why do we need to create a new session for every profile ?

pkg/collector/corechecks/networkconfigmanagement/networkconfigmanagement.go
```diff
@@ -153,3 +168,30 @@ func newCheck(agentConfig config.Component) check.Check {
...
+		if err != nil {
+			return nil, fmt.Errorf("unable to connect to remote device %s: %s", c.checkContext.Device.IPAddress, err)
+		}
+		commands, err := prof.GetCommandValues(profile.Version)
+		if err != nil {
+			return nil, err
```
NouhaManai96: wouldn't this mean that if one profile has an error (syntax error for example) then it would fail the matching for all the rest of profiles since we exit here instead of moving on to the next profile in the list ?
(same for the next `return nil, err` )

pkg/collector/corechecks/networkconfigmanagement/networkconfigmanagement.go
```diff
@@ -153,3 +168,30 @@ func newCheck(agentConfig config.Component) check.Check {
 		agentConfig: agentConfig,
 	}
 }
+
+// FindMatchingProfile supports testing profiles until one is found with a successful command for the device
+func (c *Check) FindMatchingProfile() (*profile.NCMProfile, error) {
```
NouhaManai96, line 173: wondering if the matching logic should be per type or all or nothing ?
meaning if a profile has the correct command to get the running config but not the right one to get the start up config then should it be considered valid or not ?
with the current implementation it's considered invalid but it means we are potentially losing one (or more) valid command(s) because of a "wrong" one

pkg/networkconfigmanagement/test/conf.d/networkconfigmanagement.d/default_profiles/_base.yaml
NouhaManai96, line 1: what is this used for exactly ?

NouhaManai96 (commented):
looks good overall !
I left some questions about the profile validation and matching logic !

pkg/networkconfigmanagement/test/conf.d/networkconfigmanagement.d/default_profiles/_base.yaml
zoedt: lol i copied for an example from the snmp profiles but i didn't clean it up LOL 

i'll amend so it has the `base` elements that is the basis (really just the `name` field rn)

pkg/collector/corechecks/networkconfigmanagement/networkconfigmanagement.go
zoedt: unfortunate need from the library - each time we want to send a "command" a new session would have to be made - more context from a question florian also had regarding sessions [in this comment thread](https://github.com/DataDog/datadog-agent/pull/39735#discussion_r2293037228)

for this method of seeing what commands "succeeds" we would have to create sessions to test the commands - as we have different methods moving forward we'll probably have something more refined

pkg/collector/corechecks/networkconfigmanagement/networkconfigmanagement.go
zoedt: good call out! i didn't add a `FindMatchingProfile` test - i will add one featuring the error case and make sure it still proceeds to the next profile

pkg/collector/corechecks/networkconfigmanagement/networkconfigmanagement.go
zoedt: good point in whether one command succeeds or not what is considered "valid" (i think we can send the running config and log some warning or metric that shows that the startup command failed but not fail the entire check)

in this case for finding the profile match - we're running the `show version` command on line 179 which isn't very clear so i may add something to make that more obvious (or whatever is closest to sharing the device's OS, version, model, etc.)

in a follow up PR i'll focus on the "parsing"/"validation" of this command - if we can parse out information regarding these above then we can intuit the profile match a lot easier 🤔

NouhaManai96: maybe the `SetProfile` can be updated to set a type of command instead of the whole profile ?

zoedt: i added a commit to amend the finding profile to match by valid response from retrieving the running config - we have to do that anyways and we care more about it as you noted!

next after this pr i'm going to look into redactions + validation rules for the profile so we are certain the profile is a match :^D

pkg/collector/corechecks/networkconfigmanagement/networkconfigmanagement.go
```diff
@@ -153,3 +169,24 @@ func newCheck(agentConfig config.Component) check.Check {
...
+	for profName, prof := range c.checkContext.ProfileMap {
+		if slices.Contains(c.checkContext.ProfileCache.GetTriedProfiles(), profName) {
```
Pierre-L42: nit: maybe we could create a function like `IsProfileTried` that does the contains check rather than having `GetTriedProfiles`?

pkg/networkconfigmanagement/profile/profile.go
```diff
@@ -0,0 +1,190 @@
...
+	files, err := os.ReadDir(profilesRoot)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read profile dir %q: %w", profilesRoot, err)
+	}
+
+	profiles := make(Map)
+	for _, file := range files {
+		filename := file.Name()
+		// Skip non yaml profiles
+		if !strings.HasSuffix(filename, ".yaml") {
```
Pierre-L42: From what I understood from the `ParseProfileFromFile`, we also want to support `.json` files, should we also not skip `.json` files here?

pkg/networkconfigmanagement/config/config.go
```diff
@@ -105,6 +107,24 @@ func NewNcmCheckContext(rawInstance integration.Data, rawInitConfig integration.
 	ncc.Namespace = initConfig.Namespace
 	ncc.MinCollectionInterval = time.Duration(initConfig.MinCollectionInterval)
 
+	// Populate the profiles map (from defaults/OOTB)
+	profMap, err := profile.GetProfileMap("default_profiles")
```
Pierre-L42, line 111: We only want to support OOTB profiles and not custom profiles for this PR?

pkg/collector/corechecks/networkconfigmanagement/networkconfigmanagement.go
```diff
@@ -50,12 +52,26 @@ func (c *Check) Run() error {
...
+	// Must defer this way because sometimes we will have to redial the remote client
 	defer func() {
 		if c.remoteClient != nil {
 			_ = c.remoteClient.Close()
 		}
 	}()
 
+	// If the check did not have inline profile explicitly defined/from cache, find the profile that works
+	if c.checkContext.ProfileCache.ProfileName == "" {
+		prof, err := c.FindMatchingProfile()
```
Pierre-L42, line 64: Is there a specific reason we do this in `Run` rather than `Configure` of the check?

pkg/collector/corechecks/networkconfigmanagement/networkconfigmanagement.go
Pierre-L42: Just throwing an idea: maybe we could also try all profiles (or until one have all commands that succeeds) and use the profile where the most commands succeeded? this could make the `FindMatchingProfile` function slower though if there is a lot of profiles

pkg/networkconfigmanagement/config/config.go
zoedt: yes! see context in the PR description / pasted again here!

> regarding exclusion of a user_profiles directory
a directory for user-made profiles is not currently being supported to begin with - we hope to slowly provide

>another todo is to support some "overrides" from the DeviceInstance if there are any use cases for custom command changes, but without as much maintenance as user profiles may be (as seen by SNMP profiles).

> slack thread for context [here](https://dd.slack.com/archives/C08NTAS3A1E/p1756919426988199), screenshot for posterity below:
<details>
<summary>🖼️ screenshot mentioned above</summary>
image
</details>

pkg/networkconfigmanagement/profile/profile.go
zoedt: good catch, i forgot to remove this when i refactored to incl. json - will update this!

pkg/networkconfigmanagement/config/config.go
Pierre-L42: Sorry missed that, thanks for reminding it!

zoedt: np!! ty for watching out for it, it's a good question!

pkg/collector/corechecks/networkconfigmanagement/networkconfigmanagement.go
zoedt: hmm i think it was mostly that for the "matching" functionality - we'd need to connect to the device and i wanted to make the `Configure` mostly straightforward and not involve any device connection at that point - there may be additional functionality for the "matching" in the future where there could be additional involvement such as querying for the sysobjectid which i also think doesn't necessarily need to live in the `Configure` function but lmk your thoughts!

i tried to avoid any additional extra work from the `Run` functionality by skipping over profiles if already tried and err out as soon as possible

wdyt? i'm not feeling too strongly if it feels better to move it and connect to the client in the `Configure` step but it just seemed like a lot - i can encapsulate it in a separate function if that helps as well if the issue is the longness of the `Run` function?

Pierre-L42: Thanks for the explanation, I'm fine with as it is now then!

Pierre-L42 (approved):

PR #40765, opened 2025-09-09
[dyninst/object] Simplify computing compression lengths

piob-io:
This fixes a bug that occurs when compressed length is longer than compressed one.

Add a cli for running irgen.


ajwerner (approved):
Thank you!

PR #40766, opened 2025-09-09
[CONTP-841]: Fixed a bug for handling hostpath on Admission Controller

JatinSingh007:
### What does this PR do?

Fixes an issue in the Admission Controller where the UDS socket host path was hardcoded to **/var/run/datadog**. This caused the controller to ignore the hostSocketPath configured via the Helm chart or Operator. As a result, mutated pods could not access APM or DogStatsD sockets when custom paths were used.

This PR updates:
	•   The Agent config to expose new config keys for **trace_agent_host_socket_path** and  **dogstatsd_agent_host_socket_path**
	•.  The mutator code to consume those config values instead of relying on hardcoded defaults
	•.  Unit tests to validate both default behavior and user-defined socket paths

### Motivation
When users deploy the Agent with a custom socket path, e.g:

```
datadog:
  dogstatsd:
    hostSocketPath: /var/run/datadog-custom
  apm:
    hostSocketPath: /var/run/datadog-custom
clusterAgent:
  admissionController:
    mutateUnlabelled: true
    enabled: true
```

the sockets are created correctly under /var/run/datadog-custom, but the Admission Controller still injects /var/run/datadog.

Impact:
	•   Mutated pods fail to connect to the expected sockets
	•.  Metrics and traces are lost when workloads rely on auto-injected UDS mounts

Root Cause:
	•  The admission mutation webhook injects hardcoded paths instead of reading the configured values.

**Deploying the agents**
      • Clone the repo datadog-agent
      • Build the images for your agent and cluster-agent
      • Set up your own simple values.yaml 
```
datadog:
  apiKey: '' <provide your api key here>
  clusterName: '' <provide your clustername here>
  kubelet:
    tlsVerify: false
  logs:
    containerCollectAll: true
    enabled: true
  dogstatsd:
    hostSocketPath: /var/run/datadog-custom
  apm:
    hostSocketPath: /var/run/datadog-custom
clusterAgent:

  enabled: true
  shareProcessNamespace: false
  admissionController:
    mutateUnlabelled: true
    enabled: true
  image:

    repository: '' <provide your cluster agent repo here>
    tag: '' <provide your cluster agent tag here>
    doNotCheckTag: true    
```
 • Deploy the agents with `helm install latest-release -f values.yaml`

**Setup and deploy a test redis app**
 • Deploy your own redis app into the cluster with the following template:

```
apiVersion: v1
kind: Pod
metadata:
  name: redis
spec:
  containers:
  - name: redis
    image: redis
    ports:
    - containerPort: 80
```
 • Run the app with `kubectl apply -f template.yaml` and check the configuration with `kubectl get pods redis -o yaml > redis_output.txt 
`

**Verify the Volume injection**

Check the new deployed redis app pods configs and check volumes and volume mounts. Ideally the host path should be same as we defined in the values.yaml.

```
 - hostPath:
      path: /var/run/datadog-custom <- Same as hostpath described above
      type: DirectoryOrCreate
      name: datadog
```

```
volumeMounts:
    - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
      name: kube-api-access-d9wdj
      readOnly: true
    - mountPath: /var/run/datadog
      name: datadog
      readOnly: true
```

We can see the type of volume injection is DirectoryOrCreate, if it is socket we will see something like `<directory>/apm.socket` or `<directory>/dsd.socket`  at hostPath.


pkg/clusteragent/admission/mutate/config/config_test.go
```diff
@@ -431,13 +436,15 @@ func TestInjectSocket_VolumeTypeSocket(t *testing.T) {
 		{
 			name:                    "no csi driver",
 			withCSIDriver:           false,
+			hostSocketPathDsd:       "/var/run/datadog-custom/dsd.socket",
+			hostSocketPathApm:       "/var/run/datadog-custom/apm.socket",
```
adel121: Could you please add a test case where `hostSocketPathDsd` and `hostSocketPathApm` are not equal ?

adel121 (commented):
Great work Jatin!

I left a small comment.

dustmop (approved):

pkg/config/setup/config.go
```diff
@@ -842,6 +842,8 @@ func InitConfig(config pkgconfigmodel.Setup) {
 	config.BindEnvAndSetDefault("admission_controller.inject_config.endpoint", "/injectconfig")
 	config.BindEnvAndSetDefault("admission_controller.inject_config.mode", "hostip") // possible values: hostip / service / socket / csi
 	config.BindEnvAndSetDefault("admission_controller.inject_config.local_service_name", "datadog")
+	config.BindEnvAndSetDefault("admission_controller.inject_config.trace_agent_host_socket_path", "/var/run/datadog") // configure base host socket path to avoid hardcoding
```
adel121:
```suggestion
	config.BindEnvAndSetDefault("admission_controller.inject_config.trace_agent_host_socket_path", "/var/run/datadog")
```

pkg/clusteragent/admission/mutate/config/mutator.go
```diff
@@ -193,24 +212,103 @@ func (i *Mutator) injectSocketVolumes(pod *corev1.Pod, withCSI bool) bool {
...
+			volume      corev1.Volume
+			volumeMount corev1.VolumeMount
+		}
+		var plans []plannedMount
+
 		if withCSI {
-			// csiDSDSocketDirectory is used because we can't mount two different volumes at the same mount path
-			// this is a limitation in the configuration of the admission controller
-			volume, volumeMount = buildCSIVolume(DatadogVolumeName, i.config.socketPath, csiDSDSocketDirectory, true, i.config.csiDriver)
+			if csiDSDSocketDirectory == csiAPMSocketDirectory {
```
adel121: I think here we are not comparing the right variables. 

`csiDSDSocketDirectory` and `csiAPMSocketDirectory` are constant strings, they will never be equal.

Maybe you meant to compare the host path of the trace agent and dogstatsd sockets ?

adel121: Since this was not caught by unit tests, it means that either we are missing unit tests for this if branch, or we have a bug in the unit tests.

pkg/clusteragent/admission/mutate/config/mutator.go
```diff
@@ -287,12 +380,12 @@ func injectExternalDataEnvVar(pod *corev1.Pod) (injected bool) {
 	return
 }
 
-func buildHostPathVolume(volumeName, path string, hostpathType corev1.HostPathType, readOnly bool) (corev1.Volume, corev1.VolumeMount) {
+func buildHostPathVolume(volumeName, hostpath string, path string, hostpathType corev1.HostPathType, readOnly bool) (corev1.Volume, corev1.VolumeMount) {
```
adel121, line 383:
```suggestion
func buildHostPathVolume(volumeName, hostpath string, mountPath string, hostpathType corev1.HostPathType, readOnly bool) (corev1.Volume, corev1.VolumeMount) {
```

pkg/clusteragent/admission/mutate/config/mutator.go
```diff
@@ -26,27 +26,35 @@ import (
...
+		localServiceName:         datadogConfig.GetString("admission_controller.inject_config.local_service_name"),
+		traceAgentHostSocket:     datadogConfig.GetString("admission_controller.inject_config.trace_agent_host_socket_path"),
+		dogStatsDAgentHostSocket: datadogConfig.GetString("admission_controller.inject_config.dogstatsd_agent_host_socket_path"),
+		traceAgentSocket:         datadogConfig.GetString("admission_controller.inject_config.trace_agent_socket"),
+		dogStatsDSocket:          datadogConfig.GetString("admission_controller.inject_config.dogstatsd_socket"),
+		socketPath:               datadogConfig.GetString("admission_controller.inject_config.socket_path"),
+		typeSocketVolumes:        datadogConfig.GetBool("admission_controller.inject_config.type_socket_volumes"),
+		csiDriver:                datadogConfig.GetString("csi.driver"),
+		dsdSuffix:                "/dsd",
+		apmSuffix:                "/apm",
```
adel121: Since these are constants, there is no need to define them as struct properties.

pkg/clusteragent/admission/mutate/config/mutator.go
```diff
@@ -126,6 +132,25 @@ func (i *Mutator) MutatePod(pod *corev1.Pod, _ string, _ dynamic.Interface) (boo
...
+			apmMountBase := i.config.socketPath
+			dsdMountBase := i.config.socketPath
+
+			if useCSI || (i.config.dogStatsDAgentHostSocket != i.config.traceAgentHostSocket) {
+				apmMountBase = i.config.socketPath + i.config.apmSuffix
+				dsdMountBase = i.config.socketPath + i.config.dsdSuffix
+			}
+
+			traceURLSocketEnvVar.Value = "unix://" + apmMountBase + "/apm.socket"
+			dogstatsdURLSocketEnvVar.Value = "unix://" + dsdMountBase + "/dsd.socket"
```
adel121: This can be simplified to always split the socket paths. 

For instance, we can always put the sockets at: `/var/run/datadog/apm/apm.socket` and `/var/run/datadog/dsd/dsd.socket`. 

This will work in regardless of whether the sockets have the same host paths or not, and it makes the implementation much simpler. The code will essentially get reduced to simply:

```
apmMountBase := i.config.socketPath + i.config.apmSuffix
dsdMountBase := i.config.socketPath + i.config.dsdSuffix

traceURLSocketEnvVar.Value = "unix://" + apmMountBase + "/apm.socket"
dogstatsdURLSocketEnvVar.Value = "unix://" + dsdMountBase + "/dsd.socket"
```

When creating the volume, we have 3 cases:
- Mounting volumes with type Socket: in this case it is a direct mount from hostpath to mount path, resulting in 2 volumes (CSI case also falls into this category).
- Mounting volumes with type DirectoryOrCreate, where apm and dsd have **different** hostpaths. This is similar to the first case, we will add and mount 2 volumes, mounting the hostpath of each socket separately.
- Mounting volumes with type DirectoryOrCreate, where apm and dsd have **same** hostpaths. In this case it is sufficient to mount volume named `datadog`, mounting the common hostpath only once at the common mountpath.

pkg/clusteragent/admission/mutate/config/mutator.go
```diff
@@ -160,20 +185,23 @@ func (i *Mutator) injectSocketVolumes(pod *corev1.Pod, withCSI bool) bool {
...
+			csiVolumeType  csiInjectionType
+			hostsocketpath string
 		}{
 			DogstatsdSocketVolumeName: {
 				socketpath: strings.TrimPrefix(
 					i.config.dogStatsDSocket, "unix://",
 				),
-				csiVolumeType: csiDSDSocket,
+				csiVolumeType:  csiDSDSocket,
+				hostsocketpath: i.config.dogStatsDAgentHostSocket + i.config.dsdSuffix + ".socket",
```
adel121: If I am not mistaken, this is assuming that the socket name on the host is fixed to `apm.socket` or `dsd.socket`, but this is configurable via helm chart. The user might set the following:

```
datadog:
   dogstatsd:
      socketPath: /var/run/datadog/apm.sock
```

With this, the agent will create a socket called `apm.sock`, but the admission controller will be assuming the socket is named `apm.socket` on the host.

Additionally, why are we adding the dsdSuffix to the hostpath? I thought the suffix should be only added to the mountpath.

pkg/clusteragent/admission/mutate/config/mutator.go
```diff
@@ -160,20 +185,23 @@ func (i *Mutator) injectSocketVolumes(pod *corev1.Pod, withCSI bool) bool {
...
 			},
 			TraceAgentSocketVolumeName: {
 				socketpath: strings.TrimPrefix(
 					i.config.traceAgentSocket, "unix://",
 				),
-				csiVolumeType: csiAPMSocket,
+				csiVolumeType:  csiAPMSocket,
+				hostsocketpath: i.config.traceAgentHostSocket + i.config.apmSuffix + ".socket",
```
adel121: same comment as above

pkg/clusteragent/admission/mutate/config/mutator.go
adel121: The correct name of the socket is accessible via `config.GetString("dogstatsd_socket")`.

JatinSingh007: `For instance, we can always put the sockets at: /var/run/datadog/apm/apm.socket and /var/run/datadog/dsd/dsd.socket.` I think this won't work when we have same hostpaths with the current code, we will have to mount on different dir in that case too.

pkg/config/setup/config.go
```diff
@@ -842,9 +842,11 @@ func InitConfig(config pkgconfigmodel.Setup) {
 	config.BindEnvAndSetDefault("admission_controller.inject_config.endpoint", "/injectconfig")
 	config.BindEnvAndSetDefault("admission_controller.inject_config.mode", "hostip") // possible values: hostip / service / socket / csi
 	config.BindEnvAndSetDefault("admission_controller.inject_config.local_service_name", "datadog")
+	config.BindEnvAndSetDefault("admission_controller.inject_config.trace_agent_host_socket_path", "/var/run/datadog")
+	config.BindEnvAndSetDefault("admission_controller.inject_config.dogstatsd_agent_host_socket_path", "/var/run/datadog")
 	config.BindEnvAndSetDefault("admission_controller.inject_config.socket_path", "/var/run/datadog")
-	config.BindEnvAndSetDefault("admission_controller.inject_config.trace_agent_socket", "unix:///var/run/datadog/apm.socket")
-	config.BindEnvAndSetDefault("admission_controller.inject_config.dogstatsd_socket", "unix:///var/run/datadog/dsd.socket")
+	config.BindEnvAndSetDefault("admission_controller.inject_config.trace_agent_socket_filename", "apm.socket")
+	config.BindEnvAndSetDefault("admission_controller.inject_config.dogstatsd_agent_socket_filename", "dsd.socket")
```
adel121: I think these values are already extractable from the existing configs. 

dogstatsd socket filename can be obtained from `config.GetString("dogstatsd_socket")`
apm socket filename can be obtained from `config.GetString("apm_config.receiver_socket")`

The former is already passed down to all containers, including the cluster agent, so the admission controller already has access to it.

The latter is only passed down to the node agent containers (see example [here](https://github.com/DataDog/helm-charts/blob/6827c62cffb014492974c34b285f0df967446c9e/charts/datadog/templates/_container-agent.yaml#L116-L119)). So this should be passed down to the cluster agent container as well. 
To achieve this we can move the env var to [this](https://github.com/DataDog/helm-charts/blob/6827c62cffb014492974c34b285f0df967446c9e/charts/datadog/templates/_components-common-env.yaml) template to ensure it is included in all containers.

The problem with adding the configs `"admission_controller.inject_config.trace_agent_socket_filename"` and `"admission_controller.inject_config.dogstatsd_agent_socket_filename"` is that it duplicates the configuration; if the user changes only one of them without changing the other one, they will go out of sync, resulting in unexpected behavior.

michaelcretzman (approved):

adel121 (approved):

adel121 (approved):
Great work!!

comp/trace/config/setup.go
```diff
@@ -80,12 +81,19 @@ func prepareConfig(c corecompcfg.Component, tagger tagger.Component, ipc ipc.Com
...
 	cfg.GitCommit = version.Commit
-	cfg.ReceiverSocket = defaultReceiverSocket
 
 	// the core config can be assumed to already be set-up as it has been
 	// injected as a component dependency
 	// TODO: do not interface directly with pkg/config anywhere
+
 	coreConfigObject := c.Object()
+
+	if runtime.GOOS == "windows" || runtime.GOOS == "darwin" {
```
ichinaski: nit: I don't have a strong preference to decide whether to use build tags or `runtime.GOOS` conditional, but since we already have `config_.go` files for each major OS, maybe we can keep the os-specific differences there? perhaps just a static `const UDSSupport bool` would do it. Then on this file, just check if `UDSSupport` is `true` or `false`.

JatinSingh007: Yes got the idea making the changes

JatinSingh007: I don't know if I got the idea of UDSSupport bool, but I implemented a change with putting the default value for apm in the build files of the specific OS. Let me know if these changes are sufficient.

ichinaski: Nice, having the defaults on those files looks nice and consistent with other variables 👍 

I would then suggest dropping this conditional here, and just do:
```go
cfg.ReceiverSocket = coreConfigObject.GetString("apm_config.receiver_socket")
```

With the default values it won't populate it for `windows` and `darwin`, and if the customer misconfigures them on those platforms, they'll get an error message -- I think this is better and more explicit to show something is wrong.

JatinSingh007: ahh yes makes sense there is no need for checking runtime now thanks

comp/trace/config/setup.go
```diff
@@ -80,12 +80,15 @@ func prepareConfig(c corecompcfg.Component, tagger tagger.Component, ipc ipc.Com
 	cfg.DDAgentBin = defaultDDAgentBin
 	cfg.AgentVersion = version.AgentVersion
 	cfg.GitCommit = version.Commit
-	cfg.ReceiverSocket = defaultReceiverSocket
 
 	// the core config can be assumed to already be set-up as it has been
 	// injected as a component dependency
 	// TODO: do not interface directly with pkg/config anywhere
+
```
ichinaski:
```suggestion
```

comp/trace/config/setup.go
```diff
@@ -80,12 +80,15 @@ func prepareConfig(c corecompcfg.Component, tagger tagger.Component, ipc ipc.Com
...
 	coreConfigObject := c.Object()
+
+	cfg.ReceiverSocket = coreConfigObject.GetString("apm_config.receiver_socket")
```
ichinaski: this should go _below_ the `coreConfigObject == nil` check

ichinaski (approved):
One last comment, but leaving the PR approved.

PR #40772, opened 2025-09-09
Implement listmount

kovagsm:
### What does this PR do?

Implement the new syscalls `listmount` and `statmount` instead of procfs for the mount resolver.

### Motivation

To have better visibility of all the mountpoints we have, and speed up synchronization

Locally, deployed to several clusters

### DIFF
--- a/pkg/security/resolvers/mount/mountlister.go
+++ b/pkg/security/resolvers/mount/mountlister.go
@@ -3,6 +3,9 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2016-present Datadog, Inc.
 
+//go:build linux
+
+// Package mount holds mount related files
 package mount
 
 import (
--- a/pkg/security/resolvers/mount/mountlister.go
+++ b/pkg/security/resolvers/mount/mountlister.go
@@ -1,0 +1,297 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+package mount
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"golang.org/x/sys/unix"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strconv"
+	"syscall"
+	"unsafe"
+)
+
+// Adjust for your arch/kernel if needed (these match your current file)
+const (
+	SysStatmount = 457
+	SysListmount = 458
+
+	LSMTRoot = ^uint64(0)
+
+	StatmountSbBasic       = 0x00000001
+	StatmountMntBasic      = 0x00000002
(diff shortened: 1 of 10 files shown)


pkg/security/resolvers/mount/mountlister.go
```diff
@@ -0,0 +1,302 @@
...
+	"path/filepath"
+	"runtime"
+	"strconv"
+	"syscall"
+	"unsafe"
+)
+
+// Adjust for your arch/kernel if needed (these match your current file)
+const (
+	SysStatmount = 457
```
theop-dd: Instead of hardcoding this value, can we use something similar to `unix.SYS_STATMOUNT`?

pkg/security/resolvers/mount/mountlister.go
```diff
@@ -0,0 +1,302 @@
...
+	SysListmount = 458
```
theop-dd: Same here with `unix.SYS_LISTMOUNT`?

theop-dd (approved):

pkg/security/resolvers/mount/mountlister.go
kovagsm: Good catch! Yes I can. updated

pkg/security/resolvers/mount/mountlister.go
kovagsm: Also updated

pkg/security/resolvers/mount/resolver.go
```diff
@@ -126,6 +129,58 @@ func (mr *Resolver) IsMountIDValid(mountID uint32) (bool, error) {
...
+
+// HasListMount returns true if the kernel has the listmount() syscall, false otherwise
+func (mr *Resolver) HasListMount() bool {
+	_, _, errno := unix.Syscall(SysListmount, 0, 0, 0)
+	return errno != unix.ENOSYS
+}
+
+// SyncCacheFromListMount Snapshots the current mountpoints using the listmount api
+func (mr *Resolver) SyncCacheFromListMount() error {
+	mounts, err := GetAll("/proc")
```
YoannGh: We should use `kernel.ProcFSRoot()` here instead of hardcoding the "/proc/" path

pkg/security/resolvers/mount/resolver.go
```diff
@@ -126,6 +129,58 @@ func (mr *Resolver) IsMountIDValid(mountID uint32) (bool, error) {
...
+		RootStr:       root,
+		Origin:        model.MountOriginListmount,
+		Visible:       true,
+		Detached:      false,
+	}
+}
+
+// HasListMount returns true if the kernel has the listmount() syscall, false otherwise
+func (mr *Resolver) HasListMount() bool {
+	_, _, errno := unix.Syscall(SysListmount, 0, 0, 0)
```
YoannGh:
```suggestion
	_, _, errno := unix.Syscall(unix.SYS_LISTMOUNT, 0, 0, 0)
```

